PR #232, opened 2018-03-23
Fix: Don't reference preview before its loaded

### LINKED ISSUE #229: Box.Preview is accessed before script has finished loading
I'm running a simple test of the Box React elements and am encountering an issue when loading a Preview from the `ContentExplorer` component.

From testing, it seems the `load` event attached to the dynamically inserted `script` tag is being fired prior to the script adding the `Box` object to `window`. Because the `loadPreview` callback attempts to read the `Preview` property before it's attached to `window`, I encounter the following error: `Uncaught TypeError: Cannot read property 'Preview' of undefined`.

By moving the global access to after the call to `isPreviewLibraryLoaded`, I'm able to properly load a ContentPreview.

Currently, the callback is as follows:
```typescript
const { Preview } = global.Box;
const { token, collection, ...rest }: Props = this.props;

if (!this.isPreviewLibraryLoaded() || !file || !token || this.preview) {
    return;
}
```

Adjusting it to the following seems to alleviate the issue:
```typescript
const { token, collection, ...rest }: Props = this.props;

if (!this.isPreviewLibraryLoaded() || !file || !token || this.preview) {
    return;
}

const { Preview } = global.Box;
```

priyajeet:
Fixes #229

### DIFF
--- a/src/components/ContentPreview/ContentPreview.js
+++ b/src/components/ContentPreview/ContentPreview.js
@@ -343,13 +343,14 @@ class ContentPreview extends PureComponent<Props, State> {
      * @return {void}
      */
     loadPreview = (): void => {
-        const { Preview } = global.Box;
         const { token, collection, ...rest }: Props = this.props;
         const { file }: State = this.state;
 
         if (!this.isPreviewLibraryLoaded() || !file || !token || this.preview) {
             return;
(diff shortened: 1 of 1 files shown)


boxcla:

PR #235, opened 2018-03-26
Fix: Pass in correct options to preview to allow dl/print

jeremypress:


boxcla:

src/components/ContentPreview/ContentPreview.js
```diff
@@ -360,6 +362,7 @@ class ContentPreview extends PureComponent<Props, State> {
             header: 'none',
             skipServerUpdate: true,
             useHotkeys: false,
+            showDownload: canDownload,
```
jeremypress: @priyajeet print and download are coupled by the same option (showDownload) in preview, so passing in a separate option for print is not needed.

src/components/ContentPreview/ContentPreview.js
```diff
@@ -343,7 +345,7 @@ class ContentPreview extends PureComponent<Props, State> {
      * @return {void}
      */
     loadPreview = (): void => {
-        const { token, collection, ...rest }: Props = this.props;
+        const { token, collection, canDownload, ...rest }: Props = this.props;
```
priyajeet: For backwards compatibility, add `showDownload` in the line above, so that its not part of `rest`.
And below, pass in showDownload: canDownload || !!showDownload

src/components/ContentPreview/ContentPreview.js
priyajeet: Need to guard the download() and print() functions below too.
And then pass canDownload to the header to not show the buttons.

src/components/ContentPreview/ContentPreview.js
priyajeet: Actually might be better to make another local function in this file for now

```js
canDownload() {
    // showDownload is a prop that preview library uses and can be passed by the user
    const { showDownload, canDownload }: Props = this.props;
    return canDownload || !showDownload;
}
```



Then in the line above, use `lodash/omit` to get rid of things that we don't want sending to the preview library that the user passes in, because we are overriding those.

```js
this.preview.show(file.id, token, {
        ...omit(rest, [
                'collection',
                'showDownload',
                'skipServerUpdate',
                'header',
                'container',
                'useHotKeys'
            ])
}
```

PR #239, opened 2018-03-28
Fix: Do not lower case event.key

priyajeet:
Also fixes render method for sidebar
Updates box-react-ui


boxcla:

PR #241, opened 2018-03-28
New: Add item expiration to details sidebar

jeremypress:


boxcla:

src/components/ContentSidebar/SidebarNotices.js
```diff
@@ -7,46 +7,70 @@
...
 };
 
+/**
+ * Adds one minute to a ISO formatted date string
+ *
+ * @private
+ * @param {string} expiration - an item or shared link expiration
+ * @return {string} an ISO formatted date string
+ */
+function addMinuteToExpiration(expiration: string) {
```
priyajeet: Probably move this and https://github.com/box/box-ui-elements/blob/master/src/util/datetime.js

into

box-react-ui/blob/master/src/utils/datetime.js

And reference that

src/components/ContentSidebar/SidebarVersions.js
```diff
@@ -27,9 +26,9 @@ const SidebarVersions = ({
     }
 
     return (
-        <SidebarSection>
+        <div>
```
priyajeet: Need a wrapping div? Can we just return the component itself (and internally add block style to it if needed)

src/components/ContentSidebar/SidebarNotices.js
```diff
@@ -7,46 +7,70 @@
...
+    }
+
+    if (itemExpiration) {
+        itemExpiration = addMinuteToExpiration(itemExpiration);
     }
 
     return (
-        !!sharedLinkExpiration && (
-            <SidebarSection>
+        <div>
```
priyajeet: If div is not needed for layout, you can try https://reactjs.org/docs/fragments.html

src/components/ContentSidebar/SidebarNotices.js
jeremypress: I'm up for moving the datetime.js stuff. But this method is specific to adding 1 minute for item expiration. The DateTime utils will already have generic time adding.

jeremypress: Ported them over, will remove from here once that's merged

src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -71,8 +71,12 @@ const DetailsSidebar = ({
 
     return (
         <SidebarContent hasTitle={hasTitle} title={<FormattedMessage {...messages.sidebarDetailsTitle} />}>
-            {hasVersions && <SidebarVersions onVersionHistoryClick={onVersionHistoryClick} versions={versions} />}
-            {hasNotices && <SidebarNotices file={file} />}
+            {(hasVersions || hasNotices) && (
+                <SidebarSection>
```
jeremypress: @priyajeet if both of these children end up being null, we still get the styles associated with the sidebar section, so an empty div with padding gets rendered. What's the best way to prevent this without moving the logic checking if we have notices or versions here?

wenboyu2:

| Bundle        | Diff | Before | After |
| ------------- |:----:|:------:|:-----:|
| sidebar.js | 1 KB | 486 KB | 487 KB |
| tree.js | 0 B | 498 KB | 498 KB |
| explorer.js | 1 KB | 682 KB | 683 KB |
| picker.css | 0 B | 28.4 KB | 28.4 KB |
| picker.js | 0 B | 599 KB | 599 KB |
| uploader.js | 0 B | 501 KB | 501 KB |
| preview.js | 0 B | 547 KB | 547 KB |
| uploader.css | 0 B | 18.7 KB | 18.7 KB |
| explorer.css | 0 B | 38.8 KB | 38.8 KB |
| tree.css | 0 B | 10.6 KB | 10.6 KB |
| preview.css | 0 B | 25.8 KB | 25.8 KB |
| sidebar.css | 0 B | 24.6 KB | 24.6 KB |
| Total | 2 KB | 3.38 MB | 3.38 MB |

Compared against previously merged PR #251

src/components/ContentSidebar/SidebarContent.scss
```diff
@@ -17,7 +17,7 @@
         bottom: 0;
         left: 0;
         overflow: auto;
-        padding-top: 20px;
+        padding-top: 10px;
```
jeremypress: changed to match legacy sidebar padding

src/components/ContentSidebar/__tests__/__snapshots__/SidebarVersions-test.js.snap
```diff
@@ -53,60 +53,28 @@ ShallowWrapper {
...
-        "props": Object {
-          "className": "",
-          "onClick": undefined,
-          "versionCount": 2,
-        },
-        "ref": null,
-        "rendered": null,
-        "type": [Function],
-      },
+      "rendered": null,
```
jeremypress: @DanDeMicco not sure why these are rendering null now, I just removed the <SidebarSection>

src/components/ContentSidebar/SidebarNotices.js
```diff
@@ -4,49 +4,73 @@
...
+ *
+ * @private
+ * @param {string} expiration - an item or shared link expiration
+ * @return {string} an ISO formatted date string
+ */
+export function addMinuteToExpiration(expiration: string) {
+    let expirationDate = new Date(expiration);
+    // One minute is added to account for dates set via a date picker.
+    // These dates will actually be stored as 11:59PM the night before the item expires.
+    expirationDate = addTime(expirationDate, ONE_MINUTE_IN_MS);
```
priyajeet: Any reason we need to new Date() above, as well as do .toISOString()?
Datefield should be localization formatting.

Does adding this inline, make the code more un-readable? Because it can then remove the if() blocks.

```js
<DateField date={addTime(expirationDate, ONE_MINUTE_IN_MS)} dateFormat={NOTICE_DATE_FORMAT} relative={false} />
```

src/components/ContentSidebar/SidebarNotices.js
```diff
@@ -4,49 +4,73 @@
  * @author Box
  */
 
-import React from 'react';
+import React, { Fragment } from 'react';
```
priyajeet: Just do `import * as React from 'react';` and then use <React.Fragment>
New-er adopted style.

src/components/ContentSidebar/DetailsSidebar.scss
```diff
@@ -8,3 +8,11 @@
         margin-bottom: 0;
     }
 }
+
+.bcs-details-content {
+    margin: 0 20px 20px 30px;
+}
+
+.bcs-details-content:empty {
```
priyajeet: Just double check if IE11 is fine with this.

src/components/ContentSidebar/SidebarNotices.js
jeremypress: addTime takes in either a date or a timestamp. The API returns an ISO formatted string, so we need to at least convert it to a Date. I can remove the toISO string part, not needed.

src/components/ContentSidebar/DetailsSidebar.scss
jeremypress: Since IE9 https://developer.mozilla.org/en-US/docs/Web/CSS/:empty

Issue #183, opened 2018-02-18
Content Explorer popup menu appears in wrong location


scherroman:
### Steps
I am using the [Box Content Explorer](https://developer.box.com/docs/box-content-explorer) React element in a [Semantic UI React Modal](https://react.semantic-ui.com/modules/modal) element.

Using [box-ui-elements 3.6.1](https://github.com/box/box-ui-elements) and [semantic-ui-react 0.78.2](https://github.com/Semantic-Org/Semantic-UI-React).

### Expected Behavior
The triple dot popup menu appears directly below the cursor after clicking the triple dot for a file.

### Actual Behavior
The triple dot popup menu appears far below and off to the side of the cursor.

![print marker file explorer random menu location](https://user-images.githubusercontent.com/7923938/36226936-58d3f6d0-119d-11e8-9fb0-38119d3b3be2.JPG)

The larger the screen size, the further below and off to the side the popup appears.

### Test Case

[Here's a Codepen showing this](https://codepen.io/scherroman/pen/YeeyYL), using Box UI Elements 3.4.1 explorer.js  and Semantic UI 2.2.14 .js

tonyjin:
Thanks for the report, we'll take a look.

priyajeet:
This will be a tricky one to fix as one solution may break another use case. Unfortunately don't really have a proper solution for this at the moment. We use Tether.js/react-tether which is sensitive to the stacking context and is not perfect. Fixed/absolute positioning by other 3rd party modals may interfere with it. We have experimented with popper.js as a replacement, which has its own problems. Will need some time to investigate.

priyajeet:
Should be fixed in version 4.4.0.
However you will have to add a z-index (maybe a 1000 for your example) to `.dropdown-menu-element` for the menu to show above your modal.

PR #243, opened 2018-03-30
Breaking: Render dropdowns and tooltips popups into body

priyajeet:
Previously they used to render into the UI Element's container element. This has been causing issues when the UI Element was rendered in some 3rd party library modal (#183) due to stacking context issues (absolute/fixed positioning) and compatibility with tether.js positioning via bodyElement.

This change removes use of bodyElement making tether render the popups directly into the document body. While this breaks a bit of encapsulation, at least now tether should be able to position them better.

However the tooltip or dropdown menu can be hidden due to z-index issues. In that case, the parent app, folks using the UI Element, need to override and add z-index based on their app's use case. 

Classes to add z-index to would be `dropdown-menu-element` and `tooltip-element`. And z-index only needs to be added/overridden if needed.


boxcla:

PR #244, opened 2018-03-30
Fix: Only focus preview when auto focus is allowed

priyajeet:
So that the preview is not focusing itself and making the parent page jump. Auto focus helps with keyboard shortcuts. Autofocus is set to true when using content explorer.

Fixes typo in options to prevent preview library attach its own hotkey handlers.

Also fixes references to React 16 in test files. UI Elements aren't compatible with React 15.


boxcla:

PR #246, opened 2018-03-30
Chore: Respect autoFocus and useHotkeys properties

priyajeet:
autoFocus is false by default, not necessary for hotkeys to work, but helpful on 1st load as hot keys only work when preview has focus.
useHotkeys is true by default.


boxcla:

PR #248, opened 2018-03-30
Fix: IE11 issue with left navigation icon in preview

priyajeet:


boxcla:

src/components/ContentPreview/ContentPreview.scss
```diff
@@ -30,6 +30,10 @@
         right: 0;
     }
 
+    .bcpr-navigate-left {
```
pramodsum: Do we need to match the border radius similar to the navigate right?

priyajeet: I should just remove that, not sure whats its doing

PR #256, opened 2018-04-05
Chore: Add props for metric and error callbacks

JustinHoldstock:
@priyajeet Does this look like correct changes to accommodate handling preview errors and metrics? Do I need to make changes to the ES6 Wrapper, and if so, should those just be class properties that are passed, similar to `onInteraction()` but set as `noop` for the time being?

TODO: Tests!

PR #270, opened 2018-04-18
Update: Modify token service to account for multiple tokens

priyajeet:
Can return a read and write token, preview should only be handed the read token.
Token generator function should either return a simple string token, or should return an object that may have read and write token.
When requesting multiple tokens per file, token generator should return a structure like

{
  file_1: {
    read: 'token',
    write: 'token'
  },
  file_2: {
    read: 'token',
    write: 'token'
  },
  ...
}


boxcla:

boxcla:

Compared against previously merged PR #256

src/util/TokenService.js
```diff
@@ -48,7 +48,7 @@ class TokenService {
      * @param {string} tokenOrTokenFunction - Optional token or token function
      * @return {Promise} that resolves to a token
      */
-    static getToken(id: string, tokenOrTokenFunction: Token): Promise<?string> {
+    static async getToken(id: string, tokenOrTokenFunction: Token): Promise<?string> {
         // Make sure we are getting typed ids
         // Tokens should either be null or undefuned or string or functions
```
jeremypress: undefined

src/util/TokenService.js
```diff
@@ -48,7 +48,7 @@ class TokenService {
      * @param {string} tokenOrTokenFunction - Optional token or token function
      * @return {Promise} that resolves to a token
      */
-    static getToken(id: string, tokenOrTokenFunction: Token): Promise<?string> {
+    static async getToken(id: string, tokenOrTokenFunction: Token): Promise<?string> {
```
jeremypress: Doesn't look like this function accepts an array of IDs anymore, can you update the jsdoc?

src/util/TokenService.js
```diff
@@ -59,27 +59,24 @@ class TokenService {
...
-                resolve(token);
-            } else if (typeof token === 'object') {
-                resolve(token[id]);
-            } else {
-                reject(error);
-            }
-        });
+        // that on resolution returns a token.
+        const token = await tokenOrTokenFunction(id);
+        if (!token || typeof token === 'string') {
```
jeremypress: Any reason why this is split between two cases that return the same thing? If you want to keep it split, it might make more sense to have the undefined/null case, and then a case where the token exists, either being a string or a well formed object.

src/util/TokenService.js
```diff
@@ -105,25 +102,56 @@ class TokenService {
...
-            // $FlowFixMe Already checked null above
-            const token = await tokenOrTokenFunction(ids);
-            if (!token || typeof token === 'object' || typeof token === 'string') {
-                resolve(createIdTokenMap(ids, token));
-            } else {
-                reject(error);
-            }
-        });
+        const token = await tokenOrTokenFunction(ids);
+        if (!token || typeof token === 'object' || typeof token === 'string') {
```
jeremypress: Why do we have to support a single string token coming back here? Shouldn't `getToken` be called in that case?

priyajeet: I am gonna nuke bunch of the code here, not really needed

src/util/TokenService.js
priyajeet: I'll combine them

src/util/TokenService.js
priyajeet: fixed

PR #271, opened 2018-04-18
New: box-react-ui upgrade script

pramodsum:
- Automates upgrading of box-react-ui inside the repository

### DIFF
--- a/build/upgrade_brui.sh
+++ b/build/upgrade_brui.sh
@@ -1,0 +1,107 @@
+#!/bin/bash
+
+export NODE_PATH=$NODE_PATH:./node_modules
+
+update_version() {
+    CURRENT_VERSION="$(cat package.json | grep \"box-react-ui\"\: -m 1 | sed 's/.*\^\(.*\)\".*/\1/')";
+    LATEST_VERSION="$(npm show box-react-ui version)";
+
+    # Upgrade peer dependency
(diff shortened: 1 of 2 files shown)


boxcla:

Compared against previously merged PR #256

pramodsum:
This script won't merge the change automatically. It will create a commit and push it your fork. You still have to manually create a PR and merge that in

priyajeet:
I don't think we need any script for this. Since this should ideally be done as part of any other commit that requires an upgrade. We will not be doing any upgrades on its own.

jeremypress:
@priyajeet I agree that it shouldn't be a part of the release process, but it doesn't hurt to have a script that makes upgrading BRUI easier.

pramodsum:
I didn't include this as a part of the release process. It's more an optional script to run if we need to upgrade BRUI

priyajeet:
I do not want additional scripts added unless they are really necessary. This one doesn't seem necessary as the same can be accomplished by just running `yarn upgrade-interactive --latest` or `yarn upgrade box-react-ui` or `yarn add box-react-ui@VERSION --peer/dev`. If at all, you should just have the command directly inside package.json scripts section without a shell script.

Should not be adding unnecessary things to maintain. And box-react-ui is a 3rd party dep like React or React-dom etc. We aren't adding scripts for them, so should not add for box-react-ui either.

PR #258, opened 2018-04-10
Update: access stats API integration

DanDeMicco:
This commit contains the access stats API integration. It also contains a minor change to versions which will make it not render if the file is a box note.

PR #283, opened 2018-04-27
New: version and tasks API

DanDeMicco:
This adds the version and task API's which will be used in activity feed. It also adds some API methods and a new class which help reduce boilerplate in the API files.


boxcla:

boxcla:

Compared against previously merged PR #258

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -502,6 +660,10 @@ class ContentSidebar extends PureComponent<Props, State> {
                                 accessStatsError={accessStatsError}
                                 fileError={fileError}
                                 versionError={versionError}
+                                tasks={tasks}
```
priyajeet: At some point we should consider collapsing sections feature wise if we see too many props per feature. Something like what the activity feed already does for its own component

So something like
```jsx
tasks = {{
    tasks: tasks,
    error: tasksError,
    onCreate: onTaskCreate,
    ...
}}
```

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -182,16 +186,40 @@ class ContentSidebar extends PureComponent<Props, State> {
...
+        if (fileId) {
+            this.fetchFile(fileId);
+            if (hasAccessStats) {
+                this.fetchFileAccessStats(fileId);
+            }
+            if (hasActivityFeed) {
+                this.fetchComments(fileId, false);
+                this.fetchTasks(fileId);
+                this.fetchVersions(fileId, false);
+            } else if (hasVersions) {
```
priyajeet: I am guessing `else if` because it only shows if there is no activity feed?

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -182,16 +186,40 @@ class ContentSidebar extends PureComponent<Props, State> {
...
+    }
+
+    /**
+     * Fetches the data for the sidebar
+     *
+     * @param {Object} Props the component props
+     * @param {boolean} hasFileIdChanged true if the file id has changed
+     */
+    fetchData({ fileId, hasVersions, hasActivityFeed, hasAccessStats }: Props, hasFileIdChanged?: boolean) {
+        if (hasFileIdChanged) {
```
priyajeet: This is not needed for componentDidMount(), so should move it inside componentWillMount()
I am assuming you are adding this for the loading indicator to show.

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -182,16 +186,40 @@ class ContentSidebar extends PureComponent<Props, State> {
...
 
         const hasTokenChanged = newToken !== token;
         const hasFileIdChanged = newFileId !== fileId;
         const currentFileId = newFileId || fileId;
 
         if (currentFileId && (hasTokenChanged || hasFileIdChanged)) {
-            this.fetchFile(currentFileId);
-            if (hasVersions) {
-                this.fetchVersions(currentFileId);
+            this.fetchData(nextProps, hasFileIdChanged);
```
priyajeet: currentFileId may not be needed with this change. Maybe just check newFileId

src/api/Base.js
```diff
@@ -123,6 +124,75 @@ class Base {
...
         return this.cache;
     }
+
+    /**
+     * Generic success handler
+     *
+     * @param {Object} data the response data
+     * @param {Function} successCallback the success callback
+     */
+    successHandler(data: Object, successCallback: Function) {
```
priyajeet: Few file already set successCallback/errorCallback on the `this` object, so I wonder maybe one can make this more generic, with the args just callback. Because I am assuming this is just to simplify the destroyed check.

```
successHandler(callback: Function) {
   if (!this.isDestroyed()) {
      callback();
    }
}
```


// Item.js
rename(...successCallback, errorCallback,...) {
        ...

        return this.xhr
            .put({ url: `${this.getUrl(id)}`, data: { name } })
            .then(this.successHandler(this.renameSuccessHandler))
            .catch(this.errorHandler);
}

src/api/Base.js
```diff
@@ -123,6 +124,75 @@ class Base {
...
+        }
+    }
+
+    /**
+     * Generic error handler
+     *
+     * @param {Object} data the response data
+     * @param {Function} errorCallback the error callback
+     */
+    errorHandler(error: Object, errorCallback: Function) {
```
priyajeet: There is an errorHandler already in one of the other files I think. Probably should consolidate.

src/api/APIFactory.js
```diff
@@ -264,6 +276,34 @@ class APIFactory {
     }
 
     /**
+     * API for comments
+     *
+     * @param {boolean} shouldDestroy - true if the factory should destroy before returning the call
+     * @return {CommentsAPI} CommentsAPI instance
+     */
+    getCommentsAPI(shouldDestroy: boolean): CommentsAPI {
+        if (shouldDestroy) {
```
priyajeet: Not sure if we have a use case yet, but I am thinking we may have to extend the destroying part to destroy itself only. In other words, getting comments should destroy only getting prior comments but not getting tasks.

This all works fine right now, since the `file` fetching API destroys everything, so getting comments for one file and comments for next file don't overlap.

priyajeet: Or maybe we will never have use case for smaller APIs to destroy anything, and only the File/Folder api should destroy since they logically represent working on a brand new file or folder which will destroy all other sub apis.

src/components/ContentSidebar/ContentSidebar.js
DanDeMicco: Agreed, the list is getting quite long

src/components/ContentSidebar/ContentSidebar.js
DanDeMicco: Actually just caught a mistake there that my code doesnt match my comment. 

My thought was that we only need to fetch all the versions if there is an activity feed, else we can just do 1 call (potentially with a limit 1 or something), to render the version component in the details tab. I am now looking to see if this can be solved by adding a new field to the get file info call, then this else if can be avoided entirely.

src/components/ContentSidebar/ContentSidebar.js
DanDeMicco: Did you mean in ```componentWillReceiveProps``` ? I added this so that the sidebar wouldn't have stale data for things such as comments when the current fileId changes

src/api/Base.js
DanDeMicco: Looks like Item has one. Will consolidate them

src/api/APIFactory.js
DanDeMicco: Yeah I think that makes sense. Basically just destroy the old API, and not all of them. For now, I dont think there should be any issues but should think about the scenarios of if individual API's will have to be destroyed. The only one I can think of is collaborations for now

DanDeMicco:
Pushed some changes

src/api/APIFactory.js
```diff
@@ -131,6 +143,14 @@ class APIFactory {
             this.fileAccessStatsAPI.destroy();
             delete this.fileAccessStatsAPI;
         }
+        if (this.tasksAPI) {
```
jeremypress: This is getting pretty verbose/repeatable, any ideas on how we could refactor?

src/api/Base.js
```diff
@@ -123,6 +129,77 @@ class Base {
...
+    }
+
+    /**
+     * Generic error handler
+     *
+     * @param {Object} data the response data
+     * @param {Function} errorCallback the error callback
+     */
+    errorHandler(error: Object): void {
+        if (!this.isDestroyed() && this.errorCallback) {
```
jeremypress: Might as well check if it's a function if you are checking for it's existence?

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -66,8 +75,12 @@ type State = {
     file?: BoxItem,
     accessStats?: FileAccessStats,
     versions?: FileVersions,
+    comments?: Comments,
```
jeremypress: It looks like we're always setting it but it may be undefined. would that be `comments: ?Comments`

src/api/Base.js
DanDeMicco: I think flow will take care of this

src/components/ContentSidebar/ContentSidebar.js
DanDeMicco: I think the way I have it is correct. It can possibly be undefined, or not be in the state at all. See https://flow.org/en/docs/types/primitives/#toc-optional-object-properties

src/components/ContentSidebar/ContentSidebar.js
DanDeMicco: Looks like I can just use the ```version_number``` field to render sidebarVersions, so adding that field and removing the else if

src/api/APIFactory.js
priyajeet: I think the solution can be just aborting the request since axios allows us to do that.
https://github.com/axios/axios#cancellation

Before it wasn't possible, but now one can use that.

src/api/Base.js
priyajeet: Only statically, probably should add this https://github.com/brigand/babel-plugin-flow-react-proptypes

src/api/OffsetBasedAPI.js
```diff
@@ -0,0 +1,105 @@
...
+                ...data,
+                entries: entries.concat(data.entries)
+            };
+            const totalCount = data.total_count;
+
+            if (shouldFetchAll && this.hasMoreItems(newOffset, totalCount)) {
+                this.get(id, successCallback, errorCallback, newOffset, limit, fields, shouldFetchAll);
+            }
+
+            this.successHandler(this.data, successCallback);
```
priyajeet: Do you think for consistency we should make this also
this.successCallback  = successCallback
and don't pass it around like error

src/api/OffsetBasedAPI.js
```diff
@@ -0,0 +1,105 @@
...
+        return queryParams;
+    }
+
+    /**
+     * Determines if the API has more items to fetch
+     * @param {number} offset the offset from the start to start fetching at
+     * @param {number} totalCount the total number of items
+     * @return {boolean} true if there are more items
+     */
+    hasMoreItems(offset: number, totalCount?: number): boolean {
```
priyajeet: Double check if you aren't falling for the same gotcha about total count issue I did that caused this patchset
https://github.com/box/box-ui-elements/commit/2980e0ca42e08194c3c3ab51c4268a775cf6e8cf

src/api/APIFactory.js
DanDeMicco: Was just looking at that earlier. I can create another ticket for that, but I think its out of scope for this change

src/api/Base.js
DanDeMicco: Will add the function type check

src/api/OffsetBasedAPI.js
DanDeMicco: Good point, will do that

src/api/OffsetBasedAPI.js
DanDeMicco: Based on my understanding, I think my implementation is correct here

src/api/APIFactory.js
priyajeet: can also just move the objects into another object.

this.apis = { file: fileAPI, ... }

in that way you can just iterate over the object.keys and destroy them all

src/api/Base.js
```diff
@@ -123,6 +134,78 @@ class Base {
...
     }
+
+    /**
+     * Generic success handler
+     *
+     * @param {Object} data the response data
+     * @param {Function} successCallback the success callback
+     */
+    successHandler(data: Object): void {
+        if (!this.isDestroyed() && this.successCallback && typeof this.successCallback === 'function') {
```
priyajeet: Middle check unnecessary
You can also bypass the checks by defaulting to lodash/noop.

src/api/Base.js
```diff
@@ -123,6 +134,78 @@ class Base {
...
+    }
+
+    /**
+     * Generic API get
+     *
+     * @param {string} id the file id
+     * @param {Function} successCallback the success callback
+     * @param {Function} errorCallback the error callback
+     */
+    async get(id: string, successCallback: Function, errorCallback: Function): Promise<void> {
```
priyajeet: Doesn't matter here, but probably not use keywords. get / set are js keywords.

src/api/Base.js
```diff
@@ -123,6 +134,78 @@ class Base {
...
+
+    /**
+     * Generic error handler
+     *
+     * @param {Object} data the response data
+     * @param {Function} errorCallback the error callback
+     */
+    errorHandler(error: Object): void {
+        if (!this.isDestroyed() && this.errorCallback && typeof this.errorCallback === 'function') {
+            this.errorCallback(error);
```
priyajeet: The signature of this is different than what it was in Item.js
error in the args, I think it will be a Error event.
This event will have the response, which in turn will have the data.
I think you want to send that data in the errorCallback

src/api/OffsetBasedAPI.js
```diff
@@ -0,0 +1,106 @@
...
+
+            const entries = this.data ? this.data.entries : [];
+            this.data = {
+                ...data,
+                entries: entries.concat(data.entries)
+            };
+            const totalCount = data.total_count;
+
+            if (shouldFetchAll && this.hasMoreItems(newOffset, totalCount)) {
+                this.get(id, successCallback, errorCallback, newOffset, limit, fields, shouldFetchAll);
```
priyajeet: need to pass in success/error callback if its already on the `this` object?

src/api/Base.js
DanDeMicco: Hmm yeah there are get and set methods which you can have in classes. I wasnt able to find any documentation for using get and set as method names, nor does eslint complain. For future reference, can you point me to the documentation where you saw this?

src/api/Base.js
DanDeMicco: Will remove

src/api/Base.js
DanDeMicco: I think I removed it because I didnt see any error handling other than a console.log. Also, returning the raw error object could potentially be more helpful for debugging, since I assume console messages will be stripped in production bundle

src/api/OffsetBasedAPI.js
DanDeMicco: In this case, yeah since its a recursive call and it sets the this variables in the function. One way to get rid of the args I could create a helper method which takes all the args, and create a helper function without the callbacks

DanDeMicco: Refactored this method a bit, ended up renaming it to offsetGet and using a helper method

src/api/Base.js
DanDeMicco: Changed this to be like the logic in Item.js

src/api/APIFactory.js
DanDeMicco: I added a ticket for this so we can get back to it later

PR #284, opened 2018-05-01
Chore: Adding general and details tab interaction targets

jeremypress:

PR #288, opened 2018-05-02
Breaking: Introduce contentSidebarProps to pass to ContentSidebar

jeremypress:
@priyajeet, leaving hasSidebar because without it, in order to show the sidebar you would have to pass in an empty `contentSidebarProps` object. With this change that boolean stays as is, and then any additional props will get passed to the sidebar (except for `showSidebar`).


boxcla:

boxcla:

Compared against previously merged PR #284

src/components/ContentPreview/ContentPreview.js
```diff
@@ -713,7 +718,8 @@ class ContentPreview extends PureComponent<Props, State> {
             language,
             messages,
             className,
-            showSidebar,
+            contentSidebarProps,
+            contentSidebarProps: { showSidebar },
```
DanDeMicco: I dont think this should be here

src/components/ContentPreview/ContentPreview.js
```diff
@@ -46,7 +46,9 @@ type Props = {
     isSmall: boolean,
     autoFocus: boolean,
     useHotkeys: boolean,
-    showSidebar?: boolean,
+    contentSidebarProps: {
```
DanDeMicco: Why is this a mandatory field? Can it not be a defaultProp {}?

src/components/ContentPreview/ContentPreview.js
jeremypress: eslint errors that i'm not using de-structuring if I don't include both

src/components/ContentPreview/ContentPreview.js
jeremypress: nope!

src/components/ContentPreview/ContentPreview.js
jeremypress: I need to access both contentSidebarProps and contentSidebarProps.showSidebar

DanDeMicco: 👍 I assumed it was a typeo, but turns out this is how you can do nested destructuring. TIL.

IMO, ```const {showSidebar} = contentSidebarProps``` on the next line is less confusing, but am fine with either way

priyajeet: yeah this looks weird, just put it on a new line

src/components/ContentPreview/ContentPreview.js
```diff
@@ -113,7 +115,10 @@ class ContentPreview extends PureComponent<Props, State> {
         onLoad: noop,
         onMetric: noop,
         onNavigate: noop,
-        collection: []
+        collection: [],
+        contentSidebarProps: {
```
priyajeet: PreviewDialog.js needs to change

priyajeet:
Probably move hasSidebar inside then.
And perhaps rename showSidebar to isSidebarVisible.
(for consistency booleans starting with is/has/can).

```jsx
<ContentExplorer
    contentPreviewProps={{
        contentSidebarProps: {
            hasSidebar: boolean,
            isSidebarVisible: boolean
        }
    }}
/>

<ContentPreview
    {...contentPreviewProps}
/>
```

```jsx
<ContentPreview
    contentSidebarProps={{
        hasSidebar: boolean,
        isSidebarVisible: boolean
    }}
/>

<ContentSidebar
    {...contentSidebarProps}
/>
```

```jsx
<ContentSidebar
    isSidebarVisible
/>
```

priyajeet:
Actually I back tracked, keep it out, see updated comment above

jeremypress:
^ Those all make sense except for the last one. Why would we need to pass isVisible to the sidebar directly if we're rendering it directly? Wouldn't that be controlled by the parent app?

src/components/ContentPreview/ContentPreview.js
jeremypress: even if it only passes in `hasSidebar` currently? In the future when the explorer supports passing in props through the sidebar then it will have to forward them as well.

priyajeet: nvm, I thought showSidebar was being passed from it

PR #297, opened 2018-05-08
Fix: ContentUploader file input should reset properly

wenboyu2:

PR #298, opened 2018-05-08
Fix: add error and success interceptors for requests and responses

jeremypress:


boxcla:

jeremypress:
Will need to update dev docs and codepen once we cut our next release.

src/util/Xhr.js
```diff
@@ -58,13 +65,8 @@ class Xhr {
         this.sharedLinkPassword = sharedLinkPassword;
         this.axios = axios.create();
 
-        if (typeof responseInterceptor === 'function') {
```
jeremypress: @priyajeet axios will do nothing if you pass in null or a non function, so i'm avoiding checking the type of 4 things.

boxcla:

Compared against previously merged PR #297

src/util/Xhr.js
```diff
@@ -48,7 +53,9 @@ class Xhr {
         sharedLink,
         sharedLinkPassword,
         responseInterceptor,
-        requestInterceptor
+        responseErrorInterceptor,
+        requestInterceptor,
```
priyajeet: Why did we decide to split the interceptors?

jeremypress:
They serve different functions. The normal interceptor allows you to modify the response, while the error interceptor allows you react to errors

priyajeet:
I don't think we should be over complicating the interface.
I think this is acceptable. Based on what kind of errors the API responds with, people are likely are not going to muck around with the data and if they do, they will just want to log or map box API response to one of their error.

```js
errorHandler = (data) => {
    // Do something with the data like logging
    return Promise.reject(data);
}

responseInterceptor = (response) => {
    if (response instanceof Error) {
        return errorHandler(response.data)
    } else {
        // do something with response
        return response;
    }
}

element.show(id, token, {
    responseInterceptor
})
```

priyajeet:
Actually a better and simpler implementation will be:


```js
// Client code
interceptor(data: Object, error?: Error) {
 // do something with data
 return data;
}

element.show(id, token, { responseInterceptor: interceptor })

// Our xhr.js
successInterceptor = interceptor;
errorInterceptor = (error) => { return Promise.reject(interceptor(error.response.data, error)) }
this.axios.interceptors.response.use(successInterceptor, errorInterceptor);
```

jeremypress:
That sounds fine. I'll update my change but we should wait for your API change so I can fully test that the object is easy to work with.

priyajeet:
ok I can merge those, since designs changed, i'll just do it in another commit.

src/util/Xhr.js
```diff
@@ -59,7 +59,11 @@ class Xhr {
         this.axios = axios.create();
 
         if (typeof responseInterceptor === 'function') {
-            this.axios.interceptors.response.use(responseInterceptor);
+            const successInterceptor = responseInterceptor;
+            // Called on any non 2xx response
+            const errorInterceptor: Function = (error: { response: Object }) =>
```
priyajeet: I would probably guard and check if there is even a response on the error object. Since in their docs https://github.com/axios/axios#handling-errors they do check if `error.reponse` exists.

BTW that section also shows what could be a request error interceptor use case, aka the request was never made as caught by the lack of `error.response` but existence of `error.request`. Though still not sure for us when that would be the case.

Also, thinking more about it. trying to think what exactly the error interceptor should be returning in its rejected promise to our code after interception has done its job. Maybe we should just return the Error object with new data that may have been modified in the interception to keep the same return if there was no interception.

```js
const errorInterceptor = (error: $AxiosError<any>) => {
    const { response = { data = {} } } = error;
    const { data = {} } = response;
    error.response.data = responseInterceptor(data, error); // the interceptor will modify the data
    return Promise.reject(error); // return the error that came in but with modified data
}
```

In our code, it will hit https://github.com/box/box-ui-elements/blob/master/src/api/Base.js#L157
When there is an error but no interceptor, then the error object is sent in there as-is.
When there is an interceptor, then error object is sent in there with possibility intercepted data.

wdyt?

priyajeet: btw can move the function out to the class level too rather having it nested

jeremypress: Is there any value to letting developers manipulate the error object before we reject with it? From my understanding people want to intercept errors (read only) so they can do custom stuff based on the status code, but the elements aren't built to render differently if that object is modified right? It might be useful if we supported a custom error message passed into that object.

If we want to allow them to modify the data, we need to document that they have to return something in the error case as well. Or, we could see what their interceptor returns, and if it's undefined we just pass in the unmodified error object. This is starting to feel like a pre-optimization to me.

priyajeet:
Still then just reject the promise with the error object as above, while ignoring the return value from the interceptor. Won't be a breaking change in that case.

src/util/Xhr.js
```diff
@@ -56,21 +57,38 @@ class Xhr {
...
         }
     }
 
+    /**
+     * Error interceptor that wraps the passed in responseInterceptor
+     *
+     * @param {Object} error - Error object from axios
+     * @return {Promise} rejected promise with error info
+     */
+    errorInterceptor(error: Object): Promise<any> {
```
priyajeet: => fat arrow so you dont have to bind.
Also, didnt you want to return `Promise.reject(error)` per your prior comment so that you dont want to send our code a potentially modified error object

PR #299, opened 2018-05-08
Fix: flow error in UploadStateContent

wenboyu2:

PR #301, opened 2018-05-09
Update: Pass preview instance to ContentPreview

tonyjin:


boxcla:
Verified that @tonyjin has signed the CLA. Thanks for the pull request!

boxcla:

Compared against previously merged PR #299

src/components/ContentPreview/ContentPreview.js
```diff
@@ -43,6 +43,7 @@ import '../base.scss';
 import './ContentPreview.scss';
 
 type Props = {
+    previewInstance?: any, // Allow instance of Preview library to be passed in
```
priyajeet: I probably wont add it as props to keep the interface clean since this is more of a internal webapp use case for now. Just set it directly via componentRef. Maybe add a function in this file called `setPreviewInstance(preview) { this.preview = preview }`

src/components/ContentPreview/ContentPreview.js
```diff
@@ -409,7 +417,8 @@ class ContentPreview extends PureComponent<Props, State> {
             useHotkeys: false
         };
 
-        this.preview = new Preview();
+        // Use instance of Preview from props if provided
+        this.preview = previewInstance || new Preview();
```
priyajeet: Don't think it should be here, because you only want it set the 1st initial time. Every other time it should create a new instance when things like navigation happen.

priyajeet: I think you may have to extract out the lines below, including listeners, in its own function like `createPreviewInstance(preview) { ... }` where it can be called from the constructor if it has been primed already.

src/components/ContentPreview/ContentPreview.js
```diff
@@ -378,10 +411,10 @@ class ContentPreview extends PureComponent<Props, State> {
      * @return {void}
      */
     loadPreview = async (): Promise<void> => {
-        const { token: tokenOrTokenFunction, onError, onMetric, collection, ...rest }: Props = this.props;
+        const { token: tokenOrTokenFunction, collection, ...rest }: Props = this.props;
         const { file }: State = this.state;
 
-        if (!this.isPreviewLibraryLoaded() || !file || !tokenOrTokenFunction || this.preview) {
```
priyajeet: Removing this guard, can cause the preview to re-show the preview when anything else changes in the component like the sidebar or something. Since loadPreview is called from multiple places, the idea is that it should only re-instantiate preview ---> add listeners ---> call show, if and only if the prior preview instance was destroyed, which only happens when file id changes.

You will have to test.

src/components/ContentPreview/ContentPreview.js
```diff
@@ -363,25 +368,58 @@ class ContentPreview extends PureComponent<Props, State> {
...
+     * @return {boolean}
+     */
+    canDownload(): boolean {
         // showDownload is a prop that preview library uses and can be passed by the user
         const { showDownload, canDownload }: Props = this.props;
         const { file }: State = this.state;
         const isFileDownloadable =
             getProp(file, 'permissions.can_download', false) && getProp(file, 'is_download_available', false);
-        return isFileDownloadable && canDownload && showDownload;
+        return !!isFileDownloadable && !!canDownload && !!showDownload;
```
priyajeet: isFileDownloadable is defaulted to boolean

src/components/ContentPreview/ContentPreview.js
```diff
@@ -196,13 +195,19 @@ class ContentPreview extends PureComponent<Props, State> {
...
-     * Called after shell re-mounts
+     * After component updates, check if file ID or token has changed and if so, load Preview again.
      *
      * @private
      * @return {void}
      */
-    componentDidUpdate(): void {
-        this.loadPreview();
+    componentDidUpdate(prevProps: Props): void {
+        const { fileId, token }: Props = this.props;
```
priyajeet: Why not just check it this.preview exists?
Because componentWillRecvProps will just nuke it with the same conditions as below?

tonyjin: Ah this is annoying. You're right based on this code I can just check whether or not this.preview exists, but that reminds me that Instant Preview depends on the file being cached in the preview library's cache. Just means I need to either need to warm up not only ContentPreview's cache but also the preview library's cache OR on hover, call `updateFileCache` on that single file before calling `prefetch()`...

PR #302, opened 2018-05-09
New: Populate @mentions user list in AF

pramodsum:
- Allow canceling of axios API requests
- Add framework for marker-based paging of API requests
- Chore: Remove collaborator query override


boxcla:

src/api/MarkerBasedAPI.js
```diff
@@ -0,0 +1,147 @@
...
+            if (this.shouldFetchAll && hasMoreItems) {
+                return this.markerGetRequest();
+            }
+
+            this.successHandler(this.data);
+        } catch (error) {
+            this.errorHandler(error);
+        }
+
+        return Promise.resolve();
```
priyajeet: No return needed either.
I see similar being done in OffSetBasedApi.
Can remove from there too.

src/api/MarkerBasedAPI.js
```diff
@@ -0,0 +1,147 @@
...
+            const params = this.getQueryParameters(this.marker, this.limit, this.params);
+
+            const { data }: { data: Data } = await this.getData(this.id, params);
+
+            const entries = this.data ? this.data.entries : [];
+            this.data = {
+                ...data,
+                entries: entries.concat(data.entries)
+            };
+            this.marker = this.data ? this.data.next_marker : '';
```
priyajeet: `this.data` won't ever be null here...

src/api/MarkerBasedAPI.js
```diff
@@ -0,0 +1,147 @@
...
+     * @private
+     */
+    async markerGetRequest(): Promise<void> {
+        if (this.isDestroyed()) {
+            return Promise.resolve();
+        }
+
+        // Make the XHR request
+        try {
+            const params = this.getQueryParameters(this.marker, this.limit, this.params);
```
priyajeet: All args are available on `this`, don't need to pass them.

src/api/MarkerBasedAPI.js
```diff
@@ -0,0 +1,147 @@
...
+
+            const { data }: { data: Data } = await this.getData(this.id, params);
+
+            const entries = this.data ? this.data.entries : [];
+            this.data = {
+                ...data,
+                entries: entries.concat(data.entries)
+            };
+            this.marker = this.data ? this.data.next_marker : '';
+            const hasMoreItems = this.hasMoreItems(this.marker);
```
priyajeet: Same comment as prior `this` one

src/api/MarkerBasedAPI.js
```diff
@@ -0,0 +1,147 @@
...
+     *
+     * @return the query params object
+     */
+    getQueryParameters(marker: string, limit: number, params: ?Object): Object {
+        const queryParams: Params = {
+            marker,
+            limit
+        };
+
+        return Object.assign(queryParams, params);
```
priyajeet: Probably just use spread operator like its done on line 89 to be consistent.
Also begs the question, if this function is really needed since merge would be a one liner.

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -8,11 +8,13 @@ import getProp from 'lodash/get';
 import noop from 'lodash/noop';
 import classNames from 'classnames';
 
+import API from '../../../../api';
```
priyajeet: API should not be imported here, it should only be inside ContentSidebar.js which is the main container and the functions to make calls should be passed down.

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -46,10 +48,20 @@ type Props = {
     }
 };
 
+type Collaborator = {
+    id: string,
+    name: string,
+    item: {
```
priyajeet: item?

src/components/ContentSidebar/ActivityFeed/approval-comment-form/ApprovalCommentForm.scss
```diff
@@ -1,6 +1,8 @@
 @import '../../../variables';
 
 .be .bcs-comment-input-container {
+    @import '~box-react-ui/lib/styles/common/overlay';
```
priyajeet: Do all overlay styles needed?

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -676,6 +672,7 @@ class ContentSidebar extends PureComponent<Props, State> {
                         {shouldRender ? (
                             <Sidebar
                                 file={file}
+                                api={this.api}
```
priyajeet: Should not be passed down

src/flowTypes.js
```diff
@@ -105,14 +105,25 @@ export type BoxItemPermission = {
 };
 
 export type User = {
-    avatarUrl: string,
-    email: string,
+    avatarUrl: ?string,
+    email: ?string,
```
priyajeet: dont think the user object has anything called an email
and its avatar_url per the docs

probably ? should come before : if the field is optional

src/flowTypes.js
```diff
@@ -105,14 +105,25 @@ export type BoxItemPermission = {
...
     id: string,
     login: string,
     name: string,
     type: 'user'
 };
 
+export type UserCollection = {
+    total_count?: number,
+    entries?: User[],
```
priyajeet: There is a mix match in this file, but since we have started to use it in other places, lets stick with Array<Foo> instead of Foo[]

src/util/Xhr.js
```diff
@@ -376,6 +383,10 @@ class Xhr {
      * @return {void}
      */
     abort(): void {
+        if (this.axios && this.axiosSource) {
+            this.axiosSource.cancel('Operation canceled due to new request.');
```
priyajeet: Double check if this does console logging.
Since we don't really want it to log.

Also don't think the `if` check does anything since it will always be true as they are never set to null.

pramodsum: 👍 that was mostly for my testing! Removing it

src/flowTypes.js
pramodsum: Changed it to avatar_url and moved the ?.
The reason email was included here is because BRUI expects this format for @mention selectors. The response from the endpoint is modified to use email instead of login for the user in the current ActivityFeed. 
```
type DefaultSelectorRowProps = {
    item?: {
        name?: string,
        email?: string,
    },
};
```

Would it make sense to do something like this for our case? 
```
export type User = {
    avatar_url?: string,
    email?: string,
    id: string,
    login: string,
    name: string,
    type: 'user'
};

export type Collaborator = {
    id: string,
    name: string,
    item: User,
}
```

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
pramodsum: Moving to this 
```
export type User = {
    avatar_url?: string,
    email?: string,
    id: string,
    login: string,
    name: string,
    type: 'user'
};

export type Collaborator = {
    id: string,
    name: string,
    item: User,
}
```

src/flowTypes.js
pramodsum: Went through and updated it everywhere else in flowTypes.js

src/components/ContentSidebar/ActivityFeed/approval-comment-form/ApprovalCommentForm.scss
pramodsum: We need all except one small toggle one at the bottom so it made more sense to just import the whole thing

src/api/MarkerBasedAPI.js
pramodsum: Just used the spread operator in markerGetRequest() instead

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -142,22 +160,139 @@ class ActivityFeed extends React.Component<Props, State> {
...
+     * Re-formats file collaborators response
+     *
+     * @private
+     * @param {Array<User>} contacts - Collection of collaborated users
+     * @return {Array<Collaborator>}
+     */
+    parseContacts = (contacts: Array<User>): Array<Collaborator> => {
+        const parsedContacts = [];
+
+        contacts.forEach((collab) => {
```
DanDeMicco: Use .map here

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -142,22 +160,139 @@ class ActivityFeed extends React.Component<Props, State> {
...
+                this.mentionCollabsAPI.destroy();
+            }
+        }
+
+        const api = getProp(this.props, 'api', null);
+        this.mentionCollabsAPI = api.getFileCollaboratorsAPI();
+
+        this.mentionSelectorTimeout = setTimeout(() => {
+            this.fetchFileCollaborators(this.mentionCollabsAPI, searchStr, this.getMentionContactsSuccessCallback);
+        }, 2);
```
DanDeMicco: Why a 2ms timeout here? Seems like a very small amount of time. Are you trying to accomplish debouncing?

src/api/MarkerBasedAPI.js
DanDeMicco: I dont think it can be removed, unless you want to suppress the eslint error

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
pramodsum: Just realized there's an actual way to debounce this... Updating the timeout to something realistic. This was something I used just for testing

src/api/MarkerBasedAPI.js
pramodsum: this.data might be null depending on if the previous await has returned or not.

src/flowTypes.js
pramodsum: Or would you recommend 
```
export type CollabItem = {
    email: string,
    id: string,
    name: string,
};

export type Collaborator = {
    id: string,
    name: string,
    item: CollabItem,
}
```

pramodsum: Scratch all of that. I think I can just use SelectorItem

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -631,6 +640,95 @@ class ContentSidebar extends PureComponent<Props, State> {
...
+     * @return {void}
+     */
+    getMentionSelectorContacts = debounce((searchStr: string): void => {
+        // Do not fetch without filter
+        if (!searchStr || searchStr.trim() === '') {
+            return;
+        }
+
+        this.api.getFileCollaboratorsAPI(true).markerGet({
+            id: getProp(this.props, 'fileId', null),
```
DanDeMicco: is getProp necessary?

src/api/MarkerBasedAPI.js
DanDeMicco: You might be able to remove by changing ```return this.markerGetRequest();``` to ```this.markerGetRequest(); return;```

src/components/ContentSidebar/ContentSidebar.js
pramodsum: Removed

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.scss
```diff
@@ -12,6 +12,10 @@
         list-style: none;
     }
 
+    li {
+        line-height: 20px;
```
jeremypress: Is this eyeballed or referencing styles from webapp?

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -688,6 +697,97 @@ class ContentSidebar extends PureComponent<Props, State> {
...
+    /**
+     * File @mention contacts fetch success callback
+     *
+     * @private
+     * @param {API} api - Box API instance
+     * @param {string} searchStr - Search string to filter file collaborators by
+     * @param {Function} successCallback - Fetch success callback
+     * @return {void}
+     */
+    getApproverWithQuery = debounce((searchStr: string): void => {
```
jeremypress: When do these 2 get called? On every keydown in the textarea but then is debounced here?

src/api/MarkerBasedAPI.js
```diff
@@ -0,0 +1,136 @@
...
+        errorCallback: Function,
+        marker?: string,
+        limit?: number,
+        params?: Object,
+        shouldFetchAll?: boolean
+    }): Promise<void> {
+        this.id = id;
+        this.marker = marker;
+        this.limit = limit;
+        this.params = params;
```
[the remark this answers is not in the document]
pramodsum: Oops forgot. Params contains at least marker and limit. However, it can also contain other query params for the endpoint, such as filter_term for the collaborators endpoint

src/api/Base.js
```diff
@@ -180,6 +181,16 @@ class Base {
...
     }
 
+    /**
+     * Formats the api response to usable data
+     * @param {Object} response the api response data
+     * @return {Object} the formatted api response data
+     */
+    formatResponse(response: Object): Object {
+        // Default returns unmodified response
+        return response;
```
jeremypress: Let's look into using responseInterceptors possibly

src/api/FileCollaborators.js
```diff
@@ -0,0 +1,46 @@
...
+            return {
+                id,
+                name,
+                item: { ...collab, email: login }
+            };
+        });
+
+        return {
+            ...response,
+            entries: formattedEntries
```
jeremypress: We should discuss for this endpoint (and others) what the value is of including the original response with our modified one.

src/api/MarkerBasedAPI.js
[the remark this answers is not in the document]
jeremypress: Can you add the other potential properties as optional so people know what might be in there?

src/api/MarkerBasedAPI.js
```diff
@@ -0,0 +1,136 @@
...
+        successCallback: Function,
+        errorCallback: Function,
+        marker?: string,
+        limit?: number,
+        params?: Object,
+        shouldFetchAll?: boolean
+    }): Promise<void> {
+        this.id = id;
+        this.marker = marker;
+        this.limit = limit;
```
jeremypress: I don't think these all need to be class properties. It looks like they only get passed down to markerGetRequest and then to hasMoreItems?

pramodsum: Yea. I can remove them as class variables (and do the same for OffsetBasedAPI)

src/api/MarkerBasedAPI.js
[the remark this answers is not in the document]
pramodsum: Feels weird to add properties as optional here because they're essentially similar to the fields property for offset based APIs. It's pretty arbitrary and depends on the specific API endpoint itself

src/api/Base.js
```diff
@@ -96,6 +96,7 @@ class Base {
      * @return {void}
      */
     destroy(): void {
+        this.xhr.abort();
```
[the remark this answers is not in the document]
priyajeet: Was not needed before axios

src/api/Base.js
priyajeet: One way to avoid having another inheritance-function, is to override the successHandler itself in the chidlren.
Duplication of the one-line isDestroyed will happen though, but maybe thats fine.

src/api/MarkerBasedAPI.js
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    /**
+     * Helper for get
+     *
+     * @private
+     */
+    async markerGetRequest(): Promise<void> {
+        if (this.isDestroyed()) {
+            return Promise.resolve();
```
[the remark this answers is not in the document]
priyajeet: I can take a look but should be doable, see Metadata.js

src/api/Base.js
jeremypress: `        if (!this.isDestroyed() && typeof this.successCallback === 'function') {`
Looks like we're checking this in base already. Do we need need this elsewhere?

jeremypress: @pramodsum I can use the same logic to format response data from a successful comment/task creation, possibly more if the other edit/delete endpoints return the same object structure.

pramodsum: 👍

src/api/FileCollaborators.js
pramodsum: Only going to include the formatted props in the response

src/api/FileCollaborators.js
```diff
@@ -0,0 +1,48 @@
...
+        const collaborators = entries.map((collab) => {
+            const { id, name, login } = collab;
+            return {
+                id,
+                name,
+                item: { ...collab, email: login }
+            };
+        });
+
+        if (!this.isDestroyed() && typeof this.successCallback === 'function') {
```
priyajeet: Probably just at the starting of this function you can add the guard, so that you don't even bother doing the above processing.

```if (this.isDestroyed() ...) { return; }```

Also, will successCallback ever be not a function in this case?

File: src/api/MarkerBasedAPI.js (hunk 23 above)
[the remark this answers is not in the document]
Comment (priyajeet) [reply src/api/MarkerBasedAPI.js]:
```
    async markerGetRequest(): Promise<void> {
        if (this.isDestroyed()) {
            return;
        }

        // Make the XHR request
        try {
            const queryParams: Params = {
                marker: this.marker,
                limit: this.limit,
                ...this.params
            };

            const { data }: { data: Data } = await this.getData(this.id, queryParams);

            const entries = this.data ? this.data.entries : [];
            this.data = {
                ...data,
                entries: entries.concat(data.entries)
            };
            this.marker = data.next_marker;
            const hasMoreItems = this.hasMoreItems();
            if (this.shouldFetchAll && hasMoreItems) {
                this.markerGetRequest();
            }

            this.successHandler(this.data);
        } catch (error) {
            this.errorHandler(error);
        }
    }
```

has no eslint errors

File: src/api/FileCollaborators.js (hunk 24 above)
Comment (pramodsum) [reply src/api/FileCollaborators.js]: Not sure if it will ever not be a function. The overridden function checked for that case

File: src/api/MarkerBasedAPI.js (hunk 23 above)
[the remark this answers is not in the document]
Comment (pramodsum) [reply src/api/MarkerBasedAPI.js]: It seems like this ends up calling the successHandler twice which isn't what we want.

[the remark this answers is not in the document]
Comment (pramodsum) [reply src/api/MarkerBasedAPI.js]: Since this is a recursive call we need to explicitly return the promise to prevent the successHandler from being called n times.

File: src/api/MarkerBasedAPI.js [hunk 25]
```diff
@@ -0,0 +1,136 @@
...
+};
+
+type Data = {
+    next_marker: string,
+    entries: Array<any>
+};
+
+class MarkerBasedApi extends Base {
+    /**
+     * @property {Data}
```
jeremypress: do we need to double up on docs like this (type 2 both places)? Kinda verbose if we're not going to add a comment with any context as to what the property is for

pramodsum: The type definition is there above 
```
type Data = {
    next_marker: string,
    limit: number,
    entries: Array<any>
};
```

pramodsum: Similarly it's there for OffsetBasedAPI ```
type Data = {
    entries: Array<any>,
    total_count: number
};
```

File: src/api/MarkerBasedAPI.js (hunk 23 above)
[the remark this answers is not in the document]
Comment (pramodsum) [reply src/api/MarkerBasedAPI.js]: Got something similar to work

File: src/api/FileCollaborators.js [hunk 26]
```diff
@@ -0,0 +1,47 @@
...
+            return;
+        }
+
+        const { entries } = data;
+        const collaborators = entries.map((collab) => {
+            const { id, name, login } = collab;
+            return {
+                id,
+                name,
+                item: { id, name, email: login }
```
JustinHoldstock: Can you follow suit with Comment type, and change all refs from `email` to `login`?

pramodsum: We're making use of an underlying BRUI DraftJSMentionSelector component which requires the user item to be formatted that way :-/

PR #303, opened 2018-05-09
Add user api to content sidebar

JustinHoldstock:
- Add UsersAPI
- Added ContentSidebar prop `user`. If a `string`, will fetch user info for that id. If `User` object, then we don't have to make an API request. if `undefined`, then it'll fetch the user info for the user the auth token was made for.
- Pass currentUser to Sidebar, as is

** Tests to follow **


boxcla:

src/components/ContentSidebar/Sidebar.js
```diff
@@ -118,12 +121,12 @@ const Sidebar = ({
         />
     );
 
-    if (!hasActivityFeed) {
+    if (!hasActivityFeed || !currentUser) {
```
jeremypress: Not sure about this, I think this check is about props that the parent app has passed in, not the success/failure of API calls.

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -218,6 +223,11 @@ class ContentSidebar extends PureComponent<Props, State> {
                 this.fetchComments(fileId);
                 this.fetchTasks(fileId);
                 this.fetchVersions(fileId);
+                if (typeof user === 'string' || typeof user === 'undefined') {
```
jeremypress: Thinking about this again. If FEF can't pass us a well formed user object, then there's really no point of them passing a User ID. We can just get that with /users/me.

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -434,6 +464,17 @@ class ContentSidebar extends PureComponent<Props, State> {
...
 
+    /**
+     * User fetch success callback
+     *
+     * @private
+     * @param {Object} currentUser - User info object
+     * @return {void}
+     */
+    fetchCurrentUserSuccessCallback = (currentUser: User): void => {
+        this.setState({ currentUser, currentUserError: undefined });
```
jeremypress: Does the User flow type match the API response? I thought we were going to format from snake_case to camelCase?

src/api/Users.js
```diff
@@ -0,0 +1,27 @@
...
+class Users extends Base {
+    /**
+     * API URL for Users
+     *
+     * @param {string} [id] - A box user id. Defaults to 'me' if empty.
+     * @return {string} base url for users
+     */
+    getUrl(id?: string): string {
+        const userId = id || 'me';
+        return `${this.getBaseApiUrl()}/users/${userId}?fields=login,avatar_url,name`;
```
jeremypress: I have a change that makes fields programatic, i'll let you know when that's in. This is fine but in case we want to use the users endpoint for other stuff we should eventually make it more flexible.

boxcla:

Compared against previously merged PR #300

src/api/Users.js
DanDeMicco: I dont think we should request avatar_url as according to the docs it is deprecated. Should just use the avatar endpoint

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -366,6 +376,26 @@ class ContentSidebar extends PureComponent<Props, State> {
...
+     * Handles a failed file user info fetch
+     *
+     * @private
+     * @param {Error} e - API error
+     * @return {void}
+     */
+    fetchCurrentUserErrorCallback = (e: Error) => {
+        this.setState({
+            currentUser: undefined,
+            currentUserError: {
```
DanDeMicco: What is your plan for using this?

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -366,6 +376,26 @@ class ContentSidebar extends PureComponent<Props, State> {
...
     };
 
+    /**
+     * Handles a failed file user info fetch
+     *
+     * @private
+     * @param {Error} e - API error
+     * @return {void}
+     */
+    fetchCurrentUserErrorCallback = (e: Error) => {
```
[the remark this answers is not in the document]
DanDeMicco: I think we need to think about what happens if an API call fails (any of them) for activity feed

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -546,6 +587,21 @@ class ContentSidebar extends PureComponent<Props, State> {
...
 
+    /**
+     * Fetches a Users info
+     *
+     * @private
+     * @param {string} [id] - User id. If missing, gets user that the current token was generated for.
+     * @return {void}
+     */
+    fetchCurrentUser(id?: string = '', shouldDestroy?: boolean = false): void {
+        if (shouldRenderSidebar(this.props)) {
```
DanDeMicco: Why is this if check needed here?

src/components/ContentSidebar/Sidebar.js
DanDeMicco: +1 to Jeremy's comment. We should still render activity feed, but it should be loading

src/api/Users.js
priyajeet: ^ if that, then is fields needed? default values suffice?

src/components/ContentSidebar/ContentSidebar.js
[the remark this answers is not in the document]
priyajeet: Probably will have to loop in design for this.
Maybe some persistent inline notification above the feed?
Or maybe some aleert icon in the corner.

src/components/ContentSidebar/Sidebar.js
JustinHoldstock: Done!

src/components/ContentSidebar/ContentSidebar.js
JustinHoldstock: Done

src/components/ContentSidebar/ContentSidebar.js
JustinHoldstock: As discussed in scrum, this is for another PR/task. I'll format it before passing to `setState()`

i18n/en-US.properties
```diff
@@ -56,6 +56,8 @@ be.createDialogErrorTooLong = This folder name is too long.
 be.createDialogLabel = New Folder
 # Text for create folder dialog
 be.createDialogText = Please enter a name.
+# Current user error message
+be.currentUserErrorHeaderMessage = Something went wrong when fetching current user.
```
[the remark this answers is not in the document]
JustinHoldstock: done

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -218,6 +223,11 @@ class ContentSidebar extends PureComponent<Props, State> {
                 this.fetchComments(fileId);
                 this.fetchTasks(fileId);
                 this.fetchVersions(fileId);
+                if (typeof user === 'string' || typeof user === 'undefined') {
+                    this.fetchCurrentUser(user);
+                } else {
+                    this.setState({ currentUser: user, currentUserError: undefined });
```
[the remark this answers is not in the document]
JustinHoldstock: done

src/components/ContentSidebar/ContentSidebar.js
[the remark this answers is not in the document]
JustinHoldstock: Leaving as is until we figure out how we want to display this

src/components/ContentSidebar/ContentSidebar.js
JustinHoldstock: Not sure yet. In the case that it returns, it means that an auth token was made for/by no one, which I'm not sure how possible that is. Anyone else know?

src/components/ContentSidebar/ContentSidebar.js
JustinHoldstock: Following suit with other API requests. No reason to re-render all of these components if nothing is being shown

src/components/ContentSidebar/ContentSidebar.js
JustinHoldstock: Talking to Zach about this, right now

src/components/ContentSidebar/ContentSidebar.js
jeremypress: I would say don't set the state then right? We want the state item to be of our type, not from the API

src/api/Users.js
```diff
@@ -0,0 +1,27 @@
...
+class Users extends Base {
+    /**
+     * API URL for Users
+     *
+     * @param {string} [id] - A box user id. Defaults to 'me' if empty.
+     * @return {string} base url for users
+     */
+    getUrl(id?: string): string {
+        const userId = id || 'me';
+        return `${this.getBaseApiUrl()}/users/${userId}?fields=login,name`;
```
jeremypress: Fields should be addable to the API call directly now. e.g https://github.com/box/box-ui-elements/commit/df2069cf64a4fac9457e5778547373d2fe42d801#diff-b7fcdd8925731f98f2c2f5413e508d6aR542

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -80,11 +82,13 @@ type State = {
     versions?: FileVersions,
     comments?: Comments,
     tasks?: Tasks,
+    currentUser?: User,
```
jeremypress: What's the difference between currentUser and User, who would the User be?

i18n/en-US.properties
```diff
@@ -1,3 +1,5 @@
+# Title for the Activity Feed tab
+be.activityFeedTitle = Activity
```
priyajeet: This was removed, did you rebase?

priyajeet: its on line 214

src/api/Users.js
priyajeet: fields= needed?
It should get that info without fields specified

src/components/messages.js
```diff
@@ -756,6 +756,16 @@ const messages: { [string]: MessageDescriptor } = defineMessages({
         description: 'File access stats error message',
         defaultMessage: 'Something went wrong when fetching the access stats.'
     },
+    currentUserErrorHeaderMessage: {
+        id: 'be.currentUserErrorHeaderMessage',
+        description: 'Current user error message',
+        defaultMessage: 'Something went wrong when fetching the current user.'
+    },
+    activityFeedTitle: {
```
priyajeet: Probably remove this duplicate

i18n/en-US.properties
JustinHoldstock: I did rebase. It's still in `messages.js`, in master
https://github.com/box/box-ui-elements/blob/master/src/components/messages.js#L333

src/api/Users.js
JustinHoldstock: You are correct! We don't need fields for this call!

PR #318, opened 2018-05-16
Fix: Ensure getAvatarUrl gets passed to Comment.js + fix version messages

pramodsum:


boxcla:

PR #320, opened 2018-05-16
Chore: Adding create/update/delete endpoints for comments

jeremypress:


boxcla:

jeremypress:
Tests on the way

src/api/Base.js
```diff
@@ -165,54 +193,152 @@ class Base {
...
         throw new Error('Implement me!');
     }
 
+    /**
+     * Attaches success and error callbacks for a given API call
+     * @param {Function} successCallback - Success function
+     * @param {Function} errorCallback - Error function
+     */
+    attachCallbacks(successCallback: Function, errorCallback: Function) {
+        if (typeof successCallback === 'function') {
```
DanDeMicco: Due to the fact that flow checks that the arguments are a function, this helper function isnt necessary

src/api/Base.js
```diff
@@ -165,54 +193,152 @@ class Base {
...
+            this.errorHandler(error);
+        }
+    }
+
+    /**
+     * Generic API put
+     *
+     * @param {string} id - The file id
+     * @param {string} url - The url to post to
+     * @param {Object} data - The data to post
```
DanDeMicco: put not post, for url as well

src/api/Base.js
```diff
@@ -165,54 +193,152 @@ class Base {
...
+     * @param {Function} successCallback - The success callback
+     * @param {Function} errorCallback - The error callback
+     * @param {Object} data optional data to delete
+     */
+    async delete(
+        id: string,
+        url: string,
+        successCallback: Function,
+        errorCallback: Function,
+        data?: Object = {}
```
DanDeMicco: Maybe see if you can remove the ? since you are defaulting

src/api/Comments.js
```diff
@@ -5,13 +5,15 @@
...
+import { getTypedFileId } from '../util/file';
+import { PERMISSION_CAN_COMMENT, PERMISSION_CAN_EDIT_COMMENT, PERMISSION_CAN_DELETE_COMMENT } from '../constants';
 
 class Comments extends OffsetBasedAPI {
     /**
-     * API URL for comments
+     * API URL for comments on a file
      *
-     * @param {string} [id] - a box file id
+     * @param {string} [id] - A box file id
```
DanDeMicco: Whats with the brackets?

src/api/Comments.js
```diff
@@ -30,7 +32,15 @@ class Comments extends OffsetBasedAPI {
             return;
         }
 
-        const comments = data.entries.map((comment: Comment) => {
+        // There is no response data when deleting a comment
+        if (!data) {
+            return;
```
DanDeMicco: You should call the successCallback

DanDeMicco: Also, change data: any -> data:Object

src/api/Comments.js
```diff
@@ -40,6 +50,143 @@ class Comments extends OffsetBasedAPI {
...
     };
+
+    /**
+     * API URL for comments endpoint
+     *
+     * @param {string} [id] - A box comment file id
+     * @return {string} base url for comments
+     */
+    getCommentsUrl(id?: string): string {
+        const suffix: string = id ? `/${id}` : '';
```
DanDeMicco: Maybe can make it look prettier

const url = ${this.getBaseApiUrl()}/comments
return id ? url/{id} : url

src/api/Comments.js
```diff
@@ -40,6 +50,143 @@ class Comments extends OffsetBasedAPI {
...
+        successCallback,
+        errorCallback
+    }: {
+        file: BoxItem,
+        message: string,
+        taggedMessage?: string,
+        successCallback: Function,
+        errorCallback: Function
+    }): Promise<void> {
+        const { id = '', permissions } = file;
```
DanDeMicco: Why are you defaulting id?

src/api/Item.js
```diff
@@ -155,7 +155,7 @@ class Item extends Base {
      * @param {Boolean} recursive - True for folders
```
DanDeMicco: Since you touched this function, remove this since the param isnt there :)

src/api/Comments.js
jeremypress: flow doesn't understand that my try catch will guarantee that ID is available by the time we get to line 99. So it thinks i'm possibly calling getTypedFileId on `undefined`. Trying to refactor around this

src/api/Comments.js
```diff
@@ -40,6 +50,143 @@ class Comments extends OffsetBasedAPI {
...
+
+        try {
+            this.checkItem(id, permissions);
+            this.checkPermission(permissions, PERMISSION_CAN_COMMENT);
+        } catch (e) {
+            errorCallback(e);
+            return Promise.reject();
```
priyajeet: Should not return promises unless we are actually trying to read it.

src/api/Comments.js
```diff
@@ -40,6 +50,143 @@ class Comments extends OffsetBasedAPI {
...
+        file,
+        commentId,
+        successCallback,
+        errorCallback
+    }: {
+        file: BoxItem,
+        commentId: string,
+        successCallback: Function,
+        errorCallback: Function
+    }): Promise<void> {
```
priyajeet: no async/await in here, no need to return

priyajeet: same goes for other functions

src/api/Base.js
```diff
@@ -165,54 +193,152 @@ class Base {
...
+    ): Promise<void> {
+        if (this.isDestroyed()) {
+            return;
+        }
+
+        this.attachCallbacks(successCallback, errorCallback);
+
+        // Make the XHR request
+        try {
+            const { responseData } = await this.xhr.post({ id, url, data });
```
priyajeet: shouldn't this be { data } since thats what axios sends.
response.data

src/api/Base.js
```diff
@@ -165,54 +193,152 @@ class Base {
...
+        errorCallback: Function,
+        data?: Object = {}
+    ): Promise<void> {
+        if (this.isDestroyed()) {
+            return;
+        }
+
+        this.attachCallbacks(successCallback, errorCallback);
+
+        // Make the XHR request
```
priyajeet: It looks like all these non-get functions are same, only thing thats diff is this.xhr[operation]

PR #329, opened 2018-05-22
Fix: Fix data returned for update/delete, task properties

jeremypress:


boxcla:

boxcla:

Compared against previously merged PR #323

src/api/Comments.js
```diff
@@ -48,10 +48,18 @@ class Comments extends OffsetBasedAPI {
             return;
         }
 
-        // Manually create the entries array if creating/updating a comment
-        const entries = data.entries ? data.entries : [data];
+        // We don't have entries when updating/creating a comment
+        if (!data.entries) {
+            const comment = {
```
JustinHoldstock: Since the code here is similar/the same as the `map`below, why not extract into a util function, and reuse?
```
extractComment(data) {
    const { tagged_message, message }
     return {
        ...data,
        tagged_message: tagged_message !== '' ? tagged_message : message
    }
}

File: src/api/Tasks.js [hunk 2]
```diff
@@ -48,10 +48,18 @@ class Tasks extends Base {
             return;
         }
 
-        // Manually create the entries array if creating/updating a task
-        const entries = data.entries ? data.entries : [data];
+        // We don't have entries when updating/creating a task
+        if (!data.entries) {
+            const task = {
```
JustinHoldstock: See above comment on extracting as util method you can reuse here and in the `map`

PR #331, opened 2018-05-23
New: Comments can now be posted to the API

jeremypress:

PR #332, opened 2018-05-23
Handle create success for comment

JustinHoldstock:
Comments are relying on: https://github.com/box/box-ui-elements/pull/331
Another series of commits to come with tests and tasks


boxcla:

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -11,7 +11,7 @@ import classNames from 'classnames';
 import ActiveState from './ActiveState';
 import ApprovalCommentForm from '../approval-comment-form';
 import EmptyState from './EmptyState';
-import { collapseFeedState, shouldShowEmptyState } from './activityFeedUtils';
+import { collapseFeedState, shouldShowEmptyState, uuidv4 } from './activityFeedUtils';
```
priyajeet: Is this UUID being persisted?
If this is only client side, and you just want an ID temporarily, use lodash/uniqueid

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -606,15 +607,32 @@ class ContentSidebar extends PureComponent<Props, State> {
         }
     }
 
+    /**
+     * Adds a comment to the state and return the new comment.
+     *
+     * @param {Function} resolve - Resolver function for onCommentCreate() returned Promise
```
jeremypress: I would rename this so that we know it's the callback from the activity feed.

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -606,15 +607,32 @@ class ContentSidebar extends PureComponent<Props, State> {
...
+     * @param {Comment} comment - The newly created comment from the API
+     * @return {void}
+     */
+    addComment(resolve: Function, comment: Comment): void {
```
jeremypress: I also think this naming is a bit confusing.` onCommentCreate` makes the API call and this only updates the state. maybe follow the pattern of other success handlers, `onCommentCreateSucces`?

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -80,18 +81,80 @@ class ActivityFeed extends React.Component<Props, State> {
...
+        const { feedItems } = this.state;
+        feedItems.find((item, index) => {
+            if (item.id === uuid) {
+                itemIndex = index;
+                return true;
+            }
+            return false;
+        });
+
+        // Replace item in the feed items or set as most recent item.
```
jeremypress: Can you explain this case a bit more? If you're updating a pending item, then why would you be straight up adding it to the sorted array? Shouldn't there always be a pending item to be replaced?

src/components/ContentSidebar/ContentSidebar.js
JustinHoldstock: This is not the callback from the AF. I'll clean this up a tad to that instead of taking a resolver function, it just adds to the comment entries.

src/components/ContentSidebar/ContentSidebar.js
JustinHoldstock: done

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
JustinHoldstock: In this case, there really shouldn't be a missing pending item. If there is, we just won't do anything. Updated that code

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -80,18 +81,80 @@ class ActivityFeed extends React.Component<Props, State> {
...
+        const feedItem = {
+            created_at: date,
+            created_by: currentUser,
+            modified_at: date,
+            isPending: true,
+            ...itemBase
+        };
+
+        const { feedItems } = this.state;
+        feedItems.unshift(feedItem);
```
DanDeMicco: Typically in react it is an anti-pattern to modify state directly, and the items should be treated as immutable. You should always return a new array or object instead e.g.
```javascript
this.setState({ feedItems: [feedItem, ...this.state.feedItems] });
```

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -80,18 +81,80 @@ class ActivityFeed extends React.Component<Props, State> {
...
+     * @param {any} args - Data returned by the Comment component on comment creation.
      * @return {void}
      */
     createComment = ({ text, hasMention }: { text: string, hasMention: boolean }): void => {
-        // create a placeholder pending comment
-        // create actual comment and send to Box V2 api
-        // call user passed in handlers.comments.create, if it exists
-        const createComment = getProp(this.props, 'handlers.comments.create', noop);
-        createComment(text, hasMention);
+        const uuid = uniqueId();
```
priyajeet: uniqueId('comment_');

so that its prefixed specific to things used in this file

JustinHoldstock: Done!

boxcla:

Compared against previously merged PR #338

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
JustinHoldstock: Cleaned this up! Thanks for the tip

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -589,15 +589,30 @@ class ContentSidebar extends PureComponent<Props, State> {
...
+     *
+     * @param {Comment} comment - The newly created comment from the API
+     * @return {void}
+     */
+    onCommentCreateSuccess(comment: Comment): void {
+        const { comments } = this.state;
+        if (comments && comments.entries) {
+            const newComments = { ...comments };
+            newComments.entries.push(comment);
+            this.setState({ comments: newComments });
```
DanDeMicco: Do you want to increment the total_count as well? Also, this is a mutation, can do something similar to what you did above

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -69,18 +70,78 @@ class ActivityFeed extends React.Component<Props, State> {
...
+    };
+
+    /**
+     * Replace a feed item with new feed item data.
+     *
+     * @param {Comment | Task} feedItem - API returned feed item data.
+     * @param {string} uuid - Unique ID of the feed item to replace.
+     * @return {void}
+     */
+    updateFeedItem = (feedItem: Comment | Task, uuid: string): void => {
```
DanDeMicco: i would change this variable name to be id, cause in the case of update we will have the actual id, not the uuid

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -69,18 +70,78 @@ class ActivityFeed extends React.Component<Props, State> {
...
+        let itemIndex = null;
+        const { feedItems } = this.state;
+        const newFeedItems = feedItems.slice();
+
+        newFeedItems.find((item, index) => {
+            if (item.id === uuid) {
```
DanDeMicco: You can also use .findIndex instead of find and one line this

```newFeedItems.findIndex((item, index) => item === uuid```

File: src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js [hunk 10]
```diff
@@ -69,18 +70,78 @@ class ActivityFeed extends React.Component<Props, State> {
...
+                itemIndex = index;
+                return true;
+            }
+            return false;
+        });
+
+        // Replace item in the feed items or set as most recent item.
+        if (itemIndex !== null) {
+            newFeedItems[itemIndex] = feedItem;
+            this.setState({ feedItems: newFeedItems });
```
DanDeMicco: Can one line with ```Object.assign([], array, {<index>: newItem});```

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -69,18 +70,78 @@ class ActivityFeed extends React.Component<Props, State> {
...
+     * Replace a feed item with new feed item data.
+     *
+     * @param {Comment | Task} feedItem - API returned feed item data.
+     * @param {string} uuid - Unique ID of the feed item to replace.
+     * @return {void}
+     */
+    updateFeedItem = (feedItem: Comment | Task, uuid: string): void => {
+        let itemIndex = null;
+        const { feedItems } = this.state;
+        const newFeedItems = feedItems.slice();
```
DanDeMicco: This function actually be replaced with a single array.map

```
this.setState({ feedItems: this.state.feedItems.map( (item) => {
// if the matching item, return the replaced item
//else return the original item
```

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -69,18 +70,78 @@ class ActivityFeed extends React.Component<Props, State> {
...
+        const comment = {
+            id: uuid,
+            tagged_message: text,
+            type: 'comment'
+        };
+
+        this.addPendingItem(comment);
+
+        const createComment = getProp(this.props, 'handlers.comments.create', Promise.resolve({}));
+        createComment(text, hasMention).then((commentData) => {
```
DanDeMicco: Am doing something similar right now with updating tasks. What do you think of this pattern instead, since we are already using callbacks?

```
    onTaskUpdate = (
        taskId: string,
        message: string,
        successCallback: (task: Task) => void = noop,
        errorCallback: (e: Error) => void = noop,
        dueAt?: string
    ) => {
        const { file } = this.state;

        if (!file) {
            throw getBadItemError();
        }

        this.api.getTasksAPI(false).updateTask({
            file,
            taskId,
            message,
            dueAt,
            successCallback: (task: Task) => {
                successCallback(task);
                this.onTaskUpdateSuccessCallback(task);
            },
            errorCallback: (e: Error) => {
                errorCallback(e);
                this.errorCallback(e);
            }
        });
    };
```

src/components/ContentSidebar/ContentSidebar.js
JustinHoldstock: Good idea! Done

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
JustinHoldstock: done!

src/components/ContentSidebar/ActivityFeed/activity-feed/__tests__/ActivityFeed-test.js
```diff
@@ -192,4 +193,172 @@ describe('components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed', ()
...
+            instance.updateFeedItem = (data) => {
+                expect(data.tagged_message).toBe(message);
+                done();
+            };
+            instance.createComment({ text: 'irrelevant text', hasMention: false });
+        });
+    });
 });
+
+// createComment = ({ text, hasMention }: { text: string, hasMention: boolean }): void => {
```
DanDeMicco: dead code

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -69,18 +70,103 @@ class ActivityFeed extends React.Component<Props, State> {
...
+     * Add a placeholder pending feed item.
      *
-     * @param {string} text - Comment text
-     * @param {boolean} hasMention - If this comment contains an at mention
+     * @param {Object} itemBase - Base properties for item to be added to the feed as pending.
+     * @return {void}
+     */
+    addPendingItem = (itemBase: Object): void => {
+        const { currentUser } = this.props;
+        // create a placeholder pending feed item
```
pramodsum: Is this comment necessary?

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -612,15 +612,40 @@ class ContentSidebar extends PureComponent<Props, State> {
...
 
+    /**
+     * Adds a comment to the comments state and increases total_count.
+     *
+     * @param {Comment} comment - The newly created comment from the API
+     * @return {void}
+     */
+    onCommentCreateSuccess(comment: Comment): void {
+        const { comments } = this.state;
+        if (comments && comments.entries) {
```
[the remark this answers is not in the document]
priyajeet: Right now you will have to pass in comments to the API. 👎 

But in the future, your API code should be like

```js
// create comment
const comments = this.cache.get('comments_123');
comments.total_count += 1;
comments.entries.push(comment);
this.cache.set('comments_123', comments);

// fetch comments
...
if (this.cache.has('comment_123')) {
    return this.cache.get('comments_123');
}
...

// onCommentCreate
this.api.getCommentsAPI(false).createComment({
    successCallback: fetchComments
}
```

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -638,8 +663,14 @@ class ContentSidebar extends PureComponent<Props, State> {
         this.api.getCommentsAPI(false).createComment({
```
priyajeet: All these API functions should ideally not be {} based, args should be concrete/specific.

src/components/ContentSidebar/ContentSidebar.js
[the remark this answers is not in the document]
JustinHoldstock: @phora, this isn't passing comments into the API, though. This is just updating the `comments` state to include the newly created comment. Are you suggesting all feed item types that are currently in `state` (`versions`, `tasks`, `comments`) be stored in their respective API layers? If so, this will require a bigger refactor than we should be doing in this PR

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
JustinHoldstock: No it isn't!

PR #361, opened 2018-06-07
Update: Clicking on @mentions will use the getUserProfileUrl prop

pramodsum:
- ContentSidebar now contains a prop for getUserProfileUrl() which allows developers to pass in a custom profile url for each Box user ID
- If no method is passed in, @mention clicks will not redirect to a user profile page
- Uses the same method to correctly link comment/task creator's name to their profile


boxcla:

boxcla:

Compared against previously merged PR #355

src/components/ContentSidebar/ActivityFeed/comment/Mention.js
```diff
@@ -8,16 +8,56 @@ import * as React from 'react';
...
-            {children}
-        </a>
-    );
+type State = {
+    profileUrl: ?string
 };
 
+class Mention extends React.PureComponent<Props, State> {
+    state = {
+        profileUrl: '#'
```
jeremypress: Why does the state need to start as a hash?

jeremypress:
What did Varun/Brad say about the case where no method is passed in? Potentially we should make the @ mention look not clickable since we have no profile to direct the user to.

pramodsum:
If we don't want @mentions to look clickable, it might just make sense to use the message instead of the tagged message for comments/tasks w/o a passed in getUserProfileUrl() method

src/components/ContentSidebar/ActivitySidebar.js
```diff
@@ -30,7 +30,8 @@ type Props = {
     mentionSelectorContacts?: SelectorItems,
     commentsError?: Errors,
     tasksError?: Errors,
-    getAvatarUrl: (string) => Promise<?string>
+    getAvatarUrl: (string) => Promise<?string>,
+    getUserProfileUrl?: Function
```
JustinHoldstock: Do you want to type this the same way you typed the other `getUserProfileUrl` props?

pramodsum: yup! I missed this one!

pramodsum:
If no method to determine the profile url is passed in, then @mentions will render as plain strings and not be clickable

src/components/ContentSidebar/ActivityFeed/comment/Mention.js
```diff
@@ -7,17 +7,60 @@ import * as React from 'react';
...
-            {children}
-        </a>
-    );
+type State = {
+    profileUrl: ?string
 };
 
+class Mention extends React.PureComponent<Props, State> {
+    state = {
+        profileUrl: null
```
jeremypress: Do you need to initialize it if its optional?

src/components/ContentSidebar/ActivityFeed/comment/Mention.js
```diff
@@ -7,17 +7,60 @@ import * as React from 'react';
...
+    render() {
+        const { children, ...rest } = this.props;
+        const { profileUrl } = this.state;
+
+        return profileUrl ? (
+            <a {...rest} style={{ display: 'inline-block' }} href={profileUrl}>
+                {children}
+            </a>
+        ) : (
+            children
```
jeremypress: Will this still have the @ sign?

pramodsum: Yup it'll show up like this -> @mention

src/components/ContentSidebar/ActivityFeed/comment/Mention.js
```diff
@@ -7,17 +7,58 @@ import * as React from 'react';
...
 
+class Mention extends React.PureComponent<Props, State> {
+    state = {};
+
+    /**
+     * Success handler for getting avatar url
+     *
+     * @param {string} avatarUrl the user avatar url
+     */
+    getProfileUrlHandler = (profileUrl?: string) => {
```
JustinHoldstock: Is it possible that the get function fails? Maybe this could be called something like `getProfileUrlSuccess`, and you could have a corresponding `fail` that goes in a `catch` in the function below?

PR #373, opened 2018-06-13
Chore: Cleaning up optional vs. required AF props

pramodsum:


boxcla:

boxcla:

Compared against previously merged PR #371

src/components/ContentSidebar/ActivityFeed/approval-comment-form/ApprovalCommentForm.js
```diff
@@ -86,7 +86,7 @@ class ApprovalCommentForm extends React.Component<Props, State> {
             return;
         }
 
-        if (formData.addApproval === 'on') {
+        if (formData.addApproval === 'on' && createTask) {
```
jeremypress: Shouldn't we not be rendering task related stuff if we don't have createTask callback? This form data would never exist in that case. Same goes for other two checks you added.

pramodsum: True. We don't render them but since they're optional props and it's possible that you have only one set of methods available to you (task OR comment). Since some of the methods are optional which could trigger this method to be called, we need to account for them in the methods.

flow-typed/box-ui-elements.js
```diff
@@ -512,6 +512,11 @@ type Collaborators = {
     entries: Array<SelectorItem>
 };
 
+type Translations = {
+    translationEnabled?: boolean,
+    onTranslate?: Function
```
JustinHoldstock: Does this function need to return anything?

src/components/ContentSidebar/ActivityFeed/comment/Comment.js
```diff
@@ -84,7 +79,9 @@ class Comment extends React.Component<Props, State> {
     approvalCommentFormSubmitHandler = (): void => this.setState({ isInputOpen: false, isEditing: false });
     updateTaskHandler = (args: any): void => {
         const { onEdit } = this.props;
-        onEdit(args);
+        if (onEdit) {
```
JustinHoldstock: If you wanna remove the `if`, you could do `const { onEdit = noop } = this.props;`

Your call!

src/components/ContentSidebar/ActivityFeed/comment/InlineEdit.js
```diff
@@ -13,7 +13,7 @@ import messages from '../../../messages';
 
 type Props = {
     toEdit: Function,
-    id: Function,
+    id: string,
```
JustinHoldstock: Thanks for catching this

flow-typed/box-ui-elements.js
pramodsum: Nope it sets the `translatedTaggedMessage` prop in Comment.js

PR #384, opened 2018-06-18
Chore: Consolidating Box specific props

jeremypress:
In this PR I pass the whole object all the way down, but we could de-structure it at a higher level if desired.


boxcla:

PR #387, opened 2018-06-18
Update: Populate AF tasks with task assignments

pramodsum:
- **NOTE**: Tasks with a 'rejected' status are improperly returned from the
API sometimes w/ an unknown or 'incomplete' resolution_status. Setting any task
assignments without an explicit resolution_status to be an 'incomplete'
assignment by default.


boxcla:

src/api/Tasks.js
```diff
@@ -46,30 +46,26 @@ class Tasks extends Base {
...
-        // We don't have entries when updating/creating a task
-        if (!data.entries) {
-            this.successCallback(this.format(data));
-            return;
-        }
-
-        const tasks = data.entries.map(this.format);
-        this.successCallback({ ...data, entries: tasks });
-    };
+    getAssignments(
```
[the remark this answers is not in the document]
pramodsum: Calling `https://api.box.com/2.0/tasks/task_id/assignments` for each task id is better than calling the assignments endpoint for each assignment inside each task

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -74,6 +79,7 @@ type State = {
     versions?: FileVersions,
     comments?: Comments,
     tasks?: Tasks,
+    tasksWithAssignments?: Tasks,
```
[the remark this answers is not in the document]
pramodsum: Same type just one has assignments updated and one doesn't

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -468,6 +475,53 @@ class ContentSidebar extends PureComponent<Props, State> {
...
+     * File task assignment fetch success callback
+     *
+     * @private
+     * @param {TaskAssignments} assignments - Box task assigment
+     * @return {void}
+     */
+    fetchTaskAssignmentsSuccessCallback = (assignments: TaskAssignments): void => {
+        const { entries, total_count } = this.state.tasks;
+        this.setState({
+            tasksWithAssignments: {
```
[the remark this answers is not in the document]
pramodsum: We cannot update the tasks state at any point because that would cause a render(). We need to maintain an internal state and then only set the tasks object in the state when all the assignments have been fetched

src/api/Base.js
```diff
@@ -219,10 +219,24 @@ class Base {
      * @param {Function} successCallback - The success callback
      * @param {Function} errorCallback - The error callback
      * @param {Object} params request params
+     * @param {string} url - API url
+     * @returns {Promise}
      */
-    get(id: string, successCallback: Function, errorCallback: Function, params?: Object): void {
-        const url = this.getUrl(id);
-        this.makeRequest(HTTP_GET, id, url, successCallback, errorCallback, params);
+    get({
```
JustinHoldstock: Would it be worth updating `post`, `put`, and `delete` to work with a similar fashion of arguments?

src/components/ContentSidebar/ActivityFeed/task/Task.js
```diff
@@ -115,35 +111,38 @@ class Task extends React.Component<Props> {
...
+                                        case TASK_REJECTED:
+                                            return <RejectedAssignment {...assigned_to} key={assigned_to.id} />;
+                                        case TASK_INCOMPLETE:
+                                        default:
+                                            return (
+                                                <PendingAssignment
+                                                    {...assigned_to}
+                                                    key={assigned_to.id}
+                                                    onTaskApproval={() =>
+                                                        onTaskAssignmentUpdate(id, assignmentId, TASK_APPROVED)
```
JustinHoldstock: Can you make sure that when you extract `onTaskAssigment` from props, you default to `noop`? Just to make sure this doesn't blow up

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -525,11 +532,63 @@ class ContentSidebar extends PureComponent<Props, State> {
...
-        this.setState({ tasks, tasksError: undefined });
+    fetchTaskAssignmentsSuccessCallback = (assignments: TaskAssignments): void => {
+        const { tasksWithoutAssignments } = this.state;
+        if (!tasksWithoutAssignments) {
+            return;
+        }
+
+        const { entries, total_count } = tasksWithoutAssignments;
+        this.setState({
+            tasksWithoutAssignments: {
```
JustinHoldstock: should this, instead, be just `tasks`? Id assume that once all tasks had assignments, we could finally render in the AF

JustinHoldstock: Ah, I see down below you set via `tasks: this.state.tasksWithoutAssignments`

pramodsum: That method updates the assignments for only that particular task so eventually when all those promises are resolved, we can update tasks

src/api/Base.js
pramodsum: I can do that in a follow up commit 👍

src/components/ContentSidebar/ActivityFeed/task/Task.js
pramodsum: 👍

boxcla:

Compared against previously merged PR #390

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -525,11 +532,63 @@ class ContentSidebar extends PureComponent<Props, State> {
      * File tasks fetch success callback
      *
      * @private
-     * @param {Object} tasks - Box task
+     * @param {Object} tasks - Box task without assignments
+     * @return {void}
+     */
+    fetchTasksSuccessCallback = (tasksWithoutAssignments: Tasks): void => {
+        this.setState({ tasksWithoutAssignments, tasksError: undefined });
+        this.fetchTaskAssignments(tasksWithoutAssignments);
```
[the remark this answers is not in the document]
jeremypress: You could either:
1. Block rendering if the task state object doesn't have a boolean set (not sure if this is a bad react practice)
2. Pass the tasks info into task assignment method and then to its callbacks using arrow functions (I do this in a few places in ActivityFeed.js

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -1037,9 +1121,11 @@ class ContentSidebar extends PureComponent<Props, State> {
         const { fileId = '' } = this.props;
         if (SidebarUtils.canHaveSidebar(this.props)) {
             if (typeof user === 'undefined') {
-                this.api
```
jeremypress: can you update the docstring? There's no more ID, just user. Also that typeof check seems weird

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -121,6 +127,7 @@ class ContentSidebar extends PureComponent<Props, State> {
         versions: undefined,
         comments: undefined,
         tasks: undefined,
+        tasksWithoutAssignments: undefined,
```
jeremypress: not needed anymore?

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -1005,12 +1046,53 @@ class ContentSidebar extends PureComponent<Props, State> {
...
+            params: {
+                fields: TASK_ASSIGNMENTS_FIELDS_TO_FETCH.toString()
+            }
+        };
+
+        let tasks = tasksWithoutAssignments;
+        const { entries } = tasksWithoutAssignments;
+        const taskAssignmentPromises = [];
+        entries.forEach((task) => {
+            const promise = this.api.getTasksAPI(shouldDestroy).getAssignments(
```
jeremypress: Since we're passing in the URL now do you just want to call `get` directly?

jeremypress: get doesn't return a promise, how does this work exactly?

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -522,14 +529,48 @@ class ContentSidebar extends PureComponent<Props, State> {
...
+     * Update Tasks to include task assignments
      *
      * @private
-     * @param {Object} tasks - Box task
-     * @return {void}
+     * @param {Array<TaskAssignment>} entries - Box task assignment entries
+     * @param {TaskAssignments} assignments - Box task assigments
+     * @return {TaskAssignments} Updated Box task assignments
+     */
+    populateTaskAssignments(entries: Array<TaskAssignment>, assignments: TaskAssignments): TaskAssignments {
```
jeremypress: Can you make the params a little more descriptive? Somehow indicate that entries are coming from the original tasks call and these assignments are what we are going to be pushing in.

jeremypress: Was there a reason you didn't want to pass the whole task object instead of just the entries array?

src/components/ContentSidebar/ContentSidebar.js
JustinHoldstock: @jeremypress  If you'll notice, Sumedha has modified it to return a promise

pramodsum: `get()` isn't it's own async method. whereas `makeRequest()` is. I modified `get()` to return the promise from `makeRequest()`

src/components/ContentSidebar/ContentSidebar.js
pramodsum: I think we can refactor the logic this method outside of this PR. This change was just moving the get request to use an object rather than individual params

src/components/ContentSidebar/ContentSidebar.js
pramodsum: This is updating the assignments within a specific task. Otherwise would have to loop through again to figure out which task is being updated

pramodsum:
yup! @JustinHoldstock https://github.com/box/box-ui-elements/pull/387/files#diff-f26ce185574104ea036d6d0575437843R70

src/api/Base.js
pramodsum: Done in #407

PR #388, opened 2018-06-18
Fix: Include a file ID when getting the current user

jeremypress:


boxcla:

boxcla:

Compared against previously merged PR #385

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -972,11 +973,19 @@ class ContentSidebar extends PureComponent<Props, State> {
...
      */
     fetchCurrentUser(user?: User, shouldDestroy?: boolean = false): void {
+        const { fileId = '' } = this.props;
         if (shouldRenderSidebar(this.props)) {
             if (typeof user === 'undefined') {
+                const getUrl = this.api.getUsersAPI(false).getUrl();
                 this.api
                     .getUsersAPI(shouldDestroy)
-                    .get('', this.fetchCurrentUserSuccessCallback, this.fetchCurrentUserErrorCallback);
+                    .makeRequest(
```
priyajeet: You should not bleed api stuff in here as much as possible.
Change the get function to to take an id and provide it instead.

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -972,11 +973,19 @@ class ContentSidebar extends PureComponent<Props, State> {
      * @return {void}
      */
     fetchCurrentUser(user?: User, shouldDestroy?: boolean = false): void {
+        const { fileId = '' } = this.props;
         if (shouldRenderSidebar(this.props)) {
             if (typeof user === 'undefined') {
+                const url = this.api.getUsersAPI(false).getUrl();
```
priyajeet: You should not bleed api stuff in here as much as possible. Otherwise we are replicating some internals here.
Change the get function to to take an id and provide it to xhr instead.

jeremypress: you're right, I switched up the Users API since it's the only endpoint that's base getURL won't consider an ID

PR #386, opened 2018-06-18
Fix: Design feedback and bug fixes

priyajeet:
Fixes sidebar menu ordering with activity feed first
Removes the need of hasSidebar prop
Determines hasSidebar via other props
Does not render the sidebar if no data to render for skills
Minor layout and width fixes for sidebar
Moves sidebar view logic into parent ContentSidebar component
Improves sidebar closing/opening based on viewport width

PR #391, opened 2018-06-19
Breaking: Refactoring details tab specific props

jeremypress:


boxcla:

boxcla:

Compared against previously merged PR #386

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -105,13 +98,9 @@ class ContentSidebar extends PureComponent<Props, State> {
         getPreviewer: noop,
         currentUser: undefined,
         hasSkills: false,
-        hasProperties: false,
         hasMetadata: false,
-        hasNotices: false,
```
jeremypress: Will need to figure out the best place to default these. Probably in DetailsSidebar

priyajeet: default these here itself since they are needed in this file

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -45,23 +45,16 @@ type Props = {
     currentUser?: User,
     getPreviewer: Function,
     hasSkills: boolean,
-    hasProperties: boolean,
+    detailsSidebarProps: Object,
```
priyajeet: So the type for this will be DetailsSidebarProps. This type will be exported from DetailsSidebar.js same way I am exporting ContentSidebarBarProps at the bottom of ContentSidebar.js

jeremypress: @priyajeet this is causing a few warnings
```warning.js:33 Warning: Failed prop type: The prop `hasClassification` is marked as required in `DetailsSidebar`, but its value is `undefined`.```

```warning.js:33 Warning: Failed prop type: The prop `detailsSidebarProps.hasClassification` is marked as required in `ContentSidebar`, but its value is `undefined`.
    in ContentSidebar```

My change makes sense to me, all of those types have a default and those defaults will get passed down. Not sure why that causes a warning.

Comment (jeremypress) [reply src/components/ContentSidebar/ContentSidebar.js]: Actually, these warnings should be here. If the parent doesn't pass in these props warnings should be thrown as they are required (even though they have a default). Since we're giving them a default, why not make them optional?

File: src/components/ContentSidebar/ContentSidebar.js [hunk 3]
```diff
@@ -1189,6 +1187,17 @@ class ContentSidebar extends PureComponent<Props, State> {
...
 
+    /**
+     * Returns detailsSidebarProps replacing missing props with their defaults
+     *
+     * @private
+     * @return {DetailsSidebarProps} the merged detailsSidebarProps
+     */
+    getdetailsSidebarProps(): DetailsSidebarProps {
+        const { detailsSidebarProps } = this.props;
+        return extend(ContentSidebar.defaultProps.detailsSidebarProps, detailsSidebarProps);
```
jeremypress: This will allow us to nest further if we decide to put objects in detailsSidebarProps

JustinHoldstock: Should the `DetailsSidebar` already have these default props? Are there any cases where the DetailsSidebar is used outside of the ContentSidebar?

src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -91,4 +91,15 @@ const DetailsSidebar = ({
     </SidebarContent>
 );
 
+export type DetailsSidebarProps = {
```
JustinHoldstock: Curious as to the reasoning of the different prop type, considering the substantial overlap with the above `Props` type.

src/components/ContentSidebar/ContentSidebar.js
jeremypress: As far as I know, the tabs we're building (details, AF, skills etc.) Will always be used in the content sidebar

src/components/ContentSidebar/DetailsSidebar.js
jeremypress: The above type is everything the details tab expects. This type a way to document what we expect to be passed into the sidebar by the consuming app. For flow purposes, this needs to be a type so it can be properly typed in `ContentSidebar`

JustinHoldstock: Ah, I see. Are compound types available in Flow? If so, could we have `Props` include the props that DO NOT exist in `DetailsSidebarProps`, and have it extend/include `DetailsSidebarProps`;

IE) 
```
type DetailsSidebarProps = {
  my_prop: boolean
};

type Props = {
  your_prop: string
} | DetailsSidebarProps;
```

jeremypress: yeah I can try to do something like that!

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -14,6 +14,7 @@ import noop from 'lodash/noop';
 import cloneDeep from 'lodash/cloneDeep';
 import LoadingIndicator from 'box-react-ui/lib/components/loading-indicator/LoadingIndicator';
 import Sidebar from './Sidebar';
+import type { DetailsSidebarProps } from './DetailsSidebar';
```
priyajeet: move type imports to the bottom before scss

src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -17,32 +17,35 @@ import SidebarFileProperties from './SidebarFileProperties';
 
 import './DetailsSidebar.scss';
 
+export type DetailsSidebarProps = {
```
priyajeet: all exports should be at the bottom

src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -17,32 +17,35 @@ import SidebarFileProperties from './SidebarFileProperties';
...
+    hasNotices?: boolean,
+    hasProperties?: boolean,
+    hasAccessStats?: boolean,
+    hasClassification?: boolean,
+    hasVersions?: boolean,
+    onAccessStatsClick?: Function,
+    onClassificationClick?: Function,
+    onVersionHistoryClick?: Function
```
priyajeet: might as well add your retentionpolicy one too?

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -1343,23 +1321,16 @@ class ContentSidebar extends PureComponent<Props, State> {
...
-                                hasClassification={hasClassification}
                                 hasActivityFeed={hasActivityFeed}
                                 onDescriptionChange={this.onDescriptionChange}
                                 accessStats={accessStats}
-                                onAccessStatsClick={onAccessStatsClick}
-                                onClassificationClick={onClassificationClick}
-                                onVersionHistoryClick={onVersionHistoryClick}
                                 onSkillChange={this.onSkillChange}
-                                hasVersions={hasVersions}
                                 accessStatsError={accessStatsError}
```
priyajeet: don't you want to move the errors related to the details sidebar in there too?
As well as the onDescriptionChange.

You can merge these internal ones into detailsSidebarProps like

detailsSidebarProps={{
 foo: bar,
...detailsSidebarProps
}}

jeremypress: good call.

src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -17,32 +17,37 @@ import SidebarFileProperties from './SidebarFileProperties';
 
 import './DetailsSidebar.scss';
 
+type ExternalProps = {
```
jeremypress: @JustinHoldstock name ideas?

JustinHoldstock: I actually like the current name! We always assume that `Props` is an internal type, whereas External explains something different than what we are expecting. Do you want to export that, or is it not longer referenced outside of here?

PR #397, opened 2018-06-22
Update: Task assignments can be updated via the API

pramodsum:
- The task assignments API updates the assignment message rather than the resolution state, therefore displaying the task assignment state based on the value of message (if available) otherwise based on the returned resolution_state
- This will also display rejected tasks properly on the initial fetch


boxcla:

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -305,7 +305,31 @@ class ActivityFeed extends React.Component<Props, State> {
      */
     updateTaskAssignment = (taskId: string, taskAssignmentId: string, status: string): void => {
         const updateTaskAssignment = this.props.onTaskAssignmentUpdate || noop;
-        updateTaskAssignment(taskId, taskAssignmentId, status);
+        const { feedItems } = this.state;
+        const task = feedItems.find((item) => !!(item.id === taskId));
+        updateTaskAssignment(taskId, taskAssignmentId, status, (assignment) => {
```
JustinHoldstock: Would it be possible to extract some of this into another method?

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -295,6 +295,40 @@ class ActivityFeed extends React.Component<Props, State> {
...
+                    ...updatedAssignment,
+                    resolution_state
+                };
+            }
+            return item;
+        });
+
+        this.updateFeedItem(
+            {
+                ...task,
```
jeremypress: Does the call to `updateFeedItem` work if you only pass the updates?

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -541,7 +541,8 @@ class ContentSidebar extends PureComponent<Props, State> {
                 const assignment = assignments.entries.find((a) => a.id === item.id);
                 if (assignment) {
                     return {
-                        ...assignment
+                        ...assignment,
+                        resolution_state: assignment.message.toLowerCase() || assignment.resolution_state
```
jeremypress: Are we calling this method on read, update, and eventually create? Also, when does messages become populated vs. resolution state? Is it only there after updates. It would be great if we could look at one field.

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -305,7 +339,15 @@ class ActivityFeed extends React.Component<Props, State> {
...
         const updateTaskAssignment = this.props.onTaskAssignmentUpdate || noop;
-        updateTaskAssignment(taskId, taskAssignmentId, status);
+        const { feedItems } = this.state;
+        const task = feedItems.find((item) => !!(item.id === taskId));
+        if (!task) {
+            return;
+        }
+
+        updateTaskAssignment(taskId, taskAssignmentId, status, (updatedAssignment) =>
+            this.updateTaskAssignmentSuccessCallback(task, updatedAssignment)
```
jeremypress: When you add the failure case, please update the item with the error object like we do in other failure cases.

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -295,6 +295,40 @@ class ActivityFeed extends React.Component<Props, State> {
...
+     * @param {Task} task - Box task
+     * @param {TaskAssignment} updatedAssignment - New task assignment from API
+     * @return {void}
+     */
+    updateTaskAssignmentSuccessCallback = (task: Task, updatedAssignment: TaskAssignment) => {
+        const { entries, total_count } = task.task_assignment_collection;
+
+        const assignments = entries.map((item: TaskAssignment) => {
+            if (item.id === updatedAssignment.id) {
+                const resolution_state = updatedAssignment.message.toLowerCase() || updatedAssignment.resolution_state;
```
jeremypress: Don't need to look at resolution_state anymore in this case

boxcla:

Compared against previously merged PR #401

PR #401, opened 2018-06-22
Chore: Return promises from all API CRUD operations

pramodsum:


boxcla:

PR #437, opened 2018-07-06
New: UploadsManager drag and drop upload empty state

wenboyu2:
- adds an `isUploadsManagerVisible` prop to the `ContentUploader` to allow the UploadsManager to be shown when there's no ongoing uploads
- adds an empty state to the UploadsManager which instructs user to drop file to upload

### DIFF
--- a/src/components/ContentUploader/ContentUploader.js
+++ b/src/components/ContentUploader/ContentUploader.js
@@ -468,7 +468,7 @@ class ContentUploader extends Component<Props, State> {
         const { rootFolderId } = this.props;
 
         // Convert files from the file API to upload items
-        const newItems = this.getNewFiles(files).map((file) => {
+        const newItems = files.map((file) => {
             const uploadFile = getFile(file);
             const uploadAPIOptions = getFileAPIOptions(file);
             const { name, size } = uploadFile;
--- a/src/components/ContentUploader/ContentUploader.js
+++ b/src/components/ContentUploader/ContentUploader.js
@@ -233,7 +233,7 @@ class ContentUploader extends Component<Props, State> {
         const { rootFolderId } = this.props;
         const { itemIds } = this.state;
 
-        return Array.from(files).filter((file) => !(getFileId(file, rootFolderId) in itemIds));
+        return Array.from(files).filter((file) => !itemIds[getFileId(file, rootFolderId)]);
     };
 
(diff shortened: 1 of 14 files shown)


boxcla:
Verified that @wenboyu2 has signed the CLA. Thanks for the pull request!

boxcla:

Compared against previously merged PR #435

src/components/ContentUploader/ContentUploader.js
```diff
@@ -233,57 +236,64 @@ class ContentUploader extends Component<Props, State> {
      * @param {Array<UploadFileWithAPIOptions | File>} files
      */
     getNewFiles = (files: Array<UploadFileWithAPIOptions | File>) => {
+        const { rootFolderId } = this.props;
         const { itemIds } = this.state;
 
-        return Array.from(files).filter((file) => !(this.getFileId(file) in itemIds));
+        return Array.from(files).filter((file) => !(getFileId(file, rootFolderId) in itemIds));
```
priyajeet: Not sure using "in" is the best. Isn't it outdated?
Better is to use array.includes

src/components/ContentUploader/ContentUploader.js
```diff
@@ -233,57 +236,64 @@ class ContentUploader extends Component<Props, State> {
...
+        });
+
         clearTimeout(this.resetItemsTimeout);
 
-        const firstFile = getFile(files[0]);
+        const firstFile = getFile(newFiles[0]);
 
-        if (firstFile.webkitRelativePath) {
-            this.addFilesWithRelativePathToQueue(files, itemUpdateCallback);
+        if (firstFile.webkitRelativePath && !isRelativePathIgnored) {
```
tonyjin: Can you add some comments here? It's not clear to me why we only add the first item in this situation

wenboyu2: Thanks for you offline feedback, I updated the FolderUpload class interface which now only represents one folder.

src/components/ContentUploader/ContentUploader.js
wenboyu2: `itemIds` is an object `{ [string]: boolean }`
I changed this to be just an object lookup instead, `!itemIds[getFileId(file, rootFolderId)]`

PR #438, opened 2018-07-06
Fix: Fetching task assignment resolution states

pramodsum:
* Chore: Ensure pending tasks assignments have an incomplete resolution_state

### DIFF
--- a/src/components/ContentSidebar/ActivityFeed/task/Task.js
+++ b/src/components/ContentSidebar/ActivityFeed/task/Task.js
@@ -116,11 +116,17 @@ class Task extends React.Component<Props> {
                                                 <PendingAssignment
                                                     {...assigned_to}
                                                     key={assigned_to.id}
-                                                    onTaskApproval={() =>
-                                                        onAssignmentUpdate(id, assignmentId, TASK_APPROVED)
+                                                    onTaskApproval={
+                                                        isPending
+                                                            ? noop
+                                                            : () =>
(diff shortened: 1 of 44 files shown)


boxcla:

src/components/ContentSidebar/ActivityFeed/task/Task.js
```diff
@@ -130,6 +129,8 @@ class Task extends React.Component<Props> {
                                                     }
                                                 />
                                             );
+                                        default:
```
jeremypress: Should we avoid rendering the whole task if there is no assignee? We would have to bump the check up a level.

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -547,31 +547,6 @@ class ContentSidebar extends PureComponent<Props, State> {
...
 
-    /**
-     * Update Tasks to include task assignments
-     *
-     * @private
-     * @param {Array<TaskAssignment>} entries - Box task assignment entries
-     * @param {TaskAssignments} assignments - Box task assigments
-     * @return {TaskAssignments} Updated Box task assignments
-     */
-    populateTaskAssignments(entries: Array<TaskAssignment>, assignments: TaskAssignments): TaskAssignments {
```
jeremypress: Gracias

src/components/ContentSidebar/ActivityFeed/task/Task.js
pramodsum: I can make sure we check for that.

PR #445, opened 2018-07-10
Chore: Removing NSP check

jeremypress:


boxcla:

jeremypress:
Let's keep this running locally since it still seems to work. We can fully remove after deprecation on Sept 30. Removing the webhook will unblock our PRs

PR #449, opened 2018-07-10
Fix: Ensures restored version is only added on first feed load

pramodsum:


boxcla:

boxcla:

Compared against previously merged PR #444

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -427,7 +427,10 @@ class ActivityFeed extends React.Component<Props, State> {
         const { file } = this.props;
         const { restored_from, modified_at, file_version } = file;
 
-        if (restored_from) {
+        // Ensures restored version is only added on first feed load
+        const hasRestoredVersion = versions.entries.find((version) => version.action === VERSION_RESTORE_ACTION);
```
DanDeMicco: Is there a better way you can detect this rather than looping through all the versions?

pramodsum: ^ looking at the last element in the versions array instead now :)

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -427,7 +427,9 @@ class ActivityFeed extends React.Component<Props, State> {
         const { file } = this.props;
         const { restored_from, modified_at, file_version } = file;
 
-        if (restored_from) {
+        // Ensures restored version is only added on first feed loads
+        const lastVersion = versions.total_count ? versions.entries[versions.total_count - 1] : null;
```
DanDeMicco: Maybe change to : {} instead of null so the checking of lastVersion is not needed below?

PR #468, opened 2018-07-25
New: Support legacy and editable auto captions

jeremypress:
tests on the way


boxcla:

boxcla:

Compared against previously merged PR #465

src/components/ContentSidebar/Skills/SidebarSkills.js
```diff
@@ -28,86 +28,125 @@ type Props = {
...
-        case SKILLS_KEYWORD:
-            return SKILLS_TARGETS.KEYWORDS.CARD;
-        case SKILLS_FACE:
-        case SKILLS_TIMELINE:
-            return SKILLS_TARGETS.FACES.CARD;
-        case SKILLS_TRANSCRIPT:
-            return SKILLS_TARGETS.TRANSCRIPTS.CARD;
-        default:
-            return '';
+class SidebarSkills extends React.PureComponent<Props> {
```
priyajeet: Do all this in SkillsSidebar.js

priyajeet:
Ok, here is your another issue. Skills sidebar tab currently opens by default, so your code gets executed. If the skills sidebar is no longer the default in the future, then your code won't reach.

jeremypress:
@priyajeet True, we could either move this farther out to contentSidebar. I've settled on this approach because doing anything in preview would be blocked by the read token/ The token used by T3

priyajeet:
Another issue, the viewer load event may get fired before the sidebar is rendered.

jeremypress:
Before I add the event listener I can check to see if the viewer is loaded, and call the method right away. I can move the initial setup to contentsidebar, and then keep the update on edit in the component.

PR #469, opened 2018-07-25
Fix: Missing locales and test

priyajeet:


boxcla:

PR #472, opened 2018-07-25
Chore: Add resin component to sidebar

JustinHoldstock:


boxcla:

boxcla:

Compared against previously merged PR #471

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -93,6 +93,9 @@ const activityFeedInlineError: Errors = {
         content: messages.activityFeedItemApiError
     }
 };
+
+const RESIN_INFOPANE_TAG = 'infopane';
```
priyajeet: src/interactionTargets.js

priyajeet: also call it 'sidebar' to be consistent

priyajeet: Wait this is adding a component. We aren't adding any resin tags other than target in UI elements.

jeremypress: Is there a way we could allow for generic attributes in content preview that get passed in by EUA? Similar to how many BRUI components do?

jeremypress: that won't help much since we already have the page labeled...

jeremypress: We could pass it in the sidebar props and apply ...rest to the DOM element like BRUI does in many instances

JustinHoldstock: `infopane` was what is requested as the component name, to be consistent with Webapp. Leaving as is

PR #476, opened 2018-07-26
Fix: Skills/Metadata navigating logic

priyajeet:
Minor CSS and style fixes.
Minor fixes to transcript copy.
Fixing test files with new props style.


boxcla:

PR #490, opened 2018-07-31
New: Style guide for UI Elements

priyajeet:
Usage:

`yarn styleguide`
will use default folder and token.

`TOKEN=mytoken FILEID=myfileid FOLDERID=myfolderid yarn styleguide`
will use the specified token, file and folder respectively.

`http://localhost:6060`
needs to be whitelisted in the developer console

Styleguide only tests react components.
Wrapper libraries should continue be tested via test/*.html files.

TODO: Still need to test gh-pages stuff. Will do some test builds.


boxcla:

package.json
```diff
@@ -56,14 +57,16 @@
     "lint": "NODE_ENV=dev ./node_modules/.bin/eslint src && ./node_modules/.bin/stylelint 'src/**/*.scss'",
     "npm": "OUTPUT=dist LANGUAGE=en-US REACT=true yarn run build-prod",
     "nsp": "nsp check --reporter summary",
-    "major": "./build/release.sh -m && ./build/publish.sh",
-    "minor": "./build/release.sh -n && ./build/publish.sh",
-    "patch": "./build/release.sh -p && ./build/publish.sh",
+    "major": "./build/release.sh -m && ./build/publish.sh && ./build/publish-examples.sh",
```
jstoffan: Minor: Should `publish-examples.sh` be called from `publish.sh`, instead? It'll help ensure that the examples are always kept updated, even if `publish.sh` is called out of band (like by CI, or something).

build/publish-examples.sh
```diff
@@ -0,0 +1,165 @@
...
+        echo "----------------------------------------------------"
+        echo "Error in build_assets!"
+        echo "----------------------------------------------------"
+        exit 1
+    fi
+
+    # Publish gh-pages
+    if ! push_to_gh_pages; then
+        echo "----------------------------------------------------"
+        echo "Error in push_to_npm!"
```
jeremypress: push_to_pages

PR #481, opened 2018-07-26
Fix: show latest version if it is an upload in activity feed

DanDeMicco:

PR #494, opened 2018-08-02
Chore: Feed API refactor

DanDeMicco:
Moves the state out of ActivityFeed, and encapsulate all the API logic for ContentSidebar and move it into the Feed API and the ActivitySidebar container


boxcla:

boxcla:

Compared against previously merged PR #481

src/api/Feed.js
```diff
@@ -0,0 +1,1022 @@
...
+        successCallback: Function,
+        errorCallback: Function
+    ): void => {
+        if (!file.id) {
+            throw getBadItemError();
+        }
+
+        this.id = file.id;
+        this.updateFeedItem({ isPending: true }, taskId);
+        const taskAssignmentsAPI = new TaskAssignmentsAPI(this.options);
```
JustinHoldstock: I've noticed you're new-ing each kind of API where it's used. Could you make them once, and then cache and reuse them, instead?

src/api/Feed.js
```diff
@@ -0,0 +1,1022 @@
...
+     * Update task error callback
+     *
+     * @param {Object} e - Axios error
+     * @param {Function} errorCallback - The error callback
+     * @return {void}
+     */
+    updateTaskErrorCallback = (e: $AxiosXHR<any>, errorCallback: Function) => {
+        this.errorCallback(e);
+        if (!this.isDestroyed()) {
+            errorCallback(e);
```
JustinHoldstock: Regardless of this being destroyed, shouldn't a passed in error callback always fire when an error occurred? How will the parent of this know what to do?

src/api/Feed.js
```diff
@@ -0,0 +1,1022 @@
...
+        const updates = omit(task, TASK_ASSIGNMENT_COLLECTION);
+        this.updateFeedItem(
+            {
+                ...updates,
+                isPending: false
+            },
+            task.id
+        );
+        if (!this.isDestroyed()) {
+            successCallback(task);
```
JustinHoldstock: See above comment about error callback and the state of this being destroyed

src/api/Feed.js
```diff
@@ -0,0 +1,1022 @@
...
+        }
+
+        this.id = file.id;
+        this.updateFeedItem({ isPending: true }, commentId);
+
+        this.commentsAPI.deleteComment({
+            file,
+            commentId,
+            permissions,
+            successCallback: () => {
```
JustinHoldstock: I know we like to use arrow functions, but do we get anything out of it that couldn't be done via `.bind()`? In the end we're making a new function object that's bound to the calling scope.

ie) `successCallback: this.deleteFeedItem.bind(this, commentId, successCallback)`

JustinHoldstock: This comment also refers to other places where we use an arrow function, in this manner, with no arguments

src/api/Feed.js
```diff
@@ -0,0 +1,1022 @@
...
+        const task = {
+            due_at: dueAtString,
+            id: uuid,
+            is_completed: false,
+            message,
+            task_assignment_collection: {
+                entries: pendingAssignees,
+                total_count: pendingAssignees.length
+            },
+            type: 'task'
```
JustinHoldstock: Move 'task' to a const

src/api/Feed.js
```diff
@@ -0,0 +1,1022 @@
...
+    /**
+     * Fetches the task assignments for each task
+     *
+     * @param {Array} tasksWithoutAssignments - Box tasks
+     * @return {Promise}
+     */
+    fetchTaskAssignments(tasksWithoutAssignments: Tasks): Promise<?Tasks> {
+        const requestData = {
+            params: {
+                fields: TASK_ASSIGNMENTS_FIELDS_TO_FETCH.toString()
```
JustinHoldstock: Could you assign the stringed version of `TASK_ASSIGNMENTS_FIELDS_TO_FETCH ` to a const, and reuse it?

src/api/Feed.js
```diff
@@ -0,0 +1,1022 @@
...
+     * Callback for successful creation of a Comment.
+     *
+     * @param {Object} e - The axios error
+     * @param {string} id - ID of the feed item to update
+     * @param {Function} errorCallback - the error callback
+     * @return {void}
+     */
+    createCommentErrorCallback = (e: $AxiosXHR<any>, id: string, errorCallback: Function) => {
+        const errorMessage =
+            e.status === 409 ? messages.commentCreateConflictMessage : messages.commentCreateErrorMessage;
```
JustinHoldstock: Make a const for the `409`?

src/api/Feed.js
```diff
@@ -0,0 +1,1022 @@
...
+
+    /**
+     * Destroys all the task assignment API's
+     *
+     * @return {void}
+     */
+    destroyTaskAssignments() {
+        if (Array.isArray(this.taskAssignmentsAPI)) {
+            this.taskAssignmentsAPI.forEach((api) => api.destroy());
+            this.taskAssignmentsAPI = [];
```
JustinHoldstock: What's our standard on emptying arrays? Creating a new array, or reusing the one we have via `array.length =  0`?

src/components/ContentSidebar/ActivitySidebar.js
```diff
@@ -17,73 +24,435 @@ type ExternalProps = {
...
+            permissions,
+            (comment: Comment) => {
+                this.feedSuccessCallback();
+                onCommentDelete(comment);
+            },
+            this.feedErrorCallback
+        );
+
+        // need to load the pending item
+        this.fetchFeedItems();
```
JustinHoldstock: Since `feedSuccessCallback()` and `feedErrorCallback()` call `fetchFeedItems()` is this necessary? Unless the comment explains that... :)

src/api/Feed.js
```diff
@@ -0,0 +1,1022 @@
...
+    }
+
+    /**
+     * Creates a key for the cache
+     *
+     * @param {string} id folder id
+     * @return {string} key
+     */
+    getCacheKey(id: string): string {
+        return `feedItems_${id}`;
```
priyajeet: Probably move this in constants.js with other keys

src/api/Feed.js
```diff
@@ -0,0 +1,1022 @@
...
+     * Gets the feed items
+     *
+     * @param {BoxItem} file - The file to which the task is assigned
+     * @param {Function} successCallback - the success callback
+     * @param {Function} errorCallback - the error callback
+     */
+    feedItems(file: BoxItem, successCallback: Function, errorCallback: Function): void {
+        const { id } = file;
+
+        if (!id) {
```
priyajeet: Needed anymore?

src/api/Feed.js
```diff
@@ -0,0 +1,1022 @@
...
+        const versionsPromise = this.fetchVersions();
+        const commentsPromise = this.fetchComments();
+        const tasksPromise = this.fetchTasks();
+
+        Promise.all([versionsPromise, commentsPromise, tasksPromise]).then((feedItems) => {
+            const versions: ?FileVersions = feedItems[0];
+            const versionsWithRestoredVersion = this.addCurrentVersion(versions, file);
+            const unsortedFeedItems = [versionsWithRestoredVersion, ...feedItems.slice(1)];
+            const sortedFeedItems = this.sortFeedItems.apply(null, unsortedFeedItems);
+            this.setCachedItems(id, sortedFeedItems);
```
priyajeet: Should this line above be guarded too in the if () block below?
Though maybe priming the cache for future might be useful.

src/api/Feed.js
```diff
@@ -0,0 +1,1022 @@
...
+    }
+
+    /**
+     * Sort valid feed items, descending by created_at time.
+     *
+     * @param {Array<?Comments | ?Tasks | ?FileVersions>} args - Arguments list of each item container
+     * type that is allowed in the feed.
+     * @return {Array<?Comments | ?Tasks | ?FileVersions>} the sorted feed items
+     */
+    sortFeedItems(...args: Array<?Comments | ?Tasks | ?FileVersions>): FeedItems {
```
priyajeet: Maybe move this inside util/sorter.js?

src/api/Feed.js
```diff
@@ -0,0 +1,1022 @@
...
+    }
+
+    /**
+     * Fetches the comments for a file
+     *
+     * @return {Promise} - the file comments
+     */
+    fetchComments(): Promise<?Comments> {
+        this.commentsAPI = new CommentsAPI(this.options);
+        return new Promise((resolve) => {
```
priyajeet: no async/await?

src/api/Feed.js
```diff
@@ -0,0 +1,1022 @@
...
+
+        if (this.tasksAPI) {
+            this.tasksAPI.destroy();
+            delete this.tasksAPI;
+        }
+
+        this.destroyTaskAssignments();
+    }
+}
+
```
priyajeet: Pretty large file, should one have individual files for comments / tasks etc?

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -666,611 +442,6 @@ class ContentSidebar extends PureComponent<Props, State> {
...
-        if (!file) {
-            throw getBadItemError();
-        }
-
-        this.api.getTasksAPI(false).createTask({
-            file,
-            message: text,
-            dueAt,
-            successCallback: (task: Task) => {
-                this.createTaskSuccessCallback(task, assignees, successCallback, errorCallback);
```
priyajeet: 😍

src/api/Feed.js
DanDeMicco: I technically could, but It is a pattern within elements to make a new API for each request, so they are cancelable

src/api/Feed.js
DanDeMicco: If the API is destroyed, the parent shouldnt care about the response anymore

src/api/Feed.js
DanDeMicco: yea that would be cleaner 👍

src/api/Feed.js
DanDeMicco: I have heard of this approach but have not dug into the details of what happens when you set the length to 0. Does it remove all the references?

src/components/ContentSidebar/ActivitySidebar.js
DanDeMicco: Those are called after the call is completed. We need to fetch the feed items immediately after the API is invoked, since the cache is updated with the pending item and we want to update the UI

src/api/Feed.js
DanDeMicco: This was copy + paste from another file, but saw the only thing which referenced this was in this file, so just ```.toString```'d it in the fields file and exported that 👍

src/api/Feed.js
DanDeMicco: done

src/api/Feed.js
DanDeMicco: done

src/api/Feed.js
DanDeMicco: Nope! removed

src/api/Feed.js
```diff
@@ -0,0 +1,1022 @@
...
+            this.commentsAPI.destroy();
+            delete this.commentsAPI;
+        }
+
+        if (this.versionsAPI) {
+            this.versionsAPI.destroy();
+            delete this.versionsAPI;
+        }
+
+        if (this.tasksAPI) {
```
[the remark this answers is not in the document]
DanDeMicco: Yep its done as part of ```this.destroyTaskAssignments();```

src/api/APIFactory.js
```diff
@@ -20,6 +20,7 @@ import FileAccessStatsAPI from './FileAccessStats';
 import UsersAPI from './Users';
 import MetadataAPI from './Metadata';
 import FileCollaboratorsAPI from './FileCollaborators';
+import FeedAPI from './Feed';
```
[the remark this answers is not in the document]
DanDeMicco: Technically they aren't needed in APIFactory anymore. I will create another ticket to clean this file up

src/api/Feed.js
DanDeMicco: Can you give an example of what you have in mind? Mixin the prototypes in the classes to Feed via Object.assign?

src/api/Feed.js
DanDeMicco: done

src/api/Feed.js
DanDeMicco: done

src/api/Feed.js
jstoffan: Should the various component services return a promise directly rather than wrapping it at the consumer level? Or is that not possible due to async/await being used in the low-level modules?

src/api/Feed.js
```diff
@@ -0,0 +1,1022 @@
...
+
+        return new Promise((resolve) => {
+            this.tasksAPI.get({
+                id: this.id,
+                successCallback: (tasks) => {
+                    this.fetchTaskAssignments(tasks).then(resolve);
+                },
+                errorCallback: (err) => {
+                    this.errorCallback(err);
+                    resolve();
```
jstoffan: Should these instead call `reject`?

src/api/Feed.js
```diff
@@ -0,0 +1,1022 @@
...
+        errorCallback: Function
+    ): void => {
+        if (!file.id) {
+            throw getBadItemError();
+        }
+
+        this.id = file.id;
+        this.updateFeedItem({ isPending: true }, taskId);
+        const taskAssignmentsAPI = new TaskAssignmentsAPI(this.options);
+        this.taskAssignmentsAPI.push(taskAssignmentsAPI);
```
jstoffan: I find this logic a bit confusing. The instance property name is the same as the local variable name, but their data models are totally different. Can we clarify the naming?

src/api/Feed.js
```diff
@@ -0,0 +1,1022 @@
...
+            return;
+        }
+
+        this.id = id;
+        this.hasError = false;
+        const versionsPromise = this.fetchVersions();
+        const commentsPromise = this.fetchComments();
+        const tasksPromise = this.fetchTasks();
+
+        Promise.all([versionsPromise, commentsPromise, tasksPromise]).then((feedItems) => {
```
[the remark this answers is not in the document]
DanDeMicco: That is a possibility and a small change, but I do not see that being a use case (yet) and the goal is to keep the API's self contained. I think it can be revisited when the use case comes up

src/api/Feed.js
```diff
@@ -0,0 +1,1022 @@
...
+     * @return {Array<?Comments | ?Tasks | ?FileVersions>} the sorted feed items
+     */
+    sortFeedItems(...args: Array<?Comments | ?Tasks | ?FileVersions>): FeedItems {
+        const feedItems = [];
+        args.forEach((itemContainer) => {
+            if (itemContainer) {
+                feedItems.push(...itemContainer.entries);
+            }
+        });
+        return feedItems.sort((a, b) => Date.parse(b.created_at) - Date.parse(a.created_at));
```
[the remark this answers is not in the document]
DanDeMicco: Agreed, but I wanted to avoid refactoring each and every method. Some of these methods are direct copy + paste from existing files. I can take a look at this function anyways

src/api/Feed.js
DanDeMicco: It was intentional to call resolve. Since I am using Promise.all above, a reject would immediately resolve the promise and not let the partial state be there

src/api/Feed.js
DanDeMicco: Lower level API's use callbacks, so thats why I wrapped it in a promise for convenience

src/api/Feed.js
DanDeMicco: 👍 will change it up

PR #496, opened 2018-08-06
Breaking: Metadata UI design fixes

priyajeet:
Improves sidebar toggling behavior
Fixes examples page with new file/folder ids
Upgrades box-react-ui peer dependency

### DIFF
--- a/src/components/ContentSidebar/MetadataSidebar.js
+++ b/src/components/ContentSidebar/MetadataSidebar.js
@@ -7,6 +7,7 @@
 import * as React from 'react';
 import { FormattedMessage } from 'react-intl';
 import MetadataInstanceEditor from 'box-react-ui/lib/features/metadata-instance-editor/MetadataInstanceEditor';
+import TemplateDropdown from 'box-react-ui/lib/features/metadata-instance-editor/TemplateDropdown';
 import LoadingIndicator from 'box-react-ui/lib/components/loading-indicator/LoadingIndicator';
 import LoadingIndicatorWrapper from 'box-react-ui/lib/components/loading-indicator/LoadingIndicatorWrapper';
 import InlineError from 'box-react-ui/lib/components/inline-error/InlineError';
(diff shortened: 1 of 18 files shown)


boxcla:

boxcla:

Compared against previously merged PR #495

src/components/ContentSidebar/MetadataSidebar.scss
```diff
@@ -1,9 +1,30 @@
+@import '../variables';
+
 .be.bcs.bcs-metadata {
+    .bcs-metadata-header {
+        align-items: center;
+        border-bottom: 1px solid $off-white;
+        display: flex;
+        height: 60px;
```
jstoffan: It looks like there's duplication here and in `SidebarTitle.scss`. Should there be two components, a header (display, height, margin, etc.) and a title (font-size and font-weight)? That way the size of the header will always be consistent between panels.

src/components/ContentSidebar/SkillsSidebar.js
```diff
@@ -174,7 +175,13 @@ class SkillsSidebar extends React.PureComponent<Props, State> {
         const { cards, errors }: State = this.state;
 
         return (
-            <SidebarContent title={<FormattedMessage {...messages.sidebarSkillsTitle} />}>
+            <SidebarContent
+                title={
+                    <SidebarTitle>
```
jstoffan: Would naming this component `<SidebarContentTitle>` make where it's intended to be used more explicit? The semantics styling seems very specific to this area.

src/components/makeResponsive.js
```diff
@@ -72,9 +73,9 @@ function makeResponsive(Wrapped: React.ComponentType<any>): React.ComponentType<
          * @param {Component} react component
          * @return {void}
          */
-        onResize = ({ bounds: { width } }: { bounds: ClientRect }) => {
+        onResize = debounce(({ bounds: { width } }: { bounds: ClientRect }) => {
             this.setState({ size: this.getSize(width) });
-        };
+        }, 500);
```
jstoffan: Is this low enough for the change in width to not be noticeable when the window is resized continually (back-and-forth, for example)?

src/components/ContentSidebar/MetadataSidebar.js
```diff
@@ -157,9 +158,28 @@ class MetadataSidebar extends React.PureComponent<Props, State> {
...
+                            <FormattedMessage {...messages.sidebarMetadataTitle} />
+                        </h3>
+                        {showEditor &&
+                            canEdit && (
+                                <TemplateDropdown
+                                    hasTemplates={templates && templates.length !== 0}
+                                    isDropdownBusy={false}
+                                    onAdd={this.onAdd}
+                                    templates={templates}
+                                    usedTemplates={editors && editors.map((editor) => editor.template)}
```
jstoffan: Minor: It looks like this `editors` truthy check and the various default parameter assignments could be replaced by setting `editors: []` in the initialState instance property.

priyajeet: Unfortunately don't even need that since I have this guarded with showEditor boolean. But flow hates that.

src/components/makeResponsive.js
priyajeet: It's a bit noticeable, we can adjust later if someone complains, but the use case of window resizing continuously is more of a corner case.

PR #502, opened 2018-08-07
Fix: Allow preview to error on failed files call

jeremypress:
Advantages to this approach
1. Preview will now retry with a more minimal files call
2. Error logging will still be captured
3. Error UI is provided by preview

With this change, the sidebar will still not render if the original files call fails. So there could be cases where preview succeeds on the minimal files call but the sidebar still fails. Open to suggestions on how to render a minimal error sidebar, but right now all the sidebar render logic relies on the presence of a file object.

Tests on they way.

### DIFF
--- a/src/components/ContentPreview/ContentPreview.js
+++ b/src/components/ContentPreview/ContentPreview.js
@@ -76,7 +76,8 @@ type Props = {
 };
 
 type State = {
-    file?: BoxItem
+    file?: BoxItem,
+    fileError?: boolean
 };
 
 const InvalidIdError = new Error('Invalid id for Preview!');
@@ -215,16 +216,19 @@ class ContentPreview extends PureComponent<Props, State> {
      * @return {boolean}
      */
     shouldLoadPreview(prevProps: Props, prevState: State): boolean {
-        const { file }: State = this.state;
+        const { file, fileError }: State = this.state;
         const { file: prevFile }: State = prevState;
         let loadPreview = false;
 
         // Load preview if file version ID has changed
         if (file && file.file_version && prevFile && prevFile.file_version) {
             loadPreview = file.file_version.id !== prevFile.file_version.id;
-        } else {
+        } else if (file) {
             // Load preview if file object has newly been popuplated in state
             loadPreview = !prevFile && !!file;
+        } else {
+            // Load preview if there was a problem with the Elements files call
+            loadPreview = !!fileError;
         }
 
         return loadPreview;
@@ -423,11 +427,13 @@ class ContentPreview extends PureComponent<Props, State> {
         const { token: tokenOrTokenFunction, collection, ...rest }: Props = this.props;
         const { file }: State = this.state;
 
-        if (!this.isPreviewLibraryLoaded() || !file || !tokenOrTokenFunction) {
+        const fileId = file ? this.getFileId(file) : rest.fileId;
+
+        if (!this.isPreviewLibraryLoaded() || !tokenOrTokenFunction) {
             return;
         }
 
-        const typedId: string = getTypedFileId(this.getFileId(file));
+        const typedId: string = getTypedFileId(fileId);
         const token: TokenLiteral = await TokenService.getReadToken(typedId, tokenOrTokenFunction);
 
         const previewOptions = {
@@ -443,8 +449,11 @@ class ContentPreview extends PureComponent<Props, State> {
             this.setPreviewInstance(new Preview());
         }
 
-        this.preview.updateFileCache([file]);
-        this.preview.show(file.id, token, {
+        if (file) {
+            this.preview.updateFileCache([file]);
+        }
+
+        this.preview.show(fileId, token, {
             ...previewOptions,
             ...omit(rest, Object.keys(previewOptions))
         });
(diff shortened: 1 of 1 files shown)


boxcla:

boxcla:

Compared against previously merged PR #501

src/components/ContentPreview/ContentPreview.js
```diff
@@ -228,16 +229,19 @@ class ContentPreview extends PureComponent<Props, State> {
...
         const { file: prevFile }: State = prevState;
         let loadPreview = false;
 
         // Load preview if file version ID has changed
         if (file && file.file_version && prevFile && prevFile.file_version) {
             loadPreview = file.file_version.id !== prevFile.file_version.id;
-        } else {
+        } else if (file) {
             // Load preview if file object has newly been popuplated in state
             loadPreview = !prevFile && !!file;
```
jstoffan: Is `!!file` needed here?

src/components/ContentPreview/ContentPreview.js
```diff
@@ -228,16 +229,19 @@ class ContentPreview extends PureComponent<Props, State> {
...
-        const { file }: State = this.state;
+        const { file, fileError }: State = this.state;
         const { file: prevFile }: State = prevState;
         let loadPreview = false;
 
         // Load preview if file version ID has changed
         if (file && file.file_version && prevFile && prevFile.file_version) {
             loadPreview = file.file_version.id !== prevFile.file_version.id;
-        } else {
+        } else if (file) {
```
jstoffan: Minor: It seems like this should be an outer check or a guard clause instead, since it's also checked above.

src/components/ContentPreview/ContentPreview.js
```diff
@@ -493,7 +502,17 @@ class ContentPreview extends PureComponent<Props, State> {
...
+    };
+
+    /**
+     * File fetch success callback
+     *
+     * @param {Object} file - Box file
+     * @return {void}
+     */
+    fetchFileErrorCallback = (): void => {
+        this.setState({ fileError: true });
```
priyajeet: Question, do we need to set the fileError state? Can we just try to reload the preview directly from the error callback? Setting the state will cause a re-render for react, and it seems that this variable is being used for BL

priyajeet: We also should add the retry logic in here like its in preview library. It tries to load it 5 times, and then dies.

priyajeet: You can maintain it in `this` and not necessarily need a state variable.
this.retryCount = 0 on successful fetch.

In that way you can eventually remove the logic from preview library.

jeremypress: We could do retry here, but we don't set up any loading UI / the progress bar until we call preview.show(), so we would be blocked on rendering anything. Eventually we should expose that logic so that the element can control that kind of behavior. 

In the interim I could set a class property instead of using state as you suggested to avoid the re-render.

src/components/ContentPreview/ContentPreview.js
jeremypress: nice, I refactored to check for existence of the file first so we only have to check once.

src/components/ContentPreview/ContentPreview.js
```diff
@@ -219,8 +220,10 @@ class ContentPreview extends PureComponent<Props, State> {
         const { file: prevFile }: State = prevState;
         let loadPreview = false;
 
-        // Load preview if file version ID has changed
-        if (file && file.file_version && prevFile && prevFile.file_version) {
+        if (!file) {
+            loadPreview = !!this.fileError;
+            // Load preview if file version ID has changed
```
priyajeet: Can you move comment inside the below block

priyajeet: last if/else block `loadPreview = !prevFile;` since !!file is always true

priyajeet:
```js
if (!file) { loadPreview = !!this.fileError; }
else if (!prevFile) { loadPreview = true; }
else { loadPreview = file.file_version.id !== prevFile.file_version.id; }
```

I think can be simplified to that, file_version should always be on file object, right? If not the check can remain

src/components/ContentPreview/ContentPreview.js
```diff
@@ -423,11 +426,13 @@ class ContentPreview extends PureComponent<Props, State> {
         const { token: tokenOrTokenFunction, collection, ...rest }: Props = this.props;
```
priyajeet:
```js
const { token: tokenOrTokenFunction, collection, fileId, ...rest }: Props = this.props;
```

src/components/ContentPreview/ContentPreview.js
```diff
@@ -423,11 +426,13 @@ class ContentPreview extends PureComponent<Props, State> {
...
         const { file }: State = this.state;
 
-        if (!this.isPreviewLibraryLoaded() || !file || !tokenOrTokenFunction) {
+        const fileId = file ? this.getFileId(file) : rest.fileId;
```
priyajeet: extract fileId out

PR #509, opened 2018-08-14
Update: Improve file fetch isolation across elements

priyajeet:
Fetching file details are now isolated between explorer/picker, preview and sidebar.
The file API will fetch file fields that are missing from the cached file object.

eg:

If explorer wants { a, b } => it will request for a and b
If preview wants { a, b, c } => it will request for b and c using cache from explorer if available
If sidebar wants { b, c, d } => it will request for d using cache from preview if available

Also changes babelrc to have presets per environment so that we target latest chrome and firefox only for dev builds which makes testing easier. It will not need babel transform to use generators as such break points will work for debugging. Only will work in dev mode.

**TODO: Fix broken tests / add more tests**

### DIFF
--- a/src/api/Folder.js
+++ b/src/api/Folder.js
@@ -10,8 +10,8 @@ import flatten from '../util/flatten';
 import sort from '../util/sorter';
 import FileAPI from '../api/File';
 import WebLinkAPI from '../api/WebLink';
-import { getFieldsAsString } from '../util/fields';
-import { CACHE_PREFIX_FOLDER, X_REP_HINTS } from '../constants';
+import { CACHE_PREFIX_FOLDER } from '../constants';
+import { FOLDER_FIELDS_TO_FETCH } from '../util/fields';
 import { getBadItemError } from '../util/error';
 
 const LIMIT_ITEM_FETCH = 1000;
@@ -56,16 +56,6 @@ class Folder extends Item {
      * @property {Function}
      */
     errorCallback: Function;
-
-    /**
-     * @property {boolean}
-     */
-    includePreviewFields: boolean;
-
-    /**
-     * @property {boolean}
-     */
-    includePreviewSidebarFields: boolean;
 
     /**
      * Creates a key for the cache
@@ -220,9 +210,8 @@ class Folder extends Item {
                 params: {
                     offset: this.offset,
                     limit: LIMIT_ITEM_FETCH,
-                    fields: getFieldsAsString(this.includePreviewFields, this.includePreviewSidebarFields)
-                },
-                headers: { 'X-Rep-Hints': X_REP_HINTS }
+                    fields: FOLDER_FIELDS_TO_FETCH.toString()
+                }
             })
             .then(this.folderSuccessHandler)
             .catch(this.errorHandler);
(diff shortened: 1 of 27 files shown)


boxcla:

src/api/Folder.js
```diff
@@ -237,19 +226,17 @@ class Folder extends Item {
      * @param {Function} successCallback - Function to call with results
      * @param {Function} errorCallback - Function to call with errors
      * @param {boolean|void} [forceFetch] - Bypasses the cache
-     * @param {boolean|void} [includePreview] - Optionally include preview fields
-     * @param {boolean|void} [includePreviewSidebar] - Optionally include preview sidebar fields
+     * @param {boolean|void} [includePreviewFields] - Optionally include preview fields
```
jstoffan: These params are no longer present in the function signature. Also, `options` is missing.

src/api/Folder.js
```diff
@@ -237,19 +226,17 @@ class Folder extends Item {
      * @param {Function} successCallback - Function to call with results
      * @param {Function} errorCallback - Function to call with errors
      * @param {boolean|void} [forceFetch] - Bypasses the cache
```
jstoffan: No longer present in the function signature.

src/api/Folder.js
```diff
@@ -237,19 +226,17 @@ class Folder extends Item {
...
+    getFolder(
         id: string,
         sortBy: SortBy,
         sortDirection: SortDirection,
         successCallback: Function,
         errorCallback: Function,
-        forceFetch: boolean = false,
-        includePreviewFields: boolean = false,
-        includePreviewSidebarFields: boolean = false
+        options: Object = {}
```
jstoffan: Should this be typed to indicate what options are supported?

priyajeet: yup yup

src/util/fields.js
```diff
@@ -124,75 +129,32 @@ export const COMMENTS_FIELDS_TO_FETCH = [
...
+        return fields;
     }
 
-    const fields = getFields(includePreview, includePreviewSidebar).filter(
-        (field) => BOX_ITEM_OPTIONAL_FIELDS.indexOf(field) < 0
-    );
-
-    // Some fields like metadata have dots in it. Only use the root prop.
-    return fields.every((field) => typeof getProp(file, field.split('.')[0]) !== 'undefined');
+    return fields.filter((field: string) => {
```
jeremypress: Why not rely on the whole string? We may end up at a point where different components want different metadata fields.

test/preview.html
```diff
@@ -19,7 +19,7 @@
         <link rel="stylesheet" type="text/css" href="../dev/en-US/explorer.css" />
         <script src="../dev/en-US/explorer.js"></script>
         <script>
-            var ACCESS_TOKEN = 'htXetWuPFla33pfZiEFr4WKaOPIjH4ye';
```
jeremypress: 👍

src/util/fields.js
priyajeet: Because when you make a request for ?fields=foo.bar.baz, the API will return `{ foo: null }` suggesting there is no `foo` and not `{ foo: { bar: null } }` suggesting there is no `baz`. So if I use the full path, then it will always consider `foo.bar.baz` to be missing. Only solution is that I need to check `foo.bar.baz` and then `foo.bar` and then `foo` to correctly figure out the missing field.

priyajeet: Looks pretty ugly/hacky

```js
function findMissingFields(file, fields) {
    return fields.filter((field) => {
        let hasField = _.has(file, field);
        let hasParent = false;

        while (field && !hasField) {
            field = field.substring(0, field.lastIndexOf('.'));
            hasParent = _.has(file, field)
        }

        return !hasField || hasParent;
    });
}
```

src/util/Cache.js
```diff
@@ -39,7 +40,7 @@ class Cache {
      */
     merge(key: string, value: any): void {
         if (this.has(key)) {
-            this.set(key, Object.assign({}, this.get(key), value));
+            this.set(key, merge({}, this.get(key), value));
         } else {
             throw new Error(`Key ${key} not in cache!`);
```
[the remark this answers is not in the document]
priyajeet: Specifically kept that for debugging purposes and enforcing usage in the correct way. Want merge to be a conscious decision to merge into something that exists. Can re-visit later.

boxcla:

Compared against previously merged PR #513

src/components/ContentExplorer/ContentExplorer.js
```diff
@@ -1354,4 +1344,5 @@ class ContentExplorer extends Component<Props, State> {
     }
 }
 
+export { ContentExplorer as ContentExplorerComponent };
```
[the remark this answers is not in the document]
priyajeet: This is the pattern we generally follow in other places.

src/api/Recents.js
```diff
@@ -208,9 +195,7 @@ class Recents extends Base {
         sortDirection: SortDirection,
         successCallback: Function,
         errorCallback: Function,
-        forceFetch: boolean = false,
-        includePreviewFields: boolean = false,
-        includePreviewSidebarFields: boolean = false
+        options: Object = {}
```
jstoffan: Should this use `FetchOptions` type as well?

PR #511, opened 2018-08-15
Fix: Filtering to only add load time to load event, if valid

JustinHoldstock:
Filter so that we:
A) We don't attach file_info_time to *all* preview metric events
B) We don't log invalid load times

### DIFF
--- a/src/components/ContentPreview/ContentPreview.js
+++ b/src/components/ContentPreview/ContentPreview.js
@@ -93,12 +93,24 @@ type PreviewTimeMetrics = {
 };
 
 // Emitted by preview's 'preview_metric' event
-type PreviewLoadMetrics = {
+type PreviewMetrics = {
+    error?: Object,
+    event_name?: string,
     value: number, // Sum of all available load times.
     file_info_time: number,
     convert_time: number,
     download_response_time: number,
-    full_document_load_time: number
+    full_document_load_time: number,
+    timestamp: string,
+    file_id: string,
+    file_version_id: string,
+    content_type: string,
+    extension: string,
+    locale: string,
+    rep_type: string,
+    client_version: string,
+    browser_name: string,
+    logger_session_id: string
 };
 
 const InvalidIdError = new Error('Invalid id for Preview!');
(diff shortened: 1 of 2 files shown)


boxcla:

boxcla:

Compared against previously merged PR #508

src/components/ContentPreview/ContentPreview.js
```diff
@@ -425,22 +437,37 @@ class ContentPreview extends PureComponent<Props, State> {
...
-            ...previewLoadMetrics,
-            file_info_time: totalFetchFileTime,
-            value: (previewLoadMetrics.value || 0) + totalFetchFileTime
-        });
+        if (event_name === 'load') {
+            const totalFetchFileTime = this.getTotalFileFetchTime();
+            const totalTime = (previewMetrics.value || 0) + totalFetchFileTime;
+
+            // If an unnatural fetch time occurs or is invalid, don't log the event
+            if (!totalFetchFileTime || !totalTime) {
```
DanDeMicco: Is it possible for ```totalFetchFileTime``` to be 0? What is the harm of emitting the event if the fetch time is 0?

jeremypress: the Elements cache is in memory instead of the browser cache. I could see it being 0 depending on how performance.now() rounds

src/components/ContentPreview/ContentPreview.js
```diff
@@ -425,22 +437,37 @@ class ContentPreview extends PureComponent<Props, State> {
...
+        };
 
         // We need to add in the total file fetch time to the file_info_time and value (total)
         // as preview does not do the files call
-        onMetric({
-            ...previewLoadMetrics,
-            file_info_time: totalFetchFileTime,
-            value: (previewLoadMetrics.value || 0) + totalFetchFileTime
-        });
+        if (event_name === 'load') {
```
jeremypress: Would this make sense as a constant? It's probably somewhere else in Elements already

src/components/ContentPreview/ContentPreview.js
JustinHoldstock: It shouldn't be possible for it to take 0 milliseconds to grab file info, even from browser cache. If it is 0, I would assume that something went wrong: when getting it from the cache OR requesting it via the API

DanDeMicco: Can you quickly verify that when navigating between previews that the time is never 0 when the item is in the cache? Cache, meaning local cache (aka Cache.js), not browser cache?

jeremypress: @JustinHoldstock any update on this?

JustinHoldstock: I've confirmed that we can have 0 ms values when a file is cached. We may need to rethink strategy on reducing redundant load time calls OR just begin filtering out total load times of 0 when calculating our load metrics (from our analytics queries). I'm leaning towards the latter, and removing this guard

PR #510, opened 2018-08-14
Chore: removed display style from bcs-comment-mention

JustinHoldstock:
Doesn't look like `display: inline-block` makes a difference in any browser. Images below are with it removed.

Chrome:
screen shot 2018-08-14 at 10 41 37 am

IE 11:
screen shot 2018-08-14 at 10 41 20 am

Safari:
screen shot 2018-08-14 at 2 28 12 pm

Firefox:
screen shot 2018-08-14 at 2 31 31 pm

PR #512, opened 2018-08-16
Fix: Destroy plain upload API on cancel

jeremypress:


boxcla:

boxcla:

Compared against previously merged PR #510

src/api/uploads/MultiputPart.js
```diff
@@ -298,10 +298,6 @@ class MultiputPart extends BaseMultiput {
      * @return {void}
      */
     cancel(): void {
-        if (this.xhr && typeof this.xhr.abort === 'function') {
```
jeremypress: We don't need to abort since it's done in this.destroy below

src/api/uploads/PlainUpload.js
```diff
@@ -156,11 +156,8 @@ class PlainUpload extends BaseUpload {
             return;
         }
 
-        if (this.xhr && typeof this.xhr.abort === 'function') {
-            this.xhr.abort();
```
jeremypress: Same here

src/api/uploads/MultiputPart.js
pramodsum: Can you add the same safety check around the abort in super.destroy()?

jeremypress: According to @DanDeMicco flow should be protecting us here.

PR #519, opened 2018-08-21
New: Adding Open With skeleton, wrapper, and test file

jeremypress:

PR #520, opened 2018-08-22
Chore: Adding Open With network layer

jeremypress:


boxcla:

boxcla:

src/api/OpenWith.js
```diff
@@ -0,0 +1,25 @@
...
+     *
+     * @param {string} [id] - a box file id
+     * @return {string} base url for files
+     */
+    getUrl(id: string): string {
+        if (!id) {
+            throw new Error('Missing file id!');
+        }
+
+        return `${this.getBaseApiUrl()}/files/${id}/open_with_integrations`;
```
priyajeet: Should this inherit from File.js?

src/api/Base.js
```diff
@@ -341,9 +341,8 @@ class Base {
 
         // $FlowFixMe
         const xhrMethod: Function = this.xhr[method].bind(this.xhr);
-
         try {
-            const { data } = await xhrMethod({ id: getTypedFileId(id), url, ...requestData });
+            const { data } = await xhrMethod({ ...requestData, id: getTypedFileId(id), url });
```
priyajeet: Not sure if this should be done. And likely will break stuff. Because now you are sending a fileId always and in some cases we want a folder id. Don't think any id should be sent here, and callers should be sending those

priyajeet: Base.js should not have anything related to file or folder

jeremypress: I'll move it back

The reason why we set the ID here is to help enforce that the ID is getting typed correctly. If folder related stuff wants to start using this then we will need to move this out.

src/api/OpenWith.js
jeremypress: the getURL method in file assumes you're only passing in an ID, we would need to pass in a full suffix. We're not doing any other operations related to the file with this endpoint, so I think it's fine to leave as is.

src/api/AppIntegrations.js
```diff
@@ -0,0 +1,60 @@
...
+        if (!fileId) {
+            throw new Error('Missing file id!');
+        }
+
+        const executeURL = `${this.getUrl(integrationId)}/execute`;
+        const body = {
+            data: {
+                item: {
+                    id: fileId,
+                    type: 'file'
```
JustinHoldstock: Should we put `file` as a constant somewhere, and reuse it?

src/api/APIFactory.js
```diff
@@ -206,6 +218,21 @@ class APIFactory {
             delete this.fileCollaboratorsAPI;
         }
 
+        if (this.feedItemsAPI) {
```
jstoffan: If this change isn't related to this patchset, can we move it to another PR? It'll help ensure that changes don't get inadvertently lost if we need to revert.

src/api/AppIntegrations.js
```diff
@@ -0,0 +1,61 @@
...
+import { TYPE_FILE } from '../constants';
+
+class AppIntegrations extends Base {
+    /**
+     * API URL for Open With
+     *
+     * @param {string} [id] - a box integration app ID
+     * @return {string} base url for app integrations
+     */
+    getUrl(id: string): string {
```
jstoffan: Minor: Rename to `integrationId` to be consistent with the `execute` method.

PR #521, opened 2018-08-23
Fix: Replace missing modified_by with placeholder user

JustinHoldstock:


boxcla:

boxcla:

src/api/Versions.js
```diff
@@ -37,10 +38,14 @@ class Versions extends OffsetBasedAPI {
...
         const { entries } = data;
-        const versions = entries.map((version: BoxItemVersion) => ({
-            ...version,
-            action: version.trashed_at ? ACTION.delete : ACTION.upload
-        }));
+        const versions = entries.map((version: BoxItemVersion) => {
+            const { modified_by } = version;
+            return {
+                ...version,
+                modified_by: modified_by || PLACEHOLDER_USER,
```
jeremypress: Should we type this with flow?

JustinHoldstock:
Result:
screen shot 2018-08-27 at 4 19 39 pm

src/constants.js
```diff
@@ -236,3 +236,6 @@ export const UNAUTHORIZED_CODE = 403;
 export const VERSION_RESTORE_ACTION = 'restore';
 export const VERSION_DELETE_ACTION = 'delete';
 export const VERSION_UPLOAD_ACTION = 'upload';
+
+/* ------------------ Placeholder Feed Items ------------------------- */
+export const PLACEHOLDER_USER = { type: 'user', id: '2', name: '' };
```
jeremypress: Can we default this without adding this user? There may be cases where it's not actually an anonymous user so this data would be inaccurate.

JustinHoldstock: As discussed, leaving as is.

Issue #466, opened 2018-07-23
Getting "part_digest_retries_exceeded" error when uploading file with size > 50mb with Content Uploader


MaximeMarck:
Elements version: 4.4.0

Desktop:
 - OS: Windows 7
 - Browser: Chrome
 - Version: 67

Steps to reproduce the problem:
1. Create an HTML page to display a Content Uploader such as the one provided by documentation https://codepen.io/box-platform/pen/QvqGwr
2. Update the ACCESS_TOKEN to use a generated Developer Token 
3. Go to the page and try to upload a document with a size greater than 50 Mb

What is the expected behavior? 
Document should be uploaded.

What went wrong?
Content uploader displaying "Something went wrong with the upload. Please try again".

In the Network tab of the console, I can see that a call is made to https://upload.box.com/api/2.0/files/upload_sessions and the server sends back a response with the parts and endpoints.

Then normally several calls are made to endpoints to send documents but nothing here.
After a few minutes 2 calls are made to https://upload.box.com/api/2.0/files/upload_sessions/[ID]/log 
with this request payload: {"event_type":"part_digest_retries_exceeded","event_info":"{}"}

Then a HTTP Delete is issued to https://upload.box.com/api/2.0/files/upload_sessions/[ID]

Additional context:
The page is hosted on an internal server named http://webtest/sitename
http://webtest was added to the CORS allowed origins.
Small files uploads works properly.

Same page works in Internet Explorer 11 and Firefox.
It doesn't work in Chrome. 

The Codepen example works properly in Chrome so it seems to be related to the fact that the Content Uploader is hosted on an internal server?

DanDeMicco:
Hey @MaximeMarck this was fixed in v5.0 https://github.com/box/box-ui-elements/releases, can you try updating your version to v5.1.0 https://cdn01.boxcdn.net/platform/elements/5.1.0/en-US/uploader.js

DanDeMicco:
Thats strange it works everywhere but your server. When you tried 5.1.0 on your server, did you use the CDN or npm? 

Can you try seeing what the error messages are in the console? You should be able to send ```consoleError:true```  as a prop when initializing

DanDeMicco:
Just investigated this with @wenboyu2 it seems for whatever reason, when you are not using localhost, the upload fails when your server that is hosting the uploader is not on HTTPS. Can you try changing your local server to use HTTPS and give it another shot?

wenboyu2:
[...]

PR #525, opened 2018-08-29
Fix: uploader falls back to regular upload API if chunked unsupported

DanDeMicco:
#466

### DIFF
--- a/src/components/ContentUploader/UploaderUtils.js
+++ b/src/components/ContentUploader/UploaderUtils.js
@@ -1,0 +1,7 @@
+class UploaderUtils {
+    static isMultiputSupported() {
+        return window.location.protocol === 'https:' && window.crypto && window.crypto.subtle;
+    }
+}
+
+export default UploaderUtils;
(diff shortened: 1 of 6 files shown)


boxcla:

boxcla:

JustinHoldstock:
One quick note, does this work in IE 11? The MSDN page says that IE 11 requires the ms prefix

`var cryptoObj = window.crypto || window.msCrypto; // for IE 11`

src/components/ContentUploader/UploaderUtils.js
```diff
@@ -0,0 +1,7 @@
+class UploaderUtils {
+    static isMultiputSupported() {
```
wenboyu2: can this be a function in `src/util/uploads.js`?

src/components/ContentUploader/UploaderUtils.js
```diff
@@ -0,0 +1,7 @@
+class UploaderUtils {
```
wenboyu2: // @flow

src/components/ContentUploader/UploaderUtils.js
DanDeMicco: Didn't notice that file existed, oops! will move it there :)

wenboyu2: or maybe `webcrypto.js`

wenboyu2:
@JustinHoldstock that's a great point, we should also check for `msCrypto` and verify if you can still upload from IE 11.
`src/util/webcrypto.js L:13`

DanDeMicco:
Good catch, verified ie11 works correctly after making the change

PR #538, opened 2018-09-07
Fix: Add empty state when there is no metadata applied

priyajeet:


boxcla:

src/components/ContentSidebar/MetadataSidebar.js
```diff
@@ -217,6 +218,7 @@ class MetadataSidebar extends React.PureComponent<Props, State> {
         const showLoadingIndicator = !hasError && !showEditor;
         const canEdit = this.canEdit();
         const showTemplateDropdown = showEditor && canEdit;
+        const showEmptyContent = showEditor && ((editors: any): Array<MetadataEditor>).length === 0;
```
jstoffan: The inline casting here is hard on the eyes.

PR #540, opened 2018-09-07
Chore: fix occurred typos in messages

visylvius:
What this PR does:

Changes a typo in the messages key from occured to => occurred

PR #543, opened 2018-09-10
Chore: Refactor Open With based on API changes

jeremypress:


boxcla:

boxcla:

Compared against previously merged PR #540

src/components/OpenWith/OpenWith.js
```diff
@@ -112,7 +121,26 @@ class OpenWith extends PureComponent<Props, State> {
...
+     * @return {void}
+     */
+    fetchOpenWithData() {
+        const { fileId, extension }: Props = this.props;
+        if (extension) {
+            this.setState({ extension });
+        } else {
+            this.api
+                .getFileAPI()
+                .getFile(fileId, this.fetchExtensionSuccessCallback, noop, {
```
ConradJChan: There's no dependency between this call (when there's no extension) with the `getOpenWithIntegrations` call right?

jeremypress: Correct, the icon can async load before or after we get back integrations. We'll use a generic icon until this comes back.

src/components/OpenWith/OpenWith.js
```diff
@@ -112,16 +121,60 @@ class OpenWith extends PureComponent<Props, State> {
...
+    }
+
+    /**
+     * Fetches file extension and Open With data.
+     *
+     * @return {void}
+     */
+    fetchOpenWithData() {
+        const { fileId, extension, language }: Props = this.props;
+        if (extension) {
```
jstoffan: It seems like this operation should happen separately: once in the `initialState` construction and once again in componentDidUpdate if the extension changed.

src/components/OpenWith/OpenWith.js
```diff
@@ -112,16 +121,60 @@ class OpenWith extends PureComponent<Props, State> {
...
+    /**
+     * After component updates, re-fetch Open With data if appropriate.
+     *
+     * @return {void}
+     */
+    componentDidUpdate(prevProps: Props): void {
+        const { fileId: currentFileId }: Props = this.props;
+        const { fileId: previousFileId }: Props = prevProps;
+
+        if (currentFileId !== previousFileId) {
```
jstoffan: Would we only want to do this if the extension changes?

src/components/OpenWith/OpenWith.js
```diff
@@ -112,16 +121,60 @@ class OpenWith extends PureComponent<Props, State> {
...
-        const { fileId }: Props = this.props;
+        this.fetchOpenWithData();
+    }
+
+    /**
+     * Fetches file extension and Open With data.
+     *
+     * @return {void}
+     */
+    fetchOpenWithData() {
```
jstoffan: Should this be named `fetchFileExtension`?

src/api/OpenWith.js
```diff
@@ -5,14 +5,11 @@
  */
 
 import Base from './Base';
-import AppIntegrationsAPI from './AppIntegrations';
+// Temporary until API is fixed
```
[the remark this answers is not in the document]
jeremypress: yep, forgot to update.

src/components/OpenWith/OpenWith.js
jeremypress: It currently fetches the extension and the open with data

src/components/OpenWith/OpenWith.js
jeremypress: If the file ID changes, then we need to get new open with data and a new extension. Either that will be provided via props or we'll fetch for it.

src/components/OpenWith/OpenWith.js
jeremypress: The extension can/should only change if the file ID changes. That's why I have them grouped currently.

src/components/OpenWith/OpenWith.js
jstoffan: Ah, okay. I didn't realize the method was doing two different things. It seems like it should still be two methods to me, but it's not really a big deal.

PR #544, opened 2018-09-11
Breaking: Upgrade React to 16.5

priyajeet:
Upgrades few other packages and peer dependencies
Fixes snapshots due to enzyme changes
Updates some prettier styling
Upgrades preview version

React 16.3 --> 16.4/16.5 upgrade effects the use of use of `getDerivedStateFromProps`
 https://reactjs.org/blog/2018/05/23/react-v-16-4.html#bugfix-for-getderivedstatefromprops

> Essentially, getDerivedStateFromProps is now called every time a component is rendered, regardless of the cause of the update. Previously, it was only called if the component was re-rendered by its parent, and would not fire as the result of a local setState.


boxcla:

boxcla:

Compared against previously merged PR #543

PR #545, opened 2018-09-11
New: Render Open With button

jeremypress:
Todo:
[] Tests
[x] Refactor iconProps logic in the button component


boxcla:

src/components/OpenWith/OpenWithButton.js
```diff
@@ -0,0 +1,69 @@
...
+    displayIntegration,
+    tooltipText,
+    isDisabled = false,
+    isLoading = true,
+}: Props) => {
+    const { displayName = null, appIntegrationId = null } =
+        displayIntegration || {};
+    let DisplayIcon = ICON_FILE_MAP[displayName || 'default'];
+
+    let iconProps = {
```
jeremypress: Would love some feedback on how to handle this more cleanly. The issue here is that depending on what component we use FileIcon vs. a special integration icon e.g. IconAdobeSign, the props will be different.

height, width vs. dimension.

Also, the specific integration icons have a different viewbox, and require extra padding.

src/components/OpenWith/IconFileMap.js
```diff
@@ -0,0 +1,15 @@
+import IconAdobeSign from 'box-react-ui/lib/icons/adobe-sign/IconAdobeSign';
```
jeremypress: This will be reused by the dropdown. Eventually this may be removable depending on API changes.

src/components/OpenWith/OpenWith.js
```diff
@@ -61,6 +69,8 @@ class OpenWith extends PureComponent<Props, State> {
         extension: '',
         isDropdownOpen: false,
         integrations: null,
+        isLoading: true,
```
jstoffan: Should this also be set to true in `componentDidUpdate` if the file changed? Or is it initial load only?

src/components/OpenWith/OpenWith.js
```diff
@@ -208,22 +217,85 @@ class OpenWith extends PureComponent<Props, State> {
...
+     *
+     * @private
+     * @return {?string | ?Element}
+     */
+    getTooltip(): ?string | ?Element {
+        const { fetchError, isLoading, integrations }: State = this.state;
+        const displayIntegration = this.getDisplayIntegration();
+        if (isLoading) {
+            return null;
+        } else if (fetchError) {
```
jstoffan: Using `else if` isn't necessary with guard clauses, as the prior `return` will short-circuit the remaining conditions.

boxcla:

Compared against previously merged PR #544

src/components/OpenWith/OpenWith.js
```diff
@@ -208,22 +217,85 @@ class OpenWith extends PureComponent<Props, State> {
...
+        const displayIntegration = this.getDisplayIntegration();
+
         return (
-            integrations && (
-                <Internationalize language={language} messages={intlMessages}>
-                    <button> Open With </button>
-                </Internationalize>
-            )
+            <Internationalize language={language} messages={intlMessages}>
+                <div id={this.id} className="be bcow">
```
jstoffan: Should the class be `be-bcow`?

src/components/OpenWith/OpenWith.js
```diff
@@ -208,22 +217,85 @@ class OpenWith extends PureComponent<Props, State> {
...
+        /* no-op */
+    }
+
+    /**
+     * Gets the tooltip text for the OpenWith button
+     *
+     * @private
+     * @return {?string | ?Element}
+     */
+    getTooltip(): ?string | ?Element {
```
jstoffan: Should this live in the OpenWithButton component, instead? Its job is to basically examine the props passed in below and choose which text to display.

src/components/OpenWith/OpenWith.scss
```diff
@@ -0,0 +1,14 @@
+.be.bcow {
+    button {
```
jstoffan: Try to avoid tag names in CSS when possible.

src/components/OpenWith/OpenWithButton.js
```diff
@@ -0,0 +1,69 @@
...
+const OpenWithButton = ({
+    extension,
+    onClick,
+    displayIntegration,
+    tooltipText,
+    isDisabled = false,
+    isLoading = true,
+}: Props) => {
+    const { displayName = null, appIntegrationId = null } =
+        displayIntegration || {};
```
jstoffan: Consider assigning defaults values with the associated function params.

src/components/OpenWith/OpenWithButton.js
```diff
@@ -0,0 +1,69 @@
...
+    extension: ?string,
+    displayIntegration: ?Integration,
+    onClick: ?Function,
+    tooltipText?: string | Object,
+    isDisabled: boolean,
+    isLoading: boolean,
+    icon?: string,
+};
+
+const INTEGRATION_ICON_CLASS = 'integration-icon';
```
jstoffan: What's the value of having this as a constant? It isn't exported and seems to only be used once.

src/components/OpenWith/OpenWithButton.js
```diff
@@ -0,0 +1,69 @@
...
+    extension,
+    onClick,
+    displayIntegration,
+    tooltipText,
+    isDisabled = false,
+    isLoading = true,
+}: Props) => {
+    const { displayName = null, appIntegrationId = null } =
+        displayIntegration || {};
+    let DisplayIcon = ICON_FILE_MAP[displayName || 'default'];
```
jstoffan: Make sure to handle a case where we have a display name that doesn't map to a known key.

src/components/OpenWith/OpenWithButton.js
jstoffan: I think the simple solution would be to handle "default" separately:

```
const IntegrationIcon = displayName && ICON_FILE_MAP[displayName];

...(in render)
{ IntegrationIcon ? <IntegrationIcon height={26} width={26} /> : <FileIcon dimension={26} /> }
```

src/components/OpenWith/OpenWith.js
jeremypress: There's a couple of things not passed into the component that this function needs to consider. 
1. If there was an error
2. if there's more than one integration

Are you suggesting to include those and push the logic down to the component?

jstoffan: I'm not sure how the dropdown menu is going to play into this architecture, but it looks like this is set up as a container -> component relationship with OpenWithButton at the moment. This file is already doing a lot, so passing those props down and allowing the button to handle rendering logic doesn't seem all that bad to me.

src/components/OpenWith/OpenWith.js
jeremypress: The other elements have been adding them separately

src/components/OpenWith/OpenWith.js
jeremypress: yes, this should be reset whenever the file ID has changed.

src/components/OpenWith/OpenWithButton.js
jeremypress: no value, I thought I was going to use it in more than one place.

src/components/OpenWith/OpenWith.js
jeremypress: the BRUI dropdown accepts a child header item, so I'm expecting

if <=1 integration  <OpenWithButton />
if > 1 integration <OpenWithDropdown> <OpenWithButton /> <OpenWithDropdown>

I think passing an error and a length of integrations is reasonable.

PR #548, opened 2018-09-11
Update: adding placeholder user when not returned by API

ConradJChan:
Affected APIs are File, Comments, and Tasks. Looks for the user
fields `created_by`, `modified_by`, `assigned_to`, `owned_by`

### DIFF
--- a/src/api/Base.js
+++ b/src/api/Base.js
@@ -16,10 +16,7 @@ import {
     HTTP_POST,
     HTTP_PUT,
     HTTP_DELETE,
-    PLACEHOLDER_USER,
 } from '../constants';
-import set from 'lodash/set';
-import { USER_FIELDS } from '../util/fields';
 
 class Base {
     /**
@@ -395,27 +392,6 @@ class Base {
         } catch (error) {
             this.errorHandler(error);
         }
-    }
-
-    /**
-     * Fill user properties that are null in an object
-     *
-     * @param {Object} obj - some object
-     * @return {Object} new object with user placeholder
-     */
-    fillUserPlaceholder(obj: Object): Object {
-        const newObj = { ...obj };
-
-        USER_FIELDS.forEach((field: string) => {
-            if (
-                Object.prototype.hasOwnProperty.call(newObj, field) &&
-                newObj[field] === null
-            ) {
-                set(newObj, field, PLACEHOLDER_USER);
-            }
-        });
-
-        return newObj;
     }
 }
 
--- a/src/api/Base.js
+++ b/src/api/Base.js
@@ -16,7 +16,10 @@ import {
     HTTP_POST,
     HTTP_PUT,
     HTTP_DELETE,
+    PLACEHOLDER_USER,
 } from '../constants';
+import set from 'lodash/set';
+import { USER_FIELDS } from '../util/fields';
 
 class Base {
     /**
@@ -392,6 +395,27 @@ class Base {
         } catch (error) {
             this.errorHandler(error);
         }
+    }
+
+    /**
+     * Fill user properties that are null in an object
+     *
+     * @param {Object} obj - some object
+     * @return {Object} new object with user placeholder
+     */
+    fillUserPlaceholder(obj: Object): Object {
+        const newObj = { ...obj };
+
+        USER_FIELDS.forEach((field: string) => {
+            if (
+                Object.prototype.hasOwnProperty.call(newObj, field) &&
+                newObj[field] === null
+            ) {
+                set(newObj, field, PLACEHOLDER_USER);
+            }
+        });
+
+        return newObj;
     }
 }
 
--- a/src/api/Feed.js
+++ b/src/api/Feed.js
@@ -13,6 +13,7 @@ import {
     TASKS_FIELDS_TO_FETCH,
     VERSIONS_FIELDS_TO_FETCH,
     TASK_ASSIGNMENTS_FIELDS_TO_FETCH,
+    fillUserPlaceholder,
 } from '../util/fields';
 import CommentsAPI from './Comments';
 import VersionsAPI from './Versions';
@@ -24,6 +25,7 @@ import {
     VERSION_UPLOAD_ACTION,
     VERSION_RESTORE_ACTION,
     TYPED_ID_FEED_PREFIX,
+    PLACEHOLDER_USER,
 } from '../constants';
 import { sortFeedItems } from '../util/sorter';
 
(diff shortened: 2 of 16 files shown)


boxcla:

src/api/Tasks.js
```diff
@@ -195,6 +196,27 @@ class Tasks extends Base {
...
         });
     }
+
+    /**
+     * Overrides the Base.successHandler to handle filling in a placeholder user if
+     * task created_by or tas assignment assigned_to field is null
+     *
+     * @param {Object} data the api response data
+     */
+    successHandler = (data: Object): void => {
```
DanDeMicco: Can this be typed?

src/api/__tests__/Tasks-test.js
```diff
@@ -212,5 +212,36 @@ describe('api/Tasks', () => {
...
+
+        describe('successHandler()', () => {
+            test('should just return if tasks instance is destroyed', () => {
+                tasks.successCallback = successCallback;
+                tasks.destroyed = true;
+                tasks.successHandler();
+                expect(successCallback).not.toHaveBeenCalled();
+            });
+
+            test('should throw error if no data.entries', () => {
```
DanDeMicco: I think this test is labelled incorrectly

src/api/Tasks.js
```diff
@@ -195,6 +196,27 @@ class Tasks extends Base {
...
+        if (this.isDestroyed() || typeof this.successCallback !== 'function') {
+            return;
+        }
+
+        if (!data.entries) {
+            this.successCallback(data);
```
DanDeMicco: If there isnt an entries array, should be be an error? Or should you return some default structure, since data is incorrect

boxcla:

Compared against previously merged PR #547

src/api/Tasks.js
ConradJChan: I'll change it to return an empty default object `{ entries: [], total_count: 0 }`

src/util/fields.js
```diff
@@ -187,6 +195,26 @@ function fillMissingProperties(
     return newObj;
 }
 
+/**
+ * Fill user properties that are null in an object
+ * @param {Object} obj - some object
+ * @return {Object} new object with user placeholder
+ */
+function fillUserPlaceholder(obj: Object): Object {
```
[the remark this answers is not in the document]
ConradJChan: sounds reasonable to me

src/api/File.js
```diff
@@ -189,12 +189,18 @@ class File extends Item {
...
+            // If fields are user fields, make sure if value is null to prepopulate
+            // with placeholder user
+            const filledInData = this.fillUserPlaceholder(
+                dataWithMissingFields,
+            );
+
             // Cache check is again done since this code is executed async
             if (cache.has(key)) {
-                cache.merge(key, dataWithMissingFields);
+                cache.merge(key, filledInData);
```
priyajeet: I am also not sure if the cache should be having mocked out objects. I think the component should have defaults set on its end and not at the data layer. Since this assumes that mocked out placeholder is going to be used by all components.

src/api/Base.js
```diff
@@ -16,7 +16,10 @@ import {
     HTTP_POST,
     HTTP_PUT,
     HTTP_DELETE,
+    PLACEHOLDER_USER,
 } from '../constants';
+import set from 'lodash/set';
```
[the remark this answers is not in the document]
ConradJChan: can do

src/api/Base.js
```diff
@@ -393,6 +396,27 @@ class Base {
...
         }
     }
+
+    /**
+     * Fill user properties that are null in an object
+     *
+     * @param {Object} obj - some object
+     * @return {Object} new object with user placeholder
+     */
+    fillUserPlaceholder(obj: Object): Object {
```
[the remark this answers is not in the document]
ConradJChan: Just moved it from there 😅. But looking at the other methods in Base.js I think you're right, this seems more like a utility than some API core functionality. Will move it back

src/api/File.js
ConradJChan: So if I'm reading you right, I'll pass the placeholder'ed data straight through to the `successHandler` but leave the cache as the API response was received so if people really want the API truth, they can hit the cache?

priyajeet: More like you have the placeholder in the component that needs it.

src/api/Base.js
```diff
@@ -393,6 +396,27 @@ class Base {
...
+        const newObj = { ...obj };
+
+        USER_FIELDS.forEach((field: string) => {
+            if (
+                Object.prototype.hasOwnProperty.call(newObj, field) &&
```
pramodsum: should this be `newObj.hasOwnProperty(field)`?

ConradJChan: I had it as that originally... then eslint told me I was wrong, but based on the eslint rules now it appears that calling `hasOwnProperty` on newObj is acceptable again!

src/util/fields.js
```diff
@@ -182,11 +190,29 @@ function fillMissingProperties(
...
+ * Fill user properties that are null in an object
+ *
+ * @param {Object} obj - some object
+ * @return {Object} new object with user placeholder
+ */
+function fillUserPlaceholder(obj: Object): Object {
+    const newObj = { ...obj };
+
+    USER_FIELDS.forEach((field: string) => {
+        if (newObj.hasOwnProperty(field) && newObj[field] === null) {
```
priyajeet: can lodash/has be used here?

src/api/Feed.js
```diff
@@ -24,6 +25,7 @@ import {
     VERSION_UPLOAD_ACTION,
     VERSION_RESTORE_ACTION,
     TYPED_ID_FEED_PREFIX,
+    PLACEHOLDER_USER,
```
priyajeet: One thing I am curious, what are the use cases when we have to fill in placeholder users? I notice that we are inserting user id with 2, which is reserved for anonymous users. But we don't show feed for anonymous users.

Instead of putting in placeholder users, is it better to just do null checks in components with the assumption that the user object may be absent? Or is some special UI being generated for placeholder user?

ConradJChan: @JustinHoldstock can correct me if I'm wrong but I heard that it's for the use case where a user has been deleted from the system, in which case the API will return null for the user. 

I think at the end of the day, the components should probably have null checks knowing that users could come in as null. This should fix the issue for ContentSidebar but say another ui-element ends up using comment (annotations?) they'd have to make a similar fix or make the fix in Comment itself. We can look at handling this in the annotations react refactoring

priyajeet: If thats the case I would say we should just go ahead and do the simpler null checks / defensive approach for the components. Each component should know what to do with a potential null value. I also know that there is at least one use case where we use that null check to not show something. So if we add a placeholder, that will result in an incorrect experience, specially if the placeholder is actually a possible value.

ConradJChan: Ok let me close this PR then and do a new one with handling just in the components affected by this issue in sidebar

PR #560, opened 2018-09-14
New: Render Open With dropdown

jeremypress:


boxcla:

boxcla:

Compared against previously merged PR #559

src/components/OpenWith/OpenWithDropdownMenuItem.js
```diff
@@ -0,0 +1,34 @@
...
+
+const OpenWithDropdownMenuItem = ({
+    integration: { displayName, displayDescription },
+    onClick,
+}: Props) => {
+    const Icon = ICON_FILE_MAP[displayName] || IconFileDefault;
+    return (
+        <MenuItem onClick={onClick}>
+            <Icon />
+            <span>
```
jstoffan: Try to avoid markup that doesn't have semantic value.

src/components/OpenWith/OpenWithButtonContents.js
```diff
@@ -0,0 +1,24 @@
...
+
+type Props = {
+    children: any,
+};
+
+const OpenWithButtonContents = ({ children }: Props) => (
+    <React.Fragment>
+        {children}
+        <span className="btn-header-text">
+            <FormattedMessage {...messages.open} />{' '}
```
jstoffan: `{' '}`?

src/components/OpenWith/OpenWithDropdownMenuItem.js
jeremypress: Based on this layout, I think we need a container so that the p tags don't appear in line. The container that holds the text and the icon is using flexbox with default row direction.

PR #564, opened 2018-09-18
New: Execute integrations

jeremypress:


boxcla:

flow-typed/box-ui-elements.js
```diff
@@ -563,6 +563,14 @@ type AppIntegrationAPIMiniItem = {
     type: APP_INTEGRATION,
 }
 
+type ExecuteAPI = {
+    url: string,
+    params: ?Array<{key: string, value: string}>,
+    integration_type: string,
+    method: "POST" | "GET",
+
```
jstoffan: Extra line break here.

src/components/OpenWith/OpenWith.js
```diff
@@ -192,9 +200,64 @@ class OpenWith extends PureComponent<Props, State> {
...
+
+    /**
+     * Opens the integration in a new tab based on the API data
+     *
+     * @private
+     * @param {ExecuteAPI} executeData - API response on how to open an executed integration
+     * @return {void}
+     */
+    executeIntegrationSuccessHandler = (executeData: ExecuteAPI) => {
+        switch (executeData.method.toLowerCase()) {
```
jstoffan: The `toLowerCase()` here tripped me up for a second when I got to the case statements below. Should we keep everything uppercase?

src/components/OpenWith/OpenWith.js
```diff
@@ -192,9 +200,64 @@ class OpenWith extends PureComponent<Props, State> {
...
+            case HTTP_POST:
+                this.setState({ executePostData: executeData }, () => {
+                    const form: ?HTMLFormElement = document.querySelector(
```
jstoffan: Can we access the form element via a ref, instead?

src/components/OpenWith/OpenWith.js
```diff
@@ -192,9 +200,64 @@ class OpenWith extends PureComponent<Props, State> {
...
+                        `#bcow-execute-form-${this.id}`,
+                    );
+
+                    if (!form) {
+                        this.executeIntegrationErrorHandler(
+                            Error('execution form unavailable'),
```
JustinHoldstock: Message as a constant

src/components/OpenWith/OpenWith.js
```diff
@@ -192,9 +200,64 @@ class OpenWith extends PureComponent<Props, State> {
...
+                    form.submit();
+                });
+
+                break;
+            case HTTP_GET:
+                window.open(executeData.url, '_blank');
+                break;
+            default:
+                this.executeIntegrationErrorHandler(
+                    Error('unknown invocation type'),
```
JustinHoldstock: Message as a constant

src/components/OpenWith/OpenWith.js
```diff
@@ -192,9 +200,64 @@ class OpenWith extends PureComponent<Props, State> {
...
+                        );
+                        return;
+                    }
+
+                    form.submit();
+                });
+
+                break;
+            case HTTP_GET:
+                window.open(executeData.url, '_blank');
```
JustinHoldstock: '_blank' as a constant

src/components/OpenWith/OpenWithButton.js
```diff
@@ -69,9 +69,8 @@ const OpenWithButton = ({
             position="bottom-center"
         >
             <Button
-                data-attribute-id={id}
                 isDisabled={isDisabled}
-                onClick={onClick}
+                onClick={onClick ? () => onClick(id) : null}
```
JustinHoldstock: `noop` instead of `null`?

boxcla:

Compared against previously merged PR #561

src/components/OpenWith/OpenWith.js
```diff
@@ -192,9 +200,64 @@ class OpenWith extends PureComponent<Props, State> {
...
+     * Opens the integration in a new tab based on the API data
+     *
+     * @private
+     * @param {ExecuteAPI} executeData - API response on how to open an executed integration
+     * @return {void}
+     */
+    executeIntegrationSuccessHandler = (executeData: ExecuteAPI) => {
+        switch (executeData.method.toLowerCase()) {
+            case HTTP_POST:
+                this.setState({ executePostData: executeData }, () => {
```
JustinHoldstock: Can you pull the callback out into it's own function?

src/components/OpenWith/OpenWith.js
jstoffan: Consider handling the case where the window wasn't openable, which should return `null` here.

src/components/OpenWith/OpenWith.js
```diff
@@ -244,6 +308,9 @@ class OpenWith extends PureComponent<Props, State> {
                             integrations={integrations}
                         />
                     )}
+                    {executePostData && (
+                        <ExecuteForm formData={executePostData} id={this.id} />
```
jstoffan: It looks like `ExecuteAPI`, `executePostData`, and `formData` all represent the same data. Should the names be consolidated?

test/openwith.html
```diff
@@ -8,26 +8,23 @@
...
             .container {
                 display: flex;
                 height: 100%;
                 width: 100%;
                 align-items: center;
                 justify-content: center;
             }
-
             .container > div {
+                height: 100%;
```
jstoffan: Mixing `height: 100%` with flex containers/children (especially in column mode) can cause subtle bugs in older browsers. Can this be accomplished using only flexbox?

src/components/OpenWith/ExecuteForm.js
```diff
@@ -0,0 +1,32 @@
...
+
+type Props = {
+    formData: ExecuteAPI,
+    id: string,
+};
+
+const ExecuteForm = ({ formData: { url, params }, id }: Props) => (
+    <React.Fragment>
+        <form
+            id={`bcow-execute-form-${id}`}
```
priyajeet: Is ID needed? Can we use React.createRef() and pass back reference

src/components/OpenWith/OpenWithButton.js
jeremypress: Is there a benefit? The function won't get called by react if it's `null`

src/components/OpenWith/OpenWith.js
priyajeet: probably have getWindow() or getGlobal() in dom utils.js
So that you can stub testing out

priyajeet: Double check if popup blocker doesn't block code.
Normally browsers block opening of windows when that open happens async to a user action. For example in a callback.

src/components/OpenWith/OpenWithButton.js
```diff
@@ -14,7 +14,7 @@ import ICON_FILE_MAP from './IconFileMap';
 import messages from '../messages';
 
 type Props = {
-    displayIntegration: ?Integration | ?Object,
+    displayIntegration?: ?Integration | ?Object,
```
priyajeet: Lots of ? 
are all needed?

src/components/OpenWith/ExecuteForm.js
```diff
@@ -0,0 +1,32 @@
...
+            action={url}
+            target="_blank"
+            acceptCharset="utf-8"
+            method="post"
+            encType="multipart/form-data"
```
priyajeet: Form defaults are application/x-www-form-urlencoded
multipart/form-data is mostly for binary
https://stackoverflow.com/questions/4007969/application-x-www-form-urlencoded-or-multipart-form-data

Also,

src/components/OpenWith/OpenWith.js
jeremypress: yes, we should not lower case it. I was trying to avoid making an extra constant.

flow-typed/box-ui-elements.js
```diff
@@ -563,6 +563,14 @@ type AppIntegrationAPIMiniItem = {
     type: APP_INTEGRATION,
 }
 
+type ExecuteAPI = {
+    url: string,
+    params: ?Array<{key: string, value: string}>,
+    integration_type: string,
+    method: "POST" | "GET",
```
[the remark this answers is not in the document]
jeremypress: those are lower cased :(

src/components/OpenWith/OpenWithDropdownMenu.js
```diff
@@ -24,7 +24,7 @@ const OpenWithDropdownMenu = ({ integrations, onClick }: Props) => (
                     <OpenWithDropdownMenuItem
                         integration={integration}
                         onClick={onClick}
-                        key={integration.appIntegrationId}
+                        key={integration.displayName}
```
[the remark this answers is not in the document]
jeremypress: I can go back to ID, I think when I made this change I thought multiple entries could use the same integration.

src/components/OpenWith/ExecuteForm.js
jeremypress: yeah that sounds better

src/components/OpenWith/OpenWithButton.js
jeremypress: I think I can probably get rid of that object, I'll play around with it.

flow-typed/box-ui-elements.js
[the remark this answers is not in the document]
priyajeet: make them upper :)

src/components/OpenWith/OpenWithButton.js
priyajeet: Should you even render the button if onClick is absent then?
Also, css below isn't prefixed 'integration-icon'

src/components/OpenWith/OpenWithDropdownMenuItem.js
```diff
@@ -15,12 +15,12 @@ type Props = {
 };
 
 const OpenWithDropdownMenuItem = ({
-    integration: { displayName, displayDescription },
+    integration: { displayName, displayDescription, appIntegrationId },
     onClick,
 }: Props) => {
     const Icon = ICON_FILE_MAP[displayName] || IconFileDefault;
```
priyajeet: I think this is being repeated.
Probably have a function inside IconFileMap.js and export that instead.
getIcon()

priyajeet: https://github.com/box/box-ui-elements/pull/564/files#diff-bb9a336e6b9e61655031483d8276ded2R64

src/components/OpenWith/OpenWithDropdownMenuItem.js
```diff
@@ -15,12 +15,12 @@ type Props = {
...
     return (
-        <MenuItem onClick={onClick}>
+        <MenuItem onClick={() => onClick(appIntegrationId)}>
```
priyajeet: class string prefixing?

src/components/OpenWith/OpenWith.js
priyajeet: will this case ever get executed?

src/components/OpenWith/OpenWith.js
priyajeet: Will !form ever be true?

src/components/OpenWith/OpenWith.js
jstoffan: Fair enough. You'll often see the pattern of `const myThing = new MyThing()` when instantiating classes. Most languages handle type declaration within classes. I figured it might make sense here as well, but it's up to you.

src/components/OpenWith/OpenWithDropdownMenu.js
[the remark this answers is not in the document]
jstoffan: Should it be some kind of combination of integration.id and appIntegrationId, then?

src/components/OpenWith/OpenWith.js
jeremypress: @priyajeet if the API adds a new integration type without a client change this could happen.
@ConradJChan good point, my plan is to add an OnError() callback that the developers can pass in. They'll see the message in that case. What else do you think would be useful besides a string message?

src/components/OpenWith/OpenWithButton.js
jeremypress: yes, empty and error cases don't have a on click

src/components/OpenWith/OpenWith.js
jeremypress: I'll include the error, so they can see the stack trace if they are developing.

src/components/OpenWith/OpenWithDropdownMenu.js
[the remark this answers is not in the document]
jeremypress: There's only one ID, i'll make sure it's unique

src/components/OpenWith/OpenWithDropdownMenuItem.js
jeremypress: Yeah Jared pointed this out as well. It's a good idea but I need to be able to distinguish if It's the default icon or not for style purpose. I could do an instance of check in the component...

flow-typed/box-ui-elements.js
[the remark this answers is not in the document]
jeremypress: https://github.com/box/box-ui-elements/pull/565

src/components/OpenWith/OpenWith.js
jeremypress: that depends on if setstate can fail to invoke the callback, not sure if that's possible

src/components/OpenWith/OpenWith.js
ConradJChan: Not sure :) Stack trace would be helpful for developing, but even a more descriptive string like, "<method> is an unsupported open with execute integration type"

jeremypress: sounds good!

src/components/OpenWith/ExecuteForm.js
jeremypress: React states that passing up the reference isn't ideal. Daniel had the nice suggestion to do the submit in `componentDidMount` of the form component itself.

src/components/OpenWith/OpenWith.js
jeremypress: it will block code, isn't that to be expected? After the popup's get enabled you have to click again.

package.json
```diff
@@ -1,3 +1,4 @@
+
```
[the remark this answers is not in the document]
jeremypress: I didn't touch it, but maybe some automated process did that.

flow-typed/box-ui-elements.js
```diff
@@ -563,6 +563,13 @@ type AppIntegrationAPIMiniItem = {
     type: APP_INTEGRATION,
 }
 
+type ExecuteAPI = {
+    url: string,
+    params: ?Array<{key: string, value: string}>,
```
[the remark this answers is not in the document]
jeremypress: added

src/components/OpenWith/OpenWith.js
```diff
@@ -192,9 +204,69 @@ class OpenWith extends PureComponent<Props, State> {
...
+     */
+    executeIntegrationSuccessHandler = (executePostData: ExecuteAPI): void => {
+        const { method, url } = executePostData;
+        switch (method) {
+            case HTTP_POST:
+                this.setState({ executePostData });
+                break;
+            case HTTP_GET:
+                // Use a unique combination of integration + file ID to avoid creation of new windows
+                const windowRef = window.open(url);
```
DanDeMicco: Is the above comment correct? I am assuming you are meaning the 2nd param to window.open

src/components/OpenWith/OpenWith.js
```diff
@@ -192,9 +204,69 @@ class OpenWith extends PureComponent<Props, State> {
...
+
+                if (!windowRef) {
+                    this.executeIntegrationErrorHandler(
```
DanDeMicco: Have you considered opening the integration in a new tab instead of a new window?

src/components/OpenWith/OpenWithButton.js
```diff
@@ -69,13 +70,14 @@ const OpenWithButton = ({
             position="bottom-center"
         >
             <Button
-                data-attribute-id={id}
                 isDisabled={isDisabled}
-                onClick={onClick}
+                onClick={onClick ? () => onClick(id) : noop}
```
DanDeMicco: Typically you want to avoid inline arrow functions, as a new function is created on each render, thus the sub components will be rendered. It probably won't make much of a difference though

src/components/OpenWith/OpenWith.js
jeremypress: whoops, I need to remove that since we aren't doing any optimizations with window opening yet.

src/components/OpenWith/OpenWith.js
jeremypress: I don't think I can control that from JS? Pal mentioned that there is a setting on their side that dictates. Currently some integrations open via a new tab using this method.

DanDeMicco: Nevermind about this. How its executed is dependent on the users browser settings 👍

PR #569, opened 2018-09-19
New: Open With works inside of Content Preview

jeremypress:


boxcla:

Compared against previously merged PR #568

jeremypress:
sep-19-2018 14-10-51

src/components/ContentPreview/ContentPreview.js
```diff
@@ -1055,10 +1059,13 @@ class ContentPreview extends PureComponent<Props, State> {
                     {hasHeader && (
                         <Header
                             file={file}
+                            token={token}
                             onClose={onClose}
                             onPrint={this.print}
                             canDownload={this.canDownload()}
                             onDownload={this.download}
+                            showOpenWithButton={showOpenWithButton}
```
jstoffan: Can this be folded into `openWithProps`, like `openWithProps.show`?

src/components/ContentPreview/Header.js
```diff
@@ -37,6 +43,8 @@ const Header = ({
     const closeMsg = intl.formatMessage(messages.close);
     const printMsg = intl.formatMessage(messages.print);
     const downloadMsg = intl.formatMessage(messages.download);
+    const fileId = file && file.id;
```
jstoffan: Should this guard live in the OpenWith component itself?

src/components/OpenWith/OpenWith.js
```diff
@@ -292,20 +293,25 @@ class OpenWith extends PureComponent<Props, State> {
...
+            className,
+        }: Props = this.props;
         const {
             fetchError: error,
             isLoading,
             integrations,
             executePostData,
         }: State = this.state;
 
+        const styleClassName = classNames('be bcow', className);
```
jstoffan: May want to consider doing `const className = classNames('be bcow', this.props.className);`.

src/components/OpenWith/OpenWith.js
```diff
@@ -332,4 +338,5 @@ class OpenWith extends PureComponent<Props, State> {
     }
 }
 
+export type OpenWithProps = Props;
```
jstoffan: Can this be exported in line with its declaration above and be renamed on import?

src/components/ContentPreview/ContentPreview.js
jeremypress: my only issue with that is that `show` would never actually be applied to the element. So we would want to remove it when spreading the props on the component?

src/components/OpenWith/OpenWith.js
jeremypress: sounds reasonable

src/components/OpenWith/OpenWith.js
jeremypress: I think that would work, I'm following the pattern of the other elements that are used in content explorer like preview and the sidebar.

src/components/ContentPreview/Header.js
jeremypress: Good catch. I'll add a guard there to avoid unnecessary network calls, but this guard protects us from accessing a property on an undefined file prop since it's an optional prop.

PR #573, opened 2018-09-21
New: Callbacks for execution and errors

jeremypress:


boxcla:

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -229,21 +238,24 @@ class ContentOpenWith extends PureComponent<Props, State> {
...
      */
-    executeIntegrationSuccessHandler = (executePostData: ExecuteAPI): void => {
-        const { method, url } = executePostData;
+    executeIntegrationSuccessHandler = (executeData: ExecuteAPI): void => {
+        const { onExecute }: Props = this.props;
+        const { method, url } = executeData;
         switch (method) {
             case HTTP_POST:
-                this.setState({ executePostData });
+                this.setState({ executePostData: executeData });
```
jstoffan: Should the state property also be renamed?

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -229,21 +238,24 @@ class ContentOpenWith extends PureComponent<Props, State> {
...
      * @private
-     * @param {ExecuteAPI} executePostData - API response on how to open an executed integration
+     * @param {ExecuteAPI} executeData - API response on how to open an executed integration
 
      * @return {void}
      */
-    executeIntegrationSuccessHandler = (executePostData: ExecuteAPI): void => {
-        const { method, url } = executePostData;
+    executeIntegrationSuccessHandler = (executeData: ExecuteAPI): void => {
+        const { onExecute }: Props = this.props;
```
jstoffan: Since this is only used once, you could use it inline: `this.props.onExecute(this.executeId);` Same goes for most of the other callback references in this file.

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -254,6 +266,8 @@ class ContentOpenWith extends PureComponent<Props, State> {
                 } else {
                     windowRef.opener = null;
                     windowRef.location = url;
+                    onExecute(this.executeId);
+                    this.executeId = null;
```
jstoffan: Should there be a wrapping handler to avoid the duplication of nulling out this property?

src/components/ContentOpenWith/ContentOpenWith.js
jeremypress: the state property is only in the post case, while as this data from the API could be used for POST or GET. Hence I've left it general here but made it more specific in state

src/components/ContentOpenWith/ContentOpenWith.js
jeremypress: yep, will add that in.

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -229,21 +237,23 @@ class ContentOpenWith extends PureComponent<Props, State> {
...
-    executeIntegrationSuccessHandler = (executePostData: ExecuteAPI): void => {
-        const { method, url } = executePostData;
+    executeIntegrationSuccessHandler = (executeData: ExecuteAPI): void => {
+        const { method, url } = executeData;
         switch (method) {
             case HTTP_POST:
-                this.setState({ executePostData });
+                this.setState({ executePostData: executeData });
                 break;
             case HTTP_GET:
```
jstoffan: Consider using `setState` for the GET case as well, which would allow the render method to handle the different integration types. You could then move the `onExecute` invocation to the `setState` "post-render" callback.

PR #576, opened 2018-09-24
Fix: Fix sandboxed iframe case for GET integrations

jeremypress:
Fixes the codepen case where we are operating in a sandboxed iframe. Setting the URL first allows the redirect to happen, although chrome still throws errors.

PR #577, opened 2018-09-25
Fix: ContentPreview navigation issues

DanDeMicco:
TODO: tests

### DIFF
--- a/src/components/ContentPreview/ContentPreview.js
+++ b/src/components/ContentPreview/ContentPreview.js
@@ -81,6 +81,8 @@ type State = {
     file?: BoxItem,
     isFileError: boolean,
     isReloadNotificationVisible: boolean,
+    currentFileId?: string, // the currently displayed file id in the collection
+    prevFileId?: string, // the prop "fileId" which is needed to implement getDerivedStateFromProps
 };
 
 // Emitted by preview's 'load' event
@@ -181,6 +183,7 @@ class ContentPreview extends PureComponent<Props, State> {
             apiHost,
             requestInterceptor,
             responseInterceptor,
+            fileId,
         } = props;
 
         this.id = uniqueid('bcpr_');
@@ -194,7 +197,11 @@ class ContentPreview extends PureComponent<Props, State> {
             requestInterceptor,
             responseInterceptor,
         });
-        this.state = { ...this.initialState };
+        this.state = {
+            ...this.initialState,
+            currentFileId: fileId,
+            prevFileId: fileId,
+        };
     }
 
     /**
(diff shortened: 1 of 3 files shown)


boxcla:

src/components/ContentPreview/ContentPreview.js
```diff
@@ -237,28 +244,38 @@ class ContentPreview extends PureComponent<Props, State> {
...
         this.loadScript();
 
-        this.fetchFile(fileId);
+        this.fetchFile(this.state.currentFileId);
         this.focusPreview();
     }
 
+    static getDerivedStateFromProps(props: Props, state: State) {
+        const { fileId } = props;
+        if (state.prevFileId !== fileId && fileId !== state.currentFileId) {
```
priyajeet: What if the incoming fileId is actually the previous file id. As in the parent actually set the prior one. In that case we have to cancel getting the current one and fetch the prior one again.

src/components/ContentPreview/ContentPreview.js
```diff
@@ -828,14 +847,11 @@ class ContentPreview extends PureComponent<Props, State> {
         const fileId =
             typeof fileOrId === 'object' ? fileOrId.id || '' : fileOrId;
 
+        this.setState({
+            currentFileId: fileId,
```
priyajeet: Should navigate below be as a callback? Or does it not matter

src/components/ContentPreview/ContentPreview.js
```diff
@@ -598,12 +613,16 @@ class ContentPreview extends PureComponent<Props, State> {
...
         }
-
-        const typedId: string = getTypedFileId(this.getFileId(file));
+        const fileId = this.getFileId(file);
+        const typedId: string = getTypedFileId(fileId);
         const token: TokenLiteral = await TokenService.getReadToken(
             typedId,
             tokenOrTokenFunction,
         );
+        if (fileId !== this.state.currentFileId) {
```
priyajeet: State is destructured above.
Should this check be moved above, so that token call/XHR may be prevented?

src/components/ContentPreview/ContentPreview.js
priyajeet: Can you align the equalities? fileId === FOO, that is on the same side :)

src/components/ContentPreview/ContentPreview.js
DanDeMicco: Yes, it is destructured above, but ```currentFileId``` cannot move there. This code is async, so the value of the state could change during this time and the value of ```currentFileId``` would be stale

DanDeMicco: Moving it above is also not something which has to be done, as we definitely want it in after the await. Adding an additonal check before is not needed, as the function wouldnt be called in the first place if they were different (see ```componentDidUpdate```)

src/components/ContentPreview/ContentPreview.js
DanDeMicco: It doesn't matter

boxcla:

Compared against previously merged PR #576

PR #578, opened 2018-09-25
Chore: Create new windows immediately to avoid popup blockers

jeremypress:


boxcla:

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -368,6 +372,7 @@ class ContentOpenWith extends PureComponent<Props, State> {
                         <ExecuteForm
                             onSubmit={this.onExecuteFormSubmit}
                             executePostData={executePostData}
+                            windowRef={this.windowRef}
```
jstoffan: Consider passing down only what the component needs, which is the window name. It doesn't seem to use the window object for anything else.

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -229,6 +230,12 @@ class ContentOpenWith extends PureComponent<Props, State> {
      */
     onIntegrationClick = (appIntegrationId: string): void => {
         const { fileId }: Props = this.props;
+        // window.open() is immediately invoked to avoid popup-blockers
+        // The name is included to be the target of a form if the integration is a POST integration
+        this.windowRef = window.open(
```
priyajeet: Can this be migrated to something like windowUtils.js?

jstoffan: Consider persisting the window name as an instance property and nulling out the window object in the POST case. It'll make it clear that the window object itself is still only used in the GET case, whereas the name is used for multiple things.

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -256,22 +263,19 @@ class ContentOpenWith extends PureComponent<Props, State> {
...
-                    windowRef.location = url;
-                    windowRef.opener = null;
-                    this.onExecute();
                 }
+
                 // Prevents abuse of window.opener
                 // see here for more details: https://mathiasbynens.github.io/rel-noopener/
-                windowRef.opener = null;
-                windowRef.location = url;
+                this.windowRef.location = url;
```
priyajeet: Could also be part of your windowUtils.js

setLocation(window, string?)
setOpener(window, string?)

jstoffan: It seems like mocking the global.window property would be more efficient. Otherwise you're basically writing a wrapper around every window property.

priyajeet: Mocking global properties can cause tests to fail unless the original version is properly restored

boxcla:

Compared against previously merged PR #576

src/components/ContentOpenWith/ContentOpenWith.js
jstoffan: Take care to reset it, if it's even necessary. It seems like it'd be worth a little experimentation. Also, it looks like `global` _is_ `window` in Jest. Maybe it already mocks these properties?

https://github.com/facebook/jest/issues/3692#issuecomment-408132529

jstoffan: As an alternative, one strategy I’ve used in the past is to store a reference as `this.window = window` via `componentDidMount`.

jeremypress: @priyajeet do you see value in a separate constants file outside of testability? `this.window = window` would solve that imo.

src/components/ContentOpenWith/ContentOpenWith.js
jeremypress: I'm not sure I see the value in that. This would have to be done in the switch statement of the successhandler where we know the integration type. We're still using the window object in the POST case, but we are targeting it via its name. 

It also may be useful later if we try to minimize our creation of new windows.

jstoffan: Fair enough.

src/components/ContentOpenWith/ContentOpenWith.js
priyajeet: The window/document utility can be limited to just this, and not needed to add wrapper for every property set

```js
// @flow

function getWindow() {
    return window;
}

function getDocument() {
    return document;
}

export { getWindow, getDocument };
```

priyajeet: saved this.window instance also sounds good too.
Though I would just do this.window = getWindow() in that case if the componentWillMount() function is tested, in which case you will be like expect(getWindow).to.be.called()

jeremypress: is the window object not available in jest? 

```componentDidMount();
expect(this.window).toEqual(window);```

priyajeet: I think its `global` instead

PR #582, opened 2018-10-01
Update: exposing annotations in ContentPreview header

ConradJChan:


boxcla:

boxcla:

Compared against previously merged PR #583

src/components/ContentPreview/ContentPreview.js
```diff
@@ -43,38 +43,39 @@ import '../base.scss';
...
-    measureRef: Function,
-    onLoad: Function,
-    onNavigate: Function,
-    onDownload: Function,
-    onClose?: Function,
-    language: string,
-    messages?: StringMap,
+    autoFocus: boolean,
     cache?: APICache,
+    canAnnotate?: boolean,
```
jeremypress: Does this prop reflect permissions or is it a configuration flag?

src/components/ContentPreview/ContentPreview.js
```diff
@@ -625,11 +643,13 @@ class ContentPreview extends PureComponent<Props, State> {
...
         const previewOptions = {
+            container: `#${this.id} .bcpr-content`,
+            header: 'none',
+            showAnnotations: this.canAnnotate(),
             showDownload: this.canDownload(),
             skipServerUpdate: true,
-            header: 'none',
-            container: `#${this.id} .bcpr-content`,
             useHotkeys: false,
+            headerElement: `#${this.id} .bcpr-header`,
```
jeremypress: finish alphabetizing? :)

src/components/ContentPreview/Header.js
```diff
@@ -17,85 +19,129 @@ import ContentOpenWith from '../ContentOpenWith/ContentOpenWith';
...
+                            aria-label={drawMsg}
+                        >
+                            <IconDrawAnnotationMode
+                                color={COLOR_999}
+                                width={18}
+                                height={18}
+                            />
+                        </PlainButton>
+                    )}
+                    {canAnnotate && (
```
jeremypress: You could batch all of these under one logical canAnnotate &&. Also, how hare configurations for enabling/disabling certain modes reflected here? This looks like an all or nothing check

src/components/ContentPreview/Header.scss
```diff
@@ -5,8 +5,9 @@
 
     .be & .bcpr-bcow-btn,
     .be & .btn-plain.bcpr-btn {
-        line-height: 0;
-        padding: 0 15px 0 5px;
+        display: flex;
```
jeremypress: Double check how this looks with open with!

src/components/ContentPreview/ContentPreview.js
ConradJChan: just config, pass the option through to box content preview and eventually box annotations. I believe box annotations is where the permissions are checked

src/components/ContentPreview/Header.js
ConradJChan: Good point about the different modes for different doc types. The way this is specified, is to add a `bp-is-hidden` class to the buttons and allow Box Annotations itself to show the ones that are valid. I uncovered an issue switching between file types that the different annotate buttons would not get hidden if they didn't apply to the file type. Will update with logic for this

src/components/ContentPreview/Header.scss
ConradJChan: screen shot 2018-10-04 at 3 02 12 pm

src/components/ContentPreview/ContentPreview.js
```diff
@@ -625,10 +643,12 @@ class ContentPreview extends PureComponent<Props, State> {
         }
 
         const previewOptions = {
+            container: `#${this.id} .bcpr-content`,
+            header: 'none',
+            headerElement: `#${this.id} .bcpr-header`,
```
priyajeet: Is it possible to refactor preview to not take / muck with DOM?
Can we just fix this by passing in functions, for example onAnnotationEnd() or something that preview annotations can trigger to toggle the button.

pramodsum: I think we can add that onto a later refactor of BoxAnnotations

src/components/ContentPreview/Header.js
```diff
@@ -17,85 +19,127 @@ import ContentOpenWith from '../ContentOpenWith/ContentOpenWith';
...
-                        className="bcpr-btn"
-                        onClick={onClose}
-                        title={closeMsg}
-                        aria-label={closeMsg}
-                    >
-                        <IconClose color={COLOR_999} width={24} height={24} />
-                    </PlainButton>
-                )}
+                    )}
+                    {canAnnotate && (
```
pramodsum: Would this toggle appropriately if the user has disabled a particular annotations type? https://github.com/box/box-annotations/blob/master/docs/enabling-types.md

ConradJChan: yup, it looks like it does. Should work the same as in preview since we're relying on Box Annotations to do the hiding and showing for us.

PR #583, opened 2018-10-02
Fix: Sidebar null user in modified_by

ConradJChan:


boxcla:

src/components/ContentSidebar/ActivityFeed/activity-feed/activityFeedUtils.js
```diff
@@ -27,14 +29,17 @@ export function collapseFeedState(feedState: ?FeedItems): FeedItems {
             previousFeedItem.action === ItemTypes.upload
         ) {
             const {
-                modified_by: prevModifiedBy,
```
jeremypress: Not sure if this is worth renaming. This is still the previous modified by user, it just may be null

src/components/ContentSidebar/ActivityFeed/activity-feed/activityFeedUtils.js
```diff
@@ -45,15 +50,17 @@ export function collapseFeedState(feedState: ?FeedItems): FeedItems {
                 [prevModifiedBy.id]: { ...prevModifiedBy },
             };
 
+            const modifiedBy = tmpCurModifiedBy || PLACEHOLDER_USER;
```
jeremypress: This definitely works, but could we potentially use a user with no ID? It's slightly less accurate to use the placeholder user since they get counted as anonymous.

jeremypress: If we do stick with the anonymous user, how does that look in the UI when they upload/modify a new version?

src/components/ContentSidebar/ActivityFeed/activity-feed/activityFeedUtils.js
ConradJChan: i could not rename, but then that approach I would need to use a variable I can modify, since I need to do the check to fallback onto `PLACEHOLDER_USER` instead of `const`. Any preference?

boxcla:

Compared against previously merged PR #580

src/components/ContentSidebar/ActivityFeed/activity-feed/activityFeedUtils.js
```diff
@@ -45,15 +50,17 @@ export function collapseFeedState(feedState: ?FeedItems): FeedItems {
                 [prevModifiedBy.id]: { ...prevModifiedBy },
```
jstoffan: Should we just exclude collaborators without a valid user object?

src/components/ContentSidebar/ActivityFeed/activity-feed/activityFeedUtils.js
jeremypress: resolved, we agreed this is cool.

src/components/ContentSidebar/ActivityFeed/activity-feed/activityFeedUtils.js
ConradJChan: I tried this out and it seems misleading because if we have 2 collaborators and 1 is null, then the UI will show that "<name of collaborator> uploaded versions x - y". Will leave it for now and we can address this as part of the follow on work to clean up this code/UX

src/components/ContentSidebar/ActivityFeed/activity-feed/activityFeedUtils.js
ConradJChan: shows "__ uploaded a version" which is the same as Amsterdam currently so leaving it as is.

PR #584, opened 2018-10-05
Fix: Sidebar only renders initially after file info is returned

jeremypress:


boxcla:

jeremypress:
This PR addresses the issue where we initially render the loading sidebar, the files call comes back, and since there are no tabs to show the sidebar disappears.

Brad's desired behavior:
1. Initially don't show anything until we know we have at least 1 tab to show
2. When navigating between files, it's ok to show the loading sidebar, and then hide it if there's nothing to show.

boxcla:

Compared against previously merged PR #583

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -98,6 +99,7 @@ class ContentSidebar extends PureComponent<Props, State> {
 
     initialState: State = {
         isVisible: true,
+        isNavigating: false,
```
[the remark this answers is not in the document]
jeremypress: This was requested by Brad. We may want to make this a prop, but we won't initially show any loading state until the files call comes back with this change.

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -63,7 +63,7 @@ type Props = {
 type State = {
     view?: SidebarView,
     file?: BoxItem,
-    isVisible: boolean,
+    isVisible?: boolean,
```
DanDeMicco: You can make this not be optional, and set it in the constructor

jeremypress: Flow says I can't :(

jeremypress:
@DanDeMicco pointed out that isVisible was doing almost exactly what we needed. By flipping it to false by default we won't render anything until the files call comes back. @priyajeet are there any edge cases you can think of?

src/components/ContentSidebar/__tests__/ContentSidebar-test.js
```diff
@@ -423,7 +423,6 @@ describe('components/ContentSidebar/ContentSidebar', () => {
         let wrapper;
         let instance;
         let fetchFileSuccessCallback;
-        let fetchFileErrorCallback;
```
jeremypress: Cleanup from detailsSidebar refactor.

PR #588, opened 2018-10-09
New: Add loading UI when integrations are opened

jeremypress:


boxcla:

src/components/ContentOpenWith/IntegrationPortal.js
```diff
@@ -0,0 +1,64 @@
...
+        Array.from(document.styleSheets).forEach(styleSheet => {
+            const copiedStyleSheet = document.createElement('link');
+            copiedStyleSheet.rel = 'stylesheet';
+            copiedStyleSheet.href = styleSheet.href;
+            integrationWindowRef.document.head.appendChild(copiedStyleSheet);
+        });
+    }
+
+    render() {
+        return createPortal(this.props.children, this.containerEl);
```
jeremypress: After my BRUI change goes in we may be able to use their Portal component.

src/components/ContentOpenWith/IntegrationPortal.js
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * Copies stylesheets to the new window.
+     *
+     * @private
+     * @return {void}
+     */
+    copyStyles() {
+        const { integrationWindowRef }: Props = this.props;
+        // The new window will have no CSS, so we copy all style sheets as a safe way
+        // of ensuring required styles are present
```
jeremypress: @jstoffan this prevents the renaming issue we talked about yesterday.

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -286,7 +295,7 @@ class ContentOpenWith extends PureComponent<Props, State> {
                 );
         }
 
-        this.windowRef = null;
+        integrationWindowRef = null;
```
ConradJChan: Should this be `this.setState({ integrationWindowRef: null })`?

src/components/ContentOpenWith/IntegrationPortal.js
```diff
@@ -0,0 +1,64 @@
...
+        Array.from(document.styleSheets).forEach(styleSheet => {
+            const copiedStyleSheet = document.createElement('link');
```
ConradJChan: would this be a good case to use document fragments (https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment) instead of `appendChild` in a loop?

src/components/ContentOpenWith/ContentOpenWith.js
jeremypress: I want to set the actual reference to null here for cleanup purposes. I'm not sure if clearing the actual state here could wipe our loading UI early..

src/components/ContentOpenWith/IntegrationPortal.js
jeremypress: I will check that out!

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -373,11 +389,26 @@ class ContentOpenWith extends PureComponent<Props, State> {
                             integrations={integrations}
                         />
                     )}
+                    {integrationWindowRef && (
+                        <IntegrationPortal
+                            integrationWindowRef={integrationWindowRef}
+                        >
+                            <div className="bcow-portal-container">
+                                <LoadingIndicator
```
jeremypress: This will gain some complexity when I add the error state in the next PR

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -260,13 +268,14 @@ class ContentOpenWith extends PureComponent<Props, State> {
      * @return {void}
      */
     executeIntegrationSuccessHandler = (executeData: ExecuteAPI): void => {
+        let { integrationWindowRef }: State = this.state;
```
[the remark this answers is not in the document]
jeremypress: I was previously maintaining a class property and the state, would that be more clear? The class property would allow us to update the window reference with a new URL/clear it when ready, and the state property tells the component when to render the portal.

src/components/ContentOpenWith/ContentOpenWith.scss
```diff
@@ -41,3 +41,18 @@
...
+
+.bcow-portal-container {
+    align-items: center;
+    display: flex;
+    height: 100%;
+    justify-content: center;
+    width: 100%;
+
+    .bcow-portal-loading-indicator {
+        div {
```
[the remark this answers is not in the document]
jeremypress: I don't think so. We need to change the color of the 3 child divs of the indicator, not the indicator div itself.

src/components/ContentOpenWith/ContentOpenWith.js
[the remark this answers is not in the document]
jstoffan: It seems like you're trying to use the existence of the window ref as a shortcut for some other state. It might clear things up to create a separate property for "show the loading indicator or not" state.

src/components/ContentOpenWith/ContentOpenWith.scss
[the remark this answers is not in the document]
jstoffan: Fair enough. Separately, would it make sense to leave it as the default color? If this change is meant to mimic Google's UX, it's probably not going to extend as well to other vendors.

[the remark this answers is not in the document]
jeremypress: It's meant to mimic platform standards where we opt for black over the Box blue.

src/components/ContentOpenWith/IntegrationPortal.js
jeremypress: @ConradJChan, added the fragment. Should improve performance if there are a ton of stylesheets on the users page.

src/components/ContentOpenWith/ContentOpenWith.scss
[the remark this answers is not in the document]
jeremypress: Added a comment to justify the style override

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -235,10 +239,15 @@ class ContentOpenWith extends PureComponent<Props, State> {
         const { fileId }: Props = this.props;
         // window.open() is immediately invoked to avoid popup-blockers
         // The name is included to be the target of a form if the integration is a POST integration
-        this.windowRef = this.window.open(
+        this.integrationWindow = this.window.open(
```
JustinHoldstock: Would it be beneficial to have a WindowUtils file?

PR #593, opened 2018-10-15
New: Error UI for failed integration execution

jeremypress:


boxcla:

boxcla:

Compared against previously merged PR #582

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -389,18 +396,38 @@ class ContentOpenWith extends PureComponent<Props, State> {
                             integrations={integrations}
                         />
                     )}
-                    {shouldRenderIntegrationPortal && (
+                    {shouldRenderLoadingIntegrationPortal && (
```
ConradJChan: What are your thoughts on combining this render block to be `shouldRenderLoadingIntegrationPortal || shouldRenderErrorIntegrationPortal` to save on the duplication of the `IntegrationPortal` and containing div?

jeremypress: Good point. I tried this and ran into some weird issues when trying to copy they styles. I'll try again and figure out what's actually going on.

src/components/ContentOpenWith/ContentOpenWith.scss
```diff
@@ -1,6 +1,8 @@
 @import '../variables';
 
 .be.bcow {
+    font-family: Lato, "Helvetica Neue", Helvetica, Arial, sans-serif;
```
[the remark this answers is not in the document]
jeremypress: This is necessary because the fonts are not a part of the component, but assumed to be on the page. Adding this here will also transfer to the new tab. I'll look into the variable.

src/components/ContentOpenWith/IntegrationPortal.js
```diff
@@ -4,69 +4,51 @@
...
-    }
-
-    /**
-     * Setup UI in the new window.
-     *
-     * @private
-     * @inheritdoc
-     * @return {void}
-     */
-    componentDidMount() {
```
jstoffan: It's usually a convention to do DOM manipulation within `componentDidMount`. What's the reasoning for moving to a function?

src/components/ContentOpenWith/IntegrationPortal.js
```diff
@@ -4,69 +4,51 @@
...
-            const copiedStyleSheet = integrationWindow.document.createElement(
-                'link',
-            );
-            copiedStyleSheet.rel = 'stylesheet';
-            copiedStyleSheet.href = styleSheet.href;
-            integrationWindow.document.head.appendChild(copiedStyleSheet);
-        });
-    }
+    copyStyles(integrationWindow);
+    integrationWindow.document.body.appendChild(containerElement);
```
ConradJChan: does this account for if the integration window isn't closed, we don't append the containerElement again?

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -389,14 +396,50 @@ class ContentOpenWith extends PureComponent<Props, State> {
                             integrations={integrations}
                         />
                     )}
-                    {shouldRenderIntegrationPortal && (
+                    {(shouldRenderLoadingIntegrationPortal ||
```
JustinHoldstock: Looks like there's a lot of duplicated markdown inside of here. Can you extract the error rendering portion out to a method?

Also, it looks like it's possible to render two of the same `IntegrationPortals` if `shouldRenderErrorIntegrationPortal` is true. Is this intended?

src/components/ContentOpenWith/IntegrationPortal.js
jeremypress: If you portal a component to a new tab and close that tab, it does not seem to unmount from React’s perspective.

jstoffan: Hm, okay. I think you'd still want to follow the convention, though. If you need to force the component to remount, can you use a `key` prop to ensure it's unique?

src/components/ContentOpenWith/ContentOpenWith.js
jstoffan: It looks like the old markup was accidentally left in the changeset?

jeremypress: forgot to delete that second one. It should not be there

src/components/ContentOpenWith/IntegrationPortal.js
jeremypress: good catch. Right now it will just re-append itself. Not sure if that actually changes anything but we probably shouldn't be doing it.

src/components/ContentOpenWith/ContentOpenWith.js
jeremypress: @JustinHoldstock does this seem more reasonable?

src/components/ContentOpenWith/IntegrationPortal.js
jeremypress: fair enough, I'll give that a shot.

src/components/ContentOpenWith/IntegrationPortal.js
```diff
@@ -24,7 +24,15 @@ class IntegrationPortal extends PureComponent<Props> {
...
 
         const { integrationWindow }: Props = this.props;
-        this.containerElement = integrationWindow.document.createElement('div');
+        const existingContainer = integrationWindow.document.querySelector(
+            'div',
+        );
+
+        // If the integration window isn't closed, we should reuse the container div and avoid
+        // creating another one.
+        this.containerElement =
```
jeremypress: @jstoffan I'm creating a div here but actually manipulating it in `componentDidMount`. This is because render references the container Element so it needs to be created first.

src/components/ContentOpenWith/ContentOpenWith.js
JustinHoldstock: Almost! Is there anyway you can reduce the duplicate error markdown?

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -389,15 +396,32 @@ class ContentOpenWith extends PureComponent<Props, State> {
...
+                        shouldRenderErrorIntegrationPortal) && (
                         <IntegrationPortal
                             integrationWindow={this.integrationWindow}
+                            key={uniqueid('integration_portal_')}
```
jstoffan: Will this create a new unique ID per-render? Is that desired or should it be tied to the integration somehow?

src/components/ContentOpenWith/IntegrationPortal.js
```diff
@@ -24,7 +24,15 @@ class IntegrationPortal extends PureComponent<Props> {
         super(props);
 
         const { integrationWindow }: Props = this.props;
-        this.containerElement = integrationWindow.document.createElement('div');
+        const existingContainer = integrationWindow.document.querySelector(
```
jstoffan: You may want to apply an `id` to the div when it's created so you can select it more easily/safely.

src/components/ContentOpenWith/ContentOpenWith.js
jstoffan: @JustinHoldstock, can you clarify what duplication you mean? It looks pretty DRY to me.

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -389,15 +396,32 @@ class ContentOpenWith extends PureComponent<Props, State> {
                             integrations={integrations}
                         />
                     )}
-                    {shouldRenderIntegrationPortal && (
+                    {(shouldRenderLoadingIntegrationPortal ||
+                        shouldRenderErrorIntegrationPortal) && (
                         <IntegrationPortal
```
jstoffan: Super minor, but you could consider moving this to a separate component to clean up this file's render method. The error/loading state seems to dominate it at the moment.

src/components/ContentOpenWith/ContentOpenWith.js
jeremypress: You can open the same integration multiple times in the error case. So that needs to cause a re-render

src/components/ContentOpenWith/IntegrationPortal.js
jeremypress: will do, that's safer

src/components/ContentOpenWith/ContentOpenWith.js
JustinHoldstock: @jstoffan @jeremypress Here, and then directly below here are two identical pieces of markdown for an `ErrorMask` element. I'd extract that to a function (`getOpenWithError()`).

```
// First instance as follows
                                {shouldRenderErrorIntegrationPortal ? (
                                    <ErrorMask
                                        errorHeader={
                                            <FormattedMessage
                                                {...messages.executeIntegrationOpenWithErrorHeader}
                                            />
                                        }
                                        errorSubHeader={
                                            <FormattedMessage
                                                {...messages.executeIntegrationOpenWithErrorSubHeader}
                                            />
                                        }
                                    />
                                ) : (
                                    <LoadingIndicator
                                        className="bcow-portal-loading-indicator"
                                        size="large"
                                    />
                                )}

// then below that, but inside an integration portal
                    {shouldRenderErrorIntegrationPortal && (
                        <IntegrationPortal
                            integrationWindow={this.integrationWindow}
                        >
                            <div className="be bcow bcow-portal-container">
                                <ErrorMask
                                    errorHeader={
                                        <FormattedMessage
                                            {...messages.executeIntegrationOpenWithErrorHeader}
                                        />
                                    }
                                    errorSubHeader={
                                        <FormattedMessage
                                            {...messages.executeIntegrationOpenWithErrorSubHeader}
                                        />
                                    }
                                />
                            </div>
                        </IntegrationPortal>
                    )}
```

Both ErrorMasks are the same
```
                                <ErrorMask
                                    errorHeader={
                                        <FormattedMessage
                                            {...messages.executeIntegrationOpenWithErrorHeader}
                                        />
                                    }
                                    errorSubHeader={
                                        <FormattedMessage
                                            {...messages.executeIntegrationOpenWithErrorSubHeader}
                                        />
                                    }
                                />
```

JustinHoldstock: Sorry! Looks like either Github or Chrome was showing out of date results, marking as resolved

src/components/ContentOpenWith/ContentOpenWith.js
jeremypress: As per our discussion, switching to a regular `Component`

i18n/en-US.properties
```diff
@@ -110,6 +110,10 @@ be.errorOpenWithDescription = Opening this file with other services is currently
 be.errorSkill = Error
 # Message shown when there is an error.
 be.errorState = A network error has occurred while trying to load.
+# Header message to the user when an Open With integration fails to execute
+be.executeIntegrationOpenWithErrorHeader = We're sorry, this integration is currently unavailable.
```
priyajeet: ' -> ’
accented

There are a couple of others I see too, maybe search/replace

src/components/ContentOpenWith/IntegrationPortalContainer.js
```diff
@@ -0,0 +1,51 @@
...
+type Props = {
+    hasError: boolean,
+    integrationWindow: any,
+};
+
+const IntegrationPortalContainer = ({ hasError, integrationWindow }: Props) => {
+    return (
+        <IntegrationPortal
+            integrationWindow={integrationWindow}
+            key={uniqueid('integration_portal_')}
```
priyajeet: key for?

src/components/messages.js
```diff
@@ -993,6 +993,19 @@ const messages: { [string]: MessageDescriptor } = defineMessages({
             'Message to the user when there is at least one Open With integration available',
         defaultMessage: 'Open this file with a partner service',
     },
+    executeIntegrationOpenWithErrorHeader: {
+        id: 'be.executeIntegrationOpenWithErrorHeader',
+        description:
+            'Header message to the user when an Open With integration fails to execute',
+        defaultMessage:
+            "We're sorry, this integration is currently unavailable.",
```
priyajeet: accented ’

src/components/ContentOpenWith/IntegrationPortalContainer.js
jeremypress: For some reason the portal doesn't unmount when I close the new tab, so the key forces a remount each time.

PR #595, opened 2018-10-17
Fix: ContentUploads overwrite failure issue

baipenghan:
Also gives a prop to determine overwritten or not.

### DIFF
--- a/src/api/uploads/PlainUpload.js
+++ b/src/api/uploads/PlainUpload.js
@@ -79,7 +79,7 @@ class PlainUpload extends BaseUpload {
         }
 
         const attributes = JSON.stringify({
-            name: this.file.name,
+            name: this.fileName,
             parent: { id: this.folderId },
         });
 
@@ -145,6 +145,7 @@ class PlainUpload extends BaseUpload {
         this.folderId = folderId;
         this.fileId = fileId;
         this.file = file;
+        this.fileName = this.file.name;
         this.successCallback = successCallback;
         this.errorCallback = errorCallback;
         this.progressCallback = progressCallback;
--- a/src/components/ContentUploader/ContentUploader.js
+++ b/src/components/ContentUploader/ContentUploader.js
@@ -68,6 +68,7 @@ type Props = {
     onMinimize?: Function,
     onUpload: Function,
     onUpload: Function,
+    overwrite: boolean,
     requestInterceptor?: Function,
     responseInterceptor?: Function,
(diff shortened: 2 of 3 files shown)


boxcla:
Hi @baipenghan, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

src/components/ContentUploader/ContentUploader.js
```diff
@@ -730,7 +732,7 @@ class ContentUploader extends Component<Props, State> {
             errorCallback: error => this.handleUploadError(item, error),
             progressCallback: event => this.handleUploadProgress(item, event),
             successCallback: entries => this.handleUploadSuccess(item, entries),
-            overwrite: true,
+            overwrite: overwrite,
```
DanDeMicco: Nitpick: overwrite not needed on right

src/components/ContentUploader/ContentUploader.js
```diff
@@ -68,6 +68,7 @@ type Props = {
     onMinimize?: Function,
     onUpload: Function,
     onUpload: Function,
+    overwrite?: boolean,
```
DanDeMicco: I dont think this needs to be optional, since you are defaulting it to true

src/api/uploads/PlainUpload.js
```diff
@@ -79,7 +79,7 @@ class PlainUpload extends BaseUpload {
         }
 
         const attributes = JSON.stringify({
-            name: this.file.name,
+            name: this.fileName || this.file.name,
```
DanDeMicco: I see this is used in a couple places such as in ```makePreflightRequest```. Is it possible to refactor the class a bit to always have this.fileName be the correct name?

src/components/ContentUploader/ContentUploader.js
baipenghan: you're right

src/components/ContentUploader/ContentUploader.js
baipenghan: done

src/api/uploads/PlainUpload.js
baipenghan: Here it depends on whether its subclass set this.fileName or not, and also at the beginning this.fileName is not set. I don't want to affect the existing behavior in case I break something.

baipenghan:
CLA signed

PR #598, opened 2018-10-18
Fix: file id not updating in ContentPreview after props change

DanDeMicco:


boxcla:

boxcla:

Compared against previously merged PR #596

src/components/ContentPreview/__tests__/ContentPreview-test.js
```diff
@@ -791,6 +791,82 @@ describe('components/ContentPreview/ContentPreview', () => {
...
+            expect(wrapper.state('prevFileId')).toBe(initialFileId);
+            expect(wrapper.state('currentFileId')).toBe(newFileId);
+
+            // URL update
+            wrapper.setProps({
+                fileId: newFileId,
+                foo: 'baz',
+            });
+            expect(wrapper.state('currentFileId')).toBe(newFileId);
+            expect(wrapper.state('prevFileId')).toBe(newFileId);
```
jstoffan: I may be out of the loop here, but shouldn't the "previous id" by set to `initialFileId` here? Same question below.

src/components/ContentPreview/ContentPreview.js
```diff
@@ -256,7 +256,7 @@ class ContentPreview extends PureComponent<Props, State> {
 
     static getDerivedStateFromProps(props: Props, state: State) {
         const { fileId } = props;
-        if (fileId !== state.prevFileId && fileId !== state.currentFileId) {
+        if (fileId !== state.prevFileId) {
```
jstoffan: Seems like this component might be simplified by using a `key` based on the fileId prop?

src/components/ContentPreview/__tests__/ContentPreview-test.js
DanDeMicco: No, I think maybe the variable name is confusing:
```
prevFileId?: string, // the prop "fileId" which is needed to implement getDerivedStateFromProps
```

lets walk through a simple example of file id 1 and 2

render file id 1:
```
props: 
fileId: 1

prev state: null

new state: 
prevFileId: 1,
currentFileId:1
```

user navigates to file id 2 via navigation arrows:
```
props: 
fileId: 1

prev state:
prevFileId: 1,
currentFileId: 1

new state: 
prevFileId: 1,
currentFileId: 2
```

application listens to navigate event, updates url, and updates fileId prop to 2:
```
props:
fileId: 2

prev state: 
prevFileId: 1,
currentFileId: 2

new state: 
prevFileId: 2,
currentFileId: 2
```

user clicks back button to file id 1:
```
props:
fileId: 1

prev state: 
prevFileId: 2,
currentFileId: 2

new state:
prevFileId: 1,
currentFileId: 1
```

src/components/ContentPreview/ContentPreview.js
DanDeMicco: It could definitely be simplified by doing this, but using ```key``` would cause the component to unmount and remount

jstoffan: Ah, okay. In that case, could we switch from using `PureComponent` to checking if the fileId changed in `shouldComponentUpdate`? This can happen in a follow-up PR, as well; fixing the bug is the higher priority.

src/components/ContentPreview/__tests__/ContentPreview-test.js
jstoffan: I see, so it's a snapshot of the previous state, rather than the previous file. Fair enough, though a little confusing at first glance.

DanDeMicco: Yeah, good feedback. I renamed it in my latest commit to hopefully make more sense 👍

PR #604, opened 2018-10-22
Chore: Testing integration components

jeremypress:


src/components/ContentOpenWith/IntegrationPortal.js
```diff
@@ -25,6 +27,7 @@ class IntegrationPortal extends PureComponent<Props> {
     constructor(props: Props) {
         super(props);
 
+        this.document = document;
```
jstoffan: Consider moving this to `componentDidMount`, since it doesn't seem related to the portal creation.

src/components/ContentOpenWith/__tests__/ExecuteForm-test.js
```diff
@@ -0,0 +1,47 @@
...
+            submit: submitStub,
+        },
+    };
+    const onSubmitStub = jest.fn();
+
+    const getWrapper = props => shallow(<ExecuteForm {...props} />);
+
+    describe('componentDidMount', () => {
+        it('should submit the form and call callback on mount', () => {
+            React.createRef = jest.fn().mockReturnValue(ref);
```
jstoffan: Consider moving the mocking of this method to a top-level `React` module, similar to `build/react-intl-mocks.js`.

src/components/ContentOpenWith/__tests__/ExecuteForm-test.js
```diff
@@ -0,0 +1,47 @@
...
+            const wrapper = getWrapper({
+                executePostData,
+                onSubmit: onSubmitStub,
+            });
+            const instance = wrapper.instance();
+
+            instance.componentDidMount();
+            expect(submitStub).toBeCalled();
```
jstoffan: You could also test the values being passed back to these methods.

src/components/ContentOpenWith/__tests__/ExecuteForm-test.js
```diff
@@ -0,0 +1,47 @@
...
+
+    const submitStub = jest.fn();
+    const ref = {
+        current: {
+            submit: submitStub,
+        },
+    };
+    const onSubmitStub = jest.fn();
+
+    const getWrapper = props => shallow(<ExecuteForm {...props} />);
```
jstoffan: It might be simpler to use `mount` since you're dealing with refs, rather than unwrapping the instance and setting the property manually.

src/components/ContentOpenWith/__tests__/IntegrationPortal-test.js
```diff
@@ -0,0 +1,122 @@
...
+        const wrapper = getWrapper({
+            integrationWindow: {
+                // use normal document in our integrationWindow
+                document,
+            },
+        });
+
+        wrapper.instance();
+        const containerElement = document.querySelector('div');
+        expect(!!containerElement).toBeTruthy();
```
jstoffan: Is this assertion necessary? It looks like it's testing the logic of the test itself. If it is needed, is the `!!` necessary?

src/components/ContentOpenWith/__tests__/IntegrationPortal-test.js
```diff
@@ -0,0 +1,122 @@
...
+
+    it('should create the container if needed, and assign an id ', () => {
+        const wrapper = getWrapper({
+            integrationWindow: {
+                // use normal document in our integrationWindow
+                document,
+            },
+        });
+
+        wrapper.instance();
```
jstoffan: The Enzyme docs mention this is used to return the underlying instance. Does calling it in isolation have a desirable side effect? Should we use `mount`, instead?

src/components/ContentOpenWith/__tests__/IntegrationPortal-test.js
```diff
@@ -0,0 +1,122 @@
...
+                },
+            });
+            const instance = wrapper.instance();
+            instance.containerElement = document.createElement('div');
+            instance.copyStyles = jest.fn();
+
+            instance.componentDidMount();
+
+            expect(instance.copyStyles).toBeCalled();
+            expect(!!document.querySelector('div')).toBeTruthy();
```
jstoffan: The "truthy" here would lead me to believe that this explicit casting isn't necessary.

src/components/ContentOpenWith/__tests__/IntegrationPortal-test.js
jeremypress: you're correct, that call isn't needed in this case.

src/components/ContentOpenWith/__tests__/IntegrationPortal-test.js
jeremypress: I wanted to make sure the div is there, I guess it's redundant given the ID check on the next line.

src/components/ContentOpenWith/__tests__/ExecuteForm-test.js
jeremypress: These are called without arguments, not sure what you mean?

jstoffan: Ah, okay. My point was just that you usually want to test that the proper values (or lack thereof) were passed to handlers.

PR #606, opened 2018-10-22
Fix: Update MultiputUpload successHandler to parse the response

baipenghan:
If MultiputUpload is used as Upload Embed Widget, the format of the response data of xhr is different. Instead of having "entries" as an attribute in "data", "data" itself is entries[0].


boxcla:

PR #617, opened 2018-10-25
Chore: Upgrade to Babel 7

jstoffan:


boxcla:

boxcla:

Compared against previously merged PR #613

babel.config.js
```diff
@@ -0,0 +1,57 @@
...
+        ],
+    ],
+    env: {
+        dev: {
+            plugins: ['flow-react-proptypes'],
+        },
+        npm: {
+            plugins: [
+                [
+                    'babel-plugin-transform-require-ignore',
```
priyajeet: I think we should be able to remove this one completely from npm.
So that babel output has SCSS imports, which makes it unnecessary to import the CSS file.
(like BRUI)

But that can be another commit.

src/components/makeResponsive.js
```diff
@@ -17,7 +17,7 @@ import {
     CLASS_IS_MEDIUM,
 } from '../constants';
 
-type Props = {
+type PropsShape = {
```
priyajeet: Whats this?

jstoffan: The class properties transform throws a fit unless the function is typed with `Props: { whatever }`. This was the simplest way to keep the function typed without throwing errors on compile. I can change the name if you have an alternative suggestion.

babel.config.js
jstoffan: I'd like to do that separately, since I believe it would be a breaking change. It would probably lead to duplicate stylesheets showing up for users who import the components, as they may already have the compiled stylesheet imported at the top level of their project.

PR #618, opened 2018-10-30
Chore: Update DEVELOPING.md config files and build scripts

jstoffan:

PR #619, opened 2018-10-30
Chore: Remove reuse of integration browser tabs

jeremypress:


boxcla:

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -247,10 +247,12 @@ class ContentOpenWith extends PureComponent<Props, State> {
...
         const { fileId }: Props = this.props;
         // window.open() is immediately invoked to avoid popup-blockers
-        // The name is included to be the target of a form if the integration is a POST integration
+        // The name is included to be the target of a form if the integration is a POST integration.
+        // A uniqueid is used to force the browser to open a new tab every time, while still allowing
+        // a form to reference a given tab.
         this.integrationWindow = this.window.open(
             '',
-            `OpenWithIntegration-${appIntegrationId}`,
+            `${uniqueid(appIntegrationId)}`,
```
jeremypress: I need to use a uniq because
1. We want to open a new window every time, and using the same name will reuse the window, leading to bugs and more complexity
2. We need a way for the form in ExecuteForm to reference an existing tag, afaik the window name is the only way to do this.

boxcla:

Compared against previously merged PR #618

src/components/ContentOpenWith/ContentOpenWith.js
ConradJChan: Was looking at https://developer.mozilla.org/en-US/docs/Web/API/Window/open, says that if window name is empty then the browser will create a new window every time. Would that also work?

jeremypress: unfortunately not for reason 2. The form needs a way to reference that window I opened, and I'm pretty sure window.name is the only way to do that. I could try adding the name after window.open? Otherwise I don't know of a way to post a form to an existing window without using the name.

ConradJChan: Gotcha, didn't realize you had to post to another window

PR #622, opened 2018-10-31
Fix: flipped sorting order of preview sidebar

ChenCodes:
screen shot 2018-10-30 at 8 02 27 pm

### DIFF
--- a/src/util/sorter.js
+++ b/src/util/sorter.js
@@ -59,7 +59,7 @@ export default function(
 }
 
 /**
- * Sort valid feed items, descending by created_at time.
+ * Sort valid feed items, ascending by created_at time.
  *
  * @param {Array<?Comments | ?Tasks | ?FileVersions>} args - Arguments list of each item container
  * type that is allowed in the feed.
@@ -76,7 +76,7 @@ export function sortFeedItems(
 
             return items;
         }, [])
-        .sort((a, b) => Date.parse(b.created_at) - Date.parse(a.created_at));
+        .sort((a, b) => Date.parse(a.created_at) - Date.parse(b.created_at));
 
     return feedItems;
 }


boxcla:
Hi @ChenCodes, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

alexkrolick:
What happens to the initial scroll position when there is more than 1 page of results? Should it scroll to the bottom?

ChenCodes:
> What happens to the initial scroll position when there is more than 1 page of results? Should it scroll to the bottom?

It should be scrolling to the bottom when the page loads, I'll be implementing this in the next commit

src/api/Feed.js
```diff
@@ -106,9 +106,10 @@ class Feed extends Base {
     setCachedItems(id: string, items: FeedItems) {
         const cache = this.getCache();
         const cacheKey = this.getCacheKey(id);
+        const sortedItems = sortFeedItems({ entries: items });
```
jeremypress: Is there a reason we need to sort again after it's done in the Promise.all above?

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -46,6 +46,22 @@ class ActivityFeed extends React.Component<Props, State> {
...
+        }
+    }
+
+    componentDidUpdate(prevProps) {
+        if (
+            this.feedContainer &&
+            prevProps.feedItems === undefined &&
+            this.props.feedItems
+        ) {
+            this.feedContainer.scrollTop = this.feedContainer.scrollHeight;
```
jeremypress: What if a comment is edited? Then we would jump to the bottom when the props change?

DanDeMicco: I think this looks ok. Its checking if the feedItems go from undefined to a truthy value

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -46,6 +46,22 @@ class ActivityFeed extends React.Component<Props, State> {
 
     feedContainer: null | HTMLElement;
 
+    componentDidMount() {
+        if (this.feedContainer) {
```
jstoffan: Can we avoid duplicating this logic in mount, update, and render by moving the if-check and assignment to a helper method?

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -46,6 +46,22 @@ class ActivityFeed extends React.Component<Props, State> {
...
+            this.feedContainer.scrollTop = this.feedContainer.scrollHeight;
+        }
+    }
+
+    componentDidUpdate(prevProps) {
+        if (
+            this.feedContainer &&
+            prevProps.feedItems === undefined &&
+            this.props.feedItems
```
jstoffan: It seems a little odd to check for `undefined` explicitly above, then check truthiness here. Is there a scenario where `feedItems` is null or zero?

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -188,7 +204,7 @@ class ActivityFeed extends React.Component<Props, State> {
                     <ApprovalCommentForm
                         onSubmit={() => {
```
ConradJChan: Do we have a preference to avoid anonymous functions in render methods?

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
alexkrolick: What if the length changes due to a new comment being added? Does that need special handling?

src/components/ContentSidebar/ActivityFeed/activity-feed/__tests__/ActivityFeed-test.js
```diff
@@ -135,6 +135,39 @@ describe('components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed', ()
...
 
+    test('should set scrollTop to be the scrollHeight if feedContainer ref is set', () => {
+        const wrapper = shallow(<ActivityFeed currentUser={currentUser} />);
+        const instance = wrapper.instance();
+        instance.feedContainer = {
+            scrollTop: 0,
+            scrollHeight: 100,
+        };
+
+        instance.componentDidMount();
```
alexkrolick: Instead of calling component instance methods, can we interact just through props and DOM events? That makes it possible to refactor the component without breaking the test.

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
ChenCodes: When a new comment is added, the onSubmit prop in ApprovalCommentForm will automatically scroll the div to the bottom

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
jeremypress: Ideally yes. This one can be avoided because it doesn't rely on any params?

src/api/Feed.js
ChenCodes: The add comment flow doesn't have a Promise.all that handles sorting. This setCachedItems is being called by addPendingItem on L845 which doesn't sort the feedItems

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
ChenCodes: It looks like feedItems can potentially be null. I'll adjust this to check for falsiness of prevProps.feedItems instead

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
ChenCodes: addressed

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
ChenCodes: addressed

src/api/Feed.js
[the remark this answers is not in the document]
ChenCodes: addressed

src/components/ContentSidebar/ActivityFeed/activity-feed/__tests__/ActivityFeed-test.js
ChenCodes: Will address this in a bit

ChenCodes:
Addressed all comments except for the test comment ! Please take another look
@DanDeMicco @jstoffan @alexkrolick @ConradJChan @jeremypress

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -46,6 +46,30 @@ class ActivityFeed extends React.Component<Props, State> {
...
 
+    componentDidMount() {
+        this.scrollFeedContainerToBottom(false);
+    }
+
+    componentDidUpdate(prevProps) {
+        this.scrollFeedContainerToBottom(true, prevProps);
+    }
+
+    scrollFeedContainerToBottom = (
```
ConradJChan: jsdoc?

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -46,6 +46,30 @@ class ActivityFeed extends React.Component<Props, State> {
...
+    ) => {
+        if (
+            componentWasUpdated &&
+            this.feedContainer &&
+            !prevProps.feedItems &&
+            this.props.feedItems
+        ) {
+            this.feedContainer.scrollTop = this.feedContainer.scrollHeight;
+        } else if (!componentWasUpdated && this.feedContainer) {
+            this.feedContainer.scrollTop = this.feedContainer.scrollHeight;
```
ConradJChan: it appears these two conditions lead to the same result, could they be combined somehow?

PR #623, opened 2018-10-31
Fix: moved saved versions to be within File Properties section

ChenCodes:
screen shot 2018-10-31 at 12 30 31 pm

PR #624, opened 2018-11-01
Update: flip sort order of preview sidebar items

ChenCodes:
screen shot 2018-10-31 at 9 51 37 pm

### DIFF
--- a/src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
+++ b/src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
@@ -72,6 +72,11 @@ class ActivityFeed extends React.Component<Props, State> {
         }
     }
 
+    /**
+     * Detects whether or not the empty state should be shown.
+     * @param {object} currentUser - The user that is logged into the account
+     * @param {object} feedItems - Items in the activity feed
+     */
     isEmpty = ({ currentUser, feedItems }: Props = this.props): boolean =>
         !currentUser ||
         !feedItems ||
--- a/src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
+++ b/src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
@@ -50,57 +50,56 @@ class ActivityFeed extends React.Component<Props, State> {
         this.scrollFeedContainerToBottom(false);
     }
 
-    componentDidUpdate(prevProps, prevState) {
+    componentDidUpdate(prevProps: Props, prevState: State) {
         this.scrollFeedContainerToBottom(true, prevProps, prevState);
     }
 
+    /**
+     * Determines when the sidebar should scroll to the bottom
+     * @param {boolean} componentWasUpdated - True if this call came from componentDidUpdate
+     * @param {object} prevProps - Exists only if this call came from componentDidUpdate
+     * @param {object} prevState - Exists only if this call came from componentDidUpdate
+     */
     scrollFeedContainerToBottom = (
         componentWasUpdated: boolean,
         prevProps?: Props,
         prevState?: State,
     ) => {
-        /**
-         * If our feedItems just got populated then we should scroll to the bottom.
-         */
-        if (
-            componentWasUpdated &&
-            this.feedContainer &&
-            !prevProps.feedItems &&
-            this.props.feedItems
-        ) {
-            this.feedContainer.scrollTop = this.feedContainer.scrollHeight;
-        } else if (
-            componentWasUpdated &&
-            this.feedContainer &&
-            prevProps.feedItems &&
-            prevProps.feedItems.length < this.props.feedItems.length
-        ) {
-            /**
-             * If we added a new comment we should scroll to the bottom.
-             */
-            this.feedContainer.scrollTop = this.feedContainer.scrollHeight;
-        } else if (!componentWasUpdated && this.feedContainer) {
-            /**
-             * On initial load we should scroll to the bottom
-             */
-            this.feedContainer.scrollTop = this.feedContainer.scrollHeight;
-        } else if (
-            this.feedContainer &&
-            this.state.isInputOpen !== prevState.isInputOpen
-        ) {
-            /**
-             * When we open the input we should scroll to the bottom
-             */
-            this.feedContainer.scrollTop = this.feedContainer.scrollHeight;
-        } else if (
-            prevProps &&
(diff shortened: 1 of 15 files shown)


boxcla:

boxcla:

Compared against previously merged PR #623

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -46,13 +46,72 @@ class ActivityFeed extends React.Component<Props, State> {
...
+     * @param {object} prevProps - Exists only if this call came from componentDidUpdate
+     * @param {object} prevState - Exists only if this call came from componentDidUpdate
+     */
+    scrollFeedContainerToBottom = (
+        componentWasUpdated: boolean,
+        prevProps?: Props,
+        prevState?: State,
+    ) => {
+        if (this.feedContainer) {
+            if (componentWasUpdated) {
```
jstoffan: This introduction of this boolean is adding a lot of complexity, since it requires existence checks on `prevProps` and `prevState`. Instead, focusing on reusing the common logic, which checks `feedContainer` and resets the `scrollTop` attribute.

```
    componentDidMount() {
        this.resetFeedScroll();
    }

    componentDidUpdate(prevProps, prevState: State) {
        const {
            currentUser: previousUser,
            feedItems: prevFeedItems,
        } = prevProps;
        const { currentUser, feedItems: currFeedItems } = this.props;
        const hasFewerItems = prevFeedItems && currFeedItems && prevFeedItems.length < currFeedItems.length;
        const hasNewItems = !prevFeedItems && currFeedItems;
        const hasNewUser = currentUser !== previousUser;

        if (hasFewerItems || hasNewItems || hasNewUser) {
            this.resetFeedScroll();
        }

        // TODO: Handle state changes similarly
    }

    resetFeedScroll() {
        if (this.feedContainer) {
            this.feedContainer.scrollTop = this.feedContainer.scrollHeight;
        }
    }
```

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.scss
```diff
@@ -38,7 +38,8 @@
         .bcs-activity-feed-task,
         .bcs-activity-feed-comment {
             border-bottom: 1px solid $ffive;
-            padding-bottom: 8px;
+            margin-top: 20px;
```
jstoffan: Are these CSS changes necessary to support the order change?

ChenCodes: I can move this into a separate PR
@jstoffan

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -46,13 +46,50 @@ class ActivityFeed extends React.Component<Props, State> {
...
+        const { currentUser, feedItems: currFeedItems } = this.props;
+        const { isInputOpen: prevIsInputOpen } = prevState;
+        const { isInputOpen: currIsInputOpen } = this.state;
+
+        const hasMoreItems =
+            prevFeedItems &&
+            currFeedItems &&
+            prevFeedItems.length < currFeedItems.length;
+        const hasNewItems = !prevFeedItems && currFeedItems;
+        const hasNewUser = currentUser !== previousUser;
```
jeremypress: Is this going from anonymous to logged in?

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -46,13 +46,50 @@ class ActivityFeed extends React.Component<Props, State> {
...
+        const hasInputOpened = currIsInputOpen !== prevIsInputOpen;
```
jeremypress: Can the input open such that the focus handler on the ApprovalCommentForm won't be called? Otherwise we may be able to remove that case.

ChenCodes: I don't think so. It looks like the only case of the input being set to open is on L96 - this.setState({ isInputOpen: true })

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
ChenCodes: The name for this is kind of confusing. The reason that I'm checking for the difference in user is I have a case where if the tab changes from ____ to Activity, I want to call resetScrollFeed(). I noticed that the prop that was changing was the user object (changing from undefined to defined) so I thought I would use that as an indicator of the tabs having changed

jstoffan: Hmm, okay. I think using the user object as a proxy for tab change is a bit fragile. Since tabs are unmounted when switching from one to another, shouldn't `componentDidMount` catch the case you describe?

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -186,11 +223,7 @@ class ActivityFeed extends React.Component<Props, State> {
                 </div>
                 {showApprovalCommentForm ? (
                     <ApprovalCommentForm
-                        onSubmit={() => {
-                            if (this.feedContainer) {
-                                this.feedContainer.scrollTop = 0;
-                            }
-                        }}
+                        onSubmit={() => this.resetFeedScroll()}
```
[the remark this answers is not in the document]
ChenCodes: Addressed

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -46,13 +46,56 @@ class ActivityFeed extends React.Component<Props, State> {
...
+            prevFeedItems.length < currFeedItems.length;
+        const hasNewItems = !prevFeedItems && currFeedItems;
+        const hasInputOpened = currIsInputOpen !== prevIsInputOpen;
+
+        if (hasLoaded || hasMoreItems || hasNewItems || hasInputOpened) {
+            this.resetFeedScroll();
+        }
+    }
+
+    isEmpty = ({ currentUser, feedItems }: Props = this.props): boolean =>
```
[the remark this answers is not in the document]
ChenCodes: Yep

PR #635, opened 2018-11-08
Fix: Metadata sidebar fixes

priyajeet:
- Remove classification and non-properties global templates from being able to be added via drop down
- Don't show sidebar if feature is disabled and no metadata to show
- Show sidebar even when feature is disabled but there is metadata to show

### DIFF
--- a/src/api/Metadata.js
+++ b/src/api/Metadata.js
@@ -18,6 +18,7 @@ import {
     METADATA_SCOPE_ENTERPRISE,
     METADATA_SCOPE_GLOBAL,
     METADATA_TEMPLATE_PROPERTIES,
+    METADATA_TEMPLATE_CLASSIFICATION,
     METADATA_TEMPLATE_SKILLS,
     FIELD_METADATA_SKILLS,
     CACHE_PREFIX_METADATA,
@@ -68,6 +69,20 @@ class Metadata extends File {
      */
     getMetadataTemplateUrl(scope: string): string {
         return `${this.getBaseApiUrl()}/metadata_templates/${scope}`;
+    }
+
+    /**
+     * Returns the custom properties template
+     *
+     * @return {Object} temaplte for custom properties
+     */
+    getCustomPropertiesTemplate(): MetadataEditorTemplate {
+        return {
+            id: uniqueId('metadata_template_'),
+            scope: METADATA_SCOPE_GLOBAL,
+            templateKey: METADATA_TEMPLATE_PROPERTIES,
+            hidden: false,
+        };
     }
 
     /**
@@ -125,7 +140,9 @@ class Metadata extends File {
         }
 
         return getProp(templates, 'data.entries', []).filter(
-            template => !template.hidden,
+            template =>
+                !template.hidden &&
+                template.templateKey !== METADATA_TEMPLATE_CLASSIFICATION,
         );
     }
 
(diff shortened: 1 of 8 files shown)


boxcla:

Compared against previously merged PR #632

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -308,15 +314,54 @@ class ContentSidebar extends PureComponent<Props, State> {
...
+        const canHaveMetadataSidebar =
+            !isFeatureEnabled &&
+            SidebarUtils.canHaveMetadataSidebar(this.props);
+
+        if (canHaveMetadataSidebar) {
+            this.api
+                .getMetadataAPI(true)
+                .getEditors(
+                    file,
+                    ({ editors }: { editors?: Array<MetadataEditor> }) =>
```
jstoffan: Consider moving these to named variable, since this is pretty hard to read right now.

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -308,15 +314,54 @@ class ContentSidebar extends PureComponent<Props, State> {
...
+     *
+     * @private
+     * @param {Object} file - Box file
+     * @return {void}
+     */
+    fetchFileSuccessCallback = (file: BoxItem): void => {
+        const { metadataSidebarProps }: Props = this.props;
+        const {
+            getMetadata,
+            isFeatureEnabled = true,
```
jstoffan: The naming here is unclear given that we're in ContentSidebar. Which feature is this referring to? Should the default value live in `defaultProps`?

src/components/ContentSidebar/SidebarUtils.js
```diff
@@ -95,7 +96,36 @@ class SidebarUtils {
...
+        editors?: Array<MetadataEditor>,
+    ): boolean {
+        const { metadataSidebarProps = {} }: ContentSidebarProps = props;
+        const {
+            isFeatureEnabled = true,
+        }: MetadataSidebarProps = metadataSidebarProps;
+
+        return (
+            SidebarUtils.canHaveMetadataSidebar(props) &&
+            (isFeatureEnabled || (Array.isArray(editors) && editors.length > 0))
```
jstoffan: Pretty difficult to read with the nested conditions. Consider breaking them out.

src/api/Metadata.js
```diff
@@ -501,23 +519,46 @@ class Metadata extends File {
         }
 
         try {
+            // Metadata instances
```
jstoffan: This method is turning into a bit of a monster. Can it be broken out into multiple methods?

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -182,7 +183,7 @@ class ContentSidebar extends PureComponent<Props, State> {
             this.fetchData(nextProps);
         } else if (!hasBeenToggled && hasVisibilityChanged) {
             this.setState({
-                view: this.getDefaultSidebarView(file, nextProps),
+                view: this.getDefaultSidebarView(nextProps, file, editors),
```
jstoffan: Can this concept be removed from state and just be derived from props in `render`? It seems like we're going through a lot of trouble to pass `file` and `editors` around to make this work.

src/api/Metadata.js
```diff
@@ -501,23 +519,46 @@ class Metadata extends File {
...
                 MetadataEditorTemplate,
-            > = customPropertiesTemplate
-                .concat(await this.getTemplates(id, METADATA_SCOPE_ENTERPRISE))
-                .concat(await this.getTemplates(id, METADATA_SCOPE_GLOBAL));
+            > = hasMetadataFeature
+                ? await this.getTemplates(id, METADATA_SCOPE_ENTERPRISE)
+                : [];
+
+            // Get global templates defined by box
+            const globalTemplates: Array<
```
DanDeMicco: Can these be done in parallel?

PR #640, opened 2018-11-13
Update: refresh activity feed cache in background when component mounted

DanDeMicco:


boxcla:

PR #641, opened 2018-11-13
feat(errors): add a default error component for uncaught exceptions

DanDeMicco:


boxcla:

src/components/ErrorBoundary/defaultError.scss
```diff
@@ -0,0 +1,5 @@
+.be {
```
priyajeet: Uppercase file name

src/components/ErrorBoundary/DefaultError.js
```diff
@@ -0,0 +1,23 @@
+// @flow
+import * as React from 'react';
+import ErrorMask from 'box-react-ui/lib/components/error-mask/ErrorMask';
+import { FormattedMessage } from 'react-intl';
+import messages from '../messages';
+import './defaultError.scss';
+
+const DefaultError = () => (
+    <section className="be-default-error">
+        <ErrorMask
```
priyajeet: BTW, I haven't seen ErrorMask used at other places, and when we had it for the sidebar before, product/design made me remove it in lieu of just a red error. Not sure whats the status of this component.

src/components/ErrorBoundary/DefaultError.js
```diff
@@ -0,0 +1,23 @@
+// @flow
+import * as React from 'react';
+import ErrorMask from 'box-react-ui/lib/components/error-mask/ErrorMask';
+import { FormattedMessage } from 'react-intl';
+import messages from '../messages';
+import './defaultError.scss';
```
priyajeet: Pascal case file name for scss

src/components/ErrorBoundary/ErrorBoundary.js
```diff
@@ -34,11 +35,13 @@ class ErrorBoundary extends React.Component<Props, State> {
     }
 
     render() {
-        if (this.state.error) {
-            return this.props.component;
+        const { errorComponent: ErrorComponent, children } = this.props;
```
priyajeet: Can you also add the props onError check above

src/components/ErrorBoundary/ErrorBoundary.js
```diff
@@ -34,11 +35,13 @@ class ErrorBoundary extends React.Component<Props, State> {
...
+        const { error } = this.state;
+        if (error) {
+            return <ErrorComponent error={error} />;
         }
 
-        return this.props.children;
+        return children;
```
jstoffan: If the value is only used once locally, is there a reason to destructure it at the top of the method?

src/components/ErrorBoundary/__tests__/ErrorBoundary-test.js
```diff
@@ -16,41 +16,43 @@ describe('components/ErrorBoundary', () => {
...
+    describe('componentDidCatch()', () => {
+        test('should set the error state and call the onError callback', () => {
             const onError = jest.fn();
             const wrapper = getWrapper({
                 onError,
             });
 
             simulateError(wrapper);
 
+            expect(wrapper.state('error')).toBe(wrappedError);
```
jstoffan: It doesn't seem ideal to test the internal logic and state of the component. Can we test its inputs/outputs, instead?

src/components/ErrorBoundary/__tests__/ErrorBoundary-test.js
```diff
@@ -16,41 +16,43 @@ describe('components/ErrorBoundary', () => {
...
+            const ErrorComponent = () => <div>Error</div>;
             const wrapper = getWrapper({
-                component: <div>Error</div>,
+                errorComponent: ErrorComponent,
             });
 
             simulateError(wrapper);
 
-            expect(wrapper).toMatchSnapshot();
+            expect(wrapper.find(ErrorComponent).exists()).toBe(true);
```
jstoffan: Does looking at the snapshot not assert this already?

src/components/ErrorBoundary/__tests__/ErrorBoundary-test.js
```diff
@@ -16,41 +16,43 @@ describe('components/ErrorBoundary', () => {
         wrapper.find(WrappedComponent).simulateError(wrappedError);
     };
 
-    describe('should render', () => {
+    describe('render()', () => {
```
jstoffan: Most of the test assertions read a bit oddly with this and their other changes:

`components/ErrorBoundary render() the wrapped component when no error is thrown`
`components/ErrorBoundary render() should render default error component an error is thrown`

jstoffan:
The commit message could be more clear about the reason for the change.

src/components/ErrorBoundary/__tests__/ErrorBoundary-test.js
DanDeMicco: I changed this to follow the existing pattern we use in EUA and elements

src/components/ErrorBoundary/ErrorBoundary.js
DanDeMicco: The defaultProp sets it to noop, so should be good

src/components/ErrorBoundary/DefaultError.js
DanDeMicco: Checking on which component we should use

src/components/ErrorBoundary/__tests__/ErrorBoundary-test.js
DanDeMicco: It does, but I find snapshot tests to be very brittle and it is not explicit what your output should be.

I normally wouldn't have touched this assertion, but this functionality was not working when I tested with a custom component (would only work with a dom node not a react component)

alexkrolick:
One of the react team made a reference error boundary implementation, maybe we could use that instead? https://www.npmjs.com/package/react-error-boundary

Then you just need to implement the fallback UI.

It's not super complex either way.

src/components/ErrorBoundary/ErrorBoundary.js
```diff
@@ -7,11 +7,12 @@
 import * as React from 'react';
 import noop from 'lodash/noop';
 import { ERROR_CODE_UNEXPECTED_EXCEPTION, IS_ERROR_DISPLAYED } from '../../constants';
+import DefaultError from './DefaultError';
 
 type Props = {
-    component: any,
+    ErrorComponent: React.Element,
```
jstoffan: Should this be lowercase? I think react-router uses lowercase for their route component prop.

src/components/ErrorBoundary/__tests__/ErrorBoundary-test.js
jstoffan: That seems fine, but can we make sure that the sentences are consistent and make sense when read? See the examples above.

PR #650, opened 2018-11-20
feat: add prop for emitting errors to ContentPreview, and ContentSidebar

DanDeMicco:


boxcla:

src/util/error.js
```diff
@@ -23,8 +23,36 @@ export function getBadUserError(): Error {
 
 export function isUserCorrectableError(status: number) {
     return (
+        !status ||
```
jstoffan: This will start showing "client side" errors to the user again, including CONNECTION_RESET, which we've been seeing due to proxy issues. Is this necessary as part of this change?

src/util/error.js
```diff
@@ -23,8 +23,36 @@ export function getBadUserError(): Error {
...
 }
+
+export function enhanceErrorForLogging(
+    error: Error,
+    type: ErrorTypes,
+    code: string,
+    contextInfo: Object = {},
+): ElementsError {
+    // $FlowFixMe
+    const enhancedError: ElementsError = error;
```
jstoffan: I think reassigning here will still modify the original object. Did you mean to make a copy?

src/components/ErrorBoundary/ErrorBoundary.js
```diff
@@ -27,18 +33,60 @@ class ErrorBoundary extends React.Component<Props, State> {
...
+
         if (this.state.error) {
-            return this.props.component;
+            return component;
         }
 
-        return this.props.children;
+        return React.cloneElement(this.props.children, {
+            ...rest,
+            ref: this.props.forwardedRef,
```
jstoffan: Needing to explicitly define a forwardedRef wasn't necessary previously and seems odd to surface as part of props. Can we use `forwardRef` instead?

src/components/ErrorBoundary/ErrorBoundary.js
```diff
@@ -27,18 +33,60 @@ class ErrorBoundary extends React.Component<Props, State> {
...
+        const enhancedError: ElementsError = enhanceErrorForLogging(e, type, code, contextInfo);
+
+        onError(enhancedError);
+    };
+
     render() {
+        const { children, component, ...rest } = this.props;
+
         if (this.state.error) {
-            return this.props.component;
```
jstoffan: Was this not working properly or it is a style-only change? The `component` property looks to only be used once.

src/components/ErrorBoundary/ErrorBoundary.js
```diff
@@ -27,18 +33,60 @@ class ErrorBoundary extends React.Component<Props, State> {
...
+    constructor(props: Props) {
+        super(props);
+        this.updateOnError(props.onError);
+    }
+
+    componentDidUpdate() {
+        this.updateOnError(this.props.onError);
+    }
+
+    componentDidCatch(error: Error, info: Object): void {
```
jstoffan: Minor, but I believe the second param, `errorInfo`, has a flow-type defined in React.

src/components/ContentSidebar/SkillsSidebar.js
```diff
@@ -196,4 +196,4 @@ class SkillsSidebar extends React.PureComponent<Props, State> {
 }
 
 export { SkillsSidebar as SkillsSidebarComponent };
-export default withErrorBoundary(withAPIContext(SkillsSidebar));
+export default withErrorBoundary(ERROR_TYPE_CONTENT_SIDEBAR)(withAPIContext(SkillsSidebar));
```
jstoffan: Should each sidebar have its own type rather than using its parent?

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -226,10 +229,16 @@ class ContentSidebar extends PureComponent<Props, State> {
...
      */
-    errorCallback = (error: $AxiosXHR<any>): void => {
+    errorCallback = (error: Error): void => {
         /* eslint-disable no-console */
         console.error(error);
         /* eslint-enable no-console */
+
+        /* eslint-disable react/prop-types */
+        this.props.onError(error, ERROR_TYPE_CONTENT_SIDEBAR, ERROR_CODE_FETCH_FILE, {
+            [IS_ERROR_DISPLAYED]: true,
```
jstoffan: It looks like this is already handled by `ErrorBoundary`?

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -226,10 +229,16 @@ class ContentSidebar extends PureComponent<Props, State> {
...
      * @return {void}
      */
-    errorCallback = (error: $AxiosXHR<any>): void => {
+    errorCallback = (error: Error): void => {
         /* eslint-disable no-console */
         console.error(error);
         /* eslint-enable no-console */
+
+        /* eslint-disable react/prop-types */
+        this.props.onError(error, ERROR_TYPE_CONTENT_SIDEBAR, ERROR_CODE_FETCH_FILE, {
```
jstoffan: If this is passed into the `withErrorBoundary` HOC, does it need to be passed in to `onError` each time? Can we avoid passing the name in somehow?

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -226,10 +229,16 @@ class ContentSidebar extends PureComponent<Props, State> {
      * @param {Error} error - Error object
      * @return {void}
      */
-    errorCallback = (error: $AxiosXHR<any>): void => {
+    errorCallback = (error: Error): void => {
         /* eslint-disable no-console */
         console.error(error);
         /* eslint-enable no-console */
+
+        /* eslint-disable react/prop-types */
```
jstoffan: Is this to avoid checking if `onError` is defined? Seems like we'd want to leave eslint on if possible here.

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -226,10 +229,16 @@ class ContentSidebar extends PureComponent<Props, State> {
      * @param {Error} error - Error object
      * @return {void}
      */
-    errorCallback = (error: $AxiosXHR<any>): void => {
+    errorCallback = (error: Error): void => {
```
jstoffan: Would this be `ElementsXhrError` now, as well? It looks like it's been updated in the child sidebars.

src/components/ContentSidebar/ActivitySidebar.js
```diff
@@ -405,7 +421,9 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
                 },
             },
         });
-        this.errorCallback(e);
+
+        const error = createErrorFromResponse(e);
+        this.errorCallback(error, ERROR_CODE_FETCH_CURRENT_USER);
```
jstoffan: It seems like error codes should be defined in the API layer rather than in the containers. Otherwise any consuming component would need to duplicate this logic.

src/api/Feed.js
```diff
@@ -138,9 +158,9 @@ class Feed extends Base {
 
         this.id = id;
         this.hasError = false;
-        const versionsPromise = this.fetchVersions();
-        const commentsPromise = this.fetchComments(permissions);
-        const tasksPromise = this.fetchTasks();
+        const versionsPromise = this.fetchVersions(onError);
```
jstoffan: Should `onError` instead be a `.catch` clause of `Promise.all` below? That way we can avoid passing the callback around and just throw errors from the child methods.

src/util/error.js
DanDeMicco: I added it cause `CONNECTION_RESET` is a user correctable action IMO. They could have network issues

src/components/ErrorBoundary/ErrorBoundary.js
DanDeMicco: I did this because I did not want to pass ```component``` prop to the children

src/components/ContentSidebar/ContentSidebar.js
DanDeMicco: Can you explain further?

src/components/ContentSidebar/ContentSidebar.js
DanDeMicco: It seems like its an eslint bug. I have it in the flow types above but it still complains

src/components/ContentSidebar/ContentSidebar.js
DanDeMicco: I assume you mean the ```type```? Yeah that can definitely be removed

src/components/ContentSidebar/ContentSidebar.js
DanDeMicco: 👍 good catch

src/components/ErrorBoundary/ErrorBoundary.js
```diff
@@ -27,18 +33,60 @@ class ErrorBoundary extends React.Component<Props, State> {
...
     render() {
+        const { children, component, ...rest } = this.props;
+
         if (this.state.error) {
-            return this.props.component;
+            return component;
         }
 
-        return this.props.children;
+        return React.cloneElement(this.props.children, {
```
jstoffan: It looks like `children` is destructured above.

src/util/error.js
DanDeMicco: It was intentional, I did not feel a deep clone was necessary. I think the error is pretty much thrown away after emitted

src/components/ErrorBoundary/ErrorBoundary.js
DanDeMicco: Will change this

src/util/error.js
jstoffan: In that case, you don't need the variable, right? Can you just add the property directly to `error`?

src/components/ContentSidebar/ContentSidebar.js
jstoffan: Sure, sorry. It looks like ErrorBoundary already sets `IS_ERROR_DISPLAYED` to true for uncaught exceptions. Can ErrorBoundary also handle decorating the `onError` callback to set the value there as well? It looks like it's always set to true right now.

src/components/ErrorBoundary/ErrorBoundary.js
DanDeMicco: I took a look, doesnt seem to be exposed. Maybe upgrading flow will fix it

src/components/ContentExplorer/ContentExplorer.js
```diff
@@ -1418,4 +1420,4 @@ class ContentExplorer extends Component<Props, State> {
 }
 
 export { ContentExplorer as ContentExplorerComponent };
-export default makeResponsive(ContentExplorer);
+export default withErrorBoundary(ERROR_TYPE_CONTENT_EXPLORER)(makeResponsive(ContentExplorer));
```
jeremypress: I thought we were only modifying the EUA used elements?

src/components/ContentSidebar/SkillsSidebar.js
jeremypress: This could be helpful as other teams start to inject in the sidebar.

src/constants.js
```diff
@@ -179,6 +179,30 @@ export const ERROR_CODE_ITEM_NAME_INVALID = 'item_name_invalid';
...
+export const ERROR_CODE_FETCH_TASK_ASSIGNMENT = 'fetch_task_assignment_error';
+export const ERROR_CODE_CREATE_COMMENT = 'create_comment_error';
+export const ERROR_CODE_CREATE_TASK = 'create_task_error';
+export const ERROR_CODE_CREATE_TASK_ASSIGNMENT = 'create_task_assignment_error';
+export const ERROR_CODE_DELETE_COMMENT = 'delete_comment_error';
+export const ERROR_CODE_DELETE_TASK = 'delete_task_error';
+export const ERROR_CODE_UPDATE_TASK = 'update_task_error';
+export const ERROR_CODE_UPDATE_TASK_ASSIGNMENT = 'update_task_assignment_error';
+export const ERROR_CODE_UNEXPECTED_EXCEPTION = 'unexpected_exception_error';
+export const ERROR_CODE_UNKNOWN = 'unknown_error';
```
jeremypress: I know it's a common error, but what about the 404 when fetching the avatar?

src/components/ContentPreview/ContentPreview.js
```diff
@@ -476,6 +485,19 @@ class ContentPreview extends PureComponent<Props, State> {
...
     }
 
+    /**
+     * Handler for 'preview_error' preview event
+     * @param {PreviewError} previewError - the error data emitted from preview
+     * @return {void}
+     */
+    onPreviewError = ({ error, ...rest }: PreviewError): void => {
+        const elementsError = createErrorFromResponse(error);
+        this.props.onError(elementsError, ERROR_TYPE_PREVIEW, error.code || ERROR_CODE_UNKNOWN, {
```
priyajeet: weird that eslint doesnt complain about this.props vs { onProps } = this.props

src/util/error.js
```diff
@@ -23,8 +23,36 @@ export function getBadUserError(): Error {
...
 export function isUserCorrectableError(status: number) {
     return (
+        !status ||
         status === HTTP_STATUS_CODE_RATE_LIMIT ||
         status === HTTP_STATUS_CODE_UNAUTHORIZED ||
         status >= HTTP_STATUS_CODE_INTERNAL_SERVER_ERROR
     );
 }
+
+export function enhanceErrorForLogging(
```
priyajeet: All exports should be at the bottom of the file

src/util/error.js
```diff
@@ -23,8 +23,36 @@ export function getBadUserError(): Error {
...
+        code,
+        context_info: {
+            ...contextInfo,
+        },
+    };
+
+    return enhancedError;
+}
+
+export function createErrorFromResponse(response: ElementsXhrError): Error {
```
priyajeet: export at the bottom

src/util/error.js
DanDeMicco: It was mostly for flow typing

src/components/ErrorBoundary/withErrorBoundary.js
```diff
@@ -1,17 +1,17 @@
...
- * @file withErrorBoundary HOC
+ * @file withErrorBoundary HOC which adds error boundaries as well as error error logging
  * @author Box
  */
 
 import * as React from 'react';
 import ErrorBoundary from './ErrorBoundary';
 
-const withErrorBoundary = (WrappedComponent: React.ComponentType<any>) =>
+const withErrorBoundary = (errorType: ErrorTypes) => (WrappedComponent: React.ComponentType<any>) =>
```
priyajeet: Minor, probably the type should also be singular `ErrorType` because reading it I thought it was array of errors

src/components/ErrorBoundary/ErrorBoundary.js
```diff
@@ -27,18 +33,60 @@ class ErrorBoundary extends React.Component<Props, State> {
...
+
+        onError(enhancedError);
+    };
+
     render() {
+        const { children, component, ...rest } = this.props;
+
         if (this.state.error) {
-            return this.props.component;
+            return component;
```
priyajeet: move to destructuring

src/components/ContentExplorer/ContentExplorer.js
DanDeMicco: I added this because all of the preview test files actually use `ContentExplorer`

src/components/ErrorBoundary/ErrorBoundary.js
```diff
@@ -27,18 +33,60 @@ class ErrorBoundary extends React.Component<Props, State> {
...
+            });
         });
     }
 
+    updateOnError(onErrorProp: ?Function) {
+        // We only want to update onError if it is the top level onError.
+        // For example if component tree is ContentPreview -> ContentSidebar, onError should be the
+        // onError prop passed to ContentPreview, and should be used for ContentSidebar
+        if (typeof onErrorProp === 'function') {
+            onError = onErrorProp;
```
jstoffan: If you use multiple Elements in an app, would this still work correctly?

src/util/error.js
DanDeMicco: As discussed, removing this for now

src/components/ContentSidebar/ContentSidebar.js
DanDeMicco: I can't totally remove it, since in `ContentPreview` we should distinguish between a preview and ContentPreview error. I will refactor slightly to add to end of the arguments and make it optional, defaulting to the prop

src/components/ContentSidebar/ActivitySidebar.js
DanDeMicco: Agreed

src/api/Feed.js
DanDeMicco: The problem with that is that the calls below are not actually rejecting when they fail, just resolving undefined due to displaying of partial data if possible.

If I were to throw and added a `.catch`, the `.then` would never be executed

src/components/ContentPreview/ContentPreview.js
```diff
@@ -703,19 +735,22 @@ class ContentPreview extends PureComponent<Props, State> {
                 isReloadNotificationVisible: true,
             });
         }
+
+        this.retryCount = RETRY_COUNT + 5;
```
jstoffan: Stray change?

src/components/ErrorBoundary/ErrorBoundary.js
```diff
@@ -27,18 +31,65 @@ class ErrorBoundary extends React.Component<Props, State> {
...
+        const { children, component, ...rest } = this.props;
+
         if (this.state.error) {
-            return this.props.component;
+            return component;
         }
 
-        return this.props.children;
+        // $FlowFixMe doesn't know about forwardRef (https://github.com/facebook/flow/issues/6103)
+        // return React.forwardRef((props: Object, ref: React.Ref<any>) =>
```
jstoffan: Stray commented code?

src/components/ErrorBoundary/ErrorBoundary.js
```diff
@@ -4,13 +4,17 @@
...
+import * as React from 'react';
 import noop from 'lodash/noop';
+import { ERROR_CODE_UNEXPECTED_EXCEPTION } from '../../constants';
 
 type Props = {
     children?: any,
     component: any,
-    onError: Function,
+    errorOrigin: ErrorOrigins,
+    forwardedRef: ?React.ElementRef<any>,
```
jstoffan: Unused?

src/components/ErrorBoundary/withErrorBoundary.js
```diff
@@ -1,17 +1,17 @@
 /**
  * @flow
- * @file withErrorBoundary HOC
+ * @file withErrorBoundary HOC which adds error boundaries as well as error error logging
```
jstoffan: `error error`

src/components/ErrorBoundary/ErrorBoundary.js
```diff
@@ -27,18 +31,65 @@ class ErrorBoundary extends React.Component<Props, State> {
...
+     * @param {Error} e - the error which occured
+     * @param {string} type - the error type to identify where the error occured
+     * @param {string} code - the error code to identify what error occured
+     * @param {Object} contextInfo - additional information which may be useful for the consumer of the error
+     * @return {void}
+     */
+    handleError = (
+        error: ElementsXhrError,
+        code: string,
+        isErrorDisplayed: boolean = false,
```
jstoffan: Does this need to be a separate param if it's being injected into `context_info` anyway? Seems like it could move into the object that's passed in.

src/components/ErrorBoundary/withErrorBoundary.js
```diff
@@ -1,17 +1,17 @@
...
 import * as React from 'react';
 import ErrorBoundary from './ErrorBoundary';
 
-const withErrorBoundary = (WrappedComponent: React.ComponentType<any>) =>
+const withErrorBoundary = (errorOrigin: ErrorOrigins) => (WrappedComponent: React.ComponentType<any>) =>
     // $FlowFixMe doesn't know about forwardRef (https://github.com/facebook/flow/issues/6103)
     React.forwardRef((props: Object, ref: React.Ref<any>) => (
-        <ErrorBoundary>
-            <WrappedComponent {...props} ref={ref} />
+        <ErrorBoundary {...props} ref={ref} errorOrigin={errorOrigin}>
```
jstoffan: Moving the `ref` from the wrapped component to the error boundary is likely going to cause bugs.

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -226,10 +228,16 @@ class ContentSidebar extends PureComponent<Props, State> {
...
      * @return {void}
      */
-    errorCallback = (error: $AxiosXHR<any>): void => {
+    errorCallback = (error: ElementsXhrError): void => {
         /* eslint-disable no-console */
         console.error(error);
         /* eslint-enable no-console */
+
+        /* eslint-disable react/prop-types */
+        this.props.onError(error, ERROR_CODE_FETCH_FILE, true, {
```
jstoffan: If you destructure `onError` per @priyajeet's suggestion, does eslint still complain?

DanDeMicco: Yes it does, weird.

src/constants.js
```diff
@@ -179,6 +179,31 @@ export const ERROR_CODE_ITEM_NAME_INVALID = 'item_name_invalid';
...
+
+/* ------------------ Error Origins ---------------------- */
+export const ORIGIN_CONTENT_PREVIEW = 'content_preview';
```
priyajeet: Do we need this granularity? of defining the origin which sidebar it is?
Since error codes are specific to origins, you can identify the origin from the error code.
Also, for high level UI Elements, cant we just re-use X-Box-Client-Name header value...

src/components/ErrorBoundary/withErrorBoundary.js
DanDeMicco: The ref is being forwarded there as well

DanDeMicco: Ah I see what you mean, moving it back to the wrappedComponent

src/api/Feed.js
```diff
@@ -25,6 +24,18 @@ import {
     VERSION_RESTORE_ACTION,
     TYPED_ID_FEED_PREFIX,
     HTTP_STATUS_CODE_CONFLICT,
+    ERROR_CODE_FETCH_COMMENTS,
```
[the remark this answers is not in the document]
priyajeet: We do this for resin btw

priyajeet:
Make sure to remove breaking change from the body

src/api/Feed.js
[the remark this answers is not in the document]
DanDeMicco: I will revisit this in a future commit

src/api/Feed.js
```diff
@@ -678,21 +673,40 @@ class Feed extends Base {
...
-    errorCallback = (error: $AxiosXHR<any>): void => {
-        console.error(error); // eslint-disable-line no-console
-        this.hasError = true;
+    errorCallback = (
+        hasError: boolean = false,
+        e: ElementsXhrError,
+        errorCallback?: ErrorCallback,
+        code?: string,
+    ): void => {
+        if (hasError) {
```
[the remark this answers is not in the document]
DanDeMicco: It determines if the UI should show an error or not. This will be false in the case it is not a user correctable error, but should still call the error callback

src/api/Feed.js
```diff
@@ -678,21 +673,40 @@ class Feed extends Base {
...
+            this.hasError = true;
+        }
+
+        if (!this.isDestroyed() && errorCallback && code) {
+            errorCallback(e, code, {
+                error: e,
+                [IS_ERROR_DISPLAYED]: hasError,
```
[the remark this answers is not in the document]
DanDeMicco: That makes sense in most cases, but in this case the Feed API determines if the consumer should display an error

src/constants.js
DanDeMicco: Granularity can be useful for uncaught exceptions in error boundaries

PR #658, opened 2018-11-27
refactor(sidebar): cleanup of content sidebar

priyajeet:
- make object attributes in ascending order
- remove cwrp
- default state properties
- break up file fetch and metadata fetch

Use Cases
- When the user has not interfered with the sidebar
  - The sidebar should open when viewport is large enough (current threshold 1000)
  - The sidebar should close when viewport is small
  - When navigating from file to file, tabs should switch based on this order of precedence: skills > activity > details > metadata

- When the user has interfered with the sidebar
  - Sidebar should remain closed if the user closed it when navigating between files
  - Sidebar should stick to prior chosen tab if it exists
  - Sidebar should revert to the tab based on above precedence if prior tab doesnt exist and then maintain this tab for future files

- When navigating between files sidebar should show loading indicator
- When opening file for 1st time, then sidebar should only show up after file load, so that we dont have two loading indicators
- Sidebar should not show on shared file page unless skills exist
- Sidebar should show for shared folder page with details and optionally skills

### DIFF
--- a/src/components/ContentSidebar/ContentSidebar.js
+++ b/src/components/ContentSidebar/ContentSidebar.js
@@ -69,7 +69,7 @@ type State = {
     hasBeenManuallyToggled: boolean,
     isLoading: boolean,
     metadataEditors?: Array<MetadataEditor>,
-    view: SidebarView,
+    view?: SidebarView,
 };
 
 class ContentSidebar extends React.PureComponent<Props, State> {
@@ -127,7 +127,7 @@ class ContentSidebar extends React.PureComponent<Props, State> {
             token,
         });
 
-        this.state = { hasBeenManuallyToggled: false, isLoading: true, view: SIDEBAR_VIEW_NONE };
+        this.state = { hasBeenManuallyToggled: false, isLoading: true };
     }
 
     /**
@@ -267,7 +267,7 @@ class ContentSidebar extends React.PureComponent<Props, State> {
 
         // Only reset the view if prior view is no longer applicable
         if (
-            view === SIDEBAR_VIEW_NONE ||
+            !view ||
             (view === SIDEBAR_VIEW_SKILLS && !canDefaultToSkills) ||
             (view === SIDEBAR_VIEW_ACTIVITY && !canDefaultToActivity) ||
             (view === SIDEBAR_VIEW_DETAILS && !canDefaultToDetails) ||
@@ -402,7 +402,7 @@ class ContentSidebar extends React.PureComponent<Props, State> {
             'be bcs',
             {
                 [`bcs-${((view: any): string)}`]: !!view,
-                'bcs-is-open': !!view,
+                'bcs-is-open': !!view && view !== SIDEBAR_VIEW_NONE,
             },
             className,
         );
--- a/src/components/ContentSidebar/ContentSidebar.js
+++ b/src/components/ContentSidebar/ContentSidebar.js
@@ -94,13 +94,6 @@ class ContentSidebar extends React.PureComponent<Props, State> {
         metadataSidebarProps: {},
     };
 
-    initialState: State = {
-        file: undefined,
-        hasBeenManuallyToggled: false,
-        metadataEditors: undefined,
-        view: SIDEBAR_VIEW_NONE,
-    };
-
     /**
      * [constructor]
      *
@@ -132,8 +125,7 @@ class ContentSidebar extends React.PureComponent<Props, State> {
             token,
         });
 
-        // Clone initial state to allow for state reset on new files
-        this.state = { ...this.initialState };
+        this.state = { hasBeenManuallyToggled: false, view: SIDEBAR_VIEW_NONE };
     }
 
     /**
@@ -260,6 +252,7 @@ class ContentSidebar extends React.PureComponent<Props, State> {
         const canDefaultToMetadata = SidebarUtils.shouldRenderMetadataSidebar(this.props, metadataEditors);
 
         // Calculate the default view with latest props
+        // This is the current order of importance for defaulting.
         if (canDefaultToSkills) {
             newView = SIDEBAR_VIEW_SKILLS;
         } else if (canDefaultToActivity) {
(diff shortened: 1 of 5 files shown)


boxcla:

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -81,22 +80,25 @@ class ContentSidebar extends PureComponent<Props, State> {
...
+        hasMetadata: false,
+        hasSkills: false,
+        isLarge: true,
         metadataSidebarProps: {},
     };
 
     initialState: State = {
         file: undefined,
+        hasBeenManuallyToggled: false,
+        metadataEditors: undefined,
```
jstoffan: This is already undefined when left out of the initial state.

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -248,31 +232,32 @@ class ContentSidebar extends PureComponent<Props, State> {
      * @param {Object} file - Box file
      * @return {string} Sidebar view to use
      */
-    getDefaultSidebarView(props: Props, file?: BoxItem, editors?: Array<MetadataEditor>): SidebarView {
-        const { view, hasBeenToggled }: State = this.state;
-        const { isLarge, defaultView }: Props = props;
+    getDefaultSidebarView(file?: BoxItem, metadataEditors?: Array<MetadataEditor>): SidebarView {
```
jstoffan: We had talked previously about moving this to being derived in `render` rather than living in state. It seems like that would clean up the logic around current/default view considerably. Do you want to tackle that as part of this change?

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -300,68 +285,84 @@ class ContentSidebar extends PureComponent<Props, State> {
...
+
+        // Only need to fetch metadata if the feature is disabled
+        // but the parent app has not hidden the metadata sidebar because
+        // the file may have metadata. Use case of this would be a free
+        // user who doesn't have the feature but is collabed on a file
+        // from a user who added metadata on the file.
+        // If the feature is enabled we always end up showing the metadata
+        // sidebar irrespective of there being any existing metadata or not.
+        // In that case let the metadata sidebar fetch the metadata when
+        // the user clicks on that tab.
```
jstoffan: Can this be summarized a bit more?

src/components/ContentSidebar/ContentSidebar.js
priyajeet: I mentioned to you that we cant do that. We need to keep view in state, since the future view depends upon the prior view. Also, once new view is calculated, that needs to be set to the state, hence it cannot be part of render

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -37,41 +37,40 @@ import '../modal.scss';
 import './ContentSidebar.scss';
 
 type Props = {
-    fileId?: string,
-    isLarge?: boolean,
-    clientName: string,
+    activitySidebarProps: ActivitySidebarProps,
```
priyajeet: Changes here are only alphabetical sorting

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -81,24 +80,20 @@ class ContentSidebar extends PureComponent<Props, State> {
     api: API;
 
     static defaultProps = {
+        activitySidebarProps: {},
```
priyajeet: Changes here are only alphabetical sorting

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -108,30 +103,29 @@ class ContentSidebar extends PureComponent<Props, State> {
     constructor(props: Props) {
         super(props);
         const {
-            cache,
-            token,
-            sharedLink,
-            sharedLinkPassword,
             apiHost,
+            cache,
```
priyajeet: Changes here are only alphabetical sorting

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -108,30 +103,29 @@ class ContentSidebar extends PureComponent<Props, State> {
...
             clientName,
             requestInterceptor,
             responseInterceptor,
+            sharedLink,
```
priyajeet: Changes here are only alphabetical sorting

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -296,72 +274,88 @@ class ContentSidebar extends PureComponent<Props, State> {
...
      *
      * @private
-     * @param {string} id - File id
      * @param {Object|void} [fetchOptions] - Fetch options
      * @return {void}
      */
-    fetchFile(id: string, fetchOptions: FetchOptions = {}): void {
-        if (SidebarUtils.canHaveSidebar(this.props)) {
-            this.api.getFileAPI().getFile(id, this.fetchFileSuccessCallback, this.errorCallback, {
+    fetchFile(fetchOptions: FetchOptions = {}): void {
```
priyajeet: This function should completely go away if/when each sidebar fetch their own file object

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -296,72 +274,88 @@ class ContentSidebar extends PureComponent<Props, State> {
...
-     * Only set file if there is data to show in the sidebar.
-     * Skills sidebar doesn't show when there is no data.
+     * Fetches file metadata editors if required
      *
      * @private
-     * @param {Object} file - Box file
      * @return {void}
      */
-    fetchFileSuccessCallback = (file: BoxItem): void => {
+    fetchMetadata(): void {
```
priyajeet: This function will remain and change, so will also account for skills data too, since skills is part of metadata. Right now we are fetching skills twice, one via /files?fields=metadata.global.boxSkillsCards and once via /files/metadata

DanDeMicco:
Should this use refactor: as part of the commit message?

PR #660, opened 2018-11-28
fix(sidebar): Use Metadata API for classification

JustinHoldstock:


boxcla:

src/components/ContentSidebar/SidebarFileProperties.js
```diff
@@ -55,7 +57,7 @@ const SidebarFileProperties = ({
     isLoading,
     intl,
 }: Props) => {
-    const value = getProp(file, `${FIELD_METADATA_CLASSIFICATION}.${KEY_CLASSIFICATION_TYPE}`);
+    const value = getProp(classificationInfo, `${KEY_CLASSIFICATION_TYPE}`);
```
priyajeet: ` ` not needed

src/api/Metadata.js
```diff
@@ -257,6 +267,58 @@ class Metadata extends File {
...
+     * Gets classification for a file.
+     *
+     * @param {BoxItem} file - File object for which we are getting classification
+     * @param {Function} successCallback - Success callback
+     * @param {Function} errorCallback - Error callback
+     * @param {boolean} forceFetch - True to clear the cache and make an api call
+     * @return {Promise}
+     */
+    async getClassification(
+        file: BoxItem,
```
priyajeet: Since you are not using the file object anymore to get classification out, get rid of it from the SIDEBAR PROPERTIES TO FETCH. In this way metadata (at least this one) is isolated doesn't break the files API call

src/api/Metadata.js
```diff
@@ -257,6 +267,58 @@ class Metadata extends File {
...
+        }
+
+        const cache: APICache = this.getCache();
+        const key = this.getClassificationCacheKey(id);
+        this.successCallback = successCallback;
+        this.errorCallback = errorCallback;
+
+        // Clear the cache if needed
+        if (forceFetch) {
+            cache.unset(key);
```
ConradJChan: Even if we `forceFetch`, should we be clearing the cache entry? I'm thinking that in case the API errors out, at least we are in the same state as before versus having lost what we had already fetched

JustinHoldstock: It appears that ALL metadata calls clear out their cache if that flag is present. I think we want to to remain consistent. In regards to having what we had before, the calling component's state shouldn't be updated with new data, so in any case, old data will still be present in the UI

priyajeet: There is another option, that is not yet in this method (can be added if needed) which is `refreshCache`, which is when you dont unset the cache but still fetch. Most api options take both forceFetch and refreshCache.

src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -196,52 +201,36 @@ class DetailsSidebar extends React.PureComponent<Props, State> {
...
     }
 
-    /**
-     * Sucess callback for classification change
-     *
-     * @private
-     * @return {void}
-     */
-    classifiationChangeSuccessCallback = (file: BoxItem) => {
+    fetchClassificationSuccessCallback = (classificationInfo: any): void => {
```
DanDeMicco: Can you type this?

src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -196,52 +201,36 @@ class DetailsSidebar extends React.PureComponent<Props, State> {
...
     };
 
-    /**
-     * Error callback for classification change
-     *
-     * @private
-     * @return {void}
-     */
-    classifiationChangeErrorCallback = () => {
+    fetchClassificationErrorCallback = (error: $AxiosXHR<any>): void => {
```
DanDeMicco: error type is incorrect

src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -196,52 +201,36 @@ class DetailsSidebar extends React.PureComponent<Props, State> {
...
         this.setState({
```
DanDeMicco: Can you utilize the `onError` which I recently added?

src/components/ContentSidebar/SidebarFileProperties.js
```diff
@@ -11,10 +11,11 @@ import ItemProperties from 'box-react-ui/lib/features/item-details/ItemPropertie
 import LoadingIndicatorWrapper from 'box-react-ui/lib/components/loading-indicator/LoadingIndicatorWrapper';
 import getFileSize from 'box-react-ui/lib/utils/getFileSize';
 import withErrorHandling from './withErrorHandling';
-import { FIELD_METADATA_CLASSIFICATION, KEY_CLASSIFICATION_TYPE } from '../../constants';
+import { KEY_CLASSIFICATION_TYPE } from '../../constants';
 import { INTERACTION_TARGET, DETAILS_TARGETS } from '../../interactionTargets';
 
 type Props = {
+    classificationInfo?: any,
```
DanDeMicco: Can you add a flow type?

src/api/Metadata.js
DanDeMicco: IMO we should implement refreshCache here

src/components/ContentSidebar/DetailsSidebar.js
DanDeMicco: This (along with error callback) has a typeo. Can you correct it?

DanDeMicco: and add jsdocs

src/components/ContentSidebar/DetailsSidebar.js
DanDeMicco: jsdocs here as well

src/components/ContentSidebar/DetailsSidebar.js
JustinHoldstock: Yup, wanted to get structure down before adding new types (If I need to) and handling via onError

boxcla:

Compared against previously merged PR #656

flow-typed/box-ui-elements.js
```diff
@@ -671,3 +671,15 @@ type ErrorContextProps = {
         origin: ErrorOrigins,
     ) => void,
 };
+
+type ClassificationInfo = {
+    $canEdit: boolean,
```
priyajeet: Everything with $ is generic for metadata instance, so you can be like MetadataInstance & ClassificationInfo

flow-typed/box-ui-elements.js
```diff
@@ -485,10 +485,10 @@ type Errors = {
 };
 
 type FileAccessStats = {
-    preview_count: number,
-    download_count: number,
-    comment_count: number,
-    edit_count: number,
+    preview_count?: number,
```
priyajeet: Part of this commit?

priyajeet: Did 2 commits get combined?

src/api/Metadata.js
```diff
@@ -257,6 +267,71 @@ class Metadata extends File {
...
+        }
+
+        // Return the Cache value if it exists
+        if (cache.has(key)) {
+            const cachedItem = cache.get(key);
+            if (cachedItem) {
+                const { hasError } = cachedItem;
```
priyajeet: Whats hasError?
Also if there is an error, then don't cache it in the 1st place

src/api/Metadata.js
```diff
@@ -257,6 +267,71 @@ class Metadata extends File {
...
+     * @param {Function} errorCallback - Error callback
+     * @param {boolean} forceFetch - True to clear the cache and make an api call
+     * @param {boolean} refreshCache - True to
+     * @return {Promise}
+     */
+    async getClassification(
+        file: BoxItem,
+        successCallback: Function,
+        errorCallback: Function,
+        forceFetch: boolean = false,
```
priyajeet: { forceFetch, refreshCache } as object options. See File.js
flowtype is FetchOptions

src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -197,49 +218,69 @@ class DetailsSidebar extends React.PureComponent<Props, State> {
...
      *
      * @private
+     * @param {ElementsXhrError} error - API error
      * @return {void}
      */
-    classifiationChangeErrorCallback = () => {
+    fetchClassificationErrorCallback = (error: ElementsXhrError): void => {
+        let classificationError;
+
+        if (getProp(error, 'status') === HTTP_STATUS_CODE_FORBIDDEN) {
```
priyajeet: Why do we need to show error? If the feature is turned off, then hasClassification should be false from the webapp level.

src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -269,7 +310,16 @@ class DetailsSidebar extends React.PureComponent<Props, State> {
...
 
-        const { accessStats, accessStatsError, file, fileError, isLoading }: State = this.state;
+        const {
+            accessStats,
+            accessStatsError,
+            classificationInfo,
+            classificationError,
+            file,
+            fileError,
+            isLoadingAccessStats,
```
priyajeet: Do we need 2 loading states? Are they being used locally? If not, then you can just:

componentDidMount() {
    this.fetchData();
}

fetchData() {
     if (hasAccessStats || hasClassification) {
     }
}

success/error callbacks for each of the 2, setState({ isLoading: false })

src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -269,7 +310,16 @@ class DetailsSidebar extends React.PureComponent<Props, State> {
             bannerPolicy,
         }: Props = this.props;
 
-        const { accessStats, accessStatsError, file, fileError, isLoading }: State = this.state;
+        const {
+            accessStats,
+            accessStatsError,
+            classificationInfo,
```
priyajeet: nitpick, can we just call it classification to match accessStats? or change accessStats to accessStatsInfo

src/components/ContentSidebar/DetailsSidebar.js
priyajeet: I do notice you are &&ing below, shouldn't the details properties show when avail...like not wait for everything to load?

flow-typed/box-ui-elements.js
JustinHoldstock: What the heck. That shouldn't be in there

flow-typed/box-ui-elements.js
JustinHoldstock: Like so, with new $props going to MetadataInstance?

```
type ClassificationInfo = {
    Box__Security__Classification__Key?: string
} & MetadataInstance;
```

src/api/Metadata.js
JustinHoldstock: Actually, lemme kill that. Didn't need it in the first place

src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -63,21 +66,35 @@ class DetailsSidebar extends React.PureComponent<Props, State> {
...
         }
+
+        if (hasClassification) {
+            this.fetchClassification();
+        }
+    }
+
+    componentWillUnmount() {
+        // Reset loading state to allow loading spinner
+        this.setState({
```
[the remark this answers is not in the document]
JustinHoldstock: Ah, didn't realize that. Will remove

src/components/ContentSidebar/DetailsSidebar.js
JustinHoldstock: Deal!

src/components/ContentSidebar/DetailsSidebar.js
JustinHoldstock: You mean the unauthorized error? Will remove, since that makes the most sense

JustinHoldstock: Ah, but in the case that a developer outside of webapp runs into this, though...

src/components/ContentSidebar/DetailsSidebar.js
JustinHoldstock: If classification comes back before access stats does, then we have no loader, but ALSO no access stats until they're loaded (which looks weird). I agree they should show before everything is ready, but the UI looks funky when things are popping in *without* a loading spinner

src/api/Metadata.js
```diff
@@ -257,6 +267,64 @@ class Metadata extends File {
...
+        }
+
+        const cache: APICache = this.getCache();
+        const key = this.getClassificationCacheKey(id);
+        this.successCallback = successCallback;
+        this.errorCallback = errorCallback;
+
+        // Clear the cache if needed
+        if (options.forceFetch) {
+            cache.unset(key);
```
DanDeMicco: why unset if its a forceFetch? Maybe just `if (cache.has(key) && !options.forceFetch)`

src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -158,10 +172,14 @@ class DetailsSidebar extends React.PureComponent<Props, State> {
         }
 
         this.setState({
-            isLoading: false,
+            isLoadingAccessStats: false,
             accessStats: undefined,
             accessStatsError,
         });
+
+        this.props.onError(error, ERROR_CODE_FETCH_ACCESS_STATS, {
```
DanDeMicco: This is addressed as part of https://github.com/box/box-ui-elements/pull/661/files so you can remove, thanks for adding it though :)

src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -197,48 +215,69 @@ class DetailsSidebar extends React.PureComponent<Props, State> {
...
+        }
+
         this.setState({
-            isLoading: false,
+            classification: undefined,
+            classificationError,
+            isLoadingClassification: false,
+        });
+
+        this.props.onError(error, ERROR_CODE_FETCH_CLASSIFICATION, {
```
DanDeMicco: In #661 I am changing each API to return the error code, not the component. Can you follow this pattern?

`(error: ElementsXhrError, code: string)`

src/api/Metadata.js
JustinHoldstock: This is how we use `forceFetch` in multiple places. Following suit, here. Leaving as is

src/components/ContentSidebar/DetailsSidebar.js
JustinHoldstock: DONE!

src/api/Metadata.js
priyajeet: unset is here, because the line below checks if its set below

src/components/ContentSidebar/DetailsSidebar.js
JustinHoldstock: done!

src/api/Metadata.js
```diff
@@ -257,6 +268,65 @@ class Metadata extends File {
...
+     */
+    async getClassification(
+        file: BoxItem,
+        successCallback: Function,
+        errorCallback: Function,
+        options: FetchOptions = {},
+    ): Promise<void> {
+        const { id }: BoxItem = file;
+        if (!id) {
+            errorCallback(getBadItemError());
```
DanDeMicco: One more minor thing, technically this should return the code as well

src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -282,6 +322,7 @@ class DetailsSidebar extends React.PureComponent<Props, State> {
                         onAccessStatsClick={onAccessStatsClick}
                         file={file}
                         {...accessStatsError}
+                        {...classificationError}
```
DanDeMicco: Why is `classificationError` being passed to access stats?

src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -197,48 +206,70 @@ class DetailsSidebar extends React.PureComponent<Props, State> {
...
      * @private
+     * @param {ElementsXhrError} error - API error
+     * @param {string} code - Error code
      * @return {void}
      */
-    classifiationChangeErrorCallback = () => {
+    fetchClassificationErrorCallback = (error: ElementsXhrError, code: string): void => {
+        let classificationError;
+
+        if (getProp(error, 'status') === HTTP_STATUS_CODE_FORBIDDEN) {
```
DanDeMicco: In what case will this be forbidden?

DanDeMicco: Do we want to show an error message in this case? I am not sure, but I think it should silently fail

src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -197,48 +206,70 @@ class DetailsSidebar extends React.PureComponent<Props, State> {
...
+            classificationError = {
+                error: messages.fileClassificationPermissionsError,
+            };
+        } else {
+            classificationError = {
```
DanDeMicco: I think this should be an inlineError

JustinHoldstock: done

src/components/ContentSidebar/DetailsSidebar.js
JustinHoldstock: Priyajeet also brought this up. I'll take it out, since external devs will see in a network req that the error is a 403

flow-typed/box-ui-elements.js
```diff
@@ -485,10 +489,10 @@ type Errors = {
 };
 
 type FileAccessStats = {
-    preview_count?: number,
-    download_count?: number,
-    comment_count?: number,
-    edit_count?: number,
+    preview_count: number,
```
priyajeet: Part of this PR? Or the other one

src/api/Metadata.js
```diff
@@ -257,6 +268,65 @@ class Metadata extends File {
...
+
+        // Clear the cache if needed
+        if (options.forceFetch) {
+            cache.unset(key);
+        }
+
+        // Return the Cache value if it exists
+        if (cache.has(key)) {
+            const cachedItem = cache.get(key);
+            if (cachedItem) {
```
priyajeet: null/undefined check needed?
since cache.has() checked above that there was an entry for sure.

src/components/ContentSidebar/SidebarFileProperties.js
```diff
@@ -55,7 +57,7 @@ const SidebarFileProperties = ({
     isLoading,
     intl,
 }: Props) => {
-    const value = getProp(file, `${FIELD_METADATA_CLASSIFICATION}.${KEY_CLASSIFICATION_TYPE}`);
+    const value = getProp(classification, KEY_CLASSIFICATION_TYPE);
```
priyajeet: probably a better variable name other than value

src/components/messages.js
```diff
@@ -883,6 +883,16 @@ const messages: { [string]: MessageDescriptor } = defineMessages({
         description: 'The text for when the user does not have permissions to see access stats.',
         defaultMessage: 'Sorry, you do not have permission to see the access stats for this file.',
     },
+    fileClassificationErrorHeaderMessage: {
+        id: 'be.fileClassificationErrorHeaderMessage',
+        description: 'File classification error message',
+        defaultMessage: 'Something went wrong when fetching classification.',
+    },
+    fileClassificationPermissionsError: {
```
priyajeet: unused?

flow-typed/box-ui-elements.js
JustinHoldstock: I'm honestly not sure what's going on here. When I rebased yesterday, it kept bumping this around. Will modify to see if it goes away

src/api/Metadata.js
JustinHoldstock: Great point!

src/components/ContentSidebar/SidebarFileProperties.js
JustinHoldstock: renamed to `classificationName`

src/components/messages.js
JustinHoldstock: It's dead

i18n/en-US.properties
```diff
@@ -124,6 +124,10 @@ be.expand = Expand
 be.faceSkill = Faces
 # File access stats error message
 be.fileAccessStatsErrorHeaderMessage = Something went wrong when fetching the access stats.
+# File classification error message
+be.fileClassificationErrorHeaderMessage = Something went wrong when fetching classification.
+# The text for when the user does not have permissions to see classification.
+be.fileClassificationPermissionsError = Sorry, you do not have permission to see classification for this file.
```
priyajeet: Did you regen this?

src/api/Metadata.js
```diff
@@ -257,6 +268,64 @@ class Metadata extends File {
...
+        this.errorCallback = errorCallback;
+
+        // Clear the cache if needed
+        if (options.forceFetch) {
+            cache.unset(key);
+        }
+
+        // Return the Cache value if it exists
+        if (cache.has(key)) {
+            const cachedItem = cache.get(key);
```
priyajeet: Unneeded constant

src/components/ContentSidebar/SidebarFileProperties.js
```diff
@@ -55,7 +57,7 @@ const SidebarFileProperties = ({
     isLoading,
     intl,
 }: Props) => {
-    const value = getProp(file, `${FIELD_METADATA_CLASSIFICATION}.${KEY_CLASSIFICATION_TYPE}`);
+    const classificationName = getProp(classification, KEY_CLASSIFICATION_TYPE);
```
priyajeet: classificationType?

JustinHoldstock: 👍

PR #651, opened 2018-11-21
build: semantic release and new script names for builds

priyajeet:
- Moves build folder to conf per new folder standard
- Switches build to use semantic release
- Updates release script eliminating publish/examples script
- Uses new release branch for releases

Usage

`yarn release`
Will run a release but will reset release branch to master before doing so.

`yarn release:hotfix`
Will run a release without resetting release branch to master. In this case it is assumed some patch has been cherry picked on the release branch for a hotfix release.

`source ./conf/cut_release.sh` followed by some function in that script.
If you want to run only certain aspects of the release script.

PR #661, opened 2018-11-29
refactor: add error codes to API error callbacks

DanDeMicco:
This is a follow up to #650. Mainly it adds error codes and additional flow typing to the API layer for 90% of API's which were sometimes added in components in the previous PR


boxcla:

boxcla:

Compared against previously merged PR #651

src/api/Base.js
```diff
@@ -349,8 +354,8 @@ class Base {
                 ...requestData,
             });
             this.successHandler(data);
-        } catch (error) {
-            this.errorHandler(error);
+        } catch (e) {
```
jstoffan: Curious for the reason behind this change? I think we usually want to avoid abbreviations.

src/api/Comments.js
```diff
@@ -5,7 +5,17 @@
  */
 
 import OffsetBasedAPI from './OffsetBasedAPI';
-import { PERMISSION_CAN_COMMENT, PERMISSION_CAN_DELETE, PERMISSION_CAN_EDIT } from '../constants';
+import {
+    PERMISSION_CAN_COMMENT,
+    PERMISSION_CAN_DELETE,
+    PERMISSION_CAN_EDIT,
+    ERROR_CODE_CREATE_COMMENT,
```
jstoffan: Would be great to put these behind a namespace, like `ERROR_CODE.CREATE_COMMENT` or `PERMISSION.CAN_COMMENT`

src/api/Comments.js
```diff
@@ -229,17 +242,21 @@ class Comments extends OffsetBasedAPI {
...
         try {
             this.checkApiCallValidity(PERMISSION_CAN_COMMENT, permissions, fileId);
         } catch (e) {
-            errorCallback(e);
+            errorCallback(e, this.errorCode);
             return;
         }
 
-        this.offsetGet(fileId, successCallback, errorCallback, ...rest);
+        this.offsetGet(fileId, successCallback, errorCallback, offset, limit, fields, shouldFetchAll);
```
jstoffan: Is this change related to the error codes? It seems like the defaults should live in the parent class rather than being duplicated in each child.

src/api/Feed.js
```diff
@@ -182,94 +170,66 @@ class Feed extends Base {
...
-                (e: ElementsXhrError) => {
-                    this.fetchFeedItemErrorCallback(e, errorCallback, ERROR_CODE_FETCH_VERSIONS);
-                    resolve();
-                },
-                DEFAULT_START,
-                DEFAULT_END,
-                VERSIONS_FIELDS_TO_FETCH,
-                true,
-            );
+            this.versionsAPI.getVersions(this.id, resolve, this.fetchFeedItemErrorCallback.bind(this, resolve));
```
jstoffan: It seems odd that the inner methods would resolve successfully after an error. Ideally, they would reject and the parent should catch and handle them appropriately.

src/api/Feed.js
```diff
@@ -297,6 +257,7 @@ class Feed extends Base {
         }
 
         this.id = file.id;
+        this.onError = errorCallback;
```
jstoffan: In some places we assign this value from the `onError` param and in others from the `errorCallback` param. It's not immediately clear what the distinction is or when to use one or the other.

src/api/Folder.js
```diff
@@ -315,7 +320,9 @@ class Folder extends Item {
                 },
             })
             .then(this.createSuccessHandler)
-            .catch(this.errorHandler);
+            .catch((e: $AxiosError<any>) => {
```
jstoffan: What's the value of wrapping this in an anonymous function?

src/api/Base.js
DanDeMicco: No particular reason. I saw e being used in a few places, so just changed it to be consistent

src/api/Comments.js
DanDeMicco: No not really, I can remove some of the defaults and just make them optional

src/api/Feed.js
DanDeMicco: Agreed. The reason why they resolve undefined is that `Promise.all` is being used to fetch the comments, tasks, versions. If we reject, we cannot show partial data as it will hit the .catch.

A custom promise.all could be implemented (or imported) to handle this. Another unfortunate thing about reject is that it can only return 1 value, but I need 2 (error and code from the error callback). We have an eslint rule that enforces it to be of type Error. So, I could create an Error object with the code stuck into it.

src/api/Feed.js
DanDeMicco: Would adding JSDoc to the `onError` property on the class help? Maybe changing the variable name?

```
     * @param {Function} errorCallback - the error callback which is called after data fetching is complete if there was an error
     * @param {Function} onError - the function to be called immediately after an error occurs
```

src/api/Folder.js
DanDeMicco: I think this is leftover from a refactor of this where I was before passing the error callback to the errorHandler, but then changed it to be a class property

PR #667, opened 2018-11-30
fix: Properly handle folder uploads with invalid names

patrickpaul:
Folders uploaded via ContentUploader with invalid names were not properly being handled. The creation of the folder would fail and any items contained within it would be added to the parent.

This PR adds messaging to the user when failures occur (either at the root folder level or in child folders). Additionally, it does not process child items for folders that fail to be created.

### DIFF
--- a/src/api/uploads/FolderUpload.js
+++ b/src/api/uploads/FolderUpload.js
@@ -158,12 +158,14 @@ class FolderUpload {
         successCallback: Function,
     }): Promise<any> {
         await this.folder.upload(this.destinationFolderId, errorCallback, true);
-        // Simulate BoxItem
-        successCallback([
-            {
-                id: this.folder.folderId,
-            },
-        ]);
+        // If the folder upload failed then a folderID will not be set
+        if (this.folder.folderId) {
+            successCallback([
+                {
+                    id: this.folder.folderId,
+                },
+            ]);
+        }
     }
 
     /**
--- a/src/api/uploads/FolderUploadNode.js
+++ b/src/api/uploads/FolderUploadNode.js
@@ -5,7 +5,12 @@
  */
 import noop from 'lodash/noop';
 import FolderAPI from '../Folder';
-import { STATUS_COMPLETE, ERROR_CODE_ITEM_NAME_IN_USE } from '../../constants';
+import {
+    STATUS_COMPLETE,
+    STATUS_ERROR,
+    ERROR_CODE_CHILD_FOLDER_FAILED_UPLOAD,
+    ERROR_CODE_ITEM_NAME_IN_USE,
+} from '../../constants';
 import { getFileFromEntry } from '../../util/uploads';
 
 class FolderUploadNode {
@@ -66,7 +71,12 @@ class FolderUploadNode {
         this.parentFolderId = parentFolderId;
 
         await this.createAndUploadFolder(errorCallback, isRoot);
+        if (!this.getFolderId()) {
+            // Folder was not successfully created. Do not attempt to upload its contents.
+            return;
+        }
         this.addFilesToUploadQueue(this.getFormattedFiles(), noop, true);
+
         await this.uploadChildFolders(errorCallback);
     }
 
@@ -96,17 +106,28 @@ class FolderUploadNode {
     createAndUploadFolder = async (errorCallback: Function, isRoot: boolean) => {
         await this.buildCurrentFolderFromEntry();
 
+        let errorEncountered = false;
+        let errorCode = '';
         try {
             const data = await this.createFolder();
             this.folderId = data.id;
         } catch (error) {
+            errorEncountered = true;
+            errorCode = error.code;
             // @TODO: Handle 429
             if (error.code !== ERROR_CODE_ITEM_NAME_IN_USE) {
-                errorCallback(error);
-                return;
+                if (!isRoot) {
+                    // If this is a child folder of the folder being uploaded, this errorCallback will set
+                    // an error message on the root folder being uploaded. Set a generic messages saying that a
+                    // child has caused the error. The child folder will be tagged with the error message in
+                    // the call to this.addFolderToUploadQueue below
+                    errorCallback({ code: ERROR_CODE_CHILD_FOLDER_FAILED_UPLOAD });
+                } else {
+                    errorCallback(error);
+                }
+            } else {
+                this.folderId = error.context_info.conflicts[0].id;
             }
-
-            this.folderId = error.context_info.conflicts[0].id;
         }
 
         // The root folder has already been added to the upload queue in ContentUploader
@@ -114,16 +135,29 @@ class FolderUploadNode {
             return;
         }
 
-        this.addFolderToUploadQueue([
-            {
-                extension: '',
-                name: this.name,
-                status: STATUS_COMPLETE,
-                isFolder: true,
-                size: 1,
-                progress: 100,
-            },
-        ]);
+        const folderObject: {
+            extension: string,
+            name: string,
+            status: string,
+            isFolder: boolean,
+            size: number,
+            progress: number,
+            error?: Object,
+        } = {
+            extension: '',
+            name: this.name,
+            status: STATUS_COMPLETE,
+            isFolder: true,
+            size: 1,
+            progress: 100,
+        };
+
+        if (errorEncountered) {
+            folderObject.status = STATUS_ERROR;
+            folderObject.error = { code: errorCode };
+        }
+
+        this.addFolderToUploadQueue([folderObject]);
     };
 
     /**
@@ -230,6 +264,14 @@ class FolderUploadNode {
 
             this.readEntry(reader, resolve);
         });
+    };
+
+    /**
+     * Returns the folderId
+     * @returns {string}
+     */
+    getFolderId = (): string => {
+        return this.folderId;
     };
 }
 
--- a/src/components/ContentUploader/progressCellRenderer.js
+++ b/src/components/ContentUploader/progressCellRenderer.js
@@ -6,7 +6,7 @@
 import React from 'react';
 import { FormattedMessage } from 'react-intl';
 import ItemProgress from './ItemProgress';
-import { STATUS_ERROR, STATUS_IN_PROGRESS } from '../../constants';
+import { STATUS_ERROR, STATUS_IN_PROGRESS, ERROR_CODE_CHILD_FOLDER_FAILED_UPLOAD } from '../../constants';
 
 import messages from '../messages';
 
@@ -18,12 +18,19 @@ type Props = {
  * Get error message for a specific error code
  *
  * @param {string} [errorCode]
+ * @param {string} [itemName]
  * @returns {FormattedMessage}
  */
-const getErrorMessage = (errorCode: ?string) => {
+const getErrorMessage = (errorCode: ?string, itemName: ?string) => {
     switch (errorCode) {
+        case ERROR_CODE_CHILD_FOLDER_FAILED_UPLOAD:
+            return <FormattedMessage {...messages.uploadsOneOrMoreChildFoldersFailedToUploadMessage} />;
         case 'file_size_limit_exceeded':
             return <FormattedMessage {...messages.uploadsFileSizeLimitExceededErrorMessage} />;
+        case 'item_name_invalid':
+            return (
+                <FormattedMessage {...messages.uploadsProvidedFolderNameInvalidMessage} values={{ name: itemName }} />
+            );
         case 'storage_limit_exceeded':
             return <FormattedMessage {...messages.uploadsStorageLimitErrorMessage} />;
         case 'pending_app_folder_size_limit':
@@ -34,7 +41,7 @@ const getErrorMessage = (errorCode: ?string) => {
 };
 
 export default () => ({ rowData }: Props) => {
-    const { status, error = {}, isFolder } = rowData;
+    const { status, error = {}, name, isFolder } = rowData;
     const { code } = error;
 
     if (isFolder && status !== STATUS_ERROR) {
@@ -45,7 +52,7 @@ export default () => ({ rowData }: Props) => {
         case STATUS_IN_PROGRESS:
             return <ItemProgress {...rowData} />;
         case STATUS_ERROR:
-            return getErrorMessage(code);
+            return getErrorMessage(code, name);
         default:
             return null;
     }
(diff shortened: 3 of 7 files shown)


boxcla:
Hi @patrickpaul, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

patrickpaul:
CLA signed

boxcla:

Compared against previously merged PR #664

src/api/uploads/FolderUploadNode.js
```diff
@@ -66,7 +71,12 @@ class FolderUploadNode {
         this.parentFolderId = parentFolderId;
 
         await this.createAndUploadFolder(errorCallback, isRoot);
+        if (!this.getFolderId()) {
```
patrickpaul: FYI - this does not use this.folderId because I need to be able to mock it for tests.

src/api/uploads/FolderUpload.js
```diff
@@ -158,12 +158,14 @@ class FolderUpload {
...
     }): Promise<any> {
         await this.folder.upload(this.destinationFolderId, errorCallback, true);
-        // Simulate BoxItem
-        successCallback([
-            {
-                id: this.folder.folderId,
-            },
-        ]);
+        // If the folder upload failed then a folderID will not be set
+        if (this.folder.folderId) {
```
patrickpaul: Could change this to `this.folder.getFolderId()` for consistency, even though they're synonymous.

src/constants.js
```diff
@@ -179,6 +179,7 @@ export const ERROR_CODE_ITEM_NAME_INVALID = 'item_name_invalid';
 export const ERROR_CODE_ITEM_NAME_TOO_LONG = 'item_name_too_long';
 export const ERROR_CODE_ITEM_NAME_IN_USE = 'item_name_in_use';
 export const ERROR_CODE_UPLOAD_FILE_LIMIT = 'upload_file_limit';
+export const ERROR_CODE_CHILD_FOLDER_FAILED_UPLOAD = 'child_folder_failed_upload';
```
priyajeet: ERROR_CODE_UPLOAD_CHILD_FOLDER_FAILED

src/api/uploads/FolderUploadNode.js
```diff
@@ -231,6 +265,14 @@ class FolderUploadNode {
             this.readEntry(reader, resolve);
         });
     };
+
+    /**
+     * Returns the folderId
+     * @returns {string}
```
booc0mtaco: will `folderId` be `''` or some value, or can this ever end up as `undefined`? if the latter, we'd want a maybe return type or handle that case somehow

src/components/messages.js
```diff
@@ -637,6 +637,16 @@ const messages: { [string]: MessageDescriptor } = defineMessages({
         description: 'Error message shown when pending app folder size exceeds the limit',
         defaultMessage: 'Pending app folder size limit exceeded',
     },
+    uploadsProvidedFolderNameInvalidMessage: {
+        id: 'be.uploadsProvidedFolderNameInvalidMessage',
+        description: 'Error message shown when pending folder upload contains invalid characters',
+        defaultMessage: 'Provided folder name, {name}, could not be used to create a folder.',
```
priyajeet: Wondering if a better/simpler message would be

`Could not create a folder with the name - {name}`

Def also check if the uploader is small width, text doesn't weirdly overflow.

src/components/ContentUploader/progressCellRenderer.js
```diff
@@ -18,12 +18,19 @@ type Props = {
...
  * @returns {FormattedMessage}
  */
-const getErrorMessage = (errorCode: ?string) => {
+const getErrorMessage = (errorCode: ?string, itemName: ?string) => {
     switch (errorCode) {
+        case ERROR_CODE_CHILD_FOLDER_FAILED_UPLOAD:
+            return <FormattedMessage {...messages.uploadsOneOrMoreChildFoldersFailedToUploadMessage} />;
         case 'file_size_limit_exceeded':
             return <FormattedMessage {...messages.uploadsFileSizeLimitExceededErrorMessage} />;
+        case 'item_name_invalid':
```
priyajeet: Can you change these or file a bug to change these to constants too.

src/api/uploads/FolderUpload.js
priyajeet: Yeah change it to that + make it a constant on top for re-usability below

booc0mtaco: would go with the function, as that might have some value checking logic

src/api/uploads/FolderUploadNode.js
```diff
@@ -96,34 +106,58 @@ class FolderUploadNode {
...
             const data = await this.createFolder();
             this.folderId = data.id;
         } catch (error) {
+            errorEncountered = true;
+            errorCode = error.code;
             // @TODO: Handle 429
             if (error.code !== ERROR_CODE_ITEM_NAME_IN_USE) {
-                errorCallback(error);
-                return;
+                if (!isRoot) {
```
priyajeet: Reverse the condition, so TRUTHY on top. Eliminates !

src/api/uploads/FolderUploadNode.js
```diff
@@ -96,34 +106,58 @@ class FolderUploadNode {
...
+            extension: '',
+            name: this.name,
+            status: STATUS_COMPLETE,
+            isFolder: true,
+            size: 1,
+            progress: 100,
+        };
+
+        if (errorEncountered) {
+            folderObject.status = STATUS_ERROR;
```
priyajeet: Does the folder object by default have "error"?
If not, then we should not be mutating the object type for error special casing.
Or one should adjust flowtype for the addFilesToUploadQueue() function globally.

src/api/uploads/FolderUploadNode.js
```diff
@@ -96,34 +106,58 @@ class FolderUploadNode {
...
+            folderObject.error = { code: errorCode };
+        }
+
+        this.addFolderToUploadQueue([folderObject]);
```
priyajeet: I could not find a function with that name that takes array of objects?!?

src/constants.js
patrickpaul: Wow, that's way better.

src/components/messages.js
patrickpaul: This was to preserve consistency, will explore other options.

src/components/ContentUploader/progressCellRenderer.js
patrickpaul: Will move to constants in this PR.

src/api/uploads/FolderUploadNode.js
priyajeet: So it seems there is some wierd function naming going on. This points to ContentUploader.js::addToQueue() which takes array of UploadItem, which has error object

src/api/uploads/FolderUploadNode.js
priyajeet: See comment below.

src/api/uploads/FolderUploadNode.js
```diff
@@ -96,34 +106,58 @@ class FolderUploadNode {
...
-            {
-                extension: '',
-                name: this.name,
-                status: STATUS_COMPLETE,
-                isFolder: true,
-                size: 1,
-                progress: 100,
-            },
-        ]);
+        const folderObject: {
```
priyajeet: const folderObject: UploadItem = {...}

src/api/uploads/FolderUploadNode.js
```diff
@@ -66,7 +71,12 @@ class FolderUploadNode {
         this.parentFolderId = parentFolderId;
 
         await this.createAndUploadFolder(errorCallback, isRoot);
+        if (!this.getFolderId()) {
+            // Folder was not successfully created. Do not attempt to upload its contents.
+            return;
```
priyajeet: Since this is just a small function, I think we can just do below instead of a guard check

```js
// Only upload contents if container exists
if (this.getFolderId()) {
    this.addFilesToUploadQueue(this.getFormattedFiles(), noop, true);
    await this.uploadChildFolders(errorCallback);
}
        
```

src/api/uploads/FolderUpload.js
priyajeet: Actually thinking more about this, I am not sure if we should be relying on folder id being present or not to detect an error. It feels like assumption logic, which means that somewhere else in the code you have to make sure that folder id is not set for this condition to work

priyajeet: I would think about adding try/catch in this function too, and throw a new error. In that way it will be explicit and just scoped to error handling.

priyajeet: Something like

```js
    /**
     * Upload a folder
     *
     * @public
     * @param {string} parentFolderId
     * @param {Function} errorCallback
     * @param {boolean} isRoot
     * @returns {Promise}
     */
    async upload(parentFolderId: string, errorCallback: Function, isRoot: boolean = false) {
        this.parentFolderId = parentFolderId;

		try {
        	await this.createAndUploadFolder(errorCallback, isRoot);
        } catch (error) {
        	return;
        }
        
        this.addFilesToUploadQueue(this.getFormattedFiles(), noop, true);
        await this.uploadChildFolders(errorCallback);
    }
    
    /**
     * Create folder and add it to the upload queue
     *
     * @private
     * @param {Function} errorCallback
     * @param {boolean} isRoot
     * @returns {Promise}
     */
    createAndUploadFolder = async (errorCallback: Function, isRoot: boolean) => {
        await this.buildCurrentFolderFromEntry();

        let errorObj = {};

        try {
            const data = await this.createFolder();
            this.folderId = data.id;
        } catch (error) {
            if (error.code === ERROR_CODE_ITEM_NAME_IN_USE) {
                this.folderId = error.context_info.conflicts[0].id;
            } else {
            	if (isRoot) {
            		errorCallback(error);
                    throw error; // non-recoverable error, abort
                } else {
                	errorObj = { status: STATUS_ERROR, error };
                    errorCallback({ code: ERROR_CODE_CHILD_FOLDER_FAILED_UPLOAD });
                }
            }
        }
        
        // The root folder has already been added to the upload queue in ContentUploader
        if (isRoot) {
            return;
        }

        const folderObject: UploadItem = {
            extension: '',
            name: this.name,
            status: STATUS_COMPLETE,
            isFolder: true,
            size: 1,
            progress: 100,
            ...errorObj
        };

        this.addFolderToUploadQueue([folderObject]);
    };
```

src/api/uploads/FolderUploadNode.js
patrickpaul: Resolved, no longer calls with array.

src/constants.js
```diff
@@ -179,6 +179,10 @@ export const ERROR_CODE_ITEM_NAME_INVALID = 'item_name_invalid';
 export const ERROR_CODE_ITEM_NAME_TOO_LONG = 'item_name_too_long';
 export const ERROR_CODE_ITEM_NAME_IN_USE = 'item_name_in_use';
 export const ERROR_CODE_UPLOAD_FILE_LIMIT = 'upload_file_limit';
+export const ERROR_CODE_UPLOAD_CHILD_FOLDER_FAILED = 'child_folder_failed_upload';
+export const ERROR_CODE_STORAGE_LIMIT_EXCEEDED = 'storage_limit_exceeded';
```
priyajeet: The new ones should all be ERROR_CODE_UPLOAD_*
So that it's a bit more descriptive where the error is likely being used and also helps in search.

src/api/uploads/FolderUploadNode.js
```diff
@@ -231,6 +263,14 @@ class FolderUploadNode {
             this.readEntry(reader, resolve);
         });
     };
+
+    /**
+     * Returns the folderId
+     * @returns {string}
+     */
+    getFolderId = (): string => {
```
priyajeet: `?string` ?

src/api/uploads/FolderUploadNode.js
patrickpaul: This isn't yet an `UploadItem`. This object is passed up and created as an `UploadItem` in a higher level (we don't have access to the required API property at this lower level).

src/api/uploads/FolderUploadNode.js
patrickpaul: Done.

src/api/uploads/FolderUploadNode.js
patrickpaul: The empty state is an empty string -> it is returned from the monolith that way.

src/api/uploads/FolderUploadNode.js
patrickpaul: It returns an empty string if not set, does that still need a maybe type?

src/constants.js
patrickpaul: Will fix.

src/api/uploads/FolderUploadNode.js
```diff
@@ -96,34 +105,56 @@ class FolderUploadNode {
...
-            {
-                extension: '',
-                name: this.name,
-                status: STATUS_COMPLETE,
-                isFolder: true,
-                size: 1,
-                progress: 100,
-            },
-        ]);
+        const folderObject: {
```
DanDeMicco: I think this can be refactored instead of creating a whole new type. It appears to just be missing the `api` field

flow-typed/box-ui-elements.js
```diff
@@ -370,6 +370,18 @@ type Collection = {
     totalCount?: number,
 };
 
+type FolderUploadItem = {
```
priyajeet: Unless you have time don't have to do in this commit but in your followup:

Since FolderUploadItem and UploadItem are essentially the same, other than a couple poorly labelled and optional.

`UploadFile` ---> `WebkitFileObject`

```js
type UploadItem = {
    api?: PlainUploadAPI | MultiputUploadAPI, // made optional
    item?: BoxItem, // renamed
    error?: Object,
    extension: string,
    fileObject: WebkitFileObject, // renamed, OR nativeFileObject
    name: string,
    progress: number,
    size: number,
    status: UploadStatus,
    options?: UploadItemAPIOptions,
    isFolder?: boolean,
};
```

patrickpaul: Will merge this and create a new PR. Thanks!

PR #675, opened 2018-12-03
Fix: specific styling for metadata sidebar

ChenCodes:
screen shot 2018-12-03 at 12 03 58 pm


boxcla:

priyajeet:
Should not be touching too much of a central layout. A better/more correct fix would be to remove padding-top: 15px; from .bcs-scroll-content-wrapper and then going to each sidebar and adding that padding as needed.

ChenCodes:
Closing this PR as the best fix for now is to remove position: sticky from BRUI

PR #677, opened 2018-12-03
fix(release): missing scss files

priyajeet:
Also reverts to using the `npm` env in babel config so that import statements for scss files are stripped out in the babel compile step


boxcla:

PR #680, opened 2018-12-04
fix(sidebar): remove global padding from sidebar content

priyajeet:
and let each sidebar dictate their own inner layout and padding


boxcla:

PR #681, opened 2018-12-04
fix(sidebar): only show classification error when correctable

DanDeMicco:


boxcla:

PR #684, opened 2018-12-05
fix(sidebar): pass correct query params to collaborators

DanDeMicco:


boxcla:

src/components/ContentSidebar/ActivitySidebar.js
```diff
@@ -387,9 +387,7 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         }
 
         api.getFileCollaboratorsAPI(true).getFileCollaborators(file.id, successCallback, errorCallback, {
-            params: {
```
jstoffan: Should we instead change the `markerGetRequest` to unwrap `params` from `requestData`?

PR #689, opened 2018-12-06
fix(sidebar): Do not show hidden template fields

jeremypress:
Will fix tests after feedback


boxcla:

PR #695, opened 2018-12-10
fix(sidebar): allow approval date to be reset

cweeii:
The date can come in as null, and thus the date never gets reset from the previously set date.


boxcla:

PR #696, opened 2018-12-11
fix(webapp): add event for UEW

visylvius:
Adds an event for the Upload Embed Widget in which the data that is dragged into the modal is then sent to the t3 page.


boxcla:

src/components/ContentUploader/ContentUploader.js
```diff
@@ -312,6 +313,10 @@ class ContentUploader extends Component<Props, State> {
             return;
         }
 
+        if (onStage) {
```
jeremypress: Is this the only place where files can be "staged"?

src/components/ContentUploader/ContentUploader.js
```diff
@@ -66,6 +66,7 @@ type Props = {
     onComplete: Function,
     onError: Function,
     onMinimize?: Function,
+    onStage?: Function,
```
jeremypress: Can you add the prop to our docs? https://github.com/box/box-ui-elements#props

PR #698, opened 2018-12-11
fix(uploader): add new event for upload widget

visylvius:
UEW -> Upload Embed Widget for Webapp.
onStage is the new event that gets introduced for the purposes of the UEW.

The general idea is that the way the new v2 Upload Widget wasn't properly set up for handling drag and dropping files into the UEW.
If a user did drag and drop, it was querying the ItemList React Virtualized Column to render the items name.
However for the v2 UEW it was requested that the item size would also be included in the rendered item info.
And the item list didn't have a column for item size, and it would be best if we passed a file object back to the UEW. (Props to jpress for the suggestion!)

Therefore this onStage event was introduced, The idea being that if a user drags or drops an item into the Upload embed widget it then emits an event that can be intercepted by the UEW and then used to render the proper information.

Since you can't drag folders into the UEW, I only added it to the item getNewDataForTransferItems, giving access to the file object

### DIFF
--- a/src/components/ContentUploader/ContentUploader.js
+++ b/src/components/ContentUploader/ContentUploader.js
@@ -66,7 +66,7 @@ type Props = {
     onComplete: Function,
     onError: Function,
     onMinimize?: Function,
-    onUpload: Function,
+    onBeforeUpload: (file: Array<UploadFileWithAPIOptions | File>) => void,
     onUpload: Function,
     overwrite: boolean,
     requestInterceptor?: Function,
@@ -120,6 +120,7 @@ class ContentUploader extends Component<Props, State> {
         clientName: CLIENT_NAME_CONTENT_UPLOADER,
         fileLimit: FILE_LIMIT_DEFAULT,
         uploadHost: DEFAULT_HOSTNAME_UPLOAD,
+        onBeforeUpload: noop,
         onClose: noop,
         onComplete: noop,
         onError: noop,
@@ -286,7 +287,7 @@ class ContentUploader extends Component<Props, State> {
         itemUpdateCallback: Function,
         isRelativePathIgnored?: boolean = false,
     ) => {
-        const { rootFolderId } = this.props;
+        const { onBeforeUpload, rootFolderId } = this.props;
         if (!files || files.length === 0) {
             return;
         }
@@ -311,6 +312,8 @@ class ContentUploader extends Component<Props, State> {
             this.addFilesWithRelativePathToQueue(newFiles, itemUpdateCallback);
             return;
         }
+
+        onBeforeUpload(newFiles);
 
         this.addFilesWithoutRelativePathToQueue(newFiles, itemUpdateCallback);
     };
--- a/src/wrappers/ContentUploader.js
+++ b/src/wrappers/ContentUploader.js
@@ -41,6 +41,16 @@ class ContentUploader extends ES6Wrapper {
     };
 
     /**
+     * Callback on a single pre-uploaded file. Emits 'beforeupload' event with the Box File object pre-upload.
+     *
+     * @param {Object} data - Upload item
+     * @return {void}
+     */
+    onBeforeUpload = (data: UploadFileWithAPIOptions | File): void => {
+        this.emit('beforeupload', data);
+    };
+
+    /**
      * Callback on a single successful upload. Emits 'uploadsuccess' event with Box File object of uploaded item.
      *
      * @param {BoxItem} data - Successfully uploaded item
@@ -65,6 +75,7 @@ class ContentUploader extends ES6Wrapper {
                 onClose={this.onClose}
                 onComplete={this.onComplete}
                 onError={this.onError}
+                onBeforeUpload={this.onBeforeUpload}
                 onUpload={this.onUpload}
                 modal={((modal: any): ModalOptions)}
                 {...rest}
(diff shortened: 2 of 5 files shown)


boxcla:

priyajeet:
Can you explain what this is and what it is doing, whats UEW?

visylvius:
Of course. Was in a bit of a rush yesterday, so didn't properly document. 

UEW -> Upload Embed Widget for Webapp. 
onStage is the new event that gets introduced for the purposes of the UEW. 

The general idea is that the way the new v2 Upload Widget wasn't properly set up for handling drag and dropping files into the UEW. 
If a user did drag and drop, it was querying the ItemList React Virtualized Column to render the items name. 
However for the v2 UEW it was requested that the item size would also be included in the rendered item info. 
And the item list didn't have a column for item size, and it would be best if we passed a file object back to the UEW. (Props to jpress for the suggestion!)

Therefore this onStage event was introduced, The idea being that if a user drags or drops an item into the Upload embed widget it then emits an event that can be intercepted by the UEW and then used to render the proper information. 

Since you can't drag folders into the UEW, I only added it to the item getNewDataForTransferItems, giving access to the file object. I default to your wisdom on if this a proper design pattern.

src/components/ContentUploader/ContentUploader.js
```diff
@@ -66,6 +66,7 @@ type Props = {
     onComplete: Function,
     onError: Function,
     onMinimize?: Function,
+    onStage: Function,
     onUpload: Function,
     onUpload: Function,
```
priyajeet: Can you remove one of these, looks like there are two

src/components/ContentUploader/ContentUploader.js
```diff
@@ -66,6 +66,7 @@ type Props = {
     onComplete: Function,
     onError: Function,
     onMinimize?: Function,
+    onStage: Function,
```
priyajeet: Does stage mean staging? If so, wondering if a better name would be onBeforeUpload()?

priyajeet: Also instead of Function, can you add the flowtype signature (args:type)=> etc. Starting to add that now instead of just Function.

src/components/ContentUploader/ContentUploader.js
```diff
@@ -312,6 +313,10 @@ class ContentUploader extends Component<Props, State> {
             return;
         }
 
+        if (onStage) {
```
priyajeet: For consistency, add in defaultProps like the others, and just call it.

src/wrappers/ContentUploader.js
```diff
@@ -40,6 +40,16 @@ class ContentUploader extends ES6Wrapper {
         this.emit('error', data);
     };
 
+    /**
+     * Callback when a file uploads that gets called in the t3 page for Upload Embed Widget
+     *
+     * @param {Object} data - Upload item
+     * @return {void}
+     */
+    onStage = (data: BoxItem): void => {
```
priyajeet: You didnt pass it to the component below

src/components/ContentUploader/ContentUploader.js
priyajeet: Also add a unit test for this function?

priyajeet:
> Of course. Was in a bit of a rush yesterday, so didn't properly document.
> 
> UEW -> Upload Embed Widget for Webapp.
> onStage is the new event that gets introduced for the purposes of the UEW.
> 
> The general idea is that the way the new v2 Upload Widget wasn't properly set up for handling drag and dropping files into the UEW.
> If a user did drag and drop, it was querying the ItemList React Virtualized Column to render the items name.
> However for the v2 UEW it was requested that the item size would also be included in the rendered item info.
> And the item list didn't have a column for item size, and it would be best if we passed a file object back to the UEW. (Props to jpress for the suggestion!)
> 
> Therefore this onStage event was introduced, The idea being that if a user drags or drops an item into the Upload embed widget it then emits an event that can be intercepted by the UEW and then used to render the proper information.
> 
> Since you can't drag folders into the UEW, I only added it to the item getNewDataForTransferItems, giving access to the file object. I default to your wisdom on if this a proper design pattern.

Can you add details to commit message

src/wrappers/ContentUploader.js
visylvius: Whoops, forgot to add that when I redid the PR

src/components/ContentUploader/ContentUploader.js
visylvius: There is only one. There are two onUploads.

src/wrappers/ContentUploader.js
```diff
@@ -40,6 +40,16 @@ class ContentUploader extends ES6Wrapper {
...
     };
 
+    /**
+     * Callback when a file uploads that gets called in the t3 page for Upload Embed Widget
+     *
+     * @param {Object} data - Upload item
+     * @return {void}
+     */
+    onBeforeUpload = (data: BoxItem): void => {
+        this.emit('stage', data);
```
priyajeet: event `beforeupload`

src/wrappers/ContentUploader.js
```diff
@@ -40,6 +40,16 @@ class ContentUploader extends ES6Wrapper {
         this.emit('error', data);
     };
 
+    /**
+     * Callback when a file uploads that gets called in the t3 page for Upload Embed Widget
```
priyajeet: Change the comment to suggest what the event is doing instead of what its for, since this is public api

visylvius:
Will submit a separate PR for updating the docs.

PR #700, opened 2018-12-11
fix(sidebar): Use correct timestamps for file details

jstoffan:


boxcla:

src/util/__tests__/fields-test.js
```diff
@@ -99,6 +101,8 @@ describe('util/fields', () => {
             FIELD_FILE_VERSION,
             FIELD_SHARED_LINK,
             FIELD_PERMISSIONS,
+            FIELD_CONTENT_CREATED_AT,
+            FIELD_CONTENT_MODIFIED_AT,
             FIELD_CREATED_AT,
```
jstoffan: It looks like neither `FIELD_CREATED_AT` nor `FIELD_MODIFIED_AT` are being used anywhere else in the sidebar. @priyajeet, can you think of any tricky areas I may have missed, maybe in the metadata sidebar?

priyajeet: Not from the top of my head. Explorer / Picker use something for recents but not here. Leave it for now, we can remove them later. Since they are part of the same DB row, perf hit should not be there

DanDeMicco: We are currently using modified_at in Feed.js I believe

jstoffan: Is that coming from the comment and task objects or from the file object @DanDeMicco?

DanDeMicco: It is from the file object. We use it to inject the current version

jstoffan: I'll just leave them for now.

PR #708, opened 2018-12-13
chore(box-edit): Adding box-edit lib

jeremypress:
Still working on migrating tests

### DIFF
--- a/src/api/box-edit/BoxEdit.js
+++ b/src/api/box-edit/BoxEdit.js
@@ -1,0 +1,200 @@
+// @flow
+
+import Browser from './BrowserUtils';
+import ComServerClient from './ComServerClient';
+import CONSTANTS from './constants';
+
+type stringTuple = [string, string];
+type extensionRequestQueueItem = {
+    promise: Promise<string>,
+    reject: Function,
+    resolve: Function,
+};
+
+const TIMEOUT_MS = 5000;
+const EXTENSION_CHECK_DEBOUNCE_TIME = 100;
+
+let extensionRequestTimeout: ?TimeoutID;
+
+function createRequestData(extensions: Array<string>): string {
+    return JSON.stringify({
+        request_type: 'get_default_application',
+        extension: extensions,
+    });
+}
+
+function createExecuteData(fileId, token, authCode, tokenScope): string {
+    const execData = JSON.stringify({
+        auth_code: authCode,
+        auth_token: token,
+        browser_type: Browser.getName(),
+        command_type: 'launch_application',
+        file_id: fileId.toString(),
+        token_scope: tokenScope,
+    });
+    return execData;
+}
+
+function isBlacklistedExtension(extension): boolean {
+    const { EXTENSION_BLACKLIST } = CONSTANTS;
+    let uppercaseExt = extension.toUpperCase();
+
+    // if ext has a leading ., strip it
+    if (uppercaseExt.charAt(0) === '.') {
+        uppercaseExt = uppercaseExt.substr(1);
+    }
+
+    return uppercaseExt in EXTENSION_BLACKLIST;
+}
+
+let BoxEditInstance = null;
+
+class BoxEdit {
+    client: ComServerClient;
+
+    extensionRequestQueue: Map<string, extensionRequestQueueItem>;
+
+    constructor() {
+        if (!(BoxEditInstance instanceof BoxEdit)) {
+            BoxEditInstance = this;
+        }
+        this.extensionRequestQueue = new Map();
+        return BoxEditInstance;
+    }
+
+    queueGetNativeAppNameFromLocal(extension: string): Promise<string> {
+        // There's already a pending or fulfilled request for the appname
+        if (this.extensionRequestQueue.has(extension)) {
+            const queueItem = this.extensionRequestQueue.get(extension);
+            if (!queueItem) {
+                throw new Error('Race condition re: queueGetNativeAppNameFromLocal');
+            }
+            return queueItem.promise;
+        }
+
+        const extensionRequest = {};
+        const appNameRequestPromise = new Promise((resolve, reject) => {
+            extensionRequest.resolve = resolve;
+            extensionRequest.reject = reject;
+        });
+        extensionRequest.promise = appNameRequestPromise;
+        this.extensionRequestQueue.set(extension, extensionRequest);
+
+        return appNameRequestPromise;
+    }
+
+    checkBoxEditAvailability(): Promise<any> {
+        return this.getBoxEditAvailability();
+    }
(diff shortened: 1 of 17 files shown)


src/api/box-edit/BrowserUtils.js
```diff
@@ -0,0 +1,298 @@
+// @flow
+const BROWSER_CONSTANTS = {
```
priyajeet: This file should ideally be inside util folder, and merged with Browser.js

src/api/box-edit/BoxEdit.js
```diff
@@ -0,0 +1,200 @@
...
+            }),
+        );
+
+        const resultMap: Map<string, string> = new Map();
+        extensionToAppTuples.forEach(tuple => resultMap.set(...tuple));
+
+        return Promise.resolve(resultMap);
+    }
+
+    openFile(fileTypedID: string, token: Object): Promise<any> {
```
JustinHoldstock: Flow type for `token`? `tokenDetails` might be a little closer to a correct name, since `token.data.token` sounds strange

src/api/box-edit/BoxEdit.js
```diff
@@ -0,0 +1,200 @@
...
+            token.data.token_scope,
+        );
+
+        return this.client.sendCommand(executeDataAsString, TIMEOUT_MS);
+    }
+
+    getAppForExtension(extension: string): Promise<any> {
+        try {
+            if (isBlacklistedExtension(extension)) {
+                throw new Error('blacklisted');
```
JustinHoldstock: If we want to record this in the future, could you also toss the extension that was blacklisted. ( I know we can check against the list of blacklisted extensions, but it'll allow us to see EXACTLY which ones are throwing this error)

src/api/box-edit/BoxEdit.js
```diff
@@ -0,0 +1,200 @@
...
+
+            if (!extensionRequestTimeout) {
+                extensionRequestTimeout = setTimeout(() => {
+                    this.processExtensionRequestQueue();
+                }, EXTENSION_CHECK_DEBOUNCE_TIME);
+            }
+
+            return applicationSupportRequest;
+        } catch (err) {
+            return Promise.reject();
```
JustinHoldstock: Pass the error to the promise rejection?

src/api/box-edit/BrowserUtils.js
```diff
@@ -0,0 +1,298 @@
...
+    OS: string;
+
+    device: string;
+
+    constructor() {
+        const { navigator } = window;
+        const browserData = getBrowserData(navigator, BROWSER_DATA);
+        const osInfo = getBrowserData(navigator, OS_DATA);
+
+        this.browser = browserData.identity || 'unknown';
```
JustinHoldstock: constant for `unknown`, since you're using it in so many places?

src/api/box-edit/ComServerClient.js
```diff
@@ -0,0 +1,329 @@
...
+            this.channels = this.channels.concat([
+                createHTTPChannel(appName),
+                createSafariChannel(appName),
+                createSynchronousActiveXChannel(appName),
+            ]);
+        }
+    }
+
+    isInitialized() {
+        return this.isInitialized;
```
JustinHoldstock: This will just return the function property `isInitialized()`. Maybe rename this variable to something else

src/api/box-edit/HTTPChannel.js
```diff
@@ -0,0 +1,242 @@
...
+const DEFAULT_PRIMARY_PORT = 17223;
+const DEFAULT_FALLBACK_PORT = 17224;
+const CONTENT_TYPE_HEADER = 'Content-Type';
+const CONTENT_TYPE_VALUE = 'text/plain; charset=UTF-8';
+const STATUS_NOT_RUNNING = 'notrunning';
+
+const localStore = new LocalStore();
+
+function shouldUseFallbackFirst() {
+    return !!localStore.getItem('comUseFallback');
```
JustinHoldstock: `comUseFallback` as a constant

src/api/box-edit/HTTPChannel.js
```diff
@@ -0,0 +1,242 @@
...
+        setCookie(`bgp-${bgpId}`, 'generic', aYearFromNow);
+        return comChannelName;
+    };
+
+    getComServerStatus = (timeoutMS: number) => {
+        if (!this.comServerInstallationPromise || this.comServerInstallationPromiseRejected) {
+            // Null out the promise and fetch status again (handles the case where user installed box edit within the same page load)
+            this.comServerInstallationPromise = null;
+            this.comServerInstallationPromise = this.getComServerStatusInstallationPromise(timeoutMS);
+            return this.comServerInstallationPromise;
```
JustinHoldstock: No need to return inside the `if`, if you're already returning it below :)

src/api/box-edit/SafariChannel.js
```diff
@@ -0,0 +1,123 @@
...
+    appExtensionEventResponseHandler = (responseVal: any) => {
+        const response = typeof responseVal.detail === 'string' ? JSON.parse(responseVal.detail) : responseVal.detail;
+
+        if (this.reqIdToPromiseMap.has(response.req_id)) {
+            const resolveObj = this.reqIdToPromiseMap.get(response.req_id);
+            if (resolveObj) {
+                clearTimeout(resolveObj.rejectTimeout);
+                this.reqIdToPromiseMap.delete(response.req_id);
+
+                const responseData =
```
JustinHoldstock: Could used above mentioned util method, here

src/api/box-edit/cookies.js
```diff
@@ -0,0 +1,91 @@
...
+ * @param {string} domain - (Optional) Cookie domain. Defaults to the cookie_domain used server-side.
+ * @returns {void}
+ */
+function set(name: string, value: any, expires: number, path: string = '/', domain: string = '') {
+    const { document } = window;
+    const expiresDate = new Date();
+    const cookie = [];
+
+    if (typeof expires === 'undefined') {
+        expires = expiresDate.getTime() + 1000 * 3600 * 24 * 60; // two months;
```
JustinHoldstock: Could create a `TWO_MONTHS` constant to make this a little more readable

src/api/box-edit/BrowserUtils.js
jeremypress: agreed, will do.

jeremypress: I don't know if we want to bring this model into our browser utils. The box edit version passes an instance around and keeps track of state while we rely on static methods. We can refactor their utils but at that point we would own it.

PR #709, opened 2018-12-13
feat: feature flag component

alexkrolick:


boxcla:

src/util/features.js
```diff
@@ -0,0 +1,53 @@
+// @flow
```
priyajeet: For starters I would re-organize this file, as its dealing with both components and utility functions. You can create a features folder inside

src/components/feature-checking
|-- util.js // holds the isFeatureEnabled, getFeatureConfig
|-- FeatureProvider // exports <FeatureContext.Provider>
|-- FeatureFlag // renders <FeatureContext.Consumer>
|-- FeatureContext // exports React.createContext({});
|-- flowTypes.js // exports all needed types
|-- index.js // exports all the above

src/util/features.js
```diff
@@ -0,0 +1,53 @@
...
+
+function FeatureProvider({ features = {}, children }: { features?: FeatureConfig, children: ?React.Node }) {
+    return <FeatureContext.Provider value={features}>{children}</FeatureContext.Provider>;
+}
+
+const FeatureConsumer = FeatureContext.Consumer;
+
+function FeatureFlag({
+    feature,
+    enabled = () => null,
```
priyajeet: lodash/noop?

src/util/features.js
```diff
@@ -0,0 +1,53 @@
...
+    enabled?: FeatureOptions => React.Node,
+    disabled?: FeatureOptions => React.Node,
+    children?: (Boolean, FeatureOptions) => React.Node,
+}) {
+    return (
+        <FeatureContext.Consumer>
+            {features => {
+                const isEnabled = isFeatureEnabled(features, feature);
+                const featureConfig = getFeatureConfig(features, feature);
+                if (children) return children(isEnabled, featureConfig);
```
priyajeet: children should not be a function, should be a simple React.Node

src/util/features.js
alexkrolick: It should return null for type safety, not undefined.

src/util/features.js
alexkrolick: That removes the ability to provide a fallback element, but enabled/disabled could be used instead. OK.

alexkrolick: Also, currently the feature format allows additional parameters in the feature object. These are passed to the function as arguments. Should this be removed?

```js
const features = {
  aFeature: {
    enabled: true,
    fetchUrl: '/app-api/abc',
    isNameRequired: false,
  }
}
```

priyajeet: Initially I was thinking roughly along the lines of this...

```jsx
// Somewhere in EUA
const features = {
    activityfeed: {
        hasNewUI: FF boolean from monolith,
    },
};

// withFeatureContext.js
const withFeatureContext = (WrappedComponent: React.ComponentType<any>, feature) => (props: any) => (
    <FeatureContext.Consumer>
        {features => <WrappedComponent {...props} {...features[feature]} />}
    </FeatureContext.Consumer>
);

// ActivityFeed.js
class ActivityFeed {
    render() {
        this.props.hasNewUI ? <NewUI> : <OldUI>
    }
}

export withFeatureContext(ActivityFeed, 'activityfeed')
```

src/components/FeatureChecking/FeatureFlag.js
```diff
@@ -0,0 +1,30 @@
...
+    disabled?: () => React.Node,
+    children?: React.Node,
+}) {
+    return (
+        <FeatureConsumer>
+            {features => {
+                const isEnabled = isFeatureEnabled(features, feature);
+                const featureConfig = getFeatureConfig(features, feature);
+                if (children) return isEnabled && children;
+                return isEnabled ? enabled(featureConfig) : disabled();
```
jstoffan: What is the purpose of the `enabled` and `disabled` callbacks?

src/components/FeatureChecking/util.js
```diff
@@ -0,0 +1,20 @@
+// @flow
+import get from 'lodash/get';
+
+export type FeatureOptions = {
```
jstoffan: Duplicate type declarations?

src/components/FeatureChecking/FeatureFlag.js
alexkrolick: allows you to inline logic that depends on the variables in the config object without erroring out

jstoffan: It also seems to allow arbitrary components to be rendered as the return value of this function, which I believe we wanted to avoid.

alexkrolick: No, that's the intended behavior:

```js
function BigThing() {
  return(
    <div>
      <h2>Lots going on in here</h2>
      <FeatureFlag feature="feat-a" enabled={() => <Button>Do Something</Button>} />
      <p>Some other stuff...</p>
      <FeatureFlag feature="feat-b" enabled={() =>
        <div>Other Feature Here</div>
      } />
    </div>
  )
}
```

jstoffan: I see. That should be fine as long as the functions passed to `enabled` and `disabled` live inside box-ui-elements and aren't supplied as props by the consuming app.

alexkrolick: Right, they are props of the internal FeatureFlag component itself

src/components/FeatureChecking/util.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+import get from 'lodash/get';
+import * as types from './flowTypes';
+
+function isFeatureEnabled(features: types.FeatureConfig, featureName: string) {
+    return !!get(features, `${featureName}`, false);
```
jstoffan: Is the template string necessary here? It looks like `featureName` is required to be a string as-is.

alexkrolick: Good point, that's residual from this format: `${featureName}.enabled` (requiring an object)

src/components/FeatureChecking/flowTypes.js
```diff
@@ -0,0 +1,8 @@
+// @flow
+export type FeatureOptions = {
```
priyajeet: FeatureOptions = Object?
Since thats an object

alexkrolick: "Object" is disallowed in strict mode, and is broader than this type (lots of things are objects).

alexkrolick:
I can't click the merge button, so please merge when the build is green

screen shot 2018-12-13 at 4 19 47 pm

Thanks for the reviews 👍

PR #710, opened 2018-12-14
fix(uploader): fallback for content_modified_at in ie11 and edge

DanDeMicco:


boxcla:

PR #716, opened 2018-12-14
fix(uploader): refactor anti pattern in addFilesToUploadQueue

visylvius:
What this PR Does: 

We shouldn't ideally set state by mutating it directly. One should typically call setState, in order to do state changes. 

this PR simply refactors the anti pattern and uses setState to not cause unnecessary side effects.


boxcla:

jeremypress:
looks good, lets merge after EUA rollout

PR #717, opened 2018-12-16
fix(openwith): disable integrations in the menu dropdown

jeremypress:


boxcla:

src/components/ContentOpenWith/OpenWithDropdownMenuItem.js
```diff
@@ -15,14 +15,16 @@ type Props = {
 };
 
 const OpenWithDropdownMenuItem = ({ integration, onClick }: Props) => {
-    const { displayName, displayDescription } = integration;
+    const { displayName, displayDescription, isDisabled, disabledReasons = [] } = integration;
     const Icon = ICON_FILE_MAP[displayName] || IconFileDefault;
+    // Use the first disabled reason as the description if the integration is disabled.
+    const description = (isDisabled && disabledReasons[0]) || displayDescription;
```
pramodsum: I think you checked this but the `&&` returns the second option as expected here on all browsers right?

jeremypress: It does, but I'm switching up the logic a bit to use a default message instead of the description so I removed it.

src/components/ContentOpenWith/OpenWithDropdownMenuItem.js
```diff
@@ -7,22 +7,28 @@
...
     onClick: Function,
 };
 
 const OpenWithDropdownMenuItem = ({ integration, onClick }: Props) => {
-    const { displayName, displayDescription } = integration;
+    const { displayName, displayDescription, isDisabled, disabledReasons } = integration;
     const Icon = ICON_FILE_MAP[displayName] || IconFileDefault;
+    // Use the first disabled reason as the description if the integration is disabled.
+    const description = isDisabled
+        ? disabledReasons[0] || <FormattedMessage {...messages.errorOpenWithDescription} />
```
jstoffan: Consider breaking this out to its own variable to help readability.

PR #719, opened 2018-12-18
feat(openwith): Add support for the box-edit integration

jeremypress:


boxcla:

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -194,7 +203,57 @@ class ContentOpenWith extends PureComponent<Props, State> {
...
+        });
+
+        if (boxEditIntegration && !boxEditIntegration.isDisabled) {
+            this.boxEdit = new BoxEdit();
+            // Check if Box Tools is installed and reachable
+            this.boxEdit
+                .checkBoxEditAvailability()
+                .catch(() => {
+                    boxEditIntegration.isDisabled = true;
+                    boxEditIntegration.disabledReasons = ['Install Box Tools to open this file on your desktop'];
```
jeremypress: Will become a localized string

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -194,7 +203,57 @@ class ContentOpenWith extends PureComponent<Props, State> {
...
+                        return;
+                    }
+
+                    this.boxEdit
+                        // Ping box edit server to see if we can open this file
+                        .canOpenWithBoxEdit([extension])
+                        .catch(() => {
+                            // Box Tools was unable to open this file type
+                            boxEditIntegration.isDisabled = true;
+                            boxEditIntegration.disabledReasons = ['Box Tools cannot open this file type'];
```
jeremypress: Will also become a localized string

src/components/ContentOpenWith/OpenWithButton.js
```diff
@@ -29,32 +29,53 @@ type Props = {
...
+    disabledReasons: Array<string> = [],
+    displayDescription: string,
+): ?(string | Element) => {
     if (isLoading) {
         return null;
     }
+
     if (error) {
-        return <FormattedMessage {...messages.errorOpenWithDescription} />;
+        return disabledReasons[0] || <FormattedMessage {...messages.errorOpenWithDescription} />;
```
jstoffan: Lots of duplication here. Would this work?

```
if (isLoading) {
    return null;
}

if (disabledReasons[0]) {
    return disabledReasons[0];
}

//etc.
```

src/components/ContentOpenWith/OpenWithButton.js
```diff
@@ -29,32 +29,53 @@ type Props = {
...
 const OpenWithButton = ({ error, onClick, displayIntegration, isLoading }: Props) => {
-    const { displayDescription, displayName, isDisabled: isDisplayIntegrationDisabled } = displayIntegration || {};
+    const { displayName, isDisabled: isDisplayIntegrationDisabled, extension, disabledReasons, displayDescription } =
+        displayIntegration || {};
 
     const isDisabled = !!isDisplayIntegrationDisabled || !displayName;
-    const IntegrationIcon = displayName && ICON_FILE_MAP[displayName];
-    const Icon = IntegrationIcon || IconOpenWith;
+
+    let Icon;
```
jstoffan: What's the reason for this change?

src/components/ContentOpenWith/OpenWithButton.js
```diff
@@ -29,32 +29,53 @@ type Props = {
...
+        Icon = ICON_FILE_MAP[displayName];
+    } else {
+        Icon = IconOpenWith;
+    }
 
     return (
-        <Tooltip text={getTooltip(displayDescription, error, isLoading)} position="bottom-center">
-            <Button isDisabled={isDisabled} onClick={() => onClick(displayIntegration)}>
+        <Tooltip
+            text={getTooltip(isLoading, error, isDisplayIntegrationDisabled, disabledReasons, displayDescription)}
```
jstoffan: Consider alphabetizing or grouping params together, like `isLoading` with `isDisabled`.

src/components/ContentOpenWith/OpenWithButton.js
```diff
@@ -29,32 +29,53 @@ type Props = {
...
+    if (isDisabled) {
+        return disabledReasons[0] || <FormattedMessage {...messages.emptyOpenWithDescription} />;
     }
+
     if (displayDescription) {
         return displayDescription;
     }
 
-    return <FormattedMessage {...messages.emptyOpenWithDescription} />;
+    return disabledReasons[0] || <FormattedMessage {...messages.emptyOpenWithDescription} />;
```
priyajeet: I would reduce the number of returns here specially considering some are returning something similar. The lesser the number of exit points from your function, less break points you/someone will have to put to debug what the final value ends up being. You could just figure out what message to show and at the end 

if (error) message = messages.errorOpenWithDescription;
else if (isDisabled) message = messages.emptyOpenWithDescription;
else message = messages.emptyOpenWithDescription

`return disabledReasons[0] || <FormattedMessage {...message} />`

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -194,7 +203,57 @@ class ContentOpenWith extends PureComponent<Props, State> {
...
+            this.boxEdit = new BoxEdit();
+            // Check if Box Tools is installed and reachable
+            this.boxEdit
+                .checkBoxEditAvailability()
+                .catch(() => {
+                    boxEditIntegration.isDisabled = true;
+                    boxEditIntegration.disabledReasons = ['Install Box Tools to open this file on your desktop'];
+                })
+                .then(this.getFileExtension)
+                .then(extension => {
```
priyajeet: Can this be migrated to another top level function, so that >>> arrowness reduces?

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -194,7 +203,57 @@ class ContentOpenWith extends PureComponent<Props, State> {
...
+                    // Add the extension to render the correct icon.
+                    boxEditIntegration.extension = extension;
+
+                    if (boxEditIntegration.isDisabled) {
```
priyajeet: Should this check be made into a function, because wouldn't you want it in this.getFileExtension also? To prevent a possible XHR?

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -298,6 +364,30 @@ class ContentOpenWith extends PureComponent<Props, State> {
...
+     * @param {ExecuteAPI} executeData - API response on how to open an executed integration
+
+     * @return {void}
+     */
+    executeBoxEditSuccessHandler = (executeData: ExecuteAPI): void => {
+        const { fileId, token, onExecute } = this.props;
+        const { url } = executeData;
+        const authCode = url.split(AUTH_CODE_DELIMITER)[1];
+
+        this.boxEdit.openFile(`f_${fileId}`, {
```
priyajeet: Since box edit is considered an API, shouldn't this be inside the api layer?

src/components/ContentOpenWith/IconFileMap.js
```diff
@@ -3,13 +3,15 @@ import IconFileDefault from 'box-react-ui/lib/icons/file/IconFileDefault';
...
 import IconGoogleSlides from 'box-react-ui/lib/icons/google-docs/IconGoogleSlides';
+import FileIcon from 'box-react-ui/lib/icons/file-icon/FileIcon';
 
 const ICON_FILE_MAP = {
     default: IconFileDefault,
     'Adobe Sign': IconAdobeSign,
     'Google Docs': IconGoogleDocs,
     'Google Sheets': IconGoogleSheets,
     'Google Slides': IconGoogleSlides,
+    Open: FileIcon,
```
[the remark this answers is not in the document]
jeremypress: 😢 yeah default doesn't take an extension, this one does.

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -194,7 +203,57 @@ class ContentOpenWith extends PureComponent<Props, State> {
      * @return {void}
      */
     fetchOpenWithSuccessHandler = (integrations: Array<Integration>): void => {
-        this.setState({ integrations, isLoading: false });
+        const boxEditIntegration = integrations.find(integration => {
+            return this.isBoxEditIntegration(integration);
+        });
+
+        if (boxEditIntegration && !boxEditIntegration.isDisabled) {
+            this.boxEdit = new BoxEdit();
```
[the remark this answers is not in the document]
jeremypress: Box Edit is already preserving an instance if one is previously created.

```   
let BoxEditInstance
 constructor() {
        if (!(BoxEditInstance instanceof BoxEdit)) {
            BoxEditInstance = this;
        }
        this.extensionRequestQueue = new Map();
        return BoxEditInstance;
    }```

File: src/components/ContentOpenWith/ContentOpenWith.js [hunk 12]
```diff
@@ -224,8 +283,17 @@ class ContentOpenWith extends PureComponent<Props, State> {
...
      * @return {void}
      */
-    onIntegrationClick = ({ appIntegrationId, displayName }: Integration): void => {
+    onIntegrationClick = (integration: Integration): void => {
+        const { appIntegrationId, displayName } = integration;
         const { fileId }: Props = this.props;
+
+        this.executeId = appIntegrationId;
+        if (this.isBoxEditIntegration(integration)) {
+            this.api
```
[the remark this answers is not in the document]
jeremypress: I'm using a different successhandler. I could try to combine but we do very different things if it's box edit vs. an online integration.

src/components/ContentOpenWith/ContentOpenWith.js
jeremypress: do you mean adding it to the factory? I don't want to refactor the library itself yet since we don't own it.

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -175,6 +180,10 @@ class ContentOpenWith extends PureComponent<Props, State> {
         }
     }
 
+    isBoxEditIntegration(integrationId: ?string): boolean {
```
jeremypress: add doc strings

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -194,7 +209,73 @@ class ContentOpenWith extends PureComponent<Props, State> {
...
+    };
+
+    /**
+     * Uses Box Edit to check if Box Tools can open a given file type
+     *
+     * @param {Integration} boxEditIntegration - A Box Edit or Box Edit SFC integration
+     * @return {void}
+     */
+    canOpenExtensionWithBoxEdit = (integration: Integration): Promise<Integration> => {
+        const { extension } = integration;
```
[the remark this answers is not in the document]
jeremypress: I need the whole integration to do the pass through below.

src/components/ContentOpenWith/IconFileMap.js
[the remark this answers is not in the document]
jeremypress: done.

src/components/ContentOpenWith/OpenWithButton.js
```diff
@@ -29,32 +29,41 @@ type Props = {
...
-        return <FormattedMessage {...messages.errorOpenWithDescription} />;
-    }
-    if (displayDescription) {
-        return displayDescription;
+
+    let message = <FormattedMessage {...messages.emptyOpenWithDescription} />;
+    if (disabledReasons.length > 0) {
+        [message] = disabledReasons;
+    } else if (error) {
+        message = <FormattedMessage {...messages.errorOpenWithDescription} />;
```
JustinHoldstock: This seems like an unnecessary statement, considering it's identical, above

jeremypress: What's it identical to? Here we're using the error message but we default to the empty message on L42.

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -194,7 +209,82 @@ class ContentOpenWith extends PureComponent<Props, State> {
...
+                this.setState({ integrations, isLoading: false });
+            });
+    };
+
+    /**
+     * Fetches the file extension of the current file.
+     *
+     * @return {void}
+     */
+    setIntegrationFileExtension = (): Promise<BoxItem> => {
```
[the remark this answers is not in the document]
jeremypress: yep, it was previously modifying the object but now it's not!

src/api/File.js
```diff
@@ -189,6 +190,24 @@ class File extends Item {
...
+
+    /**
+     * Gets the extension of a box file.
+     *
+     * @param {string} id - File id
+     * @param {Function} successCallback - Function to call with results
+     * @param {Function} errorCallback - Function to call with errors
+     * @return {Promise}
+     */
+    async getFileExtension(id: string, successCallback: Function, errorCallback: ElementsErrorCallback): Promise<void> {
```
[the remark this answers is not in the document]
jeremypress: It resolves the `return;` case not ideal, we have a mix of promise based/callback based logic in the API layer.

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -194,7 +209,82 @@ class ContentOpenWith extends PureComponent<Props, State> {
...
+            return;
+        }
+
+        // If Box Edit is present and enabled, we need to set its ability to locally open the given file
+        this.setIntegrationFileExtension()
+            .then(data => {
+                boxEditIntegration.extension = data.extension;
+                return boxEditIntegration;
+            })
+            .then(this.checkBoxEditAvailability)
```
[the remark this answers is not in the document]
jeremypress: giving it a shot. Also finally isn't supported on IE so this could clean things up considerably.

src/api/APIFactory.js
```diff
@@ -510,6 +516,21 @@ class APIFactory {
...
     }
+
+    /**
+     * API for Box Edit
+     *
+     * @param {boolean} shouldDestroy - true if the factory should destroy before returning the call
+     * @return {BoxEditAPI} BoxEditAPI instance
+     */
+    getBoxEditAPI(shouldDestroy: boolean): BoxEditAPI {
+        if (shouldDestroy) {
```
jstoffan: Seems a little odd to pass this in but not destroy the BoxEditAPI itself. Maybe just omit the param entirely?

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -187,16 +200,90 @@ class ContentOpenWith extends PureComponent<Props, State> {
...
+        }
+
+        try {
+            const { extension } = await this.getIntegrationFileExtension();
+            boxEditIntegration.extension = extension;
+        } catch (error) {
+            boxEditIntegration.disabledReasons.push(
+                <FormattedMessage {...messages.executeIntegrationOpenWithErrorHeader} />,
+            );
+            boxEditIntegration.isDisabled = true;
```
jstoffan: Seems like these two lines are duplicated three times, which isn't ideal.

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -187,16 +200,90 @@ class ContentOpenWith extends PureComponent<Props, State> {
...
+            this.setState({ integrations, isLoading: false });
+            return;
+        }
+
+        // If Box Edit is present and enabled, we need to set its ability to locally open the given file
+        if (!(await this.checkBoxEditAvailability())) {
```
jstoffan: I personally find this very difficult to read. If not await check?

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -187,16 +200,90 @@ class ContentOpenWith extends PureComponent<Props, State> {
...
+        );
+
+        if (!boxEditIntegration || boxEditIntegration.isDisabled) {
+            this.setState({ integrations, isLoading: false });
+            return;
+        }
+
+        try {
+            const { extension } = await this.getIntegrationFileExtension();
+            boxEditIntegration.extension = extension;
```
jstoffan: It seems like you could create a new object to contain your overrides, then spread it onto the `boxEditIntegration` object at the end of the method, avoiding duplication of the `setState` call.

src/components/ContentOpenWith/ContentOpenWith.js
DanDeMicco: Can this be done in parallel to `getIntegrationFileExtension`?

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -187,16 +200,90 @@ class ContentOpenWith extends PureComponent<Props, State> {
...
+        });
+    };
+
+    /**
+     * Uses Box Edit to check if Box Tools is installed and reachable
+     *
+     * @param {Integration} boxEditIntegration - A Box Edit or Box Edit SFC integration
+     * @return {void}
+     */
+    checkBoxEditAvailability = (): Promise<boolean> => {
```
jstoffan: The other similar method begins with `can` below. Since the promise here always resolves with a boolean, might be good to stay consistent.

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -187,16 +200,90 @@ class ContentOpenWith extends PureComponent<Props, State> {
             .getOpenWithIntegrations(fileId, language, this.fetchOpenWithSuccessHandler, this.fetchErrorHandler);
     }
 
+    setResolvedIntegrationState = (integrations: Array<Integration>) =>
```
ConradJChan: Is this used anywhere?

src/components/ContentOpenWith/OpenWithButton.js
```diff
@@ -29,32 +29,41 @@ type Props = {
...
-    const IntegrationIcon = displayName && ICON_FILE_MAP[displayName];
-    const Icon = IntegrationIcon || IconOpenWith;
+
+    const Icon = displayName ? getIcon(displayName) : IconOpenWith;
 
     return (
-        <Tooltip text={getTooltip(displayDescription, error, isLoading)} position="bottom-center">
-            <Button isDisabled={isDisabled} onClick={() => onClick(displayIntegration)}>
+        <Tooltip text={getTooltip(displayDescription, isLoading, error, disabledReasons)} position="bottom-center">
+            <Button isDisabled={isDisabled} onClick={() => onClick(displayIntegration, isLoading)}>
```
jstoffan: Does `isLoading` need to be passed back to the click handler?

src/components/ContentOpenWith/ContentOpenWith.js
jeremypress: switching to a similar format. `isBoxEditAvailable`

PR #720, opened 2018-12-19
fix(sidebar): Task assignment to use new status field

ConradJChan:


src/api/Feed.js
```diff
@@ -259,7 +259,7 @@ class Feed extends Base {
         assignmentAPI.updateTaskAssignment({
             file,
             taskAssignmentId,
-            resolutionState,
+            taskStatus,
             message,
```
jstoffan: Is message still used? Seems like it could be removed now, since we were only using it for Accepted/Rejected previously AFAIK.

src/api/TaskAssignments.js
```diff
@@ -92,14 +92,14 @@ class TaskAssignments extends Base {
...
         successCallback,
         errorCallback,
     }: {
         file: BoxItem,
         taskAssignmentId: string,
-        resolutionState: string,
+        taskStatus: string,
         successCallback: Function,
         errorCallback: ElementsErrorCallback,
         message?: string,
```
jstoffan: Can message be removed?

src/api/__tests__/Feed-test.js
```diff
@@ -748,13 +746,12 @@ describe('api/Feed', () => {
                 {
                     id: '1',
                     assigned_to: { id: '1234' },
-                    message: 'completed',
-                    resolution_state: 'completed',
+                    status: 'Completed',
```
jstoffan: Seems like `status` should always be lowercase?

src/util/__tests__/fields-test.js
```diff
@@ -139,7 +139,7 @@ describe('util/fields', () => {
     });
 
     test('should fetch correct task assignment fields', () => {
-        expect(TASK_ASSIGNMENTS_FIELDS_TO_FETCH).toEqual([FIELD_ASSIGNED_TO, FIELD_RESOLUTION_STATE, FIELD_MESSAGE]);
+        expect(TASK_ASSIGNMENTS_FIELDS_TO_FETCH).toEqual([FIELD_ASSIGNED_TO, FIELD_STATUS, FIELD_MESSAGE]);
```
jstoffan: Do we still need `FIELD_MESSAGE` now?

src/api/Feed.js
```diff
@@ -705,13 +702,12 @@ class Feed extends Base {
...
         task.task_assignment_collection.entries = assignments.map(taskAssignment => {
-            const { id, assigned_to, message, status } = taskAssignment;
+            const { id, assigned_to, status } = taskAssignment;
             return {
                 type: TASK_ASSIGNMENT,
                 id,
                 assigned_to,
-                message,
-                status: status.toLowerCase(),
+                status,
```
jstoffan: The `toLowerCase` here might still be necessary, depending on how the API is updated.

src/api/TaskAssignments.js
```diff
@@ -116,7 +113,7 @@ class TaskAssignments extends Base {
         }
 
         const requestData = {
-            data: { resolution_state: resolutionState, message },
+            data: { resolution_state: taskStatus },
```
alexkrolick: so we read `status` but write `resolution_state`?

ConradJChan: updated, we'll only be using `status` now

flow-typed/box-ui-elements.js
```diff
@@ -523,8 +523,7 @@ type TaskAssignment = {
     type: 'task_assignment',
     id: string,
     assigned_to: User,
-    resolution_state: string,
-    message: string,
+    status: string,
```
DanDeMicco: Can you update the type to be an enum?

PR #721, opened 2018-12-19
fix(sidebar): details sidebar has stale cache when switching tabs

DanDeMicco:
`ContentSidebar` passes the file object down as props to most of the child sidebars such as `DetailsSidebar`. `DetailsSidebar` updates the cache for the description, but when switching to another tab and going back to the details sidebar, it receives a stale file object as props and renders the stale description.

There are some TODO comments in this PR which will need to be resolved when refactoring files calls into each sidebar but will not have any impact for now, since `ContentSidebar` handles fetching of the file information and populating cache.


boxcla:

src/api/Metadata.js
```diff
@@ -285,20 +285,19 @@ class Metadata extends File {
...
+     * @param {string} id - The file id
      * @param {Function} successCallback - Success callback
      * @param {Function} errorCallback - Error callback
      * @param {boolean|void} [options.forceFetch] - Optionally Bypasses the cache
      * @param {boolean|void} [options.refreshCache] - Optionally Updates the cache
      * @return {Promise}
      */
     async getClassification(
-        file: BoxItem,
+        id: string,
```
jstoffan: Can we name this `fileId`? Otherwise it looks like we're "getting a classification with ID of X".

src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -54,7 +52,7 @@ type State = {
     classification?: ClassificationInfo,
     classificationError?: Errors,
     isLoadingClassification: boolean,
-    file: BoxItem,
+    file?: BoxItem,
```
jstoffan: Is this still used? If not, let's remove it.

src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -73,13 +71,22 @@ class DetailsSidebar extends React.PureComponent<Props, State> {
     constructor(props: Props) {
         super(props);
         this.state = {
-            file: props.file,
```
jstoffan: I actually didn't realize we were tracking the same data in two (or more) different states. That was bound to lead to trouble. Glad we're starting to address it.

src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -73,13 +71,22 @@ class DetailsSidebar extends React.PureComponent<Props, State> {
...
             isLoadingAccessStats: false,
             isLoadingClassification: false,
         };
     }
 
     componentDidMount() {
+        this.fetchFile();
+        this.fetchData();
```
jstoffan: The naming here is pretty vague. Can we move the content of `fetchData` directly inside `componentDidMount`, instead?

src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -91,14 +98,54 @@ class DetailsSidebar extends React.PureComponent<Props, State> {
...
     /**
-     * File description update callback
+     * Fetches a file with the fields needed for details sidebar
+     *
+     * @param {string} id - File id
+     * @param {Object|void} [fetchOptions] - Fetch options
+     * @return {void}
+     */
+    fetchFile(
+        fetchOptions: FetchOptions = {},
```
jstoffan: This parameter appears to be unused.

src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -91,14 +98,54 @@ class DetailsSidebar extends React.PureComponent<Props, State> {
     }
 
     /**
-     * File description update callback
+     * Fetches a file with the fields needed for details sidebar
+     *
+     * @param {string} id - File id
```
jstoffan: Parameter list in the docs doesn't look correct.

src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -131,6 +178,10 @@ class DetailsSidebar extends React.PureComponent<Props, State> {
     onDescriptionChange = (newDescription: string): void => {
         const { api }: Props = this.props;
         const { file }: State = this.state;
+        if (!file) {
```
jstoffan: It doesn't seem like this case can actually occur, since we return null in render if file is falsy.

src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -139,7 +190,10 @@ class DetailsSidebar extends React.PureComponent<Props, State> {
         api.getFileAPI().setFileDescription(
             file,
             newDescription,
-            this.descriptionChangeSuccessCallback,
+            () => {
+                // Refetch the file from cache
+                this.fetchFile();
```
jstoffan: Do we need to fetch the file object again after updating the description? Or can we just set state with the response as we were before?

src/api/Metadata.js
DanDeMicco: Yeah I can, was just trying to do minimal changes to reduce the diff

src/components/ContentSidebar/DetailsSidebar.js
DanDeMicco: This is the state, which is still used to store the result of the files call (cached)

src/components/ContentSidebar/DetailsSidebar.js
DanDeMicco: I actually think this might be good to have separated for now, since we will have to implement a `componentDidUpdate`

src/components/ContentSidebar/DetailsSidebar.js
DanDeMicco: I added it to appease the flow gods

src/components/ContentSidebar/DetailsSidebar.js
DanDeMicco: This is true, the response actually is the cache value. Putting it back

src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -317,6 +377,10 @@ class DetailsSidebar extends React.PureComponent<Props, State> {
             isLoadingClassification,
         }: State = this.state;
 
+        if (!file) {
+            return null; // TODO: change to loading indicator and handle errors once file call split out
```
priyajeet: You can also look at `<LoadingIndicatorWrapper>` (used in MetadataSidebar.js) if that helps at all...though I think that requires some content to overlay...

src/components/ContentSidebar/DetailsSidebar.js
jstoffan: That's my mistake, I thought I was looking in Props.

PR #722, opened 2018-12-19
new(logging): Sample PR of how we can log render times

JustinHoldstock:


boxcla:

src/components/ContentSidebar/ActivitySidebar.js
```diff
@@ -8,15 +8,23 @@ import * as React from 'react';
...
 import { DEFAULT_COLLAB_DEBOUNCE, ORIGIN_ACTIVITY_SIDEBAR } from '../../constants';
 import API from '../../api';
+import Logger from '../../logger';
 import './ActivitySidebar.scss';
+import Timer from '../../util/Timer'; // Note that this is a thin wrapper around performance api
+// To track whether or not the Activity Feed JS is loaded
+let isAFJSLoaded = false;
+const ActivityFeed = React.lazy(() => {
+    const afPromise = import('./ActivityFeed/activity-feed/ActivityFeed');
+    afPromise.then(() => (isAFJSLoaded = true));
```
JustinHoldstock: As far as I can tell, running something inside of the `then` is the only way we can know that the import has loaded. Note that Activity Feed will never load until first render occurs

src/components/ContentSidebar/ActivitySidebar.js
```diff
@@ -56,6 +65,18 @@ export const activityFeedInlineError: Errors = {
         content: messages.activityFeedItemApiError,
     },
 };
+
+// The time it takes for ActivityFeed JS to be loaded, but still loading feed items and current user.
+let hasLoadRenderMetricBeenRecorded = false;
+// The time it takes for ActivityFeed JS to be loaded and presented with complete data.
+let hasRenderMetricBeenRecorded = false;
+// According to React docs, ActivityFeed loaded via React.lazy() WILL NOT begin loading until first render occurs
+let hasAFJSBegunLoading = false;
```
JustinHoldstock: These three are in the highest scope since we only want to record this ONCE

src/components/ContentSidebar/ActivitySidebar.js
```diff
@@ -438,29 +459,59 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
             activityFeedError,
             currentUserError,
         } = this.state;
+
+        // Since this is the first render, we know that ActivityFeed will finally be lazy loaded in!
```
JustinHoldstock: This is where things get a little messy. `React.lazy()` ONLY begins loading when the first `render()` occurs. We need to start a timer when this happens. After it DOES happen, then we need to start listening for the AF to be in the loading state AND/OR the full renderable state. 

Note that this logic dictates that if a full render occurs before the load state is rendered, the time it took to get to a load state render will not be recorded. This can occur when `currentUser` and `feedItems` API requests come back before the ActivityFeed component JS is loaded

src/components/ContentSidebar/ActivitySidebar.js
```diff
@@ -438,29 +459,59 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+        if (!hasAFJSBegunLoading) {
+            Timer.mark(ACTIVITY_FEED_START_TAG);
+            hasAFJSBegunLoading = true;
+        } else if (isAFJSLoaded) {
+            // If the JS is loaded, then we can stop some timers!
+            if (!hasRenderMetricBeenRecorded && currentUser && feedItems) {
+                hasRenderMetricBeenRecorded = true;
```
JustinHoldstock: Some of the internals of this `if` are duplicated, and can be abstracted away to a logger/timer utility function

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -420,24 +428,26 @@ class ContentSidebar extends PureComponent<Props, State> {
...
-                                    hasSkills={hasSkills}
-                                    hasDetails={hasDetails}
-                                    hasMetadata={hasMetadata}
-                                    hasActivityFeed={hasActivityFeed}
-                                    onToggle={this.onToggle}
-                                    onVersionHistoryClick={onVersionHistoryClick}
-                                    currentUser={currentUser}
-                                />
-                            </APIContext.Provider>
+                            <LoggerContext.Provider value={(this.logger: any)}>
```
jstoffan: I think this will need to live in the consuming app, such as EUA. Some developers may not want or need our logs.

src/logger/Logger.js
```diff
@@ -0,0 +1,49 @@
+type LoggerOptions = {
+    onMetricLog?: Function
+};
+
+type MetricData = {
+    component: string,
+    message: string,
+    value?: string | number,
+    timestamp: string,
+    session_id: string,
```
jstoffan: `sessionId`?

src/components/ContentSidebar/ActivitySidebar.js
JustinHoldstock: Great point

src/components/ContentSidebar/ActivitySidebar.js
jeremypress: assuming the component will be created right away, could we move to the constructor?

src/components/ContentSidebar/ActivitySidebar.js
```diff
@@ -438,29 +459,59 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
             activityFeedError,
             currentUserError,
         } = this.state;
+
+        // Since this is the first render, we know that ActivityFeed will finally be lazy loaded in!
+        if (!hasAFJSBegunLoading) {
+            Timer.mark(ACTIVITY_FEED_START_TAG);
```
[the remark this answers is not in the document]
JustinHoldstock: If we put it in the root element (assuming that's ContentSidebar, in this case), we're going to have skewed results due to certain elements not rendering until they are presented/user initiated.

IE) In the ContentSidebar, we have Activity Feed, Details Sidebar, and Metadata. When the screen is small-ish, nothing is opened, but the sidebar is rendered as collapsed. If the user has to open the AF sidebar, our results are going to be skewed to "How long did the user take to open the sidebar" PLUS actual JS and API load times.

src/components/ContentSidebar/ActivitySidebar.js
JustinHoldstock: I'm doing away with this code entirely, and marking the "start" as ActivitySidebar constructor, and the Activity Feed marks are occurring inside the ActivityFeed itself (meaning that the AF JS has been loaded)

wenboyu2:
@JustinHoldstock I don't think `new` is a valid type, commitlint should complain. Did you use `--no-verify`? The commit message/PR title should be `feat(logging): ...` instead.

JustinHoldstock:
@wenboyu2 This is a POC PR. I'll have something a little more official in the future

src/logger/Logger.js
JustinHoldstock: Ah, have to lowercase to follow schema validation

JustinHoldstock: I think we'll want it back to `session_id`. The analytics repo standardizes on snake_case

src/components/ContentSidebar/ActivitySidebar.js
```diff
@@ -56,13 +59,16 @@ export const activityFeedInlineError: Errors = {
...
+
 class ActivitySidebar extends React.PureComponent<Props, State> {
     state = {};
 
     componentDidMount() {
         const { currentUser } = this.props;
         this.fetchFeedItems(true);
         this.fetchCurrentUser(currentUser);
+        // Start time to interaction timer
+        Timer.mark(ACTIVITY_SIDEBAR_TAGS.Initialized);
```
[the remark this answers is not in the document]
JustinHoldstock: I might've commented back, but that is going to give us inaccurate results. We want to know how long it took from initialization to render. If we start the mark when the file is loaded, and then the component isn't rendered until a user initiated action, our timings are going to be skewed. 

We can make *another* mark outside of the class (something like ACTIVITY_SIDEBAR_TAGS.JSReady) then we can track the time it to to load the JS a little easier, assuming that's what you're thinking of

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -67,6 +78,38 @@ class ActivityFeed extends React.Component<Props, State> {
...
+    checkAndLogRenderMetrics(currentUser?: User, feedItems?: FeedItems) {
+        const { logger } = this.props;
+        const hasLoggedRender = logger.hasLoggedEvent(ACTIVITY_FEED_COMPONENT, EVENT_TIME_TO_RENDER);
+        if (!hasLoggedRender) {
+            if (currentUser && feedItems) {
+                Timer.mark(ACTIVITY_FEED_TAGS.TimeToRender);
+                logger.logUniqueMetric(
+                    ACTIVITY_FEED_COMPONENT,
+                    EVENT_TIME_TO_RENDER,
+                    Timer.getDuration(ACTIVITY_SIDEBAR_TAGS.Initialized, ACTIVITY_FEED_TAGS.TimeToRender),
```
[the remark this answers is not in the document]
JustinHoldstock: I'm not sure that the logger needs to be coupled to the timing mechanism, does it?

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -67,6 +78,38 @@ class ActivityFeed extends React.Component<Props, State> {
...
+        }
+    }
+
+    /**
+     * Log render and load state timing events.
+     *
+     * @param {*} currentUser
+     * @param {*} feedItems
+     */
+    checkAndLogRenderMetrics(currentUser?: User, feedItems?: FeedItems) {
```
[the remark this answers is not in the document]
JustinHoldstock: As each component has its own requirements for what is considered a render metric, I think we'll want to add this to each component that requires logging. Any thoughts in a different direction would be really helpful, though

PR #723, opened 2018-12-21
 feat: add retry logic for 429 and general network errors

DanDeMicco:
This adds retry logic for any general network errors and 429. It will the request retry up to 3 times. I have excluded `ContentUploader` from this by passing a flag as it has some internal retry logic, but it will work for any elements which uses `Xhr`


boxcla:

src/util/Xhr.js
```diff
@@ -69,39 +81,90 @@ class Xhr {
...
+     */
+    shouldRetryRequest(error: $AxiosError<any>): boolean {
+        if (!this.shouldRetry || this.retryCount >= MAX_NUM_RETRIES) {
+            return false;
+        }
+
+        const { response, request } = error;
+        // Retry if there is a network error (e.g. ECONNRESET) or rate limited
+        const isNetworkError = request && !response;
+        const isRetryableStatusCode = isNetworkError || getProp(response, 'status') === HTTP_STATUS_CODE_RATE_LIMIT;
```
priyajeet: Obviously the more generic way of doing this would be to let the caller tell what HTTP codes to retry on. But if/when we get use cases of more, specially if some functions want to retry on different codes, then we can come back and refactor.

src/util/Xhr.js
```diff
@@ -69,39 +81,90 @@ class Xhr {
...
+    /**
+     * Calculate the exponential backoff time with randomized jitter. Taken from box node SDK
+     *
+     * @param {number} numRetries Which retry number this one will be. Must be > 0
+     * @returns {number} The number of milliseconds after which to retry
+     */
+    getExponentialRetryTimeoutInMs(numRetries: number): number {
+        const minRandomization = 1 - RETRY_RANDOMIZATION_FACTOR;
+        const maxRandomization = 1 + RETRY_RANDOMIZATION_FACTOR;
+        const randomization = Math.random() * (maxRandomization - minRandomization) + minRandomization;
```
JustinHoldstock: Since `RETRY_RANDOMIZATION_FACTOR` is `0.5`, `maxRandomization - minRandomization` will always be `1`.
Simplify to `Math.random() + minRandomization`

src/util/Xhr.js
```diff
@@ -69,39 +81,90 @@ class Xhr {
...
+        shouldRetry = true,
     }: Options = {}) {
         this.id = id;
         this.token = token;
         this.clientName = clientName;
         this.version = version;
         this.sharedLink = sharedLink;
         this.sharedLinkPassword = sharedLinkPassword;
-        this.responseInterceptor = responseInterceptor;
+        this.responseInterceptor = responseInterceptor || this.defaultResponseInterceptor;
```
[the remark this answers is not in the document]
DanDeMicco: Nope, axios doesn't seem to like that

src/util/Xhr.js
```diff
@@ -69,39 +81,90 @@ class Xhr {
...
-    errorInterceptor = (error: Object): Promise<any> => {
-        const errorObject = getProp(error, 'response.data', error);
-        if (typeof this.responseInterceptor === 'function') {
-            this.responseInterceptor(errorObject);
+    errorInterceptor = (error: $AxiosError<any>): Promise<any> => {
+        const shouldRetry = this.shouldRetryRequest(error);
+        if (shouldRetry) {
+            this.retryCount += 1;
+            const delay = this.getExponentialRetryTimeoutInMs(this.retryCount);
+            return new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
DanDeMicco: Doesn't look like it since this is done via axios

src/util/Xhr.js
DanDeMicco: Yeah, It should be easy to add in an optional option which implements a custom function but defaults.

src/util/Xhr.js
DanDeMicco: This is true, it was a copy + paste from node SDK. I did some more research and simplified the algorithm based off https://cloud.google.com/iot/docs/how-tos/exponential-backoff

PR #724, opened 2018-12-28
feat(openwith): Add support for 'beta' google integrations

jeremypress:
the platform version of the gsuite integrations will be considered "beta" for now. This change will allow the correct icons to be used during and after this beta period.


boxcla:

PR #726, opened 2018-12-28
fix(sidebar): details sidebar updates when changing props

DanDeMicco:


src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -78,7 +78,22 @@ class DetailsSidebar extends React.PureComponent<Props, State> {
...
         this.fetchFile();
-        this.fetchData();
+        this.fetchAccessStats();
+        this.fetchClassification();
+    }
+
+    componentDidUpdate(prevProps: Props) {
+        const { hasAccessStats, hasClassification } = this.props;
+        // Component visibility props such as hasAccessStats can sometimes be flipped after an async call
+        const hasAccessStatsVisibilityChanged = prevProps.hasAccessStats !== hasAccessStats;
```
jstoffan: Minor: It seems inconsistent to add "Visibility" to these variables; seems like it should just be `hasAccessStatsChanged`.

src/components/ContentSidebar/DetailsSidebar.js
```diff
@@ -263,7 +262,20 @@ class DetailsSidebar extends React.PureComponent<Props, State> {
      * @return {void}
      */
     fetchAccessStats(): void {
-        const { api, fileId }: Props = this.props;
+        const { api, fileId, hasAccessStats }: Props = this.props;
+        const { isLoadingAccessStats } = this.state;
+        if (!hasAccessStats) {
```
ConradJChan: Do we need to worry about a scenario like the API call is in flight, and then `hasAccessStats` flips to `false`, causing two setStates, first `accessStats: undefined` then the result of `fetchAccessStatsSuccessCallback`?

PR #727, opened 2018-12-29
feat(tasks): Add "add task" button and modal in Activity Sidebar

alexkrolick:
- [x] Adds new modal-based task create UX behind a feature flag
- [x] Improve validation handling: ported-over ApprovalCommentForm does everything on submit; this could be better
- [x] Refactor components: "Add Task" button + form can probably be pulled out of ActivitySidebar
- [x] Tests for form, button, modal

### DIFF
--- a/examples/src/features.js
+++ b/examples/src/features.js
@@ -1,2 +1,4 @@
 // feature config for feature flags
-export default {};
+export default {
+    tasks: true,
+};
--- a/examples/src/features.js
+++ b/examples/src/features.js
@@ -1,4 +1,8 @@
 // feature config for feature flags
 export default {
-    tasks: true,
+    activityFeed: {
+        tasks: {
+            createButton: true,
+        },
+    },
 };
--- a/i18n/en-US.properties
+++ b/i18n/en-US.properties
@@ -328,8 +328,12 @@ be.taskEditErrorMessage = There was an error updating this task.
 be.taskReject = Decline
 # label for button that opens task popup
 be.tasks.addTask = Add Task
+# label for task create form assignee input
+be.tasks.addTaskForm.assigneesLabel = Assignees
 # label for cancel button in create task popup
 be.tasks.addTaskForm.cancel = Cancel
+# label for task create form due date input
+be.tasks.addTaskForm.dueDateLabel = Due Date
 # label for task create form message input
 be.tasks.addTaskForm.messageLabel = Message
 # label for create button in create task popup
--- a/i18n/en-US.properties
+++ b/i18n/en-US.properties
@@ -326,6 +326,16 @@ be.taskDueDate = Due
 be.taskEditErrorMessage = There was an error updating this task.
 # Reject option for a task
 be.taskReject = Decline
+# label for button that opens task popup
+be.tasks.addTask = Add Task
+# label for cancel button in create task popup
+be.tasks.addTaskForm.cancel = Cancel
+# label for task create form message input
+be.tasks.addTaskForm.messageLabel = Message
+# label for create button in create task popup
+be.tasks.addTaskForm.submit = Add Task
+# title for task popup
+be.tasks.addTaskForm.title = Add Task
 # Tasks for approval
 be.tasksForApproval = Tasks
 # Shown instead of todays date.
--- a/src/elements/common/messages.js
+++ b/src/elements/common/messages.js
@@ -843,6 +843,31 @@ const messages: { [string]: MessageDescriptor } = defineMessages({
         defaultMessage: 'Decline',
         description: 'Reject option for a task',
     },
+    tasksAddTask: {
+        id: 'be.tasks.addTask',
+        defaultMessage: 'Add Task',
+        description: 'label for button that opens task popup',
+    },
+    tasksAddTaskFormTitle: {
+        id: 'be.tasks.addTaskForm.title',
+        defaultMessage: 'Add Task',
+        description: 'title for task popup',
+    },
+    tasksAddTaskFormMessageLabel: {
+        id: 'be.tasks.addTaskForm.messageLabel',
+        defaultMessage: 'Message',
+        description: 'label for task create form message input',
+    },
+    tasksAddTaskFormSubmitLabel: {
+        id: 'be.tasks.addTaskForm.submit',
+        defaultMessage: 'Add Task',
+        description: 'label for create button in create task popup',
+    },
+    tasksAddTaskFormCancelLabel: {
+        id: 'be.tasks.addTaskForm.cancel',
+        defaultMessage: 'Cancel',
+        description: 'label for cancel button in create task popup',
+    },
     versionDeleted: {
         id: 'be.versionDeleted',
         defaultMessage: '{ name } deleted version { version_number }',
--- a/src/elements/common/messages.js
+++ b/src/elements/common/messages.js
@@ -853,10 +853,20 @@ const messages: { [string]: MessageDescriptor } = defineMessages({
         defaultMessage: 'Add Task',
         description: 'title for task popup',
     },
+    tasksAddTaskFormAssigneesLabel: {
+        id: 'be.tasks.addTaskForm.assigneesLabel',
+        defaultMessage: 'Assignees',
+        description: 'label for task create form assignee input',
+    },
     tasksAddTaskFormMessageLabel: {
         id: 'be.tasks.addTaskForm.messageLabel',
         defaultMessage: 'Message',
         description: 'label for task create form message input',
+    },
+    tasksAddTaskFormDueDateLabel: {
+        id: 'be.tasks.addTaskForm.dueDateLabel',
+        defaultMessage: 'Due Date',
+        description: 'label for task create form due date input',
     },
     tasksAddTaskFormSubmitLabel: {
         id: 'be.tasks.addTaskForm.submit',
--- a/src/elements/content-sidebar/ActivitySidebar.js
+++ b/src/elements/content-sidebar/ActivitySidebar.js
@@ -14,6 +14,8 @@ import messages from '../common/messages';
 import { withAPIContext } from '../common/api-context';
 import { withErrorBoundary } from '../common/error-boundary';
 import { getBadUserError, getBadItemError } from '../../utils/error';
+import AddTaskButton from './AddTaskButton';
+import { FeatureFlag } from '../common/feature-checking';
 import { DEFAULT_COLLAB_DEBOUNCE, ORIGIN_ACTIVITY_SIDEBAR } from '../../constants';
 import API from '../../api';
 import './ActivitySidebar.scss';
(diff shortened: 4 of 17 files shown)


boxcla:

src/components/ContentPreview/ContentPreview.js
```diff
@@ -1078,73 +1083,89 @@ class ContentPreview extends PureComponent<Props, State> {
...
-                            {hasLeftNavigation && (
-                                <PlainButton type="button" className="bcpr-navigate-left" onClick={this.navigateLeft}>
-                                    <IconNavigateLeft />
-                                </PlainButton>
-                            )}
-                            {hasRightNavigation && (
-                                <PlainButton type="button" className="bcpr-navigate-right" onClick={this.navigateRight}>
-                                    <IconNavigateRight />
-                                </PlainButton>
+                <FeatureProvider features={features}>
```
alexkrolick: this change just wraps everything with a feature provider

src/components/ContentPreview/ContentPreview.js
```diff
@@ -1078,73 +1083,89 @@ class ContentPreview extends PureComponent<Props, State> {
...
+                                )}
+                            </div>
+                            {file && (
+                                <ContentSidebar
+                                    {...contentSidebarProps}
+                                    isLarge={isLarge}
+                                    apiHost={apiHost}
+                                    token={token}
+                                    cache={this.api.getCache()}
+                                    features={contentSidebarProps.features || features}
```
alexkrolick: ContentSidebar is a top-level export from BUIE, so it also needs to have a FeatureProvider. This passes the parent feature options down when it used in the Preview.

src/components/ContentSidebar/ActivityFeed/task-form/TaskForm.js
```diff
@@ -0,0 +1,312 @@
+/**
```
alexkrolick: this file is based on ApprovalCommentForm

src/components/ContentSidebar/ActivitySidebar.js
```diff
@@ -428,6 +432,54 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         return api.getUsersAPI(false).getAvatarUrlWithAccessToken(userId, file.id);
     };
 
+    renderAddTaskButton = () => {
```
jstoffan: This seems more appropriate as its own component with internally-managed state.

src/components/ContentSidebar/ActivitySidebar.js
```diff
@@ -428,6 +432,54 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+        const { isDisabled } = this.props;
+        const { isTaskFormOpen, currentUser } = this.state;
+        const closeTaskModal = () => this.setState({ isTaskFormOpen: false });
```
jstoffan: Try to avoid recreating event handlers that contain only static code paths.

src/components/ContentSidebar/ActivitySidebar.js
```diff
@@ -428,6 +432,54 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+        const openTaskModal = () => this.setState({ isTaskFormOpen: true });
+        const handleClickAdd = openTaskModal;
```
jstoffan: What is the reason for reassigning/renaming these variables? I would have thought this would cause an eslint warning.

src/components/ContentSidebar/ActivityFeed/task-form/TaskForm.js
```diff
@@ -0,0 +1,312 @@
...
+                            selectorOptions={approverOptions}
+                            isRequired
+                        >
+                            {approverOptions.map(({ id, name, email }) => (
+                                <ContactDatalistItem key={id} name={name} subtitle={email} />
+                            ))}
+                        </PillSelectorDropdown>
+                        <input type="hidden" name="addApproval" id="addApproval" value="on" />
+                        {/* 
+                        <TextArea
```
jstoffan: Please avoid committing commented code.

src/components/ContentSidebar/ActivityFeed/task-form/TaskForm.js
jstoffan: How similar is this file to ApprovalCommentForm? Can we avoid duplicating it?

src/components/ContentPreview/ContentPreview.js
jstoffan: Could this use a HOC to avoid the additional wrapper?

src/components/ContentPreview/ContentPreview.js
jstoffan: Is it possible for `FeatureProvider` to check if an existing `FeatureContext` value already exists and, if so, use that? That should allow the value to cascade without the need to explicitly `OR` check each time.

src/components/ContentPreview/ContentPreview.js
```diff
@@ -29,6 +29,8 @@ import { getTypedFileId } from '../../util/file';
 import { withErrorBoundary } from '../ErrorBoundary';
 import ReloadNotification from './ReloadNotification';
 import { PREVIEW_FIELDS_TO_FETCH } from '../../util/fields';
+import { FeatureProvider } from '../FeatureChecking';
+import type { FeatureConfig } from '../FeatureChecking';
```
jstoffan: Can you move this import to `flow-typed/box-ui-elements.js` and see if flow can detect it implicitly?

jstoffan:
Please open this against a feature branch (fork or canonical), such as `feature-tasks`. You can update the PR to [point at the new base branch](https://blog.github.com/2016-08-15-change-the-base-branch-of-a-pull-request/) without closing it.

src/components/ContentPreview/ContentPreview.js
alexkrolick: My preference to solve this issue would be to not import top-level exports into each other

src/components/ContentSidebar/ActivityFeed/task-form/TaskForm.js
alexkrolick: It's increasingly divergent and would require a lot of layout props like "form element order", "enable mentions", etc.

src/components/ContentPreview/ContentPreview.js
alexkrolick: no, it takes a prop

alexkrolick: unless you want to implicitly extract the prop from the parent into the provider inside the HOC

src/components/ContentPreview/ContentPreview.js
alexkrolick: Oh, we consume our own flow-typed definition? That's normally for third-party code: https://flow.org/en/docs/libdefs/creation/#declaring-a-global-function-

src/components/ContentPreview/ContentPreview.js
jstoffan: Yep, that's what I was thinking.

src/components/ContentPreview/ContentPreview.js
jstoffan: I agree that the current top-level cascade isn't ideal, but it's a much larger refactor to change that at the moment. The code snippet you posted looks like it's worth trying.

src/components/ContentSidebar/ActivityFeed/task-form/TaskForm.js
jstoffan: Can we extract any additional common sub-components or behaviors to reduce the duplication?

src/components/ContentPreview/ContentPreview.js
jstoffan: That's the convention today, at least. Seems like something to follow for now and refactor holistically in a separate PR.

src/components/ContentSidebar/ActivityFeed/task-form/index.js
```diff
@@ -0,0 +1,3 @@
+// @flow
```
jstoffan: We're still following PascalCase for directory names, at least for the time being.

src/components/ContentSidebar/ActivitySidebar.js
```diff
@@ -428,6 +434,22 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         return api.getUsersAPI(false).getAvatarUrlWithAccessToken(userId, file.id);
     };
 
+    renderAddTaskButton = () => {
+        const { isDisabled = false } = this.props;
```
jstoffan: Seems best to leave the default value to the `defaultProps` above to avoid duplication/confusion.

src/components/ContentSidebar/AddTaskButton.js
```diff
@@ -0,0 +1,86 @@
...
+    handleClickAdd = () => this.setState({ isTaskFormOpen: true });
+
+    handleClose = () => this.setState({ isTaskFormOpen: false });
+
+    handleSubmit = () => this.setState({ isTaskFormOpen: false });
+
+    render() {
+        const { isDisabled, ...passThrough } = this.props;
+        const { isTaskFormOpen } = this.state;
+        const { handleClickAdd, handleClose, handleSubmit } = this;
```
jstoffan: Minor, but the convention I've usually seen for passing handlers is `onClickAdd={this.handleClickAdd}` with `on` for the prop name and `handle` for the value.

src/components/ContentSidebar/ActivityFeed/task-form/TaskForm.js
alexkrolick: The data layer would make the most sense to refactor, the forms are pretty similar. Will see at the end if there's anything to extract. Also would be better to refactor after integration tests are written.

src/components/ContentSidebar/ActivityFeed/task-form/index.js
alexkrolick: not in this directory, this matches what's there

src/components/ContentPreview/ContentPreview.js
alexkrolick: OK, refactoring/extending the feature framework seems like it can be a separate PR

src/components/messages.js
```diff
@@ -848,6 +848,41 @@ const messages: { [string]: MessageDescriptor } = defineMessages({
...
+        id: 'be.tasks.addTaskForm.messageLabel',
+        defaultMessage: 'Message',
+        description: 'label for task create form message input',
+    },
+    tasksAddTaskFormDueDateLabel: {
+        id: 'be.tasks.addTaskForm.dueDateLabel',
+        defaultMessage: 'Due Date',
+        description: 'label for task create form due date input',
+    },
+    tasksAddTaskFormSubmitLabel: {
```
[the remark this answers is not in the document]
alexkrolick: They are coincidentally the same but not set in stone. No point in coupling prematurely.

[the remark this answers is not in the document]
alexkrolick: They are coincidentally the same but not set in stone. No point in coupling prematurely.

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -377,7 +379,7 @@ class ContentSidebar extends React.PureComponent<Props, State> {
         const styleClassName = classNames(
             'be bcs',
             {
-                [`bcs-${((selectedView: any): string)}`]: isOpen,
+                [`bcs-${selectedView || ''}`]: isOpen,
```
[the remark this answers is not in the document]
alexkrolick: Right now it results in bcs-undefined instead and breaks syntax highlighting in vscode so this is just a tweak to fix that

alexkrolick:
OK, this is rebased on the reorganized folder structure

PR #728, opened 2019-01-02
chore(sidebar): refactor / cleanup of content sidebar

priyajeet:
- make object attributes in ascending order
- remove cwrp
- default state properties
- break up file fetch and metadata fetch

Use Cases
- When the user has not interfered with the sidebar
  - The sidebar should open when viewport is large enough (current threshold 1000)
  - The sidebar should close when viewport is small
  - When navigating from file to file, tabs should switch based on this order of precedence: skills > activity > details > metadata

- When the user has interfered with the sidebar
  - Sidebar should remain closed if the user closed it when navigating between files
  - Sidebar should stick to prior chosen tab if it exists

- When navigating between files sidebar should show loading indicator
- When opening file for 1st time, then sidebar should only show up after file load, so that we dont have two loading indicators
- Sidebar should not show on shared file page unless skills exist
- Sidebar should show for shared folder page with details and optionally skills

### DIFF
--- a/flow-typed/box-ui-elements.js
+++ b/flow-typed/box-ui-elements.js
@@ -52,6 +52,7 @@ import {
     SKILLS_FACE,
     SKILLS_STATUS,
     SIZE_MEDIUM,
+    SIDEBAR_VIEW_NONE,
     SIDEBAR_VIEW_SKILLS,
     SIDEBAR_VIEW_ACTIVITY,
     SIDEBAR_VIEW_DETAILS,
@@ -608,6 +609,7 @@ type JsonPatch = {
 type JsonPatchData = Array<JsonPatch>;
 
 type SidebarView =
+    | typeof SIDEBAR_VIEW_NONE
     | typeof SIDEBAR_VIEW_SKILLS
     | typeof SIDEBAR_VIEW_DETAILS
     | typeof SIDEBAR_VIEW_METADATA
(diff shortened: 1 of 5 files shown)


boxcla:

src/api/uploads/FolderUploadNode.js
```diff
@@ -103,6 +103,9 @@ class FolderUploadNode {
             // @TODO: Handle 429
             if (error.code !== ERROR_CODE_ITEM_NAME_IN_USE) {
                 errorCallback(error);
+                if (isRoot) {
```
priyajeet: revert

flow-typed/box-ui-elements.js
```diff
@@ -52,6 +52,7 @@ import {
     SKILLS_FACE,
     SKILLS_STATUS,
     SIZE_MEDIUM,
+    SIDEBAR_VIEW_NONE,
```
priyajeet: revert file

src/components/ContentSidebar/SidebarUtils.js
```diff
@@ -88,17 +88,17 @@ class SidebarUtils {
      *
      * @private
      * @param {ContentSidebarProps} props - User passed in props
```
priyajeet: This file can also be reverted since its just variable renaming

src/components/ContentSidebar/Sidebar.js
```diff
@@ -9,7 +9,6 @@ import DetailsSidebar from './DetailsSidebar';
 import SkillsSidebar from './SkillsSidebar';
 import ActivitySidebar from './ActivitySidebar';
 import MetadataSidebar from './MetadataSidebar';
```
priyajeet: This file can also be reverted if desired, but if reverting, then the render() function of ContentSidebar.js should just send in the props needed by this file

PR #729, opened 2019-01-02
feat(openwith): Box Edit SFC support

jeremypress:


boxcla:

jeremypress:
Box Edit and Box Edit SFC are two distinct integrations. If both integrations are present (meaning they are both turned on and the token used has a high enough scope), the more powerful Box Edit integration will be used. Box Tools expects a "token_scope" to help determine what the provided token is capable of.

src/api/OpenWith.js
```diff
@@ -48,13 +54,38 @@ class OpenWith extends Base {
...
+     */
+    consolidateBoxEditIntegrations(openWithIntegrations: OpenWithAPI): OpenWithAPI {
+        let { items } = openWithIntegrations;
+        const boxEditIntegration = items.find(
+            integration => integration.app_integration.id === BOX_EDIT_INTEGRATION_ID,
+        );
+
+        if (boxEditIntegration) {
+            items = items.filter(integration => {
+                return integration.app_integration.id !== BOX_EDIT_SFC_INTEGRATION_ID;
```
jstoffan: This can be made to implicitly return to stay consistent with the `.find` expression above.

src/api/OpenWith.js
```diff
@@ -48,13 +54,38 @@ class OpenWith extends Base {
...
 
+    /**
+     * Removes the Box Edit SFC integration if the higher scoped Box Edit integration is present.
+     *
+     * @param {OpenWithAPI} openWithIntegrations - Open With integration items
+     * @return {OpenWithAPI} Open With Integrations with only one Box Edit integration
+     */
+    consolidateBoxEditIntegrations(openWithIntegrations: OpenWithAPI): OpenWithAPI {
+        let { items } = openWithIntegrations;
+        const boxEditIntegration = items.find(
```
jstoffan: Since this is used as a boolean check, consider using `items.some`, instead.

src/api/OpenWith.js
```diff
@@ -48,13 +54,38 @@ class OpenWith extends Base {
...
+        );
+
+        if (boxEditIntegration) {
+            items = items.filter(integration => {
+                return integration.app_integration.id !== BOX_EDIT_SFC_INTEGRATION_ID;
+            });
+        }
+
+        return {
+            ...openWithIntegrations,
```
jstoffan: Did you intend to spread the top level object into the same level as its child? This means that you'll have both the content of `items` and an `items` property in the result.

src/api/OpenWith.js
```diff
@@ -48,13 +54,38 @@ class OpenWith extends Base {
...
+     * @return {OpenWithAPI} Open With Integrations with only one Box Edit integration
+     */
+    consolidateBoxEditIntegrations(openWithIntegrations: OpenWithAPI): OpenWithAPI {
+        let { items } = openWithIntegrations;
+        const boxEditIntegration = items.find(
+            integration => integration.app_integration.id === BOX_EDIT_INTEGRATION_ID,
+        );
+
+        if (boxEditIntegration) {
+            items = items.filter(integration => {
```
jstoffan: This is mutating the `openWithIntegrations` param, since `items` is passed by reference, which isn't ideal. Can you just return the new object below, instead?

src/api/OpenWith.js
```diff
@@ -48,13 +54,38 @@ class OpenWith extends Base {
...
+                const consolidatedOpenWithIntegrations = this.consolidateBoxEditIntegrations(openWithIntegrations);
+                const formattedOpenWithData = this.formatOpenWithData(consolidatedOpenWithIntegrations);
                 successCallback(formattedOpenWithData);
             },
             errorCallback,
         });
     }
 
+    /**
+     * Removes the Box Edit SFC integration if the higher scoped Box Edit integration is present.
```
jstoffan: Can you update the comment to explain the "why" here, if we know?

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -188,7 +195,17 @@ class ContentOpenWith extends PureComponent<Props, State> {
...
+    }
+
+    /**
+     * Checks if a given integration is a Box Edit integration.
+     *
+     * @param {string} [integrationId] - The integration ID
+     * @return {boolean}
+     */
+    isBoxEditSFCIntegration(integrationId: ?string): boolean {
+        return integrationId === BOX_EDIT_SFC_INTEGRATION_ID;
```
jstoffan: It looks like this logic is now duplicated in the `OpenWith` component. In the future, it seems like it'd be better to have each item in the dropdown decide whether it should be shown rather than having the parent do it.

jeremypress: We would still need this logic at the top level to branch on handling successful integrations, as well as deciding what scope to pass to Box Tools. 

If each item decided whether to render or not, it would need knowledge of the other integrations. e.g. as a Box Edit SFC integration I should hide myself if the regular Box Edit integration exists.

PR #731, opened 2019-01-03
fix(sidebar): Maintain toggle state across file navigation

ConradJChan:
Introduce `isOpen` state in sidebar in order to persist toggle state when navigating across different files

### DIFF
--- a/src/components/ContentSidebar/ContentSidebar.js
+++ b/src/components/ContentSidebar/ContentSidebar.js
@@ -43,7 +43,6 @@ type Props = {
     className: string,
     clientName: string,
     currentUser?: User,
-    defaultView?: SidebarView,
     detailsSidebarProps: DetailsSidebarProps,
     fileId?: string,
     getPreview: Function,
@@ -226,15 +225,9 @@ class ContentSidebar extends React.PureComponent<Props, State> {
      */
     getSidebarView(): ?SidebarView {
         const { file, isOpen, metadataEditors, view }: State = this.state;
-        const { defaultView }: Props = this.props;
 
         if (!isOpen) {
             return undefined;
-        }
-
-        // If there was a default view provided, force use that
-        if (defaultView) {
-            return defaultView;
         }
 
         let newView = view;
@@ -318,11 +311,9 @@ class ContentSidebar extends React.PureComponent<Props, State> {
      * @return {void}
      */
     fetchFileSuccessCallback = (file: BoxItem): void => {
-        const view = this.getSidebarView();
         this.setState(
             {
                 file,
-                view,
                 isLoading: false,
             },
             this.fetchMetadata,
@@ -376,7 +367,8 @@ class ContentSidebar extends React.PureComponent<Props, State> {
             return null;
         }
 
-        const selectedView = isOpen ? this.state.view : undefined;
+        // const selectedView = isOpen ? this.state.view : undefined;
+        const selectedView = this.getSidebarView();
         const hasSkills = SidebarUtils.shouldRenderSkillsSidebar(this.props, file);
         const hasDetails = SidebarUtils.canHaveDetailsSidebar(this.props);
         const hasMetadata = SidebarUtils.shouldRenderMetadataSidebar(this.props, metadataEditors);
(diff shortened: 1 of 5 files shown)


src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -156,36 +153,32 @@ class ContentSidebar extends PureComponent<Props, State> {
...
-            this.setState({ ...this.initialState });
-            this.fetchData(nextProps);
-        } else if (!hasBeenToggled && hasVisibilityChanged) {
-            this.setState({
-                view: this.getDefaultSidebarView(nextProps, file, editors),
-            });
+        const { fileId: prevFileId, isLarge: prevIsLarge }: Props = prevProps;
+        const { view }: State = this.state;
+
+        if (fileId !== prevFileId) {
```
jeremypress: Any reason why you're removing the variables that were here before?

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -107,30 +104,30 @@ class ContentSidebar extends PureComponent<Props, State> {
...
             requestInterceptor,
             responseInterceptor,
+            sharedLink,
+            sharedLinkPassword,
+            token,
         });
 
-        // Clone initial state to allow for state reset on new files
-        this.state = { ...this.initialState };
+        this.state = { isLoading: true, isOpen: !!isLarge };
```
jeremypress: This is fine to set since this component won't unmount while navigating, and we're ok with a reset in behavior on preview close?

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -262,28 +237,11 @@ class ContentSidebar extends PureComponent<Props, State> {
             return defaultView;
```
jeremypress: What will happen if the default view is set to something like skills but there are no skills on the file? We can address that in a later change.

src/components/ContentSidebar/ContentSidebar.js
ConradJChan: Mainly that it's a different function, I think. Do you see any issues with it?

src/components/ContentSidebar/ContentSidebar.js
ConradJChan: Just checked with the T3 behavior and it is different from what we have in EUA, we should probably create a ticket to address this in elements and EUA. We'll probably need to expose a `defaultOpen` prop and some sort of toggle callback prop so that EUA can keep track of the state across preview instances

priyajeet:
@ConradJChan for UX testing, I listed the use cases here in the PR message
https://github.com/box/box-ui-elements/pull/658

src/components/ContentSidebar/ContentSidebar.js
priyajeet: What's the issue?

ConradJChan: If you toggle close the sidebar, and then 'X' out of preview, when you preview another file, the sidebar defaults to open always in EUA, but in T3 it remembers what state it was last

src/components/ContentSidebar/ContentSidebar.js
ConradJChan: oh right, sidebar is not GA'ed nor documented, so I'll remove it then

src/components/ContentSidebar/ContentSidebar.js
jeremypress: nope, just a difference in style. Fine by me

src/components/ContentSidebar/ContentSidebar.js
priyajeet: Worst case we just store it to session storage

priyajeet: Or in the in-memory cache.

ConradJChan:
Thanks @priyajeet, I've run through the use cases and it seems to be in order

priyajeet:
Can you update PR message and description

PR #732, opened 2019-01-03
test(cypress): Add support for integration tests via cypress

jstoffan:
This isn't quite ready to merge yet due to some outstanding issues:

- [x] Improve the example test and selectors
- [x] Improve eslint rules via cypress plugin
- [x] Decide on using /examples or /test harness files
- [x] Decide on nested component testing format (ContentExplorer -> ContentPreview -> ContentSidebar)
- [x] Update CI tasks to run tests on merge and release


boxcla:

.eslintrc.js
```diff
@@ -2,9 +2,16 @@ const eslintrc = require.resolve('@box/frontend/eslint/eslintrc.js');
 
 module.exports = {
     extends: [eslintrc],
```
alexkrolick: there's a cypress eslint preset that adds some helpful stuff

alexkrolick:
This is great!

.eslintrc.js
jstoffan: Yeah agreed, but we should bring it into the frontend config project, I think. This is just a temporary workaround to get support up as quickly as possible.

test/cypress/integration/ContentExplorer/ContentExplorer.spec.js
```diff
@@ -0,0 +1,32 @@
...
+        beforeEach(() => {
+            cy.contains('Sample CSV.csv').click();
+        });
+
+        afterEach(() => {
+            cy.getByTitle('Close').click();
+        });
+
+        it('Navigation within a collection keeps sidebar open', () => {
+            cy.queryByTestId('bcs-content').should('exist');
```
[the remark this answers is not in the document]
jstoffan: You can, but doing so is considered an [anti-pattern](https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements). I'm going to update the left/right buttons to have titles we can select by, though.

test/cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
pramodsum: Can we remove this?

test/cypress/plugins/index.js
```diff
@@ -0,0 +1,19 @@
+// ***********************************************************
```
pramodsum: I think this is an example file that we don't need?

test/cypress/support/index.js
```diff
@@ -0,0 +1,17 @@
+// ***********************************************************
```
pramodsum: I think this is also an example file that we don't need right?

package.json
```diff
@@ -50,6 +50,9 @@
         "build:prod:npm": "BABEL_ENV=production OUTPUT=dist LANGUAGE=en-US REACT=true yarn build:prod:dist",
         "clean": "rm -rf dist lib i18n/json i18n/*.js reports styleguide",
         "copy:styles": "cpx './src/**/*.scss' lib",
+        "cy:open": "yarn cy:wait; cypress open --project ./test",
+        "cy:run": "yarn cy:wait; cypress run --project ./test --spec test/cypress/**/*.spec.js",
+        "cy:wait": "wait-on tcp:6060",
```
ConradJChan: Are there any concerns with hardcoding the port? Parallelism issues in Travis maybe?

test/cypress.json
```diff
@@ -0,0 +1,6 @@
+{
+    "baseUrl": "http://localhost:6060/#",
```
DanDeMicco: Can we use the test folder instead, and inject the access token? This way we can use a non-hardcoded access token which could have additional scopes such as those needed for creating comments.

src/components/ContentSidebar/SidebarContent.js
```diff
@@ -14,7 +14,7 @@ type Props = {
 };
 
 const SidebarContent = ({ actions, title, children }: Props) => (
-    <div className="bcs-content">
+    <div className="bcs-content" data-testid="bcs-content">
```
DanDeMicco: It seems like we will have to add a bit of these to our code. Can we remove them with something like https://www.npmjs.com/package/babel-plugin-react-remove-properties ?

test/cypress.json
jstoffan: The example files accept a token via the `TOKEN` environment variable.

test/cypress/support/index.js
jstoffan: No, it's the central import for support file, such as `commands.js` below.

.npmignore
```diff
@@ -1,47 +1,47 @@
 *~
```
priyajeet: Can you also add code owners, commit lint and style lint files to this

src/components/ContentSidebar/SidebarContent.js
jstoffan: We can look into doing so in a follow-up PR. I'd like to avoid transforming the lib/dist output with this change, if possible.

test/cypress/integration/ContentExplorer/ContentExplorer.spec.js
```diff
@@ -0,0 +1,32 @@
+describe('ContentExplorer', () => {
+    before(() => {
```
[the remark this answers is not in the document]
jstoffan: I only need it to happen once.

test/cypress/integration/ContentExplorer/ContentExplorer.spec.js
```diff
@@ -0,0 +1,32 @@
...
+    describe('ContentSidebar', () => {
+        beforeEach(() => {
+            cy.contains('Sample CSV.csv').click();
+        });
+
+        afterEach(() => {
+            cy.getByTitle('Close').click();
+        });
+
+        it('Navigation within a collection keeps sidebar open', () => {
```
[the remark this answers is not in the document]
jstoffan: Cypress only supports Mocha/Chai, not Jest. They have an [open issue](https://github.com/cypress-io/cypress/issues/281) for it already.

test/cypress/plugins/index.js
jstoffan: I'm going to leave it for instructional purposes for future test authors. I don't think it's hurting anything, since it's not included in our bundles.

test/cypress/integration/ContentExplorer/ContentExplorer.spec.js
```diff
@@ -0,0 +1,32 @@
+describe('ContentExplorer', () => {
```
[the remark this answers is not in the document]
jstoffan: EUA is using `.spec` for Cypress scenarios, as well. I'm happy to change it, though. Can someone confirm the correct file naming?

test/cypress/integration/ContentExplorer/ContentExplorer.spec.js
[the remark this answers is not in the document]
j3tan: is there a possibility of side-effects between tests? i.e. first test sets some state inside sidebar, second test unintentionally relies on that.

.eslintrc.js
jstoffan: Good point. I'll add the Cypress plugin here shortly.

test/cypress/integration/ContentExplorer/ContentExplorer.spec.js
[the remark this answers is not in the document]
jstoffan: I suppose, yeah. I can add a `beforeEach` containing `cy.reload()`.

test/cypress/integration/ContentExplorer/ContentExplorer.spec.js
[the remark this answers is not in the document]
priyajeet: @davidjohnburrowes ☝️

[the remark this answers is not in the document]
priyajeet: @jstoffan at least for the path, you can start using content-explorer, lower case with -

test/cypress/plugins/index.js
jstoffan: This file is also expected by Cypress, from what I can tell.

test/cypress/integration/ContentExplorer/ContentExplorer.spec.js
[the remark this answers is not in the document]
jstoffan: @j3tan, I misspoke earlier with regard to EUA. Admin Console is the project upon which I partially based my original file/folder structure. I'll send the dev who worked on it a note about the naming convention issue.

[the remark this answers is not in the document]
davidjohnburrowes: @priyajeet True enough.  There's comments on the side, and the image is wrong.  I'll update.  It's also on confluence now.

test/integration/content-explorer/ContentExplorer.test.js
```diff
@@ -0,0 +1,37 @@
+// <reference types="Cypress" />
+
+describe('ContentExplorer', () => {
```
davidjohnburrowes: fwiw, this seems like it is something other than a unit test, and I'd pictured that those would end up being ContentExplorer.int.test.js (or whatever the test type is).  Not saying change it, just saying: maybe we do need some other kind of category here?

jstoffan: Cypress is an end-to-end testing library meant to simulate user interactions via a browser. It's not really intended to be used for unit tests. I originally had the files with the `.spec.js` extension, which delineates them against regular unit tests.

j3tan: the folder structure does that already

jstoffan: That's true, as well, @j3tan. In that case, it doesn't seem like the `.int.test.js` change is necessary, either.

src/components/ContentPreview/PreviewNavigation.js
```diff
@@ -0,0 +1,52 @@
...
+
+type Props = {
+    collection: Array<String | BoxItem>,
+    currentIndex: number,
+    intl: any,
+    onNavigateLeft: Function,
+    onNavigateRight: Function,
+};
+
+const PreviewNavigation = ({ collection = [], currentIndex, intl, onNavigateLeft, onNavigateRight }: Props) => {
```
DanDeMicco: I know you added a functional test, but can you add some UT's?

package.json
```diff
@@ -152,6 +158,8 @@
         "cpx": "^1.5.0",
         "css-loader": "^1.0.0",
         "cssnano": "^4.1.0",
+        "cypress": "^3.1.4",
+        "cypress-testing-library": "^2.3.5",
```
priyajeet: I did not notice that we were using another library on top of cypress. I would like to see if we can reduce the need of another library for testing and stick to just official cypress stuff that comes out of the box. Like will it cause more developer friction to write integration tests if we don't use this? Or will it be a one time hit that people re-use (like 1-2 utilities)? Or is the native cypress stuff in-adequate that we need this? I don't really have a A vs B comparison to go off or any other data to make an educated guess here.

My main rationale for not wanting this is the fact that (if) we are paying for cypress then we should stick with what they offer. When I looked at the cypress documentation, I saw this

screen shot 2019-01-04 at 10 45 25 am

They don't mention that you have to additionally use another library unless I missed it. Their whole idea is that the developer isn't picking x, y, z to get stuff working. Likewise, if we pay them and need support at some point in the future, it would be easier without non official cypress stuff. In other words, we should ideally not be playing an integration engineer between two libraries. Likewise when QA/Dev want to look at documentation, they are looking in 1 place, not two.

On the other hand, if using this library makes things substantially easier, then sure...

cc @j3tan @davidjohnburrowes

package.json
```diff
@@ -61,7 +64,9 @@
         "start:dev": "yarn setup; LANGUAGE=en-US BABEL_ENV=dev NODE_ENV=dev node --max_old_space_size=8192 node_modules/webpack-dev-server/bin/webpack-dev-server.js --config conf/webpack.config.js --mode development",
         "start:examples": "yarn setup; LANGUAGE=en-US REACT=true BABEL_ENV=dev NODE_ENV=dev node --max_old_space_size=8192 node_modules/react-styleguidist/bin/styleguidist.js server --config conf/styleguide.config.js --mode development",
         "start:npm": "yarn setup; npm-run-all build:prod:npm build:dev:es",
-        "test": "BABEL_ENV=test NODE_ENV=test yarn jest"
+        "test": "BABEL_ENV=test NODE_ENV=test yarn jest",
+        "test:e2e": "npm-run-all -p -r start:examples cy:run",
```
priyajeet: Wondering if we should rename `e2e` to `integration` OR `integration` to `e2e` at other places because I think there are subtle differences, again based on a recent doc i saw. @djordan correct me if I am wrong, integration testing ideally does not go beyond testing the stack (so front end only but multiple components/units) while e2e goes beyond that so full stack (FE + backend).

Now I think here we are doing e2e tests here, because we are using the examples page, which is actually making network requests to the API and loading real live data. So we are doing e2e tests?

test/integration/content-explorer/ContentExplorer.test.js
jstoffan: I renamed the test file to `ContentExplorer.e2e.test.js` to identify it as an end-to-end test.

PR #734, opened 2019-01-03
fix: Add dimension to make icon sizes consistent

jeremypress:
FileIcon takes a different prop `dimension` from other icons.


boxcla:

jeremypress:
I use 26 in two files and 30 in one, doesn't hurt to add as constants.

PR #737, opened 2019-01-04
perf: code splitting in ContentPreview

DanDeMicco:
This adds code splitting for ContentPreview and ContentSidebar via react.lazy. Code splitting will be disabled in dist.


src/components/AsyncLoad.js
```diff
@@ -0,0 +1,16 @@
...
+import retry from '../util/retry';
+
+type Props = {
+    loader: () => Promise<any>,
+};
+
+const AsyncLoad = (props: Props) =>
+    Loadable({
+        loading: () => null,
+        ...props,
```
jeremypress: passed in props could override loading but not the loader method, is that intended?

src/components/ContentPreview/ContentPreview.js
```diff
@@ -132,6 +132,10 @@ const RETRY_COUNT = 3; // number of times to retry network request for a file
 const MS_IN_S = 1000; // ms in a sec
 const PREVIEW_LOAD_METRIC_EVENT = 'load';
 
+const LoadableSidebar = AsyncLoad({
+    loader: () => import(/* webpackMode: "lazy", webpackChunkName: "content-sidebar" */ '../ContentSidebar'),
```
jeremypress: Where does this chunk name get set? I don't see it in our webpack config.

src/components/ContentPreview/Header.js
```diff
@@ -30,6 +30,11 @@ type Props = {
     token: ?string,
 };
 
+const LoadableContentOpenWith = AsyncLoad({
+    loader: () =>
+        import(/* webpackMode: "lazy", webpackChunkName: "content-open-with" */ '../ContentOpenWith/ContentOpenWith'),
```
jeremypress: Thanks for adding this, was this being loaded by webapp even though it will never be used?

src/util/retry.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * @flow
```
jeremypress: Is it worth extracting some of this repeated stuff out of XHR.js so it can be referenced in both places?

src/components/AsyncLoad.js
DanDeMicco: yeah I want to enforce retries

src/components/ContentPreview/ContentPreview.js
DanDeMicco: This sets the output chunk name https://webpack.js.org/api/module-methods/#magic-comments

src/components/ContentPreview/Header.js
DanDeMicco: Yes everything is loaded by webapp which is in ContentPreview. We are also loading every API even though we are only using a couple

src/util/retry.js
DanDeMicco: While there is some small similarities with the retry stuff I added in xhr, it isn't really the same. This file is a simplified retry which recursively calls itself, while xhr does some specific stuff in regards to when to retry and calls axios instead of recursively calling itself

priyajeet: Can we use function.js?

package.json
```diff
@@ -62,6 +62,7 @@
         "setup": "yarn install --frozen-lockfile; npm-run-all clean build:i18n",
         "start": "yarn start:examples",
         "start:dev": "yarn setup; LANGUAGE=en-US BABEL_ENV=dev NODE_ENV=dev node --max_old_space_size=8192 node_modules/webpack-dev-server/bin/webpack-dev-server.js --config conf/webpack.config.js --mode development",
+        "start:prod": "yarn setup; LANGUAGE=en-US BABEL_ENV=production NODE_ENV=dev node --max_old_space_size=8192 node_modules/webpack-dev-server/bin/webpack-dev-server.js --config conf/webpack.config.js --mode production",
```
priyajeet: whats this for?

src/components/ContentPreview/Header.js
priyajeet: ContentOpenWith has an index.js, is ContentOpenWith/ContentOpenWith needed?

src/components/ContentPreview/ContentPreview.js
priyajeet: For both LoadableSidebar and LoadableContentOpenWith, should we have them in their own files in their respective parent folders exported?

priyajeet:
Perhaps PR tag should be `perf:` in this case?

conf/webpack.config.js
```diff
@@ -122,6 +122,22 @@ function getConfig(isReactExternalized) {
         );
     }
 
+    if (isRelease) {
+        // For release builds, disable code splitting. https://webpack.js.org/api/module-methods/#magic-comments
+        config.module.rules = [
```
priyajeet: Array.prototype.unshift() also works

src/elements/content-sidebar/SidebarUtils.js
```diff
@@ -119,6 +131,74 @@ class SidebarUtils {
...
     }
+
+    /**
+     * Gets the title for a given sidebar view
+     *
+     * @param {string} view - the view name
+     * @return {React.Node} - the node to render
+     */
+    static getTitleForView(view: SidebarView): React.Node {
+        switch (view) {
```
jeremypress: Could you use the view as the key of the messages dict to avoid this switch?

README.md
```diff
@@ -379,6 +379,8 @@ render(
 | --- | --- | --- | --- |
 | *tbd* | - | - | - |
 
+# Code Splitting
+[Code splitting](https://webpack.js.org/guides/code-splitting/) is currently supported for the `ContentPreview` and `ContentSidebar` UI Elements. In order to use an Element with code splitting, you will need to build the Element with webpack by importing the Element from the `lib` folder in our npm package.
```
priyajeet: `es` not `lib` ?

src/elements/common/async-load/AsyncLoad.js
```diff
@@ -0,0 +1,47 @@
...
+    loader: () => Promise<any>,
+    fallback?: React.Element<any>,
+    errorComponent?: React.ComponentType<any>,
+};
+
+type State = {
+    error: ?Error,
+};
+
+const NullComponent = () => null;
```
priyajeet: Can we use undefined instead of null? Then you can use lodash/noop

src/elements/content-sidebar/__tests__/SIdebarLoading-test.js
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react';
+import { shallow } from 'enzyme';
+import SidebarLoading from '../SidebarLoading';
+
+describe('elements/content-sidebar/SidebarLoading', () => {
```
priyajeet: Filename has `I` uppercase

src/elements/content-sidebar/SidebarLoading.js
```diff
@@ -0,0 +1,23 @@
...
+import * as React from 'react';
+import LoadingIndicator from 'box-react-ui/lib/components/loading-indicator/LoadingIndicator';
+import SidebarContent from './SidebarContent';
+import './SidebarLoading.scss';
+
+type Props = {
+    title: React.Node,
+};
+
+const SidebarLoading = ({ title }: Props) => {
```
priyajeet: Could we have used `LoadingIndicatorWrapper` instead?

src/elements/content-sidebar/Sidebar.js
```diff
@@ -91,7 +93,11 @@ const Sidebar = ({
             />
         )}
         {selectedView === SIDEBAR_VIEW_METADATA && (
-            <MetadataSidebar fileId={fileId} {...metadataSidebarProps} startMarkName={MARK_NAME_JS_LOADING_METADATA} />
+            <LoadableMetadataSidebar
+                fileId={fileId}
+                {...metadataSidebarProps}
+                startMarkName={MARK_NAME_JS_LOADING_METADATA}
```
priyajeet: Is the mark name needed here? Others dont have it

src/elements/content-sidebar/SidebarUtils.js
```diff
@@ -119,6 +131,74 @@ class SidebarUtils {
...
+    }
+
+    /**
+     * Marks and gets the loader for a given sidebar view
+     *
+     * @param {String} view - the view name
+     * @param {String} markName -  the name to be used by performance.mark
+     * @return {Function} - a function which will resolve the module to load
+     */
+    static getLoaderForView(view: SidebarView, markName: string): Promise<any> {
```
priyajeet: Is it possible to have this function in its own file + same additional stuff for `content-sidebar` and `content-open-with` also in it which are currently inside ContentPreview.js and Header.js.

So that all these are in 1 place/1 file to reference?

src/elements/content-sidebar/Sidebar.js
DanDeMicco: Yes, all sidebars have this. It is passed and consumed to a HoC which wraps each sidebar container (withLogger)

src/elements/common/async-load/AsyncLoad.js
DanDeMicco: its a stateless funcitonal component, so i believe it has to return null

src/elements/content-sidebar/SidebarUtils.js
DanDeMicco: The logic here is unique for sidebar, because we are creating a start mark. The start mark for preview and sidebar are created externally

src/utils/retry.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * @flow
+ * @file A utility function to retry a function which returns a promise
+ * @author Box
+ */
+const DEFAULT_NUM_RETRIES = 2;
+const DEFAULT_INITIAL_DELAY = 500;
+
+const retry = (
```
priyajeet: Can we use https://www.npmjs.com/package/promise-retry

priyajeet: We also have that super complicated retryNumOfTimes from function.js

src/elements/common/async-load/AsyncLoad.js
alexkrolick: undefined is not a valid return value for a react component

priyajeet: We should just have this NullComponent as a low level base component inside components.
And then re-used here as well as in FeatureFlag.js

src/utils/retry.js
priyajeet: Also found this https://gitlab.com/snippets/1775781
Bit more readable

DanDeMicco: I ended up using `retryNumOfTimes`. Not the prettiest or most intuitive, but it works

PR #741, opened 2019-01-05
build: Remove data-testid properties from compiled modules

jstoffan:


babel.config.js
```diff
@@ -34,8 +34,12 @@ module.exports = {
                         extensions: ['.scss', '.css'],
                     },
                 ],
+                ['react-remove-properties', { properties: ['data-testid'] }],
```
priyajeet: Is the plugin auto-included? Since I dont see package.json and yarn diff

PR #742, opened 2019-01-05
refactor(metadata): migrate to V2 API for fetching collabed metadata

priyajeet:
- Removes the webapp specific hack to fetch cross enterprise templates
- Isolates metadata sidebar file fetch call
- Increases template fetch limit to 1000
- Extracts classification from metadata fetch
- Upgrades to latest brui for metadata fixes and UI improvements
- Adds couple of new error messages
- Fixes minor UI CSS for loading indicator


src/api/Metadata.js
```diff
@@ -80,8 +80,12 @@ class Metadata extends File {
      * @param {string} scope - metadata scope or id
      * @return {string} base url for files
      */
-    getMetadataTemplateUrl(scope: string): string {
-        return `${this.getBaseApiUrl()}/metadata_templates/${scope}`;
+    getMetadataTemplateUrl(scope: string, isInstanceId?: boolean): string {
```
JustinHoldstock: update comment docs with `isInstanceId`

src/api/Metadata.js
```diff
@@ -137,14 +136,15 @@ class Metadata extends File {
...
+        this.errorCode = ERROR_CODE_FETCH_METADATA_TEMPLATES;
         let templates = {};
         try {
             templates = await this.xhr.get({
-                url: this.getMetadataTemplateUrl(scope),
+                url: this.getMetadataTemplateUrl(scope, isInstanceId),
                 id: getTypedFileId(id),
                 params: {
-                    limit: 1000,
+                    limit: 1000, // internal hard limit is 500
```
JustinHoldstock: Can we slap this value into a meaningful constant? I'd even argue in a config/constants file that an external dev can make quick modifications to, to allow for different limits.

src/components/ContentSidebar/MetadataSidebar.js
```diff
@@ -179,6 +185,20 @@ class MetadataSidebar extends React.PureComponent<Props, State> {
         this.setState({ editors: clone });
     }
 
+    /**
+     * Instance save error handler
+     *
+     * @param {Object} oldEditor - prior editor
+     * @param {Object} newEditor - updated editor
+     * @return {void}
```
JustinHoldstock: Update code docs

src/api/Metadata.js
```diff
@@ -137,14 +136,15 @@ class Metadata extends File {
      * @param {string} scope - metadata scope
      * @return {Object} array of metadata templates
      */
-    async getTemplates(id: string, scope: string): Promise<Array<MetadataEditorTemplate>> {
+    async getTemplates(id: string, scope: string, isInstanceId?: boolean): Promise<Array<MetadataEditorTemplate>> {
+        this.errorCode = ERROR_CODE_FETCH_METADATA_TEMPLATES;
         let templates = {};
         try {
             templates = await this.xhr.get({
```
DanDeMicco: I assume this API is paginated. Can you use one of the existing API methods to fetch all the metadata to make it more future proof?

src/api/Metadata.js
```diff
@@ -153,27 +153,220 @@ class Metadata extends File {
...
+
+    /**
+     * Extracts classification for different representation in the UI.
+     *
+     * @param {string} id - box file id
+     * @param {Array} instances - metadata instances
+     * @return {Array} metadata instances without classification
+     */
+    extractClassification(id: string, instances: Array<MetadataInstance>): Array<MetadataInstance> {
+        const index = instances.findIndex(instance => instance.$template === METADATA_TEMPLATE_CLASSIFICATION);
```
DanDeMicco: can you just use Array.prototype.find?

src/constants.js
```diff
@@ -224,6 +225,7 @@ export const ERROR_CODE_UPLOAD = 'upload_error';
 export const ERROR_CODE_UNEXPECTED_EXCEPTION = 'unexpected_exception_error';
 export const ERROR_CODE_SEARCH = 'search_error';
 export const ERROR_CODE_UNKNOWN = 'unknown_error';
+export const ERROR_CODE_METADATA_CONFLICT = 'metadata_conflic_error';
```
DanDeMicco: typeo

i18n/en-US.properties
```diff
@@ -108,6 +108,8 @@ be.drawAnnotation = Drawing annotation mode
 be.editLabel = Edit
 # Message to the user when there are no Open With integrations
 be.emptyOpenWithDescription = No integrations are available for this file
+# Generic error label.
+be.error = Error
```
jeremypress: We already have this exact string in 2 other places, should we remove those and keep this generic one?

src/api/Metadata.js
jeremypress: I agree that's more readable to use `find` and then just make an `indexOf` call if you need it for the `splice` below.

src/api/Metadata.js
```diff
@@ -153,27 +153,220 @@ class Metadata extends File {
...
+        canEdit: boolean,
+    ): Promise<Array<MetadataEditor>> {
+        // All usable templates for metadata instances
+        const templates: Array<MetadataEditorTemplate> = [customPropertiesTemplate].concat(
+            enterpriseTemplates,
+            globalTemplates,
+        );
+
+        // Filter out skills and classification
+        // let filteredInstances = this.extractSkills(id, instances);
```
jeremypress: Still needed?

src/components/ContentSidebar/MetadataSidebar.js
```diff
@@ -17,86 +17,84 @@ import SidebarContent from './SidebarContent';
...
-    getMetadataEditors = (): void => {
-        const { api, file, getMetadata, isFeatureEnabled = true }: Props = this.props;
-
-        api.getMetadataAPI(true).getEditors(
-            file,
-            ({ editors, templates }: { editors: Array<MetadataEditor>, templates: Array<MetadataEditorTemplate> }) => {
-                this.setState({
-                    templates,
-                    editors: editors.slice(0),
+    commonErrorCallback(error: ElementsXhrError, code: string, newState: Object = {}) {
```
jstoffan: `onApiError`?

src/components/ContentSidebar/MetadataSidebar.js
```diff
@@ -179,6 +185,20 @@ class MetadataSidebar extends React.PureComponent<Props, State> {
...
 
+    /**
+     * Instance save error handler
+     *
+     * @param {Object} oldEditor - prior editor
+     * @param {Object} newEditor - updated editor
+     * @return {void}
+     */
+    onSaveErrorHandler(oldEditor: MetadataEditor, error: ElementsXhrError, code: string): void {
+        const clone: MetadataEditor = { ...oldEditor }; // shallow clone only needed
```
jstoffan: Why is only a shallow clone needed?

src/components/ContentSidebar/MetadataSidebar.js
```diff
@@ -17,86 +17,84 @@ import SidebarContent from './SidebarContent';
...
      */
     canEdit(): boolean {
-        const { file }: Props = this.props;
+        const { file }: State = this.state;
+        if (!file) {
+            return false;
+        }
         const { permissions = {} }: BoxItem = file;
         const { can_upload }: BoxItemPermission = permissions;
         return !!can_upload;
```
jstoffan: `return getProp(file, 'permissions.can_upload', false);`

src/components/ContentSidebar/MetadataSidebar.js
```diff
@@ -212,23 +236,113 @@ class MetadataSidebar extends React.PureComponent<Props, State> {
      * @return {void}
      */
     onModification = (id: string, isDirty: boolean) => {
-        const { editors = [] }: State = this.state;
-        const index = editors.findIndex(({ instance }) => instance.id === id);
-        if (index === -1) {
+        const oldEditor = this.getEditor(id);
+        if (!oldEditor) {
```
jstoffan: Do we need to check `!file` here, as well?

src/components/ContentSidebar/MetadataSidebar.js
```diff
@@ -212,23 +236,113 @@ class MetadataSidebar extends React.PureComponent<Props, State> {
...
+        templates,
+    }: {
+        editors: Array<MetadataEditor>,
+        templates: Array<MetadataEditorTemplate>,
+    }) => {
+        this.setState({
+            editors: editors.slice(0),
+            error: undefined,
+            isLoading: false,
+            templates,
```
jstoffan: Should we copy `templates` similar to `editors`?

src/components/ContentSidebar/MetadataSidebar.js
```diff
@@ -212,23 +236,113 @@ class MetadataSidebar extends React.PureComponent<Props, State> {
...
+        });
+    };
+
+    /**
+     * Fetches the metadata editors
+     *
+     * @return {void}
+     */
+    fetchMetadata(): void {
+        const { api, isFeatureEnabled = true }: Props = this.props;
```
jstoffan: It seems like we should rely on `defaultProps` for `isFeatureEnabled` to avoid duplication.

i18n/en-US.properties
priyajeet: I'll do a followup removing the other versions

src/api/Metadata.js
priyajeet: Since templates are limited to 500, we can hold off doing the change for now, might involve a larger change for no benefit.

src/api/Metadata.js
priyajeet: Left it for now as a hint of where to make this upcoming change

src/components/ContentSidebar/MetadataSidebar.js
priyajeet: Not needed here since file isn't used. Also function isn't accessible if no file. Other places also the same holds true, but to keep flow happy null check is there

src/components/ContentSidebar/MetadataSidebar.js
```diff
@@ -313,17 +315,27 @@ class MetadataSidebar extends React.PureComponent<Props, State> {
...
+     * On file load we should fetch metadata, but we shouldn't need to fetch
+     * if the file permissions haven't changed from a prior file fetch.
+     * Metadata editors mostly care about upload permission.
      *
      * @param {Object} file - the box file
      * @return {void}
      */
     fetchFileSuccessCallback = (file: BoxItem) => {
-        this.setState({ file }, this.fetchMetadata);
+        const uploadPermission = 'permissions.can_upload';
```
priyajeet: I'll de-dupe this string also in another PR, make it a global constant, being repeated in other areas too

src/api/Metadata.js
```diff
@@ -135,16 +151,22 @@ class Metadata extends File {
      *
      * @param {string} id - file id
      * @param {string} scope - metadata scope
+     * @param {string|void} [instanceId] - metadata instance id
      * @return {Object} array of metadata templates
      */
-    async getTemplates(id: string, scope: string): Promise<Array<MetadataEditorTemplate>> {
+    async getTemplates(id: string, scope: string, instanceId?: string): Promise<Array<MetadataEditorTemplate>> {
+        this.errorCode = ERROR_CODE_FETCH_METADATA_TEMPLATES;
```
ConradJChan: How does `this.errorCode` get used in this method?

src/api/Metadata.js
```diff
@@ -154,26 +176,223 @@ class Metadata extends File {
...
+        instance: MetadataInstance,
+        templates: Array<MetadataEditorTemplate>,
+    ): Promise<?MetadataEditorTemplate> {
+        const instanceId = instance.$id;
+        const templateKey = instance.$template;
+        const scope = instance.$scope;
+        let template = templates.find(t => t.templateKey === templateKey && t.scope === scope);
+
+        // Enterprise scopes are always enterprise_XXXXX
+        if (scope.startsWith(METADATA_SCOPE_ENTERPRISE) && !template) {
```
ConradJChan: minor: consider swapping the order of the condition so `!template` could fail faster

src/api/Metadata.js
```diff
@@ -154,26 +176,223 @@ class Metadata extends File {
...
+
+        // Filter out skills and classification
+        // let filteredInstances = this.extractSkills(id, instances);
+        const filteredInstances = this.extractClassification(id, instances);
+
+        // Create editors from each instance
+        const editors: Array<MetadataEditor> = [];
+        await Promise.all(
+            filteredInstances.map(async instance => {
+                const template: ?MetadataEditorTemplate = await this.getTemplateForInstance(id, instance, templates);
```
ConradJChan: if one of the underlying `getTemplates` API call fails for whatever reason and throws an error, what is the expected behavior here?

src/api/Metadata.js
```diff
@@ -154,26 +176,223 @@ class Metadata extends File {
...
+
+        try {
+            const customPropertiesTemplate: MetadataEditorTemplate = this.getCustomPropertiesTemplate();
+            const [instances, globalTemplates, enterpriseTemplates] = await Promise.all([
+                this.getInstances(id),
+                this.getTemplates(id, METADATA_SCOPE_GLOBAL),
+                hasMetadataFeature ? this.getTemplates(id, METADATA_SCOPE_ENTERPRISE) : Promise.resolve([]),
+            ]);
+
+            const editors = await this.getEditors(
```
ConradJChan: does this need to await the `Promise.all` above before being called?

src/api/Metadata.js
```diff
@@ -154,26 +176,223 @@ class Metadata extends File {
...
+                id,
+                instances,
+                customPropertiesTemplate,
+                enterpriseTemplates,
+                globalTemplates,
+                !!permissions.can_upload,
+            );
+
+            if (this.isDestroyed()) {
```
ConradJChan: Would something cause `isDestroyed()` to return true only now? Can this be checked earlier in the method?

src/components/ContentSidebar/MetadataSidebar.js
```diff
@@ -17,86 +19,83 @@ import SidebarContent from './SidebarContent';
...
-    getMetadataEditors = (): void => {
-        const { api, file, getMetadata, isFeatureEnabled = true }: Props = this.props;
-
-        api.getMetadataAPI(true).getEditors(
-            file,
-            ({ editors, templates }: { editors: Array<MetadataEditor>, templates: Array<MetadataEditorTemplate> }) => {
-                this.setState({
-                    templates,
-                    editors: editors.slice(0),
+    onApiError(error: ElementsXhrError, code: string, newState: Object = {}) {
```
ConradJChan: Update jsdoc to reflect params

src/api/Metadata.js
DanDeMicco: The change is trivial to implement. You just need to extend the correct api

src/api/Metadata.js
```diff
@@ -154,26 +176,223 @@ class Metadata extends File {
...
+        if (cache.has(key)) {
+            this.successHandler(cache.get(key));
+            if (!options.refreshCache) {
+                return;
+            }
+        }
+
+        try {
+            const customPropertiesTemplate: MetadataEditorTemplate = this.getCustomPropertiesTemplate();
+            const [instances, globalTemplates, enterpriseTemplates] = await Promise.all([
```
DanDeMicco: Similar to conrads comment in another spot, what happens if one of these calls fail?

src/api/Metadata.js
```diff
@@ -154,26 +176,223 @@ class Metadata extends File {
...
+            const metadata = {
+                editors,
+                templates: this.getUserAddableTemplates(
+                    customPropertiesTemplate,
+                    enterpriseTemplates,
+                    hasMetadataFeature,
+                    is_externally_owned,
+                ),
+            };
+            cache.set(key, metadata);
```
DanDeMicco: I think this might be done elsewhere in our codebase, but if we have already fetched the data we should at least set the cache, but not call the callbacks

src/components/ContentSidebar/MetadataSidebar.js
```diff
@@ -172,13 +179,28 @@ class MetadataSidebar extends React.PureComponent<Props, State> {
...
+    /**
+     * Instance save error handler
+     *
+     * @param {Object} oldEditor - prior editor
+     * @param {Object} newEditor - updated editor
+     * @param {string} code - error code
+     * @return {void}
+     */
+    onSaveErrorHandler(oldEditor: MetadataEditor, error: ElementsXhrError, code: string): void {
+        const clone: MetadataEditor = { ...oldEditor }; // shallow clone suffices for hasError setting
```
DanDeMicco: nitpick:
```
{
...oldEditor,
hasError: true
}
```

also done in other places such as `onModification`

src/api/Metadata.js
priyajeet: Its not used here, its used in the parent Base.js errorHandler()

src/api/Metadata.js
priyajeet: There is a catch in there that can re-throw, which will be caught here and the errorHandler() will be called. Notice this is surrounded with try catch

src/api/Metadata.js
priyajeet: Promise.all is inside getEditors(), so the response of that will just be 1 promise

src/api/Metadata.js
priyajeet: Needs to be after async. There are 2 asyncs going on, so it can additionally be slightly higher also, but has to remain here. I just didnt add it before since this one will take care of that too, though at the expense of a couple of wasted API request. Main goal is not to call successHandler, which this one achieves.

src/api/Metadata.js
priyajeet: See other comment

src/api/Metadata.js
priyajeet: I moved it just to guard successHandler based on another comment

PR #743, opened 2019-01-07
fix(sidebar): Fix toggle behavior

ConradJChan:
In the `onToggle` method, if prior view is falsy (only when instantiated the first time) then call `this.getSidebarView`. Also added back default view and added sanity e2e tests for sidebar

### DIFF
--- a/src/components/ContentSidebar/ContentSidebar.js
+++ b/src/components/ContentSidebar/ContentSidebar.js
@@ -220,7 +220,7 @@ class ContentSidebar extends React.PureComponent<Props, State> {
     };
 
     /**
-     * File fetch success callback that sets the file and view
+     * Determines the current sidebar tab view
      *
      * @private
      * @return {string} Sidebar view to use
--- a/src/components/ContentSidebar/ContentSidebar.js
+++ b/src/components/ContentSidebar/ContentSidebar.js
@@ -43,6 +43,7 @@ type Props = {
     className: string,
     clientName: string,
     currentUser?: User,
+    defaultView?: SidebarView,
     detailsSidebarProps: DetailsSidebarProps,
     fileId?: string,
     getPreview: Function,
@@ -189,7 +190,8 @@ class ContentSidebar extends React.PureComponent<Props, State> {
      */
     onToggle = (view: SidebarView): void => {
         const { isOpen, view: priorView }: State = this.state;
-        const isClosing = isOpen && (!priorView || view === priorView);
+        const lastView = priorView || this.getSidebarView();
+        const isClosing = isOpen && view === lastView;
 
         this.setState({
             view,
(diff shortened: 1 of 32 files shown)


test/integration/content-explorer/ContentExplorer.e2e.test.js
```diff
@@ -23,7 +23,7 @@ describe('ContentExplorer', () => {
 
         it('Navigation within a collection keeps sidebar closed', () => {
             // Manually closing the sidebar should remove its content
-            cy.getByTestId('sidebaractivity').click();
+            cy.getByTestId('sidebarskills').click();
```
jstoffan: What's the reason for changing this value?

test/integration/content-sidebar/ContentSidebar.e2e.test.js
```diff
@@ -0,0 +1,30 @@
...
+        cy.getByTestId('bcs-content').should('not.exist');
+
+        cy.getByTestId('sidebarskills').click();
+        cy.getByTestId('bcs-content').should('exist');
+    });
+
+    it('Clicking other tabs should switch the sidebar tab', () => {
+        // Sidebar should be open by default
+        cy.getByTestId('bcs-content').should('exist');
+        cy.getByTestId('sidebarskills').should('have.class', 'bcs-nav-btn-is-selected');
```
jstoffan: Consider adding a sidebar-specific `data-testid` to the sidebar content component. That will allow you to avoid keying on specific CSS classes via something like `cy.getByTestId('sidebarskills-content').should('be.visible');`

test/integration/content-explorer/ContentExplorer.e2e.test.js
ConradJChan: Test was failing for me, and when I debugged it, I found that the 'Sample Audio.mp3' file is defaulting to the skills tab, not the activity feed, so in order to toggle it closed, I had to click the skills tab button

test/integration/content-explorer/ContentExplorer.e2e.test.js
```diff
@@ -32,5 +32,49 @@ describe('ContentExplorer', () => {
...
         });
+
+        // precedence order: skills > activity > details > metadata
+        it('Navigating from file to file, tabs should switch based on order of precedence', () => {
+            // Sidebar should be open by default
+            cy.getByTestId('bcs-content').should('exist');
+            // Skills tab should be selected
+            cy.getByTestId('sidebarskills').should('have.class', 'bcs-nav-btn-is-selected');
+
+            // Navigate to previous file should default to activity feed
```
jstoffan: Is this because the previous file is a specific file or type of file? If so, it might be good to add an assertion that we're looking at the expected file name.

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -225,11 +227,17 @@ class ContentSidebar extends React.PureComponent<Props, State> {
      */
     getSidebarView(): ?SidebarView {
```
jeremypress: Update the js doc since this is called on toggle now as well.

test/integration/content-explorer/ContentExplorer.e2e.test.js
ConradJChan: Yeah, i'll add that assertion 👍

PR #746, opened 2019-01-08
fix(sidebar): Don't display @mentions without getMentionWithQuery prop

pramodsum:


boxcla:

PR #747, opened 2019-01-08
fix: Use version of query-string that supports IE 11

jeremypress:


yarn.lock
```diff
@@ -3,9 +3,9 @@
 
 
 "@babel/cli@^7.0.0":
-  version "7.1.2"
-  resolved "https://registry.yarnpkg.com/@babel/cli/-/cli-7.1.2.tgz#fc2853ae96824b3779ca85de4fd025ce3cf62a5e"
-  integrity sha512-K3WDlpBPGpoW11SLKFEBhMsITomPovsrZ/wnM3y+WStbytukDXC0OBic3yQp+j058QUw0+R/jfx2obwp1fOzcA==
+  version "7.2.3"
```
jstoffan: I think we should handle upgrading other dependencies separately to this change.

PR #748, opened 2019-01-08
feat(logger): Add logging mechanism and example usage

JustinHoldstock:


src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -35,10 +39,18 @@ type Props = {
     feedItems?: FeedItems,
 };
 
+type Props = {
+    logger: LoggerType,
```
jeremypress: Since most people won't care about the logger props could we do `propsWithContext` and leave props alone?

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -35,10 +39,18 @@ type Props = {
...
+} & PropsWithoutContext;
+
 type State = {
     isInputOpen: boolean,
 };
 
+const ACTIVITY_FEED_COMPONENT = 'activity_feed';
```
jeremypress: Would these fit in the logger constants you defined in a separate file?

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -51,8 +63,8 @@ class ActivityFeed extends React.Component<Props, State> {
     }
 
     componentDidUpdate(prevProps: Props, prevState: State) {
-        const { feedItems: prevFeedItems } = prevProps;
-        const { feedItems: currFeedItems } = this.props;
+        const { feedItems: prevFeedItems, currentUser: prevCurrentUser } = prevProps;
+        const { feedItems: currFeedItems, currentUser: currCurrentUser } = this.props;
```
jeremypress: I think currentUser would suffice here. you're differentiating the old one with `prev`

src/logger/Logger.js
```diff
@@ -0,0 +1,125 @@
...
+ */
+
+/**
+ * Generates a GUID/UUID compliant with RFC4122 version 4.
+ *
+ * @return {string} A 36 character uuid
+ */
+function uuidv4() {
+    /* eslint-disable */
+    function generateRandom(c) {
```
jeremypress: can this be defined outside of `uuidv4`? It's getting redefined every time this function is called.

src/util/Timer.js
```diff
@@ -0,0 +1,17 @@
+/**
```
jeremypress: Are we planning on adding to this? We don't save any code by using this over performance.mark directly with the current implementation.

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -220,4 +268,4 @@ class ActivityFeed extends React.Component<Props, State> {
     }
 }
 
-export default ActivityFeed;
+export default withLoggerContext(ActivityFeed);
```
jeremypress: Why wrap the consumer with a HOC but use the context API directly for `ContentSidebar.js`?

src/util/Timer.js
```diff
@@ -0,0 +1,17 @@
...
+ * @flow
+ * @file Timer Class
+ * @author Box
+ */
+
+class Timer {
```
DanDeMicco: Can we get rid of this file and just use performance?

src/logger/Logger.js
```diff
@@ -0,0 +1,125 @@
...
+
+// Unique ID of this session.
+const SESSION_ID = uuidv4();
+
+class Logger {
+    /* Function callback to invoke with a metric object */
+    onMetricLog: Function = console.log; // eslint-disable-line no-console
+
+    /* A list of unique events that have occurred */
+    uniqueEvents: Array<string> = [];
```
DanDeMicco: can you change this to be a Set?

src/logger/Logger.js
```diff
@@ -0,0 +1,125 @@
...
+     *
+     * @param {string} component - the name of the component
+     * @param {string} name - the event name
+     * @param {string} from - Timer mark for the start of the timer
+     * @param {string} to - Timer mark for the end of the timer
+     * @param {boolean} [isUnique] - If true, guarantees a unique event firing
+     * @returns {void}
+     */
+    logTimeMetric(component: string, name: string, from: string, to: string, isUnique?: boolean = false): void {
+        const tagInfo = {
```
jstoffan: What's the reasoning for passing from/to rather than duration?

src/logger/Logger.js
```diff
@@ -0,0 +1,125 @@
...
+            from,
+            to,
+        };
+
+        if (isUnique) {
+            this.logUniqueMetric(component, name, tagInfo);
```
jstoffan: Where are the high-resolution timestamps extracted from the `performance` API?

src/util/Timer.js
JustinHoldstock: I don't see why not :)

src/logger/Logger.js
DanDeMicco: This allows more flexibility. Parent app may want to use a custom start tag, specifically for the js ready event, the parent app can choose and calculate what value to record for the duration

src/logger/Logger.js
DanDeMicco: The plan is to emit them via the `onMetric` prop, and have the parent app consume them.

src/logger/Logger.js
```diff
@@ -0,0 +1,125 @@
...
+     * Create an ISO Timestamp for right now.
+     *
+     * @returns {string}
+     */
+    getTimestamp(): string {
+        return new Date().toISOString();
+    }
+}
+
+export default new Logger();
```
jstoffan: If this is instead instantiated in the `LoggerContext`, it would simplify the import/export situation quite a bit and also keep its use as a "global" singleton more consistent. It also seems like `Logger` and `LoggerContext` should live in the same directory, which ideally would live in `/src/components` somewhere..

src/logger/Logger.js
jstoffan: Flexibility often comes at the cost of simplicity. Can you provide an example of where we intend to use this flexibility to solve the current requirements?

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -67,12 +78,48 @@ class ActivityFeed extends React.Component<Props, State> {
...
+                window.performance.mark(ACTIVITY_FEED_TAGS.TimeToRender);
+                logger.logTimeMetric(
+                    ACTIVITY_FEED_COMPONENT,
+                    EVENT_TIME_TO_RENDER,
+                    ACTIVITY_SIDEBAR_TAGS.Initialized,
+                    ACTIVITY_FEED_TAGS.TimeToRender,
+                    true,
+                );
+            } else if (!logger.hasLoggedEvent(ACTIVITY_FEED_COMPONENT, EVENT_TIME_TO_LOAD_STATE)) {
+                window.performance.mark(ACTIVITY_FEED_TAGS.TimeToLoadState);
```
DanDeMicco: I am unsure if this is accurate. I think you are trying to measure time to loading indicator displayed, but it will be displayed on first render which will occur before this. Should that be from from constructor to componentDidMount?

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -67,12 +78,48 @@ class ActivityFeed extends React.Component<Props, State> {
...
+     * Log render and load state timing events.
+     *
+     * @param {*} currentUser
+     * @param {*} feedItems
+     */
+    checkAndLogRenderMetrics(currentUser?: User, feedItems?: FeedItems) {
+        const { logger } = this.props;
+        const hasLoggedRender = logger.hasLoggedEvent(ACTIVITY_FEED_COMPONENT, EVENT_TIME_TO_RENDER);
+        if (!hasLoggedRender) {
+            if (currentUser && feedItems) {
```
DanDeMicco: can you abstract this into a function `isLoading`, which will be used in render as well?

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
JustinHoldstock: Sure would!

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -67,12 +79,48 @@ class ActivityFeed extends React.Component<Props, State> {
         if (hasLoaded || hasMoreItems || hasNewItems || hasInputOpened) {
             this.resetFeedScroll();
         }
+
+        if (currCurrentUser !== prevCurrentUser || currFeedItems !== prevFeedItems) {
```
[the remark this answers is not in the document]
JustinHoldstock: We do a check inside of the soon to be renamed `checkAndLogRenderMetrics()` to see if the event has already been logged, to prevent redundant logs. I'll set this up so that it runs that portion of the if check upon currFeedItems being set, and prev being null (as simple as `(currFeedItems && !prevFeedItems)`)

src/logger/loggingConstants.js
```diff
@@ -0,0 +1,13 @@
+// Activity Sidebar
+export const ACTIVITY_SIDEBAR_TAGS = {
+    Initialized: 'activity_sidebar_init',
+    JSReady: 'activity_sidebar_js',
+};
+
+// Activity Feed
+export const ACTIVITY_FEED_TAGS = {
```
DanDeMicco: maybe rename to constants.js since its already namespaced by the folder structure

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -67,12 +79,48 @@ class ActivityFeed extends React.Component<Props, State> {
...
+
+    /**
+     * Log render and load state timing events.
+     *
+     * @param {*} currentUser
+     * @param {*} feedItems
+     */
+    checkAndLogRenderMetrics(currentUser?: User, feedItems?: FeedItems) {
+        const { logger } = this.props;
+        const hasLoggedRender = logger.hasLoggedEvent(ACTIVITY_FEED_COMPONENT, EVENT_TIME_TO_RENDER);
```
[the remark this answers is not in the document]
JustinHoldstock: I'll change the name to `hasLoggedMeaningfulRender`

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
JustinHoldstock: I'm going to be honest, if loading is occurring in first render pass, I'm not sure we're going to get any use out of this metric. I'm going to remove it, to simplify for now. We can determine where loading state is useful to track, later.

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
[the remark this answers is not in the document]
JustinHoldstock: Actually, I can just used the above created `hasLoaded`

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -67,12 +73,47 @@ class ActivityFeed extends React.Component<Props, State> {
...
+
+    /**
+     * Log time to meaningful render
+     *
+     * @param {User} currentUser - CurrentUser for the session
+     * @param {FeedItems} feedItems - Feed items populating the activity feed
+     */
+    logRenderMetrics(): void {
+        const { logger } = this.props;
+        const hasLoggedRender = logger.hasLoggedEvent(ACTIVITY_FEED_TAGS.ComponentName, EVENT_TIME_TO_RENDER);
```
DanDeMicco: Can't we use `logUniqueMetric` which will do this check?

src/components/logger/Logger.js
```diff
@@ -0,0 +1,130 @@
...
+function uuidv4() {
+    return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, generateRandom);
+}
+
+// Unique ID of this session.
+const SESSION_ID = uuidv4();
+
+class Logger {
+    /* Function callback to invoke with a metric object */
+    onMetricLog: Function = console.log; // eslint-disable-line no-console
```
DanDeMicco: maybe default this to noop, since this will spam the console

src/components/logger/Logger.js
```diff
@@ -0,0 +1,130 @@
...
+
+    /* A list of unique events that have occurred */
+    uniqueEvents: Set = new Set();
```
DanDeMicco: Nit: typing here is unnecessary unless you add a type to the set

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -172,7 +212,7 @@ class ActivityFeed extends React.Component<Props, State> {
                     className="bcs-activity-feed-items-container"
                 >
                     {isEmpty ? (
-                        <EmptyState isLoading={isLoading} showCommentMessage={showApprovalCommentForm} />
+                        <EmptyState isLoading={this.isLoading()} showCommentMessage={showApprovalCommentForm} />
```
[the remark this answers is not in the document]
JustinHoldstock: This change was requested by Daniel in a previous comment. I can revert, if you prefer

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
JustinHoldstock: Oh, it's to avoid the redundant `performance.mark()`. I'm curious if we should instead wrap performance.mark in a thin wrapper to avoid redundant marking, since each `mark()` call with the same tag will create ANOTHER mark

src/components/ContentSidebar/ActivityFeed/activity-feed/ActivityFeed.js
```diff
@@ -67,12 +73,47 @@ class ActivityFeed extends React.Component<Props, State> {
...
+     * @param {User} currentUser - CurrentUser for the session
+     * @param {FeedItems} feedItems - Feed items populating the activity feed
+     */
+    logRenderMetrics(): void {
+        const { logger } = this.props;
+        const hasLoggedRender = logger.hasLoggedEvent(ACTIVITY_FEED_TAGS.ComponentName, EVENT_TIME_TO_RENDER);
+        if (!hasLoggedRender) {
+            window.performance.mark(ACTIVITY_FEED_TAGS.TimeToRender);
+            logger.logTimeMetric(
+                ACTIVITY_FEED_TAGS.ComponentName,
```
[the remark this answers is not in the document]
JustinHoldstock: They're known in the case of this component, yes. Not sure I follow what you're asking. Would the object be owned by this component, and still passed down? In the case of a method for the event, how would that differ from what is there right now?

PR #749, opened 2019-01-08
fix(sidebar): Fix defaultView logic

ConradJChan:
Tweaking `getSidebarView()` logic so that defaultView is only returned if the state `view` has not been set (which is user changed when toggling tabs)


src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -234,7 +234,8 @@ class ContentSidebar extends React.PureComponent<Props, State> {
         }
 
         // If there was a default view provided, force use that
-        if (defaultView) {
+        // only if the view has not been set
+        if (!view && defaultView) {
```
priyajeet: Double check if there is any interaction weirdness when resizing browser window, which should collapse or collapse the sidebar assuming its not interacted with

PR #750, opened 2019-01-09
test: add i18n/aria test utils

alexkrolick:
- ~~See https://github.com/box/box-ui-elements/blob/e0fad263b5cfbb2f6f1e189eec37c78e5dcef5ad/test/integration/content-sidebar/TaskForm.e2e.test.js for example~~
- ~~adds cypress-testing-library to replace home-built getByTitle, getByTestId, and add additional ARIA selectors (e.g. getByLabelText)~~
- adds a utility to run the translation library powering react-intl inside the test suite, to test using localized strings using message ids as the identifier (uses en-us as the language preset)


test/support/commands.js
```diff
@@ -1,2 +1 @@
-Cypress.Commands.add('getByTestId', testId => cy.get(`[data-testid="${testId}"]`));
-Cypress.Commands.add('getByTitle', title => cy.get(`[title="${title}"]`));
+import 'cypress-testing-library/add-commands';
```
jstoffan: We decided put adoption of this library on hold for now. We would prefer to use `cy` methods directly or via custom commands.

test/support/i18n.js
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * @param {string} messageId id of message in i18n file
+ * @param {Object} variables substitutions for translation, e.g. counts for plurals
+ * @example cy.getByText(localize('be.greeting', {name: 'Alex'})).click()
+ */
+const localize = (messageId, variables = {}) => {
+    return getFormatter(messageId).format(variables);
+};
+
+export { localize, localize as l };
```
jstoffan: I'd like to avoid importing this utility in every test. Can you move it to a global method or add a custom Cypress command, such as `Cypress.Commands.add('l', localize);`? Let's also standardize on using the `l` alias everywhere and not expose `localize`.

test/support/i18n.js
```diff
@@ -0,0 +1,17 @@
...
+import memoize from 'lodash/memoize';
+import messages from '../../i18n/en-US';
+
+const getFormatter = memoize(messageId => new IntlMessageFormat(messages[messageId], 'en-US'));
+
+/**
+ *
+ * @param {string} messageId id of message in i18n file
+ * @param {Object} variables substitutions for translation, e.g. counts for plurals
+ * @example cy.getByText(localize('be.greeting', {name: 'Alex'})).click()
```
jstoffan: Update to use `l('be.greeting')` rather than `localize`.

test/support/i18n.js
alexkrolick: Tried to add it as a command but it turns out cy commands aren't functions you can just use inline, they need to be chainable

alexkrolick: Also there doesn't appear to be a way to set a global other than setting a value on window

test/support/index.js
```diff
@@ -14,3 +14,4 @@
 // ***********************************************************
 
 import './commands';
+import './i18n';
```
jstoffan: Is this needed if we're going to import it in each test file?

PR #751, opened 2019-01-09
chore: add feature flag provider

alexkrolick:
Sets up the feature flag provider, types, and examples


examples/src/features.js
```diff
@@ -0,0 +1,4 @@
+// feature config for feature flags
+export default {
+    tasks: true,
```
DanDeMicco: Can we have this match the existing prop style like has*, maybe `hasNewTasks`

alexkrolick: can just remove for now, I'll update the tasks PR if we want to use that convention

examples/src/withConfig.js
```diff
@@ -10,12 +11,13 @@ function withConfig(WrappedComponent) {
...
             return (
                 <WrappedComponent
-                    {...this.props}
+                    features={features}
                     fileId={FILE_ID}
                     language="en-US"
                     messages={messages}
                     rootFolderId={FOLDER_ID}
                     token={TOKEN}
+                    {...this.props}
```
jstoffan: Do all of the examples still work after this change?

alexkrolick: yes, none of them actually pass props

alexkrolick:
changing to add a `withFeatureProvider` to simplify & avoid the conflicts resulting from adding the component to the render

src/components/FeatureChecking/withFeatureProvider.js
```diff
@@ -0,0 +1,23 @@
+// @flow
+import * as React from 'react';
+import FeatureProvider from './FeatureProvider';
+import type { FeatureConfig } from './flowTypes';
+
+type WrapperProps = {
+    features: FeatureConfig,
+};
+
+function withFeatureProvider(WrappedComponent: React.ComponentType<any>) {
```
DanDeMicco: Can you add a test, maybe snapshot?

src/components/FeatureChecking/withFeatureProvider.js
```diff
@@ -0,0 +1,23 @@
...
+                <WrappedComponent {...props} />
+            </FeatureProvider>
+        );
+    }
+    const wrappedName = WrappedComponent.displayName || WrappedComponent.name || '';
+    WrapperComponent.displayName = `withFeatureProvider(${wrappedName})`;
+    return WrapperComponent;
+}
+
+export default withFeatureProvider;
```
jstoffan: Would this be more accurately named `withFeaturesProvider`, plural, since it injects the `features` prop?

src/components/FeatureChecking/withFeatureProvider.js
```diff
@@ -0,0 +1,23 @@
...
+
+type WrapperProps = {
+    features: FeatureConfig,
+};
+
+function withFeatureProvider(WrappedComponent: React.ComponentType<any>) {
+    function WrapperComponent({ features, ...props }: WrapperProps) {
+        return (
+            <FeatureProvider features={features}>
+                <WrappedComponent {...props} />
```
jstoffan: Does this forward refs as expected?

src/components/ContentPreview/ContentPreview.js
```diff
@@ -1145,4 +1147,4 @@ class ContentPreview extends PureComponent<Props, State> {
 
 export type ContentPreviewProps = Props;
 export { ContentPreview as ContentPreviewComponent };
-export default withErrorBoundary(ORIGIN_CONTENT_PREVIEW)(makeResponsive(ContentPreview));
+export default withErrorBoundary(ORIGIN_CONTENT_PREVIEW)(makeResponsive(withFeatureProvider(ContentPreview)));
```
jstoffan: We may want to consider adding something akin to redux's [compose](https://github.com/reduxjs/redux/blob/master/src/compose.js) utility.

src/components/FeatureChecking/withFeatureProvider.js
alexkrolick: https://github.com/box/box-ui-elements/pull/751/files#diff-198ef77bc61f0032026472fb03a0210b

src/components/FeatureChecking/withFeatureProvider.js
alexkrolick: well... it injects the FeatureProvider which takes a features prop 🤷‍♂️

src/components/ContentPreview/ContentPreview.js
alexkrolick: _.flow would work but it would probably still look weird with such a long line length https://lodash.com/docs/4.17.11#flow

src/components/FeatureChecking/withFeatureProvider.js
alexkrolick: fixed

alexkrolick: looks like the makeResponsive HOC doesn't though

PR #754, opened 2019-01-09
feat(openwith): Box Tools install message is now customizable

jeremypress:


i18n/en-US.properties
```diff
@@ -17,7 +17,7 @@ be.approvalSelectDate = Select a date
 # Mentioning call to action displayed below the comment input
 be.atMentionTip = @mention users to notify them.
 # Message when Box Tools cannot open a particular file type
-be.boxEditBlacklistedError = This file cannot be opened by Box Tools
+be.boxEditBlacklistedError = This file cannot be opened locally
```
jeremypress: Made this more generic since Box Tools can be overwritten in another message.

PR #755, opened 2019-01-09
fix: upgrade react-measure to fix an IE11 issue

priyajeet:
So that we use the latest resize-observer-polyfill which has an issue fixed for IE11
https://github.com/que-etc/resize-observer-polyfill/issues/20


boxcla:

PR #756, opened 2019-01-10
fix: stub out react measure flow typed for now

priyajeet:
Till we can use 
https://github.com/flow-typed/flow-typed/pull/2015/files


boxcla:

PR #762, opened 2019-01-10
fix(openwith): Fix Safari delay by opening new window instead of tab

jeremypress:


boxcla:

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -351,11 +351,17 @@ class ContentOpenWith extends PureComponent<Props, State> {
             // No window management is required when using Box Edit.
             return;
         }
+
+        // These window features will open the new window directly on top of the current window at the same
+        const windowFeatures = `left=${window.screenX},top=${window.screenY},height=${window.outerHeight},width=${
```
jstoffan: Just for fun: ``Object.entries({ left: window.screenX }).map([key, value] => `key=${value}`).join(',');``

PR #774, opened 2019-01-12
fix(metadata): filter out hidden templates

priyajeet:
hidden templates should not be addable


boxcla:

PR #776, opened 2019-01-14
fix(openwith): Add codes to OpenWith errors to fix ErrorBoundary

jeremypress:


boxcla:

src/components/ContentOpenWith/ContentOpenWith.js
```diff
@@ -420,7 +420,6 @@ class ContentOpenWith extends PureComponent<Props, State> {
                 break;
             case HTTP_GET:
                 if (!this.integrationWindow) {
-                    this.executeIntegrationErrorHandler(Error(WINDOW_OPEN_BLOCKED_ERROR));
```
jeremypress: Since we're now responding directly to user input our JS shouldn't see this error anymore.  The second window that adobe create's can still be blocked, but we don't have access to that JS.

PR #778, opened 2019-01-15
fix(uploader): ContentUploader did not properly handle multiput errors

patrickpaul:
The `uploadErrorHandler` would frequently attempt to check properties on an undefined XHR object leading an exception and the part not being properly retried. This PR fixes that.


boxcla:

PR #779, opened 2019-01-15
build(node): Upgrade deps and scripts to support node 10

jstoffan:


boxcla:

package.json
```diff
@@ -154,8 +153,8 @@
         "conventional-changelog": "^2.0.3",
         "conventional-changelog-cli": "^2.0.5",
         "conventional-github-releaser": "^3.1.2",
+        "copyfiles": "^2.1.0",
         "core-js": "^2.5.7",
-        "cpx": "^1.5.0",
```
jstoffan: This library doesn't seem very well-maintained and depends on a version of `chokidar` that is outdated (1.6.0).

PR #782, opened 2019-01-15
chore: Add logger to measure performance metrics

DanDeMicco:
Builds upon #748


boxcla:

conf/performance-mock.js
```diff
@@ -0,0 +1,3 @@
+window.performance = {
```
jstoffan: Can we move the mock files to a `/conf/mocks` folder or something similar?

flow-typed/box-ui-elements.js
```diff
@@ -688,29 +691,38 @@ type ErrorResponseData = {
...
 
-type ErrorOrigins =
-    | ORIGIN_CONTENT_SIDEBAR
-    | ORIGIN_CONTENT_PREVIEW
-    | ORIGIN_PREVIEW
-    | ORIGIN_DETAILS_SIDEBAR
-    | ORIGIN_ACTIVITY_SIDEBAR
-    | ORIGIN_SKILLS_SIDEBAR
-    | ORIGIN_METADATA_SIDEBAR;
+type ElementsOrigins =
```
jstoffan: This type represents a singular item, so `ElementOrigin` seems a more appropriate name.

flow-typed/box-ui-elements.js
```diff
@@ -688,29 +691,38 @@ type ErrorResponseData = {
...
+    onError: (error: ElementsXhrError | Error, code: string, contextInfo?: Object, origin: ElementsOrigins) => void,
 };
 
 type ElementsErrorCallback = (e: ElementsXhrError, code: string, contextInfo?: Object) => void;
 
 type ClassificationInfo = {
     Box__Security__Classification__Key?: string,
 } & MetadataInstance;
+
+type MetricTypes =
```
jstoffan: This type represents a singular item, so `MetricType` seems a more appropriate name.

package.json
```diff
@@ -111,6 +111,7 @@
         "coverageDirectory": "<rootDir>/reports",
         "collectCoverageFrom": ["src/**/*.js", "!**/node_modules/**", "!**/__tests__/**"],
         "roots": ["src"],
+        "setupFiles": ["<rootDir>/conf/performance-mock.js"],
```
jstoffan: How are the existing mocks injected? Can we standardize?

src/components/ContentPreview/ContentPreview.js
```diff
@@ -228,6 +235,13 @@ class ContentPreview extends PureComponent<Props, State> {
             currentFileId: fileId,
             prevFileIdProp: fileId,
         };
+        this.props.onMetric(
+            METRIC_TYPE_ELEMENTS_LOAD_METRIC,
+            {
+                endMarkName: MARK_NAME_JS_READY,
+            },
+            EVENT_JS_READY,
```
jstoffan: It looks like `EVENT_JS_READY` is included in two places: as a suffix in `endMarkName` and here at the top level. Is there a way to avoid the duplication?

src/components/ContentPreview/ContentPreview.js
```diff
@@ -228,6 +235,13 @@ class ContentPreview extends PureComponent<Props, State> {
             currentFileId: fileId,
             prevFileIdProp: fileId,
         };
+        this.props.onMetric(
+            METRIC_TYPE_ELEMENTS_LOAD_METRIC,
```
jstoffan: Can we standardize on `load` or `js_ready`? It looks like we're using them to mean the same thing. It seems like `load` makes the most sense. What do you think?

src/components/ContentPreview/ContentPreview.js
```diff
@@ -538,7 +552,7 @@ class ContentPreview extends PureComponent<Props, State> {
             };
         }
 
-        onMetric(metrics);
+        onMetric(METRIC_TYPE_PREVIEW_METRIC, metrics);
```
jstoffan: Is this considered a breaking change, as the function params are changing? Also, it looks like we include `event_name: PREVIEW_LOAD_METRIC_EVENT` as a property of `metrics`. Should we use `event_name` for the new metrics, as well?

src/components/ContentSidebar/ActivitySidebar.js
```diff
@@ -56,9 +59,30 @@ export const activityFeedInlineError: Errors = {
...
+    };
+
     state = {};
 
+    constructor(props: Props) {
+        super(props);
+        this.props.onMetric(
+            METRIC_TYPE_ELEMENTS_LOAD_METRIC,
+            {
+                startMarkName: null, // TODO: replace with actual start mark once code splitting implemented
```
jstoffan: Where will the new start mark come from? If it's from the parent component, should we instead pass down a callback to this component and allow the parent to handle decorating the metric payload with a start mark?

src/components/ContentSidebar/ActivitySidebar.js
```diff
@@ -466,4 +491,6 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
 
 export type ActivitySidebarProps = ExternalProps;
 export { ActivitySidebar as ActivitySidebarComponent };
-export default withErrorBoundary(ORIGIN_ACTIVITY_SIDEBAR)(withAPIContext(ActivitySidebar));
+export default withLogger(ORIGIN_ACTIVITY_SIDEBAR)(
```
jstoffan: Seems like it may be worthwhile to use `_.flow` here, as well.

src/components/ContentSidebar/ContentSidebar.js
```diff
@@ -128,6 +137,15 @@ class ContentSidebar extends React.PureComponent<Props, State> {
         });
 
         this.state = { isLoading: true, isOpen: !!isLarge };
+        /* eslint-disable react/prop-types */
```
jstoffan: Still necessary to disable eslint? The other components seem to have it enabled.

src/components/logger/Logger.js
```diff
@@ -0,0 +1,200 @@
+// @flow
+import * as React from 'react';
+import noop from 'lodash/noop';
+import { METRIC_TYPE_PREVIEW_METRIC, METRIC_TYPE_ELEMENTS_LOAD_METRIC } from '../../constants';
+
+type ElementsMetric = {
```
jstoffan: It looks like `ElementsMetricCallback` is defined in the top level flowtypes file. Should these type declarations live together?

src/components/logger/Logger.js
```diff
@@ -0,0 +1,200 @@
...
+function uuidv4() {
+    return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, generateRandom);
+}
+
+const SESSION_ID = uuidv4();
+const uniqueEvents: Set<string> = new Set();
+
+class Logger extends React.Component<Props> {
+    static defaultProps = {
+        onMetric: noop,
```
jstoffan: I saw this defined in a couple other components, as well. Is it needed in both places?

src/components/ContentPreview/ContentPreview.js
DanDeMicco: Hmm.. the `Logger` can potentially inject multiple `on*` props. Maybe `onLoad` which can take `endMarkName` as a param, and automatically include `EVENT_JS_READY`. Thoughts?

jstoffan: Yeah, I like that idea. It seems reasonable to use helper methods for commonly-logged metrics.

package.json
jstoffan: Sounds good. Having all the mocks in one folder would be great.

src/components/ContentSidebar/ActivitySidebar.js
DanDeMicco: The start mark will be in Sidebar.js. I can follow up with that in the code splitting PR since this data isn't really relevant at this point

src/elements/common/logger/Logger.js
```diff
@@ -0,0 +1,200 @@
...
+    /**
+     * Logs a unique metric event. Prevents duplicate events from being logged in the session.
+     *
+     * @param {string} type - the type of the event
+     * @param {Object} data  - the event data
+     * @param {string} name - the name of the event
+     */
+    logUniqueMetric(type: MetricTypes, name: string, data: Object): void {
+        let eventName;
+        if (type !== METRIC_TYPE_ELEMENTS_LOAD_METRIC) {
```
JustinHoldstock: Should this instead be referred to as a JS load metric? We may want to refer to events with API load times as load metrics (or not, just a thought)

src/elements/common/logger/withLogger.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * @flow
+ * @file Wraps a component with the Logger context
```
jstoffan: It doesn't look like we're using context anymore. Should this be updated?

src/elements/common/logger/Logger.js
```diff
@@ -0,0 +1,200 @@
...
+            if (!this.props.fileId) {
+                return;
+            }
+            eventName = this.createFileEventNameWith(name, this.props.fileId);
```
jstoffan: Can this logic be simplified and made less aware of specific metric types? Is the file id important/needed for the load metric?

src/elements/common/logger/Logger.js
```diff
@@ -0,0 +1,200 @@
...
+     */
+    logTimeMetric(type: MetricTypes, data: Object, name: string, isUnique: boolean): void {
+        if (!name) {
+            return;
+        }
+
+        if (isUnique) {
+            this.logUniqueMetric(type, name, data);
+        } else {
+            this.logMetric(type, name, data);
```
jstoffan: Do we have a use case for this logic right now? Or the other path?

src/elements/common/logger/Logger.js
```diff
@@ -0,0 +1,200 @@
...
+    /**
+     * Log a timing related metric.
+     *
+     * @param {string} type - the type of the event
+     * @param {Object} data  - the event data
+     * @param {string} name - the name of the event
+     * @param {boolean} isUnique  - true if the metric should be logged only once
+     * @returns {void}
+     */
+    logTimeMetric(type: MetricTypes, data: Object, name: string, isUnique: boolean): void {
```
jstoffan: If `name` is the only param required, should it be the first param in the list?

src/components/ContentSidebar/ContentSidebar.js
DanDeMicco: Yes for some reason it complains here. We actually have a check during the precommit which checks for unused eslint-disable's

DanDeMicco: IMO we should disable this rule for elements folder, as we are not using React.PropTypes

src/components/logger/Logger.js
DanDeMicco: It could, but I put it here to avoid bloating the already large `box-ui-elements.js` since its used only in this file (for now), similar to how we use `constants.js`

src/components/logger/Logger.js
DanDeMicco: I can remove it from the components, will do that when i do the tests

conf/performance-mock.js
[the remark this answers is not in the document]
DanDeMicco: It depends on usage. For my case, `window.performance.mark` is used outside of the component so I needed to set it up early on. If you are stubbing out window methods and you can do it in the beforeEach that would probably be better, using jest.spyOn

src/elements/common/logger/Logger.js
```diff
@@ -0,0 +1,200 @@
...
+
+        this.props.onMetric(metric);
+    }
+
+    /**
+     * Logs a unique metric event. Prevents duplicate events from being logged in the session.
+     *
+     * @param {string} type - the type of the event
+     * @param {Object} data  - the event data
+     * @param {string} name - the name of the event
```
JustinHoldstock: Swap `data` and `name` param docs

src/elements/common/logger/Logger.js
```diff
@@ -0,0 +1,200 @@
...
+    hasLoggedEvent(name: string): boolean {
+        return this.uniqueEvents.has(name);
+    }
+
+    /**
+     * Invokes the provided metric logging callback.
+     *
+     * @param {string} type - the type of the event
+     * @param {Object} data  - the event data
+     * @param {string} name - the name of the event
```
JustinHoldstock: Swap data and name param docs

flow-typed/box-ui-elements.js
```diff
@@ -688,29 +691,38 @@ type ErrorResponseData = {
...
-type ErrorOrigins =
-    | ORIGIN_CONTENT_SIDEBAR
-    | ORIGIN_CONTENT_PREVIEW
-    | ORIGIN_PREVIEW
-    | ORIGIN_DETAILS_SIDEBAR
-    | ORIGIN_ACTIVITY_SIDEBAR
-    | ORIGIN_SKILLS_SIDEBAR
-    | ORIGIN_METADATA_SIDEBAR;
+type ElementsOrigins =
+    | typeof ORIGIN_CONTENT_SIDEBAR
```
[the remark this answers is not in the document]
DanDeMicco: Normally yes, but this is a workaround needed to get enums working with flow. See the definition ```export const ORIGIN_CONTENT_SIDEBAR: 'content_sidebar' = 'content_sidebar';```

src/elements/common/logger/Logger.js
```diff
@@ -0,0 +1,200 @@
...
+     * @returns {void}
+     */
+    handleMetric = (type: MetricTypes, data: Object, name?: string, isUnique?: boolean = true) => {
+        if (type === METRIC_TYPE_PREVIEW_METRIC) {
+            this.props.onMetric({
+                ...data,
+                type,
+            });
+        } else if (name) {
+            this.logTimeMetric(type, data, name, isUnique);
```
JustinHoldstock: Might want to standardize on either `type, data, name, ...` params ordering, or `type, name, data, ...` ordering, across this class

src/elements/common/logger/withLogger.js
DanDeMicco: 👍

flow-typed/box-ui-elements.js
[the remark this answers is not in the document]
jeremypress: cool

src/elements/common/logger/Logger.js
DanDeMicco: Removing this for now

src/elements/common/logger/Logger.js
DanDeMicco: The event name can be used for this purpose

DanDeMicco:
Tests are going to fail on this latest commit, but I resolved most comments

src/elements/content-preview/ContentPreview.js
```diff
@@ -538,7 +544,7 @@ class ContentPreview extends PureComponent<Props, State> {
             };
         }
 
-        onMetric(metrics);
+        this.props.logger.onPreviewMetric(metrics);
```
priyajeet: We should be moving to de-structuring (even for single), thats the eslint rule we need to enable in the near future. Currently its turned off but we shouldn't add more.

https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/destructuring-assignment.md

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -56,9 +60,21 @@ export const activityFeedInlineError: Errors = {
         content: messages.activityFeedItemApiError,
     },
 };
+
+const MARK_NAME_JS_READY = `${ORIGIN_ACTIVITY_SIDEBAR}_${EVENT_JS_READY}`;
+
+window.performance.mark(MARK_NAME_JS_READY);
```
priyajeet: same comment as before

src/elements/content-sidebar/Sidebar.js
```diff
@@ -9,11 +9,16 @@ import DetailsSidebar from './DetailsSidebar';
 import SkillsSidebar from './SkillsSidebar';
 import ActivitySidebar from './ActivitySidebar';
 import MetadataSidebar from './MetadataSidebar';
+import { mark } from '../../utils/performance';
 import {
     SIDEBAR_VIEW_SKILLS,
     SIDEBAR_VIEW_ACTIVITY,
     SIDEBAR_VIEW_DETAILS,
     SIDEBAR_VIEW_METADATA,
+    ORIGIN_DETAILS_SIDEBAR,
```
jstoffan: Should these follow the `SIDEBAR_ORIGIN_SKILLS` convention used above for views?

src/elements/common/logger/Logger.js
```diff
@@ -0,0 +1,177 @@
...
+    startMarkName?: string,
+};
+
+/**
+ * Converts a character, using it as a seed, to a random integer and returns it as a string.
+ *
+ * @param {string} c - A single character
+ * @returns {string} The random character
+ */
+function generateRandom(c: string): string {
```
priyajeet: These functions should be inside their own utilities.
Also https://github.com/kelektiv/node-uuid

src/elements/common/logger/Logger.js
```diff
@@ -0,0 +1,177 @@
...
+    return v.toString(16);
+    /* eslint-enable no-bitwise */
+}
+
+/**
+ * Generates a GUID/UUID compliant with RFC4122 version 4.
+ *
+ * @return {string} A 36 character uuid
+ */
+function uuidv4() {
```
priyajeet: https://gist.github.com/jed/982883

src/elements/common/logger/Logger.js
```diff
@@ -0,0 +1,177 @@
...
+        const metric: ElementsMetric = {
+            ...data,
+            component: this.props.source,
+            name,
+            timestamp: this.getTimestamp(),
+            sessionId: this.sessionId,
+            type,
+        };
+
+        this.props.onMetric(metric);
```
priyajeet: 🙄

src/elements/common/logger/Logger.js
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    /**
+     * Preview metric handler
+     *
+     * @param {Object} data - the metric data
+     * @returns {void}
+     */
+    handlePreviewMetric = (data: Object) => {
+        this.props.onMetric({
```
priyajeet: const { onMetric } = this.props

src/elements/common/logger/Logger.js
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    /**
+     * Creates an event name meant for use with an event which is unique and meant to be logged only once
+     *
+     * @param {string} name - The event name
+     * @returns {string} A string containing the component and event name
+     */
+    createEventName(name: string): string {
+        return `${this.props.source}::${name}`;
```
priyajeet: const { source } = this.props

src/elements/common/logger/Logger.js
```diff
@@ -0,0 +1,177 @@
...
+ * Generates a GUID/UUID compliant with RFC4122 version 4.
+ *
+ * @return {string} A 36 character uuid
+ */
+function uuidv4() {
+    return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, generateRandom);
+}
+
+const SESSION_ID = uuidv4();
+const uniqueEvents: Set<string> = new Set();
```
priyajeet: IE11?

src/elements/common/logger/Logger.js
```diff
@@ -0,0 +1,177 @@
...
+    };
+
+    /**
+     * JS ready metric handler
+     *
+     * @param {Object} data - the metric data
+     * @returns {void}
+     */
+    handleReadyMetric = (data: ElementsLoadMetricData) => {
+        const { startMarkName } = this.props;
```
priyajeet: :relieved:

src/elements/common/logger/Logger.js
priyajeet: I guess it allows - add, has

src/elements/common/logger/Logger.js
DanDeMicco: Using uuid

PR #784, opened 2019-01-15
refactor: add module path aliasing to reduce relative imports

priyajeet:


jsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+    "compilerOptions": {
```
priyajeet: For VSCode intellisense

package.json
```diff
@@ -166,6 +167,7 @@
         "eslint": "^4.19.1",
         "eslint-config-airbnb": "^17.1.0",
         "eslint-config-prettier": "^3.0.1",
+        "eslint-import-resolver-babel-module": "^5.0.0-beta.1",
```
priyajeet: beta :/

babel.config.js
```diff
@@ -21,6 +21,21 @@ module.exports = {
                 messagesDir: './i18n/json',
             },
         ],
+        [
+            'module-resolver',
+            {
+                alias: {
```
jstoffan: Does the following result in the same paths?

```
'module-resolver', {
    root: ["./src"]
}
```

PR #785, opened 2019-01-15
chore: Fix localStore unit test to work with node 8

jstoffan:
It looks like the latest versions of Jest have some issues with reassigning global/window methods. I've added a workaround for now that seems to work correctly in both node 8 and node 10.


boxcla:

PR #787, opened 2019-01-15
build: make node 10 min version

priyajeet:
BREAKING: makes min required node version to build


boxcla:

PR #788, opened 2019-01-16
chore(openwith): Example integration test with Cypress

jeremypress:
I'll be adding many more tomorrow but I wanted to discuss naming/ best practices for structuring fixtures and testids

### DIFF
--- a/test/fixtures/open_with_integrations/multipleIntegrations.json
+++ b/test/fixtures/open_with_integrations/multipleIntegrations.json
@@ -1,0 +1,40 @@
+{
+    "items": [
+        {
+            "app_integration": {
+                "type": "app_integration",
+                "id": "13418"
+            },
+            "should_show_consent_popup": true,
+            "disabled_reasons": [],
+            "is_disabled": false,
+            "display_order": 1,
+            "display_name": "Open",
+            "display_description": "Open this file on your computer"
+        },
+        {
+            "app_integration": {
+                "type": "app_integration",
+                "id": "10897"
+            },
+            "should_show_consent_popup": true,
+            "disabled_reasons": [],
+            "is_disabled": false,
+            "display_order": 2,
+            "display_name": "Google Docs (beta)",
+            "display_description": "Open this file in your browser"
+        },
+        {
+            "app_integration": {
+                "type": "app_integration",
+                "id": "3282"
+            },
+            "should_show_consent_popup": true,
+            "disabled_reasons": [],
+            "is_disabled": false,
+            "display_order": 3,
+            "display_name": "Adobe Sign",
+            "display_description": "Send this document for signature"
+        }
+    ]
+}
--- a/test/integration/open-with/OpenWith.int.js
+++ b/test/integration/open-with/OpenWith.int.js
@@ -1,0 +1,22 @@
+// <reference types="Cypress" />
+
+describe('OpenWith', () => {
+    beforeEach(() => {
+        cy.server();
+    });
+
+    it('integrations will render when the dropdown is clicked', () => {
+        cy.route(
+            'GET',
+            '**/files/*/open_with_integrations',
+            'fixture:open_with_integrations/multipleIntegrations.json',
+        );
+
+        cy.visit('/ContentOpenWith');
+
+        // // Click the Open With button
+        cy.getByTestId('multipleintegrationsbutton').click();
+        // // The dropdown should render
+        cy.getByTestId('openwithdropdown').should('exist');
+    });
+});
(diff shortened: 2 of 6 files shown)


boxcla:

test/fixtures/open_with_integrations/multipleIntegrations.json
```diff
@@ -0,0 +1,40 @@
+{
```
jstoffan: I think you'll want the file/folder to be named something like `open-with/integrations-multiple.json`.

test/integration/open-with/OpenWith.int.js
```diff
@@ -0,0 +1,22 @@
...
+    });
+
+    it('integrations will render when the dropdown is clicked', () => {
+        cy.route(
+            'GET',
+            '**/files/*/open_with_integrations',
+            'fixture:open_with_integrations/multipleIntegrations.json',
+        );
+
+        cy.visit('/ContentOpenWith');
```
jstoffan: Consider moving this to `beforeEach` as it will reload the page for each test, giving you a fresh state.

test/integration/open-with/OpenWith.int.js
```diff
@@ -0,0 +1,22 @@
...
+
+        // // Click the Open With button
+        cy.getByTestId('multipleintegrationsbutton').click();
+        // // The dropdown should render
+        cy.getByTestId('openwithdropdown').should('exist');
```
jstoffan: It might be more meaningful to test the content of the dropdown rather than the dropdown itself via a testid.

test/integration/open-with/OpenWith.int.js
```diff
@@ -0,0 +1,22 @@
+// <reference types="Cypress" />
+
+describe('OpenWith', () => {
+    beforeEach(() => {
+        cy.server();
+    });
+
+    it('integrations will render when the dropdown is clicked', () => {
+        cy.route(
```
jstoffan: If other tests are going to reuse this data, consider moving it to `before` or `beforeEach`, instead.

test/fixtures/open_with_integrations/multipleIntegrations.json
jeremypress: agreed, it's not worth trying to keep one folder per endpoint, replicating them isn't a huge deal.

test/integration/open-with/OpenWith.int.js
jeremypress: Other tests will be routing with different responses. But if I write more with this data I will use a local before each to that suite.

test/integration/open-with/OpenWith.int.js
jeremypress: If you `visit` before you `route` the request will actually make it to the server. Since we're dynamically routing per test case I think we have to visit in each case.

test/integration/open-with/OpenWith.int.js
jeremypress: I've heard that using css selectors is an antipattern with cypress. Since these menu items are dynamically generated, do you have any recommendations for selecting content?

alexkrolick: `cy.contains(l("translation.id.xyz"))`

Using the i18n util to translate a content string

jeremypress: I'll check the strings directly

test/integration/open-with/OpenWith.int.js
```diff
@@ -0,0 +1,142 @@
...
+        cy.server();
+        cy.route('GET', '**/files/*?fields=extension', 'fixture:open-with/file-extension-document');
+    });
+
+    it('A single integration', () => {
+        cy.route('GET', '**/files/*/open_with_integrations', 'fixture:open-with/integration-adobe');
+        cy.visit('/ContentOpenWith');
+
+        // The button should be enabled
+        cy.getByTestId('singleintegrationbutton').should('not.have.attr', 'aria-disabled', 'true');
```
ConradJChan: consider adding constants for all these test ids that get used in many places

test/integration/open-with/OpenWith.int.js
```diff
@@ -0,0 +1,142 @@
...
+        cy.get('@menuItem')
+            .eq(2)
+            .contains('Adobe Sign');
+        cy.get('@menuItem')
+            .eq(3)
+            .contains('My custom integration');
+    });
+
+    describe('box edit', () => {
+        it('box edit is uninstalled', () => {
```
jstoffan: Having `describe('box-edit')` and `it('box-edit')` will show some repetitive text in the test descriptions.

test/integration/open-with/OpenWith.int.js
jstoffan: I think it's actually more efficient to use helper methods, like: `const getSingleButton = () => cy.getByTestId('singleintegrationbutton');`

ConradJChan: yup that makes more sense

PR #789, opened 2019-01-16
fix(build): revert back to allowing even node 8 temporarily

priyajeet:


boxcla:

PR #793, opened 2019-01-16
feat(preview): add thumbnails sidebar

ConradJChan:
Putting this up as more of a RFC than the actual PR to merge in.

The issue I'm trying to solve is that `ContentPreview` provides its own navigation buttons which exist outside of the underlying Preview SDK layout, so when thumbnails sidebar is toggled open, the `ContentPreview` navigation buttons are unaffected and the left navigation overlaps with the thumbnails sidebar.

thumbnail overlap

The approach here is to:

1. pass the `collection` into PreviewSDK
2. rely on the Preview SDK's native navigation buttons
3. rely on the `navigate` event that is emitted from the Preview SDK to plug back into the `ContentPreview` navigation logic

This requires a corresponding change in Preview SDK to accept an option to disable the native collection navigation which feels a bit awkward.

Another approach to solve this problem is to offset the absolute position of the left navigation arrow whenever the thumbnail sidebar is toggled open, however this solution is not ideal in the case where we support resizing of the thumbnail sidebar (which may not happen).

Thoughts?

### DIFF
--- a/src/elements/content-preview/ContentPreview.js
+++ b/src/elements/content-preview/ContentPreview.js
@@ -16,7 +16,6 @@ import noop from 'lodash/noop';
 import Measure from 'react-measure';
 import { decode } from 'box-react-ui/lib/utils/keys';
 import PreviewLoading from './PreviewLoading';
-import PreviewNavigation from './PreviewNavigation';
 import ContentSidebar from '../content-sidebar';
 import Header from './Header';
 import API from '../../api';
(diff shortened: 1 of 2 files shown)


boxcla:

DanDeMicco:
One potential issue i can see here is that if we rely on the preview UI showing the arrows, they may disappear upon clicking to the next file as this will destroy preview, which should in theory destroy the UI

ConradJChan:
Since resizing is no longer a requirement, I'll go with a simpler approach of listening to the thumbnail open or close viewer event and just absolutely position the left navigation button

PR #798, opened 2019-01-18
fix(preview): Disable annotations in preview by default

pramodsum:


boxcla:

PR #800, opened 2019-01-18
chore(openwith): trigger error callback if Box Tools fails to open

jeremypress:


boxcla:

PR #802, opened 2019-01-18
fix(openwith): Completely reset state on component update

jeremypress:


boxcla:

src/elements/content-open-with/ContentOpenWith.js
```diff
@@ -192,7 +192,7 @@ class ContentOpenWith extends PureComponent<Props, State> {
         }
 
         if (currentFileId !== previousFileId) {
-            this.setState({ isLoading: true, integrations: null });
+            this.setState({ ...this.initialState });
```
jstoffan: https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops

> If you want to “reset” some state when a prop changes, consider either making a component fully controlled or fully uncontrolled with a key instead.

jeremypress: Good idea, since I'm already double rendering I'll merge this fix in for now and work on a permanent solution next week.

PR #803, opened 2019-01-18
fix(uploader): Handle errors with calling logging endpoint

patrickpaul:
Fixes 2 issues that occur in error handling:

1) If the original part failure is due to 401, the logEvent will also fail with an uncaught promise.
2) Aborting a session can cause multiple DELETE calls to the session endpoint which will result in 404s.

Additionally the start of some tech debt cleanup for cases when state is modified directly.


boxcla:

PR #805, opened 2019-01-22
feat(preview): Enable thumbnails sidebar in Content Preview

ConradJChan:


boxcla:

src/elements/content-preview/ContentPreview.js
```diff
@@ -682,12 +711,14 @@ class ContentPreview extends PureComponent<Props, State> {
...
             skipServerUpdate: true,
             useHotkeys: false,
+            enableThumbnailsSidebar,
         };
         const { Preview } = global.Box;
         this.preview = new Preview();
         this.preview.addListener('load', this.onPreviewLoad);
         this.preview.addListener('preview_error', this.onPreviewError);
         this.preview.addListener('preview_metric', this.onPreviewMetric);
+        this.preview.addListener('viewerevent', this.onPreviewViewerEvent);
```
jeremypress: You should be able to add these at the top level if you want.
this.preview.addListener('thumbnailsOpen', this.onThumbnailsOpen); etc...

src/elements/content-preview/ContentPreview.js
```diff
@@ -550,6 +560,19 @@ class ContentPreview extends PureComponent<Props, State> {
         logger.onPreviewMetric(metrics);
     };
 
+    /**
+     * Event handler for viewer events from Preview. Currently only handles thumbnails open and close
+     * @param {Object} viewerEvent - the viewer event object
+     * @return {void}
+     */
+    onPreviewViewerEvent = ({ event: eventName }: PreviewViewerEvent): void => {
+        if (eventName === 'thumbnailsOpen') {
```
jeremypress: Probably worth converting this to a switch to make future additions easier if you want to continue binding to 'viewerevent'. Also probably worth a constants list in this file of preview events we're using.

src/elements/content-preview/ContentPreview.js
ConradJChan: I'll do that, seems better than adding the `onPreviewerEvent` for just these two

src/elements/content-preview/ContentPreview.js
ConradJChan: will get rid of this per your suggestion

src/elements/content-preview/ContentPreview.js
```diff
@@ -91,6 +93,7 @@ type State = {
     isReloadNotificationVisible: boolean,
     currentFileId?: string, // the currently displayed file id in the collection
     prevFileIdProp?: string, // the previous value of the "fileId" prop. Needed to implement getDerivedStateFromProps
+    isThumbnailsOpen: boolean,
```
jeremypress: nit: I think technically this should be `areThumbnailsOpen`. If you want to keep the is pattern maybe `isThumbnailSidebarOpen`?

PR #809, opened 2019-01-24
fix(sidebar): Scroll to end of input when there is overflow

jeremypress:
Currently, when there's overflow on the assignee input, entering a new assignee does not scroll to the bottom of our input wrapper. This is because the underlying component preventsDefault on enter, to prevent a submission if this component is used in a form.


boxcla:

src/elements/content-sidebar/activity-feed/approval-comment-form/ApprovalCommentForm.js
```diff
@@ -188,7 +188,12 @@ class ApprovalCommentForm extends React.Component<Props, State> {
     };
 
     handleApproverSelectorSelect = (pills: any): void => {
-        this.setState({ approvers: this.state.approvers.concat(pills) });
+        this.setState({ approvers: this.state.approvers.concat(pills) }, () => {
+            const input = document.querySelector('.pill-selector-input-wrapper');
```
jstoffan: Can we get to this via a `ref`? If not, can we query the DOM in `componentDidMount` and store the reference as a class property?

jeremypress: We would need to pass the ref down 2 levels in Elements and 2 in BRUI to get to the component we care about. 

When this component mounts, the assignee input is not visible because the user hasn't clicked the "add task" button yet. I could move this logic lower down, but then we would have to override the onSelectorInput callback at a lower level.

PR #813, opened 2019-01-25
fix(explorer): Footer navigation no longer wraps on mobile devices

jstoffan:
We've decided to move forward with a minimum width of 300px for everything except ContentSidebar. 

**ContentPicker Desktop**
new desktop

**ContentPicker Mobile**
new mobile


boxcla:

PR #817, opened 2019-01-29
chore: remove dependency on box-react-ui

priyajeet:
reference local components instead


boxcla:

flow-typed/box-ui-elements.js
```diff
@@ -275,11 +318,33 @@ type MetadataInstance = {
...
+        | typeof JSON_PATCH_OP_REMOVE
+        | typeof JSON_PATCH_OP_REPLACE
+        | typeof JSON_PATCH_OP_TEST,
+    path: string,
+    value?: any,
+};
+
+type JSONPatchOperations = Array<JSONPatch>;
+
+
```
jstoffan: Something wild happened with the return key here.

src/elements/common/base.scss
```diff
@@ -1,19 +1,19 @@
 @import './variables';
-@import '~box-react-ui/lib/styles/common/buttons';
-@import '~box-react-ui/lib/styles/common/inline-notifications';
-@import '~box-react-ui/lib/styles/modifiers/accessibility';
-@import '~box-react-ui/lib/styles/common/overlay';
-@import '~box-react-ui/lib/styles/common/pills';
-@import '~box-react-ui/lib/components/pill-selector-dropdown/PillSelector';
-@import '~box-react-ui/lib/components/notification/Notification';
+@import '../../styles/common/buttons';
```
jstoffan: Seems like this should be `@import './buttons';`, along with anything else that's an immediate neighbor.

src/elements/common/base.scss
```diff
@@ -1,19 +1,19 @@
...
+@import '../../styles/common/inline-notifications';
+@import '../../styles/modifiers/accessibility';
```
jstoffan: I'm not sure if there's a convention today, but having a common import order would be helpful. Something like:

1. variables
1. common rules/helpers
1. components
1. modifiers
1. overrides

src/elements/common/sub-header/Sort.js
```diff
@@ -6,12 +6,12 @@
 
 import React from 'react';
 import { FormattedMessage } from 'react-intl';
-import Button from 'box-react-ui/lib/components/button/Button';
-import DropdownMenu from 'box-react-ui/lib/components/dropdown-menu/DropdownMenu';
-import Menu from 'box-react-ui/lib/components/menu/Menu';
-import MenuItem from 'box-react-ui/lib/components/menu/MenuItem';
-import IconCheck from 'box-react-ui/lib/icons/general/IconCheck';
-import IconSort from 'box-react-ui/lib/icons/general/IconSort';
+import Button from 'components/button/Button';
```
jstoffan: Can we also update these to import from index.js, such as `import Button from 'components/button';`? It looks like some files already do so.

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
```diff
@@ -116,7 +116,7 @@ class TaskForm extends React.Component<Props, State> {
         this.clearForm();
     };
 
-    handleDueDateChange = (date: ?Date): void => {
+    handleDueDateChange = (date?: Date): void => {
```
jstoffan: Does this clear the state as expected when the user removes the date from the form field?

alexkrolick: No, it should be a nullable value

priyajeet: DatePicker.js cannot take a null value btw per its flow declaration.

src/elements/common/base.scss
priyajeet: Can do this later

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
jstoffan: @priyajeet, that's fine as long as it actually clears the state when using `undefined` rather than `null`. Does the component still continue to work the same way with this change?

src/elements/common/base.scss
priyajeet: This is inside the elements folder, not next to buttons?

src/elements/common/sub-header/Sort.js
priyajeet: can do it in a followup commit, but we can for most iirc

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
priyajeet: Diff seems outdated, check the new diff

PR #819, opened 2019-01-29
build: remove building examples temporarily

priyajeet:
will enable it back soon


boxcla:

PR #822, opened 2019-01-30
chore(ide): fix vscode intellisense

alexkrolick:
https://code.visualstudio.com/docs/languages/jsconfig#_best-practices

Intellisense is currently disabled because the TS server detects too many files with the new settings


boxcla:

PR #834, opened 2019-02-02
feat(features): allow inverting feature flag

alexkrolick:


boxcla:

src/elements/common/feature-checking/__tests__/FeatureChecking.test.js
```diff
@@ -76,6 +76,23 @@ describe('FeatureFlag', () => {
...
+        const enabledFn = jest.fn(() => null);
+        const disabledFn = jest.fn(() => null);
+        const foo = { otherProp: 'foo' };
+        mount(
+            <FeatureProvider
+                features={{
+                    foo,
+                }}
+            >
+                <FeatureFlag disabled={disabledFn} enabled={enabledFn} not feature="foo" />
```
jstoffan: I find this a bit confusing, personally. Why is inverting the flag necessary? If we have `enabled` and `disabled` it seems like we cover both conditions.

priyajeet: +2, adding a `not` makes it rather confusing.

priyajeet: And if we need a use case where you have to render something based on if something is not ON, can we just have a component that does it. `<InvertedFeatureFlag>` or a better name, though that's only a little bit less confusing.

alexkrolick: Here's an example: https://github.com/box/box-ui-elements/compare/master...alexkrolick:feat/tasks/new-api#diff-a844e542fc9fc2f6103b82dd5410fb75R69

I was passing `enabled` and `disabled` as inline functions, but got a request to refactor as a class method on the previous PR. That would require rewriting the function component as a class component. Adding this prop allows the use of children instead of functions while negating the feature.

alexkrolick: Added test for children prop https://github.com/box/box-ui-elements/pull/834/files#diff-a73cef8be86971e9d516ffd86e661da4R79

src/elements/common/feature-checking/__tests__/FeatureChecking.test.js
```diff
@@ -76,6 +76,46 @@ describe('FeatureFlag', () => {
...
+        mount(
+            <FeatureProvider
+                features={{
+                    foo,
+                }}
+            >
+                <FeatureFlag feature="foo">
+                    <ChildComponentEnabled />
+                </FeatureFlag>
+                <FeatureFlag>
```
jstoffan: Did you want to use `not` and `feature="foo"` here, as well?

src/elements/common/feature-checking/__tests__/FeatureChecking.test.js
```diff
@@ -76,6 +76,46 @@ describe('FeatureFlag', () => {
...
+                    <ChildComponentDisabled />
+                </FeatureFlag>
+            </FeatureProvider>,
+        );
+        expect(ChildComponentEnabled).toHaveBeenCalled();
+        expect(ChildComponentDisabled).not.toHaveBeenCalled();
+    });
+
+    test('"not" prop inverts flag with enabled/disable', () => {
+        // NOTE: "not" is recommended for use with a single child, not "enabled"/"disabled"
```
jstoffan: Seems like this comment may be better served in the component itself.

src/elements/common/feature-checking/__tests__/FeatureChecking.test.js
```diff
@@ -76,6 +76,46 @@ describe('FeatureFlag', () => {
         expect(enabledFn).toHaveBeenCalledWith(foo);
         expect(disabledFn).toHaveBeenCalled();
     });
+    test('"not" prop inverts flag with children prop', () => {
```
jstoffan: Can we add a test for the explicit `features={{ foo: false }}` and `not feature="foo"` case?

src/elements/common/feature-checking/FeatureFlag.js
```diff
@@ -9,16 +9,18 @@ function FeatureFlag({
...
     disabled?: () => React.Node,
     enabled?: types.FeatureOptions => React.Node,
     feature: string,
+    not: boolean,
 }) {
     return (
         <FeatureConsumer>
             {features => {
-                const isEnabled = isFeatureEnabled(features, feature);
+                const isEnabled = !not && isFeatureEnabled(features, feature);
```
jstoffan: Doesn't this mean that the child will never be shown if `not` is `true`, regardless of the feature flag passed in, due to the logical AND? It seems like we'd want something similar to what's below, unless I'm missing something. Double-negatives make my head hurt:

```
const isFlagEnabled = isFeatureEnabled(features, feature);
const isEnabled = not ? !isFlagEnabled : isFlagEnabled;
```

alexkrolick: @jstoffan that's right - fixed and added a test case

PR #835, opened 2019-02-02
chore(format): apply lint sort to flow def

alexkrolick:


boxcla:

PR #837, opened 2019-02-04
chore: latest pull from brui

priyajeet:
Adds a new icon.

BREAKING CHANGE: Updates react-tether peer dependency to version 1.x that uses createPortal and fixes redux provider issues inside tether-ed components.


boxcla:

package.json
```diff
@@ -220,7 +220,7 @@
         "react-modal": "^3.5.1",
         "react-process-string": "^1.2.0",
         "react-styleguidist": "^8.0.6",
-        "react-tether": "^0.6.1",
+        "react-tether": "^1.0.4",
```
j3tan: use fixed version - `1.0.3` because `1.0.4` cause uglifyJS to crash during a webpack prod run

PR #838, opened 2019-02-05
build: enable beta releases for daily usage

priyajeet:
With this change, when running `yarn release` (the way to run release on a daily basis) it will cut a release from `master` and tag it as a pre-release with version like `10.0.0-beta.1` or `10.0.0-beta.2`... For example see https://github.com/priyajeet/box-ui-elements/releases. This release will also be `dist-tag`ed on NPM as `@beta` and users will install it via `yarn add box-ui-elements@beta` which will be what EUA does on a regular basis.

Then once a week (or however long) we cut a release from `release` branch after resetting to master. This will be done via `yarn release:latest` marking it as the `@latest` NPM `dist-tag` and latest release on github for everyone else.


release.config.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-    branch: 'release',
+    branches: ['release', { name: 'master', prerelease: 'beta' }],
```
jstoffan: We should probably keep the `'+([1-9])?(.{+([1-9]),x}).x'` regex to allow [hotfixes for older versions](https://github.com/semantic-release/semantic-release/blob/beta/docs/recipes/maintenance-releases.md#releasing-a-bug-fix-for-version-10x-users).

conf/cut_release.sh
```diff
@@ -24,13 +24,14 @@ setup() {
         printf "${green}Removed existing release remote branch!${end}"
     fi
     printf "${blue}Adding release remote branch...${end}"
-    git remote add release git@github.com:box/box-ui-elements.git || return 1
+    git remote add release git@github.com:priyajeet/box-ui-elements.git || return 1
```
ConradJChan: `priyajeet/box-ui-elements.git`? Leftover from testing? Also in `package.json`

conf/cut_release.sh
```diff
@@ -40,14 +41,14 @@ setup() {
     printf "${green}Fetched from remote release branch!${end}"
 
     # Only proceed if release scripts haven't changed
-    check_release_scripts_changed || return 1
+    # check_release_scripts_changed || return 1
```
ConradJChan: not needed anymore?

conf/cut_release.sh
```diff
@@ -56,10 +57,10 @@ setup() {
     git fetch --prune release '+refs/tags/*:refs/tags/*' || return 1
     printf "${green}Pruned tags!${end}"
 
-    # Clean untracked files
-    printf "${blue}Updating remote release branch...${end}"
-    git push release release --force --no-verify || return 1
-    printf "${green}Updated remote release branch!${end}"
+    # # Clean untracked files
+    # printf "${blue}Updating remote release branch...${end}"
```
ConradJChan: can delete if not needed? Similar commented out lines in `push_new_release`

PR #839, opened 2019-02-06
build(babel): Transpile node_modules for dist builds

jeremypress:
This should prevent un-transpiled node_modules from affecting our dist builds. I tested by pulling in a package that breaks on IE 11 (query-string 6.2.0) and generating the CI dist build. The resulting JS had no arrow functions, which was breaking previously.


boxcla:

conf/webpack.config.js
```diff
@@ -138,6 +139,12 @@ function getConfig(isReactExternalized) {
                     flags: 'g',
                 },
             },
+            {
+                test: /\.(js|mjs)$/,
+                loader: 'babel-loader',
+                include: /(node_modules)/,
```
jstoffan: Seems like this could live directly in `test`.

conf/webpack.config.js
```diff
@@ -138,6 +139,12 @@ function getConfig(isReactExternalized) {
...
+                exclude: excludeFromBabel,
+            },
```
jstoffan: It looks like other internal projects pass some additional configuration. Is that necessary here?

priyajeet: Maybe we should just copy most of https://github.com/facebook/create-react-app/blob/3efcdbf328cff893810a9ed88a4cf4cf838484f5/packages/react-scripts/config/webpack.config.js#L394-L431

priyajeet: Or is the goal, that it uses our babelrc, by not providing babelrc: false?
I am not seeing those options listed here https://github.com/babel/babel-loader

conf/webpack.config.js
alexkrolick: It's conventional to specify the file extensions in `test` and the paths in `include`/`exclude` (through either a regex or `path.resolve()`), but it's not enforced: https://webpack.js.org/configuration/module/#condition

conf/webpack.config.js
alexkrolick: You need to specify `babelrc: false` and `configFile: false` so that the packages are compiled according to the config we control instead of babel settings defined in the package.

It should look like the [link @priyajeet provided](https://github.com/facebook/create-react-app/blob/3efcdbf328cff893810a9ed88a4cf4cf838484f5/packages/react-scripts/config/webpack.config.js#L394-L431). If you don't want to install `babel-preset-react-app/dependencies` you can inline the necessary configuration from the file: 

https://github.com/facebook/create-react-app/blob/master/packages/babel-preset-react-app/dependencies.js

```js
// webpack.config.js
// ...rules...
  {
    test: /\.(js|mjs)$/,
    include: /node_modules/,
    exclude: excludeFromBabel,
    loader: require.resolve('babel-loader'),
    options: {
      babelrc: false,
      cacheCompression: false,
      cacheDirectory: true,
      compact: false,
      configFile: false,
      presets: [
        [
          // 🔻install babel-preset-react-app to use this 🔻
          require.resolve('babel-preset-react-app/dependencies'),
          { helpers: true }
        ]
      ],
      sourceMaps: false,
    },
  },
```

alexkrolick:
@jeremypress deleted the `package-lock.json` file which is created by `npm` since we use `yarn.lock`

conf/webpack.config.js
jstoffan: I don't usually see `include` and `exclude` get used simultaneously, since it can be a little confusing. If it works, though, I guess it's fine.

PR #841, opened 2019-02-08
chore: splits metadata-view into query-bar and list-view

priyajeet:
Also downgrades react-tether to 1.0.1


boxcla:

PR #842, opened 2019-02-08
chore(tasks): create new task api classes

alexkrolick:


boxcla:

src/api/TaskCollaborators.js
```diff
@@ -0,0 +1,130 @@
...
+    }: {
+        errorCallback: ElementsErrorCallback,
+        file: BoxItem,
+        successCallback: Function,
+        task: { id: string },
+        user: { id: string },
+    }): void {
+        this.errorCode = ERROR_CODE_CREATE_TASK_COLLABORATOR;
+
+        const requestData = {
```
priyajeet: Should you check permissions here, from the file object, so that edits aren't allowed if permission is not set?
Likewise for other sections?

src/constants.js
```diff
@@ -6,6 +6,9 @@
 
 import Browser from 'utils/Browser';
 
+/* ------------------------ API ---------------------------- */
+export const API_PAGE_LIMIT = 1000; // default limit for paginated api calls
```
priyajeet: Can you change line 71ish export const METADATA_TEMPLATE_FETCH_LIMIT = ~~1000~~API_PAGE_LIMIT;

src/api/TaskCollaborators.js
alexkrolick: That logic has been moved into the UI code
- we want to prevent this check from being called by disabling UI elements instead of bouncing the request
- we also expect to handle auth errors resulting from server side auth checks for anything that can't be computed beforehand

PR #849, opened 2019-02-11
build: include global flow types, disable react-intl for dev

priyajeet:
* Includes global flow types inside the npm package so that parent projects can reference it.
* Also fixes references and eslint issues within the global flow types file.
* Restricts react-intl message extractor plugin to npm (used in prepush and release scripts) builds only so that the properties file is generated based on all JS modules and not just those that are referenced. NPM build touches all JS modules and such suffices to extract any and every message.
* Changes release script to build assets and locales before running semantic release to detect any potential missing locales.

### DIFF
--- a/conf/cut_release.sh
+++ b/conf/cut_release.sh
@@ -125,6 +125,9 @@ build_assets() {
     printf "${blue}Building assets...${end}"
     yarn build:npm || return 1
     printf "${green}Built assets!${end}"
+    printf "${blue}Building locales...${end}"
+    yarn build:i18n || return 1
+    printf "${green}Built locales!${end}"
 }
 
 push_to_npm() {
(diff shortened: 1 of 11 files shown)


boxcla:

conf/cut_release.sh
```diff
@@ -125,6 +125,9 @@ build_assets() {
     printf "${blue}Building assets...${end}"
     yarn build:npm || return 1
     printf "${green}Built assets!${end}"
+    printf "${blue}Building locales...${end}"
+    yarn build:i18n || return 1
```
priyajeet: When you run i18n build after npm build, the properties file will get updated. If the file gets updated which means someone missed i18n build when pushing.

jstoffan: This change should probably be in a separate PR.

PR #853, opened 2019-02-12
fix: shared link precedence and query-bar updates

priyajeet:
latest pull from brui
Shared link should take precedence in quick search item link
query-bar and list-view updates


boxcla:

visylvius:
Thanks for pulling this in!

PR #855, opened 2019-02-14
fix: metadata query bar fixes and styling

priyajeet:
Final pull from brui

Metadata - rename filter components
IconBell - adjust position of svg to center in new dimensions
Selector drop down - scroll the container, not the body.
Metadata - handle filter button error states


boxcla:

PR #857, opened 2019-02-14
feat: Add two-toned icons and improve icon code/docs

tjgupta:
This PR includes various code cleanup and improvements to icons:

- new `Icon` flow type that most icons can share
- pulled out shared code for various file icons into an IconFileBase
- simplified docs for file and folder icons and added reusable `IconsExample` component
- added new `two-toned/IconSharedLink`, `two-toned/IconChatBubble`, and `two-toned/IconExclamationMark`


boxcla:
Hi @tjgupta, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

tjgupta:
CLA signed

src/icons/badges/SharedLinkBadge.js
```diff
@@ -3,42 +3,57 @@ import * as React from 'react';
...
-        <g fill="none" fillRule="evenodd">
-            <path
-                className="fill-color"
-                d="M15 8c0-3.866-3.134-7-7-7S1 4.134 1 8s3.134 7 7 7 7-3.134 7-7zM0 8c0-4.418 3.582-8 8-8s8 3.582 8 8-3.582 8-8 8-8-3.582-8-8z"
-                fill={color}
+const SharedLinkBadge = ({
+    className = '',
+    color,
+    height = 16,
+    styleProps = {
```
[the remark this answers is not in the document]
tjgupta: Mmm yeah, good point about spreading the values, I'll fix.

I originally had more specific names that mapped directly to the SVG (like `fill`, etc), but the problem is many icons have many different paths, circles, groups, etc with different fills and strokes that aren't always globally style-able. Without some sort of abstraction, this would open it up to infinite combinations. Imagine the chat icon, which has a path with a configurable fill color for the bubble and a path that's not configurable (at least not yet) for the dots inside the bubble. Would you then allow `bubbleFill`, `bubbleStroke`, `dotsFill`, and `dotsStroke`? Just declaring `fill` wouldn't be specific enough.

I think if we feel we need that much granularity that wouldn't be so bad, but some other folks gave feedback earlier that they'd like to see a common set of basic props that allow a basic set of shared concepts across all icon groups. So for example you'd have some icons that are "outline-able", i.e. you could set an outer border of a configurable width and color. And another class of icons might allow configurable outlines and configurable foreground/background colors (it would be up to the developer/designer to determine what is the foreground and what is the background for that particular type of icon). This was the approach we landed on due to the complexities mentioned above. Make sense @jstoffan?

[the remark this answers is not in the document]
jstoffan: I understand. We do currently have a wide variety of icons and their styling can be quite different. My concern is that this approach could lead to an explosion of styling micro-formats (ThreeToneStyle, FilledStyle, InvertedStyle, etc.) and make it very difficult to ever standardize. 

Limiting ourselves to the properties I listed above would provide the same basic styling options without that risk. In the IconChat example, `stroke` could be used for the bubble outline and `fill` for its background color. We would only need something like `bubbleStroke` if we want to color the inner dots separately, which doesn't seem to be a requirement today. I would argue that if such a requirement is introduced, we should push back on it in favor of standardization.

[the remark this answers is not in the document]
tjgupta: @jstoffan I think maybe we're trying to solve the same problem in different ways. The intention of creating a few base flow types was to **prevent** an explosion of styling formats or property names. The two base styles (`OutlinedStyle` and `TwoTonedStyle`) are sort of like mix-ins, and they have a fixed set of property names. Some icon families would use only the `OutlinedStyle` (like folder and file icons). And then TwoTonedOutlinedStyle was provided as a convenience since many of the icons need both of the mix-ins. In talking with design, this is likely to be the extent of the styling needs for the near future.

I believe you're suggesting a similar thing but perhaps without the notion of mix-ins, and with different property names (`stroke` and `fill` vs. `backgroundColor` and `foregroundColor`)? I had originally started with names like `stroke` and `fill` but moved away from that because I felt like those names were confusing since there are multiple strokes and fills within any given SVG, and so that actually made it confusing to understand what the given effect was going to be. To me, `backgroundColor` conveys the net visual effect (it will set the color of the background portion of the SVG), regardless of where the color gets applied (for example, depending on the SVG, it could actually end up being set on a stroke, or there could be two different fills and it only sets the color of one fill). The point of these particular names were to give enough abstraction so that we could use common names across all icons while avoiding confusing situations where, say, a prop called `fill` was actually being applied to a stroke because of the nature of a particular SVG.

src/icons/file/IconFileAudio.js
```diff
@@ -1,34 +1,49 @@
 // @flow
 import * as React from 'react';
+import uniqueId from 'lodash/uniqueId';
```
tjgupta: This is the first of many "file type" icons I updated in this PR. The same update has been applied to all files in `src/icons/file/*` - down to `IconFileZip.js`:

- Converted to a class based component
- Added `outlineId` setup in `componentWillMount()` because each icon instance needs a unique ID for the `filter` attribute to map to.
- Added in the `OutlineFilter`.
- Wrapped the paths in a `<g>` with a filter prop if the `styleProps` are set.

I did the first one manually, and then copied it over to the rest and just updated the pieces that are different from component to component, so if you review this one, the others should be pretty much identical. TODO: in a future PR we should create some kind of `IconFileBase` component and pull out all the shared code into that. There wouldn't be a whole lot left in these files then.

src/icons/general/IconSharedLink2.js
```diff
@@ -0,0 +1,67 @@
+// @flow
```
tjgupta: This is a new icon because I couldn't find a way to retrofit our existing one and preserve backward compatibility unfortunately.

src/icons/file/IconFileAudio.js
```diff
@@ -1,34 +1,49 @@
...
-            fill="#955ca5"
-        />
-        <path
-            d="M19 20.1V17h-6v4.5a1.5 1.5 0 1 1-1.5-1.5.9.9 0 0 1 .5.1V13h8v8.5a1.5 1.5 0 1 1-1.5-1.5.9.9 0 0 1 .5.1zM13 16h6v-2h-6z"
-            fill="#955ca5"
-        />
-    </AccessibleSVG>
-);
+class IconFileAudio extends React.Component<Props> {
+    componentWillMount() {
```
priyajeet: didMount

src/icons/file/IconFileAudio.js
```diff
@@ -1,34 +1,49 @@
...
+        this.outlineId = uniqueId('outline-');
+    }
+
+    outlineId: string;
+
+    render() {
+        const { className = '', height = 32, styleProps = {}, title, width = 32 } = this.props;
+        const filterProp = styleProps.outlineColor && styleProps.outlineWidth ? `url(#${this.outlineId})` : undefined;
```
priyajeet: outlineWidth if `0` will result in falsy. Assuming that's desired.

src/icons/file/IconFileAudio.js
```diff
@@ -1,34 +1,49 @@
...
-        />
-        <path
-            d="M19 20.1V17h-6v4.5a1.5 1.5 0 1 1-1.5-1.5.9.9 0 0 1 .5.1V13h8v8.5a1.5 1.5 0 1 1-1.5-1.5.9.9 0 0 1 .5.1zM13 16h6v-2h-6z"
-            fill="#955ca5"
-        />
-    </AccessibleSVG>
-);
+class IconFileAudio extends React.Component<Props> {
+    componentWillMount() {
+        this.outlineId = uniqueId('outline-');
```
priyajeet: Is there any way to create a HOC that does all this common mounting logic or a utility function that creates the filterProp below to reduce repetition?

src/icons/file/IconFileAudio.js
priyajeet: or just a class property or constructor, willMount is deprecated

src/icons/file/IconFileAudio.js
alexkrolick: How about a hook? Could do that as a followup when we upgrade. Seems like a great use case.

priyajeet: https://github.com/box/box-ui-elements/blob/master/src/icons/accessible-svg/AccessibleSVG.js#L14

Can we re-use?

alexkrolick: Like this:

```js
function useStableInstanceId(prefix) {
  return React.useCallback(() =>
    lodash.uniqueId(prefix),
    [prefix]
  )
}
```

```jsx
function MyIcon() {
  const [id] = useStableInstanceId('outline-')
  return <svg id={id}>{/* ... */}</svg>
}

function MyOtherIcon() {
  const [id] = useStableInstanceId('box-')
  return <svg id={id}>{/* ... */}</svg>
}
```

priyajeet: AccessibleSVG already has an id, and we can add an outlineId, and just pass those down to the svg without increasing complexity

priyajeet: Is the 1st child for AccessibleSVG always going to be `<OutlineFilter id={this.outlineId} styleProps={styleProps} />` for certain stuff? Can just use a prop to determine and have that code inside AccessibleSVG itself unless we prefer another level of indirection.

<AccessibleSVG>
<OutlinedSVG>
...
</OutlinedSVG>
</AccessibleSVG>

src/icons/OutlineFilter.js
```diff
@@ -0,0 +1,28 @@
...
+const OutlineFilter = ({ id, styleProps = {} }: Props) => {
+    const { outlineColor, outlineWidth } = styleProps;
+    const outlineFilter =
+        outlineColor && outlineWidth ? (
+            <filter id={id} x="-20%" y="-20%" width="150%" height="150%">
+                <feMorphology operator="dilate" radius={styleProps.outlineWidth} in="SourceAlpha" result="morphology" />
+                <feFlood floodColor={styleProps.outlineColor} floodOpacity="1" result="flood" />
+                <feComposite in="flood" in2="morphology" operator="in" result="composite" />
+                <feMerge result="merge">
+                    <feMergeNode in="composite" result="mergeNode" />
```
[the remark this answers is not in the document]
tjgupta: I'll take a look and see, maybe this can be simplified a bit.

src/icons/general/IconSharedLink2.js
```diff
@@ -0,0 +1,67 @@
...
+import * as React from 'react';
+import uniqueId from 'lodash/uniqueId';
+
+import { BOX_BLUE } from '../../common/variables';
```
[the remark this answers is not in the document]
tjgupta: Done.

src/icons/file/IconFileAudio.js
tjgupta: Yep. Although it might be better to be more explicit with it.

src/icons/file/IconFileAudio.js
tjgupta: I set up a `IconFileBase` component to remove some redundancy in all the file icons. Then I created an `OutlinedSVG` component which removed even more. Take a look at the PR now ignoring whitespace `?w=1`. Should be much easier to go through now.

We could also switch to hooks at some point, although now I think there's probably less value from that kind of switch...

tjgupta:
Hey folks, I pulled out the outline stuff from this change and updated the description to reflect the changes made. Please take a look again with `?w=1` to ignore whitespace changes.

src/icons/file/IconFileAudio.js
```diff
@@ -1,34 +1,23 @@
...
+import type { Icon } from '../flowTypes';
 
-type Props = {
-    className?: string,
-    height?: number,
-    /** A text-only string describing the icon if it's not purely decorative for accessibility */
-    title?: string | React.Element<any>,
-    width?: number,
+const IconFileAudio = (props: Icon) => {
+    return (
```
jstoffan: Can we stick with implicit returns? It looks like doing so would eliminate most of the whitespace changes in this changeset.

src/icons/file/IconFileBase.js
```diff
@@ -0,0 +1,26 @@
...
+
+import AccessibleSVG from '../accessible-svg';
+import type { Icon } from '../flowTypes';
+
+type Props = Icon & {
+    children: React.Node,
+    fileClassName: string,
+};
+
+const IconFileBase = ({ children, className = '', fileClassName, height = 32, title, width = 32 }: Props) => {
```
jstoffan: Do we want to tie file names and classes together? Or should `fileClassName` change to something like `baseClassName` or `iconClassName`?

src/icons/flowTypes.js
```diff
@@ -0,0 +1,21 @@
...
+export type Icon = {
+    className?: string,
+    color?: string,
+    height?: number,
+    /** A text-only string describing the icon if it's not purely decorative for accessibility */
+    title?: string | React.Element<any>,
+    width?: number,
+};
+
+export type TwoTonedIcon = {
```
jstoffan: Looks like this is only used in one place currently.

src/icons/flowTypes.js
```diff
@@ -0,0 +1,21 @@
...
+    className?: string,
+    height?: number,
+    styleProps: {|
+        backgroundColor?: string,
+        foregroundColor?: string,
+    |},
+    /** A text-only string describing the icon if it's not purely decorative for accessibility */
+    title?: string | React.Element<any>,
```
jstoffan: Should this be a junction type of a base type containing className, height, title, and width?

src/icons/file/IconFileBase.js
tjgupta: Updated to `baseClassName`.

src/icons/flowTypes.js
tjgupta: Updated this type now. The `TwoTonedIcon` doesn't include `color` since we decided that will be overridden with CSS only. I think it's best to keep them separate because of that.

src/icons/flowTypes.js
tjgupta: Yes, for now it is. The hope is that we start adding more icons of this type going forward that have better customization options via CSS.

src/icons/file/IconFileAudio.js
tjgupta: Fixed.

src/icons/two-toned/IconSharedLinkVars.scss
```diff
@@ -0,0 +1,9 @@
+@mixin icon-shared-link-style-override($primaryColor, $secondaryColor) {
```
jstoffan: Is this mixin intended to be imported by the parent app? If so, can we rename it to be more generic/reusable?

src/icons/general/IconChatRound.js
```diff
@@ -2,28 +2,25 @@
...
-};
-
-const IconChatRound = ({ className = '', color = '#999', height = 24, title, width = 24 }: Props) => (
+const IconChatRound = ({ className = '', color = '#999', height = 24, title, width = 24 }: Icon) => (
     <AccessibleSVG
         className={`icon-chat-round ${className}`}
         height={height}
         title={title}
-        viewBox="0 0 24 24"
+        viewBox="0 0 32 32"
```
jstoffan: It looks like this icon is used in the Sidebar nav as the Activity icon. Does it still look the same with these changes? Could you post a quick screenshot comparison?

src/icons/two-toned/IconSharedLinkVars.scss
tjgupta: It is intended for the parent app to use. The idea @booc0mtaco had was that since we're exposing the internals of the SVG through the CSS, we could mitigate that by also providing a mixin per icon. So the parent app would style it using this mixin rather than directly styling with selectors.

booc0mtaco: Two changes for consideration: 

- rename this file so it's a [partial import](https://sass-lang.com/guide#Partials). That way we can distinguish it from an actual SASS selectors needed for future icons/components
- we don't have a convention around naming for this situation yet. I'd lean toward something like `ComponentName.XXX.scss` replacing the XXX with `variables`, `mixins`, `custom`, or some words that denote what the file is for. This would fit with how we handle media query files, and some rising consensus about naming that have happened recently

src/icons/general/IconChatRound.js
tjgupta: Yep, definitely...

Here's the theme overridden version: 
screen shot 2019-03-05 at 2 02 54 pm

And regular:
screen shot 2019-03-05 at 2 05 44 pm

src/icons/two-toned/IconSharedLinkVars.scss
tjgupta: Done

src/icons/general/IconChatRound.js
tjgupta: To clarify, these are both screenshots from **after** my change...

PR #858, opened 2019-02-14
refactor: this reverts module import aliasing

priyajeet:
Revert is needed so that source files can be used by flow for flow checking. With import aliasing flow isnt happy. Goes back to using relative imports and reverts #784. Removes babel import resolver plugin.


boxcla:

PR #859, opened 2019-02-14
refactor(query-bar): replace props.template with props.columns

ChenCodes:
- removed template prop from FilterButton
- added columns prop to FilterButton
- updated tests for both FilterButton and ColumnButton due to changes in the columns contract found in fixtures.js


boxcla:

src/features/query-bar/__tests__/ColumnButtonOverlay-test.js
```diff
@@ -24,18 +34,28 @@ describe('features/query-bar/components/ColumnButtonOverlay', () => {
...
+                description:
+                    'Should not update state due to no destinationIndex',
                 sourceIndex: 0,
                 updatedColumns: columns,
             },
-        ].forEach(({ description, sourceIndex, destinationIndex, updatedColumns }) => {
-            test(`${description}`, () => {
-                const wrapper = getWrapper();
-                wrapper.instance().onDragEnd(sourceIndex, destinationIndex);
+        ].forEach(
```
visylvius: instead of doing a forEach here could you instead use a jest.each?

src/features/query-bar/__tests__/ColumnButtonOverlay-test.js
```diff
@@ -62,27 +82,38 @@ describe('features/query-bar/components/ColumnButtonOverlay', () => {
...
                 pendingColumns: columns,
                 updatedPendingColumns: columnsWithOneColumnNotShown,
             },
-        ].forEach(({ description, column, pendingColumns, updatedPendingColumns }) => {
-            test(`${description}`, () => {
-                const wrapper = getWrapper();
-                wrapper.setState({
-                    pendingColumns,
-                });
+        ].forEach(
```
visylvius: same as above

src/features/query-bar/__tests__/ColumnButtonOverlay-test.js
ChenCodes: Will tackle this in a separate PR

src/features/metadata-instance-editor/flowTypes.js
```diff
@@ -0,0 +1,107 @@
+// @flow
```
macrotim: I'm surprised this was previously omitted. Yikes.

src/features/query-bar/components/filter/Condition.js
```diff
@@ -159,7 +160,7 @@ const Condition = ({
     const renderDeleteButton = () => {
         return (
             <div className="filter-item-delete-button">
-                <button type="button" className="delete-button" onClick={onDeleteButtonClick}>
+                <button className="delete-button" onClick={onDeleteButtonClick}>
```
jstoffan: Please make sure to always provide a `type` for button elements, as they are `type="submit"` by default.

src/features/query-bar/components/filter/Condition.js
```diff
@@ -66,12 +67,12 @@ const Condition = ({
         const fieldId = option.fieldId || condition.fieldId;
 
         const displayText = option.displayText;
-        let displayTextType = '';
-        let keyType = '';
+        let displayTextType = '',
+            keyType = '';
```
jstoffan: Please use one `let/const` per variable.

src/features/metadata-instance-editor/flowTypes.js
priyajeet: Remove this file, the types should be available globally

src/features/query-bar/QueryBar.js
```diff
@@ -10,20 +11,35 @@ import type { ColumnType } from './flowTypes';
 import './styles/QueryBarButtons.scss';
 
 type Props = {
-    activeTemplate?: MetadataTemplate,
+    activeTemplate?: Template,
```
priyajeet: revert

src/features/query-bar/components/filter/Condition.js
```diff
@@ -127,11 +128,11 @@ const Condition = ({
 
     const updateValueField = (fieldValue: Object) => {
         let displayText = '';
-        const displayTextType = VALUE_DISPLAY_TEXT;
-        const fieldId = condition.fieldId;
+        let displayTextType = VALUE_DISPLAY_TEXT;
```
jstoffan: Please use `const` if a variable is never reassigned.

priyajeet:
I would recommend doing your changes based on what this repo has. Not overwrite all the changes. I see eslint issues, and you should never be pushing with no-verify.

priyajeet:
You may have to close PR and update your commit message locally. Since commit lint checks that not the PR title. Also probably should not use your real name in tests.

src/features/query-bar/components/filter/Condition.js
```diff
@@ -176,12 +178,52 @@ const Condition = ({
         );
     };
 
+    const renderAttributeField = () => {
```
jstoffan: Generally, anything in a `renderThing` type function should probably be its own component. Such a change is outside of the scope of this PR, so don't worry about changing it now.

src/features/query-bar/components/filter/Condition.js
```diff
@@ -176,12 +178,52 @@ const Condition = ({
...
+    };
+
+    const renderOperatorField = () => {
+        const { operatorKey } = condition;
+        const operatorsForAttribute = getOperatorsForAttribute();
+        const operatorOptions = getFormattedOptions(operatorsForAttribute);
+
+        return (
+            <div className="filter-item-operator-dropdown-container">
+                <div className="filter-dropdown-single-select-field-container">
```
priyajeet: You can potentially save some bytes by making the class a constant. I see it's being used twice and its long.
Also is double-nested div needed?

src/features/query-bar/components/fixtures.js
```diff
@@ -1,66 +1,128 @@
-const initialCondition = {
-    conditionId: 0,
-    attributeDisplayText: '',
-    attributeKey: null,
+// @flow
+import type { ColumnType, ConditionType } from '../flowTypes';
```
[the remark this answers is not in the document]
ChenCodes: Yes

src/features/query-bar/components/filter/Condition.js
ChenCodes: I can address in a separate PR
No it looks like the double nested div is not needed

PR #860, opened 2019-02-15
build: Prefix scripts with yarn to enforce local binary usage

jstoffan:
Yarn all the things.


boxcla:

alexkrolick:
Isn't that the default?

jstoffan:
Hmm. The binary path shown in the console output is different with/without this change. Let me install eslint globally and check the versions.

jstoffan:
Looks like you're right @alexkrolick. The path in the console shows `eslint` without this change, but it's still the local binary, even if you have it installed globally as well. My bad!

PR #861, opened 2019-02-15
build(examples): dont assume gh-pages branch exists locally

priyajeet:


boxcla:

conf/cut_release.sh
```diff
@@ -144,7 +144,11 @@ build_examples() {
 
 push_to_gh_pages() {
     printf "${blue}Pushing styleguide to gh-pages...${end}"
-    git branch -D gh-pages || return 1
+    if git branch -D gh-pages; then
```
booc0mtaco: Might be good to suppress the error message entirely. I think this will still print it but it's not really an error for us

priyajeet: I'll try `git branch | grep -w "gh-pages"`

PR #862, opened 2019-02-15
feat(query-bar): implement connector dropdown

ChenCodes:
screen shot 2019-02-14 at 10 52 22 pm


boxcla:

src/features/query-bar/constants.js
```diff
@@ -9,6 +9,9 @@ export const OPERATOR_KEY: 'operatorKey' = 'operatorKey';
 export const VALUE: 'value' = 'value';
 export const VALUE_DISPLAY_TEXT: 'valueDisplayText' = 'valueDisplayText';
 export const VALUE_KEY: 'valueKey' = 'valueKey';
+export const WHERE: null = null;
```
priyajeet: why is this null?

src/features/query-bar/components/filter/FilterButton.js
```diff
@@ -20,6 +21,7 @@ type State = {
     areErrorsEnabled: boolean,
     conditions: Array<Object>,
     isMenuOpen: boolean,
+    selectedPrefix: string,
```
priyajeet: Should this be string or `typeof WHERE | typeof AND | typeof OR`?

src/features/query-bar/components/filter/FilterButton.js
```diff
@@ -148,18 +164,74 @@ class FilterButton extends React.Component<Props, State> {
         }
     };
 
-    deleteCondition = (index: number) => {
-        const { conditions } = this.state;
+    updateSelectedPrefix = (option: Object) => {
```
priyajeet: Can option be something more specific than `Object`?

src/features/query-bar/components/filter/FilterButton.js
```diff
@@ -148,18 +164,74 @@ class FilterButton extends React.Component<Props, State> {
...
+        const displayText = option.displayText;
+        let updatedPrefix = '';
+
+        switch (displayText) {
+            case 'AND':
```
priyajeet: the string 'AND' and 'OR' are available as constants as you used them below

src/components/select-field/BaseSelectField.js
```diff
@@ -98,7 +98,14 @@ class BaseSelectField extends React.Component<Props, State> {
         // @NOTE: breaks encapsulation but alternative is unknown child ref
         const itemEl = id ? document.getElementById(id) : null;
         if (itemEl) {
-            scrollIntoViewIfNeeded(itemEl, false);
+            let parentEl = itemEl;
```
priyajeet: I would look at if there are other available DOM functions to achieve something similar, like https://developer.mozilla.org/en-US/docs/Web/API/Element/closest

Also, it seems SelectorDropDown.js has a very similar method. I don't think we should be replicating more code like that. What you can do is create a function inside utils/dom.js and extract this function in there and then use that function here as well as in SelectorDropDown.js.

Also does it have to check similar classes in both components? If not, the classes can be passed as an array of string to that function. Also ideally classes should be constants exported from their parent function. So that 'overlay-wrapper' isn't floating around in places.

Lastly I would make this change in a seperate PR as its not really related to query-bar.

src/features/query-bar/components/filter/Condition.js
```diff
@@ -43,6 +46,7 @@ type Props = {
         fieldKeyType: string,
         valueType: any,
     ) => void,
+    updateSelectedPrefix: (option: Object) => void,
```
priyajeet: Something more specific than Object if possible

src/features/query-bar/components/filter/Condition.js
```diff
@@ -86,12 +91,12 @@ const Condition = ({
...
 
     const getFormattedOptions = (options: Array<Object>): any[] => {
-        return options.map((option, optionIndex) => {
+        return options.map(option => {
             return {
                 displayText: option.displayName,
                 fieldId: option.id,
                 type: option.type,
-                value: optionIndex,
+                value: option.displayName,
```
priyajeet: Is this displayName same as displayText in the other files?

src/features/query-bar/components/filter/Condition.js
```diff
@@ -169,17 +174,48 @@ const Condition = ({
...
+            })),
+        );
+
         return (
             <div className="filter-item-prefix-container">
-                <p className="filter-item-prefix-text">
-                    <FormattedMessage {...messages.prefixButtonText} />
-                </p>
+                {prefix === WHERE ? (
+                    <p className="filter-item-prefix-text">{formatMessage(message)}</p>
```
priyajeet: We should use `<FormattedMessage>` wherever possible.

src/features/query-bar/components/filter/Condition.js
```diff
@@ -169,17 +174,48 @@ const Condition = ({
...
+                ) : (
+                    <SingleSelectField
+                        isDisabled={false}
+                        onChange={updateSelectedPrefix}
+                        options={prefixOptions}
+                        placeholder={formatMessage(message)}
```
priyajeet: Double check, but it seems like the lower level BaseSelectField can take DOM `<FormattedMessage>` so single select field's flow type can be fixed to take React.Node

src/features/query-bar/constants.js
ChenCodes: This can be made into the 'WHERE' string. Initially I had numbers to denote AND and OR but I changed them over to strings so WHERE can be made a string as well.

src/features/query-bar/components/filter/FilterButton.js
ChenCodes: It's better to be more specific, I can go with typeof instead.

src/features/query-bar/components/filter/FilterButton.js
ChenCodes: Fixed

src/features/query-bar/components/filter/Condition.js
ChenCodes: Fixed

src/components/select-field/BaseSelectField.js
ChenCodes: I'll move to a separate PR

src/features/query-bar/components/filter/Condition.js
ChenCodes: Yes

src/features/query-bar/components/filter/Condition.js
ChenCodes: Fixed

src/features/query-bar/components/filter/Condition.js
ChenCodes: Fixed

src/features/query-bar/components/filter/FilterButton.js
ChenCodes: Fixed

src/features/query-bar/components/filter/FilterButton.js
```diff
@@ -148,18 +164,62 @@ class FilterButton extends React.Component<Props, State> {
...
+        this.setState({
+            selectedPrefix: prefix,
+        });
+
+        this.updateConditionsPrefixes(prefix);
+    };
+
+    updateConditionsPrefixes = (prefix: string) => {
+        const { conditions } = this.state;
+        const conditionsAfterDeletion = conditions.map((condition, index) => {
```
krapspark: is this actually the `conditionsAfterDeletion`? looks out of place in this function

src/features/query-bar/components/filter/FilterButton.js
```diff
@@ -148,18 +164,62 @@ class FilterButton extends React.Component<Props, State> {
...
+        const conditionsAfterDeletion = conditions.filter((condition, conditionIndex) => {
+            return conditionIndex !== index;
+        });
+
+        if (conditionsAfterDeletion.length === 0) {
+            this.setState({
+                conditions: [],
+            });
+        } else {
+            // The first condition must always have a prefix of WHERE.
```
krapspark: I'm not too familiar with how these conditions work but this arrangement feels brittle. Let's talk in person about how this is structured.

src/features/query-bar/components/filter/FilterButton.js
```diff
@@ -10,16 +10,18 @@ import Button from '../../../../components/button/Button';
...
 
-import type { ColumnType } from '../../flowTypes';
+import type { ColumnType, SelectOptionType } from '../../flowTypes';
 
 type State = {
     appliedConditions: Array<Object>,
     areErrorsEnabled: boolean,
     conditions: Array<Object>,
     isMenuOpen: boolean,
+    selectedPrefix: typeof WHERE | typeof AND | typeof OR,
```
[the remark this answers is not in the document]
ChenCodes: I think this enforces that these are all strings.

src/features/query-bar/components/filter/FilterButton.js
```diff
@@ -21,7 +21,7 @@ type State = {
     areErrorsEnabled: boolean,
     conditions: Array<Object>,
     isMenuOpen: boolean,
-    selectedPrefix: typeof WHERE | typeof AND | typeof OR,
+    selectedConnector: typeof WHERE | typeof AND | typeof OR,
```
krapspark: remove WHERE

src/features/query-bar/components/filter/FilterButton.js
```diff
@@ -164,35 +161,11 @@ class FilterButton extends React.Component<Props, State> {
...
-
-        // The first condition must always have a prefix of WHERE.
-        const firstCondition = Object.assign({}, conditionsAfterDeletion[0]);
-        firstCondition.prefix = WHERE;
-
-        const updatedConditions = [firstCondition, ...conditionsAfterDeletion.slice(1)];
-
-        this.setState({
-            conditions: updatedConditions,
+            selectedConnector: connector,
```
krapspark: are there multiple instances of `FilterButton` on a query dropdown? If so, the `selectedConnector` state should be stored in the parent component

ChenCodes: The query dropdown is within the FilterButton so that technically is the parent component

src/features/query-bar/components/filter/FilterButton.js
```diff
@@ -10,16 +10,18 @@ import Button from '../../../../components/button/Button';
...
 
-import type { ColumnType } from '../../flowTypes';
+import type { ColumnType, SelectOptionType } from '../../flowTypes';
 
 type State = {
     appliedConditions: Array<Object>,
     areErrorsEnabled: boolean,
     conditions: Array<Object>,
     isMenuOpen: boolean,
+    selectedConnector: typeof AND | typeof OR,
```
macrotim: Shouldn't the type be "ConnectorType"? It isn't clear to me why use typeof here.

src/features/query-bar/components/filter/FilterButton.js
```diff
@@ -30,16 +32,25 @@ type Props = {
...
-            conditions: [initialCondition],
+            conditions: [],
+            selectedConnector: AND,
             areErrorsEnabled: false,
             isMenuOpen: false,
         };
     }
 
+    componentDidMount() {
+        const id = this.generateConditionID();
```
macrotim: generateConditionID and createCondition seem coupled. Any reasons not to merge into one?

src/features/query-bar/components/filter/FilterButton.js
```diff
@@ -148,15 +161,23 @@ class FilterButton extends React.Component<Props, State> {
         }
     };
 
+    updateSelectedConnector = (option: SelectOptionType) => {
+        const connector = option.displayText;
+
+        this.setState({
+            selectedConnector: connector,
```
macrotim: This should be set to option.key in case displayText changes in the future.

Unlike the displayText, the key absolutely should not change, ever.

src/features/query-bar/components/filter/FilterButton.js
```diff
@@ -248,7 +269,9 @@ class FilterButton extends React.Component<Props, State> {
                                             areErrorsEnabled={areErrorsEnabled}
                                             index={index}
                                             columns={columns}
+                                            selectedConnector={selectedConnector}
                                             update={this.update}
+                                            updateselectedConnector={this.updateselectedConnector}
```
macrotim: The camel casing is off.

macrotim: This prop is named inconsistently.

QueryBar uses the "on XYZ Change" format. Similar to the "update" prop.

src/features/query-bar/components/filter/Condition.js
```diff
@@ -168,20 +177,50 @@ const Condition = ({
...
+            })),
+        );
+
         return (
-            <div className="filter-item-prefix-container">
-                <p className="filter-item-prefix-text">
-                    <FormattedMessage {...messages.prefixButtonText} />
-                </p>
+            <div className="filter-item-connector-container">
+                {index === 0 ? (
```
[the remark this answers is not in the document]
ChenCodes: Index is coming from the .map in FilterButton.js when rendering each Condition component

src/features/query-bar/components/filter/Condition.js
```diff
@@ -168,20 +177,50 @@ const Condition = ({
...
+                value: key,
+            })),
+        );
+
         return (
-            <div className="filter-item-prefix-container">
-                <p className="filter-item-prefix-text">
-                    <FormattedMessage {...messages.prefixButtonText} />
-                </p>
+            <div className="filter-item-connector-container">
```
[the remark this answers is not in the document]
ChenCodes: Not necessary

src/features/query-bar/components/filter/FilterButton.js
ChenCodes: No it makes more sense to just reference id within createCondition

src/features/query-bar/components/filter/FilterButton.js
krapspark: `typeof AND` will just resolve to string here. so you can either set it to values (`selectedConnector: AND | OR`) or just use string.

macrotim: I'd like to see new flow types AND / OR and a new ```ConnectorType = AND | OR```.

One challenge though is with naming. Will both flowTypes and constants.js export AND? My experience with flow is limited here.

src/features/query-bar/components/filter/FilterButton.js
ChenCodes: I'll be using option.value instead of option.key for now since that's how the flowtype is defined for a SelectOptionType

src/features/query-bar/components/filter/FilterButton.js
booc0mtaco: So long as we have the literal value as the type, it will do the right thing. looks like `AND` and `OR` have "value-types" which makes this evaluate properly.

screen shot 2019-02-19 at 12 00 37 pm

macrotim: Thanks for pointing it out. 

Then maybe all we need is a new ConnectorType = typeof AND | typeof OR. Thoughts?

ChenCodes: That looks good to me

src/features/query-bar/components/filter/FilterButton.js
```diff
@@ -271,7 +270,7 @@ class FilterButton extends React.Component<Props, State> {
                                             columns={columns}
                                             selectedConnector={selectedConnector}
                                             update={this.update}
-                                            updateselectedConnector={this.updateselectedConnector}
+                                            onConnectorChange={this.onConnectorChange}
```
macrotim: This might be a matter of preference. The convention I've observed is...
1. The left-side prop name uses the "on" format
2. The right-side function uses the "handle" format

It is analogous to the event name and event handler.

src/features/query-bar/components/filter/Condition.js
```diff
@@ -191,22 +191,22 @@ const Condition = ({
         }
 
         const connectorOptions = getFormattedOptions(
-            [AND, OR].map(key => ({
-                displayName: key,
-                value: key,
+            [AND, OR].map(connector => ({
+                displayName: CONNECTOR_NAMES[connector],
```
macrotim: I don't think we need CONNECTOR_NAMES today. 

What do we lose by removing this?

src/features/query-bar/components/filter/FilterButton.js
```diff
@@ -65,7 +64,8 @@ class FilterButton extends React.Component<Props, State> {
         this.setState({ isMenuOpen: !this.state.isMenuOpen });
     };
 
-    createCondition = (conditionID: string) => {
+    createCondition = () => {
+        const conditionID = this.generateConditionID();
```
macrotim: Could this be reduced to uniqueId? I don't see a purpose for generateConditionID.

src/features/query-bar/constants.js
```diff
@@ -1,17 +1,22 @@
 // @flow
 
+export const AND: 'AND' = 'AND';
+export const OR: 'OR' = 'OR';
 export const ATTRIBUTE: 'attribute' = 'attribute';
 export const COLUMN_DISPLAY_TEXT: 'columnDisplayText' = 'columnDisplayText';
 export const COLUMN_KEY: 'columnKey' = 'columnKey';
+export const CONNECTOR_NAMES = {
```
macrotim: I'm not sure about the variable name.

It's odd that index 0 means AND, index 1 means OR.

I don't see a purpose for this. What do we lose by removing it?

src/features/query-bar/flowTypes.js
```diff
@@ -28,3 +29,5 @@ export type ConditionType = {
     valueKey: string | null,
     valueType: string,
 };
+
+export type ConnectorType = typeof AND | typeof OR;
```
macrotim: Shouldn't this support null? Isn't it a valid value for ConnectorType?

src/features/query-bar/components/filter/Condition.js
ChenCodes: I can remove, just thought I'd put it there for future reference

src/features/query-bar/components/filter/FilterButton.js
ChenCodes: Yes

src/features/query-bar/flowTypes.js
ChenCodes: Currently in the UX it's not possible to have a null ConnectorType as the default is set to AND.

macrotim: I'm a bit confused. 

Shouldn't ConnectorType exist within ConditionType? How else would the component communicate connector to the V2 endpoint? 

If the answer is yes, then shouldn't the connector be nullable?

macrotim: I'd like to see Acceptance Criteria around this -> how is the Connector communicates to the calling component when user clicks Apply?

src/features/query-bar/constants.js
ChenCodes: Removing it

src/features/query-bar/flowTypes.js
ChenCodes: I don't think ConnectorType necessarily needs to live within ConditionType. Right now since the ConnectorType is either AND | OR across all conditions, we don't need to keep ConnectorType as a property of a Condition object. In the future if we are handling more complex logic then yes we will need ConnectorType to be part of the Condition object but for now since any changes to the current selected Connector is applied across all Conditions, it can remain as a type independent of Condition.

For communicating connector to the V2 endpoint, here is an example of the transform method:
```
transformConditionToSQL = (conditions: Array<Object>) => {
        const { selectedConnector } = this.state;
        let sql = '';
        conditions.forEach((condition, index) => {
            const { valueType } = condition;
            const operators = OPERATORS_FOR_ATTRIBUTE[valueType];
            if (index === 0) {
                sql += 'where ';
                sql += `columnName is ${condition.columnDisplayText} `;
                sql += `and operator is ${operators[0].displayName} `;
                sql += `and value is ${condition.value} `;
            } else {
                sql += `${selectedConnector} `;
                sql += `columnName is ${condition.columnDisplayText} `;
                sql += `and operator is ${operators[0].displayName} `;
                sql += `and value is ${condition.value} `;
            }
        });
        return sql;
    };
```

ChenCodes: As for a nullable ConnectorType would that only apply to the first Condition? For the 'where' label.

macrotim: The FilterButton and the "transform" logic should be decoupled. 

FilterButton is concerned with _display_ of conditions ("view concerns") and should have no awareness of SQL. It helps with unit testing, avoids bloating FilterButton, and makes it possible to reason about each without having to juggle multiple concepts at one time.

How would you go about decoupling i.e. to communicate Conditions + connector from FilterButton to transform.

src/features/query-bar/components/filter/Condition.js
```diff
@@ -150,91 +156,122 @@ const Condition = ({
...
+                {index === 0 ? (
+                    <p className="condition-connector-text">
+                        <FormattedMessage {...messages.connectorWhereText} />
+                    </p>
+                ) : (
+                    <SingleSelectField
+                        isDisabled={false}
+                        onChange={onConnectorChange}
+                        options={connectorOptions}
+                        placeholder={message}
```
macrotim: Why use placeholder instead of selectedValue? 

See the purpose of each prop here. 

https://git.dev.box.net/Box/box-react-ui/blob/e207854a487574b7be3ec6a7cad9c8c55afac271/src/components/select-field/SingleSelectField.js

src/features/query-bar/components/filter/Condition.js
```diff
@@ -150,91 +156,122 @@ const Condition = ({
...
                 <button className="delete-button" onClick={onDeleteButtonClick} type="button">
                     <IconClose width={deleteButtonIconWidth} height={deleteButtonIconHeight} color="#999EA4" />
                 </button>
             </div>
         );
     };
 
-    const renderPrefixField = () => {
+    const renderConnectorField = () => {
+        let message = '';
```
macrotim: The name "message" confuses me. It's passed to "placeholder" but this is not the purpose of placeholders.

ChenCodes:
Will address remaining comments in a separate PR

src/features/query-bar/components/filter/Condition.js
ChenCodes: Only selectedValue is needed, removing placeholder

PR #869, opened 2019-02-15
fix(sidebar): At mentions with no link should render inline

jeremypress:
This only affects at mentions that don't have a corresponding link.

Before:
screen shot 2019-02-15 at 2 02 37 pm

After:
screen shot 2019-02-15 at 2 01 00 pm


boxcla:

PR #870, opened 2019-02-15
fix(dropdown): Single select field hover scroll glitch

ChenCodes:
Currently there is an issue with hovering over SingleSelectField and SelectorDropdown components. The page auto scrolls whenever we hover over an item which is not the effect that we would like.

The fix -> when an item is active, we traverse up to the dom tree of that item looking for a parent class that we do want to scroll. If that element is contained in a modal, then we want to scroll the modal, and not the element itself.

Acceptance Criteria:
When a user hovers over either a SingleSelectField or an item in a SelectorDropdown
Then the page should not suddenly scroll

**To Verify this fix:**

1. SingleSelectField Flow:
Given the user is on the UI Elements styleguide
When the user goes to the Query Bar section
When the user clicks on the Modify Filters button
When the user clicks on the first field dropdown
When the user hovers over any of the items in the dropdown
Then the user should see the that page does not suddenly scroll

2. SelectorDropdown Flow:
Given the user is on the UI Elements styleguide
When the user goes to the MetadataInstanceEditor section
When the user clicks on the Add button
When the user hovers over any of the items in the dropdown
Then the user should see that the page does not suddenly scroll


boxcla:

src/components/select-field/BaseSelectField.js
```diff
@@ -1,16 +1,18 @@
...
 
+import { scrollItemIntoView } from '../../utils/dom';
 import IconCheck from '../../icons/general/IconCheck';
 import SelectButton from '../select-button';
 import DatalistItem from '../datalist-item';
 import type { SelectOptionValueProp, SelectOptionProp } from './props';
 
 import './SelectField.scss';
 
+const overlayWrapperClass = 'overlay-wrapper';
```
krapspark: all caps

src/components/selector-dropdown/SelectorDropdown.js
```diff
@@ -1,13 +1,15 @@
...
 import classNames from 'classnames';
-import scrollIntoViewIfNeeded from 'scroll-into-view-if-needed';
 import uniqueId from 'lodash/uniqueId';
 
+import { scrollItemIntoView } from '../../utils/dom';
 import ScrollWrapper from '../scroll-wrapper';
 
 import './SelectorDropdown.scss';
 
+const overlayWrapperClass = 'overlay-wrapper';
```
krapspark: all caps

src/components/select-field/BaseSelectField.js
```diff
@@ -1,16 +1,18 @@
...
 
+import { scrollItemIntoView } from '../../utils/dom';
 import IconCheck from '../../icons/general/IconCheck';
 import SelectButton from '../select-button';
 import DatalistItem from '../datalist-item';
 import type { SelectOptionValueProp, SelectOptionProp } from './props';
 
 import './SelectField.scss';
 
+const OVERLAY_WRAPPER_CLASS = 'overlay-wrapper';
```
macrotim: What if the field does not reside within an Overlay component? Does scrollItemIntoView behave correctly? What if the BaseSelectField resides within the HTML body itself? Shouldn't the body be scrolled?

src/components/select-field/__tests__/BaseSelectField-test.js
```diff
@@ -1,10 +1,15 @@
 import React from 'react';
 import sinon from 'sinon';
 
+import { scrollItemIntoView } from '../../../utils/dom';
 import BaseSelectField from '../BaseSelectField';
 
 const sandbox = sinon.sandbox.create();
 
+jest.mock('../../../utils/dom', () => ({
+    scrollItemIntoView: jest.fn().mockReturnValue(null),
```
macrotim: Is mockReturnValue needed?

ChenCodes: I can remove that

src/components/selector-dropdown/SelectorDropdown.js
```diff
@@ -74,29 +76,13 @@ class SelectorDropdown extends React.Component<Props, State> {
...
 
-    scrollItemIntoView = (id: string | null) => {
-        // @NOTE: breaks encapsulation but alternative is unknown child ref
-        const itemEl = id ? document.getElementById(id) : null;
-        if (itemEl) {
-            let parentEl = itemEl;
-            while (
-                parentEl.parentElement instanceof HTMLElement &&
-                !parentEl.classList.contains('scroll-wrap-container') && // Scroll the container, not the body when contained in a scroll-container.
-                !parentEl.classList.contains('modal') // Scroll the modal, not the body when contained in a modal.
```
macrotim: Do we no longer cover the modal use case?

ChenCodes: I will put the modal use case back. When I tested the functionality it worked for MetadataInstanceEditor but I didn't test for the modal case in EUA

src/components/select-field/BaseSelectField.js
ChenCodes: Yes, I think we should be passing in the class that denotes what should be scrolled. Instead of hardcoding overlay-wrapper as the class that should be scrolled, we need to define on a case-by-case basis which class should be scrolled and pass that in to BaseSelectField / SelectorDropdown

priyajeet: I would name the variable starting with CLASS_OVERLAY_WRAPPER.
Also search the codebase to see if there are other uses of this class name, and if so, move the constant to src/constants

src/utils/dom.js
```diff
@@ -78,3 +79,18 @@ export function focus(root: ?HTMLElement, selector?: string, focusRoot: boolean
         root.focus();
     }
 }
+
+export function scrollItemIntoView(id: ?string, className: string): void {
```
priyajeet: Probably can take in array of classNames if you want to test more than one.
You can also use element.closest instead of the loop if its available. You can check the prototype if its available https://developer.mozilla.org/en-US/docs/Web/API/Element/closest#Polyfill

priyajeet: Also, should the function just take an HTML element instead of ID?
```
function scrollIntoView(element: ?HTMLElement, classNames: Array<string>)
```

priyajeet: probably also see if you can add a small unit test for this

macrotim:
I'd like to your A.C. cover more scenarios e.g. Which README(s) should this fix affect? How can I or other engineers verify that it's working correctly?

ChenCodes:
Updated A.C. to include specific flows for verifying that this works correctly

src/utils/dom.js
macrotim: @priyajeet, is it a good idea to polyfill .closest? If not, let's consider adding to utils/dom/closest. @ChenCodes, the source code is provided in the [MDN link](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest#Polyfill).

RE: HTML Element instead of ID. I vote no. This function is looking for the closest element of the target classnames (It doesn't yet but should implement that. I've given feedback here). It behaves more like .closest. See the previous implementation here:

```            while (
                parentEl.parentElement instanceof HTMLElement &&
                !parentEl.classList.contains('scroll-wrap-container') && // Scroll the container, not the body when contained in a scroll-container.
                !parentEl.classList.contains('modal') // Scroll the modal, not the body when contained in a modal.
            ) {
                parentEl = parentEl.parentElement;
            }
```

I propose doing something like this instead of the while-loop:
```const parentEl = el.closest('html, .modal, .overlay-wrapper');
```

macrotim: I don't think we should introduce a new name i.e. "scroll _item_ into view". Why not make it consistent?

src/utils/dom.js
```diff
@@ -78,3 +79,26 @@ export function focus(root: ?HTMLElement, selector?: string, focusRoot: boolean
...
+ */
+export function scrollIntoView(itemEl: ?HTMLElement): void {
+    // @NOTE: breaks encapsulation but alternative is unknown child ref
+    if (itemEl) {
+        const parentEl = itemEl.closest('html, .modal, .overlay-wrapper');
+        scrollIntoViewIfNeeded(itemEl, false, undefined, parentEl);
+    }
+}
+
+window.Element.prototype.closest = function closest(s) {
```
priyajeet: Move this to the top of the file and put a nice comment on top of it, and guard it as shown here https://developer.mozilla.org/en-US/docs/Web/API/Element/closest#Polyfill

I also think you may need the matches part as shown in the link above due to https://caniuse.com/#search=matches IE11 using ms prefix

src/utils/domPolyfill.js
```diff
@@ -0,0 +1,22 @@
...
+
+    let el = this;
+    do {
+        if (el.matches(s)) return el;
+        el = el.parentElement || el.parentNode;
+    } while (el !== null && el.nodeType === 1);
+    return null;
+};
+
+module.exports = window.Element.prototype.closest;
```
priyajeet: dont need this last line iirc

src/utils/domPolyfill.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * Polyfills the .closest method for ELement
+ * Currently being used for scrollIntoView
+ * For reference: https://github.com/zloirock/core-js/issues/317
+ *
+ * @param {HTMLElement|null} s - the dom string that is being checked against
+ * @return {HTMLElement|null} returns either the parent element that was found or null
+ */
+window.Element.prototype.closest = function closest(s) {
```
priyajeet: For consistency, I would just drop window. part, since at other places you are referencing Element directly. Or change other places to use window. too.

src/utils/domPolyfill.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * Polyfills the .closest method for ELement
+ * Currently being used for scrollIntoView
+ * For reference: https://github.com/zloirock/core-js/issues/317
+ *
+ * @param {HTMLElement|null} s - the dom string that is being checked against
```
priyajeet: JSDoc incorrect, s is a string.
I would just get rid of the docs

src/utils/dom.js
```diff
@@ -78,3 +80,17 @@ export function focus(root: ?HTMLElement, selector?: string, focusRoot: boolean
...
+/**
+ * Scrolls the container / modal / wrapper instead of the body
+ *
+ * @param {HTMLElement} itemEl - the base dom element to search
+ * @return {void}
+ */
+export function scrollIntoView(itemEl: ?HTMLElement): void {
+    // @NOTE: breaks encapsulation but alternative is unknown child ref
+    if (itemEl) {
+        const parentEl = itemEl.closest('html, .modal, .overlay-wrapper');
```
priyajeet: `html` --> `body` since I think we should terminate at the body

src/utils/__tests__/dom-test.js
```diff
@@ -0,0 +1,23 @@
...
+        });
+
+        afterEach(() => {
+            document.body.innerHTML = '';
+        });
+
+        test('should call scrollIntoView', () => {
+            const itemEl = document.querySelector('.button');
+            scrollIntoView(itemEl);
+            expect(scrollIntoViewIfNeeded).toHaveBeenCalled();
```
priyajeet: toHaveBeenCalledWith(...) add the args to see if the function is properly passing in stuff.
Maybe add a test when parent is evaluated as null

src/components/selector-dropdown/SelectorDropdown.js
```diff
@@ -1,13 +1,15 @@
...
 import classNames from 'classnames';
-import scrollIntoViewIfNeeded from 'scroll-into-view-if-needed';
 import uniqueId from 'lodash/uniqueId';
 
+import { scrollIntoView } from '../../utils/dom';
 import ScrollWrapper from '../scroll-wrapper';
 
 import './SelectorDropdown.scss';
 
+const OVERLAY_WRAPPER_CLASS = 'overlay-wrapper';
```
priyajeet: Move this constant to Styles section of src/constants.js, and rename to CLASS_OVERLAY_WRAPPER and then use that

src/components/select-field/BaseSelectField.js
```diff
@@ -1,16 +1,18 @@
...
 
+import { scrollIntoView } from '../../utils/dom';
 import IconCheck from '../../icons/general/IconCheck';
 import SelectButton from '../select-button';
 import DatalistItem from '../datalist-item';
 import type { SelectOptionValueProp, SelectOptionProp } from './props';
 
 import './SelectField.scss';
 
+const OVERLAY_WRAPPER_CLASS = 'overlay-wrapper';
```
priyajeet: remove?

PR #875, opened 2019-02-21
fix(header-flyout): Update styles from design feedback

booc0mtaco:
- Add in new size metrics and spacing
- Update variable values to reflect feedback
- Use separated variables and mixins to allow for external control and consistency


boxcla:

PR #877, opened 2019-02-21
chore(query-bar): rename condition methods

ChenCodes:
Changes in this PR:
- Renaming methods to use 'column' instead of 'attribute'
- Rename / remove unused variables 
- Remove placeholder & messages for connectorField and operatorField 
- Add unit test for getColumnOptions


boxcla:

src/features/query-bar/components/filter/Condition.js
```diff
@@ -12,14 +12,14 @@ import ValueField from './ValueField';
 import messages from '../../messages';
 import {
     AND,
-    ATTRIBUTE,
+    COLUMN,
```
macrotim: This name might be confusing but let's leave it for now.

src/features/query-bar/components/filter/Condition.js
```diff
@@ -102,33 +102,30 @@ const Condition = ({
         });
     };
 
-    const getOperatorsForAttribute = () => {
+    const getColumnOperators = () => {
```
macrotim: Ah I love the new name.

src/features/query-bar/components/filter/Condition.js
```diff
@@ -102,33 +102,30 @@ const Condition = ({
...
 
-    const getValuesForAttribute = () => {
+    const getColumnOptions = () => {
         const { fieldId } = condition;
-        const field =
+        const column =
             columns &&
-            columns.find(column => {
-                return column.id === fieldId;
+            columns.find(c => {
```
macrotim: Could be reduced to one-liner.

src/features/query-bar/components/filter/Condition.js
```diff
@@ -102,33 +102,30 @@ const Condition = ({
...
+                return c.id === fieldId;
             });
-
-        if (field && field.options) {
-            const fieldOptions = field.options.map((option, optionIndex) => {
+        if (column && column.options) {
+            return column.options.map((o, idx) => {
```
macrotim: "option" is better than "o". Or "_option" if you have to.

src/features/query-bar/components/filter/Condition.js
```diff
@@ -102,33 +102,30 @@ const Condition = ({
...
                 return {
-                    displayName: option.key,
+                    displayName: o.key,
                     id: fieldId,
                     type: 'enum',
-                    value: optionIndex,
+                    value: idx,
```
macrotim: Shouldn't the value be option.key? Worth including in a unit test.

src/features/query-bar/components/filter/Condition.js
```diff
@@ -35,7 +35,6 @@ type Props = {
     condition: Object,
     deleteCondition: (index: number) => void,
     index: number,
-    intl: Object,
```
ChenCodes: No longer needed for this file

src/features/query-bar/flowTypes.js
```diff
@@ -3,7 +3,7 @@ import { AND, OR } from './constants';
 
 export type SelectOptionType = {
     displayText: string,
-    fieldId?: string,
+    fieldId: string,
     type?: string,
```
ChenCodes: fieldId should always be present given that a Column has been selected

ChenCodes: If a Template has not been selected, then the FilterButton would be disabled and thus not have to render any options

src/features/query-bar/components/filter/Condition.js
```diff
@@ -68,36 +67,39 @@ const Condition = ({
...
+            const { displayName, id, type } = option;
             return {
-                displayText: option.displayName,
-                fieldId: option.id,
-                type: option.type,
-                value: option.displayName,
+                displayText: displayName,
+                fieldId: id,
+                type,
+                value: displayName,
```
macrotim: I'm trying to decide if value=displayName is expected. But `options` is typed as Objects. 

Is this expected?

src/features/query-bar/flowTypes.js
```diff
@@ -3,7 +3,7 @@ import { AND, OR } from './constants';
 
 export type SelectOptionType = {
```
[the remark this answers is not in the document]
ChenCodes: OptionType works

src/features/query-bar/flowTypes.js
```diff
@@ -1,11 +1,13 @@
...
 
-export type SelectOptionType = {
+export type ConnectorType = typeof AND | typeof OR | typeof EMPTY_CONNECTOR;
+
+export type OptionType = {
     displayText: string,
     fieldId: string,
     type?: string,
-    value: string,
+    value: ConnectorType,
```
ChenCodes: Not sure if this type is correct. I tried using string | ConnectorType but that still resulted in an error for `handleConnectorChange` when trying to setState with `connector`.

ChenCodes: Using ConnectorType means that value can only be the string literals AND / OR and other strings are not allowed according to the flow message. Using Union of (string | ConnectorType) should technically allow value to be any string or AND or OR but flow still doesn't seem to like it.

krapspark: Did you try just `string`? when you do `string | ConnectorType` that basically boils down to `string`

src/features/query-bar/flowTypes.js
```diff
@@ -1,11 +1,13 @@
...
-export type SelectOptionType = {
+export type ConnectorType = typeof AND | typeof OR | typeof EMPTY_CONNECTOR;
+
+export type OptionType = {
     displayText: string,
-    fieldId?: string,
+    fieldId: string,
     type?: string,
-    value: string,
+    value: string | ConnectorType,
```
macrotim: Is it needed?

src/features/query-bar/components/fixtures.js
```diff
@@ -5,13 +5,41 @@ const initialCondition: ConditionType = {
     id: 0,
     columnDisplayText: '',
     columnKey: null,
+    fieldId: '3',
```
macrotim: This is odd. I don't see ConditionType.fieldId in flowTypes.

I see this in the Condition-test as well. What am I missing?

src/features/query-bar/flowTypes.js
```diff
@@ -1,11 +1,13 @@
...
-import { AND, OR } from './constants';
+import { AND, OR, EMPTY_CONNECTOR } from './constants';
 
-export type SelectOptionType = {
+export type ConnectorType = typeof AND | typeof OR | typeof EMPTY_CONNECTOR;
+
+export type OptionType = {
     displayText: string,
-    fieldId?: string,
+    fieldId: string,
```
macrotim: The name `fieldId` seems to incorrectly borrow from Template.Field. Could it be "id"?

src/features/query-bar/flowTypes.js
ChenCodes: It isn't needed but provides more context regarding the specific literals that value could be.

src/features/query-bar/flowTypes.js
ChenCodes: Yes

src/features/query-bar/components/filter/Condition.js
```diff
@@ -89,15 +89,15 @@ const Condition = ({
                 break;
         }
 
-        update(index, condition, displayText, displayTextType, fieldId, value, keyType, type);
+        update(index, condition, displayText, displayTextType, id, value, keyType, type);
```
macrotim: shouldn't "update" be named "onXYZChange"?

src/features/query-bar/components/filter/Condition.js
macrotim: Is this expected?

src/features/query-bar/components/filter/FilterButton.js
```diff
@@ -133,7 +133,7 @@ class FilterButton extends React.Component<Props, State> {
                 [fieldDisplayTextType]: fieldDisplayText,
                 [fieldKeyType]: fieldKey,
                 valueType,
-                fieldId,
+                fieldId: id,
```
macrotim: expected?

src/features/query-bar/flowTypes.js
```diff
@@ -24,6 +24,7 @@ export type ColumnType = {
 export type ConditionType = {
     columnDisplayText: string,
     columnKey: string | null,
+    fieldId: string,
```
macrotim: What's the purpose of this?

ChenCodes: I need fieldId in `getColumnOptions` in  Condition.js in order to find the corresponding column 

```
const getColumnOptions = () => {
        const { fieldId } = condition;
        const column = columns && columns.find(c => c.id === fieldId);
        if (column && column.options) {
            return column.options.map(option => {
                const { key } = option;
                return {
                    displayName: key,
                    id: fieldId,
                    type: 'enum',
                    value: key,
                };
            });
        }
        return [];
    };
```

src/features/query-bar/components/filter/FilterButton.js
ChenCodes: Yes, ConditionType has a property called fieldId but OptionType has a property called id. The right hand of the assignment is coming from OptionType.

src/features/query-bar/components/filter/Condition.js
ChenCodes: Yes, I can change it to be:

handleFieldChange / onFieldChange

src/features/query-bar/components/filter/Condition.js
ChenCodes: Yes `value = displayName` is expected. options is typed as Objects because it's possible for it to be either ColumnType, `{displayName: string, id: string, type: string, value: string}`, or `{displayName: string, value: ConnectorType}`.

src/features/query-bar/components/filter/FilterButton.js
macrotim: What does Condition.fieldId = Option.id mean? It isn't obvious to me.

src/features/query-bar/flowTypes.js
macrotim: Should ColumnType.options be Array<OptionType>? It's currently Array<Object>.

```c => c.id === fieldId```. Why does ColumnType.id map to ConditionType.fieldId? Why is the right side called fieldId? Is ConditionType.selectedColumnId a more appropriate name?

Alternatively, you already have ConditionType.columnKey. Is that not sufficient to lookup the Column? I'd prefer that we enable this instead stuffing in another property.

Comment (ChenCodes) [reply src/features/query-bar/flowTypes.js]: So ColumnType does not have a property called key, it has id, so I'm going to use columnId as a property instead of fieldId.

Comment (ChenCodes) [reply src/features/query-bar/flowTypes.js]: I will also be removing columnKey and replacing it with columnId.

File: src/features/query-bar/components/filter/Condition.js [hunk 17]
```diff
@@ -113,14 +111,13 @@ const Condition = ({
     };
 
     const getColumnOptions = () => {
-        const { fieldId } = condition;
-        const column = columns && columns.find(c => c.id === fieldId);
+        const { columnDisplayText } = condition;
+        const column = columns && columns.find(c => c.displayName === columnDisplayText);
```
ChenCodes: Instead of using fieldId to check which column we're looking for, we can instead use column.displayName to identify the column.

src/features/query-bar/components/fixtures.js
```diff
@@ -3,7 +3,7 @@ import type { ColumnType, ConditionType } from '../flowTypes';
 
 const initialCondition: ConditionType = {
     id: 0,
-    columnDisplayText: '',
+    columnDisplayText: 'Contract Value',
```
ChenCodes: The reason that I set 'Contract Value' here is that in the snapshot, I want to render a column that has an options property that is not null.

src/features/query-bar/__tests__/Condition-test.js
```diff
@@ -111,8 +111,8 @@ describe('features/query-bar/components/filter/Condition', () => {
     describe('getColumnOptions()', () => {
         const { fieldId } = initialCondition;
         test.each`
-            description                                                    | expectedColumnOptions
-            ${'user has opened the value dropdowhen and sees the options'} | ${columnOptions}
+            description                                                  | expectedColumnOptions
+            ${'user has opened the value dropdown and sees the options'} | ${columnOptions}
```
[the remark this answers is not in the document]
macrotim: It doesn't look resolved to me.

src/features/query-bar/__tests__/__snapshots__/Condition-test.js.snap
```diff
@@ -105,7 +95,30 @@ exports[`features/query-bar/components/filter/Condition render() should render C
       updateSelectedField={[Function]}
       updateValueField={[Function]}
       valueKey={null}
-      valueOptions={Array []}
+      valueOptions={
+        Array [
```
macrotim: I'd like to see the minimum number of options to demonstrate the expected behavior. Are 4 options necessary here?

src/features/query-bar/components/filter/Condition.js
```diff
@@ -58,88 +56,84 @@ const Condition = ({
...
+            case OPERATOR:
+                displayTextType = OPERATOR_DISPLAY_TEXT;
+                keyType = OPERATOR_KEY;
+                break;
+            case VALUE:
+                displayTextType = VALUE_DISPLAY_TEXT;
+                keyType = VALUE_KEY;
+                break;
+            default:
+                break;
```
macrotim: This will lead to silent errors and hard to debug issues.

Better to replace "break" with "throw new Exception('Invalid input');"

src/features/query-bar/components/filter/Condition.js
macrotim: I'd prefer columnKey. "Display text" is a view concern and should not drive business logic.

src/features/query-bar/__tests__/Condition-test.js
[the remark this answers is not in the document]
ChenCodes: Fixed

src/features/query-bar/__tests__/__snapshots__/Condition-test.js.snap
ChenCodes: I can cut it down to just one option

src/features/query-bar/__tests__/Condition-test.js
```diff
@@ -80,31 +78,40 @@ describe('features/query-bar/components/filter/Condition', () => {
...
             );
         });
     });
+
+    describe('getColumnOptions()', () => {
+        test.each`
+            description                                             | expectedColumnOptions
+            ${'should open the value dropdown and see the options'} | ${columnOptions}
+        `('$description', ({ expectedColumnOptions }) => {
+            const wrapper = getWrapper({});
```
macrotim: Is expectedColumnOptions somehow implicitly passed to the component? I think it should be explicit here.

src/features/query-bar/components/filter/Condition.js
```diff
@@ -58,88 +56,84 @@ const Condition = ({
...
 
     const getFormattedOptions = (options: Array<Object>): any[] => {
         return options.map(option => {
+            const { displayName, type } = option;
             return {
-                displayText: option.displayName,
-                fieldId: option.id,
-                type: option.type,
-                value: option.displayName,
+                displayText: displayName,
```
macrotim: Too bad the names are a bit inconsistent.

src/features/query-bar/components/filter/Condition.js
```diff
@@ -58,88 +56,84 @@ const Condition = ({
...
-            columns &&
-            columns.find(column => {
-                return column.id === fieldId;
-            });
-
-        if (field && field.options) {
-            const fieldOptions = field.options.map((option, optionIndex) => {
+    const getColumnOptions = () => {
+        const { columnKey } = condition;
+        const column = columns && columns.find(c => c.displayName === columnKey);
```
macrotim: ```c.displayName === columnKey```

Comparing a label with a key doesn't seem right.

src/features/query-bar/components/filter/Condition.js
ChenCodes: I checked with Loren regarding changing displayName to be displayText in the Column object but it looks like it'll involve changing 12~ files on both of our ends.

src/features/query-bar/components/filter/Condition.js
ChenCodes: Will be addressing this in a separate PR

PR #879, opened 2019-02-21
perf(sidebar): Virtualize activity feed

jeremypress:
Very much a WIP, but this PR switches the activity feed to use react-virtualized.

### DIFF
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
@@ -4,6 +4,10 @@
  */
 import * as React from 'react';
 import getProp from 'lodash/get';
+import List from 'react-virtualized/dist/commonjs/List';
+import AutoSizer from 'react-virtualized/dist/es/AutoSizer';
+import { CellMeasurer, CellMeasurerCache } from 'react-virtualized';
+
 import Comment from '../comment';
 import Task from '../task';
 import Version, { CollapsedVersion } from '../version';
@@ -25,30 +29,50 @@ type Props = {
     translations?: Translations,
 };
 
-const ActiveState = ({
-    currentUser,
-    items,
-    onCommentDelete,
-    onTaskDelete,
-    onTaskEdit,
-    onTaskAssignmentUpdate,
-    onVersionInfo,
-    translations,
-    getAvatarUrl,
-    getUserProfileUrl,
-    getMentionWithQuery,
-    mentionSelectorContacts,
-}: Props): React.Node => (
-    <ul className="bcs-activity-feed-active-state">
-        {items.map((item: any) => {
-            const { type, id, versions, permissions } = item;
+class ActiveState extends React.Component<Props> {
+    cache: any;
 
-            switch (type) {
-                case 'comment':
-                    return (
-                        <li key={type + id} className="bcs-activity-feed-comment">
+    list: any;
+
+    constructor() {
+        super();
+        this.cache = new CellMeasurerCache({
+            defaultHeight: 81,
+            fixedWidth: true,
+        });
+    }
+
+    componentDidUpdate() {
+        this.cache.clearAll();
+        this.list.forceUpdateGrid();
+    }
+
+    renderActivityCard = ({ key, index, parent, style }: any) => {
+        const {
+            items,
+            currentUser,
+            getAvatarUrl,
+            getUserProfileUrl,
+            onCommentDelete,
+            translations,
+            getMentionWithQuery,
+            mentionSelectorContacts,
+            onTaskAssignmentUpdate,
+            onTaskDelete,
+            onTaskEdit,
+            onVersionInfo,
+        } = this.props;
+
+        const item: any = items[index];
+        const { type, id, versions, permissions } = item;
+        switch (type) {
+            case 'comment':
+                return (
+                    <CellMeasurer cache={this.cache} columnIndex={0} key={key} parent={parent} rowIndex={index}>
+                        <li key={type + id} className="bcs-activity-feed-comment" style={style}>
                             <Comment
                                 {...item}
+                                cellMeasurerCache={this.cache}
                                 currentUser={currentUser}
                                 getAvatarUrl={getAvatarUrl}
                                 getUserProfileUrl={getUserProfileUrl}
@@ -60,13 +84,17 @@ const ActiveState = ({
                                 translations={translations}
                             />
                         </li>
-                    );
-                case 'task':
-                    return item.task_assignment_collection.total_count ? (
-                        <li key={type + id} className="bcs-activity-feed-task">
+                    </CellMeasurer>
+                );
+
+            case 'task':
+                return item.task_assignment_collection.total_count ? (
+                    <CellMeasurer cache={this.cache} columnIndex={0} key={key} parent={parent} rowIndex={index}>
+                        <li key={type + id} className="bcs-activity-feed-task" style={style}>
                             <Task
                                 {...item}
                                 currentUser={currentUser}
+                                cellMeasurerCache={this.cache}
                                 getAvatarUrl={getAvatarUrl}
                                 getMentionWithQuery={getMentionWithQuery}
                                 getUserProfileUrl={getUserProfileUrl}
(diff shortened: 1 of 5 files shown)


boxcla:

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -25,30 +29,50 @@ type Props = {
...
+            mentionSelectorContacts,
+            onTaskAssignmentUpdate,
+            onTaskDelete,
+            onTaskEdit,
+            onVersionInfo,
+        } = this.props;
+
+        const item: any = items[index];
+        const { type, id, versions, permissions } = item;
+        switch (type) {
```
jeremypress: This will become it's own component.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -25,30 +29,50 @@ type Props = {
...
+
+    constructor() {
+        super();
+        this.cache = new CellMeasurerCache({
+            defaultHeight: 81,
+            fixedWidth: true,
+        });
+    }
+
+    componentDidUpdate() {
```
jeremypress: We might be able to be more efficient here by only updating rows that changed.

macrotim: Else it'd re-render on any prop change.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -25,30 +29,50 @@ type Props = {
...
+        this.cache.clearAll();
+        this.list.forceUpdateGrid();
```
macrotim: It's a pretty heavy-handed approach so I'd ask if it's needed.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -60,13 +84,17 @@ const ActiveState = ({
...
+                );
+
+            case 'task':
+                return item.task_assignment_collection.total_count ? (
+                    <CellMeasurer cache={this.cache} columnIndex={0} key={key} parent={parent} rowIndex={index}>
+                        <li key={type + id} className="bcs-activity-feed-task" style={style}>
                             <Task
                                 {...item}
                                 currentUser={currentUser}
+                                cellMeasurerCache={this.cache}
```
macrotim: Does Task need the cache?

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
jeremypress: agreed, I'm working on a way to determine which row changed to only re-measure that item.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
jeremypress: nope, I'm about to push a change removing this. I was working on an internal action changing the height of a row.

alexkrolick: I'm actively working on the Task area so hopefully we don't have conflicts

jeremypress: I'm currently adding a prop, but otherwise the underlying components will not be changing too much.

priyajeet:
I don't know how much (if any at all) is a use case to search the comments. Right now you can use the search bar in the browser, but not possible without custom bar with virtualized list. Also scrolling to a comment ?comment=123 in the URL, will need another plugin for that iirc.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -82,29 +113,54 @@ const ActiveState = ({
...
+                        ref={ref => {
+                            this.list = ref;
+                        }}
+                        overscanRowCount={10}
+                        className="bcs-activity-feed-active-state-content"
+                        rowCount={this.props.items.length}
+                        height={height}
+                        rowHeight={this.cache.rowHeight}
+                        rowRenderer={this.renderActivityCard}
+                        width={340}
```
jstoffan: Is it possible to inherit this from the parent's width, specified in CSS?

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -82,29 +113,54 @@ const ActiveState = ({
...
+                return null;
+        }
+    };
+
+    render = (): React.Node => (
+        <ul className="bcs-activity-feed-active-state">
+            <AutoSizer disableWidth>
+                {({ height }) => (
+                    <List
+                        ref={ref => {
```
jstoffan: You could consider moving this to a class method.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -82,29 +113,54 @@ const ActiveState = ({
...
-    </ul>
-);
+                    </CellMeasurer>
+                );
+            default:
+                return null;
+        }
+    };
+
+    render = (): React.Node => (
```
jstoffan: Can this be a named function rather than an arrow function?

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -82,29 +113,54 @@ const ActiveState = ({
...
                         </li>
-                    );
-                case 'keywords':
-                    return (
-                        <li key={type + id} className="bcs-keywords-item">
+                    </CellMeasurer>
+                );
+            case 'keywords':
+                return (
+                    <CellMeasurer cache={this.cache} columnIndex={0} key={key} parent={parent} rowIndex={index}>
```
jstoffan: It looks like all four uses of `CellMeasurer` have the same prop values. Should it be extracted to its own simple component?

src/elements/content-sidebar/activity-feed/comment/Comment.js
```diff
@@ -71,7 +72,11 @@ class Comment extends React.Component<Props, State> {
 
     approvalCommentFormFocusHandler = (): void => this.setState({ isInputOpen: true });
 
-    approvalCommentFormCancelHandler = (): void => this.setState({ isInputOpen: false, isEditing: false });
+    approvalCommentFormCancelHandler = (): void =>
+        this.setState({ isInputOpen: false, isEditing: false }, () => {
+            const { onEditStart = noop } = this.props;
+            onEditStart();
```
jstoffan: Would this be something like `onEditStop`?

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -82,29 +113,54 @@ const ActiveState = ({
...
                         </li>
-                    ) : null;
-                case 'file_version':
-                    return (
-                        <li key={type + id} className="bcs-version-item">
+                    </CellMeasurer>
+                ) : null;
+            case 'file_version':
+                return (
+                    <CellMeasurer cache={this.cache} columnIndex={0} key={key} parent={parent} rowIndex={index}>
```
jstoffan: Does this item type need to be resized if its contents don't change?

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -25,28 +29,51 @@ type Props = {
...
+            onTaskEdit,
+            onVersionInfo,
+        } = this.props;
+
+        const item: any = items[index];
+        const { type, id, versions, permissions } = item;
+        switch (type) {
+            case 'comment':
+                return (
+                    <CellMeasurer cache={this.cache} columnIndex={0} key={key} parent={parent} rowIndex={index}>
```
jstoffan: I'm not super familiar with react-virtualized, but if I'm reading the [docs](https://github.com/bvaughn/react-virtualized/blob/master/docs/Grid.md) correctly, you might be able to pass `deferredMeasurementCache` to `List` to avoid passing `this.cache` around.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -25,28 +29,51 @@ type Props = {
...
+    componentDidUpdate() {
+        this.cache.clearAll();
+        this.list.forceUpdateGrid();
+    }
 
-            switch (type) {
-                case 'comment':
-                    return (
-                        <li key={type + id} className="bcs-activity-feed-comment">
+    onEditChange = () => {
```
jstoffan: It's a little unfortunate to have to pass this down to each component to be called when the "mode" changes. Maybe we could use context and detect when a specific state property (such as `isEditing`) in each activity item changes, then trigger the resize logic?

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -25,28 +29,51 @@ type Props = {
...
+        this.cache.clearAll();
+        this.list.forceUpdateGrid();
```
jstoffan: Would we want to call [this.list.recomputeRowHeights](https://github.com/bvaughn/react-virtualized/blob/master/docs/List.md#recomputerowheights-index-number), instead?

PR #880, opened 2019-02-22
fix(collaborator-avatars): Decouple Sass from UnifiedShareModal

nickgriffin1:
A fix for the following commit: https://github.com/box/box-ui-elements/commit/7073b499655b9644648f47d6a1163c22006a997b
Some of the Sass was missed in the refactor and must be moved to it's own file.


boxcla:

src/features/collaborator-avatars/CollaboratorList.scss
```diff
@@ -0,0 +1,84 @@
+@import '../../styles/variables';
+
+.collaborator-list {
```
priyajeet: Should this be scoped under `usm-collaborator-list` as the render() shows?

src/features/collaborator-avatars/CollaboratorList.scss
```diff
@@ -0,0 +1,84 @@
...
+    border: solid 1px $seesee;
+    border-radius: 4px;
+    box-shadow: 0 2px 3px 0 rgba(0, 0, 0, .08) inset;
+    height: 210px;
+    margin: 0;
+    min-width: 400px;
+    overflow-y: scroll;
+
+    .collaborator-list-item {
```
priyajeet: This should ideally be part of its own scss imported within its parent component

src/features/collaborator-avatars/CollaboratorList.scss
nickgriffin1: It's probably a good idea to give it more scope. I'll add that.

src/features/collaborator-avatars/CollaboratorList.scss
nickgriffin1: I split this into a second file. Let me know if I did that correctly. Why the parent instead of the component the styles are for?

priyajeet: Oh I meant the parent component for the styles which in this case would be CollaboratorListItem

nickgriffin1: Gotcha. I'll move that to the correct component.

PR #886, opened 2019-02-26
chore: add metadataListViewMessage component, got lost in shuffle

Visylvius:
What this PR Does: 

Adds the metadataListViewMessage component to the list-view directory, 

This will be added to the examples page, but i'd like to wait a little longer until the component is more finalized.

Questions: 
Should we start slowly removing the metadata-view directory?

Also, there is a test that uses a native array forEach, in the MetadataListViewMessage-test. if you would prefer I will refactor that to using a test.each


boxcla:

src/features/list-view/__tests__/MetadataListViewMessage-test.js
```diff
@@ -1,7 +1,7 @@
 // @flow
 import * as React from 'react';
 
-import Message from '../Message';
+import Message from '../components/MetadataListViewMessage';
```
Visylvius: The test at the bottom of this file uses a native Array forEach, if you would prefer I can refactor into a test.each

src/features/list-view/styles/MetadataListViewMessage.scss
```diff
@@ -0,0 +1,26 @@
+.metadata-view-header-text-container {
```
Visylvius: these class name will also need to be changed once the naming convention is finalized, as they do not currently reflect the current state of the component.

priyajeet:
Feel free to nuke out deprecated folders

src/features/list-view/__tests__/MetadataListViewMessage-test.js
priyajeet: yeah would be nice to be consistent with that, so test.each

PR #887, opened 2019-02-26
feat(appactivity): Add AppActivity API Layer

JustinHoldstock:


boxcla:

src/api/AppActivity.js
```diff
@@ -0,0 +1,144 @@
...
+].join(',');
+
+/**
+ * Map an entry from the AppActivity API to an AppActivityItem.
+ * occurred_at -> created_at
+ *
+ * @param {Object} item - A single entry in the AppActivity API entiries list
+ */
+function mapAppActivityItem(item: Object): AppActivityItem {
+    const { occurred_at, ...wOutOccurredAt } = item;
```
ConradJChan: `wOutOccurredAt` is a bit interesting, how about something more traditional like `rest`?

src/api/AppActivity.js
```diff
@@ -0,0 +1,144 @@
...
+     * Generic success handler
+     *
+     * @param {Object} data - the response data
+     */
+    successHandler = (data: Object): void => {
+        if (this.isDestroyed() || typeof this.successCallback !== 'function') {
+            return;
+        }
+
+        const { entries = [] } = data;
```
jstoffan: Looks like this could live in the method signature.

src/api/AppActivity.js
```diff
@@ -0,0 +1,144 @@
...
+
+        return `${this.getAppActivityUrl()}/${id}`;
+    }
+
+    /**
+     * Generic success handler
+     *
+     * @param {Object} data - the response data
+     */
+    successHandler = (data: Object): void => {
```
jstoffan: I think we're trying to avoid `Object` and `Function` where possible.

src/api/Feed.js
```diff
@@ -943,6 +949,70 @@ class Feed extends Base {
...
+        permissions: BoxItemPermission,
+        successCallback: Function,
+        errorCallback: ErrorCallback,
+    ): void => {
+        this.appActivityAPI = new AppActivityAPI(this.options);
+        if (!file.id) {
+            throw getBadItemError();
+        }
+
+        const { id: activityItemId } = appActivityItem.id;
```
ConradJChan: Should this just be `const { id: activityItemId } = appActivityItem;`

src/elements/common/messages.js
```diff
@@ -788,10 +788,15 @@ const messages: { [string]: MessageDescriptor } = defineMessages({
         defaultMessage: 'Completed',
         description: 'Title for checkmark icon indicating someone completed a task',
     },
+    appActivityDeleteErrorMessage: {
+        id: 'be.appActivityDeleteErrorMessage',
+        description: 'Error message when an app activity deletion fails',
+        defaultMessage: 'There was an error deleting this app activity item.',
```
jeremypress: Are we displaying the string "app activity" anywhere? A user might confused as to what an app activity item is vs. just saying item.

src/api/Feed.js
JustinHoldstock: Good catch, thanks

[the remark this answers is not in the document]
JustinHoldstock: Good catch, thanks

src/api/AppActivity.js
JustinHoldstock: 👍 Moved to use AppActivityItems instead of Object.

src/api/AppActivity.js
```diff
@@ -0,0 +1,144 @@
...
+        errorCallback,
+    }: {
+        appActivityItem: AppActivityItem,
+        errorCallback: Function,
+        file: BoxItem,
+        successCallback: Function,
+    }): void {
+        this.errorCode = ERROR_CODE_DELETE_APP_ACTIVITY;
+
+        const { id = '' } = file;
```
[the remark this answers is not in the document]
JustinHoldstock: I guess there really is no need to default. The file id is unnecessary for the delete request, as deletion happens via the AppActivity item's ID

src/api/AppActivity.js
```diff
@@ -0,0 +1,141 @@
...
+import {
+    ERROR_CODE_DELETE_APP_ACTIVITY,
+    FIELD_ACTIVITY_TEMPLATE,
+    FIELD_APP,
+    FIELD_CREATED_BY,
+    FIELD_OCCURRED_AT,
+    FIELD_RENDERED_TEXT,
+} from '../constants';
+
+const APP_ACTIVITY_FIELDS: string = [
```
jeremypress: toString will do this for you as well. Also, we have a fields util method you could add this array to.

src/api/AppActivity.js
```diff
@@ -0,0 +1,129 @@
...
+     * @return {string} Url for all app activity on a file
+     */
+    getUrl(id: string): string {
+        if (!id) {
+            throw new Error('Missing file id!');
+        }
+
+        return `${this.getAppActivityUrl(
+            id,
+        )}?item_id=${id}&item_type=file&fields=${APP_ACTIVITY_FIELDS_TO_FETCH.toString()}`;
```
jstoffan: It seems like we should be passing these values an object to `this.markerGet` as part of `requestData` rather than doing string concatenation.

JustinHoldstock: That would be a heck of a lot cleaner. Can do!

src/api/AppActivity.js
```diff
@@ -0,0 +1,120 @@
...
+    getAppActivity(id: string, successCallback: Function, errorCallback: ElementsErrorCallback, limit?: number): void {
+        const requestData = {
+            item_id: id,
+            item_type: 'file',
+            fields: APP_ACTIVITY_FIELDS_TO_FETCH.toString(),
+        };
+
+        this.markerGet({
+            id,
+            limit,
```
jeremypress: if limit is not passed in are we passing `null: null` to marketGet?

jeremypress: if we're including it to be generic, then I would argue that the fields should be appended in feed where we make the call. We don't know what fields will be needed by a different caller,

JustinHoldstock: By default, `limit` is set to 1000, for`markerGet()`. Are you suggesting we just remove it, or *also* set it to 1000 inside of Feed.js?

src/api/AppActivity.js
```diff
@@ -0,0 +1,160 @@
...
+            return;
+        }
+
+        this.permissions = {};
+        const { response } = error;
+
+        // In the case of a 404, the enterprise does not have App Activities enabled.
+        // Show no App Activity
+        if (response.status === HTTP_STATUS_CODE_NOT_FOUND) {
+            this.successHandler({});
```
jeremypress: The success handler expects an item of type AppActivityItems. Could you change the type to make the fields option or pass them in blank here?

total: 0
entries: []

src/api/AppActivity.js
```diff
@@ -0,0 +1,160 @@
...
+     * Generic error handler
+     *
+     * @param {AxiosError} error - the error response
+     */
+    errorHandler = (error: $AxiosError<any>): void => {
+        if (this.isDestroyed() && typeof this.errorCallback !== 'function') {
+            return;
+        }
+
+        this.permissions = {};
```
jeremypress: Why do we need to wipe the permissions? It looks like it gets reset on every call to `getAppActivity`.

JustinHoldstock: In the case of an error, I thought a reset would be safe, but now I'm not convinced it makes a difference (as you've pointed out)

src/api/AppActivity.js
```diff
@@ -0,0 +1,159 @@
...
+    /**
+     * Map an entry from the AppActivity API to an AppActivityItem.
+     * occurred_at -> created_at
+     * Adds permissions to item
+     *
+     * @param {Object} item - A single entry in the AppActivity API entries list
+     *
+     * @return {AppActivityItem}
+     */
+    mapAppActivityItem = (item: Object): AppActivityItem => {
```
jstoffan: Can we avoid `Object` here?

src/api/MarkerBasedAPI.js
```diff
@@ -49,6 +50,7 @@ class MarkerBasedApi extends Base {
         limit: number,
         shouldFetchAll: boolean,
         requestData: Object = {},
+        headers?: StringMap = {},
```
jstoffan: Is this change necessary? It doesn't look like this is used.

src/api/AppActivity.js
JustinHoldstock: I'll whip up another type for the API response shape

src/api/MarkerBasedAPI.js
JustinHoldstock: I meant to remove those after removing `Accept-Language` header pass-through. Thanks!

src/api/Feed.js
```diff
@@ -1201,16 +1207,18 @@ class Feed extends Base {
...
         if (cachedItems) {
-            const updatedFeedItems = cachedItems.items.map((item: Comment | Task | TaskNew | BoxItemVersion) => {
-                if (item.id === id) {
-                    return {
-                        ...item,
-                        ...updates,
-                    };
-                }
+            const updatedFeedItems = cachedItems.items.map(
+                (item: Comment | Task | TaskNew | BoxItemVersion | AppActivityItem) => {
```
jstoffan: It looks like this is duplicated in the `FeedItems` type declaration. Can it be shared via a `FeedItem` type?

src/elements/content-sidebar/ContentSidebar.js
```diff
@@ -118,6 +119,7 @@ class ContentSidebar extends React.Component<Props, State> {
             apiHost,
             cache,
             clientName,
+            language,
```
jstoffan: Is this still needed?

src/api/__tests__/AppActivity-test.js
```diff
@@ -0,0 +1,191 @@
...
+            const successCallback = jest.fn();
+            const errorCallback = jest.fn();
+            const language = 'ja-JP';
+            const requestData = {
+                fields: APP_ACTIVITY_FIELDS_TO_FETCH.toString(),
+                item_id: id,
+                item_type: 'file',
+            };
+
+            appActivity.getAppActivity(id, {}, successCallback, errorCallback, language);
```
jstoffan: It doesn't look like `getAppActivity` supports the `language` argument.

src/api/Feed.js
```diff
@@ -1344,6 +1352,74 @@ class Feed extends Base {
...
+        appActivityId: string,
+        successCallback: Function,
+        errorCallback: ErrorCallback,
+    ): void => {
+        this.appActivityAPI = new AppActivityAPI(this.options);
+        if (!file.id) {
+            throw getBadItemError();
+        }
+
+        const { id } = file;
```
jstoffan: Should this be moved to the first line so it can be used in the `!file.id` check?

src/api/Feed.js
```diff
@@ -1344,6 +1352,74 @@ class Feed extends Base {
...
+     * @param {Function} errorCallback - the function which will be called on error
+     * @return {void}
+     */
+    deleteAppActivity = (
+        file: BoxItem,
+        appActivityId: string,
+        successCallback: Function,
+        errorCallback: ErrorCallback,
+    ): void => {
+        this.appActivityAPI = new AppActivityAPI(this.options);
```
jstoffan: Should this be moved after the `!file.id` check?

src/api/Feed.js
JustinHoldstock: Sounds great!

src/elements/content-sidebar/ContentSidebar.js
JustinHoldstock: Removed, as mentioned in rendering pr

PR #888, opened 2019-02-26
feat(unified-share-modal): remove beta badge from USM

booc0mtaco:
Update tests, components, and imports to no longer render this badge.


boxcla:

priyajeet:
feat(unified-share-modal)

priyajeet:
BTW, is usm becoming the default? And if so, can we nuke the older versions?

booc0mtaco:
@priyajeet Is there documentation for what scope names are preferred? one rule could be that it should make the component's folder name when specified. 

I think this unified one will eventually be the only one available. I don't suppose we would have to worry about this being used since there was no interface for seeing it until recently. Once notes has this modal, it should be safe to remove the others

priyajeet:
Yeah thats what I was thinking, just use the folder name.

booc0mtaco:
Reminder to update scope on squash and merge as titles aren't reused for merged commits

PR #891, opened 2019-02-26
refactor(text-input): remove tag styles for inputs

priyajeet:
Does not fix other components using text-inputs or text-areas

BREAKING CHANGE: The styling classes for text inputs have changed.

### DIFF
--- a/src/styles/common/_typography.scss
+++ b/src/styles/common/_typography.scss
@@ -23,47 +23,6 @@ p {
     margin: 0 0 20px;
 }
 
-h1,
-h2,
-h3,
-h4,
-h5,
-h6 {
-    color: inherit;
-    font-family: inherit;
-    font-weight: 400;
-    line-height: 20px;
-    margin: 8px 0;
-    text-rendering: optimizeLegibility;
-}
-
-h1 {
-    font-size: 30px;
-    line-height: 36px;
-}
-
-h2 {
-    font-size: 24px;
-    line-height: 30px;
-}
-
-h3 {
-    font-size: 20px;
-    line-height: 24px;
-}
-
-h4 {
-    font-size: 16px;
-}
-
-h5 {
-    font-size: 14px;
-}
-
-h6 {
-    font-size: 12px;
-}
-
 .strong,
 strong {
     font-weight: 700;
--- a/conf/styleguide.config.js
+++ b/conf/styleguide.config.js
@@ -28,6 +28,7 @@ const allSections = [
     {
         name: 'Components',
         components: () => [
+            '../src/components/heading/Heading.js',
             '../src/components/avatar/Avatar.js',
             '../src/components/badge/Badge.js',
             '../src/components/badgeable/Badgeable.js',
--- a/src/components/heading/Heading.js
+++ b/src/components/heading/Heading.js
@@ -1,0 +1,25 @@
+// @flow
+import * as React from 'react';
+import classNames from 'classnames';
+import './Heading.scss';
+
+type Props = {
+    /** Heading contents */
+    children: React.Node,
+    /** Heading contents */
+    className?: string,
+    /** Header tag type */
+    type: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6',
+};
+
+const Heading = ({ children, className = '', type, ...rest }: Props) =>
+    React.createElement(
+        type,
+        {
+            className: classNames(`bdl-${type}`, className),
+            ...rest,
+        },
+        ...children,
+    );
+
+export default Heading;
--- a/src/components/heading/Heading.scss
+++ b/src/components/heading/Heading.scss
@@ -1,0 +1,44 @@
+@import '../../styles/variables';
+
+.bdl-h1,
+.bdl-h2,
+.bdl-h3,
+.bdl-h4,
+.bdl-h5,
+.bdl-h6 {
+    color: inherit;
+    font-family: inherit;
+    font-size: 100%;
+    font-weight: 400;
+    line-height: 20px;
+    margin: 8px 0;
+    padding: 0;
+    text-rendering: optimizeLegibility;
+}
+
+.bdl-h1 {
+    font-size: 30px;
+    line-height: 36px;
+}
+
+.bdl-h2 {
+    font-size: 24px;
+    line-height: 30px;
+}
+
+.bdl-h3 {
+    font-size: 20px;
+    line-height: 24px;
+}
+
+.bdl-h4 {
+    font-size: 16px;
+}
+
+.bdl-h5 {
+    font-size: 14px;
+}
+
+.bdl-h6 {
+    font-size: 12px;
+}
--- a/src/components/heading/README.md
+++ b/src/components/heading/README.md
@@ -1,0 +1,42 @@
+### Examples
+**H1**
+```
+<Heading type='h1'>
+    Box Design Language
+</Heading>
+```
+
+**H2**
+```
+<Heading type='h2'>
+    Box Design Language
+</Heading>
+```
+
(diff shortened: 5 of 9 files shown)


boxcla:

src/components/text-input/TextInput.js
```diff
@@ -46,29 +45,28 @@ const TextInput = ({
     labelTooltip,
     ...rest
 }: Props) => {
-    const classes = classNames(className, 'text-input-container', {
-        'show-error': !!error,
+    const classes = classNames(className, 'bdl-TextInput-container', {
+        'bdl-hasError': !!error,
```
priyajeet: No TextInput- for this as this is a state that other places can re-use. Downside of this is that people should only have it `&`ed and never styled on its own.

src/components/text-input/TextInput.scss
```diff
@@ -1,55 +1,105 @@
...
+@import '../../styles/inputs';
 
-/**************************************
- * Text Input
- **************************************/
+.bdl-TextInput-input {
+    @include common-typography;
+
+    height: 32px;
+    margin-top: 5px;
```
priyajeet: Margin should likely not be there, the component should not touch its outside

src/components/text-input/TextInput.scss
```diff
@@ -1,55 +1,105 @@
 @import '../../styles/variables';
+@import '../../styles/inputs';
 
-/**************************************
- * Text Input
- **************************************/
+.bdl-TextInput-input {
```
priyajeet: This is not scoped inside `.bdl-TextInput-container` so that this class can be used on its own if needed. For some strangling html `<input>`.

src/components/text-input/TextInput.scss
```diff
@@ -1,55 +1,105 @@
...
+    @include common-typography;
```
priyajeet: For similar reasons of not being nested inside a container, and also usefulness inside a tethered component

src/styles/common/_typography.scss
```diff
@@ -110,38 +109,3 @@ textarea {
...
-
-.txt-medium {
-    font-size: 14px;
-}
-
-.txt-large {
-    font-size: 16px;
-}
-
-.txt-x-large {
```
priyajeet: Should not be using all this stuff

src/styles/common/_forms.scss
```diff
@@ -4,144 +4,6 @@
...
-    vertical-align: top;
-}
-
-.input-x-small,
-.input-x-small[type] {
-    width: 100px;
-}
-
-.input-small,
-.input-small[type] {
```
priyajeet: Should not be using all this stuff

src/components/text-input/TextInput.scss
```diff
@@ -1,55 +1,105 @@
...
+    &[type="email"],
+    &[type="url"],
+    &[type="tel"],
+    &[type="number"] {
+        @include box-inputs;
+
+        width: 262px;
+
+        &:hover {
+            @include box-inputs-hover;
```
priyajeet: @includes will be coming from _textInputs.scss, so these will be text-input-styles, text-input-hover-styles etc.

src/components/text-input/TextInput.js
```diff
@@ -46,29 +45,28 @@ const TextInput = ({
...
                 hideLabel={hideLabel}
                 showOptionalText={!hideOptionalLabel && !isRequired}
                 text={label}
                 tooltip={labelTooltip}
             >
-                {!!description && <i className="text-input-description">{description}</i>}
+                {!!description && <i className="bdl-TextInput-description">{description}</i>}
                 <Tooltip isShown={!!error} position={errorPosition || 'middle-right'} text={error || ''} theme="error">
-                    <input ref={inputRef} required={isRequired} {...rest} />
+                    <input {...rest} className="bdl-TextInput-input" ref={inputRef} required={isRequired} />
```
jstoffan: Moving `{...rest}` seems like a risky-ish change. Moving it doesn't seem related to styling, since the class names are stripped out in the method signature.

src/components/label/StandardLabel.js
```diff
@@ -5,13 +5,19 @@ import Tooltip from '../tooltip';
 import LabelPrimitive from './LabelPrimitive';
 
 type Props = {
+    /** Add a class to the component */
```
jstoffan: This looks mispositioned.

src/components/text-input/TextInput.scss
jstoffan: Are we using `.bdl-` as the new standard prefix for common styles? Are we using BEM going forward, as well, with `TextInput` being BookCase?

src/styles/common/_forms.scss
```diff
@@ -4,144 +4,6 @@
...
-
-input[type="text"].is-invalid,
-input[type="password"].is-invalid,
-input[type="search"].is-invalid,
-input[type="email"].is-invalid,
-input[type="url"].is-invalid,
-input[type="tel"].is-invalid,
-input[type="number"].is-invalid,
-div[contentEditable="true"].is-invalid,
-textarea.is-invalid {
```
priyajeet: Dont think is-invalid is used

src/components/text-input/TextInput.js
```diff
@@ -46,29 +45,28 @@ const TextInput = ({
...
-    const classes = classNames(className, 'text-input-container', {
-        'show-error': !!error,
+    const classes = classNames(className, 'bdl-TextInput-container', {
+        'bdl-hasError': !!error,
     });
 
     return (
         <div className={classes}>
             <Label
+                className="bdl-TextInput-label"
```
priyajeet: This PR is using https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md as naming example

src/components/text-input/TextInput.scss
priyajeet: This example is following SUIT css style. Still TBD if we will use SUIT vs BEM.

src/components/text-input/TextInput.scss
```diff
@@ -1,55 +1,105 @@
...
     position: relative;
 
-    .text-input-loading {
+    &.bdl-hasError {
+        .bdl-TextInput-input {
+            border-color: $amaranth;
+        }
+    }
+
+    .bdl-TextInput-loadingIndicator {
```
priyajeet: All these `bdl-TextInput-FOO` styles here can/should be replaced with SCSS convention `&-FOO`

priyajeet:
Linting for this is a bit annoynig with poor examples/docs. But roughly using postcss-bem-linter the rules are like

```js
const customSuitSelector = component => {
    const namespace = 'bdl';
    const block = component; // [A-Z]+
    const camelCase = '[a-z][a-zA-Z0-9]*';
    const state = `(?:.${namespace}-(is|has)-${camelCase})?`;
    const element = `(?:-${camelCase})?`;
    const modifier = `(?:--${camelCase})?`;
    const attribute = '(?:\\[.+\\])?';
    return new RegExp(`^\\.${namespace}-${block}${element}${modifier}${attribute}${state}$`);
};
```

src/components/text-input/TextInput.scss
booc0mtaco: How far would we want to go with this? For example, `${namespace}-${ComponentName}` is the top-level selector name? or just `${namespace}` with `-${ComponentName}` immediately after as a nested partial selector?

priyajeet: Based on some feedback we might as well just keep full strings for searchability. Since concatenation can make that harder

priyajeet:
I busted my master branch, so had to open a diff draft PR for this

PR #892, opened 2019-02-26
feat(sidebar): Render additional provided tabs

jeremypress:
screen shot 2019-02-25 at 4 17 30 pm

### DIFF
--- a/flow-typed/box-ui-elements.js
+++ b/flow-typed/box-ui-elements.js
@@ -697,7 +697,7 @@ type Integration = {
 };
 
 type AdditionalTab = {
-    callback: typeof Function,
+    callback: (id: string, callbackData: Object) => void,
     iconUrl: string,
     id: number,
     title: ?string,
--- a/flow-typed/box-ui-elements.js
+++ b/flow-typed/box-ui-elements.js
@@ -697,8 +697,8 @@ type Integration = {
 };
 
 type AdditionalTab = {
-    callback: (id: string, callbackData: Object) => void,
-    iconUrl: string,
+    callback: (id: number, callbackData: Object) => void,
+    iconUrl?: string,
     id: number,
     title: ?string,
 };
--- a/flow-typed/box-ui-elements.js
+++ b/flow-typed/box-ui-elements.js
@@ -696,6 +696,15 @@ type Integration = {
     type: APP_INTEGRATION,
 };
 
+type AdditionalTab = {
+    callback: typeof Function,
+    element: React.Element<any>,
+    id: number,
+    title: ?string,
+};
+
+type AdditionalTabs = Array<AdditionalTab>;
+
 type Alignment = 'left' | 'right';
 
 type SidebarView =
--- a/flow-typed/box-ui-elements.js
+++ b/flow-typed/box-ui-elements.js
@@ -698,7 +698,7 @@ type Integration = {
 
 type AdditionalTab = {
     callback: typeof Function,
-    element: React.Element<any>,
+    iconUrl: string,
     id: number,
     title: ?string,
 };
--- a/src/elements/content-sidebar/AdditionalTab.js
+++ b/src/elements/content-sidebar/AdditionalTab.js
@@ -1,0 +1,22 @@
+/**
+ * @flow
+ * @file Preview sidebar additional tab component
+ * @author Box
+ */
+
+import * as React from 'react';
+import noop from 'lodash/noop';
+import Tooltip from '../common/Tooltip';
+import PlainButton from '../../components/plain-button/PlainButton';
+
+const AdditionalTab = ({ callback: callbackFn = noop, element, id, title }: AdditionalTab) => {
+    return (
+        <Tooltip position="middle-left" text={title}>
+            <PlainButton className="bcs-additional-tab-btn bcs-nav-btn" type="button" onClick={() => callbackFn(id)}>
+                {element}
+            </PlainButton>
+        </Tooltip>
+    );
+};
+
+export default AdditionalTab;
--- a/src/elements/content-sidebar/AdditionalTab.js
+++ b/src/elements/content-sidebar/AdditionalTab.js
@@ -8,6 +8,8 @@ import * as React from 'react';
 import noop from 'lodash/noop';
 import Tooltip from '../common/Tooltip';
 import PlainButton from '../../components/plain-button/PlainButton';
+import IconEllipsis from '../../icons/general/IconEllipsis';
+
 import './AdditionalTab.scss';
 
 const AdditionalTab = ({ callback: callbackFn = noop, iconUrl, id, title, ...rest }: AdditionalTab) => {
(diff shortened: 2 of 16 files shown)


boxcla:

test/sidebar-additional-tabs.html
```diff
@@ -0,0 +1,120 @@
...
+
+                    localStorage.setItem('file', fileId);
+                    document.querySelector('.file').value = fileId;
+
+                    const placeholder = document.createElement('div');
+                    placeholder.textContent = 'test';
+
+                    const sidebar = new ContentSidebar();
+                    sidebar.show(fileId, token, {
+                        hasAdditionalTabs: true,
```
priyajeet: hasRecommendedApps and recommendedApps?

flow-typed/box-ui-elements.js
```diff
@@ -696,6 +696,15 @@ type Integration = {
     type: APP_INTEGRATION,
 };
 
+type AdditionalTab = {
+    callback: typeof Function,
+    element: React.Element<any>,
+    id: number,
+    title: ?string,
```
priyajeet: status?

test/sidebar-additional-tabs.html
```diff
@@ -0,0 +1,120 @@
...
+                                    id: 200,
+                                    title: 'Open with Gmail', // used as tooltip
+                                    element: Tab('gmail'), //<RecommendedAppTab />,
+                                    callback: () => window.alert('gmail'),
+                                    status: 'ADDED',
+                                },
+                                {
+                                    id: 201,
+                                    title: 'Open with Slack',
+                                    element: Tab('slack'), //<RecommendedAppTab />,
```
priyajeet: Not sure if we should allow insertion of React.Nodes from the caller app. Maybe we keep a registry/map of the apps we want to be showing and just extract it out from there. Similar to how we do our icons.

flow-typed/box-ui-elements.js
jeremypress: I realized status (or any other prop) they care about that's not needed for rendering doesn't need to be typed by us. We'll call their callback with whatever props they give us, but the sidebar doesn't actually care about what they are.

test/sidebar-additional-tabs.html
jeremypress: Wanted to keep it generic in case a platform customer/user uses it for something that's not recommended apps. Could be useful to have some extra buttons in the sidebar nav that do custom stuff.

src/elements/content-sidebar/AdditionalTabs.scss
```diff
@@ -0,0 +1,7 @@
+@import '../../styles/variables';
+
+.bcs-nav-separator {
+    border-top: 1px solid $efive;
+    margin: 16px 8px 0 8px;
+    padding-bottom: 16px;
```
ConradJChan: Could this be rolled into the `margin` style?

src/elements/content-sidebar/AdditionalTabs.js
```diff
@@ -0,0 +1,25 @@
...
+
+type Props = {
+    tabs: AdditionalTabs,
+};
+
+const AdditionalSidebarTabs = ({ tabs }: Props) => {
+    const sidebarTabs = tabs.map(tabData => <AdditionalTab key={tabData.id} {...tabData} />);
+    return (
+        <React.Fragment>
+            <div className="bcs-nav-separator" />
```
jstoffan: Can we wrap this component in a regular `div` and give it a pseudo-element (`::before`) to achieve the visual separation?

test/sidebar-additional-tabs.html
jeremypress: 👍 Going to switch to imageURLs for now, if there's significant demand from platform customers for custom markup here we can reconsider nodes.

src/elements/content-sidebar/SidebarNav.js
```diff
@@ -72,6 +84,7 @@ const SidebarNav = ({ hasSkills, hasMetadata, hasActivityFeed, hasDetails, onTog
                 <IconMetadataThick />
             </SidebarNavButton>
         )}
+        {hasAdditionalTabs && additionalTabs && <AdditionalTabs tabs={additionalTabs} />}
```
ctuckz: Are we going to need an additional isLoading property to handle showing the loading state?

jeremypress: my thought was that if the hasAdditionalTabs prop is set to true:
1. If we don't have any additional tabs data we are loading
2. If we have valid data load it
3. If we have an empty array then render the empty state

test/sidebar-additional-tabs.html
jeremypress: @ctuckz this will change our data structure a bit, instead of passing in a react component we can just pass in the URL now.

ctuckz: How are we going to handle resin tags if we're just passing in the URL? Is putting the tags in the ui element an option, or do they need to be in EUA?

src/elements/content-sidebar/SidebarNav.js
ctuckz: Sounds good!

test/sidebar-additional-tabs.html
jeremypress: There's a way to programatically fire resin events via js as "clicks". We can do this in the callback.

If that doesn't work for you we can add an "interactionTarget" prop, and included that on the wrapper component that's in the sidebar.

ctuckz: I didn't realize we could do it programmatically - that should work fine

src/elements/content-sidebar/AdditionalTab.js
```diff
@@ -0,0 +1,27 @@
...
+
+const AdditionalTab = ({ callback: callbackFn = noop, iconUrl, id, title, ...rest }: AdditionalTab) => {
+    return (
+        <Tooltip position="middle-left" text={title}>
+            <PlainButton
+                className="bcs-additional-tab-btn bcs-nav-btn"
+                type="button"
+                onClick={() => callbackFn({ id, rest })}
+            >
+                <img className="bcs-additional-tab-icon" src={iconUrl} alt={title} />
```
jeremypress: @JustinHoldstock I don't think this is worth a shared component, it's pretty small.

src/elements/content-sidebar/AdditionalTab.scss
```diff
@@ -0,0 +1,4 @@
+.bcs-additional-tab-icon {
+    height: 24px;
```
jeremypress: This will enforce a size if a partner gives us something too big or too small.

src/elements/content-sidebar/AdditionalTab.js
```diff
@@ -0,0 +1,27 @@
...
+import PlainButton from '../../components/plain-button/PlainButton';
+import './AdditionalTab.scss';
+
+const AdditionalTab = ({ callback: callbackFn = noop, iconUrl, id, title, ...rest }: AdditionalTab) => {
+    return (
+        <Tooltip position="middle-left" text={title}>
+            <PlainButton
+                className="bcs-additional-tab-btn bcs-nav-btn"
+                type="button"
+                onClick={() => callbackFn({ id, rest })}
```
jstoffan: This looks like a typo, since `...rest` is usually re-spread in callbacks. You may want to rename this as `callbackData` or something to indicate it's intended to be passed-through as an object literal.

src/elements/content-sidebar/AdditionalTab.scss
```diff
@@ -1,4 +1,10 @@
+@import '../../styles/variables';
+
 .bcs-additional-tab-icon {
     height: 24px;
     width: 24px;
 }
+
+.bcs-additional-tab-more-icon path {
```
jstoffan: Can you use `.fill-color` here instead of `path`?

PR #893, opened 2019-02-27
feat(tasks): new task card layout

alexkrolick:
Changes are feature-flagged as new features are added for the new API

**New**

screen shot 2019-02-26 at 6 19 10 pm

**Old**
screen shot 2019-02-26 at 6 18 22 pm


boxcla:

package.json
```diff
@@ -167,8 +170,8 @@
         "cypress": "^3.1.5",
         "deepmerge": "^2.1.1",
         "draft-js": "^0.10.1",
-        "enzyme": "^3.6.0",
-        "enzyme-adapter-react-16": "^1.5.0",
+        "enzyme": "3.8.0",
```
alexkrolick: Latest is 3.9.0 but there is a new bug that affected one of our tests

flow-typed/box-ui-elements.js
```diff
@@ -618,6 +618,111 @@ type Tasks = {
     total_count: number,
 };
 
+/* New Task Types START */
+
+type ID = string;
+type ISODate = string;
+
+type UserMini = {
```
JustinHoldstock: Was there a reason you decided to created a new type, instead of using `User`? They appear to be identical

alexkrolick: I guess they are... I can only guarantee accuracy for the Tasks API though. Can alias for now?

priyajeet:
Can this PR be broken into multiple small ones

1. API changes
2. UI Changes
3. Other things like changes into package.json/base styles/things not related directly to tasks

flow-typed/box-ui-elements.js
```diff
@@ -618,6 +618,104 @@ type Tasks = {
...
+    type: 'file',
+};
+
+type FolderMini = {
+    id: ID,
+    name: string,
+    type: 'folder',
+};
+
+type TaskStatus = 'NOT_STARTED' | 'IN_PROGRESS' | 'COMPLETED' | 'APPROVED' | 'REJECTED';
```
priyajeet: Normally we try to keep the actual re-usable constants in the constants.js file so that the strings are all in one place. And just use the imported constant here via typeof

alexkrolick:
Will rebase after the sub-PRs

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -62,27 +64,58 @@ const ActiveState = ({
...
-                                onDelete={onTaskDelete}
-                                onEdit={onTaskEdit}
-                                permissions={{
-                                    can_delete: true,
-                                    can_edit: true,
-                                }}
-                                // permissions are not part of task API so hard code to true
-                                translations={translations}
+                    return (
+                        <div>
```
jstoffan: It looks like this is inside a `ul`, so using a div here will generate invalid markup.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -62,27 +64,58 @@ const ActiveState = ({
...
+                            <FeatureFlag
+                                feature="activityFeed.tasks.newCards"
+                                disabled={() =>
+                                    item.task_assignment_collection &&
+                                    item.task_assignment_collection.total_count && (
+                                        <li key={type + id} className="bcs-activity-feed-task">
+                                            <Task
```
jstoffan: The `<Task>` and `<TaskNew>` markup looks very similar between enabled/disabled. Is it possible for us to limit the feature flag to the class name to avoid the duplication?

src/elements/content-sidebar/activity-feed/activity-feed/__tests__/ActiveState-test.js
```diff
@@ -64,7 +64,7 @@ describe('elements/content-sidebar/ActiveState/activity-feed/ActiveState', () =>
     });
 
     test('should correctly render with comments, tasks, versions', () => {
-        const wrapper = shallow(<ActiveState items={items} />).dive();
+        const wrapper = mount(<ActiveState items={items} />);
```
jstoffan: We try try to use `shallow` where possible to keep tests isolated and fast. Is changing all the snapshots necessary or can the feature flip logic be tested on its own?

src/elements/content-sidebar/activity-feed/comment/Comment.scss
```diff
@@ -124,7 +124,6 @@
         }
 
         .bcs-comment-content {
-            color: $fours;
```
jstoffan: Is changing the non-task comment styles intentional?

src/elements/content-sidebar/activity-feed/task-new/PendingAssignment.js
```diff
@@ -0,0 +1,47 @@
...
+
+const PendingAssignment = ({ onTaskApproval, onTaskReject, onTaskComplete, taskType }: Props): React.Node => (
+    <div className="bcs-task-pending-assignment bcs-task-assignment-actions">
+        {taskType === TASK_TYPE_APPROVAL ? (
+            <React.Fragment>
+                <Button
+                    className="bcs-task-action-button bcs-task-check-btn"
+                    onClick={onTaskApproval}
+                    data-resin-target={ACTIVITY_TARGETS.TASK_APPROVE}
+                >
```
jstoffan: Please add `type="button"`, as these are submit buttons by default.

src/elements/content-sidebar/activity-feed/task-new/Task.js
```diff
@@ -0,0 +1,166 @@
...
+    onAssignmentUpdate: Function,
+    onDelete?: Function,
+    onEdit?: Function,
+    translatedTaggedMessage?: string,
+    translations?: Translations,
+};
+
+const MAX_AVATARS = 3;
+
+// eslint-disable-next-line react/prefer-stateless-function
```
jstoffan: Can this be a stateless functional component?

src/elements/content-sidebar/activity-feed/task-new/Task.js
```diff
@@ -0,0 +1,166 @@
...
+            translatedTaggedMessage,
+            translations,
+            getAvatarUrl,
+            getUserProfileUrl,
+            getMentionWithQuery,
+            mentionSelectorContacts,
+        } = this.props;
+
+        // Enable when edit/delete are implemented for new tasks
+        const taskPermissions = {
```
jstoffan: It looks like `permissions` is currently declared in the consumer, but is ignored?

src/elements/content-sidebar/activity-feed/task-new/Task.js
```diff
@@ -0,0 +1,166 @@
...
+            can_edit: false,
+            can_delete: false,
+        };
+
+        // // find assignment for current user if permitted
```
jstoffan: Double comment.

src/elements/content-sidebar/activity-feed/task-new/Task.js
```diff
@@ -0,0 +1,166 @@
...
+
+        const currentUserAssignment =
+            assigned_to && assigned_to.entries
+                ? assigned_to.entries.find(({ target }) => target.id === currentUser.id)
+                : null;
+
+        const assigneeCount = (assigned_to && assigned_to.entries.length) || 0;
```
jstoffan: It looks like we're checking the existence of `assigned_to` and `assigned_to.entries` twice. Can we normalize it into an array once above both checks?

src/elements/content-sidebar/activity-feed/task-new/Task.js
```diff
@@ -0,0 +1,166 @@
...
+        const hiddenAssigneeCount = assigneeCount - MAX_AVATARS;
```
jstoffan: Should we try to avoid negative integers here to reduce confusion? `Math.max(0, assigneeCount - MAX_AVATARS)`?

src/elements/content-sidebar/activity-feed/task-new/Task.js
```diff
@@ -0,0 +1,166 @@
...
+                    tagged_message={name}
+                    translatedTaggedMessage={translatedTaggedMessage}
+                    translations={translations}
+                    getAvatarUrl={getAvatarUrl}
+                    getUserProfileUrl={getUserProfileUrl}
+                    mentionSelectorContacts={mentionSelectorContacts}
+                    getMentionWithQuery={getMentionWithQuery}
+                />
+                <div className="bcs-task-assignment-container">
+                    {due_at ? (
```
jstoffan: Can this be coerced into a proper boolean so we can use `&&`?

src/elements/content-sidebar/activity-feed/task-new/Task.js
```diff
@@ -0,0 +1,166 @@
...
+                                      );
+                                  })
+                            : null}
+                        {hiddenAssigneeCount > 0 ? (
+                            <span className="bcs-task-assignment-avatar bcs-task-assignment-count">
+                                +{hiddenAssigneeCount}
+                            </span>
+                        ) : null}
+                    </div>
+                    {status === TASK_NEW_INCOMPLETE &&
```
jstoffan: Can we pull this set of conditions into a variable to help explain what "mode" we expect to be in?

src/elements/content-sidebar/activity-feed/task-new/Task.scss
```diff
@@ -0,0 +1,136 @@
...
+        right: 0;
+        width: 12px;
+
+        .fill-color {
+            fill: $white;
+        }
+
+        &.completed,
+        &.approved,
+        &.COMPLETED,
```
jstoffan: Can the statuses be normalized in React to avoid needing lowercase and all-caps versions?

src/elements/content-sidebar/activity-feed/task-new/Task.scss
```diff
@@ -0,0 +1,136 @@
...
+        justify-content: center;
+    }
+
+    .bcs-task-assignment-status {
+        display: inline-block;
+        height: 32px;
+        position: relative;
+        width: 32px;
+
+        &:not(:last-child) {
```
jstoffan: Sibling selectors will allow you to avoid `:not`:

```
& + & {
    margin-left: 3px;
}
```

src/elements/content-sidebar/activity-feed/task-new/Task.scss
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        &.rejected,
+        &.REJECTED {
+            background: $error-color;
+        }
+
+        &.pending,
+        &.incomplete,
+        &.not_started,
```
jstoffan: Are these coming directly from the response payload? Can we normalize them to use hyphens or camelCase?

src/elements/content-sidebar/activity-feed/task-new/Task.scss
```diff
@@ -0,0 +1,136 @@
...
+            font-weight: bold;
+            justify-content: center;
+        }
+    }
+
+    .bcs-task-pending-assignment {
+        margin-top: 5px;
+    }
+
+    .bcs-task-action-button:first-of-type {
```
jstoffan: I believe `first-of-type` works on tags rather than class names.

src/elements/content-sidebar/activity-feed/task-new/TaskActions.js
```diff
@@ -0,0 +1,48 @@
...
+import { TASK_TYPE_APPROVAL } from '../../../../constants';
+
+type Props = {
+    onTaskApproval: Function,
+    onTaskComplete: Function,
+    onTaskReject: Function,
+    taskType?: TaskType,
+};
+
+const TaskActions = ({ onTaskApproval, onTaskReject, onTaskComplete, taskType }: Props): React.Node => (
```
jstoffan: This looks like it may be a duplicate of the other TaskActions component.

src/elements/content-sidebar/activity-feed/task-new/TaskStatus.js
```diff
@@ -0,0 +1,53 @@
...
+        case TASK_NEW_REJECTED:
+            return <IconReject width={ICON_SIZE} height={ICON_SIZE} {...rest} />;
+        case TASK_NEW_INCOMPLETE:
+            return <IconIncompleteNoSvg width={ICON_SIZE} height={ICON_SIZE} {...rest} />;
+        default:
+            return null;
+    }
+};
+
+const StatusMessage = ({ status }: { status: TaskStatus }) => {
```
jstoffan: We usually try to stick to one component per file.

src/elements/content-sidebar/activity-feed/task-new/PendingAssignment.js
alexkrolick: The Form component relies on intercepting submit events, so this is right

src/elements/content-sidebar/activity-feed/comment/Comment.scss
alexkrolick: Yes, the designs are being updated to keep them in sync

src/elements/content-sidebar/activity-feed/task-new/Task.js
alexkrolick: Will be used when the functionality to support those permissions is added

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
alexkrolick: I'm pretty sure comments and old tasks use <p> and <div> internally, but I'll take a look to see if this is fixable for all

src/elements/content-sidebar/activity-feed/activity-feed/__tests__/ActiveState-test.js
alexkrolick: It seems like having the feature flag in the component is an implementation detail, when the test only cares about the task component being rendered. I guess we could add a testid to the featureflag wrapping Task/Task identifying it as handling task cards?

src/elements/content-sidebar/activity-feed/task-new/Task.js
alexkrolick: Yes, but I think the fact that && requires explicit coercion makes it less preferable than ternaries in general

src/elements/content-sidebar/activity-feed/task-new/Task.scss
alexkrolick: Yes we can, but is this just for consistency? The JS computation would be more expensive than the additional CSS.

src/elements/content-sidebar/activity-feed/task-new/TaskStatus.js
alexkrolick: https://github.com/airbnb/javascript/pull/2006

src/elements/content-sidebar/activity-feed/activity-feed/__tests__/ActiveState-test.js
jstoffan: I'm open to that approach or others that test the new changes; I'd just like to avoid using `mount` here and regenerating all the snapshots.

src/elements/content-sidebar/activity-feed/task-new/TaskStatus.js
jstoffan: It seems like not exporting the components would make it difficult to test them in isolation. @priyajeet, thoughts?

src/elements/content-sidebar/activity-feed/task-new/Task.scss
jstoffan: Yes, consistency would be the primary goal. Once we start using BEM/SUIT/etc., these will be even more of an issue, as it's not immediately clear why they don't follow project conventions.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
jstoffan: They likely do contain other tags, but they're all inside of a parent `li` in this file. This change introduces a `div` wrapper around the `li` for tasks.

alexkrolick: Yes, but if they use `div` internally this is just as bad as here. I'll look into fixing it for all cases.

jstoffan: I think there may be a misunderstanding. Below is the declaration for the Comment item type in this file, starting at line 49. Note that it has an `<li>` wrapper:

```
case 'comment':
    return (
        <li key={type + id} className="bcs-activity-feed-comment">
            <Comment
                {...item}
                currentUser={currentUser}
                getAvatarUrl={getAvatarUrl}
                getUserProfileUrl={getUserProfileUrl}
                onDelete={onCommentDelete}
                permissions={{
                    can_delete: getProp(permissions, 'can_delete', false),
                    can_edit: getProp(permissions, 'can_edit', false),
                }}
                translations={translations}
            />
        </li>
    );
```

alexkrolick: I see. I thought div was not allowed inside li.

src/elements/content-sidebar/activity-feed/task-new/TaskActions.js
alexkrolick: what other taskactions?

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
alexkrolick: I'm struggling to see a clean way to pull out the task_assignment_collection and the new task layout logic into the same computation using isFeatureEnabled. The assignment collection check has to be contained within the `task` case as well as the feature flag.

src/elements/content-sidebar/activity-feed/task-new/Task.scss
alexkrolick: Doesn't that select all but the _first_ element, not the _last_?

jstoffan: Yes which is why I switched to `margin-left` in the example above.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
jstoffan: The new changes look good to me. We can just nuke the old `<Task>` component once the new stuff is out 100%.

PR #898, opened 2019-02-27
test(examples): read feature config from shared location

alexkrolick:


boxcla:

PR #900, opened 2019-02-27
chore(tasks): add support for new tasks api to feed

alexkrolick:


boxcla:

src/api/Feed.js
```diff
@@ -210,6 +234,23 @@ class Feed extends Base {
...
+     * Fetches the tasks for a file
+     *
+     * @return {Promise} - the feed items
+     */
+    fetchTasksNew(): Promise<?Tasks> {
+        this.tasksNewAPI = new TasksNewAPI(this.options);
+
+        return new Promise(resolve => {
+            this.tasksNewAPI.getTasksForFile({
+                file: { id: this.id },
```
jstoffan: It seems odd to spoof a file object here via an object literal. Can the `getTasksForFile` method accept a file id, instead?

src/api/Feed.js
```diff
@@ -539,6 +668,171 @@ class Feed extends Base {
...
+     *
+     * @param {BoxItem} file - The file to which the task is assigned
+     * @param {string} id - ID of the feed item to update with the new task data
+     * @param {Task} task - API returned task
+     * @param {Array} assignees - List of assignees
+     * @param {Function} successCallback - the function which will be called on success
+     * @param {Function} errorCallback - the function which will be called on error     *
+     * @return {void}
+     */
+    async createTaskNewSuccessCallback(
```
jstoffan: I think we've been avoiding use of async/await in anything other than the `Base` API class. It looks like this file uses promises for existing methods, so it would be good to stay consistent internally.

src/api/__tests__/TaskCollaborators-test.js
```diff
@@ -5,6 +5,10 @@ let taskCollaborators;
 const BASE_URL = 'https://www.foo.com';
 const FILE_ID = 'foo';
 const USER_ID = 'userone';
+const headers = {
```
jstoffan: Can these be imported from the class under test, instead, to avoid duplication in the test?

src/api/TasksNew.js
```diff
@@ -13,6 +13,12 @@ import {
     API_PAGE_LIMIT,
 } from '../constants';
 
+// microservices need different headers than other APIs
+const headers = {
```
jstoffan: Should these be a shared constant or is the duplication intentional?

alexkrolick: This should be a temporary workaround, these APIs should soon accept the standard json headers

alexkrolick: But I am not sure whether they will all be fixed at the same time. If this isn't resolved soon we can move to the constants file.

src/api/Feed.js
alexkrolick: It uses .then(), but nothing is calling multiple dependent promise APIs yet. It creates quite a bit of extra nesting and logic to handle data flow within the function. Is there a functional reason for avoiding async/await here?

jstoffan:
Can you add some tests for the new functionality?

src/api/Feed.js
jstoffan: I see what you mean now. Looks okay as-is, then.

alexkrolick:
@jstoffan added unit tests for the new methods

PR #901, opened 2019-02-27
fix(css): extract modal resets from elements reset

alexkrolick:


boxcla:

src/elements/common/base.scss
```diff
@@ -70,6 +54,32 @@
     }
 }
 
+.be {
+    @include be-content;
```
jstoffan: It looks like most of the rules in `be-content` were declared after the `@import` statements below. I'd suggest splitting `@include reset` out:

```
.be {
    @include reset;
    
    // links have tag styles
    @import '../../styles/common/links';
    // Forms have tag styles
    @import '../../styles/common/forms';

    @include be-content;
    // etc.
}
```

PR #903, opened 2019-02-27
chore: Update scss to mirror all files, add header column

Visylvius:
What this PR does: 
Updates the ListView scss, and gives it a similar feel to the files page. 
Adds the Item Icon to the initial name column 
Before: 
screen shot 2019-02-26 at 2 02 49 pm 3
After: 
screen shot 2019-02-26 at 4 35 10 pm 3

Also: Updates a test to use Test.each instead of Array.forEach to keep tests consistent across the repo. I promised Priyajeet I would do this.


boxcla:

src/features/list-view/ListView.js
```diff
@@ -38,12 +49,17 @@ class ListView extends React.PureComponent<Props> {
             );
         }
         if (columnIndex === 0) {
+            const { cellData: nameCellData, extension } = getGridCell({ columnIndex, rowIndex });
             return (
                 <div className="list-view-name-cell" key={key} style={style}>
-                    {cellData}
+                    <div className="list-view-icon">
+                        <IconItem className="list-view-icon-item" height={25} iconType={extension} width={25} />
```
macrotim: Let's hoist file-icons up and out and keep ListView dumb i.e. it renders what EUA tells it to render. This also have the benefit of simplifying the new complexity added to your tests.

src/features/list-view/ListView.js
```diff
@@ -68,7 +84,7 @@ class ListView extends React.PureComponent<Props> {
                     fixedColumnCount={1}
                     fixedRowCount={1}
                     height={height}
-                    rowHeight={40}
+                    rowHeight={50}
```
macrotim: Will this be pulled out into constants?

src/features/list-view/ListView.js
```diff
@@ -29,6 +31,15 @@ class ListView extends React.PureComponent<Props> {
     cellRenderer = ({ columnIndex, key, rowIndex, style }: CellRendererArgs) => {
         const { getGridCell, getGridHeader } = this.props;
         const cellData = getGridCell({ columnIndex, rowIndex });
+
+        if (rowIndex === 0 && columnIndex === 0) {
```
macrotim: I see that you introduced this special case so that the first column is tagged with a special classname.

Why not use CSS e.g. ".list-view-top-left-grid .list-view-column-header"?

src/features/list-view/__tests__/ListView-test.js
```diff
@@ -71,8 +71,8 @@ describe('features/list-view/ListView', () => {
 
     describe('render()', () => {
         test('Should render the ListView component ', () => {
-            const columnHeaderClass = '.list-view-column-header';
```
macrotim: Shouldn't we still verify this was rendered?

src/features/list-view/__tests__/MetadataListViewMessage-test.js
```diff
@@ -8,28 +8,15 @@ describe('Message', () => {
...
-                message: 'noAccessForTemplate',
-            },
-        ].forEach(message => {
-            test(`${message.description}`, () => {
-                const wrapper = getWrapper(message);
-                expect(wrapper).toMatchSnapshot();
-            });
+        test.each`
+            description                                      | message
+            ${'should correctly render needRefining'}        | ${'needRefining'}
```
macrotim: The messages deserve constants.

src/features/list-view/ListView.js
Visylvius: I did try using CSS at first, using the :nth selector(1) , and :first-child selectors grabbed two elements from the dom rather just the first name column. Introducing the class allowed it to be selected properly

src/features/list-view/ListView.js
```diff
@@ -11,24 +12,26 @@ type Props = {
     getGridCell: ({|
         columnIndex: number,
         rowIndex: number,
+        style?: Object,
```
macrotim: Did we decide to omit?

src/features/list-view/__tests__/ListView-test.js
```diff
@@ -1,9 +1,28 @@
+// @flow
 import * as React from 'react';
 
 import ListView from '../ListView';
-
 // Global Declarations
-const rowData = [['', 'A', 'B', 'C'], ['', 'D', 'E', 'F']];
+
+const nameCellStyle = { height: 50, left: 0, position: 'absolute', top: 0, width: 300 };
```
ChenCodes: Could we reuse the style that's being passed in to the ListView component and import it here?

src/features/list-view/__tests__/ListView-test.js
```diff
@@ -1,9 +1,28 @@
...
+
+type Props = {|
+    className: string,
+    key: string,
+    name: string,
+|};
+
+const ListViewNameTestCell = ({ name }: Props) => <div>{name}</div>;
```
ChenCodes: Nit: Maybe we can do away with the 'Test' in the name. Since this component is being used in a test file I think it's safe to assume that these component are test components.

ListViewNameTestCell -> ListViewNameCell

ChenCodes: Same goes for the `testCell`, we can probably call that `Cell` instead.

src/features/list-view/__tests__/ListView-test.js
Visylvius: Oof, that would be a little tricky.

src/features/list-view/__tests__/__snapshots__/MetadataListViewMessage-test.js.snap
```diff
@@ -97,31 +73,22 @@ exports[`Message render should correctly render noAccessForTemplate 1`] = `
   <div
     className="message-subtitle-text"
   >
-    <FormattedMessage
-      defaultMessage="You do not have access to any files that match this template."
-      id="boxui.metadataView.noResultsForTemplateSubtitleText"
-    />
+    <FormattedMessage />
```
macrotim: Looks like we lost the messages.

src/features/list-view/styles/ListView.scss
```diff
@@ -4,29 +4,53 @@
     display: flex;
     flex-direction: row;
 
-    .list-view-column-header {
-        color: $seventy-sixers;
+    .list-view-top-left-grid .list-view-column-header {
```
macrotim: I think there's redundant properties.

src/features/list-view/styles/ListView.scss
```diff
@@ -4,29 +4,53 @@
...
     .list-view-column-cell {
         border-top: 1px solid $off-white;
         color: $seventy-sixers;
         display: flex;
-        justify-content: center;
-        padding-top: 8px;
+        justify-content: flex-start;
+        padding-top: 16px;
+
+        .list-view-name-cell {
```
macrotim: Let's put both LVNC .js and .scss in same repos. Choose one.

PR #904, opened 2019-02-28
fix(unified-share-modal): fix cases where shared link gets focused

tjgupta:
FYI @booc0mtaco


boxcla:

PR #908, opened 2019-03-01
fix(unified-share-modal): Add correct permission for editable box notes

booc0mtaco:
- make sure that we show an appropriate string for editable box notes, so that we don't say it's
view only inappropriately


boxcla:

src/features/unified-share-modal/utils/item.js
```diff
@@ -0,0 +1,48 @@
+// @flow
```
tjgupta: Should we move this file to the base utils folder? I'd guess there will be other components (particularly in the `features` folder) that could make use of this eventually. If so, should remove reference to the modal in the docs:
```
* @param {LegacyItem} item The item data received by the modal
```

booc0mtaco: hm, I'd hesitate moving it that far up since it's currently stubbed out right now. Perhaps it is a feature utility, tho?

booc0mtaco: After looking, this is actually very usm specific since other feature components spread props on the outermost interface w.r.t. item data, and the parameter type is defined in the modal's flowtype.js. I'd have to move that as well.

PR #910, opened 2019-03-01
feat(query-bar): Implement multi enum in filter button

ChenCodes:
Changes in this PR:
- Update Condition flowtype `value` and `operator` property to be `Array<string>` instead of `string`
- Use `MultiSelectField` in ValueField.js
- Add multi-enum object into columns array in `fixtures.js` for use in the styleguide
- Cleaned up `fixtures.js` by moving copy/pasted column objects to a constant and share that with other constants
- Update Condition.scss to include error state for MultiSelectField
- Handle handler logic when deleting a date from the DatePicker, ran into a bug earlier where there was an error that said `cannot toString null`
screen shot 2019-03-01 at 2 25 21 pm

### DIFF
--- a/src/features/query-bar/__tests__/FilterButton-test.js
+++ b/src/features/query-bar/__tests__/FilterButton-test.js
@@ -1,7 +1,7 @@
 // @flow
 import * as React from 'react';
 
-import { columns } from '../components/fixtures';
+import { columns, initialCondition } from '../components/fixtures';
 import FilterButton from '../components/filter/FilterButton';
 import { EQUALS, LESS_THAN, OPERATOR, VALUES } from '../constants';
 
@@ -81,7 +81,7 @@ describe('feature/query-bar/components/filter/FilterButton', () => {
                         values: [],
                     },
                 ],
-                value: LESS_THAN,
+                values: LESS_THAN,
                 property: OPERATOR,
                 newCondition: [
                     {
@@ -109,7 +109,7 @@ describe('feature/query-bar/components/filter/FilterButton', () => {
                         values: [],
                     },
                 ],
-                value: ['0'],
+                values: ['0'],
                 property: VALUES,
                 newCondition: [
                     {
@@ -120,13 +120,13 @@ describe('feature/query-bar/components/filter/FilterButton', () => {
                     },
                 ],
             },
-        ].forEach(({ description, condition, conditions, value, property, newCondition }) => {
+        ].forEach(({ description, condition, conditions, values, property, newCondition }) => {
             test(`${description}`, () => {
                 const wrapper = getWrapper();
                 wrapper.setState({
                     conditions,
                 });
-                wrapper.instance().handleFieldChange(condition, value, property);
+                wrapper.instance().handleFieldChange(condition, values, property);
 
                 expect(wrapper.state('conditions')).toEqual(newCondition);
             });
@@ -257,18 +257,13 @@ describe('feature/query-bar/components/filter/FilterButton', () => {
     });
 
     describe('closeOnClickPredicate()', () => {
-        const condition = {
-            columnId: '1',
-            id: '3',
-            operator: EQUALS,
-        };
         const conditionsWithValues = [
             {
-                ...condition,
+                ...initialCondition,
                 values: ['1'],
             },
         ];
-        const conditionsWithEmptyValues = [{ ...condition, values: [] }];
+        const conditionsWithEmptyValues = [{ ...initialCondition, values: [] }];
 
         test.each`
             description                                                                | conditions                   | shouldCloseResult
--- a/src/features/query-bar/__tests__/FilterButton-test.js
+++ b/src/features/query-bar/__tests__/FilterButton-test.js
@@ -3,7 +3,7 @@ import * as React from 'react';
 
 import { columns, initialCondition } from '../components/fixtures';
 import FilterButton from '../components/filter/FilterButton';
-import { EQUALS, LESS_THAN, OPERATOR, VALUES } from '../constants';
+import { EQUALS, LESS_THAN } from '../constants';
 
 describe('feature/query-bar/components/filter/FilterButton', () => {
     const getWrapper = (props = {}) => {
@@ -82,7 +82,6 @@ describe('feature/query-bar/components/filter/FilterButton', () => {
                     },
                 ],
                 value: LESS_THAN,
-                property: OPERATOR,
                 expectedConditions: [
                     {
                         columnId: '1',
@@ -92,13 +91,13 @@ describe('feature/query-bar/components/filter/FilterButton', () => {
                     },
                 ],
             },
-        ].forEach(({ description, condition, conditions, value, property, expectedConditions }) => {
+        ].forEach(({ description, condition, conditions, value, expectedConditions }) => {
             test(`${description}`, () => {
                 const wrapper = getWrapper();
                 wrapper.setState({
                     conditions,
                 });
-                wrapper.instance().handleOperatorChange(condition, value, property);
+                wrapper.instance().handleOperatorChange(condition, value);
 
                 expect(wrapper.state('conditions')).toEqual(expectedConditions);
             });
@@ -125,7 +124,6 @@ describe('feature/query-bar/components/filter/FilterButton', () => {
                     },
                 ],
                 values: ['0'],
-                property: VALUES,
                 expectedConditions: [
                     {
                         columnId: '1',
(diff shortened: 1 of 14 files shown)


boxcla:

src/features/query-bar/__tests__/FilterButton-test.js
```diff
@@ -263,16 +257,50 @@ describe('feature/query-bar/components/filter/FilterButton', () => {
     });
 
     describe('closeOnClickPredicate()', () => {
-        test('Should return true if Apply button was clicked', () => {
+        test('Should return true if Apply button was clicked and value is not empty', () => {
```
macrotim: The next two tests deserve to be merged with test.each.

src/features/query-bar/components/filter/Condition.js
```diff
@@ -52,10 +52,10 @@ const Condition = ({
 
     const handleOperatorChange = (option: OptionType) => {
         const { value } = option;
-        onFieldChange(condition, value, OPERATOR);
+        onFieldChange(condition, ([value]: Array<string>), OPERATOR);
```
macrotim: Ah good fix.

src/features/query-bar/components/filter/Condition.js
```diff
@@ -183,7 +183,7 @@ const Condition = ({
                         isDisabled={false}
                         onChange={handleOperatorChange}
                         options={operatorOptions}
-                        selectedValue={operator}
+                        selectedValue={operator[0]}
```
macrotim: Why this change?

src/features/query-bar/components/filter/FilterButton.js
```diff
@@ -69,13 +69,13 @@ class FilterButton extends React.Component<Props, State> {
         const { columns } = this.props;
         if (columns) {
             const firstColumn = columns[0];
-            const operator = COLUMN_OPERATORS[firstColumn.type][0].key;
+            const operator = [COLUMN_OPERATORS[firstColumn.type][0].key];
```
macrotim: Is it possible for Condition to support multiple operators? If not, let's go this route.

src/features/query-bar/components/filter/FilterButton.js
```diff
@@ -69,13 +69,13 @@ class FilterButton extends React.Component<Props, State> {
...
 
             return {
                 columnId: firstColumn.id,
                 id: conditionID,
                 operator,
-                value: null,
+                value: [],
```
macrotim: If it's plural, then let's name it that way i.e. `values.`

src/features/query-bar/components/filter/ValueField.js
```diff
@@ -4,20 +4,23 @@ import { FormattedMessage } from 'react-intl';
...
 import { VALUE } from '../../constants';
 import messages from '../../messages';
 
 import '../../styles/Condition.scss';
 
 type Props = {
-    onChange: (value: String) => void,
-    selectedValue?: string | number,
+    onChange: (value: Array<string>) => void,
+    selectedValue: Array<string>,
```
macrotim: Let's pluralize.

src/features/query-bar/components/filter/ValueField.js
```diff
@@ -4,20 +4,23 @@ import { FormattedMessage } from 'react-intl';
...
     valueOptions: Array<Object>,
     valueType: string,
 };
 
 const ValueField = ({ onChange, selectedValue, valueOptions, valueType }: Props) => {
+    const isValueSet = selectedValue.length > 0;
```
macrotim: Pluralize selectedValue.

src/features/query-bar/components/filter/ValueField.js
```diff
@@ -70,20 +73,32 @@ const ValueField = ({ onChange, selectedValue, valueOptions, valueType }: Props)
                         hideLabel
                         label="Date"
                         name="datepicker"
-                        onChange={e => onChange(e.toString())}
+                        onChange={e => {
+                            return e ? onChange([e.toString()]) : onChange([]);
```
macrotim: Could it fire with an empty event? I'm wonder if this ternary op is needed?

src/features/query-bar/components/filter/ValueField.js
```diff
@@ -39,9 +42,9 @@ const ValueField = ({ onChange, selectedValue, valueOptions, valueType }: Props)
                         hideLabel
                         label="Number input"
                         name="number field"
-                        onChange={e => onChange(e.target.value)}
+                        onChange={e => onChange([e.target.value])}
                         placeholder="Enter a number"
-                        value={selectedValue || ''}
+                        value={isValueSet ? selectedValue[0] : null}
```
macrotim: This ternary is duplicated. Can it be moved up?

src/features/query-bar/components/filter/ValueField.js
ChenCodes: It still fires with a null event.

src/features/query-bar/components/filter/ValueField.js
```diff
@@ -20,6 +20,7 @@ type Props = {
```
ChenCodes: I committed this change with the wrong message, it should be 'refactor value in ValueField'

src/features/query-bar/flowTypes.js
```diff
@@ -24,6 +24,6 @@ export type ColumnType = {
 export type ConditionType = {
     columnId: string,
     id: number,
-    operator: Array<string>,
-    value: Array<string>,
+    operator: string,
```
macrotim: Too bad `operator` isn't typed. Something to consider.

src/features/query-bar/__tests__/FilterButton-test.js
```diff
@@ -99,34 +99,34 @@ describe('feature/query-bar/components/filter/FilterButton', () => {
...
                         columnId: '1',
                         id: '5',
                         operator: EQUALS,
-                        value: null,
+                        values: [],
                     },
                 ],
-                property: 0,
-                conditionProperty: 'value',
+                value: ['0'],
```
macrotim: Plural?

src/features/query-bar/__tests__/FilterButton-test.js
```diff
@@ -257,38 +257,24 @@ describe('feature/query-bar/components/filter/FilterButton', () => {
...
-            wrapper.instance().setState({
-                conditions,
-            });
-            targetWithClassName.target.className = 'apply-filters-button';
-            const closeOnClickPredicateResult = wrapper.instance().shouldClose(targetWithClassName);
-
-            wrapper.update();
-            expect(closeOnClickPredicateResult).toEqual(true);
-        });
+        const condition = {
```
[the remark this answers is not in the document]
ChenCodes: Yes

src/features/query-bar/components/filter/Condition.js
```diff
@@ -23,7 +23,7 @@ type Props = {
     index: number,
     onColumnChange: (condition: ConditionType, columnId: string) => void,
     onConnectorChange: (option: OptionType) => void,
-    onFieldChange: (condition: ConditionType, value: Array<string>, property: string) => void,
+    onFieldChange: (condition: ConditionType, value: string | Array<string>, property: string) => void,
```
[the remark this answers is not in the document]
ChenCodes: I could break it up into two methods but then we would have two methods that both do about 95% of the same thing:

```
handleOperatorChange = (condition: ConditionType, value: string) => {
        const { conditions } = this.state;
        let newConditionIndex = 0;
        const conditionToUpdate = conditions.find((currentCondition, index) => {
            newConditionIndex = index;
            return currentCondition.id === condition.id;
        });

        const newCondition = {
            ...conditionToUpdate,
            [OPERATOR]: value,
        };

        const newConditions = conditions.slice(0);
        newConditions[newConditionIndex] = newCondition;

        this.setState({
            conditions: newConditions,
        });
    };
```

and 

```
handleValueChange = (condition: ConditionType, values: Array<string>) => {
        const { conditions } = this.state;
        let newConditionIndex = 0;
        const conditionToUpdate = conditions.find((currentCondition, index) => {
            newConditionIndex = index;
            return currentCondition.id === condition.id;
        });

        const newCondition = {
            ...conditionToUpdate,
            [VALUES]: values,
        };

        const newConditions = conditions.slice(0);
        newConditions[newConditionIndex] = newCondition;

        this.setState({
            conditions: newConditions,
        });
    };
```

ChenCodes:
I think Operator gets the point across. Is there a meaning that OperatorKey helps convey that would give the user more context?

PR #911, opened 2019-03-01
fix(pill-selector-dropdown): update suggested pills filter

nickgriffin1:
Problem: this was build under the assumption that all pills contained an "id" value, which is not true. All pill do have a "value". Value is typically either an "id" or an "email" so we can filter the list of suggested pills by both.

### DIFF
--- a/src/components/pill-selector-dropdown/SuggestedPillsRow.js
+++ b/src/components/pill-selector-dropdown/SuggestedPillsRow.js
@@ -9,19 +9,21 @@ import './SuggestedPillsRow.scss';
 
 type Props = {
     onSuggestedPillAdd?: SuggestedPillType => void,
-    selectedPillsIDs?: Array<number>,
+    selectedPillsValues?: Array<number>,
     suggestedPillsData?: SuggestedPills,
     title?: string,
 };
 
 const SuggestedPillsRow = ({
     onSuggestedPillAdd = noop,
-    selectedPillsIDs = [],
+    selectedPillsValues = [],
     suggestedPillsData = [],
     title,
 }: Props) => {
-    // Prevents pills from being rendered that are in the form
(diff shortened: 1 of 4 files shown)


boxcla:

src/components/pill-selector-dropdown/SuggestedPillsRow.js
```diff
@@ -9,19 +9,21 @@ import './SuggestedPillsRow.scss';
...
-    selectedPillsIDs = [],
+    selectedPillsValues = [],
     suggestedPillsData = [],
     title,
 }: Props) => {
-    // Prevents pills from being rendered that are in the form
-    const filteredSuggestedPillData = suggestedPillsData.filter(item => !selectedPillsIDs.includes(item.id));
+    // Prevents pills from being rendered that are in the form by checking for value which is either an ID or email
+    const filteredSuggestedPillData = suggestedPillsData.filter(
+        item => !(selectedPillsValues.includes(item.email) || selectedPillsValues.includes(item.id)),
```
priyajeet: Just double check to make sure Array `includes` is polyfilled

priyajeet: Can "email" and "id" be masked somehow? As in, you provide what to filter against.

nickgriffin1: Tested as a linked dependency on IE11/Win10 and no console error thrown. I feel like we should normally be able to assume that is working right?

In this case keyProp would be a prop passed from PillSelectorDropdown? That seems like reasonable solution that is a bit more reusable to cover other use cases. I'll change this.

nickgriffin1: I'll merge this for now and Think about how best to implement this prop in a future commit.

PR #913, opened 2019-03-04
feat(pill-selector-dropdown): add prop for filtering suggested pills

nickgriffin1:
Adds a new prop for filtering suggested data by a particular key. From the suggested mentioned in this PR by @priyajeet https://github.com/box/box-ui-elements/pull/911


boxcla:

PR #914, opened 2019-03-04
fix(flow): publish src folder for flow type references to src

priyajeet:


boxcla:

PR #915, opened 2019-03-04
chore(lint): apply precommit formatters to all files

alexkrolick:
Followup to the commit that added precommit formatting for scss, html, md. Should fix all the files to avoid future large diffs.


boxcla:

PR #916, opened 2019-03-05
build(examples): remove cached files that are git ignored

priyajeet:
So that gh-pages branch isn't getting prior committed files


boxcla:

PR #920, opened 2019-03-05
chore(usm): Update copy for suggested collabs

nickgriffin1:
"Suggested:" => "Add:"

Also updated description and id to make sure it gets picked up by Moji.


boxcla:

PR #923, opened 2019-03-06
feat(typography): heading component

priyajeet:
1. Adds h1, h2, h3, h4, h5, h6 support
2. remove tag styles and reset related to header elements

screen shot 2019-03-05 at 6 46 42 pm

- Alternatively we can create a `<Typography>` element that takes in more than just hX. Similar to material-ui.
- Alternatively we can create components for each.
- Alternatively we don't style any of these and let the parent components deal with it.

BREAKING CHANGE: The styling classes for h1-h6 have changed.

### DIFF
--- a/src/components/heading/Heading.js
+++ b/src/components/heading/Heading.js
@@ -1,0 +1,39 @@
+// @flow
+import * as React from 'react';
+import isFinite from 'lodash/isFinite';
+import classNames from 'classnames';
+import HeadingConsumer from './HeadingConsumer';
+import './Heading.scss';
+
+type Props = {
+    /** Heading contents */
+    children: React.Node,
+    /** Heading contents */
+    className?: string,
+    /** Header tag level */
+    level?: number,
+};
+
+const Heading = ({ children, className = '', level, ...rest }: Props) => {
+    return (
+        <HeadingConsumer>
+            {value => {
+                const headingLevel = parseInt(level || value, 10);
+                if (isFinite(headingLevel) && (headingLevel < 1 || headingLevel > 6)) {
+                    throw new Error('Heading level can only be 1, 2, 3, 4, 5 or 6');
+                }
+                const type = `h${headingLevel}`;
+                return React.createElement(
+                    type,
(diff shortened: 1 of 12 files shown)


boxcla:

src/components/heading/Heading.scss
```diff
@@ -0,0 +1,44 @@
+@import '../../styles/variables';
+
+.bdl-h1,
```
priyajeet: should ideally be `bdl-Heading--h1`

alexkrolick:
@phora this is pretty cool 🎢 

How about adding a variant that uses Context to set the heading level, like this: https://twitter.com/sebmck/status/1092159277127557121

src/components/heading/Heading.js
```diff
@@ -0,0 +1,39 @@
...
+const Heading = ({ children, className = '', level, ...rest }: Props) => {
+    return (
+        <HeadingConsumer>
+            {value => {
+                const headingLevel = parseInt(level || value, 10);
+                if (isFinite(headingLevel) && (headingLevel < 1 || headingLevel > 6)) {
+                    throw new Error('Heading level can only be 1, 2, 3, 4, 5 or 6');
+                }
+                const type = `h${headingLevel}`;
+                return React.createElement(
```
priyajeet: extract out to its own component for simpler testing per file

src/components/heading/Heading.scss
```diff
@@ -0,0 +1,44 @@
+@import '../../styles/variables';
+
+.bdl-Heading--h1,
```
priyajeet: This should just be .bdl-Heading since modifiers are for the rules below

src/components/heading/Heading.js
```diff
@@ -0,0 +1,39 @@
...
+                    type,
+                    {
+                        className: classNames(`bdl-Heading--${type}`, className),
```
priyajeet: css class should be `bdl-Heading bdl-Heading--TYPE`

PR #924, opened 2019-03-06
feat(usm): increase size of invite input form

nickgriffin1:
Increase the size of the form to both better support suggested pills and make the form a stronger call to action.

From...
image

To...
image

Fully expanded...
image

### DIFF
--- a/src/features/unified-share-modal/UnifiedShareModal.scss
+++ b/src/features/unified-share-modal/UnifiedShareModal.scss
@@ -121,6 +121,7 @@
     .pill-selector-input-wrapper {
         border: 1px solid $seesee;
         margin-top: 10px;
+        min-height: 75px;
 
         .pill-selector-input::placeholder {
             color: $nines;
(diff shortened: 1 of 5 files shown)


boxcla:

booc0mtaco:
`usm` => `unified-share-modal`

src/features/unified-share-modal/UnifiedShareModal.scss
```diff
@@ -121,6 +121,7 @@
     .pill-selector-input-wrapper {
         border: 1px solid $seesee;
         margin-top: 10px;
+        min-height: 75px;
```
booc0mtaco: This should be the same hight for the input fields used in the forms for send invite collaborators and email shared link forms in the USM. ideally, they'd all get this new `min-height` from a shared variable, instead of it being specified several times.

nickgriffin1: Will update with a shared variable.

src/styles/common/_forms.scss
```diff
@@ -4,6 +4,10 @@
  * Forms
  **************************************/
 
+/* Variables */
+
+$large-pill-form-min-height: $default-line-height * 4;
```
nickgriffin1: Wasn't sure on if to put here of in `variables.scss`. I decided to put it here because it's a bit more organized and having all shared variables in one file might not scale well.

PR #927, opened 2019-03-07
refactor(text-input): remove tag styles for inputs

priyajeet:
BREAKING CHANGE: The styling classes for text inputs have changed.


boxcla:

src/components/text-input/TextInput.js
```diff
@@ -46,29 +45,28 @@ const TextInput = ({
     labelTooltip,
     ...rest
 }: Props) => {
-    const classes = classNames(className, 'text-input-container', {
-        'show-error': !!error,
+    const classes = classNames(className, 'bdl-TextInput-container', {
```
jstoffan: We should ideally try to avoid qualifiers like `-container` at the root of components, since it indicates a descendent element. This should ideally just be `bdl-TextInput`.

yarn.lock
```diff
@@ -46,6 +46,26 @@
     semver "^5.4.1"
     source-map "^0.5.0"
 
+"@babel/core@>=7.1.0":
```
jstoffan: Possible to avoid having two versions of `@babel/core` in the lockfile?

src/components/text-input/TextInput.scss
```diff
@@ -1,55 +1,107 @@
...
+        appearance: textfield;
+
+        &::-webkit-search-cancel-button,
+        &::-webkit-search-results-button {
+            display: none;
+        }
+    }
+
+    &[type="number"] {
+        appearance: textfield;
```
jstoffan: Looks like this style rule can be combined with `&[type="search"]`.

src/components/text-input/TextInput.scss
```diff
@@ -1,55 +1,107 @@
...
+
+.bdl-TextInput-container {
+    @include common-typography;
 
-.text-input-container {
     margin: 0 0 20px;
     position: relative;
 
-    .text-input-loading {
+    &.bdl-has-error {
```
jstoffan: Should this be `.bdl-hasError`?

priyajeet: Yes!

priyajeet: Actually is/has is followed by a - and thereafter its camelCase.

With namespace:
.bdl-is-loading, .bdl-has-error, .bdl-is-errorShown, .bdl-has-criticalError.

Without namespace:
.is-loading, .has-error, .is-errorShown, .has-criticalError.

We can however maybe enforce just `is` and drop `has` so in the above examples

.has-criticalError --> .is-criticalErrorShown

PR #928, opened 2019-03-07
feat(versions): Add version history panel components and routes

jstoffan:
Please note that this change is intended to set up the infrastructure for viewing the version history items; the content of the panel is placeholder for now.

versions panel


boxcla:

conf/styleguide.setup.js
```diff
@@ -1,10 +1,10 @@
 import { IntlProvider } from 'react-intl';
 import features from '../examples/src/features';
 // Constants
-global.FEATURES = features;
-global.FILE_ID = __FILEID__ || '308566420378'; // eslint-disable-line
-global.FOLDER_ID = __FOLDERID__ || '51964781421'; // eslint-disable-line
-global.TOKEN = __TOKEN__ || 'aiMADZorjZDCJEfi7zREbvHBo2K70MXf'; // eslint-disable-line
+global.FEATURES = global.FEATURES || features;
```
jstoffan: This change allows us to inject these variables into the host page via Cypress.

src/elements/content-sidebar/versions/VersionsSidebar.js
```diff
@@ -0,0 +1,33 @@
...
+    <SidebarContent
+        className="bcs-versions"
+        title={
+            <React.Fragment>
+                <BackButton />
+                <FormattedMessage {...messages.sidebarVersionsTitle} />
+            </React.Fragment>
+        }
+    >
+        {error || versions.map(version => version.id)}
```
jstoffan: This is just a placeholder for the forthcoming version history item components.

src/elements/content-sidebar/versions/__tests__/VersionsSidebarContainer-test.js
```diff
@@ -0,0 +1,34 @@
+import * as React from 'react';
+import { shallow } from 'enzyme/build';
+import VersionsSidebar from '..';
+
+jest.mock('../../../common/api-context', () => ({
```
jstoffan: I think this is preferable to exporting/importing unwrapped versions of components. It allows us to avoid modifying source code just to support unit tests. I'm am open to discussion about it, though.

src/elements/common/back-button/BackButton.js
```diff
@@ -0,0 +1,29 @@
...
+ * @flow
+ * @file Back Button component
+ * @author Box
+ */
+
+import * as React from 'react';
+import { FormattedMessage } from 'react-intl';
+import { withRouter } from 'react-router-dom';
+import type { RouterHistory } from 'react-router-dom';
+import IconNavigateLeft from '../../../icons/general/IconNavigateLeft';
```
jeremypress: Thoughts on using https://opensource.box.com/box-ui-elements/#/Icons?id=iconpageback?

src/elements/common/back-button/BackButton.js
```diff
@@ -0,0 +1,29 @@
...
+import messages from '../messages';
+import PlainButton from '../../../components/plain-button';
+import './BackButton.scss';
+
+type Props = {
+    history: RouterHistory,
+};
+
+const BackButton = ({ history }: Props) => (
```
jeremypress: Since EUA uses react-router this could be useful for other consumers? Maybe not since it uses our `withRouter`.

src/elements/common/back-button/BackButton.scss
```diff
@@ -0,0 +1,8 @@
+.be-btn-back {
+    &,
+    &:hover,
```
jeremypress: why do we need to specifically target hover and focus? Do we always want 5px of margin when the back button is rendered?

src/elements/content-sidebar/SidebarPanels.js
```diff
@@ -96,31 +103,45 @@ const SidebarPanels = ({
...
+                                <LoadableActivitySidebar
+                                    currentUser={currentUser}
+                                    file={file}
+                                    onVersionHistoryClick={onVersionHistoryClick}
+                                    startMarkName={MARK_NAME_JS_LOADING_ACTIVITY}
+                                    {...activitySidebarProps}
+                                />
+                            </Route>
+                            <Route exact path={`${match.path}/versions`}>
+                                <LoadableVersionsSidebar fileId={fileId} key={fileId} />
```
jeremypress: should we pass in version sidebar props even if we don't use any real data yet?

src/elements/content-sidebar/versions/VersionsSidebar.scss
```diff
@@ -0,0 +1,8 @@
+.bcs-versions {
+    .bcs-scroll-content {
+        padding-left: 25px;
```
jeremypress: `padding: 10px 25px 0 25px;` ?

src/elements/common/back-button/BackButton.js
jstoffan: It looks like that icon is intended to be used with pagination controls. The other back buttons in EUA use the IconNavigateLeft icon.

src/elements/common/back-button/BackButton.js
jstoffan: Possibly. If that demand materializes, we can look at promoting this outside of elements to a more common location. The `withRouter` HOC should work as-is in EUA without issue.

src/elements/common/back-button/BackButton.scss
jstoffan: The default `.btn-plain` class sets margin and padding on hover and focus for some reason, so I'm forced to here as well.

src/elements/content-sidebar/SidebarPanels.js
jstoffan: I don't understand the question. The version sidebar doesn't really support much in the way of props at the moment.

src/elements/content-sidebar/versions/VersionsSidebar.scss
jstoffan: We should avoid setting style properties that we don't actually need (padding-bottom, in this case). It makes style inheritance and overrides more difficult when we constantly re-set values.

src/elements/content-sidebar/Sidebar.js
```diff
@@ -155,7 +177,7 @@ class Sidebar extends React.Component<Props, State> {
                             isOpen={isOpen}
                             key={file.id}
                             metadataSidebarProps={metadataSidebarProps}
-                            onVersionHistoryClick={onVersionHistoryClick}
+                            onVersionHistoryClick={onVersionHistoryClick || handleVersionHistoryClick}
```
ConradJChan: Is this logic to defer to the user of the ContentSidebar element in case they have provided `onVersionHistoryClick` otherwise to fall back to this new feature if enabled?

src/elements/content-sidebar/SidebarPanels.js
ConradJChan: What benefit does providing the `key` here have?

src/elements/content-sidebar/Sidebar.js
jstoffan: Yep!

src/elements/content-sidebar/SidebarPanels.js
jstoffan: It forces a remount of the component so we can avoid tracking whether the fileId changed.

jeremypress: we're doing it for the other sidebars `{...activitySidebarProps}` but it makes sense to wait until we actually do something with said props.

PR #930, opened 2019-03-07
fix(unified-share-modal): increase pill selector height

nickgriffin1:
* Creates a global variable for pill selector height


boxcla:

src/styles/variables.js
```diff
@@ -75,4 +75,5 @@ export const menuZIndex = '170'; // menu-z-index
 export const notificationsWrapperZIndex = '180'; // notifications-wrapper-z-index
 export const overlayZIndex = '190'; // overlay-z-index
 export const tooltipZIndex = '190'; // tooltip-z-index
+export const defaultLineHeight = '20px'; // default-line-height
```
booc0mtaco: hmm, I would have expected the name to be `bdlLineHeight`. Did you add these two (the JS variable, and the corresponding JSON) manually?

nickgriffin1: This was automatically generated

nickgriffin1: On that point thought, I'm open to calling it either "default" or "global".

booc0mtaco: The name here and in the JSON should match a variable in _variables.scss, but I don't see one in here with the `default-line-height` name 🤔 (this happens in build-style-vars.js)

Can you re-run `build:variables` to double-check?

nickgriffin1: Sure I ran it again and got a diff on pretty much the whole file. I'll push that up so you can check it out.

nickgriffin1: Oddly enough, git locally gives me a much different diff than when I push to GHE. Might just be git being weird. Seems like it's the same as before.

PR #931, opened 2019-03-07
 feat(content-sidebar): Loading state for additional sidebar tabs

jeremypress:
todo:
- [x] flow 
- [x] rename css based on new SUIT standards
- [x] unit tests


boxcla:

src/elements/common/_animations.scss
```diff
@@ -0,0 +1,27 @@
...
+    100% {
+        opacity: 1;
+    }
+}
+
+@mixin fade-in($fade-in-time) {
+    animation: fade-in ease-in $fade-in-time;
+}
+
+@mixin loading-pulse($start-color, $end-color, $fade-in-time) {
```
jeremypress: @ConradJChan we should be able to reuse this for the Content Picker loading state.

src/elements/content-sidebar/additional-tabs/AdditionalTab.js
```diff
@@ -0,0 +1,40 @@
...
+    title,
+    ...rest
+}: AdditionalSidebarTab) => {
+    const className = classNames('bcs-nav-btn', isLoading ? 'is-hidden' : '');
+
+    return (
+        <Tooltip position="middle-left" text={title}>
+            <PlainButton className={className} type="button" onClick={() => callbackFn({ id, callbackData: rest })}>
+                {id > 0 && iconUrl ? (
+                    <img className="bcs-additional-tab-icon" src={iconUrl} onLoad={onImageLoad} alt={title} />
```
jeremypress: Will handle the onError case in a future PR.

src/elements/common/_animations.scss
```diff
@@ -0,0 +1,27 @@
+@keyframes fade-in {
```
jstoffan: This seems likely to conflict with other mixins. Should it be namespaced with `bdl`?

src/elements/content-sidebar/__tests__/SidebarNav-test.js
```diff
@@ -5,7 +5,7 @@ import IconMagicWand from '../../../icons/general/IconMagicWand';
 import IconMetadataThick from '../../../icons/general/IconMetadataThick';
 import IconDocInfo from '../../../icons/general/IconDocInfo';
 import IconChatRound from '../../../icons/general/IconChatRound';
-import AdditionalTabs from '../AdditionalTabs';
+import AdditionalTabs from '../additional-tabs/AdditionalTabs';
```
jstoffan: I think we've started to prefer directory style imports via index.js, so `import { AdditionalTabs } from '../additional-tabs';`

src/elements/common/_animations.scss
booc0mtaco: Could we consider moving this up to `styles/common` so shared animations can be use both in existing elements and the other features/components?

src/elements/content-sidebar/__tests__/SidebarNav-test.js
```diff
@@ -81,18 +81,10 @@ describe('elements/content-sidebar/SidebarNav', () => {
...
-                    title: 'Test title',
-                    iconUrl: 'https://foo.com/icon',
-                    callback: jest.fn(),
-                    status: 'ADDED',
-                },
-            ],
+            additionalTabs: [],
         };
         const wrapper = getWrapper(props);
         expect(wrapper.find(AdditionalTabs)).toHaveLength(1);
```
jstoffan: Does this mean that the wrapper should exist, but have no children and be in a loading state? It might be better to test those things explicitly rather than using a mounted snapshot.

src/elements/content-sidebar/additional-tabs/AdditionalTabs.js
```diff
@@ -0,0 +1,72 @@
...
+    loadedTabs: boolean;
+
+    initialState: State = {
+        isLoading: true,
+    };
+
+    constructor(props) {
+        super(props);
+
+        this.numLoadedTabs = 0;
```
JustinHoldstock: Can this be taken out of the constructor, as a class field, instead?

src/elements/content-sidebar/additional-tabs/AdditionalTabs.js
```diff
@@ -0,0 +1,72 @@
...
+type Props = {
+    tabs: ?AdditionalSidebarTabs,
+};
+
+type State = {
+    isLoading: boolean,
+};
+
+class AdditionalTabs extends PureComponent<Props, State> {
+    loadedTabs: boolean;
```
JustinHoldstock: Consider following the `is` or `has` prefix for boolean types. As this stands, it currently sounds like an array of tabs.

`hasLoadedTabs` or something

src/elements/content-sidebar/additional-tabs/AdditionalTab.js
```diff
@@ -0,0 +1,40 @@
...
+const AdditionalTab = ({
+    callback: callbackFn = noop,
+    iconUrl,
+    id,
+    isLoading,
+    onImageLoad,
+    title,
+    ...rest
+}: AdditionalSidebarTab) => {
+    const className = classNames('bcs-nav-btn', isLoading ? 'is-hidden' : '');
```
jstoffan: `const className = classNames('bcs-nav-btn', { 'is-hidden': isLoading })`;

src/elements/content-sidebar/additional-tabs/AdditionalTabs.js
jeremypress: forgot to remove, that's not being used.

src/elements/content-sidebar/additional-tabs/AdditionalTabLoading.js
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react';
+import PlainButton from '../../../components/plain-button/PlainButton';
+
+const AdditionalTabLoading = () => {
+    return (
+        <PlainButton className="bcs-additional-tab-loading-btn" type="button" isDisabled>
```
jstoffan: Should this be a `div` if it's always disabled?

src/elements/content-sidebar/additional-tabs/AdditionalTabs.js
```diff
@@ -0,0 +1,70 @@
...
+    };
+
+    render() {
+        const { tabs } = this.props;
+        const { isLoading } = this.state;
+
+        return (
+            <div className="bcs-nav-additional-tabs">
+                {isLoading && <AdditionalTabsLoading />}
+                {tabs &&
```
jstoffan: Should these still be rendered if `isLoading` is true?

src/elements/content-sidebar/additional-tabs/AdditionalTabsLoading.js
```diff
@@ -0,0 +1,25 @@
+/**
+ * @flow
+ * @file Preview sidebar additional tabs components
+ * @author Box
+ */
+
+import * as React from 'react';
+import './AdditionalTabs.scss';
```
jstoffan: Can we group the SCSS imports together?

src/elements/content-sidebar/additional-tabs/AdditionalTabsLoading.scss
```diff
@@ -0,0 +1,28 @@
...
+    background-color: transparent;
+    display: flex;
+    height: 60px;
+    justify-content: center;
+    pointer-events: none;
+    position: relative;
+    width: 59px; // 1px for left border
+}
+
+$fade-in-time: 0s;
```
jstoffan: Will this conflict with other top-level global variables?

src/elements/content-sidebar/additional-tabs/MoreTabLoading.js
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react';
+import './AdditionalTabs.scss';
+import PlainButton from '../../../components/plain-button/PlainButton';
+import IconEllipsis from '../../../icons/general/IconEllipsis';
+
+const MoreTabLoading = () => {
+    return (
+        <PlainButton className="bcs-additional-tab-loading-btn" type="button" isDisabled>
```
jstoffan: Can we avoid duplication of the outer wrapper?

src/elements/content-sidebar/additional-tabs/AdditionalTabs.js
```diff
@@ -0,0 +1,70 @@
...
+        this.state = { ...this.initialState };
+    }
+
+    onImageLoad = () => {
+        const { tabs } = this.props;
+
+        this.numLoadedTabs += 1;
+
+        // If we're displaying the more tab, consider it loaded.
+        const hasMoreTab = tabs.find(tab => tab.id < 0 && !tab.iconUrl);
```
ConradJChan: does this need to be done on each `onImageLoad` invocation or can it be done once?

src/elements/content-sidebar/additional-tabs/AdditionalTabsLoading.scss
```diff
@@ -0,0 +1,28 @@
...
+
+.bcs-additional-tab-loading-content {
+    @include loading-pulse(#f4f4f4, #eee, $fade-in-time);
```
ConradJChan: Do we have variables for these colors?

src/elements/content-sidebar/additional-tabs/MoreTabLoading.js
jeremypress: the outer wrapper being the plain button? If I switch the loading component to use a div since it's always disabled, I won't be able to abstract with the actual button.

src/elements/content-sidebar/additional-tabs/AdditionalTabs.js
jeremypress: I think so. We need a way to get the image src download to kick off. If they aren't rendered this won't happen.

src/elements/content-sidebar/additional-tabs/AdditionalTabsLoading.scss
jeremypress: eee has a color but f4f4f4 does not. Updated the eee's reference.

src/elements/content-sidebar/additional-tabs/AdditionalTabs.js
jeremypress: We only need to do it once, but we need to wait until we're passed in the actual tab data, which doesn't happen on init.

src/elements/content-sidebar/__tests__/SidebarNav-test.js
jeremypress: I'm going to remove this condition. I think here I'll just test that the loading state renders, and inside the components i'll test the loaded state.

src/elements/content-sidebar/additional-tabs/AdditionalTab.js
```diff
@@ -0,0 +1,38 @@
...
+import Tooltip from '../../common/Tooltip';
+import PlainButton from '../../../components/plain-button/PlainButton';
+import IconEllipsis from '../../../icons/general/IconEllipsis';
+
+import './AdditionalTab.scss';
+
+type Props = {
+    isLoading: boolean,
+    onImageLoad: () => void,
+    ...AdditionalSidebarTab,
```
jstoffan: Should this use the `&` operator to indicate a type intersection?

src/elements/content-sidebar/additional-tabs/AdditionalTab.scss
```diff
@@ -0,0 +1,18 @@
...
+
+    height: 24px;
+    width: 24px;
+}
+
+.bdl-AdditionalTab-moreIcon .fill-color {
+    fill: $downtown-grey;
+}
+
+.bdl-AdditionalTab--is-hidden {
```
jstoffan: Modifiers are indicated with a single word, like `--primary`. States are indicated with a separate class, like `.bdl-is-hidden` or just `.is-hidden`.

src/elements/content-sidebar/additional-tabs/AdditionalTab.js
```diff
@@ -0,0 +1,38 @@
+/**
+ * @flow
+ * @file Preview sidebar additional tab component
```
jstoffan: These aren't strictly tied to Preview, so should this be something like `Sidebar Additional Tab component`?

src/elements/content-sidebar/additional-tabs/AdditionalTabs.js
```diff
@@ -0,0 +1,85 @@
...
+    numLoadedTabs: number = 0;
+
+    initialState: State = {
+        isLoading: true,
+    };
+
+    constructor(props: Props) {
+        super(props);
+
+        this.state = { ...this.initialState };
```
jstoffan: It doesn't look like we ever reset state to the initial value, so you could move state instantiation out of the constructor.

src/elements/content-sidebar/additional-tabs/AdditionalTabs.js
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+import React, { PureComponent } from 'react';
+import AdditionalTab from './AdditionalTab';
+import AdditionalTabsLoading from './AdditionalTabsLoading';
+
+import './AdditionalTabs.scss';
+
+type Props = {
+    tabs: ?Array<AdditionalSidebarTab>,
```
jstoffan: It looks like the parent component's flow type has this prop marked as optional rather than nullable.

src/elements/content-sidebar/additional-tabs/AdditionalTabs.js
```diff
@@ -0,0 +1,85 @@
...
+
+        this.state = { ...this.initialState };
+    }
+
+    componentDidUpdate() {
+        const { tabs } = this.props;
+
+        if (tabs) {
+            // If we're displaying the more options tab, consider it loaded since it doesn't use an image.
+            const hasMoreTab = tabs.find(tab => tab.id < 0 && !tab.iconUrl);
```
jstoffan: This logic being duplicated here and AdditionalTab makes it seem like the "more apps" button should be a separate component, instead.

src/elements/content-sidebar/additional-tabs/AdditionalTabsLoading.js
```diff
@@ -0,0 +1,28 @@
...
+const LOADING_TABS = [
+    AdditionalTabLoading,
+    AdditionalTabLoading,
+    AdditionalTabLoading,
+    AdditionalTabLoading,
+    AdditionalTabLoading,
+    MoreTabLoading,
+];
+
+const AdditionalTabsLoading = (): any => {
```
jstoffan: Can we avoid using `any` here?

src/elements/content-sidebar/additional-tabs/MoreTabLoading.js
```diff
@@ -0,0 +1,19 @@
...
+ * @author Box
+ */
+
+import * as React from 'react';
+import './AdditionalTabs.scss';
+import IconEllipsis from '../../../icons/general/IconEllipsis';
+
+const MoreTabLoading = () => {
+    return (
+        <div className="bdl-AdditionalTabLoading-item">
```
jstoffan: Should the class and component names match?

src/elements/content-sidebar/additional-tabs/MoreTabLoading.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * @flow
+ * @file Preview sidebar more tabs loading component
+ * @author Box
+ */
+
+import * as React from 'react';
+import './AdditionalTabs.scss';
```
jstoffan: SCSS imports should be at the end of the import list.

src/elements/content-sidebar/additional-tabs/__tests__/AdditionalTabsLoading-test.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import AdditionalTabsLoading from '../AdditionalTabsLoading';
+
+describe('elements/content-sidebar/AdditionalTabs', () => {
```
jstoffan: Do the test paths need to be updated to reflect the new subdirectory?

src/elements/content-sidebar/additional-tabs/MoreTabLoading.js
jeremypress: The styles are identical between the two components, would you recommend a more general name that applies to both?

PR #932, opened 2019-03-07
refactor(breadcrumb): remove unnecessary label prop

priyajeet:
Instead hardcode the translated value of breadcrumb


boxcla:

PR #935, opened 2019-03-08
feat(tasks): connect delete-task api for new task cards

alexkrolick:


boxcla:

src/elements/content-sidebar/activity-feed/task-new/Task.js
```diff
@@ -41,25 +41,26 @@ class Task extends React.Component<Props> {
...
             id,
             isPending,
+            mentionSelectorContacts,
+            name,
+            onAssignmentUpdate = noop,
             onDelete,
             onEdit,
-            onAssignmentUpdate = noop,
-            name,
+            permissions,
```
alexkrolick: diff here is +permissions and sorting the rest alphabetically

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -552,6 +552,14 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
             ? this.tasksApiNew.updateTaskAssignment
             : this.updateTaskAssignment;
 
+        const updateTask = isFeatureEnabled(features, 'activityFeed.tasks.newApi')
```
priyajeet: Feature strings have started to repeat. we should be making them constants as I think mentioned in a prior PR sometime back

alexkrolick: Since this is a path, we should try to come up with a flow type that can either validate the path based on the feature object or at least add the constant checks to the feature checking component (at the risk of restricting the paths further than is actually necessary). Otherwise we have several definitions for the allowed values that can get out of sync. I can look into that as a followup.

src/elements/content-sidebar/activity-feed/task-new/Task.js
```diff
@@ -41,25 +41,26 @@ class Task extends React.Component<Props> {
...
-            getAvatarUrl,
-            getUserProfileUrl,
-            getMentionWithQuery,
-            mentionSelectorContacts,
         } = this.props;
 
-        // Enable when edit/delete are implemented for new tasks
         const taskPermissions = {
+            ...permissions,
+            // can_edit: permissions.can_update,
```
jstoffan: We'd prefer to avoid commented code being committed.

alexkrolick: Will remove in follow-up implementation of edit function

PR #936, opened 2019-03-08
chore(query-bar): Reference conditions from parent app

ChenCodes:
Changes in this PR:
- FilterButton now accepts ```conditions``` prop passed down from EUA
- FilterButton's original state called ```conditions``` is now called ```transientConditions```
- Upon closing the FilterButton dropdown, ```this.state.transientConditions``` will get set to an empty Array
- Upon any changes between ```this.props.condition``` in FilterButton and ```this.state.transientConditions```, ```componentDidUpdate``` will handle hydrating local state with Redux state

### DIFF
--- a/src/features/query-bar/__tests__/FilterButton-test.js
+++ b/src/features/query-bar/__tests__/FilterButton-test.js
@@ -1,13 +1,13 @@
 // @flow
 import * as React from 'react';
 
-import { columns, initialCondition } from '../components/fixtures';
+import { columns, initialCondition, conditions } from '../components/fixtures';
 import FilterButton from '../components/filter/FilterButton';
 import { EQUALS, LESS_THAN } from '../constants';
 
 describe('feature/query-bar/components/filter/FilterButton', () => {
     const getWrapper = (props = {}) => {
-        return shallow(<FilterButton {...props} />);
+        return shallow(<FilterButton conditions={conditions} {...props} />);
     };
 
     describe('render', () => {
@@ -22,6 +22,18 @@ describe('feature/query-bar/components/filter/FilterButton', () => {
         });
     });
 
+    describe('componentDidUpdate()', () => {
+        const transientConditions = [];
+        const wrapper = getWrapper({
+            conditions,
+        });
+        wrapper.setState({
+            transientConditions,
+        });
+        wrapper.instance().componentDidUpdate();
+        expect(wrapper.state('transientConditions')).toEqual(conditions);
+    });
+
     describe('handleColumnChange()', () => {
         const condition = {
             columnId: '1',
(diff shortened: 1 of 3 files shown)


boxcla:

src/features/query-bar/__tests__/FilterButton-test.js
```diff
@@ -184,29 +196,27 @@ describe('feature/query-bar/components/filter/FilterButton', () => {
...
                 id: conditionID,
                 operator: EQUALS,
                 values: [],
             };
             wrapper.instance().setState({
-                conditions: [],
+                transientConditions: [],
             });
 
-            const condition = wrapper.instance().createCondition(conditionID);
```
ChenCodes: Removed this param because in the method it does not take any params.

src/features/query-bar/__tests__/FilterButton-test.js
```diff
@@ -184,29 +196,27 @@ describe('feature/query-bar/components/filter/FilterButton', () => {
     describe('createCondition()', () => {
         test('Should return a condition object if columns is non-empty', () => {
             const wrapper = getWrapper({ columns });
-            const conditionID = '11';
+            const conditionID = '12';
```
ChenCodes: This change is due to uniqueId generating a random number, in the case of this test, it is '12' and not '11' which I used before to pass in to ```createCondition```.

src/features/query-bar/components/filter/FilterButton.js
```diff
@@ -46,13 +47,24 @@ class FilterButton extends React.Component<Props, State> {
...
         this.setState({
-            conditions: [initialCondition],
+            transientConditions: [initialCondition],
         });
     }
 
+    componentDidUpdate() {
+        const { conditions } = this.props;
+        const { transientConditions } = this.state;
+        if (transientConditions.length === 0 && conditions.length !== transientConditions.length) {
```
ChenCodes: The only time we should update transientConditions is if there's a mismatch between ```this.props.conditions``` and ```this.state.transientConditions```.

ChenCodes: The empty array check is due to the ```onClose``` method setting transientConditions to be an empty array when the FilterButton dropdown is closed.

src/features/query-bar/__tests__/FilterButton-test.js
macrotim: We don't want tests to break randomly like this.

Alternatively, you can verify...
* That conditionId is non-empty
* columnId == 1
* Operator is EQUALS
* values == []

src/features/query-bar/components/filter/FilterButton.js
macrotim: >         if (transientConditions.length === 0 && conditions.length !== transientConditions.length) {

Could this be expressed differently e.g. "if conditions is not empty and transient conditions is empty..." 

I'd like to see "const wasFlyoutOpen = transientConditions.length === null". That logic is implied and I'd like to see it made more obvious.

The if-condition should say "if flyout was not open but is open now, then copy to transient conditions". Is this possible?

priyajeet:
whats eua?

ChenCodes:
EUA is EndUserApp

priyajeet:
Whats EndUserApp? Why does it need to be referenced in an open source repo

PR #941, opened 2019-03-09
fix(badgeable): align badges

alexkrolick:
Realigns badge based on latest design - increases offset outside of container


boxcla:

PR #945, opened 2019-03-12
fix(content-sidebar): Use mixin to fix specificity for additional tabs

jeremypress:


boxcla:

PR #946, opened 2019-03-12
docs(icons): Improve icon docs

tjgupta:
Added index pages for icons that were a cohesive set of icons. May attempt the `general` set next, though I'm not sure whether this type of documentation will work for that set... 

Note: somehow I got an updated translations file during the push. Maybe some other push did `--no-verify`?


boxcla:

examples/src/IconsExample.js
```diff
@@ -42,7 +45,7 @@ class IconsExample extends React.Component<Props, State> {
                         const Component = icon.component;
                         return (
                             <div className="icon" key={icon.name}>
-                                <Component />
+                                {icon.content ? icon.content() : <Component />}
```
tjgupta: To allow for docs that need multiple variations of a single component displayed, like this:

Screen Shot 2019-03-11 at 9 23 19 PM

src/icons/annotations/index.md
```diff
@@ -0,0 +1,26 @@
+**Annotation icons**
```
jstoffan: Does Github display `index.md` files when you navigate to the folder via a browser? If not, could we name these centralized files `README.md`?

tjgupta: Nope, doesn't seem like it. I switched them to `README.md`.

PR #948, opened 2019-03-12
feat(upload-manager): Add error case to upload manager name conflict

ChenCodes:
Screen Shot 2019-03-12 at 2 45 51 PM

Changes in this PR:
- Add error case for uploading a file with a name that already exists


boxcla:

PR #951, opened 2019-03-13
fix(usm): Update people with link description

tjgupta:
Screen Shot 2019-03-12 at 5 32 49 PM


boxcla:

PR #953, opened 2019-03-13
fix(unified-share-modal): show public link state in modal

prakash-kumar:


boxcla:

prakash-kumar:
sharedstate

src/features/unified-share-modal/UnifiedShareModal.scss
```diff
@@ -208,6 +208,24 @@
...
+        margin-top: 10px;
+
+        .shared-link-icon-globe {
+            padding-right: 5px;
+
+            .icon-globe {
+                display: inline-block;
+                margin-bottom: -2px;
+
+                .fill-color {
```
tjgupta: This icon has a `color` prop. Should we use that in this case?

prakash-kumar: was thinking about it. but went this to keep the mode of styling consistent (css).. this would be the only thing that does styling through the js if i went the other way.. what do you think?

PR #954, opened 2019-03-13
fix(tasks): improve task update/create callbacks

alexkrolick:
- fix pending state:
  - add styles to task card to match comment pending state
  - add CSS transition
- fix update logic of task assignment by refetching task to get updated status
- improve error handling of initial task create using try/catch in async/await context

### DIFF
--- a/src/api/Feed.js
+++ b/src/api/Feed.js
@@ -24,6 +24,7 @@ import {
     TYPED_ID_FEED_PREFIX,
     HTTP_STATUS_CODE_CONFLICT,
     IS_ERROR_DISPLAYED,
+    ERROR_CODE_CREATE_TASK,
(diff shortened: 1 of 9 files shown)


boxcla:

src/api/Feed.js
```diff
@@ -392,15 +393,16 @@ class Feed extends Base {
             id: taskCollaboratorId,
             status: taskCollaboratorStatus,
         };
+        const handleError = (e: ElementsXhrError, code: string) => {
```
jstoffan: It looks like having the success/error handler wrappers inline or as instance properties is common across most other API calls. Can this stay consistent as well?

alexkrolick: This needs to be passed in 2 places now, rather than just the one callback. It looks like the inline callback was being wrapped.

PR #955, opened 2019-03-13
feat(link-settings): add ability to display current access level

booc0mtaco:


boxcla:

booc0mtaco:
Screen Shot 2019-03-13 at 12 21 05 AM

src/features/shared-link-settings-modal/SharedLinkSettingsModal.js
```diff
@@ -15,8 +15,31 @@ import ExpirationSection from './ExpirationSection';
...
+ * Return the translation key based on the access level and whether the user can download
+ * @param {string} accessLevel one of 'peopleWithTheLink', 'peopleInYourCompany', or 'peopleInThisItem'
+ * @param {boolean} canDownload prop value for whether the user can currently download
+ *
+ * @return {string} message ID for the proper transloation (or empty string when nothing matches)
+ */
+function getAccessNoticeMessageId(accessLevel, canDownload) {
+    switch (accessLevel) {
+        case PEOPLE_WITH_LINK:
+            return canDownload ? 'withLinkViewDownload' : 'withLinkView';
```
priyajeet: You can use messages.xyz format directly, so that it's a reference and not a string constant, which is duplicated and harder to refactor. Also can try to minimize returns so that while debugging we aren't putting a break points in multiple lines vs just one or two returns to see what the function evaluates.

message = messages.inItem;
...

return message;
// OR return <FormattedMessage {...message}>

src/features/shared-link-settings-modal/SharedLinkSettingsModal.js
```diff
@@ -15,8 +15,31 @@ import ExpirationSection from './ExpirationSection';
...
+        case PEOPLE_IN_COMPANY:
+            return canDownload ? 'inCompanyViewDownload' : 'inCompanyView';
+        case PEOPLE_IN_ITEM:
+            return 'inItem';
+        default:
+            // return falsy meaning 'there is no key to return'
+            return '';
```
priyajeet: Should it ever be here?

booc0mtaco: it shouldn't, but we have some lint rule that requires adding `default`

src/features/shared-link-settings-modal/messages.js
```diff
@@ -58,7 +58,7 @@ const messages = defineMessages({
         id: 'boxui.share.sharedLinkSettings.passwordTitle',
     },
     vanityNamePlaceholder: {
-        defaultMessage: 'Enter a custom path',
+        defaultMessage: 'Enter a path name (12 or more characters)',
```
mattnguyen1: should this say 12 to 30 characters, since there is an actual limit?

booc0mtaco: holding for confirmation on this text string. might not include the limits at all

booc0mtaco: on second thought, there is nothing in the component to enforce this currently, so this string acts as a recommendation. I think we should consider a refactor where client-side validation can be built against props. Any server-side validation can hook to the error props

src/features/shared-link-settings-modal/SharedLinkSettingsModal.js
```diff
@@ -34,6 +63,9 @@ class SharedLinkSettingsModal extends Component {
          */
         onSubmit: PropTypes.func.isRequired,
 
+        // access level props
+        accessLevel: PropTypes.string,
```
tjgupta: Should we be removing the `warnOnPublic` prop then? Also, do you think there's value in just passing in the whole `sharedLink` object to future proof this?

booc0mtaco: oh yes, good call on removal. 

I thought about it, and I decided to add the one key value since the old modals (and this one) will eventually go away.

PR #956, opened 2019-03-13
build(webpack): Remove devServer option interfering with yarn start:dev

jstoffan:
We’ll need to look into re-enabling the option if we decide to support deep linking in the future, likely via an environment flag of some kind. It's not needed right now, though.


boxcla:

PR #962, opened 2019-03-14
fix(sidebar): Render sidebar redirect only if no other route matches

jstoffan:
`Route` components always render `children` regardless of whether their path matches or not. Routes without paths always match, so in this case the redirect is always rendered. This doesn't actually cause any issues at the moment, thanks to the wrapping `<Switch>`, but it should be fixed anyway.


boxcla:

PR #965, opened 2019-03-14
chore(appactivity): Add accept-language headers to activity GET

JustinHoldstock:
Also added headers to MarkerBasedAPI GET


boxcla:

src/api/AppActivity.js
```diff
@@ -115,20 +120,26 @@ class AppActivity extends MarkerBasedAPI {
...
         errorCallback: ElementsErrorCallback,
+        language?: string = DEFAULT_LOCALE,
     ): void {
         const requestData = {
             item_id: id,
             item_type: 'file',
             fields: APP_ACTIVITY_FIELDS_TO_FETCH.toString(),
         };
 
+        const headers = {
```
priyajeet: Can we just add this to Xhr.js for all XHRs?

JustinHoldstock: That's not a terrible idea. So you're imagining we have something like a `defautlHeaders` object that we can spread into the result going to all requests?

src/elements/content-sidebar/ContentSidebar.js
```diff
@@ -107,17 +113,23 @@ class ContentSidebar extends React.Component<Props, State> {
...
 
+        const commonHeaders = {
+            [HEADER_ACCEPT_LANGUAGE]: language || DEFAULT_LOCALE,
+        };
+
         this.api = new API({
             apiHost,
             cache,
             clientName,
+            commonHeaders,
```
jstoffan: Can we pass in `language` and allow the API class to manage the headers? Otherwise each top-level Element will need to redefine the headers.

src/utils/Xhr.js
```diff
@@ -23,6 +25,11 @@ type PayloadType = StringAnyMap | Array<StringAnyMap>;
 
 const DEFAULT_UPLOAD_TIMEOUT_MS = 120000;
 const MAX_NUM_RETRIES = 3;
+const DEFAULT_COMMON_HEADERS = {
+    Accept: 'application/json',
+    [HEADER_ACCEPT_LANGUAGE]: DEFAULT_LOCALE,
```
jstoffan: This seems like a major change in disguise, as I believe the browser sets this for the user currently. Does this cause issues for other elements that aren't passing in the user-defined locale?

JustinHoldstock: In most cases with our APIs that look at accept-language, if it's missing it defaults to en-US.

I'm going to remove this. Thanks for catching it!

src/elements/content-sidebar/ContentSidebar.js
JustinHoldstock: In this case, that's fine by me. If a need to add more headers occurs, we'll want to switch back to `headers`

src/utils/Xhr.js
```diff
@@ -92,6 +101,12 @@ class Xhr {
 
         this.axios.interceptors.response.use(this.responseInterceptor, this.errorInterceptor);
 
+        if (language) {
```
JustinHoldstock: I'm going to make a method that assembles headers from the options object, instead of this

src/utils/Xhr.js
```diff
@@ -181,6 +195,21 @@ class Xhr {
         };
     }
 
+    /**
+     * Create a default headers object
+     *
+     * @param {DefaultHeaderOptions} options - Used to create default headers
+     */
+    getDefaultHeaders({ language }: DefaultHeaderOptions): StringMap {
```
priyajeet: Can this be just part of the `getHeaders` function? Unsure having a default getting makes any difference. Specially considering `getHeaders` also does similar `if () then (`) for headers.

JustinHoldstock: Yeah, I guess I can just introduce a new class property for language, then we can remove `this.headers`

src/api/MarkerBasedAPI.js
```diff
@@ -49,6 +50,7 @@ class MarkerBasedApi extends Base {
         limit: number,
         shouldFetchAll: boolean,
         requestData: Object = {},
+        headers?: StringMap,
```
jstoffan: Where is this being used?

src/elements/content-open-with/ContentOpenWith.js
```diff
@@ -224,10 +224,10 @@ class ContentOpenWith extends PureComponent<Props, State> {
      * @return {void}
      */
     fetchOpenWithData(): void {
-        const { fileId, language }: Props = this.props;
+        const { fileId }: Props = this.props;
```
jstoffan: Does `language` need to be passed in to the API constructor if we're removing it here?

src/api/MarkerBasedAPI.js
JustinHoldstock: Removed! It'll be in history if it needs to be added back

src/elements/content-open-with/ContentOpenWith.js
JustinHoldstock: Great catch. I missed a couple of the cases where we create API instances

src/api/OpenWith.js
```diff
@@ -34,25 +28,13 @@ class OpenWith extends Base {
...
         this.errorCode = ERROR_CODE_FETCH_INTEGRATIONS;
-        const params = {
-            headers: {
-                [HEADER_ACCEPT_LANGUAGE]: locale,
-            },
-        };
 
         this.get({
             id: fileId,
-            requestData: params,
```
jstoffan: @jeremypress, can you take a look at this change?

jeremypress: @JustinHoldstock where is the default locale being handled now?

JustinHoldstock: All XHR requests, as long as `language` is being passed into the `APIFactory` instance, that owns this. Right now, ContentPreview, ContentOpenWith, and ContentSidebar API calls all receive `accept-language` headers with that value

ConradJChan:
Addressed by https://github.com/box/box-ui-elements/pull/1368

PR #967, opened 2019-03-14
docs(icons): More icon docs cleanup

tjgupta:


boxcla:

tjgupta:
@booc0mtaco I'm still trying to figure out a good way to show props for a family of icons with this style of documentation... it seems we lose the auto generated props when not using the `components` key in the config. 

I think the tradeoff of readable docs is worth the loss of the props documentation for now though. I'll play with that more and see if I can get the props back in there... or maybe we just render them manually for each family of icons?

PR #968, opened 2019-03-14
fix(countbadge): add horizontal padding to count

alexkrolick:
Screen Shot 2019-03-14 at 4 45 40 PM
Screen Shot 2019-03-14 at 4 45 22 PM


boxcla:

booc0mtaco:
Can you include a screenshot of the `badgeable` example with this change?

alexkrolick:
@booc0mtaco see attached

src/components/count-badge/CountBadge.scss
```diff
@@ -35,7 +35,7 @@
     height: 16px;
     line-height: 16px;
     min-width: 16px;
-    padding: 0 2px;
+    padding: 0 4px;
```
booc0mtaco: we'll soon start coalescing on a 4-pixel grid for our sizings, so this component is now more aligned on that

PR #970, opened 2019-03-15
feat(versions): Add version history items and navigation

jstoffan:
Adds the version history items and support for previewing prior versions.

Version History Items

Todo:
- [x] Tests


boxcla:

src/elements/content-sidebar/versions/VersionsItem.scss
```diff
@@ -0,0 +1,74 @@
...
+    flex: 0 0 40px;
+    font-size: 18px;
+    margin-right: 10px;
+    padding: 10px 5px;
+    text-align: center;
+}
+
+.bcs-VersionsItem-details {
+    flex: 1 1 100%;
+    overflow: hidden;
```
ConradJChan: Is this overflow required?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -5,43 +5,75 @@
...
+        const { onVersionChange, versionId } = this.props;
+
+        // Forward the current version id that is passed in via the wrapping route
+        if (prevVersionId !== versionId) {
+            onVersionChange(versionId);
+        }
+    }
+
+    componentWillUnmount() {
+        this.props.onVersionChange();
```
ConradJChan: What's the reason for calling `onVersionChange` while unmounting?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -5,43 +5,75 @@
...
     };
 
+    static defaultProps = {
+        onVersionChange: noop,
+    };
+
     componentDidMount() {
         const { api, fileId } = this.props;
 
         api.getVersionsAPI(false).getVersions(fileId, this.handleFetchVersionsSuccess, this.handleFetchVersionsError);
```
ConradJChan: Do we plan on showing the current version in the version history sidebar? If there is only v1, would this sidebar still show the v1?

src/components/loading-indicator/Crawler.scss
```diff
@@ -6,6 +6,7 @@
 
 .crawler {
     font-size: 0;
+    line-height: 1;
```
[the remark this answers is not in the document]
jstoffan: It vertically centers the inner divs inside the `.crawler` parent div. They're currently offset by about 5 pixels without this change.

src/elements/common/nav-button/NavButton.js
```diff
@@ -0,0 +1,61 @@
...
+    component?: any,
+    exact?: boolean,
+    onClick?: (event: SyntheticEvent<>) => void,
+    replace?: boolean,
+    strict?: boolean,
+    to: string | Location,
+};
+
+const NavButton = ({
+    activeClassName = 'bdl-is-selected',
```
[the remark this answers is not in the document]
jstoffan: I'm not a big fan of sharing class names as constants across components, since it breaks encapsulation. This class in particular also isn't really meant to share styles with anything else; it's a modifier for this specific component. I could rename it to something like `bcs-Versions-Item-is-selected` or `bcs-VersionsItem--selected`, instead?

src/elements/content-sidebar/versions/VersionsItem.js
```diff
@@ -0,0 +1,86 @@
...
+    modified_at: modifiedAt,
+    modified_by: modifiedBy = PLACEHOLDER_USER,
+    size,
+    version_number: versionNumber,
+}: Props) => {
+    const isDeleted = action === VERSION_DELETE_ACTION;
+    const className = classNames('bcs-VersionsItem', {
+        'bcs-is-disabled': isDeleted,
+    });
+    const versionName = modifiedBy.name || 'Unknown';
```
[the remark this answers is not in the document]
jstoffan: I was thinking not, since it's meant to be a replacement for a name, which aren't translated. I can add it to the messages file, though.

src/elements/content-sidebar/versions/VersionsItem.js
```diff
@@ -0,0 +1,86 @@
...
+    const versionPath = generatePath(match.path, { ...match.params, versionId });
+
+    return (
+        <NavButton
+            activeClassName="bcs-is-selected"
+            className={className}
+            component="button"
+            disabled={isDeleted}
```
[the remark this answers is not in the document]
jstoffan: Yep!

src/elements/content-sidebar/versions/VersionsItem.scss
jstoffan: Yes, in order to support text overflow of its child components.

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
jstoffan: It's essentially a shortcut to reset the current version id, since the location no longer contains one once this component is unmounted.

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
jstoffan: The data layer on this stuff is still pretty basic at the moment. The idea is to only show version history if there's more than one version, though.

src/elements/content-preview/ContentPreview.js
```diff
@@ -62,6 +62,7 @@ type Props = {
     enableThumbnailsSidebar: boolean,
     features?: FeatureConfig,
     fileId?: string,
+    fileOptions?: Object,
```
JustinHoldstock: Is there a type we can introduce?

jstoffan: I can try, but this object is defined in the Preview SDK and consists of dynamic keys (file ids), each with its own options object. It doesn't seem ideal to declare and maintain types for other libraries.

src/elements/common/nav-button/NavButton.js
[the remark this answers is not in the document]
ConradJChan: that makes sense, both those alternatives seem good to me!

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
ConradJChan: From playing around with the current webapp UI, it does looks like even with V1 we can open the version history modal, so as long as this approach is an intentional change instead of an unexpected regression we should be good

jstoffan: Good point. Will validate with product before I introduce that logic.

examples/src/features.js
```diff
@@ -8,4 +8,5 @@ export default {
             newCards: true,
         },
     },
+    versions: true,
```
jstoffan: Will nuke this before merging.

src/elements/content-sidebar/versions/VersionsItem.js
```diff
@@ -0,0 +1,88 @@
...
+}: Props) => {
+    const isDeleted = action === VERSION_DELETE_ACTION;
+    const className = classNames('bcs-VersionsItem', {
+        'bcs-is-disabled': isDeleted,
+    });
+    const versionPath = generatePath(match.path, { ...match.params, versionId });
+    const versionUser = modifiedBy.name || <FormattedMessage {...messages.versionUserUnknown} />;
+
+    return (
+        <NavButton
```
jeremypress: If i'm thinking of the right component, clicking on this adds the selected style but doesn't actually navigate you anywhere? Do we need to update our route if we haven't previewed it yet?

jstoffan: The NavButton is very similar to react-router's NavLink, just using a button instead of an anchor. Clicking on one of these does update the history route in memory and will trigger Preview to load the selected version. The NavButton will then see its `to` prop matches the current route and automatically apply its `activeClassName` to itself.

src/elements/content-preview/ContentPreview.js
```diff
@@ -338,14 +345,16 @@ class ContentPreview extends PureComponent<Props, State> {
...
-        const { file: prevFile }: State = prevState;
+        const { file, fileVersionId }: State = this.state;
+        const { file: prevFile, fileVersionId: prevFileVersionId }: State = prevState;
         const versionPath = 'file_version.id';
         const previousVersionId = getProp(prevFile, versionPath);
         const currentVersionId = getProp(file, versionPath);
         let loadPreview = false;
 
-        if (previousVersionId && currentVersionId) {
+        if (prevFileVersionId !== fileVersionId) {
```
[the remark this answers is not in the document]
jstoffan: Did my best!

PR #971, opened 2019-03-15
feat(avatar): add size prop to avatars

alexkrolick:
Adds a size prop (currently restricted to "large") for avatars. The default CSS matches the design for the components using HeaderFlyout (44px width, 14px font size).

### DIFF
--- a/src/components/avatar/Avatar.js
+++ b/src/components/avatar/Avatar.js
@@ -1,5 +1,6 @@
 // @flow
 import * as React from 'react';
+import classNames from 'classnames';
 import AvatarImage from './AvatarImage';
 import AvatarInitials from './AvatarInitials';
 import UnknownUserAvatar from '../../icons/avatars/UnknownUserAvatar';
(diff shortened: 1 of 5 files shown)


boxcla:

src/components/avatar/Avatar.scss
```diff
@@ -35,4 +35,14 @@
         display: flex;
         justify-content: center;
     }
+
+    &.large {
```
jstoffan: Can this be updated to use SUIT naming conventions? The class name is pretty generic otherwise. Something like `avatar--large`?

src/components/avatar/Avatar.js
```diff
@@ -50,8 +54,10 @@ class Avatar extends React.PureComponent<Props, State> {
     };
 
     render() {
-        const { avatarUrl, className = '', name, id }: Props = this.props;
+        const { avatarUrl, className = '', name, id, size }: Props = this.props;
         const { hasImageErrored }: State = this.state;
+        // restrict CSS classnames to known sizes
+        const sizeClass = SIZES[size] ? size : '';
```
jstoffan: Seems like this could benefit from using the `classnames` library.

src/components/avatar/Avatar.scss
alexkrolick: Not updating the whole file to use SUIT since I see several overrides for `.avatar-initial`, etc., but I'll make this change.

PR #972, opened 2019-03-15
fix(headerflyout): remove caret icon

alexkrolick:


boxcla:

src/features/header-flyout/HeaderFlyout.js
```diff
@@ -30,10 +30,15 @@ class HeaderFlyout extends React.Component<Props> {
...
 
     render() {
-        const { header, footer, flyoutButton, children, scrollRefFn, ...rest } = this.props;
+        const { header, footer, flyoutButton, children, scrollRefFn, className, ...rest } = this.props;
 
         return (
-            <Flyout closeOnClick={false} offset={HeaderFlyout.panelOffset} {...rest}>
+            <Flyout
+                closeOnClick={false}
+                offset={HeaderFlyout.panelOffset}
```
booc0mtaco: When opened, the offset should now align the flyout vertically with the button clicked. with the caret, it was offset by a little bit. If I remember correctly, we need to shift this over as well

alexkrolick: Screen Shot 2019-03-15 at 12 20 39 PM

alexkrolick:
@booc0mtaco amended commit name

PR #974, opened 2019-03-15
style(searchform): restore outline on search form buttons

cweeii:
image

image

### DIFF
--- a/src/components/search-form/SearchForm.scss
+++ b/src/components/search-form/SearchForm.scss
@@ -17,11 +17,6 @@
         padding: 0;
         width: 30px;
 
-        &:hover,
-        &:focus {
-            outline: none;
-        }
-
         path {
             transition: fill linear .1s;
         }


boxcla:

src/components/search-form/SearchForm.scss
```diff
@@ -17,11 +17,6 @@
         padding: 0;
         width: 30px;
 
-        &:hover,
```
jstoffan: This is also used in ContentExplorer and MetadataSidebar. Can you check those quickly to make sure they don't have any glaring issues after this change?

cweeii: Just took a look at both and no glaring issues.  Only difference is the outline shows up now when you tab to it.

ConradJChan:
Was there previously a different hover/focus effect in place of outline?

patrickpaul:
Given recent changes to the design spec, this PR is no longer applicable.

PR #975, opened 2019-03-15
fix(query-bar): fix tooltip text for Condition

ChenCodes:
Changes in this PR:
- `TextInput` fields now have an error tooltip that says `Please enter a value`
- Use constants instead of hardcoded strings in Condition.js and ValueField.js
- Fix UI issue with single select enum where the placeholder isn't shown when empty string is passed in

Screen Shot 2019-03-15 at 2 42 14 PM

### DIFF
--- a/i18n/en-US.properties
+++ b/i18n/en-US.properties
@@ -800,6 +800,8 @@ boxui.queryBar.selectValuePlaceholderText = Select value
 boxui.queryBar.templatesButtonText = Select Metadata
 # Text on the templates button when templates are still being loaded
 boxui.queryBar.templatesLoadingButtonText = Template Name
+# Text displayed on the Tooltip for an input field
+boxui.queryBar.tooltipEnterValueError = Please Enter a Value
 # Text displayed on the Tooltip for a date field
 boxui.queryBar.tooltipSelectDateError = Please Select a Date
 # Text displayed on the Tooltip for a value field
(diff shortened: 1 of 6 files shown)


boxcla:

i18n/en-US.properties
```diff
@@ -800,6 +800,8 @@ boxui.queryBar.selectValuePlaceholderText = Select value
 boxui.queryBar.templatesButtonText = Select Metadata
 # Text on the templates button when templates are still being loaded
 boxui.queryBar.templatesLoadingButtonText = Template Name
+# Text displayed on the Tooltip for a input field
```
ChenCodes: Grammar

src/features/query-bar/__tests__/__snapshots__/ValueField-test.js.snap
```diff
@@ -63,7 +63,7 @@ exports[`features/query-bar/components/filter/ValueField render value fields "sh
       id="boxui.queryBar.selectValuePlaceholderText"
     />
   }
-  selectedValue=""
+  selectedValue={null}
```
ChenCodes: This got updated because the placeholder only shows up when the selectedValue is null and not empty string.

src/features/query-bar/messages.js
```diff
@@ -2,6 +2,11 @@
 import { defineMessages } from 'react-intl';
 
 const messages = defineMessages({
+    tooltipEnterValueError: {
+        defaultMessage: 'Please Enter a Value',
+        description: 'Text displayed on the Tooltip for a input field',
```
ChenCodes: Grammar

PR #976, opened 2019-03-15
feat(i18n): Implement the new FormattedCompMessage component

ehoogerbeets:
Implement a new component called `FormattedCompMessage`. This translates strings which can contain HTML or other JSX components in a safe way by hiding the contents of those tags from the translators or hackers and substituting the contents from source strings into the translations. Example:

```javascript
<FormattedCompMessage id="mystring" description="translator's comment">
   You can <span class="foo">delete</span> <Link to={url}>these files</Link>, or just
  <span id="x" class="unshare">unshare them</span>.
</FormattedCompMessage>
```

`FormattedCompMessage` is built as a layer on top of react-intl, and uses it to do the actual translations. Its function is to transform a tree of React elements into a coded string, get that string translated with react-intl, transform the coded translated string back into a tree of React elements, and finally substitute the hidden React elements from the source tree back into the translation tree in the right locations.

The component depends on the new babel plugin `babel-plugin-box-i18n` to extract the coded texts from the source code and place them into the same type of json files that the react-intl babel plugin produces. In this way, the translation process downstream from there works with no further changes. The translations appear in the properties files some time later and react-intl picks them up as normal. The `FormattedCompMessage` can retrieve these translations using react-intl. The `babel-pluging-box-i18n` plugin also enforces Box's policies, such as the requirement that all strings have a unique id and a description for the translators.

The `FormattedCompMessage` component can support plurals in a React-idiomatic way by embedding `<Plural>` components in the body of the `FormattedCompMessage`:

```javascript
<FormattedCompMessage id="mystring" description="translator's comment" count={count}>
  <Plural category="one">This is the <strong>singular</strong> string.</Plural>
  <Plural category="other">This is the <strong>plural</strong> string.</Plural>
</FormattedCompMessage>
```
Note that HTML and subcomponents are allowed in the plural string as well.

The `FormattedCompMessage` component supports replacement parameters in a React-idiomatic way by embedding `<Param>` components in the body of the `FormattedCompMessage`:

```javascript
<FormattedCompMessage id="mystring" description="translator's comment">
  User <Param value={user.name} description="The user's full name"/> has already deleted this file.
</FormattedCompMessage>
```

Note that the `Param` tags can be used inside of `Plural` tags as well, as they are normal components.


boxcla:

package.json
```diff
@@ -280,5 +279,9 @@
         "scroll-into-view-if-needed": "^1.5.0",
         "tabbable": "^1.1.2",
         "uuid": "^3.3.2"
+    },
+    "dependencies": {
```
priyajeet: should these be dev deps?
Normally we dont add any deps, only dev deps and/or peer deps

priyajeet: Also I am not sure if we should be adding dependency to libraries which dont have wide support.

src/components/i18n/__tests__/FormattedCompMessage-tests.js
```diff
@@ -0,0 +1,446 @@
+import React from 'react';
```
priyajeet: file name is -tests plural, should be -test

src/components/i18n/__tests__/Param-tests.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
```
priyajeet: same comment about file name

src/components/i18n/__tests__/Plural-tests.js
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
```
priyajeet: same comment about file name

package.json
ehoogerbeets: I can move them to to peer dependencies.

These libraries are ones that I wrote so that the babel plugin and this component share the way that they convert a tree of components and strings into a single translatable string and back.

src/components/i18n/Param.js
```diff
@@ -0,0 +1,62 @@
...
+                }
+                if (React.isValidElement(this.props.value)) {
+                    return this.props.value;
+                }
+                return this.props.value.toString();
+        }
+    }
+
+    render() {
+        if ((process.env.NODE_ENV === 'test' || process.env.NODE_ENV === 'dev') && !this.props.description) {
```
priyajeet: We have this check in multiple places, can we just make it part of some utility function in say src/utils/env.js?

src/components/i18n/Param.js
```diff
@@ -0,0 +1,62 @@
...
+ * @example
+ * <pre>
+ *   <FormattedCompMessage id="x" description="y">
+ *     The file <Param value={filelist[i].path} description="Name of the file that was deleted."/> has been deleted.
+ *   </FormattedCompMessage>
+ * </pre>
+ */
+class Param extends React.Component<Props> {
+    getValue() {
+        switch (typeof this.props.value) {
```
priyajeet: const { value } = this.props;

so that we arent doing this.props.value in all places

src/components/i18n/Param.js
```diff
@@ -0,0 +1,62 @@
...
+                return '';
+
+            case 'boolean':
+            case 'number':
+                return String(this.props.value);
+
+            case 'function':
+                return this.props.value();
+
+            case 'string':
```
priyajeet: We should also make the strings into constants and share it between the files. The other file also uses them.

alexkrolick:
Can you snapshot the actual HTML for some of the output? That would clarify the result of using these snapshots. Also, some documentation for the required translation format in the messages files would help users of the component.

ehoogerbeets:
There are lots of examples in the README.md, but none of the HTML output or the strings that get translated. I will add some description there.

package.json
ehoogerbeets: Question: now the CI build fails because the peer dependencies for my new modules are not there in the travis node_modules cache. How do I update that cache with those new modules?

src/components/i18n/FormattedCompMessage.js
```diff
@@ -0,0 +1,205 @@
...
+            id, // the unique id of the string
+            defaultMessage, // The English string + HTML + components that you want translated
+            description, // An explanation of the message for the translators
+            count, // the pivot count to choose a plural form
+            children, // the components within the body
+        } = this.props;
+
+        const sourceElements = defaultMessage || children;
+
+        if (isDevEnvironment()) {
```
jstoffan: Defining these props as required in the accompanying Flow types should be adequate to ensure they're passed in.

src/components/i18n/FormattedCompMessage.js
```diff
@@ -0,0 +1,205 @@
...
+                defaultMessage: source,
+                description,
+            },
+            values,
+        );
+
+        // always wrap the translated string in a tag to contain everything
+        // and to give us a spot to record the id
+        return React.createElement(
+            tagName || 'span',
```
jstoffan: It looks like `'span'` is already defined as the defaultProp value above.

src/components/i18n/FormattedCompMessage.js
```diff
@@ -0,0 +1,205 @@
...
+            {
+                key: id,
+                'x-resource-id': id,
```
jstoffan: Can you add a comment explaining the purpose of this attribute?

src/components/i18n/Param.js
```diff
@@ -0,0 +1,73 @@
...
+                }
+                return value.toString();
+        }
+    }
+
+    render() {
+        if (isDevEnvironment() && !this.props.description) {
+            throw new Error('The description property is required on a Param component.');
+        }
+        return this.getValue();
```
jstoffan: Minor, but it seems like the logic of `getValue` can live inside `render`.

src/components/i18n/Param.js
```diff
@@ -0,0 +1,73 @@
...
+ * component. Typically, this component is self-closing.
+ *
+ * @example
+ * <pre>
+ *   <FormattedCompMessage id="x" description="y">
+ *     The file <Param value={filelist[i].path} description="Name of the file that was deleted."/> has been deleted.
+ *   </FormattedCompMessage>
+ * </pre>
+ */
+class Param extends React.Component<Props> {
```
jstoffan: Components that don't contain state or lifecycle methods should ideally be pure function components rather than classes for performance reasons.

src/components/i18n/Plural.js
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * This component does not add any functionality beyond its contents, and its only
+ * purpose is to enclose some JSX to use for a particular plural category.
+ *
+ * See the [Unicode CLDR description of plural category
+ * rules](http://cldr.unicode.org/index/cldr-spec/plural-rules) for more details.
+ */
+class Plural extends React.Component<Props> {
+    // mostly for unit testing
+    getSourceString() {
```
jstoffan: We should ideally avoid introducing code paths that are only used for testing. Can the `Composition` object be tested on its own, rather than as part of the `Plural` component?

src/constants.js
```diff
@@ -364,5 +364,22 @@ export const KEYS = {
     enter: 'Enter',
 };
 
+/* ------------------ Javascript Object Types ----------------- */
```
jstoffan: You can isolate these constants to a file in the component directory if they're not intended to be used elsewhere.

yarn.lock
```diff
@@ -8295,6 +8295,11 @@ ignore@^4.0.0, ignore@^4.0.3:
   resolved "https://registry.yarnpkg.com/ignore/-/ignore-4.0.6.tgz#750e3db5862087b4737ebac8207ffd1ef27b25fc"
   integrity sha512-cyFDKrqc/YdcWFniJhzI42+AzS+gNwmUzOSFcRCQYwySuBBBy/KjuxWLZ/FHEH6Moq1NizMOBWyTcv8O4OZIMg==
 
+ilib-tree-node@^1.2.0:
```
jstoffan: Have these been reviewed?

package.json
alexkrolick: Add peer deps to dev dependencies as well if they need to be run in tests

priyajeet: So unfortunately we can't add libraries that folks wrote on their own. We want to only use libraries that have wide community support for future maintenance. They should either be made part of this commit, so that they are reviewed or the functions extracted out here. We can't assume that some library will always be maintained when someone leaves box. Have had problems in the past.

src/components/i18n/Composition.js
```diff
@@ -0,0 +1,131 @@
+/*
+ * Utility class for the FormattedCompMessage component.
+ */
+import React from 'react';
+import * as MA from 'message-accumulator';
+import * as N from 'ilib-tree-node';
+import { JSTYPE_BOOLEAN, JSTYPE_NUMBER, JSTYPE_OBJECT, JSTYPE_STRING } from '../../constants';
+
+const MessageAccumulator = MA.default;
```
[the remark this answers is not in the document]
ehoogerbeets: message-accumulator and ilib-tree-node are both ES5 CommonJS modules, so they can't be imported normally. I can perhaps use "require" instead of import to avoid this...

src/components/i18n/Param.js
ehoogerbeets: getValue is there for unit testing so we don't have to call render

jstoffan: The inclusion of the extra method requires a class, which otherwise isn't necessary. You can use enzyme's `render` helper to test the output of the component's render method.

src/components/i18n/FormattedCompMessage.js
ehoogerbeets: This documents the resource id of the string in mojito. This can be used for a number of purposes. It is one of the most important parts of this whole component from my point-of-view as a moji guy!

When the strings are tagged with their id like that, we can make a browser plugin that will connect strings in the UI directly to mojito. (Right-click > View This String in Mojito). This allows us to do in-context linguistic review of the translations, as well as updating the strings on the fly. In fact the linguistic reviewers can submit the translation fixes directly into mojito and then preview their proposed fix in the browser on the fly! (Queue the singing angels!)

It also allows us to figure out where a string is used in the UI so that when we take screen shots for the translators, we can connect those screen shots to particular strings in mojito.

It is also very useful for debugging translation problems, as it tells us where each UI string comes from. (Which Box product, which repo, which string, etc.) Many times, especially for short strings, there are multiple places in the code where the exact same string is used. Which one do we fix? Probably not all of them... context is everything! Currently, we have to spend a lot of time sleuthing around in multiple git repos to find the exact string, and guess which one we need to update in mojito. It gets even worse when a jira ticket comes in with a screen shot in a different language telling us to fix the string to something else. How do you even type that text in the search box in mojito? (Do you know how to type that string in the screen shot in Russian, Chinese, or Japanese? I know I don't!) If the id is right there in the code, then it is easy to find and update the string in mojito.

Later, the same id can be used to implement the text A/B testing framework that the personalization/growth team is interested in building. The id will allow a product manager to right-click on any text in the UI and a browser plugin will find that id and throw up a dialog box that allows the PM to enter test variations for the A/B test, pick some criteria, and then launch the test. (Again using the right browser plugin.)

For everyone else, the spans are not seen in the UI and resource ids are not visible.

src/components/i18n/README.md
```diff
@@ -0,0 +1,296 @@
...
+Note that whitespace in the middle of the string is compressed because it doesn't matter to the translation or to the final
+HTML output.
+
+** Avoid This **
+
+Do not put brace expressions in the text to translate! They will get replaced before the FormattedCompMessage component
+is created, and therefore the source string will contain the value of the expression. When the component attempts to
+load the translation, that source string will not be found, and therefore the translation will also not be found.
+Strings with brace expressions in them will be rejected by the babel plugin when they are extracted.
+
```
jeeyi: What is the error message returned to the user by babel plugin when they have curly braces instead of using `Param`? It should be descriptive enough so that they can fix it quickly.

package.json
jstoffan: Seems like we're at a bit of an impasse with the dependencies. What are our next steps here, @priyajeet?

src/components/i18n/FormattedCompMessage.js
jstoffan: This is great context. I didn't realize we had a separate plugin to assist with translations. Can you add a shorter version of this information to the source file? That way anyone who's perusing this component in the future will be able to see its purpose immediately.

src/components/i18n/Plural.js
```diff
@@ -0,0 +1,25 @@
+// @flow
+import * as React from 'react';
+
+type Props = {
+    /** The plural category this string. (required) */
+    category: 'zero' | 'one' | 'two' | 'few' | 'many' | 'other',
```
jstoffan: Seems like this should still use the `CATEGORY_ONE` constant types to avoid duplicating them here.

src/components/i18n/Plural.js
```diff
@@ -0,0 +1,25 @@
...
+ * The categories for English are "one" for singular and "other" for plural. Both are
+ * required when writing plurals with the FormattedCompMessage component in source code.<p>
+ *
+ * This component does not add any functionality beyond its contents, and its only
+ * purpose is to enclose some JSX to use for a particular plural category.
+ *
+ * See the [Unicode CLDR description of plural category
+ * rules](http://cldr.unicode.org/index/cldr-spec/plural-rules) for more details.
+ */
+export default function Plural(props: Props) {
```
jstoffan: Nearly all of our existing components follow the format below. It would be great if these could as well for consistency's sake:

```
const Plural = ({ children }) => children;

export default Plural;
```

src/components/i18n/Plural.js
ehoogerbeets: Flow does not allow that.

src/components/i18n/FormattedCompMessage.js
```diff
@@ -0,0 +1,187 @@
...
+                );
+            }
+        }
+        // add these to the string in a particular order so that
+        // we always end up with the same string regardless of
+        // the order that the Plural elements were specified in
+        // the source code
+        const categoriesString = [
+            CATEGORY_ZERO,
+            CATEGORY_ONE,
```
jeeyi: How do we handle `=1`? We have some instances where we specifically check `count==1`.

ehoogerbeets: Ooo, good one! I'll update the code for that.

src/components/i18n/Plural.js
ehoogerbeets: Sure, I'll update the code to look like that.

src/components/i18n/README.md
ehoogerbeets: I'll make it more descriptive.

yarn.lock
ehoogerbeets: These are in SLC review right now.

src/components/i18n/Plural.js
ehoogerbeets: Changed so that we test the Plural component output directly instead of having this method for use in unit testing.

src/components/i18n/README.md
ehoogerbeets: Okay, the babel plugin is updated to mention how to fix it. (ie. use the Param tag instead.)

yarn.lock
ehoogerbeets: They have passed SLC review now.

src/components/i18n/Plural.js
ehoogerbeets: It is updated like this now.

src/components/i18n/FormattedCompMessage.js
```diff
@@ -0,0 +1,187 @@
+// @flow
+import * as React from 'react';
+import { injectIntl } from 'react-intl';
+import _ from 'lodash';
```
jstoffan: Please avoid importing the root lodash module, as much of it is unused and will increase our bundle size. You should be able to do `import isNaN from 'lodash/isnan';`

src/components/i18n/FormattedCompMessage.js
jstoffan: I still think having a brief in-code comment to explain the purpose of this attribute would be prudent.

src/components/i18n/FormattedCompMessage.js
```diff
@@ -0,0 +1,187 @@
...
+        );
+
+        // always wrap the translated string in a tag to contain everything
+        // and to give us a spot to record the id
+        return React.createElement(
+            tagName,
+            {
+                key: id,
+                'x-resource-id': id,
+            },
```
jstoffan: Would it be worthwhile to spread any other props onto the inner span? That way we could apply a className to it directly, for example.

src/components/i18n/Param.js
```diff
@@ -0,0 +1,62 @@
...
+ * the self-closing syntax.
+ *
+ * @example
+ * <pre>
+ *   <FormattedCompMessage id="x" description="y">
+ *     The file <Param value={filelist[i].path} description="Name of the file that was deleted."/> has been deleted.
+ *   </FormattedCompMessage>
+ * </pre>
+ */
+export default function Param(props: Props) {
```
jstoffan: Please update to a named constant, such as `const Param = ({ value }: Props) => ( // etc`

src/utils/env.js
```diff
@@ -0,0 +1,15 @@
...
+ * @author Box
+ */
+
+/**
+ * Return true if we are currently running in a test or development
+ * environment.
+ *
+ * @return {boolean} true if we are running in a test or dev environment
+ */
+export default function isDevEnvironment() {
```
jstoffan: The generic name of the util implies that there may be other items exported from it in the future. Should this be a named export rather than a default export?

src/utils/env.js
```diff
@@ -0,0 +1,15 @@
...
+    return process.env.NODE_ENV === 'test' || process.env.NODE_ENV === 'dev';
```
jstoffan: What happens if `process` is undefined, such as in a browser? Have the components been tested in a browser?

ehoogerbeets: https://webpack.js.org/guides/production/#specify-the-mode

I have tested it in a browser, but in a mini React app I created in an ad-hoc, only-on-my-machine sort of way. I should think about doing it more formally perhaps.

Also, it runs in the browser in the style guide.

src/utils/env.js
ehoogerbeets: If I make it a named export, I get this:

```
✖ eslint --fix found some errors. Please fix them and try committing again.

/Users/ehoogerbeets/src/box-ui-elements/src/utils/env.js
13:1  error  Prefer default export  import/prefer-default-export
```

src/utils/env.js
jstoffan: We have a babel-only build as well that outputs to the `/es` folder. It seems like this could break for customers who aren't using webpack as their build system.

src/utils/env.js
jstoffan: Export an object that contains the method:

```
export default {
    isDevEnvironment,
};
```

src/utils/env.js
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+/**
+ * Return true if we are currently running in a test or development
+ * environment.
+ *
+ * @return {boolean} true if we are running in a test or dev environment
+ */
+export default function isDevEnvironment() {
+    return (process && process.env.NODE_ENV === 'test') || process.env.NODE_ENV === 'dev';
```
jstoffan: It looks like the second clause could still throw undefined, since it's checking `process.env` without checking `process`.

src/utils/env.js
jstoffan: Thoughts on this comment, @ehoogerbeets?

src/components/i18n/FormattedCompMessage.js
```diff
@@ -0,0 +1,197 @@
...
+                throw new Error(
+                    'Cannot use count prop on a FormattedCompMessage component without giving both a "one" and "other" Plural component in the children.',
+                );
+            }
+        }
+        // add these to the string in a particular order so that
+        // we always end up with the same string regardless of
+        // the order that the Plural elements were specified in
+        // the source code
+        const categoriesString = [
```
jstoffan: Can this live as its own constant in this file rather than being re-instantiated each time?

src/components/i18n/__tests__/Composition-test.js
```diff
@@ -0,0 +1,270 @@
+import React from 'react';
+
+import Composition from '../Composition';
+
+describe('components/i18n/Composition', () => {
+    test('ComposeString', () => {
```
jstoffan: Ideally, test descriptors should explain the expectation rather than the subject under test. The latter is more appropriate for `describe`.

src/components/i18n/__tests__/FormattedCompMessage-test.js
```diff
@@ -0,0 +1,451 @@
...
+            );
+
+            const span = wrapper.find('span');
+            expect(span.prop('x-resource-id')).toEqual('test');
+
+            const a = wrapper.find('b');
+            expect(a.text()).toContain('bold!');
+        });
+
+        /*
```
jstoffan: Can we avoid checking in this commented code?

PR #980, opened 2019-03-19
chore(query-bar): pass conditions into styleguide

ChenCodes:
Changes in this PR:
- Pass conditions into styleguide for Query Bar
- Default size to be empty string for Avatar.js
- Resolves two flow errors currently being shown in the console

### DIFF
--- a/src/components/avatar/Avatar.js
+++ b/src/components/avatar/Avatar.js
@@ -55,7 +55,7 @@ class Avatar extends React.PureComponent<Props, State> {
     };
 
     render() {
-        const { avatarUrl, className, name, id, size }: Props = this.props;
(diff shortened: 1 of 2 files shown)


boxcla:

src/components/avatar/Avatar.js
```diff
@@ -55,7 +55,7 @@ class Avatar extends React.PureComponent<Props, State> {
     };
 
     render() {
-        const { avatarUrl, className, name, id, size }: Props = this.props;
+        const { avatarUrl, className, name, id, size = '' }: Props = this.props;
```
ChenCodes: I included this change as it helps clean up a flow error that got through the flow check, didn't think we would need a separate ticket + new pr just for a one liner change

PR #982, opened 2019-03-19
fix(tasks): remove temporary headers

alexkrolick:
The API has been updated to use the standard
application/json type for Accept and Content-Type


boxcla:

PR #984, opened 2019-03-19
chore(content-sidebar): Cypress tests for additional tabs

jeremypress:


boxcla:

test/integration/content-sidebar/AdditionalTabs.int.test.js
```diff
@@ -0,0 +1,108 @@
...
+        helpers.load({ hasAdditionalTabs: true, additionalTabs: tabData });
+
+        getTabs()
+            .eq(0)
+            .click()
+            .then(() => expect(stub).to.be.calledWith({ id: 200, callbackData: { status: 'ADDED' } }));
+    });
+
+    it('should properly handle a failed icon load', () => {
+        tabData[1].iconUrl = 'foobar.jpg';
```
jstoffan: I just noticed we're mutating the test data from test to test. This could cause issues with tests relying on each other.

src/elements/content-sidebar/README.md
```diff
@@ -18,6 +18,7 @@ var ContentSidebar = require('./ContentSidebar').default;
         hasMetadata
         hasSkills
         token={TOKEN}
+        {...window.ADDITIONAL_PROPS}
```
jstoffan: I don't think `window` is necessary here.

jeremypress: I was running into issues with other tests that don't define ADDITIONAL_PROPS, it's not defined and throws an error when used like this.

test/integration/content-sidebar/AdditionalTabs.int.test.js
jeremypress: will address, thanks.

PR #986, opened 2019-03-20
fix(unified-share-modal): reduce default text box height

nickgriffin1:
Reverting back to original sizes because the extra height is no longer necessary. This will also help USM to better support smaller screen sizes. 

image

image


boxcla:

PR #987, opened 2019-03-20
build(flow): update flow + flow type defs

alexkrolick:


boxcla:

PR #988, opened 2019-03-20
chore(appactivity): Put AppActivityAPI behind feature flip

JustinHoldstock:


boxcla:

src/api/Feed.js
```diff
@@ -175,7 +177,7 @@ class Feed extends Base {
         const versionsPromise = this.fetchVersions();
         const commentsPromise = this.fetchComments(permissions);
         const tasksPromise = shouldShowNewTasks ? this.fetchTasksNew() : this.fetchTasks();
-        const appActivityPromise = this.fetchAppActivity(permissions);
+        const appActivityPromise = shouldShowAppActivity ? this.fetchAppActivity(permissions) : Promise.resolve();
```
jeremypress: Is the empty state different from the 404 state?

JustinHoldstock: It should be the same, however the sort method prefers nothing compared to the Object literal

PR #991, opened 2019-03-21
fix(tasks): fix flow check

alexkrolick:


boxcla:

src/elements/content-sidebar/activity-feed/task-new/Task.js
```diff
@@ -143,7 +143,7 @@ class Task extends React.Component<Props> {
                             </span>
                         ) : null}
                     </div>
-                    {shouldShowActions ? (
+                    {currentUserAssignment && shouldShowActions ? (
```
alexkrolick: extracting this result makes flow think the currentUserAssignment could be null

PR #992, opened 2019-03-21
build(ci): remove unrelated cleanup tasks from ci checks

alexkrolick:
- I don't know if this is important, but it looked like some extra tasks are running before checks that don't need them
- Adding a flow error to confirm CI reports


boxcla:

jstoffan:
No longer needed.

PR #994, opened 2019-03-21
perf(picker): Fetch Shared Link info on select

ConradJChan:
This removes the shared link fields from the initial folder request for content picker. If `canSetShareAccess` prop is true, when an item is selected, then an API call to fetch the missing shared link fields will be invoked. Also if no shared link currently exists for the item, then one will be created with enterprise defaults.

Content Picker

- [ ] Unit tests


boxcla:

src/elements/content-picker/ContentPicker.js
```diff
@@ -820,6 +827,55 @@ class ContentPicker extends Component<Props, State> {
         });
     };
 
+    /**
+     * Show the inline shared link dropdown
+     * @param {BoxItem} item - The item (folder, file, weblink)
+     * @returns {Promise}
+     */
+    showSharedLinkDropdown = (item: BoxItem): void => {
```
jeremypress: rename? This method seems to fetch info but doesn't actually show anything.

src/elements/content-picker/shareAccessCellRenderer.js
```diff
@@ -15,11 +16,16 @@ export default (
...
     hasHitSelectionLimit: boolean,
 ) => ({ rowData }: { rowData: BoxItem }) => {
-    if (!isRowSelectable(selectableType, extensionsWhitelist, hasHitSelectionLimit, rowData)) {
+    if (!isRowSelectable(selectableType, extensionsWhitelist, hasHitSelectionLimit, rowData) || !rowData.selected) {
         return <span />;
     }
 
-    return (
+    const { allowed_shared_link_access_levels } = rowData;
+    const isLoading = !allowed_shared_link_access_levels;
```
jeremypress: Minor, but do we need a failed state here now that it's a separate call?

src/elements/content-picker/shareAccessCellRenderer.js
```diff
@@ -15,11 +16,16 @@ export default (
...
+
+    return isLoading ? (
+        <LoadingIndicator className="bcp-ShareAccessSelect--loading" />
```
jstoffan: Modifiers should be used together with their base class, like `className="bcp-ShareAccessSelect bcp-ShareAccessSelect--loading"`. Also, since this isn't actually an instance of the ShareAccessSelect component, the class should probably be renamed.

src/elements/content-picker/ContentPicker.js
```diff
@@ -820,6 +827,55 @@ class ContentPicker extends Component<Props, State> {
         });
     };
 
+    /**
+     * Show the inline shared link dropdown
+     * @param {BoxItem} item - The item (folder, file, weblink)
+     * @returns {Promise}
```
jstoffan: Looks like the JSDoc and flowtypes don't match for the return value.

src/elements/content-picker/ContentPicker.js
```diff
@@ -759,8 +761,8 @@ class ContentPicker extends Component<Props, State> {
      * @param {Object} item file or folder object
      * @return {void}
      */
-    select = (item: BoxItem): void => {
-        const { type: selectableType, maxSelectable }: Props = this.props;
+    select = (item: BoxItem, forceSharedLink: boolean = true): void => {
```
jstoffan: Can we make the latter argument an options object so we can use named parameters? Otherwise it's difficult to tell what passing `false` does here by looking at only the calling code.

src/elements/content-picker/ContentPicker.js
```diff
@@ -795,7 +797,7 @@ class ContentPicker extends Component<Props, State> {
 
             // Check if we hit the selection limit
             // Ignore when in single file selection mode.
-            if (hasHitSelectionLimit && !isSingleFileSelection) {
+            if (hasHitSelectionLimit && !isSingleFileSelection && !selectedKeys.indexOf(cacheKey)) {
```
jstoffan: This new condition seems unrelated to the selection limit. Can we update the comment or break it out into a variable so it's clear what we want to test?

src/elements/content-picker/ContentPicker.js
```diff
@@ -820,6 +827,55 @@ class ContentPicker extends Component<Props, State> {
...
+     * Show the inline shared link dropdown
+     * @param {BoxItem} item - The item (folder, file, weblink)
+     * @returns {Promise}
+     */
+    showSharedLinkDropdown = (item: BoxItem): void => {
+        const { id, type }: BoxItem = item;
+
+        switch (type) {
+            case TYPE_FOLDER:
+                this.api.getFolderAPI().getFolderFields(id, this.handleSharedLinkResponse, () => {}, {
```
jstoffan: Is swallowing the error response desirable? If so, can we use `lodash/noop`, instead?

src/elements/content-picker/ContentPicker.js
```diff
@@ -820,6 +827,55 @@ class ContentPicker extends Component<Props, State> {
...
+    };
+
+    /**
+     * Handles the shared link info by either creating a share link using enterprise defaults if
+     * it does not already exist, otherwise update the item in the state currentCollection.
+     *
+     * @param {Object} item file or folder
+     * @returns {void}
+     */
+    handleSharedLinkResponse = (item: BoxItem) => {
```
jstoffan: `handleSharedLinkSuccess`?

src/elements/content-picker/ContentPicker.js
```diff
@@ -820,6 +827,55 @@ class ContentPicker extends Component<Props, State> {
...
+        // if no shared link currently exists, create a shared link with enterprise default
+        if (!item[FIELD_SHARED_LINK]) {
+            this.changeShareAccess(null, item);
+        } else {
+            const { selected } = this.state;
+            const { id, type } = item;
+            const cacheKey = this.api.getAPI(type).getCacheKey(id);
+            // if shared link already exists, update the collection in state
+            this.updateItemInCollection(item);
+            if (item.selected && item !== selected[cacheKey]) {
```
jstoffan: Does this handle the case where the user selects a new item before the shared link response returns? May want to add a comment.

src/elements/content-picker/ContentPicker.js
```diff
@@ -845,13 +901,31 @@ class ContentPicker extends Component<Props, State> {
...
 
+    /**
+     * Updates the BoxItem in the state's currentCollection
+     *
+     * @param {Object} item file or folder object
+     * @returns {void}
+     */
+    updateItemInCollection = (item: BoxItem) => {
+        const { currentCollection } = this.state;
+        const { items = [] } = currentCollection;
```
jstoffan: Can `items` be undefined here?

src/api/Folder.js
```diff
@@ -160,7 +161,7 @@ class Folder extends Item {
             new FileAPI(this.options),
             new WebLinkAPI(this.options),
         );
-        this.itemCache = (this.itemCache || []).concat(flattened);
+        this.itemCache = union(this.itemCache, flattened);
```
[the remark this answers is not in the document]
ConradJChan: Yeah I can most likely revert this now, i think it was only an issue when I was previously fetching the folder again for the shared link info and it would duplicate the entries

src/elements/content-picker/shareAccessCellRenderer.js
ConradJChan: Makes sense, how's `bcp-share-access-loading`?

src/elements/content-picker/ContentPicker.js
ConradJChan: I'll do both thanks

src/elements/content-picker/ContentPicker.js
ConradJChan: Yeah you're right, if the user clicks another item before the XHR completes it gets cancelled. I'll add a comment about that

src/elements/content-picker/ContentPicker.js
ConradJChan: It's unlikely that `items` could be undefined here, do you see a potential issue?

test/integration/content-picker/ContentPicker.e2e.test.js
```diff
@@ -11,51 +13,89 @@ const helpers = {
...
+        );
+
+        ['319004423111', '308566419514', '308409990441'].forEach(fileId => {
+            cy.route('GET', `**/files/${fileId}?fields=allowed_shared_link_access_levels,shared_link`, {
+                ...getSharedLinkRsp,
+                id: fileId,
+            });
+
+            cy.route('PUT', `**/files/${fileId}`, {
+                ...createSharedLinkRsp,
```
jstoffan: I think we want to avoid explicit imports when possible. Can we import these via `cy.fixture`, instead?

test/integration/content-picker/ContentPicker.e2e.test.js
```diff
@@ -11,51 +13,89 @@ const helpers = {
...
+    },
 };
 
 describe('ContentPicker', () => {
+    beforeEach(() => {
+        cy.server();
+
+        cy.route(
+            'GET',
+            '**/folders/51964781421?fields=id,name,type,size,parent,extension,permissions,path_collection,modified_at,created_at,modified_by,has_collaborations,is_externally_owned,item_collection&direction=asc&limit=50&offset=0&sort=name',
```
jstoffan: Can we avoid hardcoding the file/folder ids? Otherwise it seems like the tests can be pretty brittle if the fixture data changes.

test/integration/content-picker/ContentPicker.e2e.test.js
```diff
@@ -11,51 +13,89 @@ const helpers = {
...
+    },
+    unselectRow(rowNum) {
+        this.getRow(rowNum)
+            .as('row')
             .click()
             .find('input[type="checkbox"]')
-            .should('not.be.checked'),
+            .should('not.be.checked');
+
+        cy.get('@row')
```
jstoffan: Should this helper also return `cy.get('@row')`?

src/api/Folder.js
```diff
@@ -174,33 +174,85 @@ class Folder extends Item {
...
+     * @param {Array<String>} fields Array of field strings
+     * @return {Promise}
      */
-    folderRequest(): Promise<void> {
+    folderRequest(options: FetchOptions = {}, successHandler?: Function = this.folderSuccessHandler): Promise<any> {
         if (this.isDestroyed()) {
             return Promise.reject();
         }
 
+        const { fields, noPagination } = options;
```
jstoffan: These could be destructured in the signature.

src/elements/content-picker/ContentPicker.js
```diff
@@ -757,10 +759,11 @@ class ContentPicker extends Component<Props, State> {
      *
      * @private
      * @param {Object} item file or folder object
+     * @param {boolean} options.forceSharedLink Force a shared link if no link exists
      * @return {void}
      */
-    select = (item: BoxItem): void => {
-        const { type: selectableType, maxSelectable }: Props = this.props;
+    select = (item: BoxItem, options: Object = {}): void => {
```
jstoffan: Can we use something other than the `Object` type for `options`?

src/elements/content-picker/ContentPicker.js
```diff
@@ -779,6 +782,10 @@ class ContentPicker extends Component<Props, State> {
         const cacheKey: string = this.api.getAPI(type).getCacheKey(id);
         const existing: BoxItem = selected[cacheKey];
         const existingFromCache: BoxItem = this.api.getCache().get(cacheKey);
+        const { forceSharedLink = true } = options;
```
jstoffan: These can be destructured in the function signature.

src/elements/content-picker/ContentPicker.js
```diff
@@ -779,6 +782,10 @@ class ContentPicker extends Component<Props, State> {
...
+        const existInSelected = selectedKeys.indexOf(cacheKey) !== -1;
+        const { permissions = {} }: BoxItem = existingFromCache;
```
jstoffan: Should this permissions object come from the passed-in `item`, instead?

src/elements/content-picker/ContentPicker.js
jstoffan: Consider adding a comment.

src/elements/content-picker/shareAccessCellRenderer.js
```diff
@@ -15,11 +16,24 @@ export default (
     extensionsWhitelist: string[],
     hasHitSelectionLimit: boolean,
 ) => ({ rowData }: { rowData: BoxItem }) => {
-    if (!isRowSelectable(selectableType, extensionsWhitelist, hasHitSelectionLimit, rowData)) {
+    const { permissions = {} }: BoxItem = rowData;
+    const { can_set_share_access: rowCanSetShareAccess }: BoxItemPermission = permissions;
```
jstoffan: Should this be `itemCanSetShareAccess` to match the similar usage in ContentPicker? You could also consider using `lodash/get` to cut down on the multiple destructures/default values in both spots: 

```
const itemCanSetShareAccess = getProp(rowData, 'permissions.can_set_share_access', false);
```

src/elements/content-picker/shareAccessCellRenderer.js
jstoffan: @ConradJChan, thoughts on @jeremypress's comment above?

jstoffan:
May also want to consider using the `perf` type for the commit message, as well.

src/elements/content-picker/shareAccessCellRenderer.js
ConradJChan: Yeah I checked with Zach and Rohit and since it seems like an uncommon error, just leaving it as is for now

PR #1000, opened 2019-03-25
fix(versions): Add accessible text to version history item badge

jstoffan:
Quick fix for accessibility of version number information for screen readers.


boxcla:

PR #1004, opened 2019-03-26
feat(versions): Add version history item actions dropdown

jstoffan:
Version Actions

### DIFF
--- a/src/elements/content-sidebar/versions/VersionsItemActions.js
+++ b/src/elements/content-sidebar/versions/VersionsItemActions.js
@@ -1,0 +1,110 @@
+/**
+ * @flow
+ * @file Versions Item Action component
+ * @author Box
+ */
+
+import React from 'react';
+import { FormattedMessage } from 'react-intl';
+import DropdownMenu from '../../../components/dropdown-menu';
+import IconClockPast from '../../../icons/general/IconClockPast';
+import IconDownload from '../../../icons/general/IconDownload';
+import IconEllipsis from '../../../icons/general/IconEllipsis';
+import IconOpenWith from '../../../icons/general/IconOpenWith';
+import IconTrash from '../../../icons/general/IconTrash';
+import IconUpload from '../../../icons/general/IconUpload';
+import messages from './messages';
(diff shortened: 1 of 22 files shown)


boxcla:

src/elements/content-sidebar/versions/VersionsItemActions.scss
```diff
@@ -0,0 +1,37 @@
...
+        background-color: $off-white;
+        fill: $fours;
+    }
+
+    .fill-color {
+        fill: $nines;
+    }
+
+    .icon-clock-past {
+        margin-left: -2px;
```
ConradJChan: What is this margin adjustment used for?

src/elements/content-sidebar/versions/VersionsItemActions.js
```diff
@@ -0,0 +1,110 @@
...
+                )}
+                {showPromote && (
+                    <MenuItem className="bcs-VersionsItemActions-item" onClick={onPromote}>
+                        <IconUpload {...ICON_SIZE} />
+                        <FormattedMessage {...messages.versionActionPromote} />
+                    </MenuItem>
+                )}
+                {showRestore && (
+                    <MenuItem className="bcs-VersionsItemActions-item" onClick={onRestore}>
+                        <IconClockPast height={14} width={14} />
```
ConradJChan: Does this icon need to be tweaked to be consistent with the others? (Not of the scope of this PR but just wondering)

src/elements/content-sidebar/versions/VersionsItemActions.scss
jstoffan: The clock icon has a slightly different size and viewbox than the other icons, so it needed a little extra love.

src/elements/content-sidebar/versions/VersionsItemActions.js
jstoffan: Ideally, all our icons would have the same default dimensions, padding, and viewbox, but that's not the case currently. I know @tjgupta has put some effort into this area, but my understanding is that it's a pretty large effort from both us and design.

PR #1007, opened 2019-03-26
fix(travis): use PROJECT_ROOT. Remove build entry from ignore

JustinHoldstock:
Fixes Flow issue where everything in Travis build path was being ignored


boxcla:

.travis.yml
```diff
@@ -16,6 +16,8 @@ jobs:
       script: commitlint-travis
     - name: "Code Lint"
       script: npm-run-all clean build:i18n lint
+    - name: "Flow Debug"
```
ConradJChan: Still needed?

jstoffan: Looks like it can be removed. However, if we want to keep this around, can we please fold it into the `Flow Types` job? Something like a console output saying `Running Flow on X files`, with X being `yarn flow ls | wc -l`.

JustinHoldstock: Yup. As requested by @alexkrolick

jstoffan: We can keep it, but it should be folded into the existing Flow Types task. It doesn't need to be its own task.

PR #1014, opened 2019-03-27
fix(list-view): Fixed scrollbars that were doubled up

macrotim:
Mar-27-2019 16-39-06


boxcla:

PR #1015, opened 2019-03-28
Add props for title and divider in PillSelectorDropdown

nickgriffin1:
image


boxcla:

i18n/en-US.properties
```diff
@@ -316,7 +316,7 @@ be.sidebarVersions.empty = No prior versions are available for this file.
 be.sidebarVersions.preview = Preview
 # Label for the version promote action.
 be.sidebarVersions.promote = Make Current
-# Label for the version delete action.
+# Label for the version restore action.
```
nickgriffin1: @jstoffan Let me know if this looks alright to you. Seems like a commit might have just missed a translations build.

src/components/pill-selector-dropdown/README.md
```diff
@@ -95,107 +95,3 @@ const validator = text => {
     ))}
 </PillSelectorDropdown>
 ```
-
```
nickgriffin1: I originally wrote this to be temporary for testing. Now I'm removing it.

src/components/selector-dropdown/__tests__/SelectorDropdown-test.js
```diff
@@ -423,26 +454,6 @@ describe('components/selector-dropdown/SelectorDropdown', () => {
         });
     });
 
-    describe('onItemClick', () => {
```
nickgriffin1: This component doesn't have an `onItemClick()` function so I'm deleting this unit test.

src/components/selector-dropdown/__tests__/SelectorDropdown-test.js
```diff
@@ -550,56 +561,11 @@ describe('components/selector-dropdown/SelectorDropdown', () => {
...
-                .expects('scrollItemIntoView')
-                .once()
-                .withExactArgs(id);
+        test('should call scrollIntoView', () => {
+            const scrollIntoView = jest.spyOn(domUtils, 'scrollIntoView');
             instance.setActiveItemID(id);
-        });
-    });
 
-    describe('scrollItemIntoView()', () => {
```
nickgriffin1: The component also does not have this function. It imports `scrollIntoView` which seems to cover this functionality.

ConradJChan:
Would it make sense to generalize this to accept multiple titles and dividers within the dropdown? Maybe instead of accepting the title & divider as props, accept title and divider components as children?

nickgriffin1:
@ConradJChan The idea is this would be used to callout certain items, but I wanted to keep it more generic to cover more possible use cases. The PillSelectorDropdown component expects a list of items and attaches click handlers to each item in the list, which is why I implemented this as a discrete feature. We don't want click handler attached to titles and horizontal rules. Since PillSelectorDropdown expects an array as children, we would need to modify the data structure of the children or tag the children with something to indicate this. However, I think that is a more complicated implementation and probably would be a breaking change for one of the most commonly used components.

src/components/selector-dropdown/SelectorDropdown.js
```diff
@@ -241,9 +246,20 @@ class SelectorDropdown extends React.Component<Props, State> {
                         },
                         setActiveItemID,
                     };
+
                     if (index === activeItemIndex) {
                         itemProps.isActive = true;
                     }
+
+                    if (index === dividerIndex) {
```
ConradJChan: Since these two return statements are pretty similar, would it be cleaner to do something like the following to have a single return?
```
const hasDivider = index === dividerIndex;

return (
    <React.Fragment>
        hasDivider && <hr className="SelectorDropdown-divider" />
        {React.cloneElement(item, itemProps)}
    </React.Fragment>
);
```

src/components/selector-dropdown/SelectorDropdown.scss
```diff
@@ -4,14 +4,34 @@
...
     .overlay {
         display: block;
         width: 100%;
+
+        .SelectorDropdown-divider {
+            border: 0;
+            border-bottom: 1px solid $bfive;
+            display: block;
+            height: 1px;
+            margin: 5px 5%;
```
ConradJChan: Does the left-right margin need to be a percentage? Could we use the 15px like in the title?

src/components/selector-dropdown/SelectorDropdown.js
nickgriffin1: I agree that is cleaner. I'll update this.

src/components/selector-dropdown/SelectorDropdown.scss
nickgriffin1: It doesn't need to be a percentage. I'll update this.

PR #1017, opened 2019-03-28
fix(query-bar): remove display flex for query bar buttons

ChenCodes:
Changes in this PR:
- Remove display: flex from buttonLabel in QueryBar due to the overflow: ellipsis not working
Screen Shot 2019-03-28 at 1 47 19 PM


boxcla:

PR #1018, opened 2019-03-28
fix(jest): Add env setup for jests tests. Use common TZ

JustinHoldstock:
This'll allow us to run tests from any timezone!


boxcla:

PR #1019, opened 2019-03-28
fix(query-bar): add case for no templates in enterprise

ChenCodes:
Changes in this PR:
- Handle case for when templates have been loaded but is an empty array
- Add new message to TemplateButton for when there are no templates in the enterprise
Screen Shot 2019-03-28 at 2 22 35 PM


boxcla:

ChenCodes:
Adding a unit test to check if the new case is being hit

src/features/query-bar/__tests__/TemplateButton-test.js
```diff
@@ -81,5 +81,16 @@ describe('feature/query-bar/components/TemplateButton', () => {
             const entryButtonWrapper = shallow(wrapper.instance().renderEntryButton());
             expect(entryButtonWrapper.props().className.includes(activeTemplateClassName)).toEqual(expectedReturn);
         });
+
+        test.each`
+            templates | expectedMessage             | description
+            ${null}   | ${'Template Name'}          | ${'Should render templates loading message'}
```
macrotim: Is this accurate? -> "Should render templates loading message"

ChenCodes: So the text on the button while the templates are loading is "Template Name" which is technically the templates loading message.

src/features/query-bar/components/TemplateButton.js
```diff
@@ -52,9 +52,14 @@ class TemplateButton extends React.Component<Props, State> {
         let icon;
         let text;
 
-        if (!templates) {
+        const isLoadingTemplates = !templates;
+        const noTemplatesInEnterprise = templates && templates.length === 0;
```
macrotim: I'd prefer a different name. 

Suggestions: "has templates", "count templates" combined with if(countTemplates>0).

src/features/query-bar/__tests__/TemplateButton-test.js
macrotim: Does it mean loading message? I'm not sure about that. If I see the word "loading message", I expect a spinner or similar.

My understanding is that "Template Name" is a placeholder value. Here's the definition of HTML "placeholder": The placeholder attribute specifies a short hint that describes the expected value of an input field (e.g. a sample value or a short description of the expected format).

src/features/query-bar/components/TemplateButton.js
```diff
@@ -52,9 +52,14 @@ class TemplateButton extends React.Component<Props, State> {
         let icon;
         let text;
 
-        if (!templates) {
+        const isLoadingTemplates = !templates;
+        const hasTemplates = templates && templates.length === 0;
```
macrotim: It's odd that "hasTemplates" is true when length === 0.

ChenCodes: Ah this is a typo, should be `hasNoTemplates`

src/features/query-bar/components/TemplateButton.js
```diff
@@ -53,12 +53,12 @@ class TemplateButton extends React.Component<Props, State> {
         let text;
 
         const isLoadingTemplates = !templates;
-        const hasTemplates = templates && templates.length === 0;
+        const hasNoTemplates = templates && templates.length === 0;
```
macrotim: Negatives are discouraged.

Please revise to "if (!hasTemplates)" and make appropriate fixes.

PR #1020, opened 2019-03-28
Modifier dropdown gets closed datepicker

ChenCodes:
Changes in this PR:
- Pass portaledClasses so that clicking on the Month/Day button will no longer close the entire Flyout


boxcla:

PR #1021, opened 2019-03-28
docs(css): don't add outer styling to base reset class

alexkrolick:


boxcla:

src/elements/content-preview/README.md
```diff
@@ -23,6 +23,7 @@ var ContentPreview = require('./ContentPreview').default;
         features={FEATURES}
         fileId={FILE_ID}
         token={TOKEN}
+        className="example"
```
jstoffan: Can we avoid adding an additional prop here by targeting each Element's root class name directly?

```
.be {
    &.bce,
    &.bcs,
    &.bcow,
    &.bcp,
    &.bcpr,
    &.bcu {
        box-shadow: 0 0 20px 0 rgba(85, 85, 85, .42);
    }
}

jstoffan: The current change will result in box shadows being lost from all other examples, so a solution similar to the above will be necessary for us to merge.

PR #1022, opened 2019-03-28
Smart Views columns should stretch to fill the container width

macrotim:
Before the fix, the List View displays empty whitespace to the right of the right-most column. 

With the fix...
Screen Shot 2019-03-28 at 2 11 31 PM

### DIFF
--- a/src/features/list-view/ListView.js
+++ b/src/features/list-view/ListView.js
@@ -4,10 +4,10 @@ import classNames from 'classnames';
 import noop from 'lodash/noop';
 
 import { MultiGrid } from 'react-virtualized/dist/commonjs/MultiGrid/index';
-import { FIXED_ROW_COUNT, ROW_HEIGHT } from './constants';
 import { SORT_ORDER_ASCENDING, SORT_ORDER_DESCENDING } from '../query-bar/constants';
 import IconSortChevron from '../../icons/general/IconSortChevron';
 
+import { DEFAULT_COLUMN_WIDTH, FIXED_ROW_COUNT, ROW_HEIGHT } from './constants';
 import type { CellRendererArgs, ComputeColumnWidthArgs } from './flowTypes';
 
 import './styles/ListView.scss';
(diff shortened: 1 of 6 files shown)


boxcla:

src/features/list-view/ListView.js
```diff
@@ -32,6 +32,12 @@ type CellRendererArgs = {|
     style: Object,
 |};
 
+type ComputeColumnWidthArgs = {|
```
ChenCodes: Could we move this and CellRendererArgs to a flowType.js file?

macrotim: Glad you pointed it out! I'll do that.

macrotim: Done.

src/features/list-view/ListView.js
```diff
@@ -25,12 +27,7 @@ type Props = {
     width: number,
 };
 
-type CellRendererArgs = {|
-    columnIndex: number,
-    key: string,
-    rowIndex: number,
-    style: Object,
-|};
+export const DEFAULT_COLUMN_WIDTH = 300;
```
Visylvius: can this be moved to constants?

src/features/list-view/ListView.js
```diff
@@ -67,8 +64,51 @@ class ListView extends React.PureComponent<Props> {
...
+        }
+
+        const firstColumnWidth = getClientDefinedColumnWidth(0);
+        if (columnIndex === 0) {
+            return firstColumnWidth;
+        }
+
+        // Determine if columns fill width of
+        // MultiGrid or stretching is needed.
+        let canFillWidth = false;
```
Visylvius: nitpick: can we move var declarations to the top of the function?

macrotim: Can you help me understand why you prefer that? 

Moving it to the top means any reader must jump their eyeballs up and down the function to make sense of it.

src/features/list-view/ListView.js
macrotim: Done.

src/features/list-view/ListView.js
Visylvius: Thats true, i've always followed the pattern of putting all var declarations at the top of the function. That way they're in one place and you always know where to look.

PR #1023, opened 2019-03-29
chore(accessibility): Increase the contrast of modal close button to 3:1

patrickpaul:
Increases the contrast ratio of the icon to >3:1 which is the WCAG AA standard for non-text UI.

### DIFF
--- a/src/components/modal/ModalDialog.js
+++ b/src/components/modal/ModalDialog.js
@@ -72,7 +72,7 @@ class ModalDialog extends React.Component<Props> {
                 className="modal-close-button"
                 onClick={this.onCloseButtonClick}
             >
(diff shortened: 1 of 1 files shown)


boxcla:

src/components/modal/ModalDialog.js
```diff
@@ -72,7 +72,7 @@ class ModalDialog extends React.Component<Props> {
                 className="modal-close-button"
                 onClick={this.onCloseButtonClick}
             >
-                <IconClose color="#999EA4" height={18} width={18} />
+                <IconClose color="#909090" height={18} width={18} />
```
patrickpaul: Color choice of #909090 has been approved by DG in Box Design.

PR #1027, opened 2019-04-02
fix(tasks): correct task type enum

alexkrolick:
- Fixes the task API to use the new enum format for task_types
- Removes the possibility to "fall through" to general task since the field is non-nullable and the update PUT payload must match the task type

### DIFF
--- a/conf/jest/enzyme-adapter.js
+++ b/conf/jest/enzyme-adapter.js
@@ -13,4 +13,6 @@ global.shallow = shallow;
 global.mount = mount;
 
 // testing utility functions
-global.queryAllByTestId = (wrapper, testid) => wrapper.find(`[data-testid="${testid}"]`).hostNodes();
+
+// accepts a Cheerio jQuery-style wrapper or Enzyme mount wrapper
+global.queryAllByTestId = (wrapper, testid) => wrapper.find(`[data-testid="${testid}"]`);
--- a/conf/jest/enzyme-adapter.js
+++ b/conf/jest/enzyme-adapter.js
@@ -11,3 +11,6 @@ Enzyme.configure({ adapter: new Adapter() });
 // make Enzyme functions available in all test files without importing
 global.shallow = shallow;
 global.mount = mount;
+
+// testing utility functions
+global.queryAllByTestId = (wrapper, testid) => wrapper.find(`[data-testid="${testid}"]`).hostNodes();
--- a/src/elements/content-sidebar/activity-feed/task-new/TaskActions.js
+++ b/src/elements/content-sidebar/activity-feed/task-new/TaskActions.js
@@ -5,21 +5,23 @@ import { FormattedMessage } from 'react-intl';
 import { ACTIVITY_TARGETS } from '../../../common/interactionTargets';
 import messages from '../../../common/messages';
 import Button from '../../../../components/button';
-import { TASK_TYPE_APPROVAL } from '../../../../constants';
+import { TASK_TYPE_APPROVAL, TASK_TYPE_GENERAL } from '../../../../constants';
 
 type Props = {|
     onTaskApproval: Function,
     onTaskComplete: Function,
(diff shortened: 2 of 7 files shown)


boxcla:

conf/jest/enzyme-adapter.js
```diff
@@ -11,3 +11,6 @@ Enzyme.configure({ adapter: new Adapter() });
 // make Enzyme functions available in all test files without importing
 global.shallow = shallow;
 global.mount = mount;
+
+// testing utility functions
+global.queryAllByTestId = (wrapper, testid) => wrapper.find(`[data-testid="${testid}"]`).hostNodes();
```
jstoffan: Is `.hostNodes()` needed or can we return the wrappers themselves? I think we want to avoid interacting directly with DOM nodes when possible.

src/elements/content-sidebar/activity-feed/task-new/TaskActions.js
```diff
@@ -14,35 +14,42 @@ type Props = {|
     taskType?: TaskType,
 |};
 
-const TaskActions = ({ onTaskApproval, onTaskReject, onTaskComplete, taskType }: Props): React.Node => (
-    <div className="bcs-task-pending-assignment bcs-task-assignment-actions">
-        {taskType === TASK_TYPE_APPROVAL ? (
+const TaskActions = ({ onTaskApproval, onTaskReject, onTaskComplete, taskType }: Props): React.Node => {
+    let action = null;
```
jstoffan: Previously, all non-`TASK_TYPE_APPROVAL` cases returned the single-button "complete" action. Do we intend to support an empty task or should we short-circuit and return null if the `taskType` isn't recognized?

conf/jest/enzyme-adapter.js
alexkrolick: I want to count only the actual rendered testids, not everywhere the prop is passed. The Enzyme wrappers would find the react node, any child component that passes props through, and final DOM node.

src/elements/content-sidebar/activity-feed/task-new/TaskActions.js
alexkrolick: It no longer makes sense to have a fallback value since the enums passed to the update endpoints are strictly bound to the task type.

conf/jest/enzyme-adapter.js
jstoffan: Should we be using `render` rather than `mount`, in that case? If the final rendered markup is what we want?

src/elements/content-sidebar/activity-feed/task-new/TaskActions.js
jstoffan: It looks like the `taskType` prop is optional, though. Thoughts about returning null rather than the div wrapper if `taskType` is not provided?

alexkrolick: I'll change the flow type to make the type non nullable

conf/jest/enzyme-adapter.js
alexkrolick: In this case a static render might work... but that doesn't work for anything stateful or that tests interactions, right? I thought this helper would be fairly useful in other types of tests.

jstoffan: Possibly, but I think we should probably add global methods only when their value is very high. Having a helper to find/filter on `data-testid` attribute values seems valuable, but returning `hostNodes` less so. 

It also looks like the `.render` [helper](https://airbnb.io/enzyme/docs/api/ReactWrapper/render.html) may also be useful to accomplish this inline.

alexkrolick: https://github.com/box/box-ui-elements/pull/1027/commits/3c9e928ad94d86f97d86c7af5ced1897dd2f3431

PR #1028, opened 2019-04-02
FilterButton.onFilterChange should return Date as an object

macrotim:
FilterButton.onFilterChange previously returned Date as strings which is harder to consume in the client. Now returns Date objects.

Additional changes that I made:
* Introduced ConditionValueType
* Removed `columns` from fixtures as well as a few others variables and refactored dependencies e.g. unit tests and examples.
* Tried to improve readability and comprehension of a number of unit tests


boxcla:

PR #1030, opened 2019-04-02
fix(details-sidebar): classification for collaborators

priyajeet:
Removes fetching of classification from within the details sidebar and relies on the value passed in.
Classification value was being passed in via the `bannerPolicy` prop which is now renamed to `classification` with a more descriptive object structure.


boxcla:

priyajeet:
@ahakanbaba Yeah, we dont show that currently, even before this PR

PR #1032, opened 2019-04-03
style(preview): Add transition to prev file button

ConradJChan:
Also adds back the opacity transition for the prev/next file buttons


boxcla:

src/elements/content-preview/ContentPreview.scss
```diff
@@ -41,7 +41,7 @@
         position: absolute;
         top: 50%;
         transform: translateY(-50%);
-        transition-property: opacity;
+        transition: opacity .5s, left .3s cubic-bezier(.4, 0, .2, 1);
```
jeremypress: Where does the left value get changed?

jstoffan: The `opacity` entry here needs a timing function. Also, do we need to transition `right`, as well? Should they be the same duration?

ConradJChan: When thumbnails sidebar is opened: https://github.com/box/box-ui-elements/pull/1032/files/444bc2254419aba9b36fb8b4d76fe6208396c674#diff-c8b2e5e96e87b90de7160a27e8aa2805R84

jeremypress: right should be ok because there's nothing that moves it on that side?

jstoffan: Since this is in a compound rule, it's applied to the right arrow as well. Seems like it should live in the left arrow only, then? The duration and timing functions could live in the compound rule, while the properties get split up?

PR #1034, opened 2019-04-03
feat(content-sidebar): Add in beta feedback task modal component

booc0mtaco:
Create reusable component, and use it for the content sidebar. Add snapshot test for component


boxcla:

src/features/beta-feedback/BetaFeedbackBadge.js
```diff
@@ -0,0 +1,55 @@
...
+
+type Props = {
+    className?: string,
+    formUrl: string,
+    tooltip?: boolean,
+};
+
+const BetaFeedbackBadge = ({ className = '', tooltip = false, ...rest }: Props) => {
+    const classes = classNames('bdl-HeaderFeedbackBadge', className);
+    const { formUrl } = rest;
```
jeremypress: Any reason why you're unpacking formUrl from `rest` instead of naming it like the other props?

src/features/beta-feedback/BetaFeedbackBadge.js
```diff
@@ -0,0 +1,55 @@
...
+    return (
+        <span className={classes}>
+            <span id="bdl-HeaderFeedbackBadge-ariaLabel" aria-hidden="true" hidden>
+                <FormattedMessage
+                    id="be.feedbackFormDescription"
+                    defaultMessage="Beta Feedback Form"
+                    description="Accessible text used to describe the form used for feedback"
+                />
+            </span>
+            <Link href={formUrl} target="_blank" aria-labelledBy="bdl-HeaderFeedbackBadge-ariaLabel">
```
jeremypress: Do we want the link to appear if `formUrl` is ''? Or is the url always going to be valid if we have the tasks feature on?

src/features/beta-feedback/BetaFeedbackBadge.js
```diff
@@ -0,0 +1,55 @@
...
+    ) : (
+        <BetaBadge className="bdl-HeaderFeedbackBadge-betaBadge" />
+    );
+
+    // TODO: tooltip may require constrainToScrollParent & constrainToWindow in some contexts
+    return (
+        <span className={classes}>
+            <span id="bdl-HeaderFeedbackBadge-ariaLabel" aria-hidden="true" hidden>
+                <FormattedMessage
+                    id="be.feedbackFormDescription"
```
jeremypress: To avoid writing these out you can `import messages from '../messages';`, and then spread the object into the `FormattedMessage` component. 

`<FormattedMessage {...messages. feedbackFormDescription} />`

alexkrolick: I am starting to lean towards using FormattedMessage inline:
- Because defineMessages returns a new object, there is no way to easily jump to the variable definition for a message, nor to easily find usages of a message. For that reason I think there a lot of dead messages that don't show up in any components any more.
- messages.js is an indirection in the best case

src/features/beta-feedback/BetaFeedbackBadge.js
booc0mtaco: I initially wanted to put just the defaulted params on L18, and unpack the rest from `rest`, but we ended up only having the one const like this. Personally, I find the `({ ...}: Props) => {` a bit hard to read when it gets long or spans multiple lines.

src/features/beta-feedback/BetaFeedbackBadge.js
booc0mtaco: For tasks, the URL will be specified, but that value is set up in the client app's feature config. This guards against it getting `undefined`.

src/features/beta-feedback/BetaFeedbackBadge.js
booc0mtaco: ah yeah, initially there was only one message to format, so wanted to avoid creating a whole module for one string. if we had 3+, I'd convert over to a messages module

booc0mtaco: @alexkrolick we can describe some rules around it, and I think there's a case for both, given a few criteria. For instance, things like:

- well-defined strings that are used across many components, versus component-specific strings
- number of strings defined for a specific component
- likelihood that a message may become obsolete

jstoffan: I don’t see any other examples of inline FormattedMessage components in Elements. Can you move the definition into a messages module to stay consistent with the rest of the project?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -546,6 +546,7 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
             getApproverWithQuery,
             approverSelectorContacts,
             getAvatarUrl,
+            feedbackUrl: getFeatureConfig(features, 'activityFeed.tasks').feedbackUrl || '',
```
jstoffan: This could use `lodash/get` directly, since that's what `getFeatureConfig` uses under the hood: 

```
feedbackUrl: getProp(features, 'activityFeed.tasks.feedbackUrl', '');
```

src/features/beta-feedback/BetaFeedbackBadge.js
alexkrolick: https://github.com/box/box-ui-elements/pull/1037 turns out Flow can work with jump-to-definition if you don't cast the result of `defineMessages`

PR #1036, opened 2019-04-03
feat(unified-share-modal): suggestions in dropdown

nickgriffin1:


src/features/unified-share-modal/ContactsField.js
```diff
@@ -28,15 +28,15 @@ type Props = {
     onContactAdd: Function,
     onContactRemove: Function,
     onInput?: Function,
-    onSuggestedCollaboratorAdd: Contact => void,
     selectedContacts: Array<Contact>,
-    suggestedCollaborators?: Array<Contact>,
+    suggestedCollaborators?: Object,
```
shahzadaziz: Is it possible to define type for this?

src/features/unified-share-modal/ContactsField.js
```diff
@@ -50,16 +50,40 @@ class ContactsField extends React.Component<Props, State> {
 
         this.state = {
             contacts: [],
+            numSuggestedShowing: 0,
             pillSelectorInputValue: '',
         };
     }
 
+    addSuggestedContacts = (contacts: Array<Contact>) => {
+        const { suggestedCollaborators = {} } = this.props;
```
shahzadaziz: Maybe we can use `defaultProps` here instead.

src/features/unified-share-modal/ContactsField.js
nickgriffin1: I can make it more specific. I'll have to check the flow docs. It basically an Object where every key and value is the same format.

src/features/unified-share-modal/ContactsField.js
nickgriffin1: IMO it's better to use this pattern because it keeps the logic closer to the code and gives more context if that context is needed. For example, you might want to know when the results are empty versus when the prop isn't passed within the component.

src/features/unified-share-modal/ContactsField.js
nickgriffin1: Update with improved flow typing

src/features/unified-share-modal/__tests__/ContactsField-test.js
```diff
@@ -107,6 +149,21 @@ describe('features/unified-share-modal/ContactsField', () => {
...
+                suggestedCollaborators: { '12345': { id: '12345' }, '23456': { id: '23456' } },
+            });
+            const addSuggestedContactsMock = jest.fn();
+
+            wrapper.setState({ pillSelectorInputValue: 'x@' });
+
+            wrapper.instance().addSuggestedContacts = addSuggestedContactsMock;
+            wrapper.instance().filterContacts(contactsFromServer);
+
+            expect(addSuggestedContactsMock).toHaveBeenCalledWith([expectedContacts[0]]);
```
FuzzyTree: Is there a test for verifying that addSuggestedContactsMock was not called when suggestedCollaborators is empty?

nickgriffin1: I'll add a test for that case.

PR #1037, opened 2019-04-03
fix(types): remove typecast that broke jump-to-definition

alexkrolick:


boxcla:

jstoffan:
Nice catch!

PR #1041, opened 2019-04-04
fix(query-bar): conditions are reinitialized upon opening Flyout

ChenCodes:
Changes in this PR:
- In `componentDidUpdate`, check to see if the Flyout has just been opened. Every time that the Flyout is opened, we will reinitialize `transientConditions` to be the conditions that the parent app sent down
- Use cloneDeep instead of slice when referencing `conditions`
- Repurpose `areAllValid` to take in both transientConditions and props.conditions as there are separate cases for when we want to check the validity of conditions from props and local conditions


boxcla:

src/features/query-bar/components/filter/FilterButton.js
```diff
@@ -56,12 +56,11 @@ class FilterButton extends React.Component<Props, State> {
...
-    componentDidUpdate() {
-        const { conditions } = this.props;
-        const { transientConditions } = this.state;
-        const wasFlyoutClosed = transientConditions.length === 0;
-        const shouldHydrateConditions = transientConditions.length !== conditions.length;
-        if (wasFlyoutClosed && shouldHydrateConditions) {
+    componentDidUpdate(prevProps: Props, prevState: State) {
+        const { isMenuOpen, transientConditions } = this.state;
+        const { isMenuOpen: prevIsMenuOpen } = prevState;
+        const wasApplyClicked = transientConditions.length === 0 && isMenuOpen && !prevIsMenuOpen;
```
macrotim: I'm curious what behaves differently if you omit "transientConditions.length === 0"

src/features/query-bar/__tests__/FilterButton-test.js
```diff
@@ -57,15 +57,18 @@ describe('feature/query-bar/components/filter/FilterButton', () => {
     });
 
     describe('componentDidUpdate()', () => {
-        const transientConditions = [];
-        const wrapper = getWrapper({
-            conditions,
-        });
-        wrapper.setState({
-            transientConditions,
+        test('should reinitialize conditions from props.conditions when the flyout is opened after clicking the Apply Button', () => {
```
macrotim: What if we relaxed the requirement to -> "whenever flyout is opened"

src/features/query-bar/components/filter/FilterButton.js
ChenCodes: Here is the behavior when the above is removed:

On initial load, No Filters Applied is being shown (due to Redux having no initialized conditions).

src/features/query-bar/__tests__/FilterButton-test.js
```diff
@@ -43,7 +43,7 @@ describe('feature/query-bar/components/filter/FilterButton', () => {
         });
 
         test('Should set areErrorsEnabled to true for Condition if not all conditions are valid', () => {
```
macrotim: Please rephrase: "if any condition is invalid"

src/features/query-bar/__tests__/FilterButton-test.js
```diff
@@ -57,15 +57,21 @@ describe('feature/query-bar/components/filter/FilterButton', () => {
...
     describe('componentDidUpdate()', () => {
-        const transientConditions = [];
-        const wrapper = getWrapper({
-            conditions,
-        });
-        wrapper.setState({
-            transientConditions,
+        test.each`
+            isMenuOpen | hasAppliedFilters | expected      | should
+            ${true}    | ${true}           | ${conditions} | ${'should reinitialize conditions from props.conditions when flyout is opened and apply button pressed'}
```
macrotim: It's misleading that the test is described as "and apply button pressed".

src/features/query-bar/components/filter/FilterButton.js
```diff
@@ -41,10 +42,11 @@ class FilterButton extends React.Component<Props, State> {
         super(props);
 
         this.state = {
-            transientConditions: this.props.conditions.slice(0),
-            selectedConnector: AND,
             areErrorsEnabled: false,
+            hasAppliedFilters: false,
             isMenuOpen: false,
+            selectedConnector: AND,
+            transientConditions: this.props.conditions.slice(0),
```
macrotim: There's a bug here. It's doing only a shallow copy which falls short of an immutable copy. Should use lodash.cloneDeep. 

I see multiple references to slice under src/features/query-bar//components.

src/features/query-bar/components/filter/FilterButton.js
```diff
@@ -56,12 +58,11 @@ class FilterButton extends React.Component<Props, State> {
...
-        const { conditions } = this.props;
-        const { transientConditions } = this.state;
-        const wasFlyoutClosed = transientConditions.length === 0;
-        const shouldHydrateConditions = transientConditions.length !== conditions.length;
-        if (wasFlyoutClosed && shouldHydrateConditions) {
+    componentDidUpdate(prevProps: Props, prevState: State) {
+        const { hasAppliedFilters, isMenuOpen } = this.state;
+        const { isMenuOpen: prevIsMenuOpen } = prevState;
+        const wasFlyoutOpened = isMenuOpen && !prevIsMenuOpen;
+        if (wasFlyoutOpened && hasAppliedFilters) {
```
macrotim: The purpose of `hasAppliedFilters` isn't obvious. Can you remind me?

src/features/query-bar/components/filter/FilterButton.js
```diff
@@ -118,6 +119,7 @@ class FilterButton extends React.Component<Props, State> {
                 onFilterChange(transientConditions);
             }
             this.setState({
+                hasAppliedFilters: true,
                 isMenuOpen: false,
                 transientConditions: [],
```
macrotim: Can you remind me why transientConditions=[] instead of null? The latter seems more appropriate.

macrotim:
@ChenCodes and I agreed to discard this change-set and hoist logic out of BUI instead.

src/features/query-bar/__tests__/FilterButton-test.js
```diff
@@ -58,16 +58,14 @@ describe('feature/query-bar/components/filter/FilterButton', () => {
 
     describe('componentDidUpdate()', () => {
         test.each`
-            isMenuOpen | hasAppliedFilters | expected      | should
-            ${true}    | ${true}           | ${conditions} | ${'should reinitialize conditions from props.conditions when flyout is opened and apply button pressed'}
-            ${true}    | ${false}          | ${[{}]}       | ${'should not reinitialize conditions from props.conditions when flyout is opened and apply button has not been pressed'}
-        `('$should', ({ isMenuOpen, hasAppliedFilters, expected }) => {
+            isMenuOpen | expected      | should
+            ${true}    | ${conditions} | ${'should reinitialize conditions from props.conditions when flyout is opened'}
```
macrotim: Is test.each needed?

src/features/query-bar/components/filter/FilterButton.js
```diff
@@ -257,8 +247,13 @@ class FilterButton extends React.Component<Props, State> {
         const { transientConditions, areErrorsEnabled, isMenuOpen, selectedConnector } = this.state;
 
         const numberOfConditions = conditions.length;
+        const areAllValid = this.areAllValid(conditions);
+        const shouldShowDefaultText = !areAllValid || numberOfConditions === 0;
```
macrotim: Can you explain this change? -> "!areAllValid"

What's the user pov?

src/features/query-bar/__tests__/FilterButton-test.js
ChenCodes: Not needed

src/features/query-bar/components/filter/FilterButton.js
ChenCodes: On the first load of the component, the Filter Button should show the Modify Filters text. The check for this in 
```const shouldShowDefaultText = !areAllValid || numberOfConditions === 0;```

In the areAllValid helper call, since the parent app is passing down a single initial condition that is not valid, we do not want to be display 1 Filters (since the condition is not valid), rather we want to display the Modify Filters text.

As for the numberOfConditions check, this is for handling the scenario where the user has deleted all of their conditions and then pressed the Apply button. Since conditions from parent app is now an empty array, we should also show the Modify Filters text.

File: src/features/query-bar/components/filter/FilterButton.js [hunk 10]
```diff
@@ -41,30 +42,27 @@ class FilterButton extends React.Component<Props, State> {
...
-                transientConditions: this.props.conditions.slice(0),
-            });
+    componentDidUpdate(prevProps: Props, prevState: State) {
+        const { isMenuOpen } = this.state;
+        const { isMenuOpen: prevIsMenuOpen } = prevState;
+        const wasFlyoutOpened = isMenuOpen && !prevIsMenuOpen;
+        if (wasFlyoutOpened) {
+            if (this.props.conditions.length === 0) {
+                this.setState({
+                    transientConditions: [this.createCondition()],
```
macrotim: if empty condition
1. and empty columns, then transientC = []
1. and non-empty columns, then transientC = [{...}]

In createCondition(), if no columns then throw Error('Columns required');

src/features/query-bar/components/filter/FilterButton.js
macrotim: Is this still needed after moving createCondition back into FilterButton?

src/features/query-bar/__tests__/FilterButton-test.js
```diff
@@ -28,8 +33,8 @@ describe('feature/query-bar/components/filter/FilterButton', () => {
             expect(Button.props().isDisabled).toBeFalsy();
         });
 
-        test('Should close the menu and empty out transientConditions when Apply button is clicked', () => {
-            const wrapper = getWrapper({ conditions: [] });
+        test('Should close the menu and clear out transientConditions when Apply button is clicked', () => {
+            const wrapper = getWrapper({ conditions: validConditions });
```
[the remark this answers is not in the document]
ChenCodes: A condition is considered valid if the values prop has a length greater than 0. What I can do is instead reference initialCondition as such:

```
const validConditions = [{ ...initialCondition, values: [1] }];
const incompleteConditions = [initialCondition];
```

src/features/query-bar/__tests__/Condition-test.js
```diff
@@ -76,7 +78,7 @@ describe('features/query-bar/components/filter/Condition', () => {
     });
 
     describe('handleValueChange()', () => {
-        const condition = initialCondition;
+        const condition = invalidCondition;
```
macrotim: The new name `invalidCondition` is more descriptive and helpful.

I assume that all tests are compatible with the concept of "invalid condition".

src/features/query-bar/__tests__/FilterButton-test.js
```diff
@@ -1,26 +1,26 @@
...
-    columnForDateType,
-    conditions as fixtureConditions,
-    initialCondition,
-} from '../components/fixtures';
+import { columnForTemplateFieldName, columnForDateType } from '../components/fixtures';
 import FilterButton from '../components/filter/FilterButton';
+import type { ConditionType } from '../flowTypes';
 import { EQUALS, LESS_THAN } from '../constants';
 
+const validCondition: ConditionType = { columnId: '3', id: '0', operator: EQUALS, values: [1] };
```
macrotim: '3' should be replaced with columnForTemplateFieldName.id (or whatever column uses id=3).

src/features/query-bar/__tests__/Condition-test.js
```diff
@@ -1,9 +1,11 @@
 // @flow
 import * as React from 'react';
 
-import { columnWithEnumType, initialCondition } from '../components/fixtures';
+import { columnWithEnumType } from '../components/fixtures';
 import Condition from '../components/filter/Condition';
+import { EQUALS } from '../constants';
 
+const invalidCondition: ConditionType = { columnId: '3', id: '0', operator: EQUALS, values: [] };
```
macrotim: '3' should be replaced with columnForTemplateFieldName.id (or whatever column uses id=3).

src/features/query-bar/__tests__/FilterButton-test.js
```diff
@@ -62,12 +62,19 @@ describe('feature/query-bar/components/filter/FilterButton', () => {
     });
 
     describe('componentDidUpdate()', () => {
+        const initialCondition = {
+            columnId: '7',
```
macrotim: Please replace. Otherwise this test will break if the fixture changes.

src/features/query-bar/__tests__/FilterButton-test.js
```diff
@@ -247,27 +254,18 @@ describe('feature/query-bar/components/filter/FilterButton', () => {
 
         test('Should return an empty object if columns is empty', () => {
```
macrotim: Rephrase to reflect change.

src/features/query-bar/__tests__/FilterButton-test.js
```diff
@@ -247,27 +254,18 @@ describe('feature/query-bar/components/filter/FilterButton', () => {
...
-        ];
-        const conditionsWithEmptyValues = [{ ...initialCondition, values: [] }];
-
         test.each`
-            description                                                                | transientConditions          | shouldCloseResult
-            ${'Should return true if Apply button was clicked and value is not empty'} | ${conditionsWithValues}      | ${true}
-            ${'Should return false if Apply button was clicked and value is empty'}    | ${conditionsWithEmptyValues} | ${false}
+            description                                                                | transientConditions  | shouldCloseResult
+            ${'Should return true if Apply button was clicked and value is not empty'} | ${validConditions}   | ${true}
+            ${'Should return false if Apply button was clicked and value is empty'}    | ${invalidConditions} | ${false}
```
macrotim: This improves readability.

src/features/query-bar/__tests__/Condition-test.js
ChenCodes: Yes they are compatible

PR #1042, opened 2019-04-04
fix(unified-share-modal): move shared link toggle tooltip to middle left

ittaishay:
avoids overlapping other elements in USM

Screen Shot 2019-04-03 at 6 39 37 PM


boxcla:

booc0mtaco:
Does this only apply to an FTUX modal in the unified share modal?

ittaishay:
> Does this only apply to an FTUX tooltip in the unified share modal?

@booc0mtaco yup, one that I added about a week ago

PR #1035, opened 2019-04-03
 feat(versions): Add support for version delete and promote actions

jstoffan:
I also consolidated the version navigation in the container to avoid duplicating the logic.

Todo:
- [x] Unit tests for container

PR #1043, opened 2019-04-04
 refactor(versions): Move history logic from version item to container

jstoffan:
This change ended up being fairly large, so I wanted to break it into its own pull request. It should be merged before #1035. The new delete and promote actions will require programatic access to set the currently-selected version. Managing this state in the container centralizes the navigation logic nicely.


boxcla:

PR #1044, opened 2019-04-04
feat(unified-share-modal): support for default invite permission level

prakash-kumar:


boxcla:

src/features/unified-share-modal/InviteePermissionsMenu.js
```diff
@@ -66,8 +67,15 @@ class InviteePermissionsMenu extends Component<Props> {
...
-        const { inviteePermissionsButtonProps, inviteePermissionLevel, disabled, itemType } = this.props;
-
+        const {
+            inviteePermissions,
+            inviteePermissionsButtonProps,
+            inviteePermissionLevel,
+            disabled,
+            itemType,
+        } = this.props;
+        const defaultPermissionLevel = getDefaultPermissionLevel(inviteePermissions);
```
shahzadaziz: Is there a reason why we can't pass `inviteePermissions` downstream from USM since we are already doing the same logic there. We can refactor this as a USM method `getInviteePermission` or something.

src/features/unified-share-modal/__tests__/__snapshots__/UnifiedShareModal-test.js.snap
```diff
@@ -125,7 +125,7 @@ exports[`features/unified-share-modal/UnifiedShareModal closeEmailSharedLinkForm
             <InviteePermissionsMenu
               changeInviteePermissionLevel={[Function]}
               disabled={true}
-              inviteePermissionLevel="Editor"
+              inviteePermissionLevel=""
```
tjgupta: Is there a stub you should update so this still has a value?

PR #1045, opened 2019-04-04
feat(utils): add fuzzySearch function

nickgriffin1:


boxcla:

src/utils/__tests__/fuzzySearch-test.js
```diff
@@ -0,0 +1,28 @@
...
+    test('should return false when content is empty', () => {
+        expect(fuzzySearch('foo', '')).toBe(false);
+    });
+
+    test('should return false when seach is empty', () => {
+        expect(fuzzySearch('', 'foo bar')).toBe(false);
+    });
+
+    test('should only do exact matching when search length equals content length', () => {
+        expect(fuzzySearch('foo bat', 'foo bar')).toBe(false);
```
alexkrolick: It seems like it would be handy to include fuzzy results after the exact matches here?

PR #1047, opened 2019-04-04
fix(query-bar): incorrect text on ColumnButton after hiding

ChenCodes:
Changes in this PR:
- When one column is hidden, show `1 Column Hidden` on the Column Button instead of `1 Columns Hidden`

Screen Shot 2019-04-04 at 3 40 36 PM


boxcla:

i18n/en-US.properties
```diff
@@ -808,6 +808,8 @@ boxui.presence.timeSinceLastPreviewed = Previewed {timeAgo}
 boxui.queryBar.addFilterButtonText = + Add Filter
 # Text on the apply filter button, on click applies the filters
 boxui.queryBar.applyFiltersButtonText = Apply
+# Text on the columns button, if one column has been hidden then it will display this text
+boxui.queryBar.columnHiddenButtonText = {number} Column Hidden
```
macrotim: Wouldn't you replace {number} with 1?

src/features/query-bar/__tests__/ColumnButton-test.js
```diff
@@ -28,6 +28,26 @@ describe('features/query-bar/components/ColumnButton', () => {
             const wrapper = getWrapper({ template });
             expect(wrapper).toMatchSnapshot();
         });
+
+        test('should render ColumnButton with correct message when all columns are visible', () => {
```
macrotim: test.each?

src/features/query-bar/components/ColumnButton.js
```diff
@@ -67,6 +67,28 @@ class ColumnButton extends React.Component<Props, State> {
 
         const buttonClasses = classNames('query-bar-button', numberOfHiddenColumns !== 0 ? 'is-active' : '');
 
+        let columnsButtonText;
+        let message;
+        switch (true) {
```
macrotim: ?

Please switch to if().

src/features/query-bar/__tests__/ColumnButton-test.js
```diff
@@ -28,6 +27,21 @@ describe('features/query-bar/components/ColumnButton', () => {
             const wrapper = getWrapper({ template });
             expect(wrapper).toMatchSnapshot();
         });
+
+        const allVisibleColumns = [columnForDateType, columnWithFloatType];
```
patrickpaul: -> `visibleColumns`

i18n/en-US.properties
```diff
@@ -808,6 +808,8 @@ boxui.presence.timeSinceLastPreviewed = Previewed {timeAgo}
 boxui.queryBar.addFilterButtonText = + Add Filter
 # Text on the apply filter button, on click applies the filters
 boxui.queryBar.applyFiltersButtonText = Apply
+# Text on the columns button, if one column has been hidden then it will display this text
+boxui.queryBar.columnHiddenButtonText = 1 Column Hidden
```
patrickpaul: We don't need multiple messages... make use of `react-intl`. The unified message could look like:

`{count, plural, =1 {Column} other {Columns} Hidden`

PR #1049, opened 2019-04-05
 feat(tasks): Add general and approval task type icons

shahzadaziz:


boxcla:

src/icons/tasks/IconTaskApproval.js
```diff
@@ -0,0 +1,30 @@
+// @flow
+import * as React from 'react';
+
+import AccessibleSVG from '../accessible-svg';
+
+type Props = {
+    className?: string,
+    color?: string,
```
jstoffan: I'm under the impression we intend to style icons via CSS classes exclusively going forward. @tjgupta, can you take a quick look?

conf/styleguide.config.js
```diff
@@ -174,6 +174,14 @@ const allSections = [
                 name: 'States',
                 content: '../src/icons/states/README.md',
             },
+            {
+                name: 'Tasks',
+                components: () => [
```
jstoffan: We've started to prefer a single content page for multiple small components. See `name: 'States'` above for an example.

src/icons/tasks/IconTaskGeneral.js
```diff
@@ -0,0 +1,35 @@
...
+type Props = {
+    className?: string,
+    color?: string,
+    height?: number,
+    /** A text-only string describing the icon if it is not purely decorative for accessibility */
+    title?: string,
+    width?: number,
+};
+
+const IconTaskGeneral = ({ className = '', color = '#26C281', height = 30, title, width = 30 }: Props) => (
```
jstoffan: It looks like these icons are using different default dimensions and viewbox values than most others. Is that intentional?

src/icons/tasks/IconTaskApproval.js
tjgupta: Yep, I can work with @shahzadaziz to conform this to our two-toned icon style.

conf/styleguide.config.js
booc0mtaco: @jstoffan when using README.md, that requires ... `require()` for each component, correct? with the individual files, the markdown files seem to infer which component to import.

src/icons/tasks/IconTaskGeneral.js
shahzadaziz: yes it was intentional.

conf/styleguide.config.js
jstoffan: Yes, that's true. The READMEs serve a dual purpose as Github documentation and generators for the example site. Having a list of `require` statements at the top of the README seems like a reasonable tradeoff, but if that's a concern, then the other icon READMEs will also need to be reverted to single components. 

@booc0mtaco @tjgupta, can you sync up on the desired strategy here?

PR #1050, opened 2019-04-05
chore(debugging): Add Elements version to window when components mount

jeremypress:


boxcla:

src/utils/globals.js
```diff
@@ -0,0 +1,9 @@
+// @flow
+export default function setElementsVersion() {
+    if (window.Box) {
+        window.Box.Elements = {
```
jstoffan: It may be a bit safer to lazily create the `Elements` object and add defaults to avoid the eslint issues:

```
window.Box.Elements = window.Box.Elements || {};
window.Box.Elements.version = __VERSION__ || '';
```

src/utils/globals.js
```diff
@@ -0,0 +1,9 @@
+// @flow
+export default function setElementsVersion() {
```
jstoffan: Since this is in a generic `globals` util, should this be a named export rather than the deafult?

src/elements/content-picker/ContentPicker.js
```diff
@@ -206,6 +207,8 @@ class ContentPicker extends Component<Props, State> {
             isLoading: false,
             errorCode: '',
         };
+
+        setElementsVersion();
```
jstoffan: This would imply that, if we want to add additional items to the global object, we'd need one function call per line. Would something like `globals.init()` give us more breathing space?

package.json
```diff
@@ -112,7 +112,8 @@
             "box-ui-elements-locale-data": "<rootDir>/i18n/en-US.js",
             "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "<rootDir>/conf/jest/fileMock.js",
             "\\.(css|less|scss)$": "<rootDir>/conf/jest/mocks/styleMock.js",
-            "react-virtualized/dist/es": "react-virtualized/dist/commonjs"
+            "react-virtualized/dist/es": "react-virtualized/dist/commonjs",
+            "globals": "<rootDir>/conf/jest/mocks/globalsMock.js"
```
jstoffan: Can we define `__VERSIONS__` in `env-setup.js`?

jeremypress: Tried this, and it doesn't seem to work. We can't set __VERSIONS__ directly because it doesn't exist globally yet, and setting global.__VERSIONS__ is different from what the plugin does, so we can't read that value in our util method.

src/elements/content-picker/ContentPicker.js
jeremypress: I left it as a specific method in case different consumers only care about certain properties.

src/utils/globals.js
```diff
@@ -0,0 +1,9 @@
+// @flow
+// eslint-disable-next-line import/prefer-default-export
+export function setElementsDebugInfo() {
```
jstoffan: You could consider exporting this as part of an object to avoid the eslint rule above.

src/utils/globals.js
```diff
@@ -0,0 +1,9 @@
...
+    if (window.Box) {
+        window.Box.Elements = window.Box.Elements || {};
+        // $FlowFixMe
```
jstoffan: Might be worth adding a comment as to why this is needed.

ConradJChan:
Might be overkill, but would there be any value in creating a HOC which handles these debug type things but for now would call `globalUtils.setElementsDebugInfo()`.

jeremypress:
@ConradJChan I think that would be a good next step once we learn what the use case will expand to going forward.

PR #1057, opened 2019-04-08
style(elements): Fix various style specificity issues; fix modal height

jstoffan:
This PR addresses the following issues:

- **Critical**: Preview, rename, share, and create folder modals do not show at all in ContentExplorer or ContentPicker
- Access Stats items flash on hover in EUA due to specificity issues with PlainButton
- Access Stats anchor text and icons are not vertically centered
- Access Stats anchor text has its own hover treatment rather than using that of its parent button
- Breadcrumb list show a large gap before subsequent breadcrumbs when viewing a folder with a short name
- Metadata instance editor date fields did not take up the full width of the sidebar in the same way as standard text inputs

### DIFF
--- a/conf/styleguide.styles.scss
+++ b/conf/styleguide.styles.scss
@@ -1,13 +1,10 @@
 .be {
-    &.bcow {
-        width: 100px;
+    &.bce,
+    &.bcp {
+        height: 600px;
     }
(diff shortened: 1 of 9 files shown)


boxcla:

PR #1058, opened 2019-04-09
feat(utils): add fuzzySearch function

smoorebox:


boxcla:
Verified that @smoorebox has signed the CLA. Thanks for the pull request!

src/utils/__tests__/fuzzySearch-test.js
```diff
@@ -0,0 +1,88 @@
...
+        const testSet = [];
+        for (let i = 0; i < 1000; i += 1) {
+            testSet.push('alskdgjd;fjgenrtgushdfg;sadfjgsdfadslkfhagbsdfgsdfgsdfgsdfgsdgasdfa');
+        }
+        const startTime = new Date();
+        const result = testSet.every(content => {
+            return fuzzySearch('adslkfhagbsdfgsdfgsdfgsdfgsdgasdfa', content);
+        });
+        const endTime = new Date();
+        expect(endTime - startTime).toBeLessThan(250);
```
smoorebox: Not sure if this is an OK thing to do, can remove this test if it's not.

nickgriffin1: I like it. I don't often see perf testing in a unit test. I don't see any reason why we can't reduce `testSet` and the expected time by a factor of 10 though.

smoorebox: Have that change ready to go, will get a few more eyes on it before doing that.

src/utils/fuzzySearch.js
```diff
@@ -0,0 +1,55 @@
...
+        if (subScore > totalScore) {
+            totalScore = subScore;
+        }
+    }
+    if (totalScore > 0) {
+        const maxGroups = Math.min(maxGaps, searchLength);
+        // minScore is calculated as a near-worst-case score given an even distribution of gaps
+        // since the algorithm rewards streak of characters breaking them up evenly is the worst case
+        // minimum score should also be better than just each character individually
+        const minScore = Math.max(maxGroups * 2 ** Math.floor(searchLength / maxGroups - 1), searchLength + 1);
```
FuzzyTree: double ** ?

src/utils/fuzzySearch.js
```diff
@@ -0,0 +1,55 @@
...
+        if (contentLength - i < searchLength) {
+            break;
+        }
+        let searchIndex = 0;
+        let currentScore = 0;
+        let subScore = 0;
+        for (let j = i; j < contentLength; j += 1) {
+            if (content[j] === search[searchIndex]) {
+                searchIndex += 1;
+                currentScore += 1 + currentScore;
```
FuzzyTree: is currentScore supposed to double here?

src/utils/fuzzySearch.js
```diff
@@ -0,0 +1,55 @@
...
+const fuzzySearch = (search: string, content: string, minCharacters: number = 3, maxGaps: number = 2) => {
+    content = content.toLowerCase().replace(/\s/g, '');
+    search = search.toLowerCase().replace(/\s/g, '');
+    const contentLength = content.length;
+    const searchLength = search.length;
+    if (searchLength < minCharacters || searchLength > contentLength) {
+        return false;
+    }
+    let totalScore = 0;
+    for (let i = 0; i < contentLength; i += 1) {
```
FuzzyTree: nit: use `++` instead of `+= 1`?

src/utils/fuzzySearch.js
smoorebox: Yep, it's the Math.pow operator, linting made me do it.

src/utils/fuzzySearch.js
smoorebox: Yeah, it's supposed to increase exponentially, I'll add a comment.

src/utils/fuzzySearch.js
smoorebox: Linting made me do it.

src/utils/fuzzySearch.js
FuzzyTree: the linter doesn't let you use Math.pow?

smoorebox: Yes, it specifically told me to change it to the native operator, new to me as well since I'm rusty.

j3tan: it's all the rage: http://2ality.com/2016/02/exponentiation-operator.html

j3tan: may want some parens though just for clarity

src/utils/__tests__/fuzzySearch-test.js
FuzzyTree: Seems like this test has the potential to be flaky

src/utils/fuzzySearch.js
```diff
@@ -0,0 +1,55 @@
...
+ * @author Box
+ *
+ * @param {string} search User input search string
+ * @param {string} content Content to search over for matches
+ * @param {number} minCharacters Minimum number of search characters before matching anything, default 3
+ * @param {number} maxGaps Approximate maximum number of gaps in the search string to tune fuzzyness, default 2
+ * @returns {boolean} If a match is found
+ */
+const fuzzySearch = (search: string, content: string, minCharacters: number = 3, maxGaps: number = 2) => {
+    content = content.toLowerCase().replace(/\s/g, '');
```
jstoffan: Although our eslint config allows it and these are pass-by-value, I think we usually avoid param reassignment in most cases. Can we assign these to some new local variables?

src/utils/fuzzySearch.js
```diff
@@ -0,0 +1,55 @@
...
+            totalScore = subScore;
+        }
+    }
+    if (totalScore > 0) {
+        const maxGroups = Math.min(maxGaps, searchLength);
+        // minScore is calculated as a near-worst-case score given an even distribution of gaps
+        // since the algorithm rewards streak of characters breaking them up evenly is the worst case
+        // minimum score should also be better than just each character individually
+        const minScore = Math.max(maxGroups * 2 ** Math.floor(searchLength / maxGroups - 1), searchLength + 1);
+        return totalScore >= minScore;
```
jstoffan: Per @priyajeet, I think we've been trying to avoid multiple non-guard returns to make debugging a bit easier. Can we introduce a `let result = false;` local variable after the initial length-based guard clause?

src/utils/fuzzySearch.js
```diff
@@ -0,0 +1,55 @@
...
+    }
+    return false;
+};
+
+export default fuzzySearch;
```
jstoffan: The comment above ("Basic fuzzy search alogrithm") implies that we may introduce other algorithms in the future. Should we export a named function to provide flexibility for such? Maybe even rename the file to just `search`?

smoorebox: I'll remove "basic", I think it's best described as a fuzzy search util.

PR #1060, opened 2019-04-09
feat(card): create card component

ittaishay:
## Full View
Screen Shot 2019-04-08 at 10 50 30 PM

## Small Width View
Screen Shot 2019-04-08 at 10 54 42 PM

### DIFF
--- a/src/components/card/Card.js
+++ b/src/components/card/Card.js
@@ -1,0 +1,26 @@
+/* @flow */
+import * as React from 'react';
+import classNames from 'classnames';
+
+import './Card.scss';
+
+type Props = {
+    children: React.Node,
+    className?: string,
+    hasPadding?: boolean,
+    title?: React.Node,
+};
+
+const Card = ({ children, className = '', hasPadding = true, title, ...rest }: Props) => (
+    <div
+        className={classNames(`bdl-Card ${className}`, {
+            'bdl-Card--padded': hasPadding,
+        })}
+        {...rest}
+    >
+        {title && <div className="bdl-Card-title">{title}</div>}
+        {children}
+    </div>
+);
+
+export default Card;
--- a/src/components/card/Card.scss
+++ b/src/components/card/Card.scss
@@ -1,0 +1,22 @@
+@import '../../styles/variables';
+
+.bdl-Card {
+    background-color: $white;
+    border-radius: 3px;
+    box-shadow: 0 1px 3px rgba(0, 0, 0, .2);
+    margin: 0 auto;
+    max-width: 900px;
+
+    .bdl-Card-title {
+        font-size: 20px;
+        font-weight: bold;
+        margin-bottom: 30px;
+        text-align: center;
+    }
+}
+
+.bdl-Card--padded {
+    padding: 40px 60px 45px;
+}
+
+@import './Card.media';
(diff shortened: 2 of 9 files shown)


boxcla:

src/components/card/Card.js
```diff
@@ -0,0 +1,26 @@
+/* @flow */
+import * as React from 'react';
+import classNames from 'classnames';
+
+import './Card.scss';
+
+type Props = {
+    children?: React.Node,
```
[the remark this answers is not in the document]
ittaishay: nope, changed to required

src/components/card/Card.js
```diff
@@ -0,0 +1,26 @@
...
+    className?: string,
+    removePadding?: boolean,
```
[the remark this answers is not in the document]
ittaishay: done

priyajeet:
We do have a thumbnail card component, can that at all be re-purposed and re-used for text?

src/components/card/Card.scss
```diff
@@ -0,0 +1,22 @@
+@import '../../styles/variables';
+
+.bdl-Card {
+    background-color: $white;
+    border-radius: 3px;
+    box-shadow: 0 1px 3px rgba(0, 0, 0, .2);
+    margin: 0 auto;
+    max-width: 900px;
+
+    .bdl-Card-title {
```
jstoffan: Please avoid nesting classes unless necessary, as it increases specificity. The SUIT naming convention also makes it easier to avoid.

src/components/card/Card.scss
```diff
@@ -0,0 +1,22 @@
+@import '../../styles/variables';
+
+.bdl-Card {
+    background-color: $white;
+    border-radius: 3px;
+    box-shadow: 0 1px 3px rgba(0, 0, 0, .2);
+    margin: 0 auto;
+    max-width: 900px;
```
jstoffan: What is the reason for selecting this value for max width? It seems tailored to a specific use-case, which isn't ideal for a reusable component.

src/components/card/Card.scss
```diff
@@ -0,0 +1,22 @@
+@import '../../styles/variables';
+
+.bdl-Card {
+    background-color: $white;
+    border-radius: 3px;
+    box-shadow: 0 1px 3px rgba(0, 0, 0, .2);
+    margin: 0 auto;
```
jstoffan: This will center the card in its container, which may not be desirable in many circumstances. Please consider multiple use cases for common components.

src/components/card/Card.js
```diff
@@ -0,0 +1,26 @@
...
+    children: React.Node,
+    className?: string,
+    hasPadding?: boolean,
+    title?: React.Node,
+};
+
+const Card = ({ children, className = '', hasPadding = true, title, ...rest }: Props) => (
+    <div
+        className={classNames(`bdl-Card ${className}`, {
+            'bdl-Card--padded': hasPadding,
```
jstoffan: CSS classes ideally should avoid including specific visual aspects in their names.

src/components/card/Card.js
```diff
@@ -0,0 +1,26 @@
...
+
+type Props = {
+    children: React.Node,
+    className?: string,
+    hasPadding?: boolean,
+    title?: React.Node,
+};
+
+const Card = ({ children, className = '', hasPadding = true, title, ...rest }: Props) => (
+    <div
```
jstoffan: Should this be an `article` with a heading tag to represent the `title`?

src/components/card/Card.scss
priyajeet: I would not add margins at all, that depends/affects on the parent layout, component should ideally be encapsulated and full bleed

ittaishay:
closing this PR for now, will re-evaluate per @jstoffan's comment

PR #1061, opened 2019-04-09
feat(icon): add location error state

Sunyang730:
Screen Shot 2019-04-09 at 8 29 32 AM


boxcla:

src/icons/states/LocationErrorState.js
```diff
@@ -0,0 +1,93 @@
...
+const LocationErrorState = ({
+    className = '',
+    color = boxBlue,
+    height = 111,
+    opacity = 0.1,
+    title,
+    width = 130,
+}: Props) => (
+    <AccessibleSVG
+        className={`LocationErrorState ${className}`}
```
priyajeet: For BUIE, you would prefix the class with `bdl`. So in this case `bdl-LocationErrorState`

src/icons/states/LocationErrorState.js
```diff
@@ -0,0 +1,93 @@
...
+        height={height}
+        title={title}
+        viewBox="0 0 140 111"
+        width={width}
+    >
+        <g fill="none" fillRule="evenodd">
+            <path
+                fill={color}
```
priyajeet: Check the other state icons, and if they are doing it, add classes like fill-color or stroke-color (yeah old style for now) to each path entry for custom branding override. Ignore if the state should not have custom branding override. I think we just either use black or blue?

Sunyang730: We use black for custom branding. I don't see the classes in other state icon and the color seems to be coming just from the prop itself.

src/icons/states/LocationErrorState.js
Sunyang730: I thought `bdl` namespaces applies when we have class that has Box specific style. Is that not the case?

Sunyang730: nvm We want to make sure this does not have conflict with other styles that the user might have. Curious, any reason we don't want to use classnames?

priyajeet: you can use classnames if you want. Though with just a single appending its not as useful

Sunyang730: Can get rid of the extra space :) but not going to do it for this commit.

src/icons/states/LocationErrorState.js
priyajeet: yeah but you will see things like

className="stroke-color"
className="fill-color"

in a few icons which is what gets overridden for custom branding

src/icons/states/LocationErrorState.js
Sunyang730: Going to add classnames since I am addressing comments.

src/icons/states/LocationErrorState.js
```diff
@@ -0,0 +1,115 @@
...
+            />
+            <path
+                className="fill-color"
```
priyajeet: should this just be stroke color?

Sunyang730: hmm, I guess we are not changing the fill. Yes, will update.

PR #1064, opened 2019-04-09
fix(tasks): update design for due date label

booc0mtaco:
This includes changes to the tasks experience, and some necessary fixes to the handling of relative dates within thresholds.

A few changes allow proper snapshotting, and fix some unhandled test warnings that occur in Jest.


boxcla:

src/elements/common/messages.js
```diff
@@ -858,6 +858,11 @@ const messages = defineMessages({
         defaultMessage: 'Decline',
         description: 'Reject option for a task',
     },
+    taskDueDateLabel: {
+        id: 'be.tasks.taskDueDate',
+        defaultMessage: 'Due: ',
```
alexkrolick: Is this the right way to translate something like "Due: Jan 19 at 2pm"? It seems like the position of the date string may vary across languages and RTL setting.

src/elements/content-sidebar/activity-feed/task-new/Task.js
```diff
@@ -119,7 +119,6 @@ class Task extends React.Component<Props> {
                     onDelete={onDelete}
                     onEdit={onEdit}
                     permissions={taskPermissions}
-                    messageHeader={<TaskDueDate dueDate={due_at} status={status} />}
```
jstoffan: Can the `messageHeader` prop be removed from `Comment`? I think we added it recently to support the new task design?

src/elements/common/messages.js
jstoffan: Agreed, this seems a bit odd. Can we use a variable for the date, at least?

booc0mtaco: this is actually true for the previously used translation, since it has no label separator. in this case, "Due: " functions as a field label so I feel this is fine. 

The other field is part of a sentence structure, and should actually be one full message instead of two. I didn't change it to avoid complicating this review

booc0mtaco: Fair point on the ordering; this will always put the label to the left of the date because of the two components. I can fix this

src/elements/content-sidebar/activity-feed/task-new/Task.js
booc0mtaco: I can take it out, but was worried that since it was both requested to add this *and* remove it, removing it entirely might be premature

booc0mtaco:
Screen Shot 2019-04-09 at 9 59 50 PM

src/components/time/ReadableTime.js
```diff
@@ -2,31 +2,33 @@
 import React from 'react';
 import { FormattedMessage, FormattedRelative, FormattedDate } from 'react-intl';
 
+import { ONE_HOUR_MS } from '../../constants';
 import { isToday, isYesterday, isCurrentYear } from '../../utils/datetime';
 
 import messages from './messages';
 
 type Props = {
-    /** The timestamp which should be used to display the date */
```
booc0mtaco: The comments for the props were in the wrong order.

src/features/presence/__tests__/__snapshots__/Presence-test.js.snap
```diff
@@ -46,7 +46,7 @@ exports[`features/presence/Presence render() Tests for presence autofly - GROWTH
                 id="boxui.presence.timeSinceLastPreviewed"
                 values={
                   Object {
-                    "timeAgo": <UNDEFINED
+                    "timeAgo": <FormattedRelative
```
booc0mtaco: adding the update to the mock react intl file prevents this from rending invariant errors in the test now, and allows for more meaningful snapshots in future components (more to fix in the mocks tho).

PR #1065, opened 2019-04-09
chore(content-preview): Trigger callback when a new version is previewed

jeremypress:
This callback will allow ContentPreview/ContentExplorer or other consumers to control the header when a version is previewed.

Open Questions:
1. Should we send the whole version info back in the callback? This will allow consumers to skip additional network calls and directly add the new version info into their redux store/state.
2. Can we always assume that the current version will be at the beginning of the versions array?

### DIFF
--- a/flow-typed/box-ui-elements.js
+++ b/flow-typed/box-ui-elements.js
@@ -966,4 +966,9 @@ type NavigateOptions = {
     isToggle?: boolean,
 };
 
-type OnVersionChange = (version: ?BoxItemVersion, additionalVersionInfo: ?Object) => void;
+type AdditionalVersionInfo = {
+    isCurrentVersion: boolean,
+    updateVersionToCurrent: () => void,
+};
+
+type OnVersionChange = (version: ?BoxItemVersion, additionalVersionInfo: ?AdditionalVersionInfo) => void;
--- a/flow-typed/box-ui-elements.js
+++ b/flow-typed/box-ui-elements.js
@@ -965,3 +965,5 @@ type ContentSidebarFeatures = {
 type NavigateOptions = {
     isToggle?: boolean,
 };
+
+type OnVersionChange = (version: ?BoxItemVersion, additionalVersionInfo: ?Object) => void;
--- a/src/elements/content-preview/ContentPreview.js
+++ b/src/elements/content-preview/ContentPreview.js
@@ -75,6 +75,7 @@ type Props = {
     onDownload: Function,
     onLoad: Function,
     onNavigate: Function,
+    onVersionChange: (versionId?: string, isCurrentVersion?: boolean) => void,
     previewLibraryVersion: string,
     requestInterceptor?: Function,
     responseInterceptor?: Function,
@@ -187,6 +188,7 @@ class ContentPreview extends PureComponent<Props, State> {
         onError: noop,
         onLoad: noop,
         onNavigate: noop,
+        onVersionChange: noop,
         previewLibraryVersion: DEFAULT_PREVIEW_VERSION,
         showAnnotations: false,
         staticHost: DEFAULT_HOSTNAME_STATIC,
@@ -1048,7 +1050,9 @@ class ContentPreview extends PureComponent<Props, State> {
      *
      * @param {string} versionId - The version id to set as current
      */
-    onVersionChange = (versionId?: string) => {
+    onVersionChange = (versionId?: string, isCurrentVersion?: boolean) => {
+        const { onVersionChange }: Props = this.props;
+        onVersionChange(versionId, isCurrentVersion);
         this.setState({ selectedVersionId: versionId });
     };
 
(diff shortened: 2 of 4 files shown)


boxcla:

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -54,7 +54,9 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
 
         // Forward the current version id that is passed in via the wrapping route
         if (prevVersionId !== versionId) {
-            onVersionChange(versionId);
+            const { versions } = this.state;
+            const isCurrentVersion = versionId === versions[0].id;
+            onVersionChange(versionId, isCurrentVersion);
```
jstoffan: I think we should consider passing another callback here as a param. The callback param could be used by the host app to completely exit the version history preview. 

I'd also suggest moving `isCurrentVersion` to a property in an object to clean up the method signature.

src/elements/content-preview/ContentPreview.js
```diff
@@ -74,6 +74,7 @@ type Props = {
     onDownload: Function,
     onLoad: Function,
     onNavigate: Function,
+    onVersionChange: (versionId?: string, isCurrentVersion?: boolean) => void,
```
jstoffan: Should the function signature move to its own named flow type, since it's used in several disparate components?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -54,7 +54,9 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
 
         // Forward the current version id that is passed in via the wrapping route
         if (prevVersionId !== versionId) {
-            onVersionChange(versionId);
+            const { versions } = this.state;
+            const isCurrentVersion = versionId === versions[0].id;
```
jstoffan: Necessary to check id or does simple object equality work?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
jeremypress: I think the behavior on close is to actual go back to previewing the current version.
Either way a callback could be helpful here. My other approach was for the consumer to call preview.show() on the file again which would reset to the current version.

src/elements/content-preview/ContentPreview.js
jeremypress: sure, will be a part of the final PR.

ConradJChan:
I think it would make sense to return a copy of the `BoxItemVersion` in the callback, especially in the case where the application wants to update its header, it has the information there already instead of, as you said, having to make extra network calls to retrieve it

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
jstoffan: I don't think the consumer has direct access or control over when the `preview.show()` method gets called. Even if it did, that could get pretty tricky given the internal state tracking in Elements.

flow-typed/box-ui-elements.js
```diff
@@ -965,3 +965,5 @@ type ContentSidebarFeatures = {
 type NavigateOptions = {
     isToggle?: boolean,
 };
+
+type OnVersionChange = (version: ?BoxItemVersion, additionalVersionInfo: ?Object) => void;
```
ConradJChan: can `additionalVersionInfo` have a flow type too to avoid Object

src/elements/content-preview/ContentPreview.js
```diff
@@ -1046,10 +1048,13 @@ class ContentPreview extends PureComponent<Props, State> {
...
-     * @param {string} versionId - The version id to set as current
+     * @param {string} [version] - The version that is now previewed
+     * @param {object} [additionalVersionInfo] - extra info about the version
      */
-    onVersionChange = (versionId?: string) => {
-        this.setState({ selectedVersionId: versionId });
+    onVersionChange = (version?: BoxItemVersion, additionalVersionInfo?: Object): void => {
+        const { onVersionChange }: Props = this.props;
+        onVersionChange(version, additionalVersionInfo);
+        this.setState({ selectedVersionId: getProp(version, 'id') });
```
ConradJChan: When would `version` be undefined? Is `selectedVersionId` as undefined a valid state?

jeremypress: Whenever the `versionSidebarContainer` unmounts, it calls onVersionChange() with no version ID. It also gets set to undefined directly when ContentPreview destroys preview. I think it's used as a way to clear the preview logic around needing to reload, and it can be used to set the header back to the normal one before when a fresh preview is happening. 

in `shouldLoadPreview`,  ``` if (selectedVersionId !== prevSelectedVersionId) {```

will always be true when the prev ID is undefined, which will force preview to reload. The header will also be forced to reset since it has no version information.

ConradJChan: 👍 Thanks for the explanation!

PR #1066, opened 2019-04-09
feat(tooltip): render children when disabled or text is missing

priyajeet:
The tooltip will not show and just render its children when `text` is missing or if `isDisabled` is `true`.

### DIFF
--- a/src/components/tooltip/__tests__/Tooltip-test.js
+++ b/src/components/tooltip/__tests__/Tooltip-test.js
@@ -203,6 +203,22 @@ describe('components/tooltip/Tooltip', () => {
 
             expect(wrapper.find('[role="tooltip"]').hasClass('is-error')).toBe(true);
         });
+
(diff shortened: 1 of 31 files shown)


boxcla:

src/components/tooltip/__tests__/Tooltip-test.js
```diff
@@ -203,6 +203,22 @@ describe('components/tooltip/Tooltip', () => {
 
             expect(wrapper.find('[role="tooltip"]').hasClass('is-error')).toBe(true);
         });
+
+        test('should render children only when tooltip is disabled', () => {
```
priyajeet: bogus

jeremypress:
`src/elements/common/Tooltip.js` We added this before the merger which does something similar if you want to re-use.

priyajeet:
> `src/elements/common/Tooltip.js` We added this before the merger which does something similar if you want to re-use.

Plan is to nuke that

PR #1067, opened 2019-04-09
fix(versions): Add support for null user values for version items

jstoffan:


boxcla:

ConradJChan:
Can we add a unit test?

PR #1069, opened 2019-04-10
fix(sidebar): Sidebar nav button tooltips should display on hover

jstoffan:
Tooltips


boxcla:

src/elements/content-sidebar/SidebarNavButton.js
```diff
@@ -22,12 +22,12 @@ const SidebarNavButton = ({ children, interactionTarget, onNavigate, sidebarView
...
-                {({ match }) => {
-                    const isToggle = !!match && match.isExact;
+        <Route path={sidebarPath}>
+            {({ match }) => {
+                const isToggle = !!match && match.isExact;
 
-                    return (
+                return (
+                    <Tooltip position="middle-left" text={tooltip}>
                         <NavButton
```
priyajeet: NavButton also has <Route> in it, should tooltip be inside there around the actual component?

jstoffan: It worked as-is when I tested it locally. The NavButton component forwards all its props to its child, which may explain why it works properly.

jstoffan:
I'll put adding an e2e test on the todo list, but I think we have other work to address at the moment. This will at least fix the bug in the immediate term.

PR #1074, opened 2019-04-10
feat(tasks): Add task type icon to task creator avatar

booc0mtaco:

### DIFF
--- a/src/elements/content-sidebar/activity-feed/comment/Comment.js
+++ b/src/elements/content-sidebar/activity-feed/comment/Comment.js
@@ -28,6 +28,7 @@ import './Comment.scss';
 import { PLACEHOLDER_USER } from '../../../../constants';
 
 type Props = {
+    avatarRenderer?: React.Node => React.Element<any>,
     created_at: string | number,
     created_by: User,
     currentUser?: User,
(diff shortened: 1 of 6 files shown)


boxcla:

booc0mtaco:
Screen Shot 2019-04-10 at 4 12 41 PM

src/elements/content-sidebar/activity-feed/comment/Comment.js
```diff
@@ -93,6 +94,7 @@ class Comment extends React.Component<Props, State> {
 
     render(): React.Node {
         const {
+            avatarRenderer = identity,
```
jeremypress: This is really cool. Do you mind dropping a comment explaining what this does?

booc0mtaco: Sure! there are three parts:

- [`identity`](https://lodash.com/docs/4.17.11#identity) let's you get exactly the thing you sent as a parameter back when you call a function. `(prop) => { return prop }`. *Using this as a default value for the renderer, so it will maintain the previous behavior*
- in the implementation, we wrap the original component with `avatarRenderer` so that, by default, it will render just itself. so that's lower down in the comment component
- lastly, a consumer who wishes to render a wrapped avatar for some reason (what we needed), or a completely different avatar, could use the renderer for this purpose.

in theory, this approach could be extended in different ways like most renderProps implementations. For instance, there could be additional parameters in `Comment` that are used in the render prop that specify avatar metrics, or the data used for the avatar, in cases where we want to render something entirely different.

PR #1076, opened 2019-04-11
feat(avatars): add loading states and new colors

alexkrolick:


boxcla:

src/elements/content-sidebar/activity-feed/__tests__/Avatar-test.js
```diff
@@ -9,66 +9,57 @@ describe('elements/content-sidebar/ActivityFeed/Avatar', () => {
...
-                .find('AvatarInitials')
-                .exists(),
-        ).toBe(false);
-        expect(getAvatarUrl).not.toBeCalledWith(user.id);
+    test('should pass isPending prop to wrapped Avatar getAvatarUrl prop is passed', () => {
+        const getAvatarUrl = jest.fn().mockResolvedValue('avatar.jpg');
+        const wrapper = getWrapper({ user, getAvatarUrl });
+        const wrappedAvatar = wrapper.find('Avatar').first();
+        expect(wrappedAvatar.props()).toEqual(expect.objectContaining({ isPending: true }));
+        expect(wrapper).toMatchInlineSnapshot(`
```
booc0mtaco: can we add these to the snapshots file since this component already had one?

alexkrolick: The file is empty/deleted now, the inline snaps replaced them

src/components/avatar/Avatar.js
```diff
@@ -55,21 +61,28 @@ class Avatar extends React.PureComponent<Props, State> {
...
         let avatar;
-        if (avatarUrl && !hasImageErrored) {
-            avatar = <AvatarImage onError={this.onImageError} url={avatarUrl} />;
-        } else if (name) {
-            avatar = <AvatarInitials id={id} name={name} />;
+        if (!isPending && avatarUrl && !hasImageErrored) {
+            avatar = <AvatarImage onError={this.onImageError} url={`${avatarUrl}`} />;
+        } else if (!isPending && name) {
+            avatar = <AvatarInitials name={name} />;
+        } else if (isPending) {
```
jeremypress: Could this go first and then we could avoid the !isPending for the other checks?

src/components/avatar/Avatar.js
```diff
@@ -55,21 +61,28 @@ class Avatar extends React.PureComponent<Props, State> {
...
+        const backgroundColor = hasNoAvatar || hasImageErrored ? initialsBackgroundColor : loadingBackgroundColor;
+
         let avatar;
-        if (avatarUrl && !hasImageErrored) {
-            avatar = <AvatarImage onError={this.onImageError} url={avatarUrl} />;
-        } else if (name) {
-            avatar = <AvatarInitials id={id} name={name} />;
+        if (!isPending && avatarUrl && !hasImageErrored) {
+            avatar = <AvatarImage onError={this.onImageError} url={`${avatarUrl}`} />;
+        } else if (!isPending && name) {
```
jeremypress: If name is required when we don't have an avatar URL, can we skip the pending state and go straight to name if avatarUrl doesn't exist?

src/components/avatar/Avatar.scss
```diff
@@ -22,27 +29,20 @@
         width: 100%;
     }
 
-    .avatar-icon,
-    .avatar-image {
-        @extend %circle;
+    img.avatar-image {
+        image-rendering: optimizeSpeed;
```
jeremypress: Deprecated according to MDN https://developer.mozilla.org/en-US/docs/Web/CSS/image-rendering#Browser_compatibility

src/components/avatar/Avatar.scss
```diff
@@ -4,14 +4,21 @@
...
+$size-default: 32px;
+$size-large: 44px;
+
 .avatar {
+    border-radius: 100%;
     display: inline-block;
-    height: 32px;
+    height: $size-default;
     position: relative;
+    transition: .3s;
```
ConradJChan: Is it possible to be more explicit about what properties are transitioned?

src/elements/content-sidebar/activity-feed/Avatar.js
```diff
@@ -14,46 +14,46 @@ type Props = {
 
 type State = {
     avatarUrl?: ?string,
+    isPending: boolean,
 };
 
+// Note: setting a key is required, re-render with a new user is not implemented
```
jeremypress: Which key are you referring to? I don't see one in this component.

src/components/avatar/AvatarImage.js
```diff
@@ -1,14 +1,16 @@
...
 
 type Props = {
     className?: string,
+    id?: string | number,
     onError?: Function,
     url: string,
 };
 
-const AvatarImage = ({ className = '', url, onError }: Props) => (
+const AvatarImage = ({ className = '', url, onError, ...rest }: Props) => (
```
ConradJChan: Would it make sense to destructure `id` explicitly?

src/components/avatar/Avatar.js
```diff
@@ -20,6 +22,10 @@ type Props = {
     className?: string,
     /** Users id */
     id?: string | number,
+    /** force loading state if avatarUrl is not passed
+     *  (useful if avatar url is determined asynchronously to avoid flashing background color)
+     */
+    isPending?: boolean,
```
priyajeet: Can this be achieved without isPending? By using key prop with url value? When URL is undefined/null, then always show the flashing thingy, and when url is there, then show the avatar.

src/components/avatar/Avatar.js
```diff
@@ -55,21 +61,28 @@ class Avatar extends React.PureComponent<Props, State> {
...
+        const initialsBackgroundColor = `${getColor(id)}`;
+        const backgroundColor = hasNoAvatar || hasImageErrored ? initialsBackgroundColor : loadingBackgroundColor;
+
         let avatar;
-        if (avatarUrl && !hasImageErrored) {
-            avatar = <AvatarImage onError={this.onImageError} url={avatarUrl} />;
-        } else if (name) {
-            avatar = <AvatarInitials id={id} name={name} />;
+        if (!isPending && avatarUrl && !hasImageErrored) {
+            avatar = <AvatarImage onError={this.onImageError} url={`${avatarUrl}`} />;
```
priyajeet: url={`${avatarUrl}`} ?
whats wrong with url={avatarUrl}

src/elements/content-sidebar/activity-feed/Avatar.js
alexkrolick: It needs a key when used, because it doesn't have a componentDidUpdate to handle refetching when the user identity changes `<Avatar key={user.id} />`

src/components/avatar/AvatarImage.js
alexkrolick: Woops, `id` should be removed from props, it was meant to be the user id, not the item `id`.

src/components/avatar/Avatar.js
alexkrolick: The existing code says it is "required" in a comment but it still handled the UnknownUser case (where there is no name or avatarUrl). I left it the same way.

src/components/avatar/Avatar.js
alexkrolick: The pending state is for when the URL has not been determined; the wrapper avatar does an async call to get an authenticated download URL.

src/elements/content-sidebar/activity-feed/Avatar.js
jeremypress: gotcha, thanks

src/components/avatar/Avatar.js
alexkrolick: One other idea is to merge the implementation of the element avatar (with `getAvatarUrl`) and the component (with `isPending`), since the `isPending` prop exists to support the state of the elements avatars that resolve async.

src/components/avatar/Avatar.js
```diff
@@ -55,21 +61,28 @@ class Avatar extends React.PureComponent<Props, State> {
...
         const classes = classNames(['avatar', className, { [`avatar--${size}`]: SIZES[size] }]);
 
+        const hasNoAvatar = !avatarUrl && !isPending;
+        const loadingBackgroundColor = lightGray;
+        const initialsBackgroundColor = `${getColor(id)}`;
+        const backgroundColor = hasNoAvatar || hasImageErrored ? initialsBackgroundColor : loadingBackgroundColor;
+
         let avatar;
-        if (avatarUrl && !hasImageErrored) {
+        if (isPending) {
```
jeremypress: Any thoughts on using the initials in pending case if they exist? This would prevent 2 visual states for the initials case

src/elements/content-sidebar/activity-feed/Avatar.js
```diff
@@ -14,46 +14,46 @@ type Props = {
...
 
     componentDidMount() {
         this.getAvatarUrl();
     }
 
+    async getAvatarUrl() {
+        const { user, getAvatarUrl } = this.props;
+        const { isPending } = this.state;
+        // don't fetch if url exists or no handler was passed
+        if (!isPending || !getAvatarUrl) return;
```
jeremypress: do you mind expanding this out with brackets? We don't really do this anywhere else

alexkrolick: Ok, but as a followup please create a lint rule for this if you care about it.

src/elements/content-sidebar/activity-feed/Avatar.js
```diff
@@ -14,46 +14,48 @@ type Props = {
...
-     *
-     * @param {string} avatarUrl the user avatar url
-     */
-    getAvatarUrlHandler = (avatarUrl: ?string) => {
-        this.setState({
+    constructor(props: Props) {
+        super(props);
+        // short-circuit promises by checking if we already have a url
+        // or if we don't have a method to get one
+        const avatarUrl = props.user.avatar_url || null;
```
jstoffan: Seems like this should be destructured or use `lodash/get`.

src/elements/content-sidebar/activity-feed/Avatar.js
```diff
@@ -14,46 +14,48 @@ type Props = {
...
 
+    async getAvatarUrl() {
+        const { user, getAvatarUrl } = this.props;
+        const { isPending } = this.state;
+        // don't fetch if url exists or no handler was passed
+        if (!isPending || !getAvatarUrl) {
+            return;
+        }
+
+        const avatarUrl = await getAvatarUrl(user.id).catch(() => null);
```
jstoffan: Can we use a promise here, since the async logic is fairly straightforward?

src/elements/content-sidebar/activity-feed/__tests__/Avatar-test.js
```diff
@@ -9,66 +9,57 @@ describe('elements/content-sidebar/ActivityFeed/Avatar', () => {
...
         const wrapper = getWrapper({ user, getAvatarUrl });
-        wrapper.instance().getAvatarUrlHandler('foo');
-        wrapper.update();
-        expect(wrapper).toMatchSnapshot();
+        expect(getAvatarUrl).toBeCalledWith(user.id);
+        // wait a tick for async, use react-dom/test-utils/act in future versions
+        await new Promise(r => setImmediate(r));
+        const wrappedAvatar = wrapper.find('Avatar').first();
+        expect(wrappedAvatar.props()).toEqual(expect.objectContaining({ isPending: false, avatarUrl: 'avatar.jpg' }));
+        expect(wrapper).toMatchInlineSnapshot(`
```
jstoffan: We don't really use these much, it looks like. Can this live in a snapshot file?

src/styles/_variables.scss
```diff
@@ -1,19 +1,23 @@
-// colors defined by BDL
+// Box Design Language Theme Colors
 $bdl-box-blue: #0061d5 !default;
```
jstoffan: Should these blues live under `// Box Design Langauge Blues`? 🎶

src/elements/content-sidebar/activity-feed/Avatar.js
alexkrolick: Like this? Is this to avoid some Babel overhead?

```js
getAvatarUrl(user.id)
 .catch(() => null)
 .then(avatarUrl => {
 	this.setState({ avatarUrl, isPending: false });
  })
```

src/elements/content-sidebar/activity-feed/Avatar.js
alexkrolick: `get` isn't typesafe, so I've been avoiding it. Destructuring looks like this:

```js
const { user: { avatar_url: avatarUrl = null } } = props
```

src/styles/_variables.scss
alexkrolick: Actually we should remove the "blues" section until it's decided whether the scale will have 3 or 5 colors

src/components/avatar/Avatar.js
```diff
@@ -55,21 +61,28 @@ class Avatar extends React.PureComponent<Props, State> {
     };
 
     render() {
-        const { avatarUrl, className, name, id, size = '' }: Props = this.props;
+        const { avatarUrl, className, name, id, size = '', isPending = false }: Props = this.props;
```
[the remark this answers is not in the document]
alexkrolick: Will followup for the size one since the `= ''` here is not the default prop value, it's just used for the object lookup. We can add a "medium" value for that and make it the default.

src/components/avatar/Avatar.js
jstoffan: @alexkrolick, can you address or respond to the outstanding comments?

alexkrolick: Working on it. Design signed off on not showing the initials for the loading state - which would entail putting the `` tag on top of the initials in any case.

src/elements/content-sidebar/activity-feed/Avatar.js
alexkrolick: Assuming this is because async/await might get compiled into generators. Changing.

jstoffan: Yep, it does. Yay IE! Thanks for updating.

src/components/avatar/Avatar.js
```diff
@@ -55,21 +61,28 @@ class Avatar extends React.PureComponent<Props, State> {
...
         const { hasImageErrored }: State = this.state;
         const classes = classNames(['avatar', className, { [`avatar--${size}`]: SIZES[size] }]);
 
+        const hasNoAvatar = !avatarUrl && !isPending;
```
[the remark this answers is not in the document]
alexkrolick: Resolving by inlining the color computation into the other if chain.

src/elements/content-sidebar/activity-feed/__tests__/Avatar-test.js
priyajeet: What was the reason of making it inline? It will be harder to diff if snapshots are inlined and large.

alexkrolick: Inline snaps are more likely to be read and understood. You shouldn't use them if the diff is large, but you shouldn't be using snapshots at all if they are _huge_.

src/elements/content-sidebar/activity-feed/Avatar.js
priyajeet: Whats wrong with const avatarUrl: string = get(props, 'user.avatar_url');

src/elements/content-sidebar/activity-feed/__tests__/Avatar-test.js
priyajeet: And why not? Even if the snapshot is 2x-3x of this one, it provides usefulness. Please move this out to its own file as thats what we have for other files and other projects.

alexkrolick: The Jest ESLint plugin has some good reasoning for these rules:

[no-large-snapshots](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/no-large-snapshots.md)

> When using Jest's snapshot capability one should be mindful of the size of created snapshots. As a best practice snapshots should be limited in size in order to be more manageable and reviewable. A stored snapshot is only as good as its review and as such keeping it short, sweet, and readable is important to allow for thorough reviews.

[prefer-inline-snapshots](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/prefer-inline-snapshots.md)

> In order to make snapshot tests more managable and reviewable toMatchInlineSnapshot() and toThrowErrorMatchingInlineSnapshot should be used to write the snapshots inline in the test file.

src/components/avatar/Avatar.js
priyajeet: Can you explain more "which would entail putting the  tag on top of the initials"
Like can't you just show initials by default and then make a request to get avatar. When that request comes or fails you replace initials with the image.

Whats the use case around "pending" state of avatars?

priyajeet: Is there a path when name is not there but pending true? In other words we cant even show initials?

PR #1077, opened 2019-04-11
test(access-stats): update mocks to avoid invariant errors in test

booc0mtaco:
This fixes snapshots where it referenced a non-existant component name and resolved to `undefined`


boxcla:

booc0mtaco:
Example:

```console
      Warning: React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.```

booc0mtaco:
One improvement will be to handle prop types passed in to the mocked i18n components for verification. This will require many more snapshot changes and some refactoring, so will break this out as a separate task

PR #1078, opened 2019-04-11
perf(avatar): memoize avatar url calls

alexkrolick:


boxcla:

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -526,11 +527,13 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
      * @param {string} fileId the file id
      * @return the user avatar URL string for a given user with access token attached
      */
-    getAvatarUrl = async (userId: string): Promise<?string> => {
-        const { file, api } = this.props;
+    getAvatarUrl = memoize<[string], Promise<?string>>(
```
jeremypress: Should we update the usersAPI to check the cache instead so that other callers of this endpoint don't have to implement memoization themselves?

alexkrolick: Sure, this is off the critical path for tasks so I'll file a ticket to follow up on it

PR #1082, opened 2019-04-11
build: Add npm registry & login checks

ConradJChan:


boxcla:

conf/cut_release.sh
```diff
@@ -187,6 +192,20 @@ check_branch_dirty() {
     fi
 }
 
+check_npm_registry() {
+    if [[ ! $(npm config get registry) =~ (${YARN_PUBLIC_REGISTRY_REGEX}|${NPM_PUBLIC_REGISTRY_REGEX}) ]] ; then
+        printf "${red}Not pointing at the right npm registry!${end}"
```
jeremypress: Maybe print a message to update .npmrc with the url we want them to use?

ConradJChan: Sounds good

booc0mtaco: Yes, let's add in a known step to address the issue in each case. Here, if they're pointed at the wrong place, they'll probably need to:

* move their current config and `npm login`
* or, do some custom commenting in their config file

I'd lean toward encouraging something like the former to avoid user error

conf/cut_release.sh
```diff
@@ -187,6 +192,20 @@ check_branch_dirty() {
...
+        return 1
+    fi
+}
+
+check_npm_login() {
+    if [[ ! $(npm whoami --registry ${NPM_PUBLIC_REGISTRY} 2>/dev/null) ]] ; then
+        printf "${red}Not logged into npm!${end}"
```
booc0mtaco: if we detect this state, could we prompt to login?

conf/cut_release.sh
```diff
@@ -187,6 +192,20 @@ check_branch_dirty() {
...
+check_npm_registry() {
+    if [[ ! $(npm config get registry) =~ (${YARN_PUBLIC_REGISTRY_REGEX}|${NPM_PUBLIC_REGISTRY_REGEX}) ]] ; then
+        printf "${red}Not pointing at the right npm registry! Make sure .npmrc points to ${NPM_PUBLIC_REGISTRY}${end}"
+        return 1
+    fi
+}
+
+check_npm_login() {
+    if [[ ! $(npm whoami --registry ${NPM_PUBLIC_REGISTRY} 2>/dev/null) ]] ; then
+        printf "${red}Not logged into npm! Try running npm login${end}"
```
booc0mtaco: Can we prompt to login if we reach this step? And read the return value to fail the release script if login doesn't succeed

conf/cut_release.sh
```diff
@@ -187,6 +192,20 @@ check_branch_dirty() {
     fi
 }
 
+check_npm_registry() {
+    if [[ ! $(npm config get registry) =~ (${YARN_PUBLIC_REGISTRY_REGEX}|${NPM_PUBLIC_REGISTRY_REGEX}) ]] ; then
+        printf "${red}Not pointing at the right npm registry! Make sure .npmrc points to ${NPM_PUBLIC_REGISTRY}${end}"
```
booc0mtaco: `~/.npmrc`?

nit: One thing we could do is move the rc file aside while running the release, and set it back once complete (or if it aborts).

conf/cut_release.sh
ConradJChan: From what I recall, making scripts interactive like this can be kind of a pain and error prone, but let me look into it, but if not, at least the corrective action is included in the error message

conf/cut_release.sh
ConradJChan: Thanks I'll update ~/.npmrc.

I'd prefer not to modify the file system. I think with the `.npmrc` & `.yarnrc` in source control this should be a corner case. But if people start running into this we can look into making that change.

conf/cut_release.sh
priyajeet: Should not add prompts to the script. If we decide to move it to Jenkins, then prompts will cause props

conf/cut_release.sh
alexkrolick: As far as I know this yarn issue is still outstanding, meaning global rc files merge _on top_ of local: https://github.com/yarnpkg/yarn/issues/5371

ConradJChan: Did some testing and it seems to be merging correctly for me with yarn 1.13.0, giving priority to the working directory rc files

alexkrolick: I'm not sure it is, they claim it is still open and will be fixed in the next major version as a breaking change

https://github.com/yarnpkg/yarn/issues/5371#issuecomment-483274723

PR #1084, opened 2019-04-11
feat(tasks): Support creation of general/approval tasks

mattnguyen1:


boxcla:

src/api/Feed.js
```diff
@@ -679,6 +679,7 @@ class Feed extends Base {
         currentUser: User,
         message: string,
         assignees: SelectorItems,
+        taskType: TaskType,
```
ConradJChan: Update the doc

src/api/Feed.js
```diff
@@ -749,13 +750,16 @@ class Feed extends Base {
...
-            task_type: 'GENERAL',
+            task_type: taskType,
             status: TASK_NEW_NOT_STARTED,
         };
 
         this.addPendingItem(this.id, currentUser, pendingTask);
 
-        const taskPayload = { name: message, due_at: dueAtString };
+        const taskPayload: TaskPayload = { name: message, due_at: dueAtString };
+        if (taskType) {
```
ConradJChan: According to the flow typing, it seems like `taskType` is required, is this if check needed?

src/elements/content-sidebar/AddTaskButton.js
```diff
@@ -25,17 +26,29 @@ type Props = {| ...AddTaskButtonProps, ...PassThroughProps |};
...
         isTaskFormOpen: false,
+        taskType: null,
     };
 
     static defaultProps = {
         isDisabled: false,
     };
 
+    getMessageForModalTitle(taskType: ?TaskType): Object {
+        switch (taskType) {
```
ConradJChan: Should `TASK_TYPE_GENERAL` be the default case if `taskType` is not provided?

mattnguyen1: the backend creates an approval task if `taskType` is not provided

src/api/Feed.js
mattnguyen1: will update doc to make it optional

alexkrolick: let's make it required - you should have to know what type of task you create

mattnguyen1: ah yeah that makes sense, I'll go ahead and make it required

src/elements/content-sidebar/AddTaskButton.js
alexkrolick: we can the type non-optional/non-nullable here too

PR #1085, opened 2019-04-11
feat(content-preview): show versions header when previewing old versions

jeremypress:
giphy

E2E tests to follow


boxcla:

src/elements/content-preview/ContentPreview.js
```diff
@@ -1110,21 +1122,25 @@ class ContentPreview extends PureComponent<Props, State> {
         if (!currentFileId) {
             return null;
         }
+
+        const onHeaderClose = isPreviewingCurrentVersion ? onClose : this.updateVersionToCurrent;
```
jeremypress: Added this here to avoid passing 2 callbacks to the header component.

src/elements/content-preview/preview-header/Header.js
```diff
@@ -0,0 +1,129 @@
+/**
```
jeremypress: This is mostly a copy/paste into the new folder.

src/elements/content-preview/preview-header/Header.js
```diff
@@ -0,0 +1,129 @@
...
+import PlainButton from '../../../components/plain-button/PlainButton';
+import IconClose from '../../../icons/general/IconClose';
+import IconDrawAnnotationMode from '../../../icons/annotations/IconDrawAnnotation';
+import IconPointAnnotation from '../../../icons/annotations/IconPointAnnotation';
+import IconPrint from '../../../icons/general/IconPrint';
+import IconDownload from '../../../icons/general/IconDownloadSolid';
+import AsyncLoad from '../../common/async-load';
+import messages from '../../common/messages';
+import { getIcon } from '../../common/item/iconCellRenderer';
+import { nines } from '../../../styles/variables';
```
jeremypress: Changed to pull from the global vars file instead of a more local constants file.

src/elements/content-preview/preview-header/PreviewHeaderComponent.js
```diff
@@ -0,0 +1,60 @@
...
+    canDownload,
+    contentOpenWithProps,
+    file,
+    isPreviewingCurrentVersion,
+    onClose,
+    onDownload,
+    onPrint,
+    token,
+}: Props) => {
+    const classes = classNames(['bcpr-header', !isPreviewingCurrentVersion && 'bcpr-header--dark']);
```
jeremypress: This could be removed if we push the outer divs into the specific headers, but would cause repetition.

src/elements/content-preview/preview-header/VersionsHeader.js
```diff
@@ -0,0 +1,48 @@
...
+    onClose?: Function,
+} & InjectIntlProvidedProps;
+
+const VersionsHeader = ({ file, intl, onClose }: Props) => {
+    const name = file ? file.name : '';
+    const closeMsg = intl.formatMessage(messages.close);
+
+    return (
+        <React.Fragment>
+            <div className="bcpr-name">
```
jeremypress: Will pull this out as it's shared with Header.js

src/elements/content-preview/preview-header/VersionsHeader.js
```diff
@@ -0,0 +1,48 @@
...
+                {file ? getIcon(24, file) : null}
+                <span>{name}</span>
+            </div>
+            <div className="bcpr-btns">
+                {onClose && (
```
jeremypress: Although this is currently the same as the normal preview header, I expect the icon to change, so I don't plan on pulling this out.

src/elements/content-preview/ContentPreview.js
```diff
@@ -31,7 +31,7 @@ import { withFeatureProvider } from '../common/feature-checking';
 import { EVENT_JS_READY } from '../common/logger/constants';
 import ReloadNotification from './ReloadNotification';
 import API from '../../api';
-import Header from './Header';
+import PreviewHeader from './preview-header';
```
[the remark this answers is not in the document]
jeremypress: It's coming from the index.js of the folder

src/elements/content-preview/preview-header/PreviewHeaderComponent.js
```diff
@@ -0,0 +1,60 @@
+/**
+ * @flow
+ * @file Preview header container component
```
jstoffan: This doesn't appear to be a container, since it doesn't connect to any apis/data sources and contains markup.

src/elements/content-preview/preview-header/PreviewHeaderComponent.js
```diff
@@ -0,0 +1,60 @@
...
+                        canDownload={canDownload}
+                        contentOpenWithProps={contentOpenWithProps}
+                        file={file}
+                        onClose={onClose}
+                        onDownload={onDownload}
+                        onPrint={onPrint}
+                        token={token}
+                    />
+                ) : (
+                    <VersionsHeader file={file} onClose={onClose} />
```
jstoffan: I would recommend combining `Header` and `VersionsHeader`. It seems like we basically just need to pass the `version` object to the `Header` component. If it's defined and doesn't match `file.file_version.id`, the component can disable the download/print buttons and apply the dark class.

src/elements/content-preview/preview-header/PreviewHeaderComponent.js
```diff
@@ -0,0 +1,60 @@
...
+    onPrint: Function,
+    token: ?string,
+} & InjectIntlProvidedProps;
+
+const PreviewHeaderComponent = ({
+    canAnnotate,
+    canDownload,
+    contentOpenWithProps,
+    file,
+    isPreviewingCurrentVersion,
```
jstoffan: Can be derived by passing in `version`, which will also be necessary to display the correct file name and type.

src/elements/content-preview/preview-header/PreviewHeader.js
```diff
@@ -42,73 +46,77 @@ const Header = ({
...
     token,
 }: Props) => {
-    const name = file ? file.name : '';
-    const id = file && file.id;
+    const fileId = file && file.id;
+    const shouldRenderOpenWith = fileId && contentOpenWithProps.show;
+    const currentVersionId = getProp(file, 'file_version.id');
+    const selectedVersionId = getProp(selectedVersion, 'id', currentVersionId);
+    const isPreviewingCurrentVersion = currentVersionId === selectedVersionId;
+    const displayItem = isPreviewingCurrentVersion ? file : selectedVersion;
```
jstoffan: Can this be simplified to `const displayedItem = selectedVersion || item;`?

src/elements/content-preview/preview-header/PreviewHeader.js
```diff
@@ -42,73 +46,77 @@ const Header = ({
...
+
     const closeMsg = intl.formatMessage(messages.close);
     const printMsg = intl.formatMessage(messages.print);
     const downloadMsg = intl.formatMessage(messages.download);
     const drawMsg = intl.formatMessage(messages.drawAnnotation);
     const pointMsg = intl.formatMessage(messages.pointAnnotation);
-    const shouldRenderOpenWith = id && contentOpenWithProps.show;
+
+    const classes = classNames(['bcpr-header', !isPreviewingCurrentVersion && 'bcpr-header--basic']);
```
jstoffan: It seems like the common convention for this is:

```
const className = classNames('bcpr-header', { 
    'bcpr-header--basic': !isPreviewingCurrentVersion 
});
```

src/elements/content-preview/ContentPreview.js
```diff
@@ -1051,10 +1055,14 @@ class ContentPreview extends PureComponent<Props, State> {
      * @param {string} [version] - The version that is now previewed
      * @param {object} [additionalVersionInfo] - extra info about the version
      */
-    onVersionChange = (version?: BoxItemVersion, additionalVersionInfo?: Object): void => {
+    onVersionChange = (version?: BoxItemVersion, additionalVersionInfo?: AdditionalVersionInfo): void => {
         const { onVersionChange }: Props = this.props;
+        this.updateVersionToCurrent = getProp(additionalVersionInfo, 'updateVersionToCurrent');
```
jstoffan: Is `getProp` needed here or can we just default the value in the function signature to an empty object?

src/elements/content-preview/ContentPreview.js
```diff
@@ -1110,21 +1119,28 @@ class ContentPreview extends PureComponent<Props, State> {
         if (!currentFileId) {
             return null;
         }
+
+        const currentVersionId = getProp(file, 'file_version.id');
+        const selectedVersionId = getProp(selectedVersion, 'id', currentVersionId);
+
+        const onHeaderClose = currentVersionId === selectedVersionId ? onClose : this.updateVersionToCurrent;
```
jstoffan: You may want to add a check to see if `this.updateVersionToCurrent` is defined.

jeremypress: I think this actually follows expected behavior which is that if its undefined the header won't render a close button.

src/elements/content-preview/preview-header/FileInfo.js
```diff
@@ -9,13 +9,17 @@ import FileIcon from '../../../icons/file-icon/FileIcon';
...
 };
 
 const FileInfo = ({ item }: Props) => (
     <div className="bcpr-name">
-        {item ? <FileIcon dimension={24} extension={item.extension} /> : null}
-        <span>{item.name}</span>
+        {item && (
+            <React.Fragment>
+                <FileIcon dimension={24} extension={item.extension} />
+                <span> {item.name} </span>
```
jstoffan: Are the spaces inside the span needed? If they're being used for spacing, it's probably better to do it via CSS.

src/elements/content-preview/preview-header/PreviewHeader.js
```diff
@@ -54,12 +54,11 @@ const PreviewHeader = ({
     const currentVersionId = getProp(file, 'file_version.id');
     const selectedVersionId = getProp(selectedVersion, 'id', currentVersionId);
     const isPreviewingCurrentVersion = currentVersionId === selectedVersionId;
-    const displayItem = isPreviewingCurrentVersion ? file : selectedVersion;
+    const displayItem = selectedVersion || file;
```
jstoffan: Seems like this could live inline with the `FileInfo` component, if you want.

src/elements/content-preview/preview-header/PreviewHeader.js
```diff
@@ -124,17 +123,22 @@ const PreviewHeader = ({
...
+                                aria-label={closeMsg}
+                                className="bcpr-btn"
+                                onClick={onClose}
+                                title={closeMsg}
+                                type="button"
+                            >
+                                <IconClose color={nines} height={24} width={24} />
+                            </PlainButton>
+                        ) : (
+                            <PlainButton className="bcpr-btn" onClick={onClose} title={backMsg} type="button">
```
jstoffan: I don't think `title` is needed here.

PR #1088, opened 2019-04-12
fix(query-bar): user loses name column via column dropdown

ChenCodes:
Changes in this PR:
- Pass down `nameColumn` from QueryBar
- When `onColumnChange` is called in ColumnButtonOverlay, insert `nameColumn` back into the columns array at index 0
Screen Shot 2019-04-12 at 10 55 44 AM


boxcla:

ChenCodes:
Closing this PR as the fix should be done in the reducer in the parent app.

PR #1089, opened 2019-04-12
fix(query-bar): blank page shown on clicking x button in date input

ChenCodes:
Changes in this PR:
- Check to see if `selectedValues` is an array of a null element, if so, return undefined
^The case above is due to the user clicking on the X button when deleting the Datepicker input

### DIFF
--- a/src/features/query-bar/components/filter/ValueField.js
+++ b/src/features/query-bar/components/filter/ValueField.js
@@ -20,11 +20,7 @@ type Props = {
 };
 
 const getDateValue = selectedValues => {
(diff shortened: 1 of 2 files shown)


boxcla:

src/features/query-bar/components/filter/ValueField.js
```diff
@@ -24,6 +24,10 @@ const getDateValue = selectedValues => {
         return undefined;
     }
 
+    if (selectedValues[0] === null) {
```
krapspark: let's include this into the check in line 23

PR #1095, opened 2019-04-15
fix(query-bar): persist changes after closing flyout

ChenCodes:
Changes in this PR:
- handle two distinct cases:
   1. whether or not the initial condition should be set
   2. whether or not we want to reinitialize conditions from the parent app

ezgif com-video-to-gif (5)

### DIFF
--- a/src/features/query-bar/__tests__/FilterButton-test.js
+++ b/src/features/query-bar/__tests__/FilterButton-test.js
@@ -76,7 +76,7 @@ describe('feature/query-bar/components/filter/FilterButton', () => {
 
         test.each`
             innerColumns | conditions         | transientConditions | expectedConditions    | should
-            ${columns}   | ${validConditions} | ${[]}               | ${validConditions}    | ${'should reinitialize conditions from props.conditions when flyout is opened and transientConditions is not empty'}
+            ${columns}   | ${validConditions} | ${[]}               | ${validConditions}    | ${'should reinitialize conditions from props.conditions when flyout is opened and props.conditions is not empty'}
             ${columns}   | ${[]}              | ${[]}               | ${[initialCondition]} | ${'should set initial condition if props.conditions is empty and transientConditions are empty'}
             ${[]}        | ${[]}              | ${[]}               | ${[]}                 | ${'should set empty array if props.conditions is empty and transientConditions are empty and columns are empty'}
         `('$should', ({ innerColumns, conditions, transientConditions, expectedConditions }) => {
--- a/src/features/query-bar/__tests__/FilterButton-test.js
+++ b/src/features/query-bar/__tests__/FilterButton-test.js
@@ -69,19 +69,21 @@ describe('feature/query-bar/components/filter/FilterButton', () => {
     describe('componentDidUpdate()', () => {
         const initialCondition = {
             columnId: columnForTemplateFieldName.id,
-            id: '2',
+            id: '1',
             operator: '=',
             values: [],
         };
+
         test.each`
-            innerColumns | conditions         | expectedConditions    | should
-            ${columns}   | ${validConditions} | ${validConditions}    | ${'should reinitialize conditions from props.conditions when flyout is opened and props.conditions is not empty'}
-            ${columns}   | ${[]}              | ${[initialCondition]} | ${'should set to initial condition when flyout is opened and props.conditions is empty'}
-            ${[]}        | ${[]}              | ${[]}                 | ${'should set to empty array when flyout is opened and both props.columns and props.conditions is empty'}
-        `('$should', ({ innerColumns, conditions, expectedConditions }) => {
+            innerColumns | conditions         | transientConditions | expectedConditions    | should
+            ${columns}   | ${validConditions} | ${[]}               | ${validConditions}    | ${'should reinitialize conditions from props.conditions when flyout is opened and transientConditions is not empty'}
+            ${columns}   | ${[]}              | ${[]}               | ${[initialCondition]} | ${'should set initial condition if props.conditions is empty and transientConditions are empty'}
+            ${[]}        | ${[]}              | ${[]}               | ${[]}                 | ${'should set empty array if props.conditions is empty and transientConditions are empty and columns are empty'}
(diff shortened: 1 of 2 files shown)


boxcla:

src/features/query-bar/components/filter/FilterButton.js
```diff
@@ -51,17 +51,21 @@ class FilterButton extends React.Component<Props, State> {
...
         const { columns } = this.props;
-        const { isMenuOpen } = this.state;
+        const { isMenuOpen, transientConditions } = this.state;
         const { isMenuOpen: prevIsMenuOpen } = prevState;
         const wasFlyoutOpened = isMenuOpen && !prevIsMenuOpen;
         if (wasFlyoutOpened) {
             const areConditionsEmpty = this.props.conditions.length === 0;
-            if (areConditionsEmpty) {
-                const transientConditions = columns && columns.length === 0 ? [] : [this.createCondition()];
+            const isInitialConditionSet = transientConditions.length > 0;
```
Visylvius: This seems like it could be relevant information needed by the component at a later date, 
instead of defining it in a variable inside a lifecycle method, 
would it make more sense to define it in state?

ChenCodes: It looks like the only usage in <FilterButton /> where the length is being checked is on L309 so it doesn't make too much sense to set it in state. Could you elaborate more on why this information would be relevant at a later date?

src/features/query-bar/__tests__/FilterButton-test.js
```diff
@@ -69,19 +69,21 @@ describe('feature/query-bar/components/filter/FilterButton', () => {
...
         };
+
         test.each`
-            innerColumns | conditions         | expectedConditions    | should
-            ${columns}   | ${validConditions} | ${validConditions}    | ${'should reinitialize conditions from props.conditions when flyout is opened and props.conditions is not empty'}
-            ${columns}   | ${[]}              | ${[initialCondition]} | ${'should set to initial condition when flyout is opened and props.conditions is empty'}
-            ${[]}        | ${[]}              | ${[]}                 | ${'should set to empty array when flyout is opened and both props.columns and props.conditions is empty'}
-        `('$should', ({ innerColumns, conditions, expectedConditions }) => {
+            innerColumns | conditions         | transientConditions | expectedConditions    | should
+            ${columns}   | ${validConditions} | ${[]}               | ${validConditions}    | ${'should reinitialize conditions from props.conditions when flyout is opened and transientConditions is not empty'}
```
krapspark: This test case looks like `transientConditions` is empty. Do you mean `conditions` is not empty?

src/features/query-bar/components/filter/FilterButton.js
```diff
@@ -51,17 +51,21 @@ class FilterButton extends React.Component<Props, State> {
...
+            const shouldSetInitialCondition = areConditionsEmpty && !isInitialConditionSet;
+            const shouldReinitializeConditions = !isInitialConditionSet;
+
+            if (shouldSetInitialCondition) {
```
krapspark: The logic here seems to be laid out in a funky way.
Because it essentially check this:
`if (a && b) {
   ...
  } else (!b) {
   ...
  }
`
which feels odd.
Can we do this instead:
`if (b) {
  if (a) {
    ...
  } else {
    ...
  }
}`

src/features/query-bar/__tests__/FilterButton-test.js
ChenCodes: Thanks for spotting this, it should be `should reinitialize conditions from props.conditions when flyout is opened and transientConditions is empty`

src/features/query-bar/components/filter/FilterButton.js
```diff
@@ -50,21 +50,25 @@ class FilterButton extends React.Component<Props, State> {
...
-                    transientConditions,
-                });
-            } else {
-                this.setState({
-                    transientConditions: cloneDeep(this.props.conditions),
-                });
+            const isInitialConditionSet = transientConditions.length > 0;
+            const shouldSetInitialCondition = conditions.length === 0;
+
+            if (!isInitialConditionSet) {
```
krapspark: now that the two conditions are back to back, it's hard to understand what each one signifies.
Can we rename the variables
`isInitialConditionSet` -> `isTransientConditionSet`
`shouldSetInitialCondition` -> `hasDefaultConditions`

We can totally talk through other options, too.

src/features/query-bar/components/filter/FilterButton.js
Visylvius: There are there cases where this value is being used. 
isInitialConditionSet, shouldSetInitialCondition, shouldReinitializeConidtion, 
Your saying that there will never be another case that these checks won't be used?

ChenCodes: So after the logic cleanup it's only being used in one spot - L61.

src/features/query-bar/__tests__/FilterButton-test.js
```diff
@@ -69,19 +69,21 @@ describe('feature/query-bar/components/filter/FilterButton', () => {
...
-            ${columns}   | ${[]}              | ${[initialCondition]} | ${'should set to initial condition when flyout is opened and props.conditions is empty'}
-            ${[]}        | ${[]}              | ${[]}                 | ${'should set to empty array when flyout is opened and both props.columns and props.conditions is empty'}
-        `('$should', ({ innerColumns, conditions, expectedConditions }) => {
+            innerColumns | conditions         | transientConditions | expectedConditions    | should
+            ${columns}   | ${validConditions} | ${[]}               | ${validConditions}    | ${'should reinitialize conditions from props.conditions when flyout is opened and props.conditions is not empty'}
+            ${columns}   | ${[]}              | ${[]}               | ${[initialCondition]} | ${'should set initial condition if props.conditions is empty and transientConditions are empty'}
+            ${[]}        | ${[]}              | ${[]}               | ${[]}                 | ${'should set empty array if props.conditions is empty and transientConditions are empty and columns are empty'}
```
macrotim: We're missing a unit test.

PR #1097, opened 2019-04-15
fix(sidebar): Details sidebar render no longer collapses on load

jstoffan:
Returning `null` in render causes the Details sidebar to momentarily collapse, which is causing issues with the sidebar navigation tooltips.


boxcla:

PR #1100, opened 2019-04-15
fix(list-view): use smaller sort icon

ChenCodes:
Changes in this PR:
- Use 11px for height and width instead of 17px to maintain consistency with parent app
Screen Shot 2019-04-15 at 4 20 19 PM


boxcla:

PR #1101, opened 2019-04-16
feat(tasks): Add in-modal error handling for task creation

mattnguyen1:


boxcla:

mattnguyen1:
Screen Shot 2019-04-15 at 6 09 59 PM

i18n/en-US.properties
```diff
@@ -361,7 +361,9 @@ be.statusSkill = Status
 # Approve option for a task
 be.taskApprove = Complete
 # Error message when a task creation fails
-be.taskCreateErrornMessage = There was an error creating this task.
+be.taskCreateErrornMessage = An error occurred while creating this task. Please try again.
```
alexkrolick: can we fix this typo while we're adding in here the title message?

```diff
- be.taskCreateErrornMessage
+ be.taskCreateErrorMessage
```

src/elements/content-sidebar/AddTaskButton.js
```diff
@@ -51,13 +53,15 @@ class AddTaskButton extends React.Component<Props, State> {
 
     handleClickMenuItem = (taskType: TaskType) => this.setState({ isTaskFormOpen: true, taskType });
 
-    handleModalClose = () => this.setState({ isTaskFormOpen: false });
+    handleModalClose = () => this.setState({ isTaskFormOpen: false, error: null });
 
-    handleSubmit = () => this.setState({ isTaskFormOpen: false });
+    handleCreateSuccess = () => this.setState({ isTaskFormOpen: false });
```
jeremypress: Do we also need to clear the error state here if the first task fails but the second one succeeds?

mattnguyen1: yeah good catch

test/integration/content-sidebar/TaskForm.e2e.test.js
```diff
@@ -48,19 +49,8 @@ describe('Create Task', () => {
...
-            // but the card temporarily appears
-            cy.getByTestId('activityfeed').within(() => {
-                cy.getByTestId('task-card')
-                    .last()
-                    .within(() => {
-                        cy.contains('valid e2e task').should('exist');
-                    });
-            });
+            // submit button should be in loading state
+            getSubmitButton().should('have.class', 'is-loading');
```
alexkrolick: Can we add an e2e test for the loading state as the deleted portion of the test did? can rename the test flow to account for the fact that we aren't going to get a write token for e2e tests any time soon: `it(shows error state after receiving server error)`

mattnguyen1: Since the loading state happens in the modal now to account for showing an error in the modal, I took out the pending task in the activity feed, so that deleted part should stay deleted. I think we can still test the task showing up in the feed if we stub out the creation response to return some sample task data and verify that the sample task shows up in the feed.

alexkrolick: That would be good, but we can also test for the error showing up right now because the server is going to error with a read token.

PR #1102, opened 2019-04-16
fix(preview): Preview now expands to fill its flex parent when possible

jstoffan:
This fixes an issue in the beta versions of Chrome 74 and 75 where Preview is not shown on the shared files page. Using `height: 100%` inside a flex parent without its own height defined can be very flaky.


boxcla:

PR #1063, opened 2019-04-09
chore(avatars): Change avatar initials to use approved colors

patrickpaul:
Per Box Design, the following changes were requested to `<AvatarInitials>`:

Force initials to be uppercase
Change font-weight from default (400) to `bold`
Change colors to use the approved list:
Avatar Replacement Colors

PR #1103, opened 2019-04-16
fix(avatars): use new font-weight and colors

alexkrolick:
- updates avatars to match new theme from design
- update of #1063 to use scss variables

Screen Shot 2019-04-16 at 2 48 58 PM


boxcla:

PR #1108, opened 2019-04-17
fix(query-bar): User should see error invalid inputs number float field

ChenCodes:
Changes in this PR:
- Add a new util method (`isFloat` )
- Show a different error message when an input is invalid for a float/number field 
- Add a column with number type into fixtures

Screen Shot 2019-04-17 at 1 27 41 PM
Screen Shot 2019-04-17 at 1 27 34 PM
Screen Shot 2019-04-17 at 1 27 29 PM
Screen Shot 2019-04-17 at 1 27 22 PM
Screen Shot 2019-04-17 at 1 27 14 PM


boxcla:

ChenCodes:
I have also checked the functionality for `FloatField` and the behavior is the same with the usage of my new float validator.

src/features/query-bar/validator.js
```diff
@@ -1,15 +1,17 @@
 // @flow
 import { FLOAT } from './constants';
 
-const floatRegex = /[^$,.\d]/;
+const isFloat = (value: string) => {
+    return !/^\s*$/.test(value) && !Number.isNaN(Number(value));
```
patrickpaul: Number.isNaN is not IE11 compatible.

src/features/query-bar/validator.js
```diff
@@ -1,15 +1,19 @@
...
 
-const floatRegex = /[^$,.\d]/;
+const isFloat = (value: string) => {
+    return !/^\s*$/.test(value) && !isNaN(Number(value));
+};
 
-const floatValidator = (value: string) => {
-    return value.match(floatRegex);
+const isInt = (value: string) => {
+    return /^\d+$/.test(value);
```
krapspark: you should check `isNaN` here as well

src/features/query-bar/validator.js
```diff
@@ -1,15 +1,19 @@
 // @flow
+import { isNaN } from 'lodash';
+
 import { FLOAT } from './constants';
 
-const floatRegex = /[^$,.\d]/;
+const isFloat = (value: string) => {
+    return !/^\s*$/.test(value) && !isNaN(Number(value));
```
krapspark: Is there a reason you aren't using the `parseFloat` function that's native to the browser?
Performance is probably similar but seems weird to implement your own regex.

src/features/query-bar/__tests__/validator-test.js
```diff
@@ -26,3 +26,33 @@ import { isValidValue } from '../validator';
...
+        const result = isFloat(value);
+        expect(result).toBe(expected);
+    });
+});
+
+describe('isInt()', () => {
+    test.each`
+        value     | expected | should
+        ${'1.0.'} | ${false} | ${'Should return false if the value is 1.0.'}
+        ${'1.0'}  | ${false} | ${'Should return false if the value is 1.0'}
```
krapspark: I'm not sure if this should be false.
in browser's native `isInteger`, 1.0 counts as an integer

src/features/query-bar/components/filter/Condition.js
```diff
@@ -109,22 +110,46 @@ const Condition = ({
         return [];
     };
 
+    const checkValidValue = (values: Array<ConditionValueType>, type: string) => {
+        let isValueInvalid = false;
```
krapspark: It seems weird that we're returning `isValueInvalid`.
With function named `checkValidValue`, I think the natural assumption will be that it returns true if the value is valid.

src/features/query-bar/components/filter/Condition.js
```diff
@@ -109,22 +110,46 @@ const Condition = ({
         return [];
     };
 
+    const checkValidValue = (values: Array<ConditionValueType>, type: string) => {
```
krapspark: nit: could be called `validateValue`

src/features/query-bar/components/filter/Condition.js
```diff
@@ -138,8 +163,7 @@ const Condition = ({
             default:
                 break;
         }
-
-        const error = areErrorsEnabled && !isValueSet ? message : null;
+        const error = isValueInvalid || (areErrorsEnabled && !isValueSet) ? message : null;
```
krapspark: I think you should return null in the beginning if there are no errors.

src/features/query-bar/messages.js
```diff
@@ -7,6 +7,16 @@ const messages = defineMessages({
         description: 'Text displayed on the Tooltip for an input field',
         id: 'boxui.queryBar.tooltipEnterValueError',
     },
+    tooltipInvalidFloatError: {
+        defaultMessage: 'Please Enter a Float',
```
krapspark: Not sure if "Float" is a useful message for all of our users. Maybe "Decimal number"?

src/features/query-bar/validator.js
ChenCodes: I can use parseFloat! I wasn't aware of that library method.

ChenCodes: Actually I can stick with my regex expression. It looks like when I use parseFloat('1.0.') this returns 1 when it should instead return NaN.

src/features/query-bar/__tests__/validator-test.js
ChenCodes: Good catch. It should be true; I will be using lodash's isInteger to make that test case pass.

src/features/query-bar/components/filter/Condition.js
ChenCodes: Do you mean to do:
```
const error = isValueInvalid || (areErrorsEnabled && !isValueSet) ? message : null;
        if (!error) {
            return null;
        }
        return error;
```

src/features/query-bar/components/filter/Condition.js
```diff
@@ -109,22 +110,43 @@ const Condition = ({
         return [];
     };
 
+    const validateValue = (values: Array<ConditionValueType>, type: string) => {
+        switch (type) {
+            case NUMBER:
+                return !isInt(String(values[0]));
```
krapspark: Isn't this still returning `true` if invalid and `false` if valid?

src/features/query-bar/components/filter/Condition.js
krapspark: I mean you can have something like this in line 136:
`      if (!isValueInvalid && (!areErrorsEnabled || isValueSet)) { return null; }`

This way, you don't have to do the switch statement if there are no errors.

Come to think of it, shouldn't you just return `null` if `areErrorsEnabled` is `false`? haha
Also, that should probably have been named `isErrorEnabled`.
We can take some time to talk through the logic here

src/features/query-bar/components/filter/Condition.js
ChenCodes: I think the naming/logic got kind of confusing. Here's the new setup:

```
const validateValue = (values: Array<ConditionValueType>, type: string) => {
        switch (type) {
            case NUMBER:
                return isInt(String(values[0]));
            case FLOAT:
                return isFloat(String(values[0]));
            default:
                break;
        }
        return true;
    };
```

validateValue will return true if the value is valid. It will return false if the value is invalid. isInt will return true if the string is an integer.

In getErrorMessage, I will have the variable be defined as `'isValueValid'` and have that be assigned to `validateValue(values, type)`.

ChenCodes:
For the comment about returning null if areErrorsEnabled is false, this wouldn't be correct because the name is a little misleading; areErrorsEnabled is meant to determine whether or not the user has already pressed on the Apply button. When the user presses on the Apply button then this boolean will get set to true, meaning that we should now show any "empty field" errors.

src/features/query-bar/components/filter/Condition.js
```diff
@@ -109,22 +110,54 @@ const Condition = ({
...
+
     const getErrorMessage = () => {
         const { values, columnId } = condition;
         const column = columns && columns.find(c => c.id === columnId);
         const type = column && column.type;
 
         const isValueSet = values.length !== 0;
+
+        let isValueValid = true;
+        if (isValueSet && type) {
```
krapspark: Now that `isValueValid` defaults to `true`, you need to set it to `false` if `isValueSet && type` is false. Alternatively, you can set `isValueValid` default value to `false`

src/features/query-bar/components/filter/Condition.js
```diff
@@ -139,9 +172,7 @@ const Condition = ({
                 break;
         }
 
-        const error = areErrorsEnabled && !isValueSet ? message : null;
-
-        return error;
+        return message;
```
krapspark: it actually looks like you can have each switch case set the `messageText` and then return `<FormattedMessage {...messageText}/>` in line 175

src/features/query-bar/validator.js
```diff
@@ -1,15 +1,19 @@
...
+};
 
-const floatValidator = (value: string) => {
-    return value.match(floatRegex);
+const isInt = (value: string) => {
```
patrickpaul: Why is this function necessary? _.isInteger

src/features/query-bar/validator.js
```diff
@@ -1,15 +1,19 @@
 // @flow
+import { isInteger, isNaN } from 'lodash';
```
jstoffan: Please do not import the root lodash module directly, as it's very likely to increase our bundle size. You can import single functions via `import isInteger from 'lodash/isinteger';`

src/features/query-bar/validator.js
ChenCodes: Will remove this method

src/features/query-bar/validator.js
```diff
@@ -1,15 +1,19 @@
...
+
 import { FLOAT } from './constants';
 
-const floatRegex = /[^$,.\d]/;
+const isFloat = (value: string) => {
```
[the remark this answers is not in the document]
ChenCodes: Yep I can move it to utils.js under src/common

src/features/query-bar/validator.js
jstoffan: I agree. This doesn't seem to justify a separate utility.

src/features/query-bar/validator.js
jstoffan: It very much seems like `lodash/isFixed` should be a separate check for all of these validations rather than having them baked in.

PR #1110, opened 2019-04-17
feat(tasks): Handle errors when performing actions on tasks

mattnguyen1:


boxcla:

mattnguyen1:
Screen Shot 2019-04-17 at 4 25 49 PM

src/elements/content-sidebar/activity-feed/task-new/Task.js
```diff
@@ -136,68 +137,74 @@ const Task = ({
...
-                getUserProfileUrl={getUserProfileUrl}
-                mentionSelectorContacts={mentionSelectorContacts}
-                getMentionWithQuery={getMentionWithQuery}
-                userHeadlineRenderer={userLinkInstance => {
-                    return (
-                        <FormattedMessage
-                            {...getMessageForTask(!!currentUserAssignment, task_type)}
-                            values={{ user: userLinkInstance }}
+        <div className="bcs-task-container">
+            {error ? <CommentInlineError {...error} /> : null}
```
Visylvius: instead of returning null, 
could you just use 
`` error && <CommentInlineError {...error} />```?

i18n/en-US.properties
```diff
@@ -358,8 +358,14 @@ be.skillUnknownError = Something went wrong with running this skill or fetching
 be.sort = Sort
 # Label for status skill card in the preview sidebar
 be.statusSkill = Status
+# Title shown when an error occurs performing an action on a task
+be.taskActionErrorTitle = Error
 # Approve option for a task
 be.taskApprove = Complete
+# Error message when approving a task fails
+be.taskApproveErrorMessage = An error has occurred while approving this task. Please refresh the page and try again.
```
jeremypress: Is there a chance that refreshing would help fix the issue more than just retrying would?

booc0mtaco: from what I understand, there are some possible edge cases where the UI preemptively updates and can lead to an undefined state, so during beta we take the conservative approach and suggest a refresh. 

- future implementation could avoid the preemptive updates entirely, and add an interface to do safe retries

jeremypress: sounds good!

PR #1111, opened 2019-04-18
fix(icons): update general and task approval icons

booc0mtaco:
Update icons to have no outline, and to fix the type used when importing them (flow)


boxcla:

src/icons/flowTypes.js
```diff
@@ -1,17 +1,19 @@
 // @flow
+import type { Element } from 'react';
```
alexkrolick: This should be ReactElement I believe

alexkrolick: https://flow.org/en/docs/react/types/ maybe not! closing.

booc0mtaco:
@jeremypress this class of icon will have fixed colors based on designs, and can be overridden by the client using the mixins in `two-toned/`. These were already in place, so they don't show up in the review

https://github.com/box/box-ui-elements/blob/2f6a1bb4cd0a907ed4b032137d939d9fb9fb4101/src/icons/two-toned/_IconTaskApproval.mixins.scss

As an example, to override in the client, one would add a custom `className` prop for the icons, then use the mixin with that class in the Sass to specify the static overrides.

src/icons/two-toned/IconTaskGeneral.js
```diff
@@ -3,24 +3,17 @@ import * as React from 'react';
...
-                </g>
-            </g>
+    <AccessibleSVG className={className} height={height} title={title} viewBox="0 0 32 32" width={width}>
```
priyajeet: no default classnames?

booc0mtaco: yeah this was missed on the original review. I was in the midsts of adding them right now!

src/icons/flowTypes.js
```diff
@@ -1,17 +1,19 @@
...
+
 export type Icon = {
     className?: string,
     color?: string,
     height?: number,
-    /** A text-only string describing the icon if it's not purely decorative for accessibility */
-    title?: string | React.Element<any>,
+    /** A string describing the icon if it's not purely decorative for accessibility */
+    title?: string | Element<any>,
```
priyajeet: btw I think we can make this all React.Node now.
Though not even sure what title element translates to though, maybe `<title>` is fine having elements in it for accessibility

booc0mtaco: yah, we inherited this type from existing icons, so I think we can lock down the types more, encouraging people to use `intl.formatMessage` or some other method to insert the text.

Checked yesterday, and title is allowed to have elements in : https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title#Usage_notes

PR #1112, opened 2019-04-18
docs(styleguide): expand props and methods by default

alexkrolick:
This changes the styleguide to expand the props panel by default, so you don't have to click into it.


boxcla:

jeremypress:
Makes sense for smaller components but for top level elements that have a lot of props the demo gets pushed way down. I'm looking to see if we could move "demo" above props & methods, which are only used by the top level elements.

jstoffan:
I don't agree with this change as a global default, as it has a negative effect on the top-level Elements. If this is something that will make your personal development experience more efficient, you're welcome to make the change locally.

PR #1116, opened 2019-04-18
feat(tasks): show full task assignee list in dropdown

alexkrolick:
- This currently renders the full assignee list in a dropdown
  when the +N icon next to the first 3 assignment avatars is clicked
- A followup commit will add pagination via infinite scroll

### DIFF
--- a/src/elements/content-sidebar/activity-feed/task-new/AssigneeStatus.js
+++ b/src/elements/content-sidebar/activity-feed/task-new/AssigneeStatus.js
@@ -27,8 +27,8 @@ const StatusIcon = ({ status, ...rest }: { status: TaskCollabStatus }) => {
     }
 };
 
-const AssignmentStatus = React.memo<Props>(({ user, status, getAvatarUrl, ...rest }: Props) => (
-    <div className="bcs-task-assignment-status" {...rest}>
+const AssignmentStatus = React.memo<Props>(({ user, status, getAvatarUrl, className, ...rest }: Props) => (
+    <div className={`bcs-task-assignment-status ${className}`} {...rest}>
         <Avatar className="bcs-task-assignment-avatar" user={user} getAvatarUrl={getAvatarUrl} />
         <StatusIcon
             status={status}
--- a/src/elements/content-sidebar/activity-feed/task-new/AssigneeStatus.js
+++ b/src/elements/content-sidebar/activity-feed/task-new/AssigneeStatus.js
@@ -1,6 +1,7 @@
 // @flow
 import * as React from 'react';
 import { FormattedMessage } from 'react-intl';
+import classNames from 'classnames';
 import camelCase from 'lodash/camelCase';
 import IconComplete from '../../../../icons/general/IconVerified';
 import IconReject from '../../../../icons/general/IconRejected';
@@ -9,6 +10,7 @@ import { TASK_NEW_APPROVED, TASK_NEW_REJECTED, TASK_NEW_COMPLETED, TASK_NEW_NOT_
 import messages from '../../../common/messages';
 
 type Props = {
+    className?: string,
     getAvatarUrl?: GetAvatarUrlCallback,
     status: TaskCollabStatus,
     user: UserMini,
@@ -28,7 +30,7 @@ const StatusIcon = ({ status, ...rest }: { status: TaskCollabStatus }) => {
 };
 
 const AssignmentStatus = React.memo<Props>(({ user, status, getAvatarUrl, className, ...rest }: Props) => (
-    <div className={`bcs-task-assignment-status ${className}`} {...rest}>
+    <div className={classNames('bcs-task-assignment-status', className)} {...rest}>
         <Avatar className="bcs-task-assignment-avatar" user={user} getAvatarUrl={getAvatarUrl} />
         <StatusIcon
             status={status}
--- a/src/elements/content-sidebar/activity-feed/task-new/Assignees.js
+++ b/src/elements/content-sidebar/activity-feed/task-new/Assignees.js
@@ -83,7 +83,8 @@ class Assignees extends React.Component<Props> {
                     {visibleAssignees}
                     {hiddenAssigneeCount > 0 && (
                         <Flyout position="top-left">
-                            <span
+                            <button
+                                type="button"
                                 className="bcs-task-assignment-count-container"
                                 data-testid="task-assignment-overflow"
                             >
@@ -92,7 +93,7 @@ class Assignees extends React.Component<Props> {
                                         {overflowLabel}
                                     </span>
                                 </Tooltip>
-                            </span>
+                            </button>
                             <Overlay>
                                 <div className="bcs-task-assignment-list">
                                     <h5 className="bcs-task-assignment-list-title">
--- a/src/elements/content-sidebar/activity-feed/task-new/Assignees.js
+++ b/src/elements/content-sidebar/activity-feed/task-new/Assignees.js
@@ -1,0 +1,112 @@
+// @flow strict
+import * as React from 'react';
+import { injectIntl, FormattedMessage } from 'react-intl';
+import type { InjectIntlProvidedProps } from 'react-intl';
+import { Flyout, Overlay } from '../../../../components/flyout';
+import Tooltip from '../../../../components/tooltip';
+import messages from '../../../common/messages';
+import AssigneeStatus from './AssigneeStatus';
+import { TASK_NEW_APPROVED, TASK_NEW_REJECTED, TASK_NEW_COMPLETED, TASK_NEW_NOT_STARTED } from '../../../../constants';
+
+const MAX_AVATARS = 3;
+
+type Props = {|
+    assignees: TaskAssigneeCollection,
+    getAvatarUrl: GetAvatarUrlCallback,
+    maxAvatars?: number,
+    ...InjectIntlProvidedProps,
+|};
+
+const statusMessages = {
+    [TASK_NEW_APPROVED]: <FormattedMessage {...messages.tasksFeedStatusAccepted} />,
+    [TASK_NEW_REJECTED]: <FormattedMessage {...messages.tasksFeedStatusRejected} />,
+    [TASK_NEW_COMPLETED]: <FormattedMessage {...messages.tasksFeedStatusCompleted} />,
+    [TASK_NEW_NOT_STARTED]: <FormattedMessage {...messages.tasksFeedStatusNotStarted} />,
+};
+
+const AssigneeTooltipLabel = React.memo(({ user, status }) => {
+    const statusMessage = statusMessages[status] || null;
+    return (
+        <div className="bcs-task-assignment-tooltip-text">
+            <div>
+                <strong>{user.name}</strong>
+            </div>
+            {statusMessage && <div>{statusMessage}</div>}
+        </div>
+    );
+});
+
+class Assignees extends React.Component<Props> {
+    state = {};
+
+    render() {
+        const { maxAvatars = MAX_AVATARS, assignees, getAvatarUrl, intl } = this.props;
+        const assigneeCount = (assignees && assignees.entries.length) || 0;
+        const hiddenAssigneeCount = assigneeCount - maxAvatars;
+        const overflowLabel = `+${hiddenAssigneeCount}`;
+        const visibleAssignees = assignees.entries
+            .slice(0, maxAvatars)
+            .map(({ id: assignmentId, target, status: assigneeStatus }) => {
+                return (
+                    <Tooltip key={assignmentId} text={<AssigneeTooltipLabel user={target} status={assigneeStatus} />}>
+                        <AssigneeStatus
+                            key={assignmentId}
+                            status={assigneeStatus}
+                            user={target}
+                            getAvatarUrl={getAvatarUrl}
+                            data-testid="task-assignment-status"
+                        />
+                    </Tooltip>
+                );
+            });
+        const hiddenAssignees = assignees.entries.map(({ id: assignmentId, target, status: assigneeStatus }) => {
+            const statusMessage = statusMessages[assigneeStatus] || null;
+            return (
+                <div key={assignmentId} className="bcs-task-assignment-list-item">
+                    <AssigneeStatus
+                        status={assigneeStatus}
+                        className="bcs-task-assignment-list-item-avatar"
+                        user={target}
+                        getAvatarUrl={getAvatarUrl}
+                        data-testid="task-assignment-status"
+                    />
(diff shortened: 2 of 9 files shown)


boxcla:

src/elements/content-sidebar/activity-feed/task-new/AssigneeStatus.js
```diff
@@ -27,8 +27,8 @@ const StatusIcon = ({ status, ...rest }: { status: TaskCollabStatus }) => {
     }
 };
 
-const AssignmentStatus = React.memo<Props>(({ user, status, getAvatarUrl, ...rest }: Props) => (
-    <div className="bcs-task-assignment-status" {...rest}>
+const AssignmentStatus = React.memo<Props>(({ user, status, getAvatarUrl, className, ...rest }: Props) => (
```
jstoffan: Looks like `className` is missing from the `Props` type. Should it also have a default value of `''`?

src/elements/content-sidebar/activity-feed/task-new/Assignees.js
```diff
@@ -0,0 +1,113 @@
...
+    [TASK_NEW_REJECTED]: <FormattedMessage {...messages.tasksFeedStatusRejected} />,
+    [TASK_NEW_COMPLETED]: <FormattedMessage {...messages.tasksFeedStatusCompleted} />,
+    [TASK_NEW_NOT_STARTED]: <FormattedMessage {...messages.tasksFeedStatusNotStarted} />,
+};
+
+const AssigneeTooltipLabel = React.memo(({ user, status }) => {
+    const statusMessage = statusMessages[status] || null;
+    return (
+        <div className="bcs-task-assignment-tooltip-text">
+            <div>
```
jstoffan: Try to avoid non-semantic markup when possible. Should this be a heading tag?

src/elements/content-sidebar/activity-feed/task-new/Assignees.js
```diff
@@ -0,0 +1,113 @@
...
+                <strong>{user.name}</strong>
+            </div>
+            {statusMessage && <div>{statusMessage}</div>}
```
jstoffan: Should this be a `p` tag or a lesser heading?

src/elements/content-sidebar/activity-feed/task-new/Assignees.js
```diff
@@ -0,0 +1,113 @@
...
+|};
+
+const statusMessages = {
+    [TASK_NEW_APPROVED]: <FormattedMessage {...messages.tasksFeedStatusAccepted} />,
+    [TASK_NEW_REJECTED]: <FormattedMessage {...messages.tasksFeedStatusRejected} />,
+    [TASK_NEW_COMPLETED]: <FormattedMessage {...messages.tasksFeedStatusCompleted} />,
+    [TASK_NEW_NOT_STARTED]: <FormattedMessage {...messages.tasksFeedStatusNotStarted} />,
+};
+
+const AssigneeTooltipLabel = React.memo(({ user, status }) => {
```
jstoffan: Can we maintain one component per file? Having multiple in one makes testing and reuse more difficult.

src/elements/content-sidebar/activity-feed/task-new/Assignees.js
```diff
@@ -0,0 +1,113 @@
...
+        </div>
+    );
+});
+
+class Assignees extends React.Component<Props> {
+    state = {};
+
+    render() {
+        const { maxAvatars = MAX_AVATARS, assignees, getAvatarUrl, intl } = this.props;
+        const assigneeCount = (assignees && assignees.entries.length) || 0;
```
jstoffan: The flow type suggests that `assignees` is always defined. Is `assignees.entries` always defined?

src/elements/content-sidebar/activity-feed/task-new/Assignees.js
```diff
@@ -0,0 +1,113 @@
...
+            {statusMessage && <div>{statusMessage}</div>}
+        </div>
+    );
+});
+
+class Assignees extends React.Component<Props> {
+    state = {};
+
+    render() {
+        const { maxAvatars = MAX_AVATARS, assignees, getAvatarUrl, intl } = this.props;
```
jstoffan: Should `MAX_AVATARS` be the default prop value?

src/elements/content-sidebar/activity-feed/task-new/Assignees.js
```diff
@@ -0,0 +1,113 @@
...
+        const assigneeCount = (assignees && assignees.entries.length) || 0;
+        const hiddenAssigneeCount = assigneeCount - maxAvatars;
+        const overflowLabel = `+${hiddenAssigneeCount}`;
```
jstoffan: It doesn't look like this is used if `hiddenAssigneeCount` is less than zero and is used only once. Can it be moved to where it's used?

src/elements/content-sidebar/activity-feed/task-new/Assignees.js
```diff
@@ -0,0 +1,113 @@
...
+                            key={assignmentId}
+                            status={assigneeStatus}
+                            user={target}
+                            getAvatarUrl={getAvatarUrl}
+                            data-testid="task-assignment-status"
+                        />
+                    </Tooltip>
+                );
+            });
+        const hiddenAssignees = assignees.entries.map(({ id: assignmentId, target, status: assigneeStatus }) => {
```
jstoffan: It looks like this maps over all assignees rather than just those that are hidden.

src/elements/content-sidebar/activity-feed/task-new/Assignees.js
```diff
@@ -0,0 +1,113 @@
...
+    render() {
+        const { maxAvatars = MAX_AVATARS, assignees, getAvatarUrl, intl } = this.props;
+        const assigneeCount = (assignees && assignees.entries.length) || 0;
+        const hiddenAssigneeCount = assigneeCount - maxAvatars;
+        const overflowLabel = `+${hiddenAssigneeCount}`;
+        const visibleAssignees = assignees.entries
+            .slice(0, maxAvatars)
+            .map(({ id: assignmentId, target, status: assigneeStatus }) => {
+                return (
+                    <Tooltip key={assignmentId} text={<AssigneeTooltipLabel user={target} status={assigneeStatus} />}>
```
jstoffan: Seems like this should be its own component, along with the list of hidden assignees.

src/elements/content-sidebar/activity-feed/task-new/Assignees.js
```diff
@@ -0,0 +1,113 @@
...
+                            getAvatarUrl={getAvatarUrl}
+                            data-testid="task-assignment-status"
+                        />
+                    </Tooltip>
+                );
+            });
+        const hiddenAssignees = assignees.entries.map(({ id: assignmentId, target, status: assigneeStatus }) => {
+            const statusMessage = statusMessages[assigneeStatus] || null;
+            return (
+                <div key={assignmentId} className="bcs-task-assignment-list-item">
```
jstoffan: Should these be using `li` tags?

src/elements/content-sidebar/activity-feed/task-new/Assignees.js
```diff
@@ -0,0 +1,113 @@
...
+                                data-testid="task-assignment-overflow"
+                            >
+                                <Tooltip text={intl.formatMessage(messages.tasksFeedMoreAssigneesLabel)}>
+                                    <span className="bcs-task-assignment-count bcs-task-assignment-avatar">
+                                        {overflowLabel}
+                                    </span>
+                                </Tooltip>
+                            </button>
+                            <Overlay>
+                                <div className="bcs-task-assignment-list">
```
jstoffan: Should this be a `ul` tag?

src/elements/content-sidebar/activity-feed/task-new/Assignees.js
```diff
@@ -0,0 +1,113 @@
...
+                                    <h5 className="bcs-task-assignment-list-title">
```
jstoffan: Is this semantically correct or is it only intended to apply a certain visual style? If the latter, it should be done via CSS while using the appropriate heading level for the heading context.

src/elements/content-sidebar/activity-feed/task-new/Task.js
```diff
@@ -177,7 +137,7 @@ const Task = ({
             />
             <div className="bcs-task-content">{!!due_at && <TaskDueDate dueDate={due_at} status={status} />}</div>
             <div className="bcs-task-content">
-                <Assignees maxAvatars={MAX_AVATARS} assignees={assigned_to} getAvatarUrl={getAvatarUrl} />
+                <Assignees maxAvatars={3} assignees={assigned_to} getAvatarUrl={getAvatarUrl} />
```
jstoffan: Looks like `maxAvatars={3}` isn't needed here, as it's the default value in the component.

src/elements/content-sidebar/activity-feed/task-new/Assignees.js
```diff
@@ -0,0 +1,113 @@
...
+                </div>
+            );
+        });
+        return (
+            <div className="bcs-task-assignment-container">
+                <div className="bcs-task-assignments">
+                    {visibleAssignees}
+                    {hiddenAssigneeCount > 0 && (
+                        <Flyout position="top-left">
+                            <button
```
jstoffan: Should this be using `PlainButton`?

src/elements/content-sidebar/activity-feed/task-new/Task.scss
```diff
@@ -142,3 +128,50 @@ $card-left-offset: 40px;
...
+    }
+}
+
+.bcs-task-assignment-status-icon {
+    @include avatar-badge-icon;
+}
+
+.bcs-task-assignment-list {
+    max-width: 100%;
+    width: 280px;
```
jstoffan: Can this be avoided by making the overlay aware of its parent's size?

alexkrolick: This is to make them all the same size matching the design, unless it would cause the container to scroll.

alexkrolick:
Screen Shot 2019-04-18 at 2 10 12 AM
Screen Shot 2019-04-18 at 2 10 17 AM

src/elements/content-sidebar/activity-feed/task-new/Assignees.js
alexkrolick: We are going to aria-label the avatar/tooltip into a single piece of information for screenreaders as a followup

src/elements/content-sidebar/activity-feed/task-new/Assignees.js
alexkrolick: This component is under development - I'm about to add state and handle infinite loading. Can refactor/break up once the final structure is stable.

src/elements/content-sidebar/activity-feed/task-new/Task.scss
```diff
@@ -142,3 +120,64 @@ $card-left-offset: 40px;
...
+}
+
+.bcs-task-assignment-list-item {
+    align-items: flex-start;
+    display: flex;
+    margin: 6px 0 0 0;
+    padding: 3px 0;
+}
+
+.bcs-task-assignment-list-item-avatar {
```
jstoffan: These names are getting a little hairy. Would be good to do a pass to convert these to SUIT-style class names, but it can wait.

src/elements/content-sidebar/activity-feed/task-new/Assignees.js
jstoffan: Thoughts on this, @alexkrolick?

src/elements/content-sidebar/activity-feed/task-new/__tests__/__snapshots__/Assignees-test.js.snap
```diff
@@ -0,0 +1,921 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`elements/content-sidebar/ActivityFeed/task-new/Assignees should open assignee list when overflow icon is clicked: assignment list 1`] = `
```
jstoffan: These are pretty humungous. Would using `shallow` provide a clearer view of the rendered component hierarchy?

src/elements/content-sidebar/activity-feed/task-new/__tests__/Assignees-test.js
```diff
@@ -0,0 +1,75 @@
...
+    });
+
+    test('should open assignee list when overflow icon is clicked', () => {
+        const max = 2;
+        const wrapper = mount(<Assignees assignees={assignees} maxAvatars={max} getAvatarUrl={mockGetAvatarUrl} />);
+        const overflowIcon = global.queryAllByTestId(wrapper, 'task-assignment-overflow').first();
+        expect(wrapper.find('Overlay')).toHaveLength(0);
+        overflowIcon.simulate('click');
+        expect(wrapper.find('Overlay')).toHaveLength(1);
+        const assignmentList = wrapper.find('.bcs-task-assignment-list');
```
jstoffan: I'm finding these a bit difficult to read. May want to add some whitespace between related blocks to open it up a bit.

src/elements/content-sidebar/activity-feed/task-new/__tests__/__snapshots__/Assignees-test.js.snap
alexkrolick: I don't know if they are super valuable really - it's a bunch of Tooltips and Overlay classes. Will probably delete in the next pass.

booc0mtaco:
One thought on whether to move "sub-components" to separate files or not... I'd say we can make a strict rule that this applies to any file attempting to export multiple components. I think that components in flight might benefit from keeping them together until the prop structure and API have settled, to avoid any premature optimizations.

src/elements/content-sidebar/activity-feed/task-new/Assignees.js
alexkrolick: I originally had this as just a helper function, but thought adding a React.memo would be a cheap optimization, so it became a component. I've found it difficult to refactor components (like Comment) when they are broken up while still being tightly coupled to the parent. I'll take a pass at re-abstracting the Task and Comment cards after this.

alexkrolick:
@booc0mtaco I think "don't _export_ multiple components" is a good rule and it's the AirBnB lint default as of a few months ago (it used to ban multiple declarations). Also keeping an eye on file length - if it goes over about 200 lines I'm in favor of strongly investigating pulling it apart if possible.

jstoffan:
Fair enough. Thanks for responding with your thoughts.

PR #1120, opened 2019-04-18
fix(sidebar): Set correct default render prop and avoid lodash import

jstoffan:
Whoops! 🤦‍♂️


boxcla:

PR #1121, opened 2019-04-18
fix(sidebar): Sidebar now remembers toggle state between sessions

jstoffan:
Todo:
- [x] Unit tests for all permutations of toggled state


boxcla:

src/elements/content-sidebar/Sidebar.js
```diff
@@ -46,17 +47,22 @@ type Props = {
 };
 
 type State = {
-    isDirty: boolean,
-    isOpen: boolean,
+    isOpen: boolean, // Local isOpen state consists of stored forced state (if any) and responsive adjustments
 };
 
+export const SIDEBAR_FORCE_KEY: 'bcs.force' = 'bcs.force';
```
jeremypress: It looks like our abstraction is using `/` under the hood as a delimiter, should we continue that here?

`${KEY_PREFIX}/${SERVICE_VERSION}/${key}`

src/elements/content-sidebar/Sidebar.js
```diff
@@ -117,6 +129,36 @@ class Sidebar extends React.Component<Props, State> {
...
+
+        return this.store.getItem(SIDEBAR_FORCE_KEY);
+    }
+
+    /**
+     * Getter for whether the sidebar has been forced open
+     * @returns {boolean} - True if the sidebar has been forced open
+     */
+    isForcedOpen() {
+        return this.isForced() !== SIDEBAR_FORCE_VALUE_CLOSED;
```
jeremypress: isForced returns a boolean, but here we're comparing to the string `closed` so this will never be true?

src/elements/content-sidebar/Sidebar.js
```diff
@@ -117,6 +129,36 @@ class Sidebar extends React.Component<Props, State> {
...
+     *
+     * @param isOpen - Optionally set the sidebar to open/closed
+     * @returns {boolean} - The sidebar open/closed state
+     */
+    isForced(isOpen?: boolean): ?boolean {
+        if (isOpen !== undefined) {
+            this.store.setItem(SIDEBAR_FORCE_KEY, isOpen ? SIDEBAR_FORCE_VALUE_OPEN : SIDEBAR_FORCE_VALUE_CLOSED);
+        }
+
+        return this.store.getItem(SIDEBAR_FORCE_KEY);
```
jeremypress: What if we fell back to`false` if the key doesn't exist? Could clean up our types a bit.

src/elements/content-sidebar/Sidebar.js
jstoffan: I actually think dot notation has the advantage here because it could be used with lodash/get.

src/elements/content-sidebar/Sidebar.js
jstoffan: I also use this method to determine if the value is set at all or is null. I'll update the flow types, though.

src/elements/content-sidebar/Sidebar.js
jstoffan: Updated the types to be more clear.

PR #1124, opened 2019-04-19
fix(activityfeed): handle unknown users in activity feed

alexkrolick:
Unknown users returned by the API look like: `{type: "user", id: null, name: null}`

This fixes the activity feed avatar component and consumers to allow returning the UnknownUser avatar instead of throwing an error


boxcla:

src/elements/content-sidebar/activity-feed/task-new/Assignees.js
```diff
@@ -66,24 +68,26 @@ class Assignees extends React.Component<Props> {
...
-                    />
-                    <div className="bcs-task-assignment-list-item-details">
-                        <div>{target.name}</div>
-                        <div className="bcs-task-assignment-list-item-status">{statusMessage}</div>
-                    </div>
-                </li>
-            );
-        });
+        const allAssignees = entries
+            .map(fillUserPlaceholder)
```
alexkrolick: adds "fillUserPlaceholder"

booc0mtaco:
test for users.js, i.e.,

```
        test('should return null if there is no user id specified', async () => {
            const url1 = await users.getAvatarUrlWithAccessToken();
            expect(url1).toBeNull();
        });
```

src/elements/content-sidebar/activity-feed/Avatar.js
```diff
@@ -38,9 +38,10 @@ class Avatar extends React.PureComponent<Props, State> {
      * @return {Promise<?string>} Promise which resolve with the avatar url string
      */
     getAvatarUrl(): Promise<?string> {
-        const { user, getAvatarUrl }: Props = this.props;
+        const { user = {}, getAvatarUrl }: Props = this.props;
+        const { avatar_url = null, id } = user;
 
-        const avatarPromise = getAvatarUrl ? getAvatarUrl(user.id) : Promise.resolve(user.avatar_url);
+        const avatarPromise = id && getAvatarUrl ? getAvatarUrl(`${id}`) : Promise.resolve(avatar_url);
```
jeremypress: id is already a string according to the type?

alexkrolick: The type for this wrapper is User but it doesn't acknowledge some of the fields as nullable, and the underlying Avatar accepts numbers for the ID.

Changing the type has cascading consequences so we can follow up later.

PR #1126, opened 2019-04-19
feat(textarea): add ability to hide optional label text

priyajeet:
feature matches what `TextInput` already has.

Screen Shot 2019-04-19 at 12 29 45 PM


boxcla:

PR #1127, opened 2019-04-19
chore(content-preview): Adding sanity e2e tests

jeremypress:


boxcla:

test/integration/content-preview/ContentPreview.e2e.test.js
```diff
@@ -36,6 +37,36 @@ describe('ContentPreview', () => {
...
+            helpers.load({
+                fileId: Cypress.env('FILE_ID_SKILLS'),
+                props: {
+                    contentSidebarProps: {
+                        hasActivityFeed: null,
+                    },
+                },
+            });
+
+            cy.wait(2000); // eslint-disable-line cypress/no-unnecessary-waiting
```
jeremypress: Looking to avoid this, but we need to wait until everything is loaded to check for the absence of the sidebar.

jstoffan: Since we're checking for a negative, you probably need a positive result first. Can you look for the existence of preview before checking for the absence of the sidebar?

test/integration/content-preview/ContentPreview.e2e.test.js
```diff
@@ -36,6 +37,36 @@ describe('ContentPreview', () => {
...
+            cy.getByTestId('bcs-content').should('not.exist');
```
jstoffan: Should we be looking for the root element instead of the inner content element?

test/integration/content-preview/ContentPreview.e2e.test.js
jeremypress: unfortunately preview is loading before the sidebar appears when it's turned on. This looks like a potential regression, I would expect the sidebar loading state instantly.

jeremypress: looks like the chunk for the sidebar is loading after preview is considered loaded

jstoffan: Can we combine the positive and negative case (in that order)? That way we can test for the existence of the sidebar, change the props of ContentPreview, then test for the sidebar's absence?

jeremypress: to get the test page to reflect new props, we either have to visit a different page and then visit this one, or call `cy.reload`. Neither of those actions cause the `onbeforeload` callback to be executed, so the component won't re-render based on the new object on the window.

jstoffan: Okay. It just seems like having a "sanity" test rely on an arbitrary wait could lead to false alarms. Maybe add a comment explaining the issue.

PR #1130, opened 2019-04-19
fix(icon-sort-chevron): user should see a larger sort icon

ChenCodes:
Changes in this PR:
- IconSortChevron's default height / width should be 12x12 as requested by design

### DIFF
--- a/src/icons/general/IconSortChevron.js
+++ b/src/icons/general/IconSortChevron.js
@@ -12,7 +12,7 @@ type Props = {
     width?: number,
 };
 
-const IconSortChevron = ({ className = '', color = '#aeaeae', height = 11, title, width = 11 }: Props) => (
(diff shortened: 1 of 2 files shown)


boxcla:

priyajeet:
Have you checked other areas this icon is being used and it looks fine with size increase?

src/icons/general/IconSortChevron.js
```diff
@@ -12,7 +12,7 @@ type Props = {
     width?: number,
 };
 
-const IconSortChevron = ({ className = '', color = '#aeaeae', height = 11, title, width = 11 }: Props) => (
+const IconSortChevron = ({ className = '', color = '#aeaeae', height = 12, title, width = 12 }: Props) => (
```
macrotim: To avoid breaking existing dependencies, you may want to override height / width instead of changing it here.

ChenCodes:
Closing this PR as it is better to just override the styles to avoid breaking dependencies.

PR #1132, opened 2019-04-22
feat(list-view): user sees edit icon on hover metadata column cells

ChenCodes:
Changes in this PR:

Screen Shot 2019-04-22 at 10 55 29 AM


boxcla:

src/features/list-view/ListView.js
```diff
@@ -56,7 +71,13 @@ class ListView extends React.PureComponent<Props> {
         const cellData = getGridCell({ columnIndex, rowIndex: rowIndex - 1 });
 
         return (
-            <div className="bdl-ListView-columnCell" key={key} style={style}>
+            <div
+                className="bdl-ListView-columnCell"
+                key={key}
+                style={style}
+                onFocus={() => this.handleCellHover(columnIndex, rowIndex - 1)}
+                onMouseOver={() => this.handleCellHover(columnIndex, rowIndex - 1)}
```
macrotim: Don't forget onMouseOut and onBlur.

src/features/list-view/ListView.js
```diff
@@ -22,12 +22,27 @@ type Props = {
...
+        columnIndex: number,
+        rowIndex: number,
+    |}) => void,
     onSortChange?: (columnIndex: number) => void,
     rowCount: number,
     width: number,
 };
 
 class ListView extends React.PureComponent<Props> {
+    gridEl: MultiGrid;
```
macrotim: Let's pass an additional prop to <MultiGrid> instead of using ref.

src/features/list-view/ListView.js
```diff
@@ -56,7 +71,13 @@ class ListView extends React.PureComponent<Props> {
         const cellData = getGridCell({ columnIndex, rowIndex: rowIndex - 1 });
```
macrotim: I'm realizing what confuses me about "rowIndex: rowIndex - 1". It should be "cellIndex: rowIndex - 1". This naming is consistent with the header versus cell convention. Thoughts?

src/features/list-view/ListView.js
```diff
@@ -22,12 +22,27 @@ type Props = {
     getGridHeader: (columnIndex: number) => any,
     getGridHeaderSort?: (columnIndex: number) => typeof SORT_ORDER_ASCENDING | typeof SORT_ORDER_DESCENDING | null,
     height: number,
+    onCellHover?: ({|
```
macrotim: On second thought, I'd prefer that we move this to EUA. I couldn't think of a reason previously but am realizing that any work in BUI tends to carry undesirable overhead.

src/features/list-view/ListView.js
ChenCodes: This makes sense. It's not immediately obvious to a developer why the rowIndex - 1 exists but calling it a cell helps give more clarity.

src/features/list-view/__tests__/ListView-test.js
```diff
@@ -49,11 +43,29 @@ describe('features/list-view/ListView', () => {
                 rowIndex,
             });
 
-            const testCell = shallow(cell);
+            expect(cell.props.children[0]).toBe(cellData);
+            expect(cell.props.className).toBe(className);
+        });
 
-            expect(testCell.text()).toEqual(cellData);
+        test.each`
```
macrotim: Let's consider merging back into its previous form. There's some copy / paste that I'd like to avoid.

src/features/list-view/examples/ListViewExamples.js
```diff
@@ -6,21 +6,25 @@ import { AutoSizer } from 'react-virtualized/dist/es/AutoSizer/index';
 import ListView from '../ListView';
 
 type Props = {
+    cellCoordinate: string,
```
macrotim: This naming doesn't seem right.

src/features/list-view/ListView.js
```diff
@@ -120,6 +121,7 @@ class ListView extends React.PureComponent<Props> {
                     fixedColumnCount={1}
                     fixedRowCount={FIXED_ROW_COUNT}
                     height={height}
+                    gridDataHash={gridDataHash} // Forces MultiGrid to re-render
```
macrotim: The comment should be adjacent to its declaration (type Props).

src/features/list-view/examples/ListViewExamples.js
```diff
@@ -6,21 +6,25 @@ import { AutoSizer } from 'react-virtualized/dist/es/AutoSizer/index';
...
     columnCount: number,
 };
 
+const defaultCellCoordinate = '-1-1';
```
macrotim: Same.

src/features/list-view/ListView.js
```diff
@@ -16,11 +16,12 @@ type Props = {
     columnCount: number,
     getColumnWidth?: (columnIndex: number) => number,
     getGridCell: ({|
+        cellRowIndex: number,
         columnIndex: number,
```
macrotim: Let's rename this to cellColumnIndex.

src/features/list-view/__tests__/ListView-test.js
```diff
@@ -49,11 +49,8 @@ describe('features/list-view/ListView', () => {
                 rowIndex,
             });
 
-            const testCell = shallow(cell);
-
-            expect(testCell.text()).toEqual(cellData);
-
-            expect(testCell.hasClass(className)).toBeTruthy();
+            expect(cell.props.children[0]).toBe(cellData);
```
macrotim: I believe it should be .text().

PR #1133, opened 2019-04-22
fix(tasks): Remove spacing after task creator name in task card

mattnguyen1:


boxcla:

PR #1134, opened 2019-04-22
fix(tasks): Allow names to wrap in task card when profile links provided

mattnguyen1:


boxcla:

mattnguyen1:
The name wraps fine when it isn't a link. Having the link class turns off the wrapping, which we don't want in this case.

PR #1135, opened 2019-04-22
fix(tasks): Make task assignee dropdown scrollable

mattnguyen1:


boxcla:

src/elements/content-sidebar/activity-feed/task-form/TaskForm.scss
```diff
@@ -53,6 +53,17 @@
             .pill-selector-input-wrapper {
                 width: auto;
             }
+
+            .overlay-wrapper {
+                border: 1px solid $sf-fog;
```
booc0mtaco: does this gray map to any of the `bdl-neutral` colors?

src/elements/content-sidebar/activity-feed/task-form/TaskForm.scss
```diff
@@ -53,6 +53,17 @@
...
+                height: 215px;
```
booc0mtaco: let's pull this up as a constant. if it can be derived or related to other measurements, we can set those as well.

src/elements/content-sidebar/activity-feed/task-form/TaskForm.scss
mattnguyen1: nope, and it seems a lot of our borders in general end up using this `sf-fog` color as well

mattnguyen1:
looks like this doesn't work correctly if there are not enough contacts, will push a fix for it

src/elements/content-sidebar/activity-feed/task-form/__tests__/TaskForm-test.js
```diff
@@ -93,6 +93,28 @@ describe('components/ContentSidebar/ActivityFeed/task-form/TaskForm', () => {
         expect(wrapper.find('PillSelectorDropdown').prop('selectorOptions').length).toBe(1);
     });
 
+    test('should not add scrollable class when there are not enough contacts in assignment dropdown', () => {
```
booc0mtaco: we can make good use of [`setProps`](https://airbnb.io/enzyme/docs/api/ShallowWrapper/setProps.html) here.

PR #1138, opened 2019-04-23
fix(tasks): task due date input handling

alexkrolick:
- remove onBlur that causes month/yr select to lose focus
- allow manual date entry as designed


boxcla:

PR #1139, opened 2019-04-23
feat(uploader): prevent upload based on file size

priyajeet:
Rough


boxcla:

ConradJChan:
`STATUS_CRITICAL` is a new error status that cannot be retried? Do we need to update the logic to be able to click to remove such files from the upload queue?

jeremypress:
for restrictions like this, file type etc. customers are probably going to want custom error messages. Ideally we could provide that without provide a prop for each string.

priyajeet:
Will have to make the error stuff default to string message if code is not provided or is bogus. And if no message either then we show our default message.

In either case onBeforeUpload should return data or promise and we should handle both, and the item returned can have the error info, which is then used to create the upload item object

PR #1140, opened 2019-04-23
feat(unified-share-modal): Add fuzzy search to USM

smoorebox:


boxcla:
Verified that @smoorebox has signed the CLA. Thanks for the pull request!

src/features/unified-share-modal/__tests__/ContactsField-test.js
```diff
@@ -111,6 +152,122 @@ describe('features/unified-share-modal/ContactsField', () => {
...
+        test('should order the suggested selectorOptions by userScore', () => {
+            const wrapper = getWrapper({
+                suggestedCollaborators: largeSuggestions,
+            });
+            const result = wrapper.instance().addSuggestedContacts(Object.values(largeSuggestions));
+            expect(wrapper.state().numSuggestedShowing).toEqual(MAX_SUGGESTIONS_TO_SHOW);
+            expect(result).toMatchSnapshot();
+        });
+
+        test('should match suggested collaborators even when options are empty if a search matches', () => {
```
nickgriffin1: It's a bit hard to tell what's going on here from the snapshot.

src/features/unified-share-modal/__tests__/ContactsField-test.js
```diff
@@ -111,6 +152,122 @@ describe('features/unified-share-modal/ContactsField', () => {
...
+
+        test('should match suggested collaborators but not duplicate', () => {
+            const wrapper = getWrapper({
+                suggestedCollaborators: largeSuggestions,
+            });
+            wrapper.setState({
+                pillSelectorInputValue: 'person',
+            });
+            const result = wrapper.instance().addSuggestedContacts(Object.values(largeSuggestions));
+            expect(result).toMatchSnapshot();
```
nickgriffin1: For this test and others, you should add an expectation to ensure the size or some other basic expectation is met. I think this redundancy is nice to have because it's so easy to update snapshot.
`expect(result).toHaveLength(x)`
or you could check for emails
`expect(result.map(contact => contact.email)). arrayContaining(<some-email>)`
or the inverse
`expect(result.map(contact => contact.email)).not. arrayContaining(<some-email>)`

booc0mtaco:
After a brief chat, will investigate extracting the list parsing logic into a separate utility to aid in generalizing this for other components/use cases.

src/utils/fuzzySearch.js
```diff
@@ -19,6 +19,9 @@
  * @returns {boolean} If a match is found
  */
 const fuzzySearch = (search: string, content: string, minCharacters: number = 3, maxGaps: number = 2) => {
+    if (typeof search !== 'string' || typeof content !== 'string') {
```
FuzzyTree: Wouldn't `search` and `content` guaranteed to be strings due to typing?

smoorebox: In theory yeah, but I ran into the case when testing so I added a safety clause to be safe.

src/features/unified-share-modal/__tests__/ContactsField-test.js
```diff
@@ -31,20 +31,23 @@ describe('features/unified-share-modal/ContactsField', () => {
             email: 'x@example.com',
             id: '12345',
             text: 'X User',
+            name: 'X User',
```
booc0mtaco: Hm, do we need both `name` and `text`? seems that `text` was already there, and is always the same as name.

src/utils/fuzzySearch.js
booc0mtaco: Most of the test files do not include `// @flow` so they don't reveal any type checking issues. this'll let the test cases violate the type contract. 

Do we want to support `search` or `content` being anything but strings?

src/utils/__tests__/fuzzySearch-test.js
```diff
@@ -71,4 +71,15 @@ describe('util/fuzzySearch', () => {
     test('should return false when search is empty', () => {
         expect(fuzzySearch('', 'foo bar')).toBe(false);
     });
+
+    test('should return false when search/content are not a string', () => {
```
booc0mtaco: We can probably remove these checks since we have the types in fuzzySearch.js. Since test files skip flow checks, the lines below don't have any flow errors.

src/features/unified-share-modal/__tests__/ContactsField-test.js
smoorebox: I don't actually know much about the API response and what the type is used by. I am mostly just adapting to what exists here, might merit a deeper look though.

src/utils/SuggestedCollabs.js
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * @param {Array<Contact>} contacts Contacts returned from the server
+ * @param {SuggestedCollabLookup} suggestedCollaborators Cached suggested collab lookup to match on
+ * @param {string} inputValue Value the user typed into the input box searching for collabs
+ * @param {number} maxSuggestions Maximum number of items to suggest/return, default 3
+ * @param {number} minCharacters Minimum number of characters before searching, default 3
+ * @returns {Array<Contact>} Suggested collaborators formatted as Contacts in recommended order
+ */
+function computeSuggestedContacts(
+    contacts: Array<Contact>,
```
nickgriffin1: One thing I wanted to point out is the data format here. A different implementation of the PillSelector use a data type like `Array<{ text: string, value: string, data: Object }>`. Where `data` is similar to `Contact`. I think this is fine because you can just pass the data field to this function or you can convert one data type to another but I think it's worth pointing out because the pill selector itself might not be happy about the `Contact` format that is returned. If the other implementation of PillSelector is going to use this you might need to add another function and param to create the data format it wants or you can refactor the other implementation to use this data format. Not a huge deal, but I think it's worth pointing out up-front.

smoorebox: Totally agree, there are a few other things like that such as whether the consumers should pass in the constants etc. I decided to keep it simple and didn't want to go too deep on restructuring all the types and such until we add the next use case(s).

src/utils/SuggestedCollabs.js
```diff
@@ -0,0 +1,92 @@
...
+    const exactMatchIds = contacts
+        .filter(option => {
+            const id = option.id.toString();
+            return id && suggestedCollaborators[id];
+        })
+        .map(option => option.id);
+
+    const suggestedCollabs = Object.keys(suggestedCollaborators).map(key => suggestedCollaborators[key]);
+
+    const suggestedExactMatches = suggestedCollabs.filter(option => {
```
jstoffan: I may be reading this wrong, but it seems like this is very similar to the logic used to compute `exactMatchIds`. Can the two filters be combined and/or simplified, perhaps eliminating one of the arrays?

src/utils/SuggestedCollabs.js
```diff
@@ -0,0 +1,92 @@
...
+                      return false;
+                  }
+                  const nameMatches = fuzzySearch(inputValue, option.name, minCharacters, MAX_GAPS_FUZZY_MATCH);
+                  const emailAddress = option.email || '';
+                  const emailAlias = emailAddress.substring(0, emailAddress.indexOf('@'));
+                  const emailMatches = inputValue.length >= minCharacters && emailAlias.indexOf(inputValue) > -1;
+                  return nameMatches || emailMatches;
+              });
+
+    const sortSuggestionList = list => {
```
jstoffan: This one method is doing quite a lot of work and is also redeclaring inner helper functions when it doesn't appear to be needed. I would really recommend breaking it into smaller functions to ease readability and testing burden.

src/features/unified-share-modal/ContactsField.js
```diff
@@ -55,20 +55,20 @@ class ContactsField extends React.Component<Props, State> {
         };
     }
 
-    addSuggestedContacts = (contacts: Array<Contact>) => {
+    getSuggestedCollaboratorArray = (): Array<SuggestedCollab> => {
```
jstoffan: Where is this used?

src/features/unified-share-modal/ContactsField.js
```diff
@@ -55,20 +55,20 @@ class ContactsField extends React.Component<Props, State> {
...
-                return id && suggestedCollaborators[id.toString()];
-            })
-            .sort((optionA, optionB) => {
-                const currentSuggestedItemA = suggestedCollaborators[optionA.id.toString()];
-                const currentSuggestedItemB = suggestedCollaborators[optionB.id.toString()];
-                return currentSuggestedItemB.userScore - currentSuggestedItemA.userScore;
-            })
-            .slice(0, 3);
+    addSuggestedContacts = (contacts: Array<Contact>): Array<Contact> => {
+        const { suggestedCollaborators = {} } = this.props;
```
jstoffan: Recommend using `static defaultProps` for default prop values to keep class behavior consistent.

src/constants.js
```diff
@@ -370,3 +370,9 @@ export const KEYS = {
 
 /* ----------------- Other ----------------------- */
 export const ONE_HOUR_MS = 3600000; // 60 * 60 * 1000
+
+/* ----------------- Fuzzy Search/Suggested Collabs ----------------------- */
```
jstoffan: This constants file is intended for global use; these values seem very specific to USM and should probably live in its feature directory.

flow-typed/box-ui-elements.js
```diff
@@ -980,3 +980,23 @@ type AdditionalVersionInfo = {
 };
 
 type OnVersionChange = (version: ?BoxItemVersion, additionalVersionInfo: ?AdditionalVersionInfo) => void;
+
+type Contact = {
```
jstoffan: The name of this type is very generic but the data it represents is very specific to a single feature.

src/utils/SuggestedCollabs.js
```diff
@@ -0,0 +1,92 @@
...
+        return list.sort((optionA: SuggestedCollab, optionB: SuggestedCollab) => {
+            const currentSuggestedItemA = suggestedCollaborators[optionA.id.toString()];
```
jstoffan: It looks like both `list` values are arrays of `SuggestedCollab` objects generated as subsets of
 the `suggestedCollaborators` map. Is this lookup actually needed? Adding some flow types for these helper methods would be helpful.

src/features/unified-share-modal/__tests__/ContactsField-test.js
jstoffan: I share @booc0mtaco's question and think it should be at least investigated before moving forward.

src/utils/SuggestedCollabs.js
```diff
@@ -0,0 +1,92 @@
+/**
+ * @flow
+ * @file Suggested Collaborator utility functions
```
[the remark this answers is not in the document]
smoorebox: It will be used by an external dependency as well unrelated to USM. This logic is currently duplicated in EUA for the create-new-folder modal as well, that was one of the primary motivations for doing this refactor. After this goes out I plan on consuming this util from there as well. Is that enough to keep it in the general utils folder?

src/features/unified-share-modal/ContactsField.js
smoorebox: It was inlined in the util, will remove, good catch.

src/constants.js
smoorebox: Yeah I moved it here because it used to be in the USM folder but needed to move out to be referenced by utils as well. Perhaps it should move to utils? I was also a little unsure on if I wanted to move thee values out to the consumers instead of them living in the util.

src/utils/SuggestedCollabs.js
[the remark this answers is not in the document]
jstoffan: I see. It definitely sounds like we're in need of a new low-level component that all of these various high-level features can share.

PR #1141, opened 2019-04-23
fix(sidebar): Show the selected class only when sidebar is open

jstoffan:


boxcla:

src/elements/content-sidebar/SidebarNavButton.js
```diff
@@ -13,18 +13,20 @@ import './SidebarNavButton.scss';
...
-const SidebarNavButton = ({ children, interactionTarget, onNavigate, sidebarView, tooltip }: Props) => {
+const SidebarNavButton = ({ children, interactionTarget, isOpen, onNavigate, sidebarView, tooltip }: Props) => {
     const sidebarPath = `/${sidebarView}`;
 
     return (
         <Route path={sidebarPath}>
             {({ match }) => {
-                const isToggle = !!match && match.isExact;
+                const isMatch = !!match && match.isExact;
+                const isActive = () => !!isOpen && isMatch;
```
ConradJChan: Does this need to be a function? Can we determine `isActive` during `render()` and let the `NavButton IsActive` prop be a boolean?

jstoffan: The goal is to mimic the behavior of [React Router's NavLink component](https://reacttraining.com/react-router/web/api/NavLink/isactive-func), which expects a function.

PR #1142, opened 2019-04-23
fix(query-bar): change representation of not equals in query bar

ChenCodes:
Changes in this PR:
- Change '!=' to '<>'


boxcla:

PR #1143, opened 2019-04-23
fix(list-view): show pointer on hover of metadata column header

ChenCodes:
Changes in this PR:
- Take in isGridHeaderSortable handler that determines whether or not a column header is sortable
- Apply ```cursor: pointer``` if they column header is sortable

ezgif com-video-to-gif (6)

### DIFF
--- a/src/features/list-view/ListView.js
+++ b/src/features/list-view/ListView.js
@@ -47,16 +47,16 @@ class ListView extends React.PureComponent<Props> {
             });
             const isHeaderSortable = isGridHeaderSortable(columnIndex);
 
-            const headerClassName = classNames('bdl-ListView-columnHeader', {
+            const buttonClassName = classNames('bdl-ListView-columnHeader', {
                 'bdl-ListView-isHeaderSortable': isHeaderSortable,
             });
             return (
                 <button
-                    className={headerClassName}
+                    className={buttonClassName}
                     key={key}
                     style={style}
                     type="button"
-                    onClick={() => onSortChange(columnIndex)}
+                    onClick={() => isHeaderSortable && onSortChange(columnIndex)}
                 >
                     {displayName}
                     {sortDirection && <IconSortChevron className={iconClassName} />}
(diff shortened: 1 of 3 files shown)


boxcla:

macrotim:
Let's obsolete this with https://github.com/box/box-ui-elements/pull/1144

src/features/list-view/ListView.js
```diff
@@ -22,33 +22,44 @@ type Props = {
...
             const displayName = getGridHeader(columnIndex);
             const sortDirection = getGridHeaderSort(columnIndex);
             const isSortAsc = sortDirection === SORT_ORDER_ASCENDING;
-            const className = classNames({
+            const iconClassName = classNames({
                 'bdl-ListView-isSortAsc': isSortAsc,
             });
+            const isHeaderSortable = isGridHeaderSortable(columnIndex);
 
+            const headerClassName = classNames('bdl-ListView-columnHeader', {
```
Visylvius: headerClassName seems generic, not descriptive enough.
Perhaps we could change this to something else?

ChenCodes: There's only one header within this ListView component so I think it's appropriate that we have a generic header class name. If there were multiple headers in this component then I would push for a more specific name.

src/features/list-view/ListView.js
```diff
@@ -22,33 +22,44 @@ type Props = {
...
+                'bdl-ListView-isHeaderSortable': isHeaderSortable,
+            });
             return (
                 <button
-                    className="bdl-ListView-columnHeader"
+                    className={headerClassName}
```
macrotim: Should be "buttonClassName".

src/features/list-view/ListView.js
```diff
@@ -22,33 +22,44 @@ type Props = {
...
                     key={key}
                     style={style}
                     type="button"
                     onClick={() => onSortChange(columnIndex)}
```
macrotim: Please add... isHeaderSortable  && onSortChange(columnIndex)

PR #1144, opened 2019-04-23
feat(list-view): Removed blue outline, enabled header pointer

macrotim:
Includes visual fixes.

### DIFF
--- a/i18n/en-US.properties
+++ b/i18n/en-US.properties
@@ -132,10 +132,6 @@ be.executeIntegrationOpenWithErrorSubHeader = Please try again later.
 be.expand = Expand
 # Label for face skill section in the preview sidebar
 be.faceSkill = Faces
-# Call-to-action text describing what to do to navigate to specified feedback form
-be.feedbackCtaText = Click to provide feedback
-# Accessible text used to describe the form used for feedback
-be.feedbackFormDescription = Beta Feedback Form
 # File access stats error message
 be.fileAccessStatsErrorHeaderMessage = Something went wrong when fetching the access stats.
 # File classification error message
@@ -358,18 +354,10 @@ be.skillUnknownError = Something went wrong with running this skill or fetching 
 be.sort = Sort
 # Label for status skill card in the preview sidebar
 be.statusSkill = Status
-# Title shown when an error occurs performing an action on a task
-be.taskActionErrorTitle = Error
 # Approve option for a task
 be.taskApprove = Complete
-# Error message when approving a task fails
-be.taskApproveErrorMessage = An error has occurred while approving this task. Please refresh the page and try again.
-# Error message when completing a task fails
-be.taskCompleteErrorMessage = An error has occurred while completing this task. Please refresh the page and try again.
 # Error message when a task creation fails
-be.taskCreateErrorMessage = An error occurred while creating this task. Please try again.
-# Title shown above error message when a task creation fails
-be.taskCreateErrorTitle = Error
+be.taskCreateErrornMessage = There was an error creating this task.
 # Error message when a task deletion fails
 be.taskDeleteErrorMessage = There was an error deleting this task.
 # Confirmation prompt text to delete task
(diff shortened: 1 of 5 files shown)


boxcla:

i18n/en-US.properties
```diff
@@ -380,70 +368,34 @@ be.taskDueDate = Due
...
 # label for button that opens task popup
 be.tasks.addTask = Add Task
-# label for menu item that opens approval task popup
-be.tasks.addTask.approval = Approval
-# description for menu item that opens approval task popup
-be.tasks.addTask.approval.description = Request an approval to move work forward
-# label for menu item that opens general task popup
-be.tasks.addTask.general = Task
-# description for menu item that opens general task popup
-be.tasks.addTask.general.description = Keep track of work that needs to get done
```
jstoffan: Why are we deleting tasks messages?

src/features/list-view/styles/ListView.scss
```diff
@@ -1,14 +1,16 @@
 @import '../../../styles/variables';
 
-.metadata-views-list-view {
+$prefix: bdl-MetadataViews-ListView;
```
jstoffan: This will be defined as a global variable. Is that intended?

src/features/list-view/styles/ListView.scss
```diff
@@ -20,16 +22,24 @@
...
+            cursor: pointer;
+        }
+
+        .bdl-icon-sort-chevron.#{$prefix}-isSortAsc {
             transform: rotate(180deg);
         }
     }
 
-    .bdl-ListView-columnCell {
+    .#{$prefix}-columnCell {
```
jstoffan: This isn't really necessary, in my opinion. Repeating the class name is acceptable. It could also probably be shortened to `.bdl-ListView`, since that's the name of the component.

src/features/list-view/styles/ListView.scss
```diff
@@ -20,16 +22,24 @@
         text-transform: none;
         width: 100%;
 
-        .bdl-icon-sort-chevron.bdl-ListView-isSortAsc {
+        &--sortable {
```
jstoffan: Please avoid this convention. It makes class names very difficult to search globally.

src/features/list-view/ListView.js
```diff
@@ -22,41 +22,42 @@ type Props = {
...
-            const isSortAsc = sortDirection === SORT_ORDER_ASCENDING;
-            const className = classNames({
-                'bdl-ListView-isSortAsc': isSortAsc,
-            });
+            const isSortable = isGridHeaderSortable(columnIndex);
 
             return (
                 <button
-                    className="bdl-ListView-columnHeader"
+                    className={classNames(`${CSS_PREFIX}-columnHeader`, {
```
jstoffan: This seems like a pretty heavy and somewhat low value abstraction for a cell that could be rendered many times. I believe this generates/calls functions under the hood after transpilation.

src/features/list-view/ListView.js
```diff
@@ -105,11 +106,20 @@ class ListView extends React.PureComponent<Props> {
         return canFillWidth ? getClientDefinedColumnWidth(columnIndex) : (width - firstColumnWidth) / (columnCount - 1);
     };
 
+    renderSortIcon(columnIndex: number) {
```
jstoffan: Seems like this is ready to live in its own component.

src/features/list-view/styles/ListView.scss
priyajeet: Final CSS will also have full classname, so having it a variable makes no difference.

src/features/list-view/styles/ListView.scss
macrotim: Good catch. I can revise to...

```.bdl-MetadataViews-ListView {
    $prefix: bdl-MetadataViews-ListView;
   ...
```

Thoughts?

jstoffan: I'm not sure I see the value of the variable. Part of the benefit of SUIT is to avoid nesting and increases in specificity. The code duplication without the variable is minimal, is it not?

src/features/list-view/styles/ListView.scss
macrotim: "ListView" is too generic. There is already a different ListView in EUA which could be moved to BUI in the near future.

i18n/en-US.properties
macrotim: I tried rebasing and yarn-run-build with no success. Do you have suggestions?

src/features/list-view/styles/ListView.scss
macrotim: There's a tradeoff between searchability and DRY. The names are an eyesore the other way.

Visylvius:
As this is a stylistic PR, can you please include a screen shot, or giphy?

src/features/list-view/styles/ListView.scss
jstoffan: Then the component files, folder, etc. should also be renamed to something more specific. Also, SUIT uses hyphens as indicators of descendent components.

src/features/list-view/ListView.js
```diff
@@ -22,41 +22,42 @@ type Props = {
     getGridHeader: (columnIndex: number) => any,
     getGridHeaderSort?: (columnIndex: number) => typeof SORT_ORDER_ASCENDING | typeof SORT_ORDER_DESCENDING | null,
     height: number,
+    isGridHeaderSortable?: (columnIndex: number) => boolean,
     onSortChange?: (columnIndex: number) => void,
     rowCount: number,
     width: number,
 };
 
+const CSS_PREFIX = 'bdl-MetadataViews-ListView';
```
Visylvius: Can this be moved to constants?

src/features/list-view/ListView.js
macrotim: Not sure if I agree on the heaviness.

src/features/list-view/ListView.js
```diff
@@ -105,11 +106,20 @@ class ListView extends React.PureComponent<Props> {
...
+
+        return sortDirection && <IconSortChevron className={classNames({ [`${CSS_PREFIX}-isSortAsc`]: isSortAsc })} />;
+    }
+
     render() {
         const { columnCount, height, rowCount, width } = this.props;
 
         return (
-            <div className="metadata-views-list-view">
+            <div className={`${CSS_PREFIX}`}>
```
Visylvius: I originally was also following this pattern.
One thing that was brought to my attention by AK was that if we were to do a global search and find on this class name we wouldn't be able to find this class, as it is being set to a const. Is this something we want?

src/features/list-view/examples/ListViewExamples.js
```diff
@@ -15,6 +15,13 @@ const getGridHeader = columnIndex => `Header ${generateRandomString(columnIndex)
 
 const getGridCell = ({ columnIndex, rowIndex }) => `Row ${rowIndex}, Column ${columnIndex}`;
 
+const isGridHeaderSortable = () => true;
+
+const onSortChange = columnIndex => {
+    // eslint-disable-next-line no-console
+    console.log(`Column header #${columnIndex} clicked`);
```
Visylvius: would you mind explaining the reason we are comitting a console log to the code base?

src/features/list-view/ListView.js
priyajeet: `bdl-ListView` since the component name is `ListView`

src/features/list-view/ListView.js
priyajeet: `<div className={CSS_PREFIX}>`

src/features/list-view/styles/ListView.scss
macrotim: There's this -> .bdl-icon-sort-chevron.bdl-MetadataViews-ListView-isSortAsc which I find barely legible.

As it stands today, I find repetitiveness of the prefix to be an eyesore and this is exacerbated as the scss file will only grow over time. 

Given the lengthiness of this prefix, I prefer DRY over other virtues.

src/features/list-view/ListView.js
priyajeet: Such multiline class names should also be taken out from within JSX so that while debugging we can put proper break points

src/features/list-view/styles/ListView.scss
macrotim: I guess I have two options:
1. Leave as bdl-ListView and risk a future collision
2. Rename the prefix but introduce a breaking change

Given my time constraints, I'll have to go with #1.

src/features/list-view/styles/ListView.scss
jstoffan: I agree, that class name isn't acceptable, but the issue isn't caused by SUIT or the conventions I describe; it's caused by the component architecture. Each component should be a separate concern (and file, preferably) with a class to match.

Example: `.bdl-MetadataList-sort---asc`

i18n/en-US.properties
jstoffan: Are you pushing with `--no-verify`?

src/features/list-view/ListView.js
macrotim: I can see the value of a global search. If that's important, then we can tradeoff DRY for searchability.

src/features/list-view/examples/ListViewExamples.js
macrotim: It's part of styleguidist -> Log to console when user clicks on header [to sort].

src/features/list-view/styles/ListView.scss
macrotim: I take it back. If searchability is important, then let's do that instead of DRY.

src/features/list-view/styles/ListView.scss
macrotim: @priyajeet It's an eyesore for readers to repeat a lengthy prefix but if we prefer searchability, then I can buy into that.

macrotim:
Closing this PR. We're going with this instead -> https://github.com/box/box-ui-elements/pull/1143

macrotim:
One more note: Moving forward, I'll prefer global searchability over DRY in CSS names.

i18n/en-US.properties
macrotim: Sometimes, I use --no-verify for incremental changes, but I immediately trigger `push` with it omitted and let it run in the background. This shortens my dev-cycles and allows me to get feedback from my team more quickly. 

Do you think it's breaking i18n?

jstoffan: It seems likely, yes. The pre-push script runs other scripts necessary to update the messages, as well as linting and unit tests. We strongly recommend against pushing with `--no-verify` for that reason.

macrotim:
My bad.  I will avoid it moving forward.

Thanks for all the great feedback!

PR #1145, opened 2019-04-24
fix(pickers): improve responsiveness when using longer strings

jeremypress:
Before: 
Screen Shot 2019-04-23 at 3 12 23 PM

After: 
Screen Shot 2019-04-24 at 3 34 55 PM

Empty state text
Before:
Screen Shot 2019-04-24 at 9 59 31 AM

After:
Screen Shot 2019-04-24 at 9 58 04 AM


boxcla:

src/elements/common/empty-state/EmptyState.scss
```diff
@@ -4,4 +4,5 @@
     flex: 1;
     flex-direction: column;
     justify-content: center;
+    text-align: center;
```
jstoffan: Unrelated?

src/elements/content-picker/Footer.scss
```diff
@@ -9,11 +9,23 @@
...
+        min-width: 0;
+    }
+
     .bcp-selected {
+        flex: 0 0 75px;
         margin: 0;
+        width: 100%;
 
         .btn-content {
-            display: flex;
```
jstoffan: Is everything inside the selected button, including the red file limit text (`(2 max)`), still vertically centered with this change?

src/elements/content-picker/Footer.scss
```diff
@@ -9,11 +9,23 @@
     justify-content: space-between;
     padding: 0 20px;
 
+    .bcp-footer-left {
+        display: flex;
+        flex: 1;
+        min-width: 0;
```
jstoffan: Is this necessary? Is it overriding something?

src/elements/content-picker/Footer.scss
```diff
@@ -9,11 +9,23 @@
...
+    }
+
     .bcp-selected {
+        flex: 0 0 75px;
         margin: 0;
+        width: 100%;
```
jstoffan: Generally not a great idea to mix flex and width. Should the basis or the width take precedence here?

src/elements/content-picker/ItemList.js
```diff
@@ -120,7 +120,7 @@ const ItemList = ({
                             height={height}
                             disableHeader
                             headerHeight={0}
-                            rowHeight={50}
+                            rowHeight={65}
```
jstoffan: Does this still look okay in wider resolutions and/or with single-line items? Can you post a screenshot?

src/elements/common/empty-state/EmptyState.scss
jeremypress: unrelated but fixes text alignment for the empty state at small widths. I can make this a separate change if you'd like.

src/elements/content-picker/ItemList.js
jeremypress: Added screenshots to the PR description.

src/elements/content-picker/Footer.scss
jeremypress: The button won't resize without it: following advice from stack overflow / https://css-tricks.com/flexbox-truncated-text/

ConradJChan:
The 65px looks pretty spacious on a wider screen, should we instead use the `isSmall` prop and add some logic around reducing the content showed on each line to avoid increasing the row height?

src/elements/content-picker/Footer.scss
jeremypress: looks like the centering is off, I'll look for a solution.

PR #1149, opened 2019-04-26
feat(formik): shareable form elements to use with formik

priyajeet:
Screen Shot 2019-04-25 at 6 05 39 PM


boxcla:

src/components/radio/RadioGroup.js
```diff
@@ -47,7 +50,7 @@ class RadioGroup extends React.Component<Props, State> {
         const { value } = this.state;
 
         return (
-            <div className={`radio-group ${className}`} key={value} onChange={this.onChangeHandler}>
```
priyajeet: I had added this few days ago, is redundant, needed on parent instead.

src/components/select-field/index.js
```diff
@@ -1,3 +1,4 @@
 // @flow
 export { default as SingleSelectField } from './SingleSelectField';
 export { default as MultiSelectField } from './MultiSelectField';
+export { default as SelectField } from './SelectField';
```
priyajeet: *Field already taken :(

flow-typed/npm/formik_vx.x.x.js
```diff
@@ -0,0 +1,283 @@
+declare module "formik" {
```
priyajeet: Had to copy paste this from an older version

src/components/checkbox/__tests__/CheckboxField-test.js
```diff
@@ -0,0 +1,33 @@
+// @flow
+
+import React from 'react';
+import CheckboxField from '../CheckboxField';
+
+describe('components/checkbox/CheckboxField', () => {
+    const getWrapper = (props = {}) => shallow(<CheckboxField {...props} />);
+
+    test('should render properly when turned on', () => {
```
jstoffan: Might be worthwhile to use `test.each` for these, since the value of `value` is the only difference.

src/components/checkbox/__tests__/CheckboxField-test.js
```diff
@@ -0,0 +1,33 @@
...
+        const wrapper = getWrapper({
+            field: {
+                name: 'checkbox',
+                value: 'value',
+                onBlur: 'onblur',
+                onChange: 'onchange',
+            },
+            label: 'Enter things',
+        });
+        expect(wrapper).toMatchSnapshot();
```
jstoffan: Consider explicitly checking `isChecked` instead or as well?

src/components/checkbox/CheckboxField.js
```diff
@@ -0,0 +1,16 @@
...
+import type { FieldProps } from 'formik';
+
+import CheckboxPrimitive from './Checkbox';
+import type { CheckboxProps } from './Checkbox';
+
+type Props = CheckboxProps & FieldProps;
+
+const CheckboxField = ({ field, ...rest }: Props) => {
+    const { value } = field;
+    return <CheckboxPrimitive {...field} {...rest} isChecked={!!value} />;
```
jstoffan: Would it be worth forwarding refs from the field to the inner primitive? If we need to call `input.focus` somewhere, for example?

src/components/select-field/SelectField.js
```diff
@@ -0,0 +1,61 @@
...
+
+function createFakeSyntheticEvent(name: string, value: SelectOptionValueProp | Array<SelectOptionValueProp>) {
+    return {
+        currentTarget: { name, value },
+        target: { name, value },
+    };
+}
+
+function onSelect(name: string, onChange: Function, options: SelectOptionProp | Array<SelectOptionProp>) {
+    if (Array.isArray(options)) {
```
jstoffan: Minor, but it seems like this could live as a ternary to avoid duplicating the `onChange` calls.

src/components/text-area/TextAreaField.js
```diff
@@ -0,0 +1,20 @@
...
+import type { FieldProps } from 'formik';
+
+import TextAreaPrimitive from './TextArea';
+import type { TextAreaProps } from './TextArea';
+
+type Props = TextAreaProps & FieldProps;
+
+const TextAreaField = ({ field, form, isRequired, ...rest }: Props) => {
+    const { name } = field;
+    const { errors, touched } = form;
```
jstoffan: This means errors will show as a user tabs through a form, right? Do we want `touched` or some combination of `dirty`/`isSubmitted`?

src/components/toggle/Toggle.js
```diff
@@ -56,8 +62,10 @@ const Toggle = ({
                     className="toggle-simple-input"
                     disabled={isDisabled}
                     name={name}
+                    onBlur={onBlur}
                     onChange={onChange}
                     type="checkbox"
+                    value={value}
```
jstoffan: Is this going to cause issues with existing usages of this component?

priyajeet: I dont expect it since Toggles are controlled by the parent and rely on the checked/isOn boolean. But I can remove it for now, its dud anyways.

src/components/text-area/TextAreaField.js
priyajeet: I can talk to design, but we can change this later too. If they are tabbing through and not filling out a required field, then showing error might be fine.

alexkrolick:
Would it be possible to update the PillSelector dropdown to conform to this API as well?

priyajeet:
@alexkrolick future patch for more fields, this is just preliminary. So pill selector + maybe a few more soon.

PR #1151, opened 2019-04-26
feat(versions): Add support for file version downloads

jstoffan:
Todo:
- [x] Fix broken tests (4)
- [x] Add unit tests for new logic


boxcla:

src/api/File.js
```diff
@@ -32,34 +34,56 @@ class File extends Item {
...
-            .then(({ data }: { data: BoxItem }) => {
-                this.successHandler(data[FIELD_DOWNLOAD_URL]);
+            .then(async ({ data }: { data: BoxItem }) => {
+                const dataUrl = data[FIELD_AUTHENTICATED_DOWNLOAD_URL];
+                const typedId = getTypedFileId(id);
+                const token: TokenLiteral = await TokenService.getReadToken(typedId, this.options.token);
+                const tokenString: ?string = token && (typeof token === 'string' ? token : token.read);
+
+                if (!dataUrl || !tokenString) {
+                    this.errorHandler({ code: this.errorCode });
```
jstoffan: I need to add a `return` after this statement.

src/api/File.js
```diff
@@ -32,34 +34,56 @@ class File extends Item {
...
+    ): Promise<void> {
         this.errorCode = ERROR_CODE_GET_DOWNLOAD_URL;
-        this.successCallback = successCallback;
         this.errorCallback = errorCallback;
+        this.successCallback = successCallback;
+
         return this.xhr
             .get({
-                url: this.getUrl(id),
+                url: this.getUrl(id, versionId),
```
jstoffan: The relationship between the `File` and `Versions` API modules is a little awkward. I think the ideal solution would be to introduce a `FileVersion` module that extends `File` and shares the base download logic. Not sure I want to tackle that right now, though.

src/api/File.js
```diff
@@ -40,30 +43,45 @@ class File extends Item {
...
+        const downloadAvailable = fileOrFileVersion[FIELD_IS_DOWNLOAD_AVAILABLE];
+        const downloadUrl = fileOrFileVersion[FIELD_AUTHENTICATED_DOWNLOAD_URL];
+
+        if (!downloadAvailable || !downloadUrl) {
+            this.errorHandler({ code: this.errorCode, message: 'Item to download missing required fields.' });
+            return;
+        }
+
+        const token: TokenLiteral = await TokenService.getReadToken(getTypedFileId(fileId), this.options.token);
+        const tokenString: ?string = token && (typeof token === 'string' ? token : token.read);
```
ConradJChan: Is this check needed? It looks like `getReadToken` already does a check to see if token is an object and to return the read property

src/elements/content-preview/preview-header/Header.scss
```diff
@@ -32,7 +53,7 @@
     }
 
     .bp-base-header {
-        padding: 0 0 0 15px;
+        padding: 0 15px;
```
ConradJChan: Where was the right 15px padding accounted for previously?

src/api/File.js
jstoffan: Indeed, nice catch. I was overcomplicating the return value due to a blind copy-paste. Will fix in the next commit.

src/elements/content-preview/preview-header/Header.scss
jstoffan: It wasn't, which was causing the button and title to be spaced differently, which is what this change aims to resolve.

PR #1153, opened 2019-04-26
feat(badgeable): Pass down props to children

stevenzyang:


boxcla:

src/components/badgeable/__tests__/Badgeable-test.js
```diff
@@ -10,6 +10,18 @@ describe('components/badgeable/Badgeable', () => {
         expect(wrapper).toMatchSnapshot();
     });
 
+    test('should pass down props to children when it is an element', () => {
+        const children = <div id="render">Some text to render</div>;
+        const wrapper = shallow(<Badgeable foo="bar">{children}</Badgeable>);
+        // console.log(wrapper.prop("foo"));
```
booc0mtaco: take this line out to tidy up

src/components/badgeable/__tests__/Badgeable-test.js
```diff
@@ -10,6 +10,18 @@ describe('components/badgeable/Badgeable', () => {
         expect(wrapper).toMatchSnapshot();
     });
 
+    test('should pass down props to children when it is an element', () => {
+        const children = <div id="render">Some text to render</div>;
+        const wrapper = shallow(<Badgeable foo="bar">{children}</Badgeable>);
```
booc0mtaco: To make this more deliberate, let's `shallow` and include the output for `children` in the wrapper directly.

```
shallow(
  <Badgeable foo="bar">
    <div id="render">etc.</div>
  </Badgeable>
);

Let's also add a test to define what should happen if `children` is more than one Element.

File: src/components/badgeable/Badgeable.js [hunk 3]
```diff
@@ -18,11 +18,19 @@ type Props = {
...
+        topRight = null,
+        bottomLeft = null,
+        bottomRight = null,
+        ...rest
+    } = props;
 
     return (
         <div className={`badgeable-container ${className}`}>
-            {children}
+            {React.isValidElement(children) ? React.cloneElement(React.Children.only(children), rest) : children}
```
jstoffan: What is the purpose of this change? I would think that since `children` should always be declarative (eslint rule), props could be passed directly rather than via `Badgeable`.

stevenzyang: We have a pattern in EUA of using `HeaderToggleButton` to wrap a button and pass a `color` prop depending on the state of the header and the theme. For NotificationCenterButton and TaskFlyout we wrap these buttons in a `Badgeable` as well. Due to this, there's currently a bug where the the inner icon element is not receiving the correct color prop from `HeaderToggleButton`, and this would fix that issue.

src/components/badgeable/__tests__/Badgeable-test.js
```diff
@@ -4,12 +4,37 @@ import Badgeable from '../Badgeable';
...
+        );
+        expect(
+            wrapper
+                .find('#render')
+                .first()
+                .prop('foo'),
+        ).toEqual('bar');
+    });
+
+    test('should throw an error when there are multiple children', () => {
```
jstoffan: The component is losing the ability to render text and multiple components, which seems like a breaking change? Does these features get used anywhere internally or externally?

src/components/badgeable/Badgeable.js
jstoffan: It sounds like the `HeaderToggleButton` component should be improved to allow arbitrary content to be rendered rather than decorated, if that's the case. Would it be possible to add a render prop to `HeaderToggleButton` so you can access the `color` prop?

stevenzyang:
Following @jstoffan's suggestion to modify HeaderToggleButton to accept a render prop instead

PR #1155, opened 2019-04-26
fix(droppable): Bind drag drop handlers when wrapped element changes

baipenghan:
This change fixes the issue that when Wrapped Component changes itself, all the drag and drop event listeners are added to the new element, thus solving the Ghost State issue where event listeners are attached to Ghost State Component first but not attached to FilesPage Component


boxcla:

cweeii:
Can you make sure there are spaces between the arrange/act/assert sections?  It's a little difficult to read the tests when they're together.

src/elements/common/droppable/makeDroppable.js
```diff
@@ -62,22 +62,37 @@ const makeDroppable = ({ dropValidator, onDrop }: { dropValidator?: Function, on
...
+         * event listeners on the new droppableEl
          */
         bindDragDropHandlers = () => {
             const droppableEl = findDOMNode(this); // eslint-disable-line react/no-find-dom-node
-            if (!droppableEl || !(droppableEl instanceof Element)) {
+            if (!droppableEl || !(droppableEl instanceof Element) || droppableEl === this.droppableEl) {
                 return;
             }
 
+            if (this.droppableEl) {
```
jstoffan: Can we move this to `componentDidUpdate`, since it's only relevant for that lifecycle method?

baipenghan: Sure. makes sense

PR #1156, opened 2019-04-26
fix(tasks): use presence of next page to compute overflow label

alexkrolick:
- Uses next_marker to detect pagination
- Fixes small bug where + is displayed initially instead of the count (+x)


boxcla:

PR #1159, opened 2019-04-27
fix(colors): add #eaeaea to neutral scale

alexkrolick:
- Updates the Box Design Language color scale to [include #eaeaea](https://github.com/box/box-ui-elements/pull/1159/files#diff-38afb9466705304822f75481bbc99308R15)
- Replaces direct usage of #eaeaea with the variable
- Updates value of neutral-03
- Updates color of activity feed item separator to the new neutral-04 color


boxcla:

PR #1160, opened 2019-04-29
feat(datetime): Add date converter from ISO8601 to RFC3339

booc0mtaco:
- convert ISO8601 long format (with zone description) to RFC3339 (fixed zone description format)

### DIFF
--- a/src/utils/__tests__/datetime-test.js
+++ b/src/utils/__tests__/datetime-test.js
@@ -1,9 +1,11 @@
 import {
     convertToMs,
+    convertISOStringtoRFC3339String,
     convertDateToUnixMidnightTime,
     convertISOStringToUTCDate,
     isToday,
     isTomorrow,
+    isValidDate,
     isYesterday,
     isCurrentYear,
     formatTime,
@@ -131,6 +133,19 @@ describe('utils/datetime', () => {
         });
     });
 
+    describe('isValidDate()', () => {
+        test.each([
+            ['2019-01-01', true],
+            ['2019-01-01T12:34:56', true],
+            ['2019-01-01T09:41:56-07:00', true],
+            ['some random string', false],
+            ['', false],
+        ])('should interpret %s as a %p date', (dateString, expected) => {
+            const date = new Date(dateString);
+            expect(isValidDate(date)).toBe(expected);
+        });
+    });
+
     describe('addTime()', () => {
         test('should correctly add time', () => {
             const TEN_MIN_IN_MS = 600000;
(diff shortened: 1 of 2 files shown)


boxcla:

src/utils/__tests__/datetime-test.js
```diff
@@ -160,4 +161,30 @@ describe('utils/datetime', () => {
...
+        });
+
+        test.each([
+            ['bad date'],
+            [''],
+            ['2018-01-01'],
+            ['2019-01-02T00:00:00'],
+            ['2019-01-03T00:00:00.000'],
+            ['2019-01-04T00:00:00.000z'],
+        ])('should throw on bad date format: %s', () => {
```
jstoffan: It doesn't look like the parameterized value is used in the test.

src/features/metadata-instance-editor/fields/DateField.js
```diff
@@ -18,29 +19,41 @@ type Props = {
...
-        value={typeof dataValue === 'string' ? convertISOStringToUTCDate(dataValue) : undefined}
-    />
-);
+const DateField = ({ dataKey, dataValue, displayName, description, intl, onChange, onRemove }: Props) => {
+    // if the JS engine thinks there's an invalid date, convert to known format if possible
+    let parsedDataValue = dataValue;
+
+    try {
+        parsedDataValue =
+            !isNaN(dataValue) && typeof dataValue === 'string' ? convertISOStringtoRFC3339String(dataValue) : dataValue;
```
jstoffan: Can this logic be encapsulated in the convert function? Otherwise I would assume every client/component will need to duplicate it.

src/features/metadata-instance-editor/fields/DateField.js
```diff
@@ -18,29 +19,41 @@ type Props = {
...
+    } catch (err) {
+        // no-op; keep the existing dataValue
```
jstoffan: Can we just assign the default value here rather than having the comment? `parsedDataValue = dataValue`?

booc0mtaco: I could do that

src/utils/__tests__/datetime-test.js
booc0mtaco: good catch!

src/features/metadata-instance-editor/fields/DateField.js
booc0mtaco: Actually, there are some factors that create the following scenario:

- I can't define a `let` and `var` inside the try-catch due to variable scope and "no-var"/"vars-on-top" respectively
- if I put a `let` at the top, I could define it with no default, and set it only in the try-catch. this adds one extra line, and achieves the same behavior as what I have already
- I don't think it's beneficial to both define `let parseDataValue = dataValue` at line 24, then set it again in the `catch` to the same value

I'm inclined to leave this one as-is

src/features/metadata-instance-editor/fields/DateField.js
booc0mtaco: Yah, I considered that. the conversion is harmless, but I also didn't want to apply it globally on LN53. If I embed this check in the conversion function, I'd have to use it for every DatePicker, even for browsers that support the ISO8601 date formats.

It occurred to me a better way to use this might be to embed it in the existing conversion functions instead of importing it in the client components. I'll try that for comparison

src/features/metadata-instance-editor/fields/DateField.js
booc0mtaco: removed this, and make the checks simpler

booc0mtaco:
@jstoffan adding one more test and re-using the matcher regex to avoid having two in this file. should be more simplified now, and reduces redundancy

PR #1161, opened 2019-04-29
feat(tasks): Add status section to task card

mattnguyen1:


boxcla:

mattnguyen1:
Screen Shot 2019-04-29 at 1 19 29 AM

i18n/en-US.properties
```diff
@@ -424,14 +424,16 @@ be.tasks.feed.headline.approval.currentUser = {user} assigned you an Approval Ta
...
+be.tasks.feed.inProgressLabel = In Progress
 # Label for button to expand flyout to see all task assignees
 be.tasks.feed.moreAssigneesLabel = See all assignees
 # Reject option for an approval task
 be.tasks.feed.rejectAction = Reject
 # Label for a rejected task
-be.tasks.feed.rejectedLabel = Task Rejected
+be.tasks.feed.rejectedLabel = Rejected
+# Label for the task status
+be.tasks.feed.statusLabel = Status:
```
alexkrolick: Looks good, my only comment is I _think_ when you use `:` you should put a variable after instead of concatenating the strings in code, because different languages might have different ways of treating this grammar

src/elements/content-sidebar/activity-feed/task-new/TaskStatus.js
```diff
@@ -29,30 +28,43 @@ const StatusIcon = ({ status, ...rest }: { status: TaskStatus }) => {
...
-            return <IconPending width={ICON_SIZE} height={ICON_SIZE} {...rest} />;
         default:
             return null;
     }
 };
 
 const StatusMessage = ({ status }: { status: TaskStatus }) => {
+    let statusMessage;
     switch (status) {
         case TASK_NEW_APPROVED:
```
jstoffan: Seems like this could be simplified to a map with key/value of status/message and moved to a constant.

src/elements/content-sidebar/activity-feed/task-new/Task.scss
```diff
@@ -74,16 +74,37 @@ $card-left-offset: 40px;
 
     .bcs-task-status {
         align-items: center;
+        color: $bdl-neutral-02;
         display: flex;
-        margin-top: 18px;
+        font-weight: bold;
+
+        > span {
```
alexkrolick: Is there some way we could avoid these brittle selectors? Maybe adding a className onto the status components?

src/elements/content-sidebar/activity-feed/task-new/Task.scss
```diff
@@ -112,6 +133,10 @@ $card-left-offset: 40px;
         margin-top: 5px;
     }
 
+    .bcs-comment-container {
+        margin-bottom: 8px;
```
alexkrolick: If we stick to margin-top for spacing, it makes it easier to compose more elements.

src/elements/content-sidebar/activity-feed/task-new/Task.scss
mattnguyen1: I'm not sure if there's a good way to avoid this; `> span` is to refer to the span that is created for `<FormattedMessage />`, which doesn't take in a class name unfortunately. Do you know if there's a better way to do this?

mattnguyen1: I can probably update the selector `> *:first-child` to be more specific by adding a class though as per your suggestion.

alexkrolick: You could wrap the FormattedMessage in another span with a className, or pass a function as child: https://github.com/yahoo/react-intl/wiki/Components#formattedmessage

mattnguyen1: Looks doing the function way works for me 💯

src/elements/content-sidebar/activity-feed/task-new/TaskStatus.js
```diff
@@ -29,33 +36,31 @@ const StatusIcon = ({ status, ...rest }: { status: TaskStatus }) => {
             return <IconReject width={ICON_SIZE} height={ICON_SIZE} {...rest} />;
         case TASK_NEW_NOT_STARTED:
         case TASK_NEW_IN_PROGRESS:
-            return <IconPending width={ICON_SIZE} height={ICON_SIZE} {...rest} />;
```
jstoffan: @alexkrolick , just to confirm: Pending status is no longer needed? It was added recently, iirc.

alexkrolick: correct, it doesn't show an icon for the initial states

src/elements/content-sidebar/activity-feed/task-new/Task.scss
```diff
@@ -72,18 +71,38 @@ $card-left-offset: 40px;
...
+
     .bcs-task-status {
         align-items: center;
+        color: $bdl-neutral-02;
         display: flex;
-        margin-top: 18px;
+        font-weight: bold;
+
+        .bcs-task-status-item:first-child {
+            margin-left: 5px;
```
alexkrolick: May be missing something, but why does it need :first-child? Can the item itself have the padding?

mattnguyen1: The `In Progress` status doesn't have an icon with it, so either the icon or the status message needs the padding

mattnguyen1: We can change this in a follow-up

PR #1164, opened 2019-04-30
feat(tooltip): expose position instance API on tooltips

drichard-box:


boxcla:
Verified that @drichard-box has signed the CLA. Thanks for the pull request!

src/components/radar/RadarAnimation.js
```diff
@@ -76,8 +78,22 @@ class RadarAnimation extends React.Component<Props> {
         position: MIDDLE_RIGHT,
     };
 
+    tetherRef = React.createRef<{ position: () => {} }>();
+
     radarAnimationID = uniqueId('radarAnimation');
 
+    isShown = () => {
+        const { isShown = true } = this.props;
```
DanDeMicco: use `defaultProps`, this doesn't need to be a function if its a just returning a prop

src/components/radar/RadarAnimation.js
```diff
@@ -76,8 +78,22 @@ class RadarAnimation extends React.Component<Props> {
         position: MIDDLE_RIGHT,
     };
 
+    tetherRef = React.createRef<{ position: () => {} }>();
```
DanDeMicco: I think this should be `React.createRef<TetherComponent>`

src/components/radar/RadarAnimation.js
drichard-box: its defaulting the value if the prop isnt provided. id prefer not using defaultProps since it makes it less clear which props are optional.

src/components/radar/RadarAnimation.js
drichard-box: best case it would be `React.createRef<typeof TetherComponent>`, but TetherComponent isnt typed so this is better

src/components/radar/RadarAnimation.js
DanDeMicco: Not sure if I agree that defaultProps makes it less clear. Its a react standard and the above flow types explain that is is optional

src/components/radar/RadarAnimation.js
DanDeMicco: If it became typed I think this would cause an error wouldn't it?

src/components/radar/RadarAnimation.js
drichard-box: yeah, the fact that it needs a comment is kind of the issue... otherwise callers are forced to compare the Props and the defaultProps in order to determine which are optional and which aren't. 

seems like its being done here already though, so im fine using it if there is a strong feeling towards it.

src/components/radar/RadarAnimation.js
drichard-box: If what became typed?

src/components/radar/RadarAnimation.js
DanDeMicco: I guess its a matter of opinion but personally I don't agree the way you have it is clearer. The consumer would have to look at the component in the `isShown` to determine what the default value is vs looking at the `defaultProps`. I would prefer to stick to the existing pattern of using `defaultProps`

src/components/radar/RadarAnimation.js
DanDeMicco: I meant `TetherComponent`. I was thinking if the flow types got added for it, it would break. Anyways this is fine since the typing doesn't do much anyways

src/components/radar/RadarAnimation.js
drichard-box: done

src/components/tooltip/Tooltip.js
```diff
@@ -160,6 +169,22 @@ class Tooltip extends React.Component<Props, State> {
...
+    isControlled = () => {
+        const { isShown: isShownProp } = this.props;
+        return typeof isShownProp !== 'undefined';
+    };
+
+    isShown = () => {
+        const { isShown: isShownProp } = this.props;
+        const isControlled = this.isControlled();
+
+        const isShown = isControlled ? isShownProp : this.state.isShown;
```
DanDeMicco: I know this is just a refactor, but this seems a bit strange to me. Should probably implement `getDerivedStateFromProps` or `componentDidUpdate`

PR #1167, opened 2019-04-30
feat(notifications): Add stroke/fill classes to NC states

mattnguyen1:
- This aligns the two icons with the same classes used for other states in BUIE, which enables for the states to use the themed (custom branding) css.


boxcla:

PR #1168, opened 2019-04-30
fix(modal): fix focus issue

Visylvius:
What this PR does: 
This always guarantees that the field that is focused in the modal starts on the left, rather than the right.

### DIFF
--- a/src/components/text-input-with-copy-button/TextInputWithCopyButton.js
+++ b/src/components/text-input-with-copy-button/TextInputWithCopyButton.js
@@ -149,7 +149,7 @@ class TextInputWithCopyButton extends React.PureComponent<Props, State> {
 
     handleFocus = (event: SyntheticEvent<>) => {
         if (this.copyInputRef) {
-            this.copyInputRef.select();
+            this.performAutofocus();
         }
 
         if (this.props.onFocus) {
@@ -167,8 +167,10 @@ class TextInputWithCopyButton extends React.PureComponent<Props, State> {
     };
 
     performAutofocus = () => {
-        if (this.copyInputRef) {
-            this.copyInputRef.select();
+        const { copyInputRef } = this;
+        if (copyInputRef) {
+            copyInputRef.select();
+            copyInputRef.scrollLeft = 0;
         }
     };
 


boxcla:

jstoffan:
@Visylvius, can you please clarify the commit message and correct it scope. It doesn't seem to be exclusively related to a modal.

src/components/text-input-with-copy-button/TextInputWithCopyButton.js
```diff
@@ -150,6 +150,9 @@ class TextInputWithCopyButton extends React.PureComponent<Props, State> {
     handleFocus = (event: SyntheticEvent<>) => {
         if (this.copyInputRef) {
```
jstoffan: I realize this code was already duplicated, but we're adding to it a bit. Could this method also call `this.performAutoFocus()` directly?

src/components/text-input-with-copy-button/TextInputWithCopyButton.js
```diff
@@ -150,6 +150,9 @@ class TextInputWithCopyButton extends React.PureComponent<Props, State> {
...
             this.copyInputRef.select();
+            if (this.copyInputRef && this.copyInputRef.scrollLeft) {
```
jstoffan: It looks like we access `this.copyInputRef` directly on the line above. Is the existence check necessary and, if so, should it be moved up?

Visylvius: The reason for this check is for the flow error, for when the input is null. Adding this check stopped it from complaining

src/components/text-input-with-copy-button/TextInputWithCopyButton.js
Visylvius: It certainly could.

src/components/text-input-with-copy-button/TextInputWithCopyButton.js
jstoffan: It's already checked on line 151, so this code is unreachable if it's falsy. I'm not seeing a flow error when I make the change on my local. Can you try it once more?

Visylvius:
Ah, my apologies. Went through a few iterations, and didn't make the changes to modal. I will glady change the commit message.

src/components/text-input-with-copy-button/TextInputWithCopyButton.js
jstoffan: It looks like flow assumes that calling a method could destroy the object, even though the return of `.select()` isn't assigned to the class property. You can do something like this to get around it:

```
const { copyInputRef } = this;

if (copyInputRef) {
    copyInputRef.select();
    copyInputRef.scrollLeft = 0;
}
```

Hooray Flow.

alexkrolick: That's an interesting find. I wonder what would happen if this was using the createRef API instead of function refs

Visylvius: That is interesting. I can definitely give that a try. 

Thats a good question Alex, we probably we should probably switch to the createRef api.

PR #1169, opened 2019-05-01
feat(tasks): Update task status to not use an icon

mattnguyen1:
- Design updated for the task status to no longer include an icon


boxcla:

mattnguyen1:
Screen Shot 2019-05-01 at 4 48 07 PM

PR #1171, opened 2019-05-02
refactor: Upgrade react and react-dom to v16.8 and refactor Avatar

j3tan:
BREAKING CHANGE: peerDep React requirement have been bumped higher


boxcla:

src/elements/content-sidebar/activity-feed/task-new/Assignees.js
```diff
@@ -73,12 +73,7 @@ class Assignees extends React.Component<Props> {
                             />
                         }
                     >
-                        <AssigneeStatus
```
j3tan: for whatever reason, the `data-testid` in the `wrapper.find` was returning both this top level function and the nested `div` causing 2x the number of objects returned. moving the data-testid into the child component fixes that without impacting usage

src/elements/content-sidebar/__tests__/ContentSidebar-test.js
```diff
@@ -164,10 +164,6 @@ describe('elements/content-sidebar/ContentSidebar', () => {
         });
 
         test('should set the state with the file and view and then call fetchMetadata', () => {
-            wrapper = getWrapper();
```
j3tan: unnecessary due to beforeEach, also was causing a random failure

package.json
```diff
@@ -175,7 +175,7 @@
         "deepmerge": "^2.1.1",
         "draft-js": "^0.10.1",
         "enzyme": "3.8.0",
-        "enzyme-adapter-react-16": "^1.10.0",
+        "enzyme-adapter-react-16": "1.12.1",
```
jstoffan: Does this need to be pinned to this specific version?

src/components/avatar/Avatar.js
```diff
@@ -30,50 +30,38 @@ type Props = {
...
-        let avatar;
-        if (avatarUrl && !hasImageErrored) {
-            avatar = <AvatarImage onError={this.onImageError} url={avatarUrl} />;
-        } else if (name) {
-            avatar = <AvatarInitials id={id} name={name} />;
-        } else {
-            avatar = <UnknownUserAvatar className="avatar-icon" />;
+    // Reset hasImageErrored state when avatarUrl changes
+    useEffect(() => {
+        if (hasImageErrored) {
```
jstoffan: Is this check necessary or should we just reset `setHasImageErrored` to false whenever the url changes?

src/components/avatar/__tests__/Avatar-test.js
```diff
@@ -66,18 +72,27 @@ describe('components/avatar/Avatar', () => {
...
+            wrapper = mount(<Avatar {...props} />);
+        });
+        expect(wrapper.find(MockAvatarImage).length).toEqual(1);
 
-        wrapper.setProps({
-            ...props,
-            avatarUrl: 'http://foo.bar/baz123_invalid_new',
+        act(() => {
+            const avatarImage = wrapper.find(MockAvatarImage);
+            avatarImage.prop('onError')();
```
jstoffan: It looks like this is manually invoking the `onError` prop callback? Can we use `.simulate` or `.simulateError`?

src/components/avatar/__tests__/Avatar-test.js
```diff
@@ -66,18 +72,27 @@ describe('components/avatar/Avatar', () => {
             avatarUrl: 'http://foo.bar/baz123_invalid',
         };
 
-        const wrapper = shallow(<Avatar {...props} />);
-
-        wrapper.instance().onImageError();
-        expect(wrapper.state('hasImageErrored')).toEqual(true);
+        let wrapper;
+        act(() => {
```
jstoffan: It seems a bit arbitrary when we're using `act` or not. Can we create a `getWrapper` helper that abstracts it away?

src/components/avatar/__tests__/Avatar-test.js
j3tan: `enzyme` devs advocate against using `simulate`. They actually recommend calling the prop directly.

src/components/avatar/Avatar.js
j3tan: good point. since this only fires when url changes, there should be no downside to just calling this always

src/components/avatar/__tests__/Avatar-test.js
j3tan: yes, but i'm not sure how that would help the additional calls to `act`.

```
const wrapper = getWrapper(<Avatar {...props} />);
wrapper.find(...);
// Auto call wrapper.update after
wrapper.act({
   ....
})
```
Something like that?

src/components/avatar/__tests__/Avatar-test.js
j3tan: https://github.com/airbnb/enzyme/issues/1606

package.json
j3tan: not fixed, can change this to ^

src/components/avatar/__tests__/Avatar-test.js
j3tan: I think for now, we should just use `act` as is and wait for enzyme to fix the update functionality.

j3tan:
> Finally, `chore` commits don't mix very well with breaking changes, since that scope doesn't usually trigger new releases. Consider using `feat`, `refactor`, etc., instead.

yeah, actually just copied the previous react upgrade PR message, but I can update to `refactor`

package.json
```diff
@@ -175,7 +175,7 @@
         "deepmerge": "^2.1.1",
         "draft-js": "^0.10.1",
         "enzyme": "3.8.0",
-        "enzyme-adapter-react-16": "^1.10.0",
+        "enzyme-adapter-react-16": "^1.12.1",
         "enzyme-to-json": "^3.3.4",
         "eslint": "^4.19.1",
```
jeremypress: Has FTL discussed adding the lint plugin for hooks?
https://www.npmjs.com/package/eslint-plugin-react-hooks
https://reactjs.org/docs/hooks-overview.html#rules-of-hooks

j3tan: yes, we will be adding these plugins. I'll discuss it at the next FTL meeting but it'll be an update to @box/frontend

PR #1172, opened 2019-05-02
feat(inline-edit): user should see instance field input

ChenCodes:
Changes in this PR:
- Pass className prop to ValueField
- Use Date.parse due to an error that showed up with the original implementation (explanation in review comments)


boxcla:

src/features/query-bar/components/filter/ValueField.js
```diff
@@ -25,8 +26,9 @@ const getDateValue = selectedValues => {
     }
 
     const value = selectedValues[0];
-    if (value instanceof Date) {
-        return value;
+
+    if (Date.parse(String(value))) {
+        return new Date(value);
```
ChenCodes: The reason for the change here is that value was coming back as the string ("April 19, 2019") and not as a Date object which was causing an error.

The correct way is to first Date.parse the string to convert it to the number of milliseconds (as a string), and then return the Date object as that's what the DatePicker expects as the value prop.

krapspark: `Date.parse` looks like it returns different results based on implementation.
Do we use any libraries that validates dates?

src/features/list-view/styles/ListView.scss
```diff
@@ -51,4 +51,12 @@
     .ReactVirtualized__Grid {
         outline: none;
     }
+
+    .ReactVirtualized__Grid__innerScrollContainer {
```
krapspark: Is there any way we can pass a class of some kind into `ReactVirtualized` so that we can style that instead of this class directly? It's usually undesirable to style a class we don't own.

ChenCodes: Discussed with Sam, will be adding a TODO regarding the limitations of our current approach and plans for a more ideal solution.

src/features/query-bar/components/filter/ValueField.js
ChenCodes: Will be reverting this change as it is not needed. It turns out formatDate was being used in the parent app incorrectly causing us to have to reformat it into a Date object.

src/features/query-bar/components/filter/ValueField.js
```diff
@@ -58,6 +60,9 @@ const ValueField = ({ onChange, selectedValues, valueOptions, valueType }: Props
             return (
                 <div className="filter-dropdown-text-field-container">
                     <TextInput
+                        className={className}
```
macrotim: Why is className set on an inner element? This effectively bars a future solution where clients can select the outermost <div/>.

ChenCodes: Classname is set here since there are unique styles that are required for each input field to make it look correct in the Grid.

macrotim: Is it possible to solve what you need _and_ put the className on the outermost div?

ChenCodes: I could have the className on L61 take in one of many possible selectors:

```'filter-dropdown-text-field-container text-input'
'filter-dropdown-text-field-container date-input'
'filter-dropdown-text-field-container select-field'```

Comment (macrotim) [reply src/features/query-bar/components/filter/ValueField.js]: .InstanceFieldInput .text-input-container {
}

Ideally, SUIT-friendly CSS would look like...
.InstanceFieldInput .ValueField .TextInput {
}

File: src/features/query-bar/components/filter/ValueField.js [hunk 4]
```diff
@@ -25,8 +28,9 @@ const getDateValue = selectedValues => {
     }
 
     const value = selectedValues[0];
-    if (value instanceof Date) {
-        return value;
+    const formattedDate = new Date(value);
```
macrotim: What makes this a "formatted date"?

macrotim:
Are no unit tests affected? Should we test the props.error?

src/features/query-bar/components/filter/ValueField.js
ChenCodes: Maybe 'date' would be a better name.

src/features/list-view/styles/ListView.scss
```diff
@@ -24,13 +24,16 @@
             cursor: pointer;
         }
 
-        .bdl-icon-sort-chevron.bdl-ListView-isSortAsc {
+        .bdl-icon-sort-chevron {
```
ChenCodes: Added this change in this PR since I didn't think we need a separate PR just for a small change. Will link this PR with the relevant ticket that is related to the icon being 12px by 12px.

src/features/query-bar/__tests__/ValueField-test.js
```diff
@@ -72,5 +72,34 @@ describe('features/query-bar/components/filter/ValueField', () => {
...
+
+        describe('when there is an error due to an invalid input', () => {
+            describe('should not show an error for a', () => {
+                test.each`
+                    valueType        | componentName          | should
+                    ${'multiSelect'} | ${'MultiSelectField'}  | ${'MultiSelectField of type multiSelect'}
+                    ${'enum'}        | ${'SingleSelectField'} | ${'SingleSelectField of type enum'}
+                    ${'date'}        | ${'DatePicker'}        | ${'DatePicker of type date'}
+                `('$should', ({ componentName, valueType }) => {
+                    const wrapper = getWrapper({ valueType, error: undefined, selectedValues: [] });
```
macrotim: Programmers should specify null instead of undefined.

src/features/query-bar/__tests__/ValueField-test.js
```diff
@@ -72,5 +72,34 @@ describe('features/query-bar/components/filter/ValueField', () => {
                 expect(component.prop(valuePropNames[componentName])).toEqual(expectedValue);
             });
         });
+
+        describe('when there is an error due to an invalid input', () => {
+            describe('should not show an error for a', () => {
```
macrotim: I need help to understand from the user's pov. Why omit error when invalid input?

macrotim: We agreed to delete line 76.

ChenCodes: Discussed in person and decided that we can get rid of L76.

src/features/query-bar/components/filter/ValueField.js
```diff
@@ -13,6 +14,8 @@ import type { ConditionValueType } from '../../flowTypes';
 import '../../styles/Condition.scss';
 
 type Props = {
+    className?: string,
```
macrotim: Remove

PR #1173, opened 2019-05-02
fix(tasks): autofocus first input in task form

alexkrolick:


boxcla:

src/elements/content-sidebar/AddTaskButton.js
```diff
@@ -63,12 +63,16 @@ class AddTaskButton extends React.Component<Props, State> {
...
+        // CSS selector for first form element
+        // Note: Modal throws an error if this fails to find an element!
+        const focusTargetSelector = '.task-modal #taskAssignees';
         return (
             <React.Fragment>
                 <AddTaskMenu isDisabled={isDisabled} onMenuItemClick={this.handleClickMenuItem} />
                 <Modal
                     className="be-modal task-modal"
                     data-testid="create-task-modal"
+                    focusElementSelector={focusTargetSelector}
```
jstoffan: It looks like we just use `focusElementSelector="input"` or `focusElementSelector=".someClassName input"` in other places. Would that work here? It would allow us to avoid adding the id to the inputProps.

alexkrolick: Ok, sure. The focusElementSelector thing is pretty dangerous since it throws if it fails.

PR #1175, opened 2019-05-03
fix(menu): Menu should only autofocus on mount if visible

jstoffan:
This will also resolve the issue where the Components section of the examples site immediately jumps to the Menu documentation.


boxcla:

PR #1177, opened 2019-05-04
feat(icon): add new globe icon

DanDeMicco:
Screen Shot 2019-05-05 at 9 37 01 AM


boxcla:

priyajeet:
There is only 1-2 places the older icon is used, we should replace that instead of adding a new one

src/icons/general/IconGlobe2.js
```diff
@@ -0,0 +1,34 @@
...
+        className={`icon-zones ${className}`}
+        height={height}
+        title={title}
+        viewBox="0 0 12 12"
+        width={width}
+    >
+        <g id="Zones_Unselected_12x12" stroke="none" strokeWidth="1" fill="none" fillRule="evenodd">
+            <path
+                d="M6,1 C3.24,1 1,3.24 1,6 C1,8.76 3.24,11 6,11 C8.76,11 11,8.76 11,6 C11,3.24 8.76,1 6,1 Z M5.5,9.965 C3.525,9.72 2,8.04 2,6 C2,5.69 2.04,5.395 2.105,5.105 L4.5,7.5 L4.5,8 C4.5,8.55 4.95,9 5.5,9 L5.5,9.965 Z M8.95,8.695 C8.82,8.29 8.45,8 8,8 L7.5,8 L7.5,6.5 C7.5,6.225 7.275,6 7,6 L4,6 L4,5 L5,5 C5.275,5 5.5,4.775 5.5,4.5 L5.5,3.5 L6.5,3.5 C7.05,3.5 7.5,3.05 7.5,2.5 L7.5,2.295 C8.965,2.89 10,4.325 10,6 C10,7.04 9.6,7.985 8.95,8.695 Z"
+                id="Shape"
```
priyajeet: id: 👎

src/icons/general/IconGlobe2.js
```diff
@@ -0,0 +1,34 @@
...
+type Props = {
+    className?: string,
+    color?: string,
+    height?: number,
+    /** A text-only string describing the icon if it's not purely decorative for accessibility */
+    title?: string | React.Element<any>,
+    width?: number,
+};
+
+const IconGlobe2 = ({ className = '', color = '#8EA6B2', height = 12, title, width = 12 }: Props) => (
```
jstoffan: Can we use a more descriptive name? What makes this different from the current globe icon? Also, most new icons are using the `TwoTonedIcon` props, which encourages styling via CSS rather than props. FYI @tjgupta

src/icons/general/IconGlobe2.js
```diff
@@ -0,0 +1,34 @@
...
+    <AccessibleSVG
+        className={`icon-zones ${className}`}
```
priyajeet: zones?

jstoffan: Is this class name correct?

src/icons/general/IconGlobe2.js
priyajeet: color should come from the variables

src/icons/general/IconGlobe2.js
```diff
@@ -0,0 +1,34 @@
...
+        height={height}
+        title={title}
+        viewBox="0 0 12 12"
+        width={width}
+    >
+        <g id="Zones_Unselected_12x12" stroke="none" strokeWidth="1" fill="none" fillRule="evenodd">
```
jstoffan: Can you run this through an optimizer?

src/icons/general/IconGlobe2.js
```diff
@@ -0,0 +1,34 @@
...
+import * as React from 'react';
+
+import AccessibleSVG from '../accessible-svg';
+
+type Props = {
+    className?: string,
+    color?: string,
+    height?: number,
+    /** A text-only string describing the icon if it's not purely decorative for accessibility */
+    title?: string | React.Element<any>,
```
priyajeet: React.Node

src/icons/general/IconGlobe2.js
priyajeet: Zones_Unselected_12x12 ?

src/icons/general/__tests__/IconGlobe2-test.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+
+import IconGlobe2 from '../IconGlobe2';
+
+describe('icons/general/IconGlobe2', () => {
```
priyajeet: Use snapshot testing for icons

src/icons/general/IconGlobe2.js
```diff
@@ -0,0 +1,34 @@
...
+        title={title}
+        viewBox="0 0 12 12"
+        width={width}
+    >
+        <g id="Zones_Unselected_12x12" stroke="none" strokeWidth="1" fill="none" fillRule="evenodd">
+            <path
+                d="M6,1 C3.24,1 1,3.24 1,6 C1,8.76 3.24,11 6,11 C8.76,11 11,8.76 11,6 C11,3.24 8.76,1 6,1 Z M5.5,9.965 C3.525,9.72 2,8.04 2,6 C2,5.69 2.04,5.395 2.105,5.105 L4.5,7.5 L4.5,8 C4.5,8.55 4.95,9 5.5,9 L5.5,9.965 Z M8.95,8.695 C8.82,8.29 8.45,8 8,8 L7.5,8 L7.5,6.5 C7.5,6.225 7.275,6 7,6 L4,6 L4,5 L5,5 C5.275,5 5.5,4.775 5.5,4.5 L5.5,3.5 L6.5,3.5 C7.05,3.5 7.5,3.05 7.5,2.5 L7.5,2.295 C8.965,2.89 10,4.325 10,6 C10,7.04 9.6,7.985 8.95,8.695 Z"
+                id="Shape"
+                fill={color}
+                fillRule="nonzero"
```
priyajeet: fill-rule default is nonzero iirc, dont need it to be explicit

src/icons/general/IconGlobe2.js
```diff
@@ -0,0 +1,34 @@
+// @flow
```
priyajeet: Should have a corresponding .md file for examples

src/icons/general/IconGlobe2.js
DanDeMicco: I was following existing examples in the naming, there are multiple which follow this style. Its pretty hard to distinguish the differences in words, but I was given this icon to be used for zones. I originally called it `IconZones` but then saw a globe and thought it would be better for it to be generically named

I can remove the color prop to enforce styling via css if that is the direction we are taking, but don't think this icon meets the two toned criteria

DanDeMicco: This color doesn't exist in variables but is used in a few other icons. Can I just add it?

src/icons/general/IconGlobe2.js
drichard-box: does this need to be in box-ui-elements? imo we could hold off until its clear that it is general/foundational enough to be here.

DanDeMicco: After talking with Lu, I am thinking the same thing. Closing this PR

PR #1178, opened 2019-05-06
fix(floatfield): always allow trailing period for float field

krapspark:
FloatField was took an optional `allowTrailingPeriod` which wasn't passed in all field validations. This caused inconsistent behavior.

Since there isn't a clear use case for disallowing trailing period, I'm removing the extra option.


boxcla:
Hi @krapspark, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

krapspark:
CLA signed

PR #1184, opened 2019-05-07
feat(versions): Add support for file version restore actions

jstoffan:
The restore endpoint has an initial spec, but has not yet been built out. This patchset should get us 90% of the way to integration when the endpoint eventually goes live.


boxcla:

PR #1185, opened 2019-05-07
fix(text-input): looses focus when tooltip toggles

priyajeet:
If text is absent for the tooltip it rendered only its children otherwise a tethered component. As such the DOM is re-rendered when tooltip gets text and thus the underlying input loses focus.

This partially reverts #1066


boxcla:

PR #1187, opened 2019-05-08
build: Disable perf hints and pikaday downcompile to avoid warnings

jstoffan:


boxcla:

PR #1189, opened 2019-05-08
fix(tasks): fix task issues

shikharmohan:
- Fixes an issue where the modal is still dirty with previously found contact selectors after closing it.
- Change verified icon green color
- Change 'Assignee(s)' to 'Assignees'
- Change assignee input field error message to say "Invalid User" instead of "Required Field" 
- Remove form validation hooks for mouseenter and focus of the submit button.


boxcla:

PR #1190, opened 2019-05-08
fix(metadataeditor): send cascadeId onRemove

Visylvius:
What this PR Does:

Adds the cascadePolicy Id to the onRemove action. 
This only has to happen because the instance/:id endpoint hasn't been created. Once it is I will remove this from the implementation.


boxcla:

Visylvius:
Talked with Tim in person, we have the relevant information to massage data on the front-end. 
therefore closing this PR.

PR #1191, opened 2019-05-08
fix(datetime): add missing support for times w/o ms

booc0mtaco:
Include a capture group so we can standardize output w.r.t. milliseconds (and match formats that lack milliseconds)


boxcla:

src/utils/datetime.js
```diff
@@ -161,29 +174,25 @@ function convertISOStringtoRFC3339String(isoString: string): string {
...
-        /**
-         * This contains an array structured:
-         * 1) the date/time portion (2018-06-13T00:00:00.000)
-         * 2) the timezone portion (e.g., Z, +03, -0400, +05:00)
-         * 3) the Z format for timezone (if matched)
-         * 4) the short format for timezone (if matched)
-         * 5) the colon-less format for timezone (if matched)
-         * 6) the colon long format for timezone (if matched)
-         */
+        let dateTime = parseDate[ISO8601_DATETIME];
```
jstoffan: Usual convention is for `let` to come after `const`.

PR #1192, opened 2019-05-09
fix(metadata): template dropdown scrollbar issue in safari

krapspark:
The template dropdown uses the Flyout component which wraps a SelectorDropdown component.
Both of these components use the Overlay component in its implementation.
This results in the TemplateDropdown component having two layers of Overlay components.
This isn't a problem for most cases for in Safari causes the vertical scroll bar to disappear because of the nested transform CSS properties.
This change manually removes the CSS transform of the inner Overlay component to fix this bug.


boxcla:
Verified that @krapspark has signed the CLA. Thanks for the pull request!

PR #1193, opened 2019-05-10
fix(versions): Retrieve and use file version permissions for actions

jstoffan:
- [x] Unit tests for new API methods


boxcla:

src/api/Versions.js
```diff
@@ -299,6 +335,25 @@ class Versions extends OffsetBasedAPI {
...
+     */
+    sortVersions(versions: ?FileVersions): ?FileVersions {
+        if (!versions) {
+            return versions;
+        }
+
+        const { entries, total_count } = versions;
+
+        return {
+            entries: [...entries].sort((a, b) => Date.parse(b.created_at) - Date.parse(a.created_at)),
```
jeremypress: Do we need to spread and repackage as an array here instead of just `entries.sort(...`?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -122,23 +120,24 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
...
     };
 
-    handleFetchSuccess = ([file, versions]): [BoxItem, FileVersions] => {
+    handleFetchSuccess = ([fileResponse, versionsResponse]): [BoxItem, FileVersions] => {
         const { api } = this.props;
-        const { entries } = api.getVersionsAPI(false).addCurrentVersion(versions, file) || {};
+        const versionsApi = api.getVersionsAPI(false);
+        const versionsWithCurrent = versionsApi.addCurrentVersion(versionsResponse, fileResponse);
+        const versionsWithPermissions = versionsApi.addPermissions(versionsWithCurrent, fileResponse);
+        const { entries: versions } = versionsApi.sortVersions(versionsWithPermissions) || {};
```
jeremypress: I get the case for the other 2, but would it make sense to sort all versions entries all the time to prevent each caller from doing it?

jstoffan: I figured it'd be best to leave it to the calling component due to the churn we had with sorting items in the activity feed.

PR #1194, opened 2019-05-10
feat(pill-selector): formik wrapper for pill selector

priyajeet:
- Adds check for composition mode in `PillSelectorDropdown`
- Adds ability to add invalid pills in the `PillSelector`
- Changes `Pill`s to have a consistent data structure `{ displayText, value }`
- Adds Formik compatible `PillSelectorDropdownField`.
  - Does not yet have all the `PillSelectorDropdown` API as part of this PR.
  - By default will allow custom inputs.
  - Validator should validate both pills and pill text.
  - Usage:
```jsx
<Field
	label="Pill Selector Field With Dropdown"
	name="pillselectordropdown"
	placeholder="Colors"
	options={[
		{ displayText: 'Red', value: 'red' },
		{ displayText: 'Green', value: 'green' },
		{ displayText: 'Blue', value: 'blue' },
	]}
	component={PillSelectorDropdownField}
	validator={pillSelectorValidator}
	dropdownRenderer={options => options.map(option => (
		<DatalistItem key={option.value} style={{ color: option.value }}>
			{option.displayText}
		</DatalistItem>
	))}
/>
```

pills


boxcla:

src/components/pill-selector-dropdown/PillSelectorDropdown.js
```diff
@@ -57,7 +59,7 @@ type Props = {
     /** Validate the given input value, and update `error` prop if necessary */
     validateForError?: Function,
     /** Called to check if pill item data is valid. The `item` is passed in. */
-    validator?: Function,
+    validator: (text: string) => boolean,
```
Sunyang730: 👏

src/components/pill-selector-dropdown/PillSelector.js
```diff
@@ -7,14 +7,15 @@ import { KEYS } from '../../constants';
 
 import Pill from './Pill';
 import SuggestedPillsRow from './SuggestedPillsRow';
-import type { SelectedOptions, SuggestedPillsFilter } from './flowTypes';
+import type { SelectedPills, SuggestedPillsFilter } from './flowTypes';
```
Sunyang730: I am always confuse on this selectedPills naming.. want to make sure I understand this correctly.. selectedPill means the pill that's currently in the pill selector?

src/components/pill-selector-dropdown/PillSelectorDropdownField.js
```diff
@@ -0,0 +1,127 @@
...
+    /** If true, user can add pills not included in selector options */
+    isCustomInputAllowed: boolean,
+    /** Input label */
+    label: React.Node,
+    /** Array or Immutable list with data for the selected options shown as pills */
+    options: Array<Option>,
+    /** A placeholder to show in the input when there are no pills */
+    placeholder: string,
+    /** Called to check if pill item data is valid. The `item` is passed in. */
+    validator: (text: string) => boolean,
```
priyajeet: This props collection is a lot lot less than what the PillSelectorDropdown.js has. For now I want to keep it small and expand only as need arises.

src/components/pill-selector-dropdown/PillSelectorDropdownField.js
```diff
@@ -0,0 +1,127 @@
...
+    /** CSS class for the component */
+    className?: string,
+    /** If true, the input control is disabled so no more input can be made */
+    disabled?: boolean,
+    /** If true, user can add pills not included in selector options */
+    isCustomInputAllowed: boolean,
+    /** Input label */
+    label: React.Node,
+    /** Array or Immutable list with data for the selected options shown as pills */
+    options: Array<Option>,
```
priyajeet: I am enforcing some specific option format now, unlike existing pill selector that works on arbitrary objects. Having arbitrary objects allows more pill customization but also forces the client using the pill selector to [write all this](https://github.com/box/box-ui-elements/blob/master/src/components/pill-selector-dropdown/README.md).

From what I see is most pill selector use case is [TEXT x] so keeping data structure same as other components like select / multi-select ==> { displayText: something, value: something }

src/components/pill-selector-dropdown/PillSelectorDropdown.js
```diff
@@ -67,33 +69,36 @@ type State = {
...
+        const { allowInvalidPills, parseItems, validator } = this.props;
 
         let pills = parseItems ? parseItems(inputValue) : parseCSV(inputValue);
 
         if (!pills) {
             return [];
         }
 
-        if (validator) {
+        if (validator && !allowInvalidPills) {
```
jstoffan: Doesn't look like the `validator` identity check is required anymore.

src/components/pill-selector-dropdown/PillSelectorDropdown.js
```diff
@@ -192,6 +199,7 @@ class PillSelectorDropdown extends React.Component<Props, State> {
                 selector={
                     <Label text={label}>
                         <PillSelector
+                            allowInvalidPills={allowInvalidPills}
```
jstoffan: Should this live after `{...inputProps}` or is its position up here meaningful?

src/components/pill-selector-dropdown/PillSelectorDropdownField.js
```diff
@@ -0,0 +1,127 @@
...
+
+type State = {
+    selectorOptions: Array<Option>,
+};
+
+class PillSelectorDropdownField extends React.PureComponent<Props, State> {
+    static defaultProps = {
+        isCustomInputAllowed: false,
+        options: [],
+        validator: () => true,
```
jstoffan: Is this needed in all three components?

src/components/pill-selector-dropdown/PillSelectorDropdownField.js
```diff
@@ -0,0 +1,127 @@
...
+    };
+
+    constructor(props: Props) {
+        super(props);
+        this.state = {
+            selectorOptions: props.options,
```
jstoffan: We're going to lose changes to this prop from the parent. If this is meant to be an uncontrolled component, should this be `props.initialOptions`?

src/components/pill-selector-dropdown/PillSelectorDropdownField.js
```diff
@@ -0,0 +1,127 @@
...
+        };
+    }
+
+    deNormalize(options: Array<Option>): Array<Pill> {
```
jstoffan: Normal seems subjective here. How about `toPills` and `toOptions`?

src/components/pill-selector-dropdown/PillSelectorDropdownField.js
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    normalize(pills: Array<Pill>): Array<Option> {
+        return pills.map((pill: Pill) => ({
+            displayText: pill.text,
+            value: pill.value,
+        }));
+    }
+
+    createFakeSyntheticEvent(name: string, value: Array<Option>) {
```
jstoffan: Is this returning the full shape of a synthetic event? If so, can we add it as a return type? If not, could we just call this `this.handleChange` and leave a comment describing the need for `currentTarget` and `target?

src/components/pill-selector-dropdown/PillSelectorDropdownField.js
```diff
@@ -0,0 +1,127 @@
...
+        return {
+            currentTarget: { name, value },
+            target: { name, value },
+        };
+    }
+
+    handleInput = (text: string) => {
+        const { options } = this.props;
+        const { field } = this.props;
```
jstoffan: The two lines above can be combined.

src/components/pill-selector-dropdown/PillSelectorDropdownField.js
```diff
@@ -0,0 +1,127 @@
...
+        const { name, onChange, value } = field;
+        const options = value.slice();
+        options.splice(index, 1);
+        onChange(this.createFakeSyntheticEvent(name, options));
+    };
+
+    render() {
+        const { selectorOptions } = this.state;
+        const { className, disabled, field, form, isCustomInputAllowed, label, placeholder, validator } = this.props;
+        const { name, value = [] } = field;
```
jstoffan: The same defaulting logic is in multiple methods. Is it possible for `field.value` to be undefined or is this mainly to make flow happy?

src/components/pill-selector-dropdown/PillSelectorDropdownField.js
alexkrolick: Also, for the situation where the handler is debounced or something, will this work with the .persist() method on React synthetic events?

src/components/pill-selector-dropdown/PillSelectorDropdownField.js
priyajeet: Yeah it can be undefined, and in the example in the PR it actually is, as it has no initial value

src/components/pill-selector-dropdown/PillSelectorDropdownField.js
priyajeet: I can rename. I dont have access to the original event, for that will have to refactor the actual pill selector. Also not sure what target it should be anyways even if we did that. Since the "target" for this component is not really the input you type in, but this whole custom component itself. Formik is fine getting a fake event target such as this, and it will not call persist() if that function does not exist.

Alternative to calling onChange with fake event target, would be to use Formik's setFieldValue, but I am trying to keep all these formik wrappers free off formik specific functions.

src/components/pill-selector-dropdown/PillSelector.js
priyajeet: Yeah, Pill at the moment is { text, value } while Option is { displayText, value }.
Eventually everything should move to Option. Option will then be a Pill or a option within the drop down.

src/components/pill-selector-dropdown/PillSelectorDropdownField.js
priyajeet: `handleInput` always uses props.options to filter so should always use latest value on props.
I can add componentDidUpdate to update selectorOptions with latest props just incase `handleInput` has not been called yet.

src/components/pill-selector-dropdown/PillSelectorDropdownField.js
```diff
@@ -0,0 +1,126 @@
...
+        return { target: { name, value } };
+    }
+
+    handleInput = (text: string) => {
+        const { options, field } = this.props;
+        const { value = [] } = field;
+        const filteredOptions = options.filter((option: Option) => {
+            // Filter out anything that does not match the display text of the options
+            const hasText = !!text && option.displayText.toLowerCase().includes(text.toLowerCase());
+            // Also filter out anything that has alrady been chosen
```
priyajeet: already

src/components/pill-selector-dropdown/PillSelectorDropdownField.js
priyajeet: Removed from the one that wasn't using it locally

src/components/pill-selector-dropdown/PillSelectorDropdown.js
```diff
@@ -163,8 +170,17 @@ class PillSelectorDropdown extends React.Component<Props, State> {
         this.handleInput({ target: { value: '' } });
     };
 
+    onCompositionStart = () => {
```
jstoffan: Looks like the convention in this file is `handle` rather than `on`, so `handleCompositionStart`.

src/components/pill-selector-dropdown/PillSelectorDropdown.js
```diff
@@ -57,44 +59,46 @@ type Props = {
...
 
-        if (validator) {
+        if (!allowInvalidPills) {
             pills = pills.filter(pill => validator(pill));
         }
 
-        // Keep the data format consistent with DatalistItem
         return pills.map(pill => ({
-            text: pill,
+            displayText: pill,
```
jstoffan: Is this change necessary? Seems like it adds a lot of weight to the PR.

src/components/pill-selector-dropdown/PillSelectorDropdown.js
```diff
@@ -193,7 +210,10 @@ class PillSelectorDropdown extends React.Component<Props, State> {
                     <Label text={label}>
                         <PillSelector
                             onChange={noop} // fix console error
+                            onCompositionEnd={this.onCompositionEnd}
+                            onCompositionStart={this.onCompositionStart}
```
jstoffan: Do these need to be up here or can they live below `{...inputProps}` with the other event handler props?

src/components/pill-selector-dropdown/PillSelectorDropdownField.js
```diff
@@ -0,0 +1,117 @@
...
+
+    handleSelect = (options: Array<Option>) => {
+        const { field } = this.props;
+        const { name, onChange, value = [] } = field;
+        onChange(this.createFakeEventTarget(name, [...value, ...options]));
+    };
+
+    handleRemove = (option: Option, index: number) => {
+        const { field } = this.props;
+        const { name, onChange, value } = field;
```
jstoffan: Is `value = []` needed here or not needed in `handleSelect`?

src/components/pill-selector-dropdown/PillSelectorDropdownField.js
```diff
@@ -0,0 +1,117 @@
...
+        dropdownRenderer: defaultDropdownRenderer,
+        inputParser: parseCSV,
+        isCustomInputAllowed: true,
+        isDisabled: false,
+        options: [],
+    };
+
+    state = { inputText: '' };
+
+    createFakeEventTarget(name: string, value: Array<Option>) {
```
jstoffan: Seems like adding a comment explaining the reason for this would be worthwhile.

src/components/pill-selector-dropdown/PillSelectorDropdown.js
priyajeet: Put here so that if someone is overriding them via inputProps that continues as today

src/components/pill-selector-dropdown/PillSelectorDropdownField.js
priyajeet: handleRemove gets called when you click on the X button on a pill which is generated using value, so in that case the value should always be there

src/components/pill-selector-dropdown/PillSelectorDropdown.js
priyajeet: I want to leave it to keep the data structure consistent with other drop downs. Not sure if I will get that opportunity again anytime soon. It also helps BFFs to send displayText and we not having to convert back/forth and also re-use of flow types.

PR #1195, opened 2019-05-11
Tasks/retry 5xx

alexkrolick:
- This adds a "retryableStatusCodes" option to the Xhr class and Base API class
- Each API class can add this.retryableStatusCodes to the instance to customize which HTTP status codes should retry
- The default is [429] , same as current behavior


boxcla:

src/utils/__tests__/Xhr-test.js
```diff
@@ -301,113 +301,39 @@ describe('util/Xhr', () => {
     });
 
     describe('shouldRetryRequest', () => {
-        const createXhrInstance = shouldRetryRequest => {
+        const createXhrInstance = options => {
             xhrInstance = new Xhr({
                 token: '123',
-                shouldRetryRequest,
```
alexkrolick: This was not the right name for the option - it is `shouldRetry`. The first test case below was only passing because retry count was set to 4 initially.

src/utils/Xhr.js
```diff
@@ -118,10 +126,17 @@ class Xhr {
             return false;
         }
 
+        // Network methods that are safe to retry
+        const idempotentMethods = [HTTP_GET, HTTP_OPTIONS, HTTP_HEAD];
+
         const { response, request } = error;
         // Retry if there is a network error (e.g. ECONNRESET) or rate limited
+        const status = getProp(response, 'status');
+        const method = getProp(request, 'method');
```
alexkrolick: TODO: confirm the HTTP method is actually available on the Axios request object

alexkrolick: https://github.com/axios/axios/blob/master/lib/core/Axios.js#L38 `request.config.method`?

src/api/TaskAssignments.js
```diff
@@ -10,9 +10,28 @@ import {
...
+    HTTP_STATUS_CODE_NOT_IMPLEMENTED,
+    HTTP_STATUS_CODE_BAD_GATEWAY,
+    HTTP_STATUS_CODE_SERVICE_UNAVAILABLE,
+    HTTP_STATUS_CODE_GATEWAY_TIMEOUT,
 } from '../constants';
 
+const RETRYABLE = [
+    HTTP_STATUS_CODE_RATE_LIMIT,
+    HTTP_STATUS_CODE_INTERNAL_SERVER_ERROR,
+    HTTP_STATUS_CODE_NOT_IMPLEMENTED,
```
jeremypress: Is it possible that retrying not_implemented could succeed?

src/api/TaskLinks.js
```diff
@@ -5,9 +5,30 @@
...
+    ERROR_CODE_CREATE_TASK_LINK,
+    HTTP_STATUS_CODE_RATE_LIMIT,
+    HTTP_STATUS_CODE_INTERNAL_SERVER_ERROR,
+    HTTP_STATUS_CODE_NOT_IMPLEMENTED,
+    HTTP_STATUS_CODE_BAD_GATEWAY,
+    HTTP_STATUS_CODE_SERVICE_UNAVAILABLE,
+    HTTP_STATUS_CODE_GATEWAY_TIMEOUT,
+} from '../constants';
+
+const RETRYABLE = [
```
jeremypress: Is there any shared constants area for tasks? Could be useful since these classes will probably keep retryable fields in sync.

src/utils/__tests__/Xhr-test.js
```diff
@@ -301,113 +301,40 @@ describe('util/Xhr', () => {
...
-        ])('should retry request %#', (shouldRetryRequest, response, request, retryCount, expected) => {
-            createXhrInstance(shouldRetryRequest);
-            xhrInstance.retryCount = retryCount;
-            const result = xhrInstance.shouldRetryRequest({
-                response,
-                request,
-            });
-            expect(result).toBe(expected);
-        });
+        test.each`
```
jeremypress: Thanks for updating this, much easier on the eyes.

src/api/TaskAssignments.js
```diff
@@ -10,9 +10,28 @@ import {
...
+    HTTP_STATUS_CODE_BAD_GATEWAY,
+    HTTP_STATUS_CODE_SERVICE_UNAVAILABLE,
+    HTTP_STATUS_CODE_GATEWAY_TIMEOUT,
+];
+
 class TaskAssignments extends Base {
```
jstoffan: Can we create a `TasksBase` that handles this logic for all the affected classes?

src/api/TaskAssignments.js
alexkrolick: Looking at other retry plugins, they retry all 5xx codes:

https://github.com/JustinBeckwith/retry-axios/blob/master/src/index.ts#L110-L120

https://github.com/softonic/axios-retry/blob/master/es/index.js#L28

jstoffan: It looks like they also retry put/delete. Do the task service endpoints throw 501 errors that we want to retry?

alexkrolick: One of them retries PUT and DELETE, the other doesn't. Technically those are supposed to be idempotent but 1) we aren't seeing many errors on those methods. 99%+ are GET. 2) Due to the side-effects on task status when task_collaborators are updated, I am not sure we want to retry PUT and DELETE for this particular case.

src/api/TaskAssignments.js
alexkrolick: Ok, that is a good way to handle the constants.

src/api/TaskAssignments.js
alexkrolick: I'll check with backend about what they recommend

alexkrolick: We are going to go with what's here. Not PUT and DELETE.

src/utils/Xhr.js
```diff
@@ -77,6 +84,7 @@ class Xhr {
         sharedLinkPassword,
         responseInterceptor,
         requestInterceptor,
+        retryableStatusCodes = [HTTP_STATUS_CODE_RATE_LIMIT],
```
jstoffan: If `options` is defined, but this property isn't, we lose the default value here. Should we put it inline on line 100?

src/utils/Xhr.js
```diff
@@ -118,10 +126,17 @@ class Xhr {
...
+        const idempotentMethods = [HTTP_GET, HTTP_OPTIONS, HTTP_HEAD].map(s => s.toLowerCase());
+
+        const { response, request, config } = error;
         // Retry if there is a network error (e.g. ECONNRESET) or rate limited
+        const status = getProp(response, 'status');
+        const method = getProp(config, 'method');
         const isNetworkError = request && !response;
-        return isNetworkError || getProp(response, 'status') === HTTP_STATUS_CODE_RATE_LIMIT;
+        const isRateLimitError = status === HTTP_STATUS_CODE_RATE_LIMIT;
+        const isOtherRetryableError = this.retryableStatusCodes.includes(status) && idempotentMethods.includes(method);
```
jstoffan: It looks like `.includes` isn't supported natively in IE11. May as well swap it out for a lodash version.

src/utils/Xhr.js
alexkrolick: Are you sure?

```
> function foo({a = 'A!', b} = {}) { return {a, b} }
undefined
> foo()
{ a: 'A!', b: undefined }
> foo({b: 'z'})
{ a: 'A!', b: 'z' }
```

jstoffan: Hm, fair enough. Brain fart, I guess.

alexkrolick:
Rebasing...

PR #1196, opened 2019-05-13
fix(unified-share-modal): add constraints to modal menus

booc0mtaco:
add some constraints so that modal menu contents stay on-screen when windows are short


boxcla:

i18n/en-US.properties
```diff
@@ -1340,5 +1340,3 @@ boxui.versionHistoryModal.versionLimitExceeded = You can restore the most recent
 boxui.versionHistoryModal.versionNumberBadge = V{versionNumber}
 # Label given to the version badge for screen readers
 boxui.versionHistoryModal.versionNumberLabel = Version number {versionNumber}
-# y
```
booc0mtaco: I'm not sure what this change is about. I think `master` may have had some message with this pattern, which you can see is not part of my commit.

jstoffan: It slipped in with the FormattedCompMessage PR. Removing it should be fine. FYI @ehoogerbeets

src/features/unified-share-modal/InviteePermissionsMenu.js
```diff
@@ -114,6 +114,8 @@ class InviteePermissionsMenu extends Component<Props> {
             plainButton
         );
 
+        // TODO: `DropdownMenu` doesn't currently handle a scenario where the menu is taller than
```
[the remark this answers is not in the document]
booc0mtaco: Right now, `DropdownMenu` has two behaviors: to open the content below the menu item, or above it if one of the constraints is violated. in the case where neither would cause the menu to fully display, both behaviors cause issues.

Perhaps the better place for this would be an issue here describing the change to make in `DropdownMenu`, but not sure what exactly would be the action.

# should it use `ScrollWrapper` and a fixed height?
# should it appear adjacent to the menu target?
# should it have native behavior and expand upward as a user traverses down into the available option?

I didn't want to prescribe the change to make, since I'm not sure which approach is best. Suggestions?

PR #1197, opened 2019-05-13
Fix styleguidest warnings for new i18n components

ehoogerbeets:
Param and Plural should not be in the styleguide.conf.  These are covered by the documentation for FormattedCompMessage already.


boxcla:

PR #1198, opened 2019-05-13
fix(versions): Update versions restore call to match latest contract

jstoffan:


boxcla:

PR #1200, opened 2019-05-14
feat(scroll): upgrade scroll in to view if needed

Sunyang730:
BREAKING CHANGE: Major version/peer dependency upgrade for [scroll-into-view-if-needed](https://github.com/stipsan/scroll-into-view-if-needed)


boxcla:

examples/src/SelectorDropdownExamples.js
```diff
@@ -132,6 +132,36 @@ const SelectorDropdownExamples = () => (
             'Straight Outta Compton',
             'Reasonable Doubt',
             'Super long name that should be truncated, we should see the dots at the very end, adding some more text here, should truncate at any second now, please truncate soon',
+            'Illmatic',
```
Sunyang730: Just increasing the size of the dropdown, can remove or edit it to make more sense if needed

src/features/metadata-instance-editor/Instance.js
```diff
@@ -134,10 +134,10 @@ class Instance extends React.PureComponent<Props, State> {
     componentDidUpdate() {
         const element = this.collapsibleRef.current;
         if (element && this.state.shouldConfirmRemove) {
-            scrollIntoViewIfNeeded(element, false, {
-                centerIfNeeded: false,
-                duration: 320,
-                easing: 'easeInOut',
+            scrollIntoViewIfNeeded(element, {
```
Sunyang730: The transition behave the same to my eyes. Not sure if we must of the duration? If so, we will need to do introduce a callback that handles that which will complicate things.

src/utils/dom.js
```diff
@@ -112,6 +112,6 @@ export function scrollIntoView(itemEl: ?HTMLElement): void {
     // @NOTE: breaks encapsulation but alternative is unknown child ref
     if (itemEl) {
         const parentEl = itemEl.closest(`.body, .modal, .${OVERLAY_WRAPPER_CLASS}`);
-        scrollIntoViewIfNeeded(itemEl, false, undefined, parentEl);
+        scrollIntoViewIfNeeded(itemEl, { boundary: parentEl });
```
Sunyang730: The place that's using this 'selectorDropdown' does not seems to be doing anything even with the old code. Anyone know how the selectorDropdown suppose to behave originally?

package.json
```diff
@@ -289,7 +289,7 @@
         "react-textarea-autosize": "^7.1.0",
         "react-virtualized": "^9.21.0",
         "regenerator-runtime": "^0.13.2",
-        "scroll-into-view-if-needed": "^1.5.0",
+        "scroll-into-view-if-needed": "2.2.20",
```
priyajeet: Add ^ to both, since we dont want to limit clients from using a newer version

src/features/metadata-instance-editor/Instance.js
```diff
@@ -4,7 +4,7 @@ import { FormattedMessage } from 'react-intl';
 import classNames from 'classnames';
 import isEqual from 'lodash/isEqual';
 import cloneDeep from 'lodash/cloneDeep';
-import scrollIntoViewIfNeeded from 'scroll-into-view-if-needed';
+import { scrollIntoView } from '../../utils/dom';
```
priyajeet: Group the import with the relative ones below, since the top section is for npm imports

PR #1204, opened 2019-05-17
feat(radio): formik wrapper for RadioButton

priyajeet:
1. Adds `<RadioButtonField>` to be used **_on its own group and outside_** `<RadioGroupField>` when the UI requires components that the RadioGroup component cannot accommodate within its DOM structure. This allows building of UI as seen in the picture below that requires text boxes be embedded within a group radio buttons. RadioGroup does not support this type of embedding.

Screen Shot 2019-05-17 at 3 25 46 PM

2. Fixes minor issue where formik `form` object was getting put inside the DOM


boxcla:

src/components/form-elements/formik/README.md
```diff
@@ -60,6 +61,24 @@ const SelectField = require('box-ui-elements/es/components/select-field').Select
                     placeholder="Text Area Field"
                     component={TextArea}
                 />
+                <b>Non-RadioGroup RadioButtons sharing the same name</b>
+                <br />
+                <br />
+                <Field
+                    isSelected={false}
+                    label="Radio Button Field 1"
+                    name="radiobutton"
```
jstoffan: This example is a bit confusing. Radio buttons with the same name are meant to be part of a group.

src/components/radio/RadioButtonField.js
```diff
@@ -0,0 +1,16 @@
...
+import type { FieldProps } from 'formik';
+
+import RadioButton from './RadioButton';
+import type { RadioButtonProps } from './RadioButton';
+
+type Props = RadioButtonProps & FieldProps;
+
+const RadioButtonField = ({ field, form, value, ...rest }: Props) => {
+    const { value: fieldValue, ...fieldRest } = field;
+    return <RadioButton {...fieldRest} {...rest} value={value} isSelected={value === fieldValue} />;
```
jstoffan: It seems like this could be accomplished by using `RadioButtonGroup` with a unique name?

src/components/checkbox/CheckboxField.js
```diff
@@ -8,7 +8,7 @@ import type { CheckboxProps } from './Checkbox';
 
 type Props = CheckboxProps & FieldProps;
 
-const CheckboxField = ({ field, ...rest }: Props) => {
+const CheckboxField = ({ field, form, ...rest }: Props) => {
```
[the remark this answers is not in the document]
priyajeet: No real issue, just that the dom had unnecessary `<input form=[Object object]]>`

src/components/form-elements/formik/README.md
priyajeet: Yeah radio will still be used as a group of radios (hence same name) just not inside our `<RadioGroup>` component since that does not allow arbitrary components.

PR #1209, opened 2019-05-20
 chore(version-history): Include current version ID in additionalVersionInfo

jeremypress:


boxcla:

src/elements/content-preview/ContentPreview.js
```diff
@@ -1054,9 +1054,10 @@ class ContentPreview extends PureComponent<Props, State> {
      */
     onVersionChange = (version?: BoxItemVersion, additionalVersionInfo?: AdditionalVersionInfo = {}): void => {
         const { onVersionChange }: Props = this.props;
+        const { file }: State = this.state;
         this.updateVersionToCurrent = additionalVersionInfo.updateVersionToCurrent;
 
-        onVersionChange(version, additionalVersionInfo);
+        onVersionChange(version, { ...additionalVersionInfo, currentVersionId: getProp(file, 'file_version.id') });
```
jeremypress: We need to pass the file's current version ID, since EUA overwrites this on each preview.

jstoffan: Does this need to come from ContentPreview or can it be passed up by the sidebar?

jeremypress: It could come from the sidebar as the 0th index of the versions array, but it's stored more explicitly here.

PR #1211, opened 2019-05-21
feat(access-label): Add text override prop

cweeii:
This change is to allow for passing in custom override message


boxcla:

booc0mtaco:
Hm, this component seems pretty simple. why not do a switch where it is used to either use `AccessLabel` or a component based on whatever condition sets the override?

booc0mtaco:
extra note: have a look at `SharedLinkAccessDescription` which appears to need some similar treatment

cweeii:
Thanks for the feedback.  Let me look at an alternative way of doing it.

PR #1217, opened 2019-05-23
feat(icons): update to new globe icon and add new globe tinycon

cweeii:
Old:
image

New:
image

New Tinycon:
image


boxcla:

src/icons/general/IconGlobe.js
```diff
@@ -12,18 +13,17 @@ type Props = {
...
+const IconGlobe = ({ className = '', color = bdlNeutral03, height = 16, title, width = 16 }: Props) => (
     <AccessibleSVG
         className={`icon-globe ${className}`}
         height={height}
         title={title}
         viewBox="0 0 16 16"
         width={width}
     >
-        <circle className="fill-color" cx="8" cy="8" fill={color} r="8" />
         <path
```
j3tan: `fill-color` class

src/icons/general/IconGlobe.js
```diff
@@ -12,18 +13,17 @@ type Props = {
     width?: number,
 };
 
-const IconGlobe = ({ className = '', color = '#000', height = 16, title, width = 16 }: Props) => (
+const IconGlobe = ({ className = '', color = bdlNeutral03, height = 16, title, width = 16 }: Props) => (
```
priyajeet: 32x32 please

priyajeet:
@cweeii can this be made into the two-toned icons? See two toned directory

src/icons/general/IconGlobe.js
cweeii: done

src/icons/general/IconGlobe.js
cweeii: @priyajeet I've made it 20x20 as it's mainly used as a badgeable icon

cweeii:
@priyajeet I don't see a need for background/foreground color for the icons I'm editing/adding, so do they need to be two-toned?

src/icons/general/__tests__/IconGlobeTinycon-test.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+
+import { white } from '../../../styles/variables';
+
+import IconGlobeTinycon from '../IconGlobeTinycon';
```
booc0mtaco: We don't have any naming scheme for the different usages of icons. @priyajeet is there any scheme that might make sense from your perspective? I was thinking a usage prefix like `Tinycon` which also implies associated default sizing, etc.

priyajeet:
@booc0mtaco @tjgupta I think we should make all svg icons same viewport 32x32 irrespective of their use. If it needs to be a 20x20 icon for badges, then the width/height should be modified. Something like this https://codesandbox.io/s/peaceful-poitras-rc47d

PR #1218, opened 2019-05-24
feat(select): make caret rotate up/down on open/close respectively

priyajeet:
Also makes collapsible and skill card caret rotate up/down on open/close respectively. This makes all 3 consistent to drop down menu carets.

caret
collap

Screen Shot 2019-05-24 at 11 09 50 AM
Screen Shot 2019-05-23 at 10 54 53 PM
Screen Shot 2019-05-23 at 10 55 05 PM
Screen Shot 2019-05-23 at 10 55 33 PM
Screen Shot 2019-05-23 at 10 55 40 PM
Screen Shot 2019-05-23 at 10 56 12 PM


boxcla:

src/elements/content-sidebar/SidebarSection.scss
```diff
@@ -18,6 +18,10 @@
     text-align: left;
     width: 100%;
 
+    .icon-caret-down {
+        transition: transform .25s;
```
macrotim: Should we move to var's?

macrotim: ...because I see some copy / paste.

priyajeet:
> started thinking about how we might introduce a `_animations.scss` to define lengths, directions, and keyframe animations that ought to be consistent

+2

src/elements/content-sidebar/SidebarSection.scss
```diff
@@ -27,12 +31,8 @@
...
     }
-
-    .icon-caret-down {
-        transform: rotateZ(-90deg);
-    }
 }
 
 .be .bcs-section-open .btn-plain.bcs-section-title .icon-caret-down {
-    transform: none;
+    transform: rotateZ(180deg);
```
nickgriffin1: Could we pull this into a shared variable. Something like $bdl-caret-transform?

priyajeet: I am not sure if we should variable-ize everything. No gaurantee that we will keep that degree forever.

src/components/collapsible/Collapsible.scss
```diff
@@ -29,6 +29,7 @@
     .collapsible-card-header-caret {
         position: absolute;
         right: 0;
+        transition: transform $bdl-transition-duration-200ms;
```
booc0mtaco: I'd lean toward `-short` instead of a precise value in the variable name, or something similar. We can define a few specific speeds, and avoid future var.s like `-300ms`, `-450ms`, etc.

I'd also consider defining the easing function too. One here is using `ease-in` and the ones without will default to `linear`. (IMO `ease-in-out` offers the smoothest timing behavior). 

This makes me think a mixin for the transition might be more useful, so we can capture all three parts. Thoughts?

priyajeet: I'll do it in a followup

PR #1219, opened 2019-05-24
style(bdl-colors): add colors to variables.scss

Visylvius:
Adds new colors to BDL


boxcla:

src/styles/_variables.scss
```diff
@@ -4,10 +4,17 @@ $bdl-secondary-blue: #003c84 !default;
 $bdl-light-blue: #2486fc !default;
 $bdl-green-light: #26c281 !default;
 $bdl-yellorange: #f5b31b !default;
+$bdl-yellow: #ffd700 !default;
 $bdl-grimace: #4826c2 !default;
 $bdl-purple-rain: #9f3fed !default;
 $bdl-watermelon-red: #ed3757 !default;
 
+// Box Design Theme Color Variations
+// Confirm with design before extending
```
macrotim: What's the meaning of "extend"?

priyajeet:
They do get auto generated

src/styles/_variables.scss
booc0mtaco: one might assume "oh if we have something with 10, maybe I can add 20" etc. wanted to put a useful message here so that we get confirmation on BDL for the color values and increments to use.

PR #1223, opened 2019-05-30
perf(sidebar): Lazy load sidebar panels without 500ms delay

ConradJChan:
Invokes the loader function first without waiting for any delay and then rely on `retryNumOfTimes` in case of error


boxcla:

src/elements/common/async-load/AsyncLoad.js
```diff
@@ -26,7 +26,7 @@ const AsyncLoad = (asyncProps: Props = {}) => {
             DEFAULT_INITIAL_DELAY,
             DEFAULT_BACKOFF_FACTOR,
         );
-    const LazyComponent = React.lazy(asyncLoadWithRetry);
+    const LazyComponent = React.lazy(() => asyncProps.loader().catch(asyncLoadWithRetry));
```
jeremypress: Will this increase our number of retries by 1?

ConradJChan: Yeah, I guess technically we try once and then retry 3 times now whereas before we do a total of 3 attempts.

jstoffan: It looks this solution duplicates a bit of code and changes the behavior slightly. Would changing `DEFAULT_INITIAL_DELAY` to 0 also work?

ConradJChan: No, not as `retryNumOfTimes` works as is, the value passed in is used in conjunction with the backoff factor to determine the next attempt, so if it is 0, it will always fire off the retry attempt immediately

PR #1224, opened 2019-05-30
feat(icons): add icons for collections

nickgriffin1:
image


boxcla:

src/icons/collections/__tests__/IconCollectionsAdd-test.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+
+import IconCollectionsAdd from '../IconCollectionsAdd';
+
+describe('icons/collections/IconCollectionsAdd', () => {
```
nickgriffin1: I'm not sure these unit tests are totally necessary. I can be convinced to remove them.

booc0mtaco:
@nickgriffin1 can we get imports where the view boxes are "0 0 32 32"

src/icons/collections/__tests__/IconCollectionsAdd-test.js
FuzzyTree: It looks like other Icons in this project use snapshots for testing icons. I think either way is ok.

nickgriffin1: I'll add snapshots. That's a good idea.

nickgriffin1:
@booc0mtaco I plan on getting and updated the icon with a corrected one at a later date.

PR #1227, opened 2019-05-30
refactor(tasks-new): extract task assignee avatars

tjgupta:
- Extracted out avatar group CSS into separate file so the component can be used independently of a Task.
- Renamed class names and components related to avatar group to be more generic so we can pull it out into a reusable `<AvatarGroup>` separate from Task at some future time.


boxcla:

src/elements/content-sidebar/activity-feed/task-new/_vars.scss
```diff
@@ -0,0 +1,8 @@
+@mixin avatar-badge-icon {
```
booc0mtaco: consider renaming to `_variables` or `_mixins` (since there are no variables defined here). Also, to be safe we should import the main _variables.scss file (which has the colors defined). SASS will de-dupe any redundant imports, and that makes this file safe to import separately, should that case arise.

alexkrolick: Alternately, make the badge a component?

tjgupta: Renamed to _mixins and added import of _variables. I could probably turn it into a reusable component (it's already a component, but it's not used in the top level avatar currently). However, I'd suggest doing that in a separate PR unless you object @alexkrolick.

src/elements/content-sidebar/activity-feed/task-new/AssigneeAvatar.js
```diff
@@ -29,17 +29,21 @@ const StatusIcon = ({ status, ...rest }: { status: TaskCollabStatus }) => {
...
-const AssignmentStatus = React.memo<Props>(({ user, status, getAvatarUrl, className, ...rest }: Props) => (
-    <div className={classNames('bcs-task-assignment-status', className)} data-testid="task-assignment-status" {...rest}>
-        <Avatar className="bcs-task-assignment-avatar" user={user} getAvatarUrl={getAvatarUrl} />
+const AssigneeAvatar = React.memo<Props>(({ user, status, getAvatarUrl, className, ...rest }: Props) => (
+    <div
+        className={classNames('bcs-avatar-group-avatar-container', className)}
+        data-testid="task-assignment-status"
+        {...rest}
+    >
+        <Avatar className="bcs-avatar-group-avatar" user={user} getAvatarUrl={getAvatarUrl} />
```
jstoffan: It seems like including the concept of `group` in the base AssigneeAvatar component is a bit of a separation of concerns issue. Couldn't it be used outside a group?

src/elements/content-sidebar/activity-feed/task-new/AvatarGroup.js
```diff
@@ -0,0 +1,127 @@
...
+
+const Datestamp = ({ date }: { date: ISODate | Date }) => {
+    return <ReadableTime timestamp={new Date(date).getTime()} alwaysShowTime relativeThreshold={0} />;
+};
+
+const AvatarDetails = React.memo(({ user, status, completedAt, className }) => {
+    const statusMessage = statusMessages[status] || null;
+    return (
+        <div className={className}>
+            <div className="bcs-avatar-group-details-name">{user.name}</div>
```
jstoffan: Can we start using SUIT for these components? The naming here is getting pretty unwieldy.

src/elements/content-sidebar/activity-feed/task-new/AvatarGroup.js
```diff
@@ -0,0 +1,127 @@
...
+        </div>
+    );
+});
+
+class AvatarGroup extends React.Component<Props> {
+    static defaultProps = {
+        maxAvatars: MAX_AVATARS,
+    };
+
+    listTitleId = `avatar-group-list-title-${uniqueId()}`;
```
jstoffan: This can be used as `uniqueId('avatar-group-list-title-') to avoid string building. Also, does this ID need to persist between renders or can we generate a new one when the props change? If the latter, this could be a stateless functional component, instead.

src/elements/content-sidebar/activity-feed/task-new/AvatarGroup.js
```diff
@@ -0,0 +1,127 @@
...
+
+    render() {
+        const { maxAvatars, assignees = {}, getAvatarUrl, intl } = this.props;
```
jstoffan: Seems like `assignees` should have a default prop value, instead.

src/elements/content-sidebar/activity-feed/task-new/AvatarGroup.scss
```diff
@@ -0,0 +1,107 @@
...
+        cursor: pointer;
+        display: flex;
+        font-weight: bold;
+        justify-content: center;
+    }
+
+    .bcs-avatar-group-avatar-container {
+        margin-left: $avatar-spacing;
+
+        &:first-of-type {
```
jstoffan: This pseudo-class works on elements tags rather than classes, so this is a bit misleading We can use sibling selectors instead to avoid the override entirely.

```
.bcs-avatar-group-avatar-container {
    & + & {
        margin-right: $avatar-spacing;
    }
}
```

src/elements/content-sidebar/activity-feed/task-new/AvatarGroup.scss
```diff
@@ -0,0 +1,107 @@
+@import '../../../common/variables';
+@import './mixins';
+
+$avatar-spacing: 4px;
```
jstoffan: This is being declared as a global variable. Is that intended?

src/elements/content-sidebar/activity-feed/task-new/AvatarGroup.js
tjgupta: Yep, updated.

src/elements/content-sidebar/activity-feed/task-new/AvatarGroup.js
tjgupta: I updated this to be `uniqueId('avatar-group-list-title-')`, but didn't yet switch to functional component for the sake of not making the diff larger.

src/elements/content-sidebar/activity-feed/task-new/AvatarGroup.js
tjgupta: Done, and renamed to be more generic (`users`) in the event we extract this out as a separate reusable component.

src/elements/content-sidebar/activity-feed/task-new/AvatarGroup.scss
tjgupta: Done

src/elements/content-sidebar/activity-feed/task-new/AvatarGroup.scss
tjgupta: Fixed.

src/elements/content-sidebar/activity-feed/task-new/AssigneeAvatar.js
tjgupta: I think maybe the confusion was with the naming? Now it's called `bcs-AvatarGroup-avatar`. 

It could perhaps be exposed independently eventually if needed, maybe something like `AvatarWithStatus`? But I think that's a few steps out.

src/elements/content-sidebar/activity-feed/task-new/AvatarGroup.scss
```diff
@@ -0,0 +1,99 @@
+@import '../../../common/variables';
+@import './mixins';
+
+.bcs-AvatarGroup {
+    display: flex;
+    margin: 10px 0;
+
+    .bcs-AvatarGroup-avatar {
```
jstoffan: Nesting should be avoided when possible to reduce specificity. SUIT naming conventions are intended to help facilitate that by scoping styles to specific component names/namespaces.

src/elements/content-sidebar/activity-feed/task-new/AvatarGroup.scss
```diff
@@ -0,0 +1,99 @@
...
+            margin-left: $avatar-spacing;
+        }
+    }
+}
+
+.bcs-AvatarGroup-listFlyout {
+    max-width: 100%;
+    width: 280px;
+
+    .bcs-AvatarGroup-list {
```
jstoffan: The renaming done so far is a great start. I think we should target these components for further improvement in the future, as well. It looks like this should probably be its own component (`bcs-AvatarList` or something similar), for example, which could then have `bcs-AvatarList-title`, `bcs-AvatarList-item`, etc.

tjgupta: Yep agreed. Since this isn't yet a fully generic reusable component separate from tasks (yet), I didn't go all the way with that, but if/when we extract it fully from the tasks stuff we can fix that.

PR #1231, opened 2019-05-31
feat(classification): new badge and layout

priyajeet:
New design for classification. Does not delete older files just yet.

Screen Shot 2019-05-30 at 11 08 59 PM

### DIFF
--- a/src/features/classification/Classification.js
+++ b/src/features/classification/Classification.js
@@ -1,0 +1,27 @@
+// @flow
+import * as React from 'react';
+
+import Tooltip from '../../components/tooltip';
+import './Classification.scss';
+
+type Props = {
+    advisoryMessage?: string,
+    isMessageInline?: boolean,
+    name: string,
+};
+
+const Classification = ({ advisoryMessage, isMessageInline = true, name }: Props) => {
+    const hasMessage = !!advisoryMessage;
+    const isTooltipEnabled = hasMessage && !isMessageInline;
+    const isInlineMessageVisible = hasMessage && isMessageInline;
+    return (
+        <article className="bdl-Classification">
+            <Tooltip isDisabled={!isTooltipEnabled} position="bottom-center" text={advisoryMessage}>
+                <h1 className="bdl-Classification-badge">{name}</h1>
+            </Tooltip>
+            {isInlineMessageVisible && <p className="bdl-Classification-advisoryMessage">{advisoryMessage}</p>}
+        </article>
+    );
+};
+
+export default Classification;
--- a/src/features/classification/Classification.scss
+++ b/src/features/classification/Classification.scss
@@ -1,0 +1,18 @@
+@import '../../styles/variables';
+
+.bdl-Classification-badge {
+    background-color: $bdl-yellorange-10;
+    border: 1px solid $bdl-yellorange;
+    border-radius: $bdl-border-radius-size;
+    display: inline;
+    font-size: 11px;
+    font-weight: 600;
+    line-height: initial;
+    margin: 0;
+    padding: 2px 5px;
+    text-transform: uppercase;
+}
+
+.bdl-Classification-advisoryMessage {
+    margin: 5px 0 0;
+}
--- a/conf/styleguide.config.js
+++ b/conf/styleguide.config.js
@@ -201,9 +201,7 @@ const allSections = [
     {
         name: 'Features',
         components: () => [
-            '../src/features/activity-feed/activity-feed/ActivityFeed.js',
-            '../src/features/in-app-message/create-message-form/CreateMessageForm.js',
-            '../src/features/in-app-message/targeting-expression-editor/TargetingExpressionEditor.js',
+            '../src/features/classification/Classification.js',
             '../src/features/invite-collaborators-modal/InviteCollaboratorsModal.js',
             '../src/features/left-sidebar/LeftSidebar.js',
             '../src/features/header-flyout/HeaderFlyout.js',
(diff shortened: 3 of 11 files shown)


boxcla:

src/features/classification/Classification.js
```diff
@@ -0,0 +1,36 @@
...
+    const borderColor = color;
+    const backgroundColor = Color(color).lighten(0.75);
+    const hasMessage = !!advisoryMessage;
+    const isTooltipEnabled = hasMessage && !isMessageInline;
+    const isInlineMessageVisible = hasMessage && isMessageInline;
+
+    return (
+        <div className="bdl-Classification">
+            <Tooltip isDisabled={!isTooltipEnabled} position="bottom-center" text={advisoryMessage}>
+                <span className="bdl-Classification-badge" style={{ backgroundColor, borderColor }}>
```
priyajeet: Contrasting font color will be later once the full pallette is known

src/features/classification/Classification.js
```diff
@@ -0,0 +1,36 @@
...
+type Props = {
+    advisoryMessage?: string,
+    color?: string,
+    isMessageInline?: boolean,
+    name: string,
+};
+
+const Classification = ({ advisoryMessage, color = bdlYellorange, isMessageInline = true, name }: Props) => {
+    const borderColor = color;
+    const backgroundColor = Color(color).lighten(0.75);
```
jstoffan: Should `Color` be lowercase, since it's not being used as a constructor?

package.json
```diff
@@ -264,6 +264,7 @@
         "@hapi/address": "^2.0.0",
         "axios": "^0.19.0",
         "classnames": "^2.2.5",
+        "color": "^3.0.0",
```
jstoffan: This is another 7.5KB of gzipped code for a single usage. Can we find a different way, please?

src/features/classification/Classification.js
```diff
@@ -0,0 +1,36 @@
...
+
+type Props = {
+    advisoryMessage?: string,
+    color?: string,
+    isMessageInline?: boolean,
+    name: string,
+};
+
+const Classification = ({ advisoryMessage, color = bdlYellorange, isMessageInline = true, name }: Props) => {
+    const borderColor = color;
```
jstoffan: Rename during destructuring.

src/features/classification/Classification.js
```diff
@@ -0,0 +1,36 @@
...
+    const backgroundColor = Color(color).lighten(0.75);
+    const hasMessage = !!advisoryMessage;
+    const isTooltipEnabled = hasMessage && !isMessageInline;
+    const isInlineMessageVisible = hasMessage && isMessageInline;
+
+    return (
+        <article className="bdl-Classification">
+            <Tooltip isDisabled={!isTooltipEnabled} position="bottom-center" text={advisoryMessage}>
+                <h3 className="bdl-Classification-badge" style={{ backgroundColor, borderColor }}>
```
jstoffan: This should be an `<h1>` if it's in a new outline context (article).

src/features/classification/Classification.scss
```diff
@@ -0,0 +1,18 @@
+@import '../../styles/variables';
+
+.bdl-Classification {
+    .bdl-Classification-badge {
```
jstoffan: Nesting is not needed using these SUIT naming conventions.

src/features/classification/Classification.scss
```diff
@@ -0,0 +1,18 @@
...
+        border: 1px solid;
```
jstoffan: I think it would help to specify a default color and include a comment that it's overridden in JS.

src/features/classification/Classification.js
priyajeet: Using as per their documentation

src/features/classification/Classification.js
priyajeet: Not necessarily, per MDN
"Each <article> should be identified, typically by including a heading (<h1>-<h6> element) as a child of the <article> element."

jstoffan: Articles introduce a new outline context that is separate from the root document's. This is the most important piece of information in this article context, so should use an `h1`.

package.json
priyajeet: its already included in our apps, its used for theming

jstoffan: Sure, but it's not currently used in Elements and will increase bundle size for platform customers. It's also used in a single place. Is there no other alternative?

src/features/classification/Classification.js
priyajeet: Its actually a class, `new` keyword is optional for instantiating anyways.

package.json
priyajeet: Have to keep library consistent otherwise there will be 2 libraries.
I haven't seen other ones, maybe tinycolor

src/features/classification/Classification.js
```diff
@@ -0,0 +1,40 @@
...
+const Classification = ({
+    advisoryMessage,
+    color: borderColor = bdlYellorange,
+    isMessageInline = true,
+    name,
+}: Props) => {
+    const backgroundColor = Color(borderColor).lighten(0.75);
+    const hasMessage = !!advisoryMessage;
+    const isTooltipEnabled = hasMessage && !isMessageInline;
+    const isInlineMessageVisible = hasMessage && isMessageInline;
```
Sunyang730: Are you seeing use case where isMessage

package.json
priyajeet: https://www.npmtrends.com/chroma-js-vs-color-vs-color-js-vs-tinycolor-vs-tinycolor2

src/features/classification/Classification.js
priyajeet: Thats the default case now

PR #1233, opened 2019-05-31
feat(navlist): add animation prop and enable for LeftSidebar

nickgriffin1:
NavList is given a new prop `enableAnimation` that will animate the height of it's contents. I think it is better to use a prop instead of just giving it a height animation so that we aren't injecting a useless component when it's not necessary. @priyajeet or @jstoffan let me know if you have any feedback for that approach.

This also includes a bug fix for the focus outline for the toggle button to prevent overflow and overlap with other components.

### DIFF
--- a/src/components/nav-sidebar/NavList.js
+++ b/src/components/nav-sidebar/NavList.js
@@ -2,25 +2,47 @@
 import * as React from 'react';
 import classNames from 'classnames';
 
+import AnimateHeight from 'react-animate-height';
+
 type Props = {
     children: React.Node,
     className?: string,
     collapsed?: boolean,
+    enableAnimation?: boolean,
     heading?: React.Node,
     placeholder?: React.Node,
     ulProps?: Object,
 };
 
-const NavList = ({ children, className = '', collapsed = false, heading, placeholder, ulProps = {} }: Props) => {
+const NavList = ({
+    children,
+    className = '',
+    collapsed = false,
+    enableAnimation = false,
+    heading,
+    placeholder,
+    ulProps = {},
+}: Props) => {
     const classes = classNames(`nav-list`, className, {
+        'is-animation-enabled': enableAnimation,
         'is-collapsed': collapsed,
     });
+
+    const childrenParsed = React.Children.map(children, link => (link ? <li>{link}</li> : null));
 
     return (
         <nav className={classes}>
             {heading ? <h2>{heading}</h2> : null}
             {placeholder}
-            <ul {...ulProps}>{React.Children.map(children, link => (link ? <li>{link}</li> : null))}</ul>
+            <ul {...ulProps}>
+                {enableAnimation && children ? (
+                    <AnimateHeight duration={300} height={!collapsed ? 'auto' : 0}>
+                        {childrenParsed}
+                    </AnimateHeight>
+                ) : (
+                    childrenParsed
+                )}
+            </ul>
         </nav>
     );
 };
--- a/src/components/nav-sidebar/NavList.js
+++ b/src/components/nav-sidebar/NavList.js
@@ -10,6 +10,8 @@ type Props = {
     collapsed?: boolean,
     enableAnimation?: boolean,
     heading?: React.Node,
+    onAnimationEnd?: () => void,
+    onAnimationStart?: () => void,
     placeholder?: React.Node,
     ulProps?: Object,
 };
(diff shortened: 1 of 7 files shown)


boxcla:

src/components/nav-sidebar/NavList.js
```diff
@@ -2,25 +2,47 @@
...
+    children,
+    className = '',
+    collapsed = false,
+    enableAnimation = false,
+    heading,
+    placeholder,
+    ulProps = {},
+}: Props) => {
     const classes = classNames(`nav-list`, className, {
+        'is-animation-enabled': enableAnimation,
```
priyajeet: Use the new style, is-animationEnabled

src/components/nav-sidebar/NavList.js
```diff
@@ -2,25 +2,47 @@
...
     placeholder?: React.Node,
     ulProps?: Object,
 };
 
-const NavList = ({ children, className = '', collapsed = false, heading, placeholder, ulProps = {} }: Props) => {
+const NavList = ({
+    children,
+    className = '',
+    collapsed = false,
+    enableAnimation = false,
```
priyajeet: isAnimationEnabled or hasAnimation

booc0mtaco:
The shadow seems to have the opposite effect that it should. When collapsed, there should be no shadow. When expanded below the bottom edge, the shadow should appear

nickgriffin1:
@booc0mtaco Good catch. I'll take a look at that.

src/components/nav-sidebar/NavList.js
```diff
@@ -2,25 +2,47 @@
 import * as React from 'react';
```
priyajeet: I would also change caret to match https://github.com/box/box-ui-elements/pull/1218

src/components/nav-sidebar/NavSidebar.scss
```diff
@@ -58,7 +59,7 @@
     }
 
     &.is-collapsed {
-        > ul,
+        :not(.is-animation-enabled) > ul,
```
priyajeet: anyway to simplify by not using not?

src/components/nav-sidebar/NavList.js
```diff
@@ -2,25 +2,47 @@
...
+    const childrenParsed = React.Children.map(children, link => (link ? <li>{link}</li> : null));
+
     return (
         <nav className={classes}>
             {heading ? <h2>{heading}</h2> : null}
             {placeholder}
-            <ul {...ulProps}>{React.Children.map(children, link => (link ? <li>{link}</li> : null))}</ul>
+            <ul {...ulProps}>
+                {enableAnimation && children ? (
+                    <AnimateHeight duration={300} height={!collapsed ? 'auto' : 0}>
```
priyajeet: swap the boolean to avoid !

priyajeet: can this be done with just css transform?

src/components/nav-sidebar/NavSidebar.scss
nickgriffin1: This is the most simple solution I could think of. The problem is that the library need to control the display of the element in order to properly animate. I also thought about overwriting the style, but thought this was cleaner.

src/components/nav-sidebar/NavList.js
nickgriffin1: Yes. I figured since we had a library to handle animating height, that it would be the preferred method. I'm definitely open to just writing this with plain old css.

src/components/nav-sidebar/NavSidebar.scss
booc0mtaco: we have something very similar in the count badge component, for comparison

src/components/nav-sidebar/NavList.js
booc0mtaco: looks like recent designs still specify that left side bar carets expand and collapse similar to file explorer conventions like in Finder.app and Windows Explorer. Probably good to double-check if this difference is intended

Screen Shot 2019-05-31 at 11 33 54 AM

priyajeet: The native design is also that way, look at the `<details>` html5 tag.

nickgriffin1: That's a good point. I'm going to stick to what is in the design which is what @booc0mtaco pointed out (similar to Mac finder).

nickgriffin1:
Closing pull request. The animation is not working as expected with some recent changes. Some of the code the CSS changes for the focus of the arrow button will be pulled into a new commit.

PR #1235, opened 2019-05-31
fix(content-explorer): rows don't become stuck after closing upload modal

rickstefanik:


boxcla:

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -763,8 +760,16 @@ class ContentExplorer extends Component<Props, State> {
         this.unselect();
         item.selected = true;
 
+        // ensure that only selected item is marked as selected in currentCollection
+        const newCollection: Collection = { ...this.state.currentCollection };
```
jeremypress: currentCollection is already being pulled out above, can you reuse that reference?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -763,8 +760,16 @@ class ContentExplorer extends Component<Props, State> {
...
+        if (newCollection.items) {
+            for (let i = 0; i < newCollection.items.length; i += 1) {
```
jeremypress: Looks like a good opportunity for `map`. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -391,14 +391,10 @@ class ContentExplorer extends Component<Props, State> {
 
         // New folder state
         const newState = {
-            selected: undefined,
             currentCollection: collection,
             rootName: id === rootFolderId ? name : '',
         };
 
-        // Unselect any rows that were selected
-        this.unselect();
```
jeremypress: Why is this no longer needed?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -763,8 +760,16 @@ class ContentExplorer extends Component<Props, State> {
...
+        const newCollection: Collection = { ...this.state.currentCollection };
+        if (newCollection.items) {
+            for (let i = 0; i < newCollection.items.length; i += 1) {
+                newCollection.items[i].selected = newCollection.items[i].id === item.id;
+            }
+        }
+
         const focusedRow = items.findIndex((i: BoxItem) => i.id === item.id);
-        this.setState({ focusedRow, selected: item }, () => {
+        this.setState({ focusedRow, selected: item, currentCollection: newCollection }, () => {
```
jstoffan: Alphabetize all the things!

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -738,6 +734,7 @@ class ContentExplorer extends Component<Props, State> {
         if (selected) {
             selected.selected = false;
         }
+        this.setState({ selected });
```
ConradJChan: Another good place to avoid state mutation, so instead of mutating the `selected` item in `this.state` and then setting it back, you can simplify this method to something like:
```
const { selected: selectedItem }: State = this.state;
this.setState({ selected: {...selectedItem, selected: false } });
```

ConradJChan:
You can also update your PR title to reflect what element your fix is for, like "fix(content-explorer): ..."

src/elements/content-explorer/ContentExplorer.js
rickstefanik: In some cases, this function is called after closing a modal.  In this scenario, the item that was selected before opening the modal should be selected after closing the modal.

I call this.updateCollection() instead of this.unselect() now.  this.updateCollection() will keep the current item selected if it is in the folder, otherwise it will make everything unselected.

PR #1236, opened 2019-05-31
fix(sidebar): More semantically accessible tabs

ConradJChan:
Moved from `nav` element to `div` with a tablist role

### DIFF
--- a/src/elements/content-sidebar/SidebarNav.js
+++ b/src/elements/content-sidebar/SidebarNav.js
@@ -5,7 +5,7 @@
  */
 
 import * as React from 'react';
-import { FormattedMessage } from 'react-intl';
+import { injectIntl, FormattedMessage } from 'react-intl';
 import IconMagicWand from '../../icons/general/IconMagicWand';
 import IconMetadataThick from '../../icons/general/IconMetadataThick';
 import IconDocInfo from '../../icons/general/IconDocInfo';
@@ -32,7 +32,7 @@ type Props = {
     hasSkills: boolean,
     isOpen?: boolean,
     onNavigate?: (SyntheticEvent<>, NavigateOptions) => void,
-};
+} & InjectIntlProvidedProps;
 
 const SidebarNav = ({
     additionalTabs,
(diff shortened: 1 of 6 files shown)


boxcla:

src/elements/content-sidebar/__tests__/__snapshots__/SidebarNav-test.js.snap
```diff
@@ -1,122 +1,265 @@
...
 
 exports[`elements/content-sidebar/SidebarNav should render the additional tabs loading state 1`] = `
-<SidebarNav
-  additionalTabs={Array []}
-  hasAdditionalTabs={true}
->
-  <nav>
-    <AdditionalTabs
-      tabs={Array []}
+Array [
```
jstoffan: Is this array coming from the `injectIntl` mock in `react-intl-mocks.js`? It seems like it should just be a decorator/passthrough.

src/elements/content-sidebar/SidebarNav.js
```diff
@@ -42,10 +42,11 @@ const SidebarNav = ({
     hasDetails,
     hasMetadata,
     hasSkills,
+    intl,
     isOpen,
     onNavigate,
 }: Props) => (
-    <nav>
+    <div aria-label={intl.formatMessage(messages.sidebarNavLabel)} role="tablist">
```
priyajeet: Have you tried this if you want to avoid adding in the intl hoc

```jsx
<span className="accessibility-hidden">
    <FormattedMessage {...messages. sidebarNavLabel} />
</span>
```

ConradJChan: Oh cool, I didn't realize there was something like this. Just tried this out and it seems to work the same. Do we have a preference on avoiding another hoc? It seems like adding the aria-label seems more semantically correct

src/elements/content-sidebar/__tests__/__snapshots__/SidebarNav-test.js.snap
ConradJChan: Yeah, this was a mistake, due to the new intl mock it's adding a component around SidebarNav with the same name, so enzyme is finding two `SidebarNav` components. I'll update the test to grab the real one

src/elements/content-sidebar/SidebarNav.js
priyajeet: According to this https://accessible360.com/accessible360-blog/use-aria-label-screen-reader-text/ it seems its better to use hidden text just incase screen reader does not support ARIA. Not sure whats better since I have seen both being used at box

ConradJChan: Interesting, a screen reader not supporting ARIA seems like a paradox 😆 . I'll change it to use the hidden text then

PR #1237, opened 2019-06-01
feat(classification): new section within the details sidebar

priyajeet:
BREAKING CHANGE: The api for providing classification data to the sidebar has changed to `{ advisoryMessage, name }`. The classification UI has changed and older version of classification removed.

Screen Shot 2019-06-10 at 4 40 45 PM

Screen Shot 2019-06-10 at 4 37 46 PM

Screen Shot 2019-06-04 at 4 29 19 PM

Also fixes:

Screen Shot 2019-06-04 at 5 36 14 PM

Cleans up a bit of color variable naming


boxcla:

src/components/collapsible/Collapsible.scss
```diff
@@ -6,7 +6,6 @@
 
     .collapsible-card-header {
         align-items: center;
-        background-color: $white;
```
priyajeet: Only bordered versions should have white background otherwise transparent

src/elements/content-sidebar/SidebarFileProperties.js
```diff
@@ -26,73 +25,38 @@ type Props = {
...
     onRetentionPolicyExtendClick,
     isLoading,
     intl,
-}: Props) => {
-    const classificationType = getProp(classification, 'type');
-
-    return (
-        <LoadingIndicatorWrapper isLoading={isLoading}>
-            <ItemProperties
-                classificationProps={
```
priyajeet: 62-73 nuked

src/elements/content-sidebar/DetailsSidebar.scss
```diff
@@ -3,6 +3,10 @@
 .bcs-details {
     .bcs-scroll-content {
         padding-top: 15px;
+
+        > .collapsible-card {
+            margin: 0 8px 20px 25px;
```
jstoffan: Can the collapsible panel be wrapped in a SidebarSection to avoid duplicating these styles?

src/elements/content-sidebar/SidebarClassification.js
```diff
@@ -0,0 +1,58 @@
...
+ * @author Box
+ */
+
+import React from 'react';
+import getProp from 'lodash/get';
+import { FormattedMessage, injectIntl } from 'react-intl';
+
+import Classification from '../../features/classification';
+import messages from '../common/messages';
+import { INTERACTION_TARGET, SECTION_TARGETS, DETAILS_TARGETS } from '../common/interactionTargets';
```
jstoffan: Minor, but can this live down with the imports from `constants?

src/elements/content-sidebar/SidebarClassification.js
```diff
@@ -0,0 +1,58 @@
...
+
+type Props = {
+    classification: ClassificationInfo,
+    file: BoxItem,
+    onEdit?: () => void,
+} & InjectIntlProvidedProps;
+
+const SidebarClassification = ({ classification, file, intl, onEdit }: Props) => {
+    const isEditable = !!onEdit && getProp(file, FIELD_PERMISSIONS_CAN_UPLOAD, false);
+    const hasClassification = !!getProp(classification, 'name');
```
jstoffan: Is `getProp` needed here? Can `classification` be null/undefined?

src/elements/content-sidebar/SidebarClassification.js
```diff
@@ -0,0 +1,58 @@
...
+import Collapsible from '../../components/collapsible';
+import PlainButton from '../../components/plain-button/PlainButton';
+import IconEdit from '../../icons/general/IconEdit';
+import { FIELD_PERMISSIONS_CAN_UPLOAD } from '../../constants';
+import './SidebarClassification.scss';
+
+type Props = {
+    classification: ClassificationInfo,
+    file: BoxItem,
+    onEdit?: () => void,
```
jstoffan: It looks like `onEdit` should accept an `event` parameter, from what I can see.

src/elements/content-sidebar/SidebarClassification.js
```diff
@@ -0,0 +1,58 @@
...
+    }
+    return (
+        <Collapsible
+            buttonProps={{
+                [INTERACTION_TARGET]: SECTION_TARGETS.CLASSIFICATION,
+            }}
+            headerActionItems={
+                isEditable ? (
+                    <PlainButton
+                        aria-label={intl.formatMessage(messages.sidebarClassificationEditBtn)}
```
jstoffan: Can this use hidden text, instead, to avoid the `injectIntl` wrapper?

src/elements/content-sidebar/SidebarClassification.js
```diff
@@ -0,0 +1,58 @@
...
+                        className="bdl-SidebarClassification-editBtn"
+                        data-resin-target={DETAILS_TARGETS.CLASSIFICATION_EDIT}
```
jstoffan: Seems like this component could be used in other sidebars, in theory.

src/elements/content-sidebar/SidebarClassification.scss
```diff
@@ -0,0 +1,6 @@
+@import '../common/variables';
+
+.bdl-SidebarClassification-editBtn {
```
jstoffan: Should this be prefixed with `.bcs`, instead? It doesn't look to be meant as a common component.

src/elements/content-sidebar/SidebarClassification.js
```diff
@@ -0,0 +1,58 @@
...
+    return (
+        <Collapsible
+            buttonProps={{
+                [INTERACTION_TARGET]: SECTION_TARGETS.CLASSIFICATION,
+            }}
+            headerActionItems={
+                isEditable ? (
+                    <PlainButton
+                        aria-label={intl.formatMessage(messages.sidebarClassificationEditBtn)}
+                        className="bdl-SidebarClassification-editBtn"
```
jstoffan: We've been trying to avoid abbreviations and element-aware classnames. Can this be `bcs-SidebarClassificiation-edit`?

src/elements/content-sidebar/SidebarSection.js
```diff
@@ -1,3 +1,5 @@
+// @deprecated, use Collapsible
```
jstoffan: Are we sure this is correct? There are multiple uses of this component that don't collapse.

src/elements/content-sidebar/SidebarClassification.js
priyajeet: Yeah, since it may not be passed in at all or the feature doesn't exist or feature exists but no classification was added

src/elements/content-sidebar/SidebarClassification.js
priyajeet: This was using hidden text in prior patchset. There was a discussion about it in the other PR, @ConradJChan would know the details but aria-label was preferred instead of hidden span, so I reverted to use aria-label too

src/elements/content-sidebar/SidebarSection.js
priyajeet: This was meant to be replaced by Collapsible since that was created using code from here + additional stuff. But we never went back and nuked this one out.

src/elements/content-sidebar/SidebarClassification.js
priyajeet: Design has changed a bit, no longer using icon, and using words. So will have to create a new component for this anyways

src/elements/content-sidebar/SidebarClassification.js
priyajeet: Wont need it anymore, changing to words instead of an icon

src/elements/content-sidebar/DetailsSidebar.scss
priyajeet: Depends, if we refactor that to just use collapsible then yeah. But if we just nuke that and use collapsible directly, then over here will work for all "cards" in the details sidebar.

src/elements/content-sidebar/SidebarClassification.scss
priyajeet: Actually on second thoughts, everything within this project should be bdl as we have been doing at other places. Having more than 1 namespace just adds more than it achieves. Right now there is bdl, be, bcs etc. A single namespace should suffice. And if be/bcs will carry styling, it should be bdl-be/bcs if really needed or one can just switch to mixins instead.

i18n/en-US.properties
```diff
@@ -547,11 +547,15 @@ boxui.breadcrumb.breadcrumbLabel = Breadcrumb
 # Icon to display more information on the checkbox
 boxui.checkboxTooltip.iconInfoText = Info
 # Button to add classification on an item
+boxui.classification.add = Add
+# Button to add classification on an item
 boxui.classification.addClassification = Add Classification
 # Header for classification section in sidebar
 boxui.classification.classification = Classification
 # Button to edit classification on an item
 boxui.classification.edit = Edit
```
priyajeet: unfortunate dupe :(

src/elements/content-sidebar/SidebarClassification.js
priyajeet: I fixed the flowtype to have ?

src/features/classification/EditClassificationButton.js
```diff
@@ -0,0 +1,35 @@
...
+import messages from './messages';
+import PlainButton from '../../components/plain-button/PlainButton';
+
+type Props = {
+    className?: string,
+    isEditing: boolean,
+    onEdit: (e: SyntheticEvent<HTMLButtonElement>) => void,
+};
+
+const EditClassificationButton = ({ className = '', isEditing, onEdit, ...rest }: Props) => {
```
priyajeet: Maybe in the future it can become more generic if there are more use cases

src/styles/_variables.scss
```diff
@@ -1,25 +1,51 @@
-// Box Design Language Theme Colors
+// Box Design Language Colors, confirm with design before extending
+// @see https://gallery.io/projects/MCHbtQVoQ2HCZTdZ0F79-fQJ/files/MCEJu8Y2hyDScSTYfw7hxWcA-L66cSQvSj4
+
+// Box/Primary Blues
 $bdl-box-blue: #0061d5 !default;
-$bdl-secondary-blue: #003c84 !default;
+$bdl-box-blue-50: #619de5 !default;
```
Sunyang730: Can't find the usage of this color in this PR might be good to update this on another PR? Will this be include in the release note as something addition?

src/features/classification/Classification.js
```diff
@@ -1,25 +1,48 @@
...
+    const isClassified = !!name;
     const hasMessage = !!advisoryMessage;
-    const isTooltipEnabled = hasMessage && !isMessageInline;
-    const isInlineMessageVisible = hasMessage && isMessageInline;
+
+    const isTooltipMessageEnabled = isClassified && hasMessage && messageStyle === STYLE_TOOLTIP;
+    const isInlineMessageEnabled = isClassified && hasMessage && messageStyle === STYLE_INLINE;
+
+    // Either the add classification badge should be visible or the "not classified" text or neither
+    const isNotClassifiedBadgeVisible = !isClassified && !messageStyle;
```
jpfranco: Just curious as to why this depends on `messageStyle`. Name might also be a bit clearer as `isClassifiedBadgeHidden`.

src/styles/_variables.scss
```diff
@@ -1,25 +1,51 @@
...
-// Box Design Language Neutral Colors
-$bdl-neutral-01: #222 !default;
-$bdl-neutral-02: #767676 !default;
-$bdl-neutral-03: #909090 !default;
+// Do not use, deprecated colors below this line
+$bdl-secondary-blue: $bdl-dark-blue !default;
+$bdl-neutral-01: $bdl-gray !default;
+$bdl-neutral-02: $bdl-gray-62 !default;
+$bdl-neutral-03: $bdl-gray-50 !default;
 $bdl-neutral-04: #eaeaea !default;
```
alexkrolick: these should be remapped as well, the 100-point scale is an evolution of this and they should have the same source of truth

src/features/classification/Classification.js
priyajeet: Gonna change it to isAddClassificationBadgeVisible

src/styles/_variables.scss
priyajeet: It was already there, few lines below

src/features/classification/messages.js
```diff
@@ -1,6 +1,11 @@
 import { defineMessages } from 'react-intl';
 
 const messages = defineMessages({
+    add: {
+        defaultMessage: 'Add',
+        description: 'Button to add classification on an item',
+        id: 'boxui.classification.add',
```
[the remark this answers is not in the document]
priyajeet: Yeah thats something I have been thinking of. Gonna note it down as a todo.

src/elements/content-sidebar/__tests__/SidebarClassification-test.js
```diff
@@ -0,0 +1,88 @@
+import React from 'react';
```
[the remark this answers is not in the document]
priyajeet: For the @deprecated stuff, no, it doesn't. Just for easier searching for future refactor. Normally it should be @TODO (name), but we dont put names in OSS

src/elements/content-sidebar/__tests__/SidebarClassification-test.js
```diff
@@ -0,0 +1,88 @@
...
+import { shallow } from 'enzyme';
+
+import Collapsible from '../../../components/collapsible';
+import SidebarClassification from '../SidebarClassification';
+
+describe('elements/content-sidebar/SidebarClassification', () => {
+    const getWrapper = props => shallow(<SidebarClassification {...props} />);
+
+    describe('render()', () => {
+        test('should render nothing when unclassified and can_upload is false', () => {
```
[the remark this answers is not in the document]
priyajeet: Unclassified is actually an incorrect term, because unclassified is a sort of classification, you can have an unclassified label. Changing it to not classified.

src/features/classification/AddClassificationBadge.scss
```diff
@@ -0,0 +1,19 @@
+@import '../../styles/variables';
+@import './mixins';
+
+.bdl-AddClassificationBadge {
+    @include bdl-ClassificationBadge;
+
+    background-color: $bdl-gray-02;
+    border-color: $bdl-gray-20;
+
+    &:hover {
```
jstoffan: Is `&:focus` needed to support keyboard navigation, as well?

src/features/classification/AddClassificationBadge.scss
```diff
@@ -0,0 +1,19 @@
...
+        background-color: $bdl-gray-05;
+    }
+}
+
+.bdl-AddClassificationBadge-name {
+    @include bdl-ClassificationBadge-name;
+
+    color: $bdl-gray-50 !important; // To inhibit overriding
```
jstoffan: We've managed to avoid this in most cases. What would override it?

src/features/classification/Classification.js
```diff
@@ -1,25 +1,48 @@
...
-            </Tooltip>
-            {isInlineMessageVisible && <p className="bdl-Classification-advisoryMessage">{advisoryMessage}</p>}
+        <article className={`bdl-Classification ${className}`}>
+            {isClassified && (
+                <ClassifiedBadge
+                    name={((name: any): string)}
+                    tooltipText={isTooltipMessageEnabled ? advisoryMessage : undefined}
+                />
+            )}
+            {isNotClassifiedBadgeVisible && <AddClassificationBadge />}
```
jstoffan: Can we avoid the negation in the variable names? What is the positive case?

src/features/classification/Classification.js
```diff
@@ -1,25 +1,48 @@
...
+            {isInlineMessageEnabled && <p className="bdl-Classification-advisoryMessage">{advisoryMessage}</p>}
+            {isNotClassifiedMessageVisible && (
+                <span className="bdl-Classification-missingMessage">
```
jstoffan: This will create a nested span. You can avoid it by using the inner render function style inside FormattedMessage.

src/features/classification/ClassifiedBadge.scss
```diff
@@ -0,0 +1,16 @@
+@import '../../styles/variables';
+@import './mixins';
+
+.bdl-ClassifiedBadge,
+.bdl-ClassifiedBadge:hover {
```
jstoffan: Can this live as `&:hover`, as the other one does? Is `&:focus` needed?

src/features/classification/messages.js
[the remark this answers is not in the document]
priyajeet: For others reading this comment, he is referring to having this on top of each message file

`const scope = `boxui.classification`;`

And then for each ID you have `` `${scope}.add` ``

Note that we rarely (never?) search code by IDs anyways, so concatenation is not a problem here. Prevents repetition and saves some bytes.

src/features/classification/_mixins.scss
```diff
@@ -0,0 +1,18 @@
...
+    align-items: center;
+    border: 1px solid;
+    border-radius: $bdl-border-radius-size;
+    display: inline-flex;
+    margin: 0;
+    padding: 2px 5px;
+}
+
+@mixin bdl-ClassificationBadge-name {
+    font-size: 10px !important; // To inhibit overriding
```
jstoffan: Can we avoid this? Overriding by whom? The parent application?

src/styles/_variables.scss
priyajeet: Let me see if I can map it to the closest grays, hmm

src/icons/general/IconAddTags.js
```diff
@@ -2,30 +2,33 @@
...
     /** A text-only string describing the icon if it's not purely decorative for accessibility */
     title?: string | React.Element<any>,
     width?: number,
 };
 
-const IconAddTags = ({ className = '', height = 15, color = '#444', title, width = 15 }: Props) => (
+const IconAddTags = ({ className = '', height = 32, color = bdlGray, title, strokeWidth = 2, width = 32 }: Props) => (
     <AccessibleSVG
-        className={`icon-add-tags ${className}`}
+        className={`bdl-iconAddTags ${className}`}
```
jstoffan: I think we've been using PascalCase for other icons, so `bdl-IconAddTags`.

src/styles/_variables.scss
```diff
@@ -1,25 +1,51 @@
...
+
+// Box/Primary Blues
 $bdl-box-blue: #0061d5 !default;
-$bdl-secondary-blue: #003c84 !default;
+$bdl-box-blue-50: #619de5 !default;
+$bdl-box-blue-05: #f2f7fd !default;
+
+// Box/Primary Grays
+$bdl-gray: #222 !default;
+$bdl-gray-62: #767676 !default;
```
jstoffan: It looks like in some places we sort ascending and others descending. Maybe standardize on the former?

src/styles/_variables.scss
priyajeet: I only added a couple of new ones that I had to use and made explicit about them grays, rest were already in the file.

src/features/classification/AddClassificationBadge.scss
priyajeet: It does get the blue outline because its wrapped within a button with a on click. So the badge itself is fine. We can tho choose to move the button in here and pass in click handlers/resin tags

src/features/classification/AddClassificationBadge.scss
priyajeet: Very unfortunate, there is code both in EUA and here that adds styling on direct tags (h1 etc) and on * that I couldn't fix without a much bigger refactor or without duplicating styles that are here and then again in EUA. Line 34 here https://github.com/box/box-ui-elements/blob/master/src/components/breadcrumb/Breadcrumb.scss#L34

This was the least offensive way for now.

src/features/classification/Classification.js
priyajeet: Fixed this with prior comment.
But the other one about the message, its called "not classified" message, so the negation is the output

src/features/classification/Classification.js
priyajeet: Yeah nested span is fine here, and in almost every other case too. Except the other file  where nested span causes issues due to styles being added to *s

src/features/classification/_mixins.scss
priyajeet: yeah, see other comment

src/icons/general/IconAddTags.js
priyajeet: good catch

src/features/classification/ClassifiedBadge.scss
priyajeet: I think I could but it was just the same style since this case badge doesn't have a hover state, same as non-hovered state. The badge ends up in breadcrumbs that have hover state, so this is just overriding it. The badge till be in a button, so the parent has its focus and hover.

src/styles/_variables.scss
booc0mtaco: Could we update the colors.md to show the new palette / categories documented here?

src/elements/content-sidebar/DetailsSidebar.js
```diff
@@ -44,7 +45,7 @@ type ExternalProps = {
     hasRetentionPolicy?: boolean,
     hasVersions?: boolean,
     onAccessStatsClick?: Function,
-    onClassificationClick: Function,
+    onClassificationClick?: Function,
```
Sunyang730: Is there a reason why we want to just use Function and not specify what kind of function it could be?

src/features/classification/messages.js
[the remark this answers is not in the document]
alexkrolick: That makes it a bit harder to use the localization helper in e2e tests and to figure out if a translation can be deleted. Not sure the abstraction is needed.

src/elements/content-sidebar/DetailsSidebar.js
priyajeet: for new stuff we are putting all specifics, this is just some dangling legacy stuff

priyajeet: i fixed it

src/features/classification/messages.js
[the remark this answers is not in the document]
priyajeet: Thats the pattern being used in AC now at least and maybe EUA too?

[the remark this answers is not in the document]
priyajeet: In any case worth considering

PR #1241, opened 2019-06-04
fix(iconworkflow): make background rect same as viewBox

frogzsj:


boxcla:

PR #1242, opened 2019-06-04
fix(icons): give collections Icons SUIT classNames

nickgriffin1:
Fixes classname for 3 near brand new icons so that they fit the SUIT naming convention.

### DIFF
--- a/src/icons/collections/IconCollections.js
+++ b/src/icons/collections/IconCollections.js
@@ -24,7 +24,7 @@ const IconCollections = ({
     width = 32,
 }: Props) => (
     <AccessibleSVG
-        className={classNames('icon-collections', className, {
+        className={classNames('IconCollections', className, {
             'is-selected': selected,
         })}
         height={height}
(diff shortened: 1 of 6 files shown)


boxcla:

alexkrolick:
I thought we were adding a bdl- or box- prefix to the suit names? @jstoffan 👀

src/icons/collections/IconCollections.js
```diff
@@ -24,7 +24,7 @@ const IconCollections = ({
     width = 32,
 }: Props) => (
     <AccessibleSVG
-        className={classNames('icon-collections', className, {
+        className={classNames('IconCollections', className, {
```
jstoffan: Common component class names should be prefixed with `bdl-`, so this would be `bdl-IconCollections`.

nickgriffin1: Ah I was under the impression that only shared variable names and common class names  such as "Button" were prefixed with `bdl`. Will fix.

PR #1243, opened 2019-06-05
fix(api): Fix avatar resolution in activity sidebar

swaroopbutala:
Default avatar url returned by the API is 28*28. The avatar is 32*32 in the UI which makes the avatar look blurry. Fixing avatar urls in activity sidebar to add the pic_type param so we can specify next size up in the API for avatars which will return 40*40 image resolution.

### DIFF
--- a/src/api/Users.js
+++ b/src/api/Users.js
@@ -54,7 +54,14 @@ class Users extends Base {
         const accessToken: TokenLiteral = await TokenService.getReadToken(getTypedFileId(fileId), this.options.token);
 
         if (typeof accessToken === 'string') {
-            const url = `${this.getAvatarUrl(userId)}?access_token=${accessToken}`;
+            const options = {
+                access_token: accessToken,
+                pic_type: 'large',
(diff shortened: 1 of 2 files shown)


boxcla:
Verified that @swaroopbutala has signed the CLA. Thanks for the pull request!

src/api/Users.js
```diff
@@ -54,7 +54,12 @@ class Users extends Base {
         const accessToken: TokenLiteral = await TokenService.getReadToken(getTypedFileId(fileId), this.options.token);
 
         if (typeof accessToken === 'string') {
-            const url = `${this.getAvatarUrl(userId)}?access_token=${accessToken}`;
+            const options = {
+                access_token: accessToken,
+                pic_type: 'large',
+            };
+            const urlParams = Object.keys(options).map(key => key + '=' + options[key]).join('&');
```
jstoffan: You can use `queryString.stringify` from the `query-string` module to build this string.

PR #1245, opened 2019-06-05
fix(content-explorer): rows no longer stuck after closing modal

rickstefanik:
New pull request for stuck rows bug, since original got messed up.


boxcla:

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -388,16 +388,18 @@ class ContentExplorer extends Component<Props, State> {
     fetchFolderSuccessCallback(collection: Collection, triggerNavigationEvent: boolean): void {
         const { onNavigate, rootFolderId }: Props = this.props;
         const { id, name, boxItem }: Collection = collection;
+        const { selected }: State = this.state;
 
         // New folder state
         const newState = {
+            currentCollection: { ...collection },
```
ConradJChan: I think this spread is unnecessary because the `collection` passed in as an argument is a new object already from the API layer

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -388,16 +388,18 @@ class ContentExplorer extends Component<Props, State> {
...
             selected: undefined,
-            currentCollection: collection,
             rootName: id === rootFolderId ? name : '',
         };
 
-        // Unselect any rows that were selected
-        this.unselect();
+        const { newCollection, newSelected } = this.updateCollection(collection, selected);
+        newState.currentCollection = newCollection;
+        newState.selected = newSelected;
```
ConradJChan: I think you could move L400 above and then declare the `newState` one time instead of doing these reassignments

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -761,12 +795,16 @@ class ContentExplorer extends Component<Props, State> {
...
 
         this.unselect();
-        item.selected = true;
 
-        const focusedRow = items.findIndex((i: BoxItem) => i.id === item.id);
-        this.setState({ focusedRow, selected: item }, () => {
-            onSelect(cloneDeep([item]));
-            callback(item);
+        const selectedItem: BoxItem = { ...item };
+        selectedItem.selected = true;
```
ConradJChan: can do this in one line: `const selectedItem: BoxItem = { ...item, selected: true };`

test/integration/content-explorer/ContentExplorer.e2e.test.js
```diff
@@ -0,0 +1,207 @@
...
+const helpers = {
+    load(additionalProps = { canUpload: true, permissions: { canUpload: true } }) {
+        cy.visit('/Elements/ContentExplorer', {
+            onBeforeLoad: contentWindow => {
+                contentWindow.PROPS = additionalProps;
+            },
+        });
+    },
+    getRow(rowNum) {
+        return cy.getByDataPreview('ContentExplorer').find(`.bce-item-row-${rowNum}`);
```
ConradJChan: A Cypress best practice we've been trying to follow is to use data-testid attributes to select elements of interests in our tests. You could do the same here and instrument the code with data-testid attribute for each row and then use `cy.getByTestId` like we do in other tests

test/integration/content-explorer/ContentExplorer.e2e.test.js
```diff
@@ -0,0 +1,207 @@
...
+    },
+    checkRows(rowNum) {
+        cy.getByDataPreview('ContentExplorer')
+            .find('.ReactVirtualized__Table')
+            .then(table => {
```
ConradJChan: Similarly, we could instrument the Table with a data-testid instead

test/integration/content-explorer/ContentExplorer.e2e.test.js
```diff
@@ -0,0 +1,207 @@
...
+                const totalRows = table[0].getAttribute('aria-rowcount');
+
+                for (let i = 0; i < totalRows; i += 1) {
+                    if (i !== rowNum) {
+                        this.getRow(i).should('not.have.class', 'bce-item-row-selected');
+                    }
+                }
+                if (rowNum >= 0) {
```
ConradJChan: could we combine the case where rowNum exists as part of the for loop?

test/integration/content-explorer/ContentExplorer.e2e.test.js
```diff
@@ -0,0 +1,207 @@
...
+        cy.visit('/Elements/ContentExplorer', {
+            onBeforeLoad: contentWindow => {
+                contentWindow.PROPS = additionalProps;
+            },
+        });
+    },
+    getRow(rowNum) {
+        return cy.getByDataPreview('ContentExplorer').find(`.bce-item-row-${rowNum}`);
+    },
+    checkRows(rowNum) {
```
ConradJChan: Maybe rename the method to be more explicit about what it does, perhaps `checkRowSelections` and `rowNum` => `selectedRow`

test/integration/content-explorer/ContentExplorer.e2e.test.js
```diff
@@ -0,0 +1,207 @@
...
+            .as('row')
+            .click()
+            .should('have.class', 'bce-item-row-selected');
+
+        // Whenever a row is selected, confirm that it is the only row that appears selected
+        this.checkRows(rowNum);
+        return cy.get('@row');
+    },
+    clickAddButton() {
+        cy.getByDataPreview('ContentExplorer')
```
ConradJChan: Similarly, we can instrument all these buttons with `data-testid`'s

test/integration/content-explorer/ContentExplorer.e2e.test.js
```diff
@@ -0,0 +1,207 @@
...
+describe('ContentExplorer', () => {
+    beforeEach(() => {
+        cy.server();
+        cy.route('GET', '**/folders/*', 'fixture:content-explorer/root-folder.json');
+    });
+
+    describe('Selection', () => {
+        it('Should not have a selected row to start', () => {
+            helpers.load();
+            helpers.checkRows(-1);
```
ConradJChan: Probably can just call it without an argument and achieve the same effect?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -725,6 +729,38 @@ class ContentExplorer extends Component<Props, State> {
...
+     * newCollection will be the given collection, with items.selected properties
+     * updated according to the given selected param. newSelected will be the selected
+     * item if it is in newCollection, otherwise undefined.
+     *
+     * @private
+     * @param {Collection} collection - collection that needs to be updated
+     * @param {?BoxItem} selected - item that should be selected in that collection (if present)
+     * @return {Object}
+     */
+    updateCollection(collection: Collection, selected: ?BoxItem): Object {
```
ConradJChan: Would it make sense to simply call `setState` here since that seems to be how it's used?

test/integration/content-explorer/ContentExplorer.e2e.test.js
rickstefanik: The rows are children of a react-virtualized table, and the table doesn't expose the html attributes of its individual rows, so I don't think I can add a data-testid attribute for the rows.  (I will add the data-testid attributes for the other components that we have control over.)

test/integration/content-explorer/ContentExplorer.e2e.test.js
rickstefanik: Same situation as above in that I don't think it's possible to add a data-testid to something inside the react-virtualized library.  I was able to add data-testid's to everything else in the file though.

test/fixtures/content-explorer/root-folder.json
```diff
@@ -0,0 +1,682 @@
+{
```
jeremypress: Any reason why we have to mock the network? The explorer should be able to return all of this info with our public token right?

test/integration/content-explorer/ContentExplorer.e2e.test.js
jstoffan: It seems like this could be simplified quite a bit by using more advanced chai or [chai-jquery selectors](https://docs.cypress.io/guides/references/assertions.html#Chai-jQuery), such as `.descendants`.

test/fixtures/content-explorer/root-folder.json
rickstefanik: The default page doesn't have the upload and share buttons, so mocking the network was the only way I could get them to appear.  Can I get the upload and share buttons to appear without mocking?

test/integration/content-explorer/ContentExplorer.e2e.test.js
```diff
@@ -0,0 +1,207 @@
...
+            .click();
+    },
+    clickMoreOptionsButton(rowNum) {
+        this.getRow(rowNum)
+            .find('.bce-btn-more-options')
+            .click();
+    },
+    clickRenameButton() {
+        cy.get('.dropdown-menu-element')
+            .contains('Rename')
```
jstoffan: User-facing text selectors should probably use the i18n `localize` helper method.

test/integration/content-explorer/ContentExplorer.e2e.test.js
```diff
@@ -0,0 +1,207 @@
+// <reference types="Cypress" />
+const helpers = {
+    load(additionalProps = { canUpload: true, permissions: { canUpload: true } }) {
```
jstoffan: Is `permissions` exposed as a top-level prop in ContentExplorer?

test/integration/content-explorer/ContentExplorer.e2e.test.js
```diff
@@ -0,0 +1,207 @@
...
+                if (rowNum >= 0) {
+                    this.getRow(rowNum).should('have.class', 'bce-item-row-selected');
+                }
+            });
+    },
+    selectRow(rowNum) {
+        this.getRow(rowNum)
+            .as('row')
+            .click()
+            .should('have.class', 'bce-item-row-selected');
```
jstoffan: This class name appears more than twice, so should probably be a constant or helper method.

test/integration/content-explorer/ContentExplorer.e2e.test.js
```diff
@@ -0,0 +1,207 @@
...
+
+        // Whenever a row is selected, confirm that it is the only row that appears selected
+        this.checkRows(rowNum);
```
jstoffan: This check should be delegated to calling code to simplify this helper method.

test/integration/content-explorer/ContentExplorer.e2e.test.js
```diff
@@ -0,0 +1,207 @@
...
+            .contains('Cancel')
+            .click();
+    },
+    clickNewFolderCreateButton() {
+        cy.get('.be-modal-btns > button')
+            .contains('Create')
+            .click();
+    },
+    clickCloseUploadModal() {
+        cy.get('.bcu-footer-left > button')
```
jstoffan: It's not usually necessary to combine `get` and `contains`, as the latter usually only matches a single item. We should also avoid using CSS selectors (and especially descendent element selectors) whenever possible in Cypress tests.

test/integration/content-explorer/ContentExplorer.e2e.test.js
```diff
@@ -0,0 +1,207 @@
...
+            helpers.selectRow(5);
+        });
+
+        it('Should open and close upload modal', () => {
+            helpers.load();
+            helpers.selectRow(2);
+            helpers.clickAddButton();
+            helpers.clickUploadButton();
+            helpers.clickCloseUploadModal();
+            helpers.checkRows(2);
```
jstoffan: It's unclear to me what these tests are doing from reading these lines. I think the assertions should live directly in the test rather than in helpers.

test/support/commands.js
```diff
@@ -1,3 +1,4 @@
 Cypress.Commands.add('getByTestId', testId => cy.get(`[data-testid="${testId}"]`));
 Cypress.Commands.add('getByTitle', title => cy.get(`[title="${title}"]`));
 Cypress.Commands.add('getTooltip', () => cy.get('[role="tooltip"]'));
+Cypress.Commands.add('getByDataPreview', dataPreview => cy.get(`[data-preview="${dataPreview}"]`));
```
jstoffan: We should try to avoid selecting on custom Styleguidist properties like this in case we want to migrate to a different framework. Navigating to the page should be enough to ensure that the test is executing on the correct element.

test/fixtures/content-explorer/root-folder.json
jstoffan: You should be able to intercept the server response and decorate it with the specific mocks you need rather than mocking the entire response.

jstoffan: Take a look at the [Response Functions section](https://docs.cypress.io/api/commands/route.html#With-Stubbing) of the Cypress docs on `cy.route`.

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -725,6 +729,38 @@ class ContentExplorer extends Component<Props, State> {
...
+        const newCollection: Collection = { ...collection };
+        const targetID = selected ? selected.id : null;
+
+        const ret = { newCollection, newSelected: undefined };
```
jstoffan: It may be more clear to build this object at the time when it's returned:

```
return {
   newCollection,
   newSelected: undefined,
};
```

It also seems a little odd that `updateCollection` returns a wrapper around the collection rather than the collection itself.

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -725,6 +729,38 @@ class ContentExplorer extends Component<Props, State> {
...
+
+        if (collection.items) {
+            ret.newCollection.items = collection.items.map(obj => {
+                const item = { ...obj, selected: obj.id === targetID };
```
jstoffan: It seems like we're tracking the same piece of information, `selected`,  twice: once in the collection and once in a top-level state property. Consolidating to one or the other seems like it would simplify things a bit.

test/integration/content-explorer/ContentExplorer.e2e.test.js
rickstefanik: No, it's not.  That line is an artifact of me trying to get permissions to work for testing.  I'll take it out in the next commit.

src/components/button/Button.js
```diff
@@ -45,6 +46,7 @@ class Button extends React.Component<Props> {
             setRef,
             type = 'submit',
             showRadar = false,
+            dataTestId,
             ...rest
```
jstoffan: I think this should forward `data-testid` to `buttonProps` automatically without needing to pass it in as a separate camelCase prop.

src/components/menu/MenuItem.js
```diff
@@ -33,7 +34,16 @@ class MenuItem extends React.Component<Props> {
...
-        const { children, className, isDisabled, isSelectItem, isSelected, showRadar, ...rest } = this.props;
+        const {
+            children,
+            className,
+            dataTestId,
+            isDisabled,
+            isSelectItem,
+            isSelected,
+            showRadar,
+            ...rest
```
jstoffan: Looks like `data-testid` will get forwarded to `menuItemProps` automatically.

src/elements/common/sub-header/Add.js
```diff
@@ -24,12 +24,12 @@ const Add = ({ onUpload, onCreate, showUpload = true, showCreate = true }: Props
         <AddButton />
         <Menu>
             {showUpload && (
-                <MenuItem onClick={onUpload}>
+                <MenuItem dataTestId="be-btn-add-upload" onClick={onUpload}>
```
jstoffan: We should try to select these and other elements via their user-facing text content, `messages.upload`, rather than data-testid whenever possible.

test/integration/content-explorer/ContentExplorer.e2e.test.js
```diff
@@ -0,0 +1,201 @@
...
+    },
+    previewItemFromRow(rowNum) {
+        this.getItemNameFromRow(rowNum).click();
+    },
+};
+
+describe('ContentExplorer', () => {
+    beforeEach(() => {
+        cy.server();
+        cy.route('GET', '**/folders/*', 'fixture:content-explorer/root-folder.json');
```
jstoffan: We should probably rename this file to `ContentExplorer.int.test.js`, since its backend is stubbed out.

test/integration/content-explorer/ContentExplorer.e2e.test.js
```diff
@@ -0,0 +1,201 @@
...
+            cy.getByTestId('content-explorer')
+                .find(`${selectedRowClassSelector}`)
+                .should('have.length', 1)
+                .should('have.class', `bce-item-row-${selectedRow}`);
+        } else {
+            cy.getByTestId('content-explorer').should('not.have.descendants', selectedRowClassSelector);
+        }
+    },
+    selectRow(rowNum) {
+        this.getRow(rowNum)
```
ConradJChan: I think you can just `return` this chain because the yielded subject should still be the row after the click and should assertion

src/elements/content-explorer/ContentExplorer.js
rickstefanik: My thinking on not calling setState in updateCollection is the following scenario: someone wants to set the state of collection, selected, and some other piece(s) of state, and then have a callback function.  This happens in the select function of ContentExplorer. If setState is called in updateCollection, then that callback wouldn't function the same, since collection and selected were set earlier .  Am I thinking about this correctly, or am I misunderstanding how callbacks from setState work?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -728,17 +717,17 @@ class ContentExplorer extends Component<Props, State> {
...
-     * item if it is in newCollection, otherwise undefined.
+     * Sets state with currentCollection updated to have items.selected properties
+     * set according to the given selected param. selected will be set to the selected
+     * item if it is in currentCollection, otherwise it will be set to undefined.
      *
      * @private
      * @param {Collection} collection - collection that needs to be updated
      * @param {?BoxItem} selected - item that should be selected in that collection (if present)
+     * @param {Function} callback - callback function that should be called after setState occurs
      * @return {Object}
```
ConradJChan: can update the @return to be `@return {void}`

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -728,17 +717,17 @@ class ContentExplorer extends Component<Props, State> {
...
-     * updated according to the given selected param. newSelected will be the selected
-     * item if it is in newCollection, otherwise undefined.
+     * Sets state with currentCollection updated to have items.selected properties
+     * set according to the given selected param. selected will be set to the selected
+     * item if it is in currentCollection, otherwise it will be set to undefined.
      *
      * @private
      * @param {Collection} collection - collection that needs to be updated
      * @param {?BoxItem} selected - item that should be selected in that collection (if present)
+     * @param {Function} callback - callback function that should be called after setState occurs
```
ConradJChan: optional parameters you can put in brackets `[callback]`

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -754,8 +743,9 @@ class ContentExplorer extends Component<Props, State> {
                 return item;
             });
         }
-
-        return { newCollection, newSelected };
+        this.setState({ currentCollection: newCollection, selected: newSelected }, () => {
```
ConradJChan: Can probably leave out the extra anonymous function and just pass in the callback reference directly

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -768,9 +758,7 @@ class ContentExplorer extends Component<Props, State> {
      */
     unselect(): void {
         const { currentCollection }: State = this.state;
-
-        const { newCollection, newSelected } = this.updateCollection(currentCollection, null);
-        this.setState({ currentCollection: newCollection, selected: newSelected });
+        this.updateCollection(currentCollection, null);
```
ConradJChan: Can we call `updateCollection` without `null`?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -795,10 +783,11 @@ class ContentExplorer extends Component<Props, State> {
 
         const selectedItem: BoxItem = { ...item, selected: true };
 
-        const { newCollection } = this.updateCollection(currentCollection, selectedItem);
+        this.updateCollection(currentCollection, selectedItem);
+
         const focusedRow: number = items.findIndex((i: BoxItem) => i.id === item.id);
 
-        this.setState({ currentCollection: newCollection, focusedRow, selected: selectedItem }, () => {
+        this.setState({ focusedRow }, () => {
```
ConradJChan: It seems like this callback should actually be used on L786 instead since it depends on `selectedItem` being set to state

PR #1249, opened 2019-06-06
fix(beta-feedback): update aria-labelledby attribute name

booc0mtaco:
Sometimes, you're supposed to use camelCase names for DOM attributes, and sometimes you're not. Fix this case where it should use all lower-case for an `aria-` attribute

<https://reactjs.org/blog/2017/09/08/dom-attributes-in-react-16.html>


boxcla:

PR #1253, opened 2019-06-07
fix(footer-indicator): move tooltip to appear above footer indicator

sahiga:
The current footer indicator tooltip appears to the right of the indicator, which causes unwanted scrollbars to appear in certain browsers and pages and long text to get cut off at the bottom of the viewport.

current-footer-indicator-tooltip

This pull request fixes the above issues by changing the position of the footer indicator tooltip from `middle-right` to `top-right`.

proposed-footer-indicator-tooltip


boxcla:

PR #1254, opened 2019-06-07
refactor(content-sidebar): make TaskModal separate component

booc0mtaco:
remove task modal from add task button component so that it can be used in other contexts

### DIFF
--- a/src/elements/content-sidebar/AddTaskButton.js
+++ b/src/elements/content-sidebar/AddTaskButton.js
@@ -3,23 +3,14 @@ import * as React from 'react';
 import AddTaskMenu from './AddTaskMenu';
 import TaskModal from './TaskModal';
 import { TASK_TYPE_APPROVAL } from '../../constants';
+import type { TaskFormProps } from './activity-feed/task-form/TaskForm';
 
-type AddTaskButtonProps = {|
+type Props = {|
     feedbackUrl: string,
     isDisabled: boolean,
     onTaskModalClose: () => void,
+    taskFormProps: TaskFormProps,
 |};
-
-// These are used by the wrapped form
-type PassThroughProps = {|
-    approverSelectorContacts: any,
-    className?: string,
-    createTask: any,
-    getApproverWithQuery?: any,
-    getAvatarUrl: any,
-|};
-
-type Props = {| ...AddTaskButtonProps, ...PassThroughProps |};
 
 type State = {
     error: ?ElementsXhrError,
@@ -50,7 +41,7 @@ class AddTaskButton extends React.Component<Props, State> {
     handleCreateError = (e: ElementsXhrError) => this.setState({ error: e });
 
     render() {
-        const { isDisabled, feedbackUrl, ...passThrough } = this.props;
+        const { isDisabled, feedbackUrl, taskFormProps } = this.props;
         const { isTaskFormOpen, taskType, error } = this.state;
 
         return (
@@ -63,7 +54,7 @@ class AddTaskButton extends React.Component<Props, State> {
                     handleCreateSuccess={this.handleCreateSuccess}
                     handleModalClose={this.handleModalClose}
                     isTaskFormOpen={isTaskFormOpen}
-                    passThrough={passThrough}
+                    taskFormProps={taskFormProps}
                     taskType={taskType}
                 />
             </React.Fragment>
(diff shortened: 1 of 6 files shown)


boxcla:

src/elements/content-sidebar/AddTaskButton.js
```diff
@@ -67,36 +53,19 @@ class AddTaskButton extends React.Component<Props, State> {
...
-                    </div>
-                </Modal>
+                <TaskModal
+                    error={error}
+                    feedbackUrl={feedbackUrl}
+                    handleCreateError={this.handleCreateError}
+                    handleCreateSuccess={this.handleCreateSuccess}
+                    handleModalClose={this.handleModalClose}
+                    isTaskFormOpen={isTaskFormOpen}
+                    passThrough={passThrough}
```
alexkrolick: How about changing this to "taskFormProps" or something since it is now a public interface rather than a private variable?

alexkrolick: Or continue the pattern and {..rest} the props that are not recognized?

alexkrolick: IRL:

- rename to `taskFormProps`, add to both AddTaskButton and TaskModal
- export type from TaskForm to converge on source of truth for the form props

src/elements/content-sidebar/TaskModal.js
```diff
@@ -0,0 +1,64 @@
...
+
+const TaskModal = (props: TaskModalProps) => {
+    // CSS selector for first form element
+    // Note: Modal throws an error if this fails to find an element!
+    return (
+        <Modal
+            className="be-modal task-modal"
+            data-testid="create-task-modal"
+            focusElementSelector={focusTargetSelector}
+            isOpen={props.isTaskFormOpen}
```
jstoffan: Props should be destructured in the function signature.

src/elements/content-sidebar/TaskModal.js
```diff
@@ -0,0 +1,64 @@
...
+            title={
+                <React.Fragment>
+                    <FormattedMessage {...getMessageForModalTitle(props.taskType)} />
+                    <BetaFeedbackBadge tooltip formUrl={props.feedbackUrl} />
+                </React.Fragment>
+            }
+        >
+            <div className="be">
+                <TaskForm
+                    {...props.passThrough}
```
jstoffan: Should this be at the bottom of the prop list to override the defaults?

src/elements/content-sidebar/TaskModal.js
```diff
@@ -0,0 +1,64 @@
...
+    handleCreateSuccess: Function,
+    handleModalClose: Function,
+    isTaskFormOpen: boolean,
+    passThrough: Object,
+    taskType: TaskType,
+};
+
+function getMessageForModalTitle(taskType: TaskType): Object {
+    switch (taskType) {
+        case TASK_TYPE_GENERAL:
```
jstoffan: Minor, but it seems like this could use an object literal instead of switch/case.

src/elements/content-sidebar/TaskModal.js
```diff
@@ -0,0 +1,64 @@
...
+    feedbackUrl: string,
+    handleCreateError: Function,
+    handleCreateSuccess: Function,
+    handleModalClose: Function,
+    isTaskFormOpen: boolean,
+    passThrough: Object,
+    taskType: TaskType,
+};
+
+function getMessageForModalTitle(taskType: TaskType): Object {
```
jstoffan: Can we be more specific with the return type and other types, as well?

src/elements/content-sidebar/TaskModal.js
booc0mtaco: why do we do this, actually? my impression was it's a stylistic thing, but is there more to it?

src/elements/content-sidebar/TaskModal.js
booc0mtaco: oh good call. I will update with some other changes in a bit and include these

src/elements/content-sidebar/TaskModal.js
booc0mtaco: there will be a little more logic here soon, so will update when that lands

PR #1258, opened 2019-06-10
refactor(tasks-new): extract TaskDueDate and TaskStatus styles

tjgupta:


boxcla:

src/elements/content-sidebar/activity-feed/task-new/TaskDueDate.scss
```diff
@@ -0,0 +1,12 @@
+@import '../../../common/variables';
+
+$error-color: $red;
```
booc0mtaco: let's rename this since this sass variable is global and this name is pretty generic. Some SUIT-able name would work

https://sass-lang.com/documentation/variables#scope

tjgupta: I took this from the previous file, but do you think there's a reason to rename it? Can we just use `$red` here?

src/elements/content-sidebar/activity-feed/task-new/TaskDueDate.scss
```diff
@@ -0,0 +1,10 @@
+@import '../../../common/variables';
+
+.bcs-TaskDueDate {
+    color: $bdl-neutral-02;
+    font-weight: bold;
+
+    &.bcs-is-taskOverdue {
+        color: $red;
```
booc0mtaco: `$bdl-watermelon-red`

PR #1260, opened 2019-06-10
feat(tasks): Add new task edit api methods

shahzadaziz:


boxcla:

src/api/Feed.js
```diff
@@ -494,6 +494,50 @@ class Feed extends Base {
...
+        this.errorCallback = errorCallback;
+        this.tasksNewAPI = new TasksNewAPI(this.options);
+        this.updateFeedItem({ isPending: true }, task.id);
+        this.tasksNewAPI.updateTask({
+            file,
+            task,
+            successCallback: (taskData: Task) => {
+                this.updateFeedItem(
+                    {
+                        ...taskData,
```
alexkrolick: does this also need to merge with the old task data (for nested values like assigned_to)?

shahzadaziz: @alexkrolick Yep, however it is outside the scope of this task. Once assignment stuff is figured out appropriate logic can be added here.

booc0mtaco:
@shahzadaziz on merge, remember to format the title and description since GH couldn't figure out what to use

`feat(tasks): Add new task edit api methods`

src/api/Feed.js
```diff
@@ -494,6 +494,50 @@ class Feed extends Base {
...
+                        isPending: false,
+                    },
+                    task.id,
+                );
+                if (!this.isDestroyed()) {
+                    successCallback();
```
jstoffan: Should we pass the task back to the callback?

src/api/Feed.js
```diff
@@ -494,6 +494,50 @@ class Feed extends Base {
...
+            task,
+            successCallback: (taskData: Task) => {
+                this.updateFeedItem(
+                    {
+                        ...taskData,
+                        isPending: false,
+                    },
+                    task.id,
+                );
+                if (!this.isDestroyed()) {
```
jstoffan: It seems like this check should occur before the `this.updateFeedItem` call?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -191,9 +191,17 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
             // need to load the pending item
             this.fetchFeedItems();
         },
-        updateTask: ({ text, id }: { id: string, text: string }): void => {
```
jstoffan: Is this still used by the "legacy" tasks? If so, does it still work properly for them? It looks like it was calling the other `updateTask` endpoint previously.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -191,9 +191,17 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
-            const { file, api } = this.props;
-            api.getFeedAPI(false).updateTask(file, id, text, this.feedSuccessCallback, this.feedErrorCallback);
+        updateTask: (task: TaskUpdatePayload): void => {
+            const { file, api, onTaskUpdate = noop } = this.props;
```
jstoffan: Default value for `onTaskUpdate` should live in the `defaultProps` static property.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -191,9 +191,17 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+            api.getFeedAPI(false).updateTaskNew(
+                file,
+                task,
+                () => {
+                    this.feedSuccessCallback();
```
jstoffan: Should this logic be moved to its own named method so it can be tested? Looks like it's missing in the unit tests right now.

src/api/Feed.js
shahzadaziz: We could, but it is not required at the moment. The form already receives latest data through props.

src/elements/content-sidebar/ActivitySidebar.js
shahzadaziz: No, this is nested under the newTaskAPI. This handler is only used if `newApi` feature is enabled

src/elements/content-sidebar/ActivitySidebar.js
shahzadaziz: I would like this be inline for now, this will be refactored in the next chunk of work.

src/api/Feed.js
shahzadaziz: I am personally conflicted about this check, looks like we do not need it and it is baked into Base class.

shahzadaziz: Removing it for now

PR #1266, opened 2019-06-12
fix(icons): updating add icon style and adding hidden eye icon

krapspark:
BREAKING CHANGE: New IconAdd with different default dimensions.

**Old Add Icon**
Screen Shot 2019-06-12 at 10 38 26 AM

**New Add Icon**
Screen Shot 2019-06-12 at 10 37 49 AM

**New Eye Hidden Icon**
Screen Shot 2019-06-13 at 4 47 36 PM


boxcla:
Verified that @krapspark has signed the CLA. Thanks for the pull request!

src/icons/general/IconAdd.js
```diff
@@ -13,17 +13,10 @@ type Props = {
...
 
 const IconAdd = ({ className = '', color = '#000000', height = 24, title, width = 24 }: Props) => (
-    <AccessibleSVG
-        className={`icon-add ${className}`}
-        height={height}
-        title={title}
-        viewBox="-603 389 16 16"
-        width={width}
-    >
+    <AccessibleSVG className={`icon-add ${className}`} height={height} title={title} viewBox="0 0 16 16" width={width}>
```
jstoffan: Good opportunity to update to SUIT? `bdl-IconAdd`.

src/icons/general/IconAdd.js
```diff
@@ -13,17 +13,10 @@ type Props = {
 };
 
 const IconAdd = ({ className = '', color = '#000000', height = 24, title, width = 24 }: Props) => (
```
priyajeet: This should also be now 32x32

priyajeet: And if you dont want it to be a breaking change, you can leave the defaults to be 24x24 in the props, but still change the viewport to 32x32

krapspark: Is the 32x32 the desired size for all icons?
Should I also make the new icon 32x32 as well?

src/icons/general/IconAdd.js
krapspark: Doesn't that make this a breaking change?

jstoffan: Yep, it does, but it needs to get done eventually. The question is if it's something we can tackle now or not.

src/icons/general/IconAdd.js
krapspark: Also, the viewport just pertains to the paint area of the icon, right? I'd need to get new assets if I wanted the viewport to be 32 x 32.

jeremypress: We've been standardizing around 32x32 for anything that looks good at that aspect ratio.

src/icons/general/IconAdd.js
jeremypress: yeah, but that should be fine because we have a lot of breaking changes going into 11.0 once we release it not as a beta.

src/icons/general/IconAdd.js
```diff
@@ -12,20 +12,19 @@ type Props = {
     width?: number,
 };
 
-const IconAdd = ({ className = '', color = '#000000', height = 24, title, width = 24 }: Props) => (
+const IconAdd = ({ className = '', color = '#222', height = 16, title, width = 16 }: Props) => (
```
jstoffan: Ideally, the default `color` value should come from our `variables` module and use one of the new `bdlNeutral` colors. Same comment for the new hidden icon.

src/icons/general/IconEyeHidden.js
```diff
@@ -0,0 +1,33 @@
...
+type Props = {
+    className?: string,
+    color?: string,
+    height?: number,
+    /** A text-only string describing the icon if it's not purely decorative for accessibility */
+    title?: string | React.Element<any>,
+    width?: number,
+};
+
+const IconEyeHidden = ({ className = '', color = '#222', height = 16, title, width = 16 }: Props) => (
```
priyajeet: color should be imported from variables, so bdlGray here

src/icons/general/IconEyeHidden.js
```diff
@@ -0,0 +1,33 @@
...
+    <AccessibleSVG
+        className={`bdl-IconEyeHidden ${className}`}
+        height={height}
+        title={title}
+        viewBox="0 0 16 16"
+        width={width}
+    >
+        <g className="fill-color" fill={color} fillRule="evenodd">
```
priyajeet: I think you can move `className="fill-color" fill={color}` to the paths below and get rid of the g and one nesting level

src/icons/general/IconEyeHidden.js
```diff
@@ -0,0 +1,32 @@
...
+type Props = {
+    className?: string,
+    color?: string,
+    height?: number,
+    /** A text-only string describing the icon if it's not purely decorative for accessibility */
+    title?: string | React.Element<any>,
+    width?: number,
+};
+
+const IconEyeHidden = ({ className = '', color = '#222', height = 32, title, width = 32 }: Props) => (
```
priyajeet: import { bdlGray } from 'box-ui-elements/es/styles/variables';
color = bdlGray

PR #1270, opened 2019-06-13
fix(comment): use overlay z-index for confirmation dialog

booc0mtaco:
Client apps which might have more deeply-nested structure will require a standard, higher z-index value. This applies one to the tethered container.


boxcla:

booc0mtaco:
@ConradJChan since the issue can happen when the component is embedded, a test for that is probably best left to the client app. in ui-elements context, that test would have passed before this change (unless it was checking for some specific z-index, which is less useful).

PR #1272, opened 2019-06-13
fix(activity-feed): Fix version ordering in Activity Feed

jeremypress:
This PR introduces a call to /files/file_id/versions/version_id using the current version to retrieve information about when the version was uploaded.

Todo
- [ ] tests


boxcla:

src/api/Feed.js
```diff
@@ -174,27 +179,28 @@ class Feed extends Base {
         }
 
         this.id = id;
+        this.versionId = file_version.id;
```
jstoffan: Would it be preferable to store `this.file`, instead, at this point?

src/api/Feed.js
```diff
@@ -228,6 +234,24 @@ class Feed extends Base {
...
+     * @return {Promise} - the file versions
+     */
+    fetchCurrentVersion(file: BoxItem): Promise<?FileVersions> {
+        this.versionsAPI = new VersionsAPI(this.options);
+
+        return new Promise(resolve => {
+            this.versionsAPI.getCurrentVersion(
+                this.id,
+                this.versionId,
+                currentVersion => resolve(this.versionsAPI.decorateCurrentVersion(currentVersion, file)),
```
jstoffan: You may be able to move this method call to a `.then` after the promise instantiation.

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -121,10 +121,19 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
...
-    handleFetchSuccess = ([fileResponse, versionsResponse]): [BoxItem, FileVersions] => {
+    handleFetchSuccess = ([fileResponse, versionsResponse, currentVersionResponse]): [
+        BoxItem,
+        FileVersions,
+        FileVersions,
+    ] => {
         const { api } = this.props;
         const versionsApi = api.getVersionsAPI(false);
-        const versionsWithCurrent = versionsApi.addCurrentVersion(versionsResponse, fileResponse);
+
```
jstoffan: Probably want to keep all the related variables grouped together?

src/api/Versions.js
```diff
@@ -242,6 +190,54 @@ class Versions extends OffsetBasedAPI {
...
+     * @param {BoxItem} file - a box file
+     * @returns {FileVersions} - a FileVersions object containing the decorated current version
+     */
+    decorateCurrentVersion(version: BoxItemVersion, file: BoxItem): FileVersions {
+        const restoredFromId = getProp(file, 'restored_from.id');
+        const isRestoredVersion = !!restoredFromId;
+
+        if (isRestoredVersion) {
+            version.action = VERSION_RESTORE_ACTION;
+            version.version_restored = version.version_number;
```
jstoffan: It looks like this is re-applying the version's number to a separate field. I think we need to look up `restored_from.id` value in the versions array and use that version number, instead. It's probably also worth adding an integration test for this behavior since it's a little fragile.

src/api/Feed.js
```diff
@@ -173,28 +173,32 @@ class Feed extends Base {
...
-            if (!this.isDestroyed()) {
-                this.setCachedItems(id, sortedFeedItems);
-                if (this.hasError) {
-                    errorCallback(sortedFeedItems);
-                } else {
-                    successCallback(sortedFeedItems);
+        Promise.all([versionsPromise, currentVersionPromise, commentsPromise, tasksPromise, appActivityPromise]).then(
+            ([versions: ?FileVersions, currentVersion: ?BoxItemVersion, ...feedItems]) => {
+                const decoratedVersion =
+                    currentVersion && versions
```
jstoffan: Can this be refined a bit to avoid the ternary?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -185,6 +194,28 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
         return new Promise((resolve, reject) => api.getVersionsAPI(false).getVersions(fileId, resolve, reject));
     };
 
+    fetchCurrentVersion = ([fileResponse, versionsResponse]): Promise<[BoxItem, FileVersions, FileVersions]> => {
```
jstoffan: Super minor, but all the `fetch` methods in here are alphabetized at the moment. Maybe `fetchVersionCurrent`? 😄

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -121,10 +121,18 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
...
+        BoxItem,
+        FileVersions,
+        FileVersions,
+    ] => {
         const { api } = this.props;
         const versionsApi = api.getVersionsAPI(false);
-        const versionsWithCurrent = versionsApi.addCurrentVersion(versionsResponse, fileResponse);
+
+        const versionsWithCurrent: FileVersions = {
+            entries: [...versionsResponse.entries, ...currentVersionResponse.entries],
```
jstoffan: Ideally, the munging and decoration logic would live in the API layer and be hidden completely behind `api.getVersionsAPI(false).getVersions()`. I'm not sure how feasible that is given the Feed API limitations, though.

src/api/Feed.js
jeremypress: yep, will revisit

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
jeremypress: agreed this is worth coming back to in a later change.

src/api/Feed.js
```diff
@@ -239,7 +239,6 @@ class Feed extends Base {
 
         return new Promise(resolve => {
             const { file_version = {} } = this.file;
-
```
jstoffan: Whitespace-only change.

src/api/Versions.js
```diff
@@ -243,7 +243,6 @@ class Versions extends OffsetBasedAPI {
             currentVersion.version_restored = restoredVersion.version_number;
         }
 
-        // Wrap the current version as a FileVersions object to make merging with the feed easier
```
jstoffan: Is this no longer applicable?

src/api/__tests__/Feed-test.js
```diff
@@ -202,24 +238,14 @@ jest.mock('../Comments', () =>
...
-            ],
-        }),
-    })),
-);
+jest.mock('../Versions', () => {
+    const version = mockFirstVersion;
+    const currentVersion = mockCurrentVersion;
+    return jest.fn().mockImplementation(() => ({
+        getVersions: jest.fn().mockReturnValue(version),
+        getCurrentVersion: jest.fn().mockReturnValue(currentVersion),
```
jstoffan: The shortcut for this should be `jest.fn(() => mockCurrentVersion)`, likewise for the above;

src/api/__tests__/Feed-test.js
```diff
@@ -202,24 +238,14 @@ jest.mock('../Comments', () =>
...
-                    modified_by: { name: 'Akon', id: 11 },
-                },
-            ],
-        }),
-    })),
-);
+jest.mock('../Versions', () => {
+    const version = mockFirstVersion;
+    const currentVersion = mockCurrentVersion;
+    return jest.fn().mockImplementation(() => ({
```
jstoffan: Shortcut:

```
jest.fn(() => ({
    // properties
}));
```

src/elements/content-sidebar/versions/__tests__/VersionsSidebarContainer-test.js
```diff
@@ -239,6 +240,31 @@ describe('elements/content-sidebar/versions/VersionsSidebarContainer', () => {
...
+    describe('fetchVersionCurrent', () => {
+        const fileVersionId = '1234';
+        test('should get the current version and add it to the versions response', () => {
+            const file = {
+                id: defaultId,
+                file_version: {
+                    id: fileVersionId,
+                },
+            };
+            const versions = 'versions';
```
jstoffan: Should this be something other than a string?

src/api/Versions.js
jeremypress: not really. Now we're merging a version into an existing list instead of arbitrarily wrapping one `BoxItemVersion` as a `FileVersions`. I updated the doc string to be more informative.

PR #1273, opened 2019-06-13
build(prepush): add check for uncommitted files before pushing

priyajeet:
So that files that may be generated during the build process are either cleaned up or committed and tested.

### DIFF
--- a/conf/prepush.sh
+++ b/conf/prepush.sh
@@ -9,6 +9,13 @@ green=$"\e[1;32m"
 blue=$"\e[1;34m"
 end=$"\e[0m\n"
 
+check_uncommitted_files() {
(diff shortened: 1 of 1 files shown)


boxcla:

conf/prepush.sh
```diff
@@ -55,6 +62,8 @@ prepush() {
     printf "${blue}-------------------------------------------------------------${end}"
     yarn build:prod:es || exit 1
 
+    check_uncommitted_files || exit 1
```
jstoffan: Should this check happen at the very beginning, instead? Or is the concern that `build:prod:es` could generate some uncommitted files?

priyajeet: yeah after build:prod:es since thats what generates the variables

PR #1275, opened 2019-06-14
fix(versions): Show server errors without hiding version history

jstoffan:


boxcla:

PR #1279, opened 2019-06-14
feat(content-explorer): Gallery View

rickstefanik:
Moved over relevant files from MDVGridView.  Have basic functionality of Gallery View in Content-Explorer.  Apologies for the large PR, mainly due to bringing files over from MDV.  I will keep PR's smaller in the future.


boxcla:

src/api/File.js
```diff
@@ -202,6 +202,65 @@ class File extends Item {
...
+            const thumbnailUrl = await this.xhr.get(xhrOptions).then(async response => {
+                const entries = response.data.representations.entries;
+
+                if (!entries.length) {
+                    return null;
+                }
+
+                // if unable to fetch jpg thumbnail, grab png of first page of file.
+                // Asset path for thumbnail is simply empty string.
+                const asset_path = entries[0].representation === 'jpg' ? '' : '1.png';
```
jeremypress: I think it's worth breaking out the png logic as a separate method, this function is getting a bit long and it's a good place to separate concerns.

src/components/pages/metadata-view-page/components/MDVGridView.js
```diff
@@ -0,0 +1,114 @@
+// @flow
```
jeremypress: `metadata-view` is also a features directory, should this go there?

src/elements/common/sub-header/SubHeaderRight.js
```diff
@@ -36,6 +48,8 @@ const SubHeaderRight = ({
 
     return (
         <div className="be-sub-header-right">
+            {isGridView ? <GridViewSlider onResize={onResize} columnCount={columnCount} /> : null}
```
jeremypress: You can avoid the `null` with `isGridView && <GridViewSlider...`

src/styles/variables.js
```diff
@@ -84,3 +84,7 @@ export const overlayZIndex = '190'; // overlay-z-index
 export const tooltipZIndex = '190'; // tooltip-z-index
 export const bdlLineHeight = '20px'; // bdl-line-height
 export const bdlBorderRadiusSize = '4px'; // bdl-border-radius-size
+export const headerHeight = '56px'; // header-height
```
jeremypress: I'm not sure if we want to add these globally, can we scope them to wherever they are used?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -377,6 +387,38 @@ class ContentExplorer extends Component<Props, State> {
...
+        if (!items) {
+            return;
+        }
+
+        const thumbnailArray = Array(items.length).fill(null);
+
+        // using for loop instead of items.forEach() to get desired await functionality
+        // TODO: This is bad. Find a better way to do this. Without using await, all requests except final
+        // one were cancelled.  Need to find a cleaner to make sure no requests get cancelled.
+        for (let i = 0; i < items.length; i += 1) {
```
jeremypress: You could use a forEach to generate all of the promises and then use `Promise.all`? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all

src/components/core/item-list/components/GridViewSlider.js
```diff
@@ -0,0 +1,57 @@
+// @flow
```
priyajeet: Should start with just this component as an isolated PR. And once this is working, replace the one within webapp with this component.

src/components/pages/metadata-view-page/components/MDVGridView.js
```diff
@@ -0,0 +1,122 @@
...
+import * as React from 'react';
```
priyajeet: We may not need this and its parent, something directly can be used from https://bvaughn.github.io/react-virtualized/#/components/Collection perhaps. Will have to try it out. If we are using Grid or any other view as an abstraction, that should be an isolated PR and likely built within the /features folder.

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -218,23 +225,26 @@ class ContentExplorer extends Component<Props, State> {
         this.id = uniqueid('bce_');
 
         this.state = {
-            sortBy,
-            sortDirection,
-            rootName: '',
+            columnCount: 5,
```
ConradJChan: `MAX_GRID_VIEW_COLUMNS`? Should the `columnCount` be defaulted to the max?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -377,6 +387,38 @@ class ContentExplorer extends Component<Props, State> {
...
     };
 
+    /**
+     * Call Box Representations API to fetch thumbnails of all files in collection.
+     * The thumbnailUrls array in state will be populated with urls to thumbnails.
+     * If a thumbnail could not be found, that item's entry in thumbnailUrls will be null.
+     *
+     * @private
+     * @param {Object} collection - item collection object
+     * @param {Boolean|void} triggerNavigationEvent - To trigger navigate event and focus grid
```
ConradJChan: doc seems out of date

src/elements/content-explorer/ContentExplorer.js
ConradJChan: I think the way it is written also fetches the thumbnails serially, whereas if you go with @jeremypress suggestion, then the images should be fetched in parallel giving a bit of a performance boost

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -377,6 +387,38 @@ class ContentExplorer extends Component<Props, State> {
...
+     * @return {void}
+     */
+    async fetchThumbnailUrls(collection: Collection, dimensions: string): Promise<void> {
```
ConradJChan: If items is the only thing being used from `collection`, we can either deconstruct that in the function declaration or we can change the input to be the actual items array

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -377,6 +387,38 @@ class ContentExplorer extends Component<Props, State> {
...
+        const { items = [] } = collection;
+        if (!items) {
```
ConradJChan: This check would always be a noop if `items` is defaulted to []

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -377,6 +387,38 @@ class ContentExplorer extends Component<Props, State> {
...
+
+        // using for loop instead of items.forEach() to get desired await functionality
+        // TODO: This is bad. Find a better way to do this. Without using await, all requests except final
+        // one were cancelled.  Need to find a cleaner to make sure no requests get cancelled.
+        for (let i = 0; i < items.length; i += 1) {
+            const item = items[i];
+            // eslint-disable-next-line no-await-in-loop
+            await this.api.getFileAPI().getFileThumbnail(item, dimensions, thumbnailUrl => {
+                thumbnailArray[i] = thumbnailUrl;
+                this.setState({ thumbnailUrls: [...thumbnailArray] });
```
ConradJChan: Is the separate `thumbnailUrls` data structure necessary? Could we the thumbnail url decorate the existing item in `currentCollection`?

src/api/File.js
```diff
@@ -202,6 +202,65 @@ class File extends Item {
...
+        const access_token = this.xhr.token;
+        if (!access_token) {
+            return;
+        }
+
+        const xhrOptions: Object = {
+            url: newUrl,
+            headers: {
+                // API will return first representation it finds, so 1024x1024 png is fallback.
+                'X-Rep-Hints': `[jpg?dimensions=${dimensions},png?dimensions=1024x1024]`,
```
ConradJChan: Seems like we're mixing the Thumbnails rep call with the Single page image rep call, do we need the png? Specifically, I see this in the docs `Therefore, the first time you request a single-page representation, you will notice the status is set to none and the metadata->pages property is not returned, which indicates that the representation is not yet generated and that you must try again after some time.`

src/api/File.js
```diff
@@ -202,6 +202,65 @@ class File extends Item {
...
+        if (this.isDestroyed()) {
+            return;
+        }
+
+        // TODO: implement cache for recently fetched thumbnails
+
+        const { id } = item;
+        const newUrl = `${this.getUrl(id)}?fields=representations`;
+
+        const access_token = this.xhr.token;
```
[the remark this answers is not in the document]
rickstefanik: I'm grabbing the token here to use as a query in the URLs of thumbnails. 
`"${thumbnailLink}?access_token=${access_token}"`
Should I not be getting the token to do this?

rickstefanik:
Jun-20-2019 13-14-59

rickstefanik:
Breaking this PR down into smaller pieces to more efficiently merge into master.

PR #1280, opened 2019-06-14
fix(scroll-if-needed): fix scroll-if-needed on selector dropdown

Sunyang730:
The auto scroll for the selector dropdown will scroll the item in to
center of the container when hover over. The correct behavior should
only scroll the item into the nearest view of the container.


boxcla:

PR #1282, opened 2019-06-17
refactor(tasks-new): switch to explicit flow types

tjgupta:
Switch to using imported flow types for new tasks components rather than using global implicit types from the flow-typed file as doing that was causing issues in consuming libraries.


boxcla:

flow-typed/box-ui-elements.js
```diff
@@ -633,125 +634,6 @@ type Tasks = {
     total_count: number,
 };
 
-/* New Task Types START */
```
priyajeet: 👍

src/elements/content-sidebar/activity-feed/task-new/flowTypes.js
```diff
@@ -0,0 +1,96 @@
...
+
+export type TaskCollabAssignee = {|
+    ...TaskCollab<'ASSIGNEE'>,
+    permissions: {|
+        can_delete: boolean,
+        can_update: boolean,
+    |},
+|};
+
+export type TaskLink = {|
```
priyajeet: I would move all exports to the bottom like we do for other files

flow-typed/box-ui-elements.js
```diff
@@ -18,6 +18,7 @@ import type { ContentSidebarProps } from '../src/elements/content-sidebar';
 import type { ContentOpenWithProps } from '../src/elements/content-open-with';
 import type { ContentPreviewProps } from '../src/elements/content-preview';
 import type { FeatureConfig } from '../src/elements/common/feature-checking';
+import type { TaskNew } from '../src/elements/content-sidebar/activity-feed/task-new/flowTypes';
```
alexkrolick: Is this path a good place to put this if we also update the `api/tasks/*` and `api/feed.js` files to reference these types?

src/elements/content-sidebar/activity-feed/task-new/TaskDueDate.js
```diff
@@ -8,6 +8,9 @@ import { ReadableTime } from '../../../../components/time';
 
 import { TASK_NEW_NOT_STARTED } from '../../../../constants';
 
+import type { TaskStatus } from './flowTypes';
+import type { ISODate } from '../../../../common/box-types';
```
priyajeet: probably rename file to flowTypes.js (or just add a new one and leave the old one for later refactor) so that it can just be common/flowTypes

src/elements/content-sidebar/activity-feed/task-new/flowTypes.js
tjgupta: Done

flow-typed/box-ui-elements.js
tjgupta: Done, moved to the `api` folder.

src/elements/content-sidebar/activity-feed/task-new/TaskDueDate.js
tjgupta: Done.

alexkrolick: We might end up with something like `flowTypes/tasks`, `/core`, `/miniObjects` etc. eventually but this is a good start - no more leaking globals

src/common/flowTypes.js
```diff
@@ -0,0 +1,28 @@
+// @flow
+import { ITEM_TYPE_FOLDER, ITEM_TYPE_FILE, ITEM_TYPE_WEBLINK } from './constants';
+
+export type inlineNoticeType = 'warning' | 'error' | 'success' | 'info';
+export type itemType = ITEM_TYPE_FOLDER | ITEM_TYPE_FILE | ITEM_TYPE_WEBLINK;
+
+export type FileMini = {
```
priyajeet: exports at the bottom like you did for other files

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -8,6 +8,8 @@ import * as React from 'react';
 import debounce from 'lodash/debounce';
 import noop from 'lodash/noop';
 import flow from 'lodash/flow';
+import type { TaskType, TaskNew } from './activity-feed/task-new/flowTypes';
```
priyajeet: can we keep type imports at the bottom

src/features/unified-share-modal/SharedLinkSection.js
```diff
@@ -17,7 +17,7 @@ import IconMail from '../../icons/general/IconMail';
 import IconClock from '../../icons/general/IconClock';
 import IconGlobe from '../../icons/general/IconGlobe';
 import { amaranth } from '../../styles/variables';
-import type { itemType as ItemType } from '../../common/box-types';
+import type { itemType as ItemType } from '../../common/flowTypes';
```
priyajeet: can just make it ItemType inside the file?

alexkrolick: +1 to uppercasing all types

src/api/Feed.js
```diff
@@ -31,6 +31,14 @@ import {
     TASK_NEW_NOT_STARTED,
     TYPED_ID_FEED_PREFIX,
 } from '../constants';
+import type {
```
jstoffan: It seems like these are our "model" types, is that correct? If so, should they instead live up here with the API flow types and be imported by the components? They could even be re-exported by the component types file, if desired.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -23,6 +23,8 @@ import AddTaskButton from './AddTaskButton';
 import SidebarUtils from './SidebarUtils';
 import { DEFAULT_COLLAB_DEBOUNCE, ORIGIN_ACTIVITY_SIDEBAR, SIDEBAR_VIEW_ACTIVITY } from '../../constants';
 import './ActivitySidebar.scss';
+import type { TaskType, TaskNew } from './activity-feed/task-new/flowTypes';
```
jstoffan: It might be nice to import these into a sidebar-specific types module, then re-export them as a single object. That way, we wouldn't need to import types from all over the place directly in our components. Thoughts?

src/common/flowTypes.js
```diff
@@ -0,0 +1,30 @@
+// @flow
+import { ITEM_TYPE_FOLDER, ITEM_TYPE_FILE, ITEM_TYPE_WEBLINK } from './constants';
+
+type InlineNoticeType = 'warning' | 'error' | 'success' | 'info';
+type ItemType = ITEM_TYPE_FOLDER | ITEM_TYPE_FILE | ITEM_TYPE_WEBLINK;
```
priyajeet: Should these be typeof ITEM_TYPE_FOLDER | typeof ITEM_TYPE_FILE | typeof ITEM_TYPE_WEBLINK?
And then in the constants file:

ITEM_TYPE_FOLDER: 'folder' = 'folder';
...

off topic: this is another syntax I learnt recently
https://flow.org/try/#0MYewdgzgLgBAygSQFoFEYF4YG8BQN-wCyAggDKkBcMAFAOQQC2AhgDYu1X3Nu0CUANHgKEUAEQQBVQlToMApgBMAlgFcGHGLXnK1fQQRiliAJQDiKGbRZMATgHM5tfpuv3HAoflEoAYsQmkACqWCnIAZkwqLFAatKERUTEeAL4A3Dg4oJCwEEoAXnJUUACeAA5yIGHwyCgAdHAk5BiaAO424Ha0qUA

src/common/flowTypes.js
```diff
@@ -0,0 +1,30 @@
...
+
+type FileMini = {
+    id: string,
+    name: string,
+    type: 'file',
```
priyajeet: If you change the constants as above, then this can be `typeof ITEM_TYPE_FILE` and the one below can be the folder counterpart

src/common/flowTypes.js
alexkrolick: That's right, we need either `typeof` or `import typeof` at the top.

There are also helpers like `$Keys<Foo>` for dealing with enums and maps

src/api/Feed.js
tjgupta: Hmm, I don't know. These types are used in multiple components and sometimes in the API files as well. Maybe so it doesn't become super complex with types all over the place and re-exporting we should go back to plan B and have something like this, with all the types related to tasks in one file independent of components and API? This approach could be used for types that span multiple components and/or API:

```
src/common/types
  core.js (what used to be src/common/box-types.js)
  tasks.js
  ... other files in the future that have types spanning multiple separate things
```

src/common/flowTypes.js
tjgupta: Done

src/api/Feed.js
jstoffan: Sounds like a good compromise.

tjgupta: Done.

src/common/types/tasks.js
```diff
@@ -0,0 +1,111 @@
...
+    permissions: {|
+        can_delete: boolean,
+        can_update: boolean,
+    |},
+    target?: ?FileMini | ?FolderMini | ?UserMini,
+    task?: TaskMini,
+    type: 'task_link',
+|};
+
+type MarkerPaginatedCollection<T> = {
```
jstoffan: Seems like this should probably still live in an even more common location, since quite a few V2 endpoints response with this format.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -23,6 +23,7 @@ import AddTaskButton from './AddTaskButton';
 import SidebarUtils from './SidebarUtils';
 import { DEFAULT_COLLAB_DEBOUNCE, ORIGIN_ACTIVITY_SIDEBAR, SIDEBAR_VIEW_ACTIVITY } from '../../constants';
 import './ActivitySidebar.scss';
+import type { TaskType, TaskNew, TaskUpdatePayload } from '../../common/types/tasks';
```
jstoffan: I think our typical convention is for the SCSS import to come last.

src/elements/content-sidebar/ActivitySidebar.js
jstoffan: This doesn't seem to be accurate based on a spot check of the components in this repo. I'm seeing something more like:

```
npm imports
npm type imports
local imports
local type imports
style imports

type defs
```

priyajeet: I guess I swapped the type and styles.

PR #1283, opened 2019-06-17
fix(uploader): MultiputUploads stop hanging in IE11

ConradJChan:
`partsUploaded` was a field that was never updated, so we entered into a race condition: (1) uploading all the chunked parts and (2) sha1Worker computing all the digests. If (2) finishes first (which is the case in probably all of the other browsers), then there is no issue. But if (1) finishes first (which is the case in IE11), committing the upload session was at the mercy of `onWorkerMessage` which was referencing the value of `partsUploaded` which was never updated and as a result, `undefined`.

The fix below was to redirect the logic when the `done` event is sent from the sha1Worker back to `processNextParts` which will commit the session

### DIFF
--- a/src/api/uploads/MultiputUpload.js
+++ b/src/api/uploads/MultiputUpload.js
@@ -82,8 +82,6 @@ class MultiputUpload extends BaseMultiput {
 
     commitSessionTimeout: TimeoutID;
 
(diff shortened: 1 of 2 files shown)


boxcla:

PR #1284, opened 2019-06-18
feat(task-new): add ability to edit task assignees

booc0mtaco:
- use `TaskCollabAssignee` for Task forms and in components
- use consistent error code for errors updating tasks
- additional tests and checks for data structures

### DIFF
--- a/flow-typed/box-ui-elements.js
+++ b/flow-typed/box-ui-elements.js
@@ -745,9 +745,11 @@ type TaskPayload = {
 };
 
 type TaskUpdatePayload = {
+    addedAssignees: SelectorItems,
     description: string,
     due_at?: ?string,
     id: string,
+    removedAssignees: Array<TaskCollabAssignee>,
 };
 
 /* New Task Types END */
--- a/src/api/Feed.js
+++ b/src/api/Feed.js
@@ -22,6 +22,7 @@ import AppActivityAPI from './AppActivity';
 import {
     ERROR_CODE_CREATE_TASK,
     ERROR_CODE_CREATE_TASK_ASSIGNMENT,
+    ERROR_CODE_UPDATE_TASK,
     HTTP_STATUS_CODE_CONFLICT,
     IS_ERROR_DISPLAYED,
     TASK_INCOMPLETE,
@@ -537,26 +538,40 @@ class Feed extends Base {
         this.errorCallback = errorCallback;
         this.tasksNewAPI = new TasksNewAPI(this.options);
         this.updateFeedItem({ isPending: true }, task.id);
-        this.tasksNewAPI.updateTask({
-            file,
-            task,
-            successCallback: (taskData: Task) => {
-                this.updateFeedItem(
-                    {
-                        ...taskData,
-                        isPending: false,
-                    },
-                    task.id,
+
+        Promise.all(task.addedAssignees.map(assignee => this.createTaskCollaborator(file, task, assignee)))
+            .then(() => {
+                return Promise.all(
+                    task.removedAssignees.map(assignee => this.deleteTaskCollaborator(file, task, assignee)),
                 );
-                if (!this.isDestroyed()) {
-                    successCallback();
-                }
-            },
-            errorCallback: (e: ElementsXhrError, code: string) => {
+            })
+            .then(() => {
+                this.tasksNewAPI.updateTask({
+                    file,
+                    task,
+                    successCallback: (taskData: Task) => {
+                        this.updateFeedItem(
+                            {
+                                ...taskData,
+                                isPending: false,
+                            },
+                            task.id,
+                        );
+
+                        if (!this.isDestroyed()) {
+                            successCallback();
+                        }
+                    },
+                    errorCallback: (e: ElementsXhrError) => {
+                        this.updateFeedItem({ isPending: false }, task.id);
+                        this.feedErrorCallback(false, e, ERROR_CODE_UPDATE_TASK);
+                    },
+                });
+            })
+            .catch((e: ElementsXhrError) => {
                 this.updateFeedItem({ isPending: false }, task.id);
-                this.feedErrorCallback(false, e, code);
-            },
-        });
+                this.feedErrorCallback(false, e, ERROR_CODE_UPDATE_TASK);
+            });
     };
 
     /**
@@ -947,12 +962,16 @@ class Feed extends Base {
      * Creates a task collaborator via the API.
      *
      * @param {BoxItem} file - The file to which the task is assigned
-     * @param {Task} task - The newly created task from the API
+     * @param {Task|TaskUpdatePayload} task - The newly created or existing task from the API
      * @param {SelectorItem} assignee - The user assigned to this task
      * @param {Function} errorCallback - Task create error callback
-     * @return {Promise<TaskAssignment}
+     * @return {Promise<TaskAssignment>}
      */
-    createTaskCollaborator(file: BoxItem, task: Task, assignee: SelectorItem): Promise<TaskCollabAssignee> {
+    createTaskCollaborator(
+        file: BoxItem,
+        task: Task | TaskUpdatePayload,
+        assignee: SelectorItem,
+    ): Promise<TaskCollabAssignee> {
         if (!file.id) {
             throw getBadItemError();
         }
@@ -966,6 +985,41 @@ class Feed extends Base {
                 file,
                 task,
                 user: assignee,
+                successCallback: resolve,
+                errorCallback: (e: ElementsXhrError) => {
+                    reject(e);
+                },
+            });
+        });
+    }
+
+    /**
+     * Deletes a task collaborator via the API.
+     *
+     * @param {BoxItem} file - The file to which the task is assigned
+     * @param {Task|TaskUpdatePayload} task - The newly deleted or existing task from the API
+     * @param {TaskCollabAssignee} assignee - The user assigned to this task
+     * @param {Function} errorCallback - Task delete error callback
+     * @return {Promise<TaskAssignment>}
+     */
+    deleteTaskCollaborator(
+        file: BoxItem,
+        task: Task | TaskUpdatePayload,
+        assignee: TaskCollabAssignee,
+    ): Promise<TaskCollabAssignee> {
+        if (!file.id) {
+            throw getBadItemError();
+        }
+
+        this.file.id = file.id;
+        return new Promise((resolve, reject) => {
+            const taskCollaboratorsAPI = new TaskCollaboratorsAPI(this.options);
+            this.taskCollaboratorsAPI.push(taskCollaboratorsAPI);
+
+            taskCollaboratorsAPI.deleteTaskCollaborator({
+                file,
+                task,
+                taskCollaborator: { id: assignee.id },
                 successCallback: resolve,
                 errorCallback: (e: ElementsXhrError) => {
                     reject(e);
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
@@ -14,7 +14,9 @@ import Keywords from '../keywords';
 import withErrorHandling from '../../withErrorHandling';
 
 type Props = {
+    approverSelectorContacts?: SelectorItems,
     currentUser?: User,
+    getApproverWithQuery?: Function,
     getAvatarUrl: GetAvatarUrlCallback,
     getMentionWithQuery?: Function,
     getUserProfileUrl?: GetProfileUrlCallback,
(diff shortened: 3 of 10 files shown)


src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
```diff
@@ -61,17 +62,36 @@ type Props = TaskFormProps & TaskFormConsumerProps & InjectIntlProvidedProps;
 type TaskFormFieldName = 'taskName' | 'taskAssignees' | 'taskDueDate';
 
 type State = {|
+    addedAssignees: SelectorItems,
```
alexkrolick: To clean up the add/remove:

- change `state.approvers` type to `Array<TaskCollabAssignee>`
- do the transform to the PillSelector format inside of render
- remove `removedAssignees`, `addedAssignees` state; add/remove from `state.approvers` directly
- do the computation for the added and removed assignees on submit by comparing to the props

src/api/Feed.js
```diff
@@ -517,26 +518,45 @@ class Feed extends Base {
...
-                this.updateFeedItem(
-                    {
-                        ...taskData,
-                        isPending: false,
-                    },
-                    task.id,
+
+        Promise.resolve()
+            .then(() => {
+                return Promise.all(
```
jstoffan: I think we may be able to avoid nesting these `Promise.all` calls. The subsequent calls can be chained off this first `Promise.all`. I think it may help to move each of these to their own methods, as well, with each accepting and resolving the parent `task` object.

```
return this.addAssignees(task)
    .then(this.removeAssignees)
    .then(this.updateTask)
    .catch(() => // catch all the things)
```

src/api/Feed.js
```diff
@@ -517,26 +518,45 @@ class Feed extends Base {
...
+                    task.addedAssignees.map(assignee => this.createTaskCollaborator(file, task, assignee)),
```
jstoffan: I assume this question has been asked, but is it possible to do this in a batch call? This scheme is going to generate a lot of round-trips to the server.

booc0mtaco: it isn't possible currently :(

src/api/Feed.js
priyajeet: also maybe more async/await syntax?

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
```diff
@@ -61,17 +62,36 @@ type Props = TaskFormProps & TaskFormConsumerProps & InjectIntlProvidedProps;
...
     dueDate?: ?Date,
     formValidityState: { [key: TaskFormFieldName]: ?{ code: string, message: string } },
     id: string,
     isLoading: boolean,
     isValid: ?boolean,
     message: string,
+    removedAssignees: Array<TaskCollabAssignee>,
 |};
 
+function getUsersFromTask(approvers: Array<TaskCollabAssignee>): SelectorItems {
```
priyajeet: This can be in its own file, something like utils.js locally

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
```diff
@@ -61,17 +62,36 @@ type Props = TaskFormProps & TaskFormConsumerProps & InjectIntlProvidedProps;
...
+    return (
+        approvers &&
+        approvers.map(taskCollab => {
```
priyajeet: approvers.map(({ target }) => {...}) will prevent multiple .s

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
```diff
@@ -61,17 +62,36 @@ type Props = TaskFormProps & TaskFormConsumerProps & InjectIntlProvidedProps;
...
     id: string,
     isLoading: boolean,
     isValid: ?boolean,
     message: string,
+    removedAssignees: Array<TaskCollabAssignee>,
 |};
 
+function getUsersFromTask(approvers: Array<TaskCollabAssignee>): SelectorItems {
+    return (
+        approvers &&
```
priyajeet: Array.isArray() is a cleaner check

flow-typed/box-ui-elements.js
```diff
@@ -745,9 +745,11 @@ type TaskPayload = {
 };
 
 type TaskUpdatePayload = {
+    addedAssignees: SelectorItems,
```
alexkrolick: Is this right? Thought it would be `Array<TaskCollabAssignee>`

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
```diff
@@ -190,14 +215,21 @@ class TaskForm extends React.Component<Props, State> {
...
+        // TODO: remove duplicates in added list
+        this.setState({ approvers: this.state.approvers.concat(pills), addedAssignees: pills });
         this.validateForm('taskAssignees');
     };
 
     handleApproverSelectorRemove = (option: any, index: number): void => {
         const approvers = [...this.state.approvers];
-        approvers.splice(index, 1);
-        this.setState({ approvers });
+        const fullApprovers = this.props.approvers;
```
priyajeet: We should be using destructure assignments whenever possible, even if it means destructing 1. Helps in diffing if in the future more things need to get added. Similar to trialing , logic. And prevents multiple this.state.*

const { approvers, removedAssignees } = this.state;
const { approvers: fullApprovers } = this.props;
const approversClone/copy = [...approvers];

src/api/Feed.js
alexkrolick: If anything, the optimization we can consider is `_.chunk(added, 5)` to reduce the number of concurrent requests

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
```diff
@@ -190,14 +215,21 @@ class TaskForm extends React.Component<Props, State> {
...
+        const updatedRemovedAssignees = this.state.removedAssignees;
+        // TODO: remove duplicates in removed list
+        // look up the approver instance in the prop
+        const deleteUser = approvers.splice(index, 1)[0];
+        const foundApprover = fullApprovers.find(approver => approver.target.id === deleteUser.id);
+        if (foundApprover) updatedRemovedAssignees.push(foundApprover);
```
priyajeet: if () {
...
}

easier to mark breakpoints

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
```diff
@@ -149,19 +171,22 @@ class TaskForm extends React.Component<Props, State> {
 
     handleValidSubmit = (): void => {
         const { id, createTask, editTask, editMode, taskType } = this.props;
-        const { message, approvers, dueDate, isValid } = this.state;
+        const { message, approvers, addedAssignees, removedAssignees, dueDate, isValid } = this.state;
         const dueDateString = dueDate && dueDate.toISOString();
```
priyajeet: btw `dueDateString` is not really a string if `dueDate` is not a date object.

src/api/Feed.js
```diff
@@ -954,6 +978,41 @@ class Feed extends Base {
...
+        file: BoxItem,
+        task: Task | TaskUpdatePayload,
+        assignee: TaskCollabAssignee,
+    ): Promise<TaskCollabAssignee> {
+        if (!file.id) {
+            throw getBadItemError();
+        }
+
+        this.id = file.id;
+        return new Promise((resolve, reject) => {
```
priyajeet: async/await

flow-typed/box-ui-elements.js
booc0mtaco: other parts of the API code expect `SelectorItems` when adding assignees to existing tasks and creating new tasks. Leaving this be for now to de-risk changes to those existing flows

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
booc0mtaco: a lot of this got factored out so resolving the comments in this block

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
booc0mtaco: This check is actually redundant now. the block was in a higher-up component originally. now that it acts on the state of `TaskForm`, which guarantees a value exists during usage, we can remove it entirely.

src/api/Feed.js
booc0mtaco: I could remove the outer `Promise.resolve` which was suggested for ease of reading. I tried a few tweaks here, and this format avoids exposing methods that don't have a use outside this specific call and keeps the order of operations we need to enforce

src/api/Feed.js
alexkrolick: API methods like this do not return a Promise, this wraps the callbacks to enable Promise chaining or async/await elsewhere in the component

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
booc0mtaco: We have a future optimization to possibly align the types across components to avoid this conversion, but put that on the back burner for now. I'm a fan of component util.s in general, tho. easier to unit test

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
booc0mtaco: fair point. I've wondered if we should get more into the habit of typing `const`s like this anyhow...

src/elements/content-sidebar/activity-feed/task-new/__tests__/__snapshots__/Task-test.js.snap
```diff
@@ -114,27 +114,43 @@ exports[`elements/content-sidebar/ActivityFeed/task-new/Task should correctly re
...
+            "id": "other-user-assignment-id",
+            "permissions": Object {
+              "can_delete": true,
+              "can_update": true,
+            },
+            "role": "ASSIGNEE",
+            "status": "COMPLETED",
+            "target": Object {
+              "id": "3",
+              "name": "Patrick Paul",
```
jstoffan: We should probably avoid names of Boxers in our tests.

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
```diff
@@ -148,26 +163,42 @@ class TaskForm extends React.Component<Props, State> {
...
+            const approverIds = approvers.map(approver => approver.id);
+            const currentApproverIds = currentApprovers.map(currentApprover => currentApprover.id);
+
             editTask(
                 {
                     id,
                     description: message,
                     due_at: dueDateString,
+                    addedAssignees: convertAssigneesToSelectorItems(
+                        currentApprovers.filter(currentApprover => approverIds.indexOf(currentApprover.id) === -1),
```
jstoffan: We can probably make good use of [helper utils](https://lodash.com/docs/4.17.11#intersection) for situations like this one.

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
```diff
@@ -70,8 +71,22 @@ type State = {|
     message: string,
 |};
 
+function convertAssigneesToSelectorItems(approvers: Array<TaskCollabAssignee>): SelectorItems {
+    return approvers.map(({ target }) => {
+        const newSelectorItem: SelectorItem = {
```
jstoffan: Consider using an implicit return, since the variable is returned immediately.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -14,7 +14,9 @@ import Keywords from '../keywords';
 import withErrorHandling from '../../withErrorHandling';
 
 type Props = {
+    approverSelectorContacts?: SelectorItems,
     currentUser?: User,
+    getApproverWithQuery?: Function,
```
jstoffan: Can the type be stronger on these callbacks? It looks like some others are.

src/elements/content-sidebar/activity-feed/task-new/__tests__/__snapshots__/Task-test.js.snap
booc0mtaco: Probably! there are definitely some instances of names in the test data. here, I changed the snapshot which displays a bit more info than before

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
booc0mtaco: we may drop this function once we determine if it's safe to use `Array<TaskCollabAssignee>` throughout the feature. If we find a continued need, we can factor this more tersely

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
booc0mtaco: good call. will add to a post-launch tracker (once we verify we have the desired behavior implemented)

PR #1285, opened 2019-06-18
feat(versions): Add resin tracking to all version history actions

jstoffan:
I also took the opportunity to update to the new color palette.


boxcla:

src/elements/content-sidebar/versions/VersionsItemActions.js
```diff
@@ -54,7 +59,7 @@ const VersionsItemActions = ({
     const { can_delete, can_download, can_preview, can_upload } = permissions;
     const showDelete = can_delete && !isDeleted && !isCurrent;
     const showDownload = can_download && !isDeleted && isDownloadable;
-    const showPreview = can_preview && !isDeleted;
+    const showPreview = can_preview && !isDeleted && !isSelected;
```
jeremypress: Previewing an unselected version seems to work in my testing, why are we removing it?

src/elements/content-sidebar/versions/VersionsItemButton.js
```diff
@@ -32,51 +39,43 @@ class VersionsItemButton extends React.Component<Props> {
...
-                    }
-                }}
-                onKeyPress={event => {
-                    if (isActivateKey(event)) {
-                        if (event.key === KEYS.space) {
-                            event.preventDefault(); // Prevent scroll on space key press
-                        }
-
-                        onActivate(event);
+                    if (isDisabled) {
```
ConradJChan: Could we set the disabled property of the button to avoid manually preventing the click event?

src/elements/content-sidebar/versions/VersionsItemActions.js
jstoffan: It was a requested change.

src/elements/content-sidebar/versions/VersionsItemButton.js
jstoffan: This is how `PlainButton` handles disabled behavior, so I'm just aping it here. I don't want to use `PlainButton` because it messes with the styles really badly in this case.

src/elements/content-sidebar/versions/VersionsItemActions.js
ConradJChan: Isn't this simply removing the option to preview an already selected version in the menu?

jeremypress: yep, totally misread this. This is valid!

PR #1287, opened 2019-06-18
feat(tasks-new): add assignee list

shikharmohan:
Replacing avatar group with the new assignee list. 

image

### DIFF
--- a/i18n/en-US.properties
+++ b/i18n/en-US.properties
@@ -390,6 +390,10 @@ be.taskEditMenuItem = Modify task
 be.taskReject = Decline
 # Error message when rejecting a task fails
 be.taskRejectErrorMessage = An error has occurred while rejecting this task. Please refresh the page and try again.
+# Button name to hide task assignee list
+be.taskShowLessAssignees = Show Less
+# Button name to expand task assignee list, additionalAssigneeCount is the number of additional task assignees that can be shown.
+be.taskShowMoreAssignees = Show {additionalAssigneeCount} More
 # label for button that opens task popup
 be.tasks.addTask = Add Task
 # label for menu item that opens approval task popup
--- a/src/elements/content-sidebar/activity-feed/task-new/AssigneeList.js
+++ b/src/elements/content-sidebar/activity-feed/task-new/AssigneeList.js
@@ -1,0 +1,162 @@
+// @flow strict
+import * as React from 'react';
+import { injectIntl, FormattedMessage } from 'react-intl';
+import type { InjectIntlProvidedProps } from 'react-intl';
+import uniqueId from 'lodash/uniqueId';
+import PlainButton from '../../../../components/plain-button';
+import ReadableTime from '../../../../components/time/ReadableTime';
+import messages from '../../../common/messages';
+import AvatarGroupAvatar from './AvatarGroupAvatar';
+import { TASK_NEW_APPROVED, TASK_NEW_REJECTED, TASK_NEW_COMPLETED, TASK_NEW_NOT_STARTED } from '../../../../constants';
+
+import './AssigneeList.scss';
+
+const MIN_ASSIGNEES_SHOWN = 3;
+const TASKS_PAGE_SIZE = 20; // service does not return the page size to the client at the moment
+
+type Props = {|
+    getAvatarUrl: GetAvatarUrlCallback,
+    minAssignees: number,
+    onExpand: Function,
+    users: TaskAssigneeCollection,
+|} & InjectIntlProvidedProps;
+
+type State = {
+    isCollapsed: boolean,
+};
+
+const statusMessages = {
+    [TASK_NEW_APPROVED]: messages.tasksFeedStatusApproved,
+    [TASK_NEW_REJECTED]: messages.tasksFeedStatusRejected,
+    [TASK_NEW_COMPLETED]: messages.tasksFeedStatusCompleted,
+    [TASK_NEW_NOT_STARTED]: null,
+};
+
+const Datestamp = ({ date }: { date: ISODate | Date }) => {
+    return <ReadableTime timestamp={new Date(date).getTime()} alwaysShowTime relativeThreshold={0} />;
+};
+
+const AvatarDetails = React.memo(({ user, status, completedAt, className }) => {
+    const statusMessage = statusMessages[status] || null;
+    return (
+        <div className={className}>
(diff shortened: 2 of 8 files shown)


boxcla:

i18n/en-US.properties
```diff
@@ -388,6 +388,10 @@ be.taskEditMenuItem = Modify task
 be.taskReject = Decline
 # Error message when rejecting a task fails
 be.taskRejectErrorMessage = An error has occurred while rejecting this task. Please refresh the page and try again.
+# Button name to hide task assignee list
+be.taskShowLessAssignees = Show Less
```
jeremypress: Show Fewer?

src/elements/content-sidebar/activity-feed/task-new/AssigneeList.js
```diff
@@ -0,0 +1,162 @@
...
+import { TASK_NEW_APPROVED, TASK_NEW_REJECTED, TASK_NEW_COMPLETED, TASK_NEW_NOT_STARTED } from '../../../../constants';
+
+import './AssigneeList.scss';
+
+const MIN_ASSIGNEES_SHOWN = 3;
+const TASKS_PAGE_SIZE = 20; // service does not return the page size to the client at the moment
+
+type Props = {|
+    getAvatarUrl: GetAvatarUrlCallback,
+    minAssignees: number,
```
jeremypress: something like initialAssignee count might help enforce the point across that this is expandable.

src/elements/content-sidebar/activity-feed/task-new/AssigneeList.js
```diff
@@ -0,0 +1,162 @@
...
+
+    listTitleId = uniqueId('assignee-list-title-');
+
+    showLessAssignees = () => {
+        this.setState({
+            isCollapsed: true,
+        });
+    };
+
+    showMoreAssignees = async () => {
```
jeremypress: Would using a promise allow you to skip the empty catch case?

src/elements/content-sidebar/activity-feed/task-new/AssigneeList.js
```diff
@@ -0,0 +1,162 @@
...
+    };
+
+    render() {
+        const { minAssignees, users, getAvatarUrl } = this.props;
+        const { isCollapsed } = this.state;
+        const { entries = [], next_marker } = users;
+        const entryCount = entries.length;
+        const hiddenAssigneeCount = Math.max(0, entryCount - minAssignees);
+        const numVisibleAssignees = isCollapsed ? minAssignees : entryCount;
+        const visibleUsers = entries
```
jeremypress: Anything with significant logic like this should be its own component imo.

src/elements/content-sidebar/activity-feed/task-new/AssigneeList.js
```diff
@@ -0,0 +1,162 @@
...
+                            className="bcs-AssigneeList-listItemDetails"
+                            user={target}
+                            status={status}
+                            completedAt={completedAt}
+                        />
+                    </li>
+                );
+            });
+
+        const maxAdditionalAssignees = TASKS_PAGE_SIZE - minAssignees;
```
jeremypress: Can the API not return more than 20?

src/elements/content-sidebar/activity-feed/task-new/AssigneeList.js
```diff
@@ -0,0 +1,162 @@
...
+        const hasMoreAssigneesThanPageSize = hiddenAssigneeCount > maxAdditionalAssignees || next_marker;
+        const additionalAssigneeCount = hasMoreAssigneesThanPageSize
+            ? `${maxAdditionalAssignees}+`
```
jeremypress: is the + needed because we know we have at least `maxAdditionalAssignees` to show but we might have more?

src/elements/content-sidebar/activity-feed/task-new/AssigneeList.js
```diff
@@ -0,0 +1,162 @@
...
+    listTitleId = uniqueId('assignee-list-title-');
+
+    showLessAssignees = () => {
+        this.setState({
+            isCollapsed: true,
+        });
+    };
+
+    showMoreAssignees = async () => {
+        const { users } = this.props;
```
ConradJChan: Could deconstruct `onExpand` here as well

src/elements/content-sidebar/activity-feed/task-new/AssigneeList.js
alexkrolick: yes

i18n/en-US.properties
alexkrolick: As-designed

src/elements/content-sidebar/activity-feed/task-new/AssigneeList.js
alexkrolick: The initial page returns 20 max (should also return `limit` so we can calculate dynamically once a bug is fixed)

src/elements/content-sidebar/activity-feed/task-new/AssigneeList.js
shikharmohan: If I use a promise it would look something like this? 

```
onExpand().then(setState).catch(setState);
```

not sure if there's a better way to skip the empty catch case w/o repeating the set state call

src/elements/content-sidebar/activity-feed/task-new/AssigneeList.js
alexkrolick: `minAssignees` isn't quite accurate since you can have less than 3 initially

src/elements/content-sidebar/activity-feed/task-new/__tests__/AssigneeList-test.js
```diff
@@ -0,0 +1,187 @@
...
+                    minAssignees={max}
+                    getAvatarUrl={mockGetAvatarUrl}
+                />,
+            );
+
+            const expandBtn = global.queryAllByTestId(wrapper, 'show-more-assignees').first();
+            expandBtn.simulate('click');
+
+            let assigneeList = global.queryAllByTestId(wrapper, 'task-assignee-list');
+            expect(assigneeList.children()).toHaveLength(3);
```
alexkrolick: could you add the testid to the task-assignee items themselves so you don't have to check `.children()`, which could have some other elements in there?

PR #1291, opened 2019-06-18
fix(comment): remove redundant aria-label

tjgupta:
The `aria-label` attributes are not needed since there is text already there that the screen reader can read.


boxcla:

src/elements/content-sidebar/activity-feed/comment/CommentMenu.js
```diff
@@ -26,7 +26,7 @@ type Props = {
 } & InjectIntlProvidedProps;
 
 const CommentMenu = (props: Props) => {
-    const { intl, isDisabled, onDeleteClick, onEditClick, permissions = {}, type } = props;
```
priyajeet: injectIntl at the bottom not needed

src/elements/content-sidebar/activity-feed/comment/Comment.js
```diff
@@ -51,12 +51,12 @@ type Props = {
     tagged_message: string,
     translatedTaggedMessage?: string,
     translations?: Translations,
-    type?: typeof COMMENT_TYPE_DEFAULT | typeof COMMENT_TYPE_TASK,
+    type: typeof COMMENT_TYPE_DEFAULT | typeof COMMENT_TYPE_TASK,
```
tjgupta: Apparently when we do ` & InjectIntlProvidedProps` in a flow type (like we had in `CommentMenu`) it resolves to `any` and stops properly flow checking, which meant that this was broken before but was passing anyway. Once I removed those props this and `isConfirming?` became flow errors.

src/elements/content-sidebar/activity-feed/comment/__tests__/Comment-test.js
```diff
@@ -173,7 +173,7 @@ describe('elements/content-sidebar/ActivityFeed/comment/Comment', () => {
 
         const instance = wrapper.instance();
 
-        expect(wrapper.find('CommentMenu').length).toEqual(2);
+        expect(wrapper.find('CommentMenu').length).toEqual(1);
```
tjgupta: I'm not sure why, but when we were doing `injectIntl()` it was creating two `CommentMenu` components in these tests. Now it's back to one.
Screen Shot 2019-06-18 at 10 29 13 PM

src/elements/content-sidebar/activity-feed/comment/CommentMenu.js
tjgupta: Done

src/elements/content-sidebar/activity-feed/comment/Comment.js
priyajeet: Not sure, but it could be because it wasn't imported on top?
import type { InjectIntlProvidedProps } from 'react-intl';

I notice if you add that, it seems to show correctly, so it is a bug that we are just doing any

src/elements/content-sidebar/activity-feed/comment/Comment.js
```diff
@@ -51,12 +51,12 @@ type Props = {
...
     userHeadlineRenderer?: React.Node => React.Element<typeof FormattedMessage>,
 };
 
 type State = {
-    isConfirming?: boolean,
+    isConfirming: boolean,
```
priyajeet: I am guessing the other two also dont need ? since its set in state

src/elements/content-sidebar/activity-feed/comment/Comment.js
tjgupta: Yep, I think something with the way we're doing the flow-typed lib defs in box-ui-elements.js is breaking those types in certain cases.

src/elements/content-sidebar/activity-feed/comment/Comment.js
tjgupta: Fixed, can you approve again please @priyajeet?

src/elements/content-sidebar/activity-feed/comment/Comment.js
alexkrolick: Yeah, we should import the types explicitly

PR #1292, opened 2019-06-18
build: disable husky for release build

priyajeet:
- Upgrades husky to the latest version
- Adds back pre-push hook except for when running a release
- Disables husky hooks for CI also
- Reverts back to checking uncommitted files during pre-push


boxcla:

PR #1293, opened 2019-06-18
feat(sidebar): Add support for toggling the sidebar externally

jstoffan:
This change moves the `isOpen` state management of the sidebar to the router, which allows external applications to toggle it open/closed as needed via a controlled `history` context. 

One alternative would be to expose `isOpen` as a top-level prop on ContentSidebar. However, it would end up being treated as a partially-controlled prop and would need to be managed/checked against internal state. This router solution seems preferable to me, as it is either controlled or not, but I'm open to feedback.

Todo
- [x] More unit tests for Sidebar


boxcla:

src/elements/common/nav-button/NavButton.js
```diff
@@ -55,7 +55,7 @@ const NavButton = React.forwardRef<Props, React.Ref<any>>((props: Props, ref: Re
 
                             if (!event.defaultPrevented && isLeftClick(event)) {
                                 const method = replace ? history.replace : history.push;
-                                method(path);
+                                method(to);
```
ConradJChan: Should this still be `path`?

src/elements/content-sidebar/Sidebar.js
```diff
@@ -136,6 +112,18 @@ class Sidebar extends React.Component<Props, State> {
...
+    /**
+     * Getter for location state properties
+     *
+     * @param key - Optionally get a specific key value from state
+     * @returns {any} - The location state or state key value
+     */
+    getLocationState(key?: string): any {
+        const { location } = this.props;
+        const { state: locationState = {} } = location;
+        return key ? getProp(locationState, key) : locationState;
```
ConradJChan: This method seems a bit overloaded imo, does it return locationState properties or the location state itself? Maybe just keep the scope to properties and return null in this ternary?

ConradJChan: And do we need `getProp` here? Seems like the user defined properties you are adding to location state are only `silent` and `open`

src/elements/common/nav-button/NavButton.js
jstoffan: Nope. I want to pass the entire location object, whether it's just a path or includes something like `state`.

src/elements/content-sidebar/Sidebar.js
jstoffan: It accepts a single optional argument as a getter, so I'm not sure I agree it's too overloaded. I'll try to simplify it, though, since I'm not using the non-key aspect right now.

src/elements/content-sidebar/Sidebar.js
```diff
@@ -166,6 +158,17 @@ class Sidebar extends React.Component<Props, State> {
         return this.isForced() !== null;
     }
 
+    /**
+     * Helper to set the local store open state based on the location open state, if defined
+     */
+    setForcedByLocation(): void {
+        const isLocationOpen = this.getLocationState('open');
+
+        if (isLocationOpen !== undefined) {
```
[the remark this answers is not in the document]
jstoffan: I'm expecting a nullable boolean. Will push up a flow annotation for it.

PR #1294, opened 2019-06-18
feat(content-explorer): import files from MDVGridView

rickstefanik:


boxcla:

src/components/core/pure-component/hocs.js
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
```
ConradJChan: Can you port over the associated `hocs-test.js` as well?

src/components/core/table/components/BaseTable.js
```diff
@@ -0,0 +1,99 @@
+/* @flow */
+import * as React from 'react';
+import classNames from 'classnames';
+import Table from 'react-virtualized/dist/es/Table';
+import 'react-virtualized/styles.css';
```
ConradJChan: Minor: maybe group the style imports together?

src/components/core/table/styles/Table.scss
```diff
@@ -0,0 +1,138 @@
+@import '~box-ui-elements/es/styles/variables';
```
ConradJChan: We can probably update these to be relative references to our UI elements files

src/components/in-app-messenger/_variables.scss
```diff
@@ -0,0 +1,16 @@
+$banner-height: 40px;
```
ConradJChan: Is `in-app-messenger` relevant to ui elements? Or is it just the styles that we need?

src/components/core/table/components/BaseTable.js
```diff
@@ -0,0 +1,99 @@
+/* @flow */
```
ConradJChan: We don't currently have a `core` directory under components in elements, can this live under just `components`? Also how does this table component differ from the existing `src/components/table` component? Maybe this can be `VirtualizedTable`?

src/components/pages/metadata-view-page/components/MDVGridView.js
```diff
@@ -0,0 +1,121 @@
+// @flow
```
ConradJChan: Similarly, could we do without the `pages/metadata-view-page/components` path? Maybe put it at `src/components/grid-view`?

src/components/pages/metadata-view-page/components/MDVGridView.js
```diff
@@ -0,0 +1,121 @@
...
+        return index;
+    };
+
+    render() {
+        const { columnCount, count, height, width } = this.props;
+        const rowCount = Math.ceil(count / columnCount);
+
+        return (
+            <BaseTable
+                className={`MDVGridView MDVGridView--columns-${columnCount}`}
```
ConradJChan: Should we remove the Metadata references so that it can be more generic? Also this classname can be updated according our SUIT conventions, something like `bdl-GridView`

src/components/pages/metadata-view-page/components/MDVGridViewSlot.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+import * as React from 'react';
+
+type Props = {
+    slotIndex: number,
+    slotRenderer: ?(slotIndex: number) => React.Element<any>,
+};
+
+function MDVGridViewSlot({ slotIndex, slotRenderer }: Props) {
```
ConradJChan: Also rename this one to remove MD references

src/components/pure-component/hocs.js
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
+
+function pure(func) {
```
priyajeet: Whats the use of this?

src/styles/_sizes.media.scss
```diff
@@ -0,0 +1,48 @@
+$small-width: 480px;
```
priyajeet: I would probably move this along with where its being used, since it will likely be used in the grid only and not globally

src/components/grid-view/GridViewSlot.scss
```diff
@@ -0,0 +1,15 @@
+@import '../../elements/common/variables';
+@import '../../styles/variables';
+
+.bdl-GridViewSlot {
+    border: 1px solid $efive;
```
priyajeet: Should switch to the closest match from here https://opensource.box.com/box-ui-elements/#/Colors (inside the imported variables)

src/components/base-table/BaseTable.js
```diff
@@ -0,0 +1,97 @@
...
+            let rowClass = 'table-row';
+            if (isStriped && index % 2 === 0) {
+                // can't target odd elements in css because row elements can disappear
+                rowClass += ' has-stripe';
+            }
+            return rowClass;
+        };
+
+        return (
+            <Table
```
priyajeet: Would a Grid be better than a Table?
https://bvaughn.github.io/react-virtualized/#/components/Grid

Since we are implementing a grid view

priyajeet: Actually now they have this layout too https://bvaughn.github.io/react-virtualized/#/components/Masonry
Pinterest style, maybe we can use this but with fixed cards. It also seems responsive, so you may not need all the media scss stuff

priyajeet: Here is a striped down/hacked version from their examples: https://codesandbox.io/s/distracted-breeze-o8lxy

src/components/pure-component/hocs.js
rickstefanik: Looks like a helper function used to convert a functional component into a class component that extends React.PureComponent.  This is one of the files I ported over from MDVGridView.  I'm only using it once in BaseTable.js, so making BaseTable a pure component itself probably makes more sense.

src/components/base-table/BaseTable.js
rickstefanik: It seems that both Collection and Masonry are better suited for variable sized cells, and don't have the concept of rows.  Reading further, Table actually imports and uses Grid internally.  Grid is designed to be scrolled both vertically and horizontally.  Table is more optimized than Grid, but is not designed to scroll horizontally.  The GridView for Content Explorer shouldn't need to scroll horizontally, so Table seems to be the best option.

priyajeet: But table has fixed columns and headers. Whereas a grid view has no concept for that and the columns should flow to the next row when viewport size changes. Our cells will also be variable sized depending upon if it's a file, folder or weblink

src/components/pure-component/hocs.js
priyajeet: Its a functional component, so probably should leave it as such.

src/components/base-table/BaseTable.js
priyajeet: BTW if you see that example link above, it only scrolls vertically and moves cells to a new row if viewport reduces

src/components/in-app-messenger/_variables.scss
rickstefanik: I moved this file to the base-table folder where it's being used, so there's no longer an in-app-messenger folder.  Should I merge these styles into another file in the base-table folder, or keep them in the separate file they're already in?

src/components/core/table/components/BaseTable.js
rickstefanik: File has been moved to `src/components/base-table/`.

Right now `src/components/table` is just an HTML table, not a React Virtualized table.  `BaseTable` is just acting as a wrapper to create a pure component of a React Virtualized table with logic for striping (if that's ever a desired feature).  I could put this `BaseTable` code directly in `GridView`, but that may lower reusability if a different element wants to use the same logic as `BaseTable`.

src/components/in-app-messenger/_variables.scss
rickstefanik: Found these styles were never actually used in our code.  Removed the file.

PR #1295, opened 2019-06-19
fix(tasks_new): add missing resin component

shikharmohan:
Events within the add/edit modal for tasks fail to send because of missing component value.


boxcla:

PR #1296, opened 2019-06-19
chore(versions): Add sidebar tests for restored versions

jeremypress:


boxcla:

test/fixtures/content-sidebar/restored-file.json
```diff
@@ -0,0 +1,102 @@
+{
```
jstoffan: Since this is a read operation, can we avoid mocking out the response? Or is there a concern that the file will accidentally get modified?

jeremypress: Since we're using the "versioned" file I think it's reasonable for someone to add a new version at some point. Since this logic only works for when current versions are restored it would break the tests.

PR #1299, opened 2019-06-19
fix(left-sidebar): add key to drop wrapper for links

nickgriffin1:
Console cleanup for the React unique key console warning. This warning is triggered because the top level component in the list can be either LeftSidebarDropWrapper or LeftSidebarLink so the key property is necessary on each.


boxcla:

booc0mtaco:
@priyajeet hmm, surprised to see the build error on this PR. I thought the config was in review?

PR #1301, opened 2019-06-19
fix(tasks-new): Reset selector contacts on modal close

shahzadaziz:


boxcla:

src/elements/content-sidebar/activity-feed/task-new/Task.js
```diff
@@ -96,7 +97,12 @@ class Task extends React.Component<Props, State> {
     };
 
     handleModalClose = () => {
-        this.setState({ isEditing: false });
+        const { onModalClose } = this.props;
+        this.setState({ isEditing: false, modalError: null });
+
+        if (onModalClose) {
```
jstoffan: Is a test necessary?

src/elements/content-sidebar/activity-feed/task-new/Task.js
```diff
@@ -96,7 +97,12 @@ class Task extends React.Component<Props, State> {
     };
 
     handleModalClose = () => {
-        this.setState({ isEditing: false });
+        const { onModalClose } = this.props;
+        this.setState({ isEditing: false, modalError: null });
```
jstoffan: The initial state for `modalError` is `undefined`. Does `null` here mean something different? If not, we should try to stay consistent and avoid `null`. Also, should the `modalError` be cleared in any other circumstances that may have been previously missed?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
```diff
@@ -31,6 +31,7 @@ type Props = {
     onTaskAssignmentUpdate?: Function,
     onTaskCreate?: Function,
     onTaskDelete?: Function,
+    onTaskModalClose?: Function,
```
priyajeet: Fine for now, but for new entries should start adding explicit flow types so () => void kind

alexkrolick: () => any is equivalent to Function, right?

src/elements/content-sidebar/activity-feed/task-new/Task.js
shahzadaziz: Added test

PR #1302, opened 2019-06-19
fix(tasks-new): task actions are not updating

alexkrolick:


boxcla:

src/elements/content-sidebar/activity-feed/task-new/AssigneeList.js
```diff
@@ -50,47 +48,20 @@ const AvatarDetails = React.memo(({ user, status, completedAt, className }) => {
...
-        isCollapsed: true,
-    };
-
+class AssigneeList extends React.Component<Props> {
     static defaultProps = {
         initialAssigneeCount: DEFAULT_ASSIGNEES_SHOWN,
         users: {},
     };
 
     listTitleId = uniqueId('assignee-list-title-');
```
jstoffan: Is this still being used? If not, this component can be converted to a stateless function component.

src/elements/content-sidebar/activity-feed/task-new/Task.js
```diff
@@ -81,6 +82,34 @@ class Task extends React.Component<Props, State> {
...
         isLoading: false,
+        isAssigneeListOpen: false,
+    };
+
+    componentDidUpdate(prevProps: Props) {
+        // reset state to props assigned_to
+        const { assigned_to } = this.props;
+        const { assigned_to: prevAssignedTo } = prevProps;
+        if (prevAssignedTo !== assigned_to) {
+            this.setState({
```
jstoffan: Is it necessary to copy the prop value into state? If so, should we be using `getDerivedStateFromProps`?

src/elements/content-sidebar/activity-feed/task-new/Task.js
```diff
@@ -81,6 +82,34 @@ class Task extends React.Component<Props, State> {
...
+                assigned_to,
+            });
+        }
+    }
+
+    handleShowMoreAssignees = () => {
+        const { assigned_to } = this.state;
+
+        if (assigned_to.next_marker) {
+            this.fetchTaskCollaborators().then(() => {
```
jstoffan: Is a `.catch` handler needed?

src/elements/content-sidebar/activity-feed/task-new/AssigneeList.js
alexkrolick: it is

src/elements/content-sidebar/activity-feed/task-new/Task.js
alexkrolick: added a noop for the rejected callback - the error state is handled in the feed api class

src/elements/content-sidebar/activity-feed/task-new/AssigneeList.js
jstoffan: I see it being used in `arial-labelledby={this.listTitleId}`, which has a typo, but I don't see any corresponding label element. Can you link to the line where the label is declared?

src/elements/content-sidebar/activity-feed/task-new/__tests__/Task-test.js
```diff
@@ -285,4 +280,49 @@ describe('elements/content-sidebar/ActivityFeed/task-new/Task', () => {
...
+        expect(assigneeList).toHaveLength(INITIAL_DISPLAY_COUNT);
+
+        const expandBtn = global.queryAllByTestId(wrapper, 'show-more-assignees').first();
+        expandBtn.simulate('click');
+
+        assigneeList = global.queryAllByTestId(wrapper, 'assignee-list-item');
+        expect(assigneeList).toHaveLength(COUNT);
+    });
+
+    test('should hide assignee list when hide button is clicked', () => {});
```
shikharmohan: Missing test.

src/elements/content-sidebar/activity-feed/task-new/AssigneeList.js
```diff
@@ -50,47 +48,20 @@ const AvatarDetails = React.memo(({ user, status, completedAt, className }) => {
...
-
     render() {
-        const { initialAssigneeCount, users, getAvatarUrl } = this.props;
-        const { isCollapsed } = this.state;
+        const { initialAssigneeCount, users, getAvatarUrl, isOpen } = this.props;
         const { entries = [], next_marker } = users;
         const entryCount = entries.length;
         const hiddenAssigneeCount = Math.max(0, entryCount - initialAssigneeCount);
-        const numVisibleAssignees = isCollapsed ? initialAssigneeCount : entryCount;
+        const numVisibleAssignees = !isOpen ? initialAssigneeCount : entryCount;
```
jstoffan: Can we invert the ternary so the first condition is the positive case?

src/elements/content-sidebar/activity-feed/task-new/AssigneeList.js
```diff
@@ -127,12 +98,12 @@ class AssigneeList extends React.Component<Props, State> {
...
                     {visibleUsers}
                 </ul>
-                {isCollapsed && hiddenAssigneeCount > 0 && (
+                {!isOpen && hiddenAssigneeCount > 0 && (
                     <span>
                         <PlainButton
                             data-resin-target="showmorebtn"
                             data-testid="show-more-assignees"
-                            onClick={this.showMoreAssignees}
+                            onClick={this.props.onExpand}
```
jstoffan: Should these be destructured in render?

src/elements/content-sidebar/activity-feed/task-new/AssigneeList.js
alexkrolick: nvm!

src/elements/content-sidebar/activity-feed/task-new/Task.js
```diff
@@ -107,49 +115,55 @@ class Task extends React.Component<Props, State> {
...
 
+        // fileid is required for api calls, check for presence
+        const fileId = get(task_links, 'entries[0].target.id');
         if (!fileId) {
-            return Promise.reject();
+            return;
         }
 
         this.setState({ isLoading: true });
+        api.getTaskCollaboratorsAPI(false).getTaskCollaborators({
```
jstoffan: This is likely outside the scope of this PR, but it seems like all the data access and formatting should be happening in some kind of API module or container component rather than being mixed into the UI component.

src/elements/content-sidebar/activity-feed/task-new/AssigneeList.js
```diff
@@ -50,113 +46,77 @@ const AvatarDetails = React.memo(({ user, status, completedAt, className }) => {
...
+        onExpand,
+    } = props;
+    const { entries = [], next_marker } = users;
+    const entryCount = entries.length;
+    const hiddenAssigneeCount = Math.max(0, entryCount - initialAssigneeCount);
+    const numVisibleAssignees = isOpen ? entryCount : initialAssigneeCount;
+    const visibleUsers = entries
+        .slice(0, numVisibleAssignees)
+        .map(({ id, target, status, completed_at: completedAt }) => {
+            return (
```
priyajeet: This should/can be its own component/file

src/elements/content-sidebar/activity-feed/task-new/AssigneeList.js
```diff
@@ -50,113 +46,77 @@ const AvatarDetails = React.memo(({ user, status, completedAt, className }) => {
...
-
-        this.setState({
-            isCollapsed: false,
-        });
-    };
+    const maxAdditionalAssignees = TASKS_PAGE_SIZE - initialAssigneeCount;
+    const hasMoreAssigneesThanPageSize = hiddenAssigneeCount > maxAdditionalAssignees || next_marker;
+    const additionalAssigneeCount = hasMoreAssigneesThanPageSize
+        ? `${maxAdditionalAssignees}+`
+        : `${hiddenAssigneeCount}`;
```
priyajeet: unnecessary `

src/elements/content-sidebar/activity-feed/task-new/AssigneeList.js
```diff
@@ -50,113 +46,77 @@ const AvatarDetails = React.memo(({ user, status, completedAt, className }) => {
...
+                <span>
+                    <PlainButton
+                        data-resin-target="showlessbtn"
+                        data-testid="show-less-assignees"
+                        onClick={onCollapse}
+                        className="lnk bcs-AssigneeList-expandBtn"
+                    >
+                        <FormattedMessage {...messages.taskShowLessAssignees} />
+                    </PlainButton>
+                </span>
```
priyajeet: span needed? here and above

src/elements/content-sidebar/activity-feed/task-new/Task.js
booc0mtaco: I was thinking the same for this one, but wasn't sure where

jstoffan: I'm sure there are lots of options.  One option would be to add a wrapping container component, maybe called TaskContainer, that only handles data access and state management. You can see an example of this pattern with the [VersionsSidebarContainer](https://github.com/box/box-ui-elements/blob/master/src/elements/content-sidebar/versions/VersionsSidebarContainer.js) and its child UI component [VersionsSidebar](https://github.com/box/box-ui-elements/blob/master/src/elements/content-sidebar/versions/VersionsSidebar.js). The rule of thumb I follow is that declared HTML should live in a separate component from data access.

shahzadaziz: We can do this as part of refactor bucket.

PR #1304, opened 2019-06-19
feat(unified-share-modal): add property to limit the number of contacts

stevegcoder:
Add property to limit the number of contact pills that can be created at one time.
Remove reset of error on contact input.  This preserves the contact limit error.

### DIFF
--- a/i18n/en-US.properties
+++ b/i18n/en-US.properties
@@ -1170,6 +1170,8 @@ boxui.unifiedShare.collaboration.userCollabText = User
 boxui.unifiedShare.collaboratorListTitle = People in '{itemName}'
 # This string is displayed as tooltip on hovering over expire icon for collab
 boxui.unifiedShare.collaborators.expirationTooltipClickableText = Access expires on {date}. Click for details.
+# Error message when more than the maximum number of contacts is entered
+boxui.unifiedShare.contactsExceedLimitError = Oops! The maximum number of collaborators that can be added at once is {maxContacts} collaborators. Please try again by splitting your invitations into batches.
 # Text shown in share modal when there is at least one external collaborators
 boxui.unifiedShare.contentSharedWithExternalCollaborators = This content will be shared with external collaborators.
 # Text used in button label to describe permission level - co-owner
--- a/src/features/unified-share-modal/EmailForm.js
+++ b/src/features/unified-share-modal/EmailForm.js
@@ -28,6 +28,7 @@ import type { SelectOptionProp } from '../../components/select-field/props';
 type Props = {
     cancelButtonProps?: Object,
     children?: React.Node,
+    contactLimit?: number,
     contactsFieldAvatars?: React.Node,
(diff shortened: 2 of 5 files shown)


boxcla:

priyajeet:
Is it a breaking change? As in the consumer will have to make changes otherwise it will break?

cweeii:
@priyajeet This shouldn't be breaking since contactLimit is an optional prop.

src/features/unified-share-modal/EmailForm.js
```diff
@@ -81,21 +82,37 @@ class EmailForm extends React.Component<Props, State> {
...
-        updateSelectedContacts([...this.props.selectedContacts, ...contacts]);
+        updateSelectedContacts([...selectedContacts, ...contacts]);
+
+        let contactsFieldError = '';
+        if (contactLimit !== undefined && selectedContacts.length + contacts.length > contactLimit) {
+            contactsFieldError = intl.formatMessage(messages.contactsExceedLimitError, {
+                maxContacts: contactLimit,
+            });
+        }
+        this.setState({ contactsFieldError });
```
cweeii: It seems like you're doing this calculation and setting state 3 times in handleContactAdd, handleContactRemove, and handleSubmit.  Can we move it to componentDidUpdate?

stevegcoder:
@priyajeet I put breaking change because it doesn't clear the error when typing into the input.  This is more of a change/break in functionality, so that's why I have it tagged.  I can remove Breaking change.

src/features/unified-share-modal/EmailForm.js
```diff
@@ -203,6 +202,23 @@ class EmailForm extends React.Component<Props, State> {
         return emailValidator(text);
     };
 
+    componentDidUpdate(prevProps: Props, prevState: State) {
```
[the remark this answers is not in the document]
stevegcoder: Hey, I added a check to disable the button when there's an error.  Also I moved the check from componentDidUpdate to onContactAdd/Remove and submit to be more explicit.

i18n/en-US.properties
```diff
@@ -1168,6 +1168,8 @@ boxui.unifiedShare.collaboration.userCollabText = User
 boxui.unifiedShare.collaboratorListTitle = People in '{itemName}'
 # This string is displayed as tooltip on hovering over expire icon for collab
 boxui.unifiedShare.collaborators.expirationTooltipClickableText = Access expires on {date}. Click for details.
+# Error message when more than the maximum number of contacts is entered
+boxui.unifiedShare.contactsExceedLimitError = Oops! You've reached the maximum number of collaborators that can be added at once ({maxContacts} collaborators). Please try again by splitting your invitation into batches.
```
cweeii: Can we change the wording to `Oops!  The maximum number of collaborators that can be added at once is {maxContacts} collaborators.  Please try again by splitting your invitations into batches.`.  I think it's more easy to localize than having something in parentheses.

PR #1305, opened 2019-06-19
fix(nav-sidebar): reduce size of toggle button to fix focus outline

nickgriffin1:
Reduces the size of the sidebar toggle button to reduce the overflow of the focus outline, which is a bit jarring when hovering other items. Below shows the reduced size.

Before
image

After
image

### DIFF
--- a/src/features/left-sidebar/LeftSidebar.md
+++ b/src/features/left-sidebar/LeftSidebar.md
@@ -241,7 +241,7 @@ const menuItems = [
     selected: false,
     showTooltip: true,
     iconComponent: IconFavorites,
-    showLoadingIndicator: true,
+    showLoadingIndicator: false,
     onToggleCollapse: onToggleCollapse,
     collapsed: state.collapsed.favorites,
     placeholder: 'Drag items here for quick access',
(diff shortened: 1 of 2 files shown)


boxcla:

nickgriffin1:
@jstoffan I'll do some measurements and look into that.

nickgriffin1:
@jstoffan Updated with before and after pics. Also updated the CSS so that the icon spins in place. I made it a centered flexbox in order to do that. Before it was rotating on an axis which caused a couple pixels of misalignment.

src/features/left-sidebar/LeftSidebar.md
```diff
@@ -241,7 +241,7 @@ const menuItems = [
     selected: false,
     showTooltip: true,
     iconComponent: IconFavorites,
-    showLoadingIndicator: true,
+    showLoadingIndicator: false,
```
jstoffan: Is this intentional?

nickgriffin1: Yes. Before the example shows favorites as loading. I think it would be more useful to display it normally.

PR #1307, opened 2019-06-20
build: fix position of husky disabling flag

priyajeet:


boxcla:

scripts/cut_release.sh
```diff
@@ -247,7 +247,7 @@ push_new_release() {
     check_untracked_files || return 1
 
     # Run the release
-    if ! yarn semantic-release --no-ci; then
+    if ! HUSKY_SKIP_HOOKS=1 yarn semantic-release --no-ci; then
```
booc0mtaco: this makes sense; semantic release needs to make some quick tweaks to the package json, release, then restore the file

PR #1308, opened 2019-06-20
refactor(task-new): componentize task assignee css

alexkrolick:
- This pulls out the CSS for the avatar component so that it imports its own stylesheet instead of relying on the AvatarGroup stylesheet
- Slight refactors:
  - Offset the border by nudging badges by 1px
  - Change avatar inner component size to 100% of the container instead of hardcoding the size twice. This makes it so adding a class to the container will set the width for both without referencing private classes or increasing specificity.

Screen Shot 2019-06-20 at 10 41 56 AM
Screen Shot 2019-06-20 at 10 41 41 AM

### DIFF
--- a/src/elements/content-sidebar/activity-feed/task-new/AssigneeList.scss
+++ b/src/elements/content-sidebar/activity-feed/task-new/AssigneeList.scss
@@ -1,7 +1,7 @@
 @import '../../../common/variables';
 @import './mixins';
 
-$avatar-size: 28px;
+$bcs-AssigneeList-avatar-size: 32px;
 
 .bcs-AssigneeList-list {
     margin: 0;
@@ -13,20 +13,17 @@ $avatar-size: 28px;
     display: flex;
     margin: 6px 0 0 0;
     padding: 3px 0;
+}
 
-    .bcs-AssigneeList-detailsStatus {
-        color: $bdl-neutral-02;
-        font-size: 12px;
-    }
+.bcs-AssigneeList-detailsStatus {
+    color: $bdl-neutral-02;
+    font-size: 12px;
 }
 
 .bcs-AssigneeList-listItemAvatar {
+    height: $bcs-AssigneeList-avatar-size;
     margin-right: 10px;
-
-    .bcs-AvatarGroup-avatar {
-        height: $avatar-size;
-        width: $avatar-size;
-    }
+    width: $bcs-AssigneeList-avatar-size;
 }
 
 .bcs-AssigneeList-listItemDetails {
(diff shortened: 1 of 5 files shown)


boxcla:

src/elements/content-sidebar/activity-feed/task-new/AssigneeList.scss
```diff
@@ -13,20 +13,17 @@ $avatar-size: 28px;
...
     padding: 3px 0;
+}
 
-    .bcs-AssigneeList-detailsStatus {
-        color: $bdl-neutral-02;
-        font-size: 12px;
-    }
+.bcs-AssigneeList-detailsStatus {
+    color: $bdl-neutral-02;
+    font-size: 12px;
```
booc0mtaco: It's funny that we have almost equal number of usages of 11px as 12px across similar use cases. We probably need to to scrub one of these sizes away...

src/elements/content-sidebar/activity-feed/task-new/AssigneeList.scss
```diff
@@ -13,20 +13,17 @@ $avatar-size: 28px;
...
     margin: 6px 0 0 0;
     padding: 3px 0;
+}
 
-    .bcs-AssigneeList-detailsStatus {
-        color: $bdl-neutral-02;
-        font-size: 12px;
-    }
+.bcs-AssigneeList-detailsStatus {
+    color: $bdl-neutral-02;
```
booc0mtaco: We could swap in the `bdl-gray` equivalent here, but that's an easy find/replace that can be done later.

src/elements/content-sidebar/activity-feed/task-new/AvatarGroupAvatar.scss
```diff
@@ -0,0 +1,20 @@
+@import '../../../common/variables';
+@import './mixins';
+
+$bcs-AvatarGroupAvatar-avatar-size: 32px;
```
booc0mtaco: Curious about component-level mixins for future use. One practice we could consider is adding variables in the components for `import`ing and re-use in features elsewhere. 

`$bcs-AvatarGroupAvatar-avatar-size: $bdl-Avatar-size;`

src/elements/content-sidebar/activity-feed/task-new/AvatarGroupAvatar.scss
```diff
@@ -0,0 +1,20 @@
...
+
+.bcs-AvatarGroupAvatar {
+    display: inline-block;
+    height: $bcs-AvatarGroupAvatar-avatar-size;
+    position: relative;
```
booc0mtaco: I think we needed this somewhere in `Badgeable`. will have a look when there's time.

src/elements/content-sidebar/activity-feed/task-new/AvatarGroupAvatar.scss
alexkrolick: We should move the support for badging into the avatar component itself, IMO. Also refactor it to support multiple sizes instead of just 32px and 44px.

src/elements/content-sidebar/activity-feed/task-new/AssigneeList.scss
alexkrolick: we were using the neutral scheme pretty consistently for tasks features, so let's commit to that in a larger scrub later

PR #1310, opened 2019-06-20
fix(header-flyout): ConstrainToWindow to fix IE11 scroll bug

shahzadaziz:
This fixes : https://jira.inside-box.net/browse/COLLAB-644


boxcla:

PR #1312, opened 2019-06-20
fix(task-new): alignment fixes/css refactor

alexkrolick:


boxcla:

src/elements/content-sidebar/activity-feed/task-new/TaskActions.scss
```diff
@@ -0,0 +1,3 @@
+button:first-of-type.bcs-TaskActions-button {
```
jstoffan: Please avoid element selectors in the future. This can be accomplished with sibling selectors, instead.

PR #1313, opened 2019-06-20
fix(tasks-new): show "a prior collaborator" for deleted users

tjgupta:


boxcla:

src/elements/common/messages.js
```diff
@@ -1264,6 +1264,12 @@ const messages = defineMessages({
         description: 'Label for back button',
         defaultMessage: 'Back',
     },
+    priorCollaborator: {
```
jstoffan: This should probably live in a task-specific messages bundle.

tjgupta: I was thinking we'd want to use this string anywhere where we show a user who could be deleted. It seems like currently this is only tasks, but I'd imagine this will be elsewhere in the future no?

jstoffan: I think we should err on the side of scoping things small rather than big. Otherwise all our messages would live in the common bundle

tjgupta: Actually it looks like all the tasks messages are already in this file.

booc0mtaco: yes, src/elements/common/messages.js contains all the strings except versions and inline delete for elements

jstoffan: Understood. Let's considering moving the messages to the tasks component directory in a follow-up PR.

PR #1315, opened 2019-06-21
fix(pill-selector): ux tweaks

jpfranco:
This update is meant to restore some behaviors that were present on the deprecated `PillSelector` from `box-react-ui` and to introduce some new ones that are needed by other use cases. All updates are configurable and turned off by default in order to avoid breaking changes.

The changes included are the following:

- Prevent dropdown from being displayed automatically on input focus. This is bad UX when the pill selector has a very large list of options.
- Allow unmatched input to be cleared automatically when it doesn't result in new pills being added. Useful when copy and pasting existing values accidentally, for example.
- Allow setting a `parseItems` function that returns an array of options (previously only an array of strings was supported).
- Allow automatically highlighting the first option when the dropdown is displayed or filtered.
- Added parser function that automatically maps copy pasted input to `Option` objects from the list of dropdown options.
- Fixed issue in which selected option would not be scrolled into view when using the arrow keys.

### DIFF
--- a/src/components/pill-selector-dropdown/PillSelectorDropdown.js
+++ b/src/components/pill-selector-dropdown/PillSelectorDropdown.js
@@ -56,6 +56,10 @@ type Props = {
     selectedOptions: SelectedOptions,
     /** Array or Immutable list with data for the dropdown options to select */
     selectorOptions: Array<Object> | List<Object>,
+    /** Determines whether or not input text is cleared automatically when it does not result in new pills being added */
+    shouldClearUnmatchedInput?: boolean,
+    /** Determines whether or not the first item is highlighted automatically when the dropdown opens */
+    shouldSetActiveItemOnOpen?: boolean,
     /** Array of suggested collaborators */
     suggestedPillsData?: Array<Object>,
     /** String decribes the datapoint to filter by so that items in the form are not shown in suggestions. */
@@ -86,6 +90,8 @@ class PillSelectorDropdown extends React.Component<Props, State> {
         placeholder: '',
         selectedOptions: [],
         selectorOptions: [],
+        shouldClearUnmatchedInput: false,
+        shouldSetActiveItemOnOpen: false,
         validator: () => true,
     };
 
@@ -104,15 +110,28 @@ class PillSelectorDropdown extends React.Component<Props, State> {
             pills = pills.filter(pill => validator(pill));
         }
 
-        return pills.map(pill => ({
-            displayText: pill,
-            text: pill, // deprecated, left for backwards compatibility
-            value: pill,
-        }));
+        const normalizedPills = pills.map(pill =>
+            typeof pill === 'string'
+                ? {
+                      displayText: pill,
+                      text: pill, // deprecated, left for backwards compatibility
+                      value: pill,
+                  }
+                : pill,
+        );
+        return normalizedPills;
     };
 
     addPillsFromInput = () => {
-        const { allowCustomPills, onInput, onPillCreate, onSelect, selectedOptions, validateForError } = this.props;
+        const {
+            allowCustomPills,
+            onInput,
+            onPillCreate,
+            onSelect,
+            selectedOptions,
+            shouldClearUnmatchedInput,
+            validateForError,
+        } = this.props;
         const { inputValue } = this.state;
 
         // Do nothing if custom pills are not allowed
@@ -131,12 +150,18 @@ class PillSelectorDropdown extends React.Component<Props, State> {
             // Reset inputValue
             this.setState({ inputValue: '' });
             onInput('');
-        } else if (validateForError && (inputValue !== '' || selectedOptions.length === 0)) {
-            /**
-             * If no pills were added, but an inputValue exists or
-             * there are no pills selected, check for errors
-             */
-            validateForError(inputValue);
+        } else {
+            if (validateForError && (inputValue !== '' || selectedOptions.length === 0)) {
+                /**
+                 * If no pills were added, but an inputValue exists or
+                 * there are no pills selected, check for errors
+                 */
+                validateForError(inputValue);
+            }
+            if (shouldClearUnmatchedInput) {
+                this.setState({ inputValue: '' });
+                onInput('');
+            }
         }
     };
 
@@ -209,6 +234,7 @@ class PillSelectorDropdown extends React.Component<Props, State> {
             suggestedPillsData,
             suggestedPillsFilter,
             suggestedPillsTitle,
+            shouldSetActiveItemOnOpen,
             validator,
         } = this.props;
 
@@ -219,6 +245,7 @@ class PillSelectorDropdown extends React.Component<Props, State> {
                 onEnter={this.handleEnter}
                 onSelect={this.handleSelect}
                 overlayTitle={overlayTitle}
+                shouldSetActiveItemOnOpen={shouldSetActiveItemOnOpen}
                 selector={
                     <Label text={label}>
                         <PillSelector
--- a/src/components/pill-selector-dropdown/PillSelectorDropdown.js
+++ b/src/components/pill-selector-dropdown/PillSelectorDropdown.js
@@ -28,6 +28,8 @@ type Props = {
     disabled: boolean,
     /** Index at which to insert a divider */
     dividerIndex?: number,
+    /** A CSS selector matching the element to use as a boundary when auto-scrolling dropdown elements into view. When not provided, boundary will be determined by scrollIntoView utility function */
+    dropdownScrollBoundarySelector?: string,
     /** Error message */
     error?: React.Node,
     /** Passed in by `SelectorDropdown` for accessibility */
@@ -226,6 +228,7 @@ class PillSelectorDropdown extends React.Component<Props, State> {
             className,
             disabled,
             dividerIndex,
+            dropdownScrollBoundarySelector,
             error,
             inputProps,
             label,
@@ -248,6 +251,7 @@ class PillSelectorDropdown extends React.Component<Props, State> {
                 onEnter={this.handleEnter}
                 onSelect={this.handleSelect}
                 overlayTitle={overlayTitle}
+                scrollBoundarySelector={dropdownScrollBoundarySelector}
                 shouldSetActiveItemOnOpen={shouldSetActiveItemOnOpen}
                 selector={
                     <Label text={label}>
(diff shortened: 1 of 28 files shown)


boxcla:
Hi @jpfranco, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

src/components/pill-selector-dropdown/PillSelectorDropdown.js
```diff
@@ -104,15 +110,28 @@ class PillSelectorDropdown extends React.Component<Props, State> {
             pills = pills.filter(pill => validator(pill));
         }
 
-        return pills.map(pill => ({
-            displayText: pill,
-            text: pill, // deprecated, left for backwards compatibility
-            value: pill,
-        }));
+        const normalizedPills = pills.map(pill =>
```
jpfranco: Note that this puts the onus on the caller to also write a validator that handles the custom values returned by their parser

i18n/en-US.properties
```diff
@@ -222,6 +222,8 @@ be.previousFile = Previous File
 be.previousSegment = Previous Segment
 # Label for print action
 be.print = Print
+# If a user has been deleted, we call the user "a prior collaborator" - meaning someone who used to be able to collaborate on the content.
+be.priorCollaborator = A Prior Collaborator
```
jpfranco: Hmm not sure how this got here.

alexkrolick: Should go away if you rebase, there was a missed i18n build

jpfranco:
> Can you provide a gif of how this new behavior is? If it's currently not as part of the examples, would be great if you can add that to the examples!

PillSelector (1)

The gifs framerate is a bit slow

src/components/pill-selector-dropdown/waitForInputFilter.js
```diff
@@ -0,0 +1,15 @@
+// @flow
+import defaultDropdownFilter from './defaultDropdownFilter';
+
+import type { Option } from './flowTypes';
+
+// Same as default filter except that it won't show the selector dropdown until the user begins typing
+// (as opposed to showing it when the input is focused)
+function waitForInputFilter(options: Array<Option>, selectedValues?: Array<Option>, inputText?: string): Array<Option> {
```
priyajeet: I am wondering if this should be part of the parent app than here. Since I am guessing you will pass this filter as a prop for the Formik `<Field>`.

On the other hand, we could create a sub-folder called `filters` and move both this and default one in there and then other filters can also be added.

A 3rd way would be that maybe we have this a a prop something like `is(should?)DropdownOpenOnFocus` and then just use that boolean within PillSelectorDropdownField to return [] if true OR do ->

Additively 4th option would be that we pass the prop from the above as a new prop to PillSelectorDropdown which passes the prop to SelectorDropdown.js and guard the handleFocus() function which is what opens it. So selectordropdown gets a new prop to prevent opening on focus like you added the prop to focus 1st item.

src/components/selector-dropdown/SelectorDropdown.js
```diff
@@ -59,8 +61,17 @@ class SelectorDropdown extends React.Component<Props, State> {
     }
 
     componentWillReceiveProps(nextProps: Props) {
```
priyajeet: While you are at it, can you rename this to UNSAFE_componentWillReceiveProps. Or if you want to be adventurous, change this function to be componentDidMount :)

src/components/pill-selector-dropdown/mapInputValuesToOptionsParser.js
```diff
@@ -0,0 +1,50 @@
...
+
+import type { Option } from './flowTypes';
+
+// Custom input parser for the PillSelectorDropdownField that automatically maps
+// a list of comma separated values to their respective Option objects. The parser
+// will try to find exact but case-insenstive matches using the option's displayText
+// property. If a token does not match displayText, the parser will also try to match
+// the string representation of the value property. This last match is useful for country
+// codes and other similar use cases.
+function mapInputValuesToOptionsParser(
```
priyajeet: probably remove the suffix Parser. mapInputValuesToOptions should suffice.

src/components/pill-selector-dropdown/mapInputValuesToOptionsParser.js
```diff
@@ -0,0 +1,50 @@
...
+    inputValue: string,
+    options: Array<Option>,
+    selectedOptions: Array<Option>,
+): Array<Option> {
+    let mappedOptions;
+    const inputTokens = parseCSV(inputValue);
```
priyajeet: Would we need to use the custom parseItems() prop here? Like in PillSelectorDropdown.js parsePills()?
let pills = parseItems ? parseItems(inputValue) : parseCSV(inputValue);

src/components/pill-selector-dropdown/mapInputValuesToOptionsParser.js
priyajeet: Also, should we think making this the default? As in having this logic be called by parsePills() in PillSelectorDropdown?

Since it feels this may be useful for other ones too, like collab pill selector. I copy paste Foo, Bar, Bam@email.com and it should use those options from the collab drop down if available?

priyajeet: "making this the default" or guarded by a prop

src/components/pill-selector-dropdown/PillSelectorDropdown.js
```diff
@@ -131,12 +150,18 @@ class PillSelectorDropdown extends React.Component<Props, State> {
...
+        } else {
+            if (validateForError && (inputValue !== '' || selectedOptions.length === 0)) {
+                /**
+                 * If no pills were added, but an inputValue exists or
+                 * there are no pills selected, check for errors
+                 */
+                validateForError(inputValue);
+            }
+            if (shouldClearUnmatchedInput) {
+                this.setState({ inputValue: '' });
```
priyajeet: probably move 162-163 to its own function, and also replace 151-152

src/components/pill-selector-dropdown/waitForInputFilter.js
jpfranco: @priyajeet do you have a preference for any of these options? I'm leaning towards the 4th since, now that I think about it, this behavior is not directly related to filtering. It probably makes more sense to handle this with a prop that is independent of filtering.

src/components/pill-selector-dropdown/mapInputValuesToOptionsParser.js
jpfranco: Nope, this function is actually meant to be used as the `parseItems` prop itself.

src/components/pill-selector-dropdown/mapInputValuesToOptionsParser.js
jpfranco: Yes I think this is useful enough to be the default but I didn't want to change the current behavior. This parser only works in the context of `PillSelectorDropdownField` which I believe is not widely used yet though. Do you want me to add a prop that turns on this parser?

src/components/pill-selector-dropdown/waitForInputFilter.js
priyajeet: 4th one would have been my pref too

src/components/pill-selector-dropdown/mapInputValuesToOptionsParser.js
priyajeet: Hmm, I would say just make it default for the Field for now. Can re-visit later.

src/components/pill-selector-dropdown/waitForInputFilter.js
jpfranco: @priyajeet on close inspection it looks like this is related to filtering after all since in general we don't want to display the dropdown when the filter value is empty. So for example we also want to close the dropdown if the user clears the input after typing, but the `SelectorDropdown` doesn't have context on that.

Are you ok with using the second option of having a "filters" folder? I think it would be good to keep the filter in this project since it could be useful to other people.

src/components/pill-selector-dropdown/PillSelectorDropdownField.js
```diff
@@ -85,6 +88,16 @@ class PillSelectorDropdownField extends React.PureComponent<Props, State> {
         onChange(this.createFakeEventTarget(name, options));
     };
 
+    handleParseItems = (inputValue: string): ?Array<Option> => {
+        const { field, inputParser, options } = this.props;
+        const { value: selectedOptions } = field;
+
+        if (inputParser) {
```
priyajeet: I would add inputParser to defaultProps, defaulting to `mapInputValuesToOptions` so that here you dont need a if / else block. And probably rename `mapInputValuesToOptions` to something like `defaultInputParser`.

src/components/selector-dropdown/SelectorDropdown.js
```diff
@@ -15,25 +15,27 @@ function stopDefaultEvent(event) {
...
-    /** Function called with the index of the selected option and the event (selected by keyboard or click) */
-    selector: React.Element<any>,
     /** Component containing an input text field and takes `inputProps` to spread onto the input element */
-    shouldScroll?: boolean,
+    selector: React.Element<any>,
     /** Boolean to indicate whether the dropdown should scroll */
+    shouldScroll?: boolean,
+    /** Determines whether or not the first item is highlighted automatically when the dropdown opens */
+    shouldSetActiveItemOnOpen?: boolean,
+    /** Optional title text that will be rendered above the list */
```
priyajeet: I thought you were going to add a new prop `shouldOpenOnFocus` or something like that?

src/components/pill-selector-dropdown/waitForInputFilter.js
priyajeet: thats works

src/components/pill-selector-dropdown/PillSelectorDropdownField.js
jpfranco: Good idea, will change this

jpfranco: @priyajeet sorry, I just remembered why I did it like this. `inputParser` and `parseItems` have different signatures. I added `options` and `selectedOptions` to `inputParser` in order to allow greater extensibility. We need this extra function in order to pass the extra parameters to `inputParser`

jpfranco:
CLA signed

src/components/pill-selector-dropdown/PillSelectorDropdownField.js
jpfranco: Had misunderstood your point. This has been updated now

PR #1316, opened 2019-06-21
feat(versions): Add support for version limits to sidebar

jstoffan:
This feature will require a change to the File V2 endpoint to support the new **version_limit** field, but is backward-compatible.


boxcla:

src/api/File.js
```diff
@@ -42,6 +42,23 @@ class File extends Item {
...
+     * Format the API response
+     *
+     * @param file
+     * @returns {BoxItem}
+     */
+    format(file: BoxItem) {
+        const { ...result } = file;
+        const { version_limit } = file;
+
+        if (version_limit !== undefined) {
```
jeremypress: looks like version_limit will either be `undefined`, or a `number`. Can we do:
`result.version_limit = version_limit || Infinity`?

jstoffan: The `version_limit` value can also be `0`, so the explicit null check is necessary. I'll update the flow type to better reflect that it's a nullable value.

PR #1317, opened 2019-06-21
fix(tasks-new): update docs and strings

booc0mtaco:
Remove comment that's no longer needed, and re-build translations file


boxcla:

PR #1318, opened 2019-06-21
fix(template-dropdown): add onMenuToggle handler to TemplateDropdown

ChenCodes:
Changes in this PR:
- add onMenuToggleHandler because in the client app there is a use case where the caret icon needs to be transformed depending on if the menu is open or not

ezgif com-video-to-gif (6)

### DIFF
--- a/src/features/metadata-instance-editor/TemplateDropdown.js
+++ b/src/features/metadata-instance-editor/TemplateDropdown.js
@@ -24,6 +24,7 @@ type Props = {
     intl: any,
     isDropdownBusy?: boolean,
     onAdd: (template: MetadataTemplate) => void,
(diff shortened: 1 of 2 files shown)


boxcla:

src/features/metadata-instance-editor/TemplateDropdown.js
```diff
@@ -218,6 +223,12 @@ class TemplateDropdown extends React.PureComponent<Props, State> {
     };
 
     onClose = () => {
+        const { onMenuToggle } = this.props;
+
+        if (onMenuToggle) {
+            onMenuToggle(false);
+        }
+
```
macrotim: Each call to `onMenuToggle` deserve unit tests.

PR #1321, opened 2019-06-21
fix(task-new): fix placement of task status checkmark

tjgupta:


boxcla:

PR #1323, opened 2019-06-21
fix(task-form): don't disable task form submit button

tjgupta:


boxcla:

PR #1325, opened 2019-06-24
chore: update metadata icon background color

ChenCodes:
As requested by design, the background for these icons should be set to none

Screen Shot 2019-06-24 at 10 45 35 AM
Screen Shot 2019-06-24 at 10 45 19 AM

### DIFF
--- a/src/icons/badges/MetadataDefaultBadge.js
+++ b/src/icons/badges/MetadataDefaultBadge.js
@@ -1,6 +1,5 @@
 // @flow
 import * as React from 'react';
-import { bdlNeutral04 } from '../../styles/variables';
 import AccessibleSVG from '../accessible-svg';
 
 type Props = {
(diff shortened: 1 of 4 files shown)


boxcla:

src/icons/badges/MetadataDefaultBadge.js
```diff
@@ -20,7 +19,7 @@ const MetadataDefaultBadge = ({ className = '', height = 16, title, width = 16 }
         width={width}
     >
         <g fill="none" fillRule="evenodd">
-            <circle cx="8" cy="8" fill={bdlNeutral04} r="7.5" stroke="#767676" />
+            <circle cx="8" cy="8" fill="none" r="7.5" stroke="#767676" />
```
ConradJChan: Since you're here can we use the variables for the stroke? Looks like `#767676` exists in `_variables.scss` but can the one from `MetadataActiveBadge` use a similar existing color?

PR #1329, opened 2019-06-24
fix(deps): downgrade axios peer dep

alexkrolick:
The bump to 0.19 24 days ago contained breaking changes


boxcla:

PR #1330, opened 2019-06-25
feat(versions): Add version group headers based on relative date

jstoffan:
grouping

### DIFF
--- a/src/elements/content-sidebar/versions/VersionsList.js
+++ b/src/elements/content-sidebar/versions/VersionsList.js
@@ -6,50 +6,32 @@
 
 import React from 'react';
 import { Route } from 'react-router-dom';
-import { FormattedMessage } from 'react-intl';
-import messages from './messages';
 import VersionsItem from './VersionsItem';
-import type { VersionActionCallback } from './Versions';
 import './VersionsList.scss';
 
 type Props = {
+    currentId: string,
     fileId: string,
-    isLoading: boolean,
-    onDelete: VersionActionCallback,
-    onDownload: VersionActionCallback,
-    onPreview: VersionActionCallback,
-    onPromote: VersionActionCallback,
-    onRestore: VersionActionCallback,
     versions: Array<BoxItemVersion>,
 };
 
-const VersionsList = ({ isLoading, versions, ...rest }: Props) => {
-    if (!isLoading && !versions.length) {
-        return (
-            <div className="bcs-VersionsList bcs-is-empty">
-                <FormattedMessage {...messages.versionsEmpty} />
-            </div>
-        );
-    }
-
-    return (
-        <ul className="bcs-VersionsList">
-            {versions.map((version, index) => (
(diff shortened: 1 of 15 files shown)


boxcla:

src/elements/content-sidebar/versions/VersionsList.js
```diff
@@ -34,20 +35,36 @@ const VersionsList = ({ isLoading, versions, ...rest }: Props) => {
...
+                const isCurrent = index === 0;
+                const previousVersion = versions[Math.max(0, index - 1)];
+                const previousGroup = getGroup(previousVersion);
+                const newGroup = getGroup(version);
+                const hasGroup = isCurrent || previousGroup !== newGroup;
+
+                return (
+                    <React.Fragment key={version.id}>
+                        {hasGroup && (
+                            <li className="bcs-VersionsList-heading">
```
ConradJChan: Is there another semantic way to do subheaders in a list? I'm wondering how screen readers might read this

src/elements/content-sidebar/versions/VersionsList.scss
```diff
@@ -11,6 +11,17 @@
...
 }
 
+.bcs-VersionsList-heading {
+    color: $bdl-gray-50;
+    font-size: 14px;
+    padding-bottom: 10px;
+    padding-top: 15px;
+
+    &:empty {
+        display: none;
```
ConradJChan: When would this occur?

jstoffan: It won't; it's a stray declaration from an earlier revision. Good catch, will remove.

src/elements/content-sidebar/versions/VersionsList.js
jstoffan: Good question. This is how our dropdown menu works, so I followed the convention here as well. I'll take a quick look online to see what I find.

src/elements/content-sidebar/versions/VersionsSidebar.js
```diff
@@ -10,42 +10,71 @@ import InlineError from '../../../components/inline-error';
...
+
+                {showEmpty && (
+                    <div className="bcs-Versions-empty">
+                        <FormattedMessage {...messages.versionsEmpty} />
+                    </div>
+                )}
+
+                {showVersions && (
+                    <ul className="bcs-Versions-menu">
+                        {Object.keys(versionGroups).map(versionGroupKey => (
```
ConradJChan: Do we need to worry about sorting the order of the versionGroups? Assuming the API data comes back in a sorted order this should work, but I'm thinking about the corner case of if version has no `createdAt` (is that even possible?) in which case it gets no heading and has a key of `''` and I don't know where that would appear in the `Object.keys` order

jstoffan: It doesn't look like `created_at` is a nullable field and the server explicitly sorts versions by that value for its response. I've removed the unnecessary existence checks from the new components.

PR #1332, opened 2019-06-25
fix(tasks): prevent resize of assignee avatar

alexkrolick:


boxcla:

PR #1333, opened 2019-06-25
fix(content-explorer): MDVGridView import fixes

rickstefanik:
Found that many of the imported styles were unnecessary or unreachable.  Removed base-table and pure-component.


boxcla:

src/components/grid-view/GridView.scss
```diff
@@ -33,7 +33,7 @@ $columns5prct: calc(20% - 8px); /* 100%/5 - 40px/5 */
         margin: 0;
     }
 
-    .table-row {
+    .bdl-GridView-tableRow {
```
ConradJChan: We should also take the opportunity to get rid of the global style variables at the top of this file

src/components/grid-view/GridView.js
```diff
@@ -1,13 +1,12 @@
...
 import { CellMeasurer, CellMeasurerCache } from 'react-virtualized/dist/es/CellMeasurer';
-import { Column } from 'react-virtualized/dist/es/Table';
+import Table, { Column } from 'react-virtualized/dist/es/Table';
 import uniqueId from 'lodash/uniqueId';
-
-import VirtualizedTable from '../virtualized-table/VirtualizedTable';
-
 import GridViewSlot from './GridViewSlot';
 
+import 'react-virtualized/styles.css';
```
ConradJChan: While we're at it, if there's no state, can we remove L33 and L39's references?

src/components/grid-view/GridView.js
```diff
@@ -102,18 +101,22 @@ class GridView extends React.Component<Props, State> {
...
                 disableHeader
                 height={height}
                 rowCount={rowCount}
                 rowGetter={this.rowGetter}
                 rowHeight={this.cache.rowHeight}
                 width={width}
+                gridClassName="bdl-GridViw-body"
+                rowClassName="bdl-GridView-tableRow"
+                sortDirection="ASC"
+                tabIndex={null}
```
ConradJChan: do we need to provide this if it's already optional?

src/components/grid-view/__tests__/GridView-test.js
```diff
@@ -0,0 +1,22 @@
...
+
+import GridView from '../GridView';
+
+describe('components/grid-view/GridView', () => {
+    test('should render()', () => {
+        const collection = { items: [{ type: 'folder', id: '001', name: 'Example Folder' }] };
+        const wrapper = shallow(
+            <GridView
+                columnCount={5}
+                count={10}
```
ConradJChan: Is count the number of items in the collection? Could we derive that from the collection itself rather than set it as a prop?

src/components/grid-view/GridView.js
```diff
@@ -54,7 +53,8 @@ class GridView extends React.Component<Props> {
     }
 
     cellRenderer = ({ dataKey, parent, rowIndex }: TableCellRendererParams) => {
-        const { columnCount, count, currentCollection, slotRenderer, onItemSelect } = this.props;
+        const { columnCount, currentCollection, slotRenderer, onItemSelect } = this.props;
+        const count = currentCollection.items ? currentCollection.items.length : 0;
```
ConradJChan: `lodash/get` is useful here to avoid this type of defensive checking, we use it in many places as such as https://github.com/box/box-ui-elements/blob/90bf2f3c90f158068b82605aaeb66e0aa2880b64/src/elements/content-sidebar/SidebarFileProperties.js#L8

PR #1334, opened 2019-06-25
fix(task-new): Minor visual tweak to add tasks menu

tjgupta:
Missed this visual difference before in my last PR. Trying to remove as many CSS overrides as we can so it matches with other similar menus like the "Open with" menu.


boxcla:

PR #1338, opened 2019-06-26
feat(icons): add collections icons for bolt and star

nickgriffin1:
image


boxcla:

booc0mtaco:
I see a class being applied for when the icon is selected, but no styles for this case. should we specify styles to control what visual changes `is-selected` applies?

Also, we should use a SUIT-able name like `bdl-is-selected` or maybe `bdl-is-icon-selected`.

nickgriffin1:
@booc0mtaco The 'is-selected' class is a legacy function. These icons could be used in LeftSidebar which applies that class to control the color of the icon for when a route is active. I believe this only could be fixed with a breaking change which I don't think is worth it.

src/icons/collections/IconCollectionsStar.js
```diff
@@ -0,0 +1,44 @@
...
+import classNames from 'classnames';
+
+import AccessibleSVG from '../accessible-svg';
+import { bdlNeutral03 } from '../../styles/variables';
+
+type Props = {
+    className?: string,
+    color?: string,
+    height?: number,
+    selected?: boolean,
```
j3tan: to andrew's point, this doesn't really belong on the Icon. `is-selected` should be passed in as a classname from the parent since these 'icons' are supposed to be dumb icon components that don't have a 'selected' state.

All the collections icons need to be changed to adhere to this

PR #1341, opened 2019-06-28
fix(sidebar): Update flexbox for Activity Feed Container

ConradJChan:
Fixes an issue in IE11 where the activity feed would render smaller than the full width and have extra whitespace on either side


boxcla:

PR #1339, opened 2019-06-27
fix(versions): Restrict actions and preview for watermarked files

jstoffan:

PR #1342, opened 2019-06-28
fix(versions): Relax restrictions for watermarked files

jstoffan:
This is a partial revert of https://github.com/box/box-ui-elements/pull/1339 due to updated requirements.


boxcla:

PR #1344, opened 2019-06-28
fix(left-sidebar): hide toggle icon on medium screen size

nickgriffin1:
Reduces the specificity of css so that the media query that applies `display: none` is not overridden by `display: flex` for the LeftSidebar.


boxcla:

PR #1345, opened 2019-06-28
fix(i18n): use plural format for task assignee message

alexkrolick:


boxcla:

i18n/en-US.properties
```diff
@@ -405,7 +405,9 @@ be.taskRejectErrorMessage = An error has occurred while rejecting this task. Ple
 # Button name to hide task assignee list
 be.taskShowLessAssignees = Show Less
 # Button name to expand task assignee list, additionalAssigneeCount is the number of additional task assignees that can be shown.
-be.taskShowMoreAssignees = Show {additionalAssigneeCount} More
+be.taskShowMoreAssignees = {additionalAssigneeCount, plural, one {Show # More} other {Show # More}}
```
jstoffan: Is the `plural` specification needed? It looks like the message variants for `one` and `other` are the same.

alexkrolick: in English, yes, but @ehoogerbeets pointed out it is different in other languages

PR #1349, opened 2019-06-29
feat(content-explorer): Add toggle button for grid view

rickstefanik:
Majority of the lines are due to the snapshot for the jest test.  In the future, I'll submit the code and tests for a PR in separate commits.

### DIFF
--- a/examples/src/features.js
+++ b/examples/src/features.js
@@ -12,4 +12,9 @@ export default {
             enabled: false,
         },
     },
+    contentExplorer: {
+        gridView: {
+            enabled: false,
+        },
+    },
 };
--- a/src/elements/common/flowTypes.js
+++ b/src/elements/common/flowTypes.js
@@ -1,4 +1,7 @@
 // @flow
+import { VIEW_MODE_GRID, VIEW_MODE_LIST } from '../../constants';
+
+type ViewMode = typeof VIEW_MODE_GRID | typeof VIEW_MODE_LIST;
 
 type ErrorType = {
     code: string,
@@ -7,5 +10,5 @@ type ErrorType = {
     message?: string,
 };
 
-// eslint-disable-next-line import/prefer-default-export
 export type { ErrorType };
+export type { ViewMode };
--- a/src/elements/common/messages.js
+++ b/src/elements/common/messages.js
@@ -127,6 +127,16 @@ const messages = defineMessages({
         description: 'Label for add action',
         defaultMessage: 'Add',
     },
+    gridView: {
+        id: 'be.gridView',
+        description: 'Label for switching to grid view',
+        defaultMessage: 'Switch to Grid View',
+    },
+    listView: {
+        id: 'be.listView',
+        description: 'Label for switching to list view',
+        defaultMessage: 'Switch to List View',
+    },
     sort: {
         id: 'be.sort',
         description: 'Label for sort action',
--- a/src/elements/common/sub-header/SubHeader.js
+++ b/src/elements/common/sub-header/SubHeader.js
@@ -7,6 +7,8 @@
 import React from 'react';
 import SubHeaderLeft from './SubHeaderLeft';
 import SubHeaderRight from './SubHeaderRight';
+import type { ViewMode } from '../flowTypes';
+import { VIEW_MODE_LIST } from '../../../constants';
 
 import './SubHeader.scss';
 
@@ -19,23 +21,27 @@ type Props = {
     onItemClick: Function,
     onSortChange: Function,
     onUpload: Function,
+    onViewModeChange?: (viewMode: ViewMode) => void,
     rootId: string,
     rootName?: string,
     view: View,
+    viewMode?: ViewMode,
 };
 
 const SubHeader = ({
-    rootId,
-    rootName,
+    canCreateNewFolder,
+    canUpload,
+    currentCollection,
+    isSmall,
+    onCreate,
     onItemClick,
     onSortChange,
-    currentCollection,
     onUpload,
-    onCreate,
-    canUpload,
-    canCreateNewFolder,
+    onViewModeChange,
+    rootId,
+    rootName,
     view,
-    isSmall,
+    viewMode = VIEW_MODE_LIST,
 }: Props) => (
     <div className="be-sub-header" data-testid="be-sub-header">
         <SubHeaderLeft
@@ -50,7 +56,9 @@ const SubHeader = ({
             canCreateNewFolder={canCreateNewFolder}
             canUpload={canUpload}
             currentCollection={currentCollection}
+            viewMode={viewMode}
             onCreate={onCreate}
+            onViewModeChange={onViewModeChange}
             onSortChange={onSortChange}
             onUpload={onUpload}
             view={view}
--- a/src/elements/common/sub-header/SubHeaderRight.js
+++ b/src/elements/common/sub-header/SubHeaderRight.js
@@ -46,13 +46,7 @@ const SubHeaderRight = ({
     return (
         <div className="be-sub-header-right">
             <FeatureFlag feature="contentExplorer.gridView.enabled">
-                {showGridButton && (
-                    <ViewModeChangeButton
-                        className="bdl-SubHeaderRight-ViewModeChangeButton"
-                        viewMode={viewMode}
-                        onViewModeChange={onViewModeChange}
-                    />
-                )}
+                {showGridButton && <ViewModeChangeButton viewMode={viewMode} onViewModeChange={onViewModeChange} />}
             </FeatureFlag>
 
             {showSort && !!sortBy && !!sortDirection && (
--- a/src/elements/common/sub-header/SubHeaderRight.js
+++ b/src/elements/common/sub-header/SubHeaderRight.js
@@ -7,7 +7,10 @@
 import React from 'react';
 import Sort from './Sort';
 import Add from './Add';
+import ViewModeChangeButton from './ViewModeChangeButton';
+import { FeatureFlag } from '../feature-checking';
 import { VIEW_FOLDER } from '../../../constants';
+import type { ViewMode } from '../flowTypes';
 import './SubHeaderRight.scss';
 
 type Props = {
@@ -17,25 +20,41 @@ type Props = {
     onCreate: Function,
     onSortChange: Function,
     onUpload: Function,
+    onViewModeChange?: (viewMode: ViewMode) => void,
     view: View,
+    viewMode: ViewMode,
 };
 
 const SubHeaderRight = ({
-    view,
-    onUpload,
-    onCreate,
-    canUpload,
     canCreateNewFolder,
+    canUpload,
     currentCollection,
+    onCreate,
+    onViewModeChange,
     onSortChange,
+    onUpload,
+    view,
+    viewMode,
 }: Props) => {
     const { sortBy, sortDirection, items = [] }: Collection = currentCollection;
+    const hasItems: boolean = items.length > 0;
     const isFolder: boolean = view === VIEW_FOLDER;
-    const showSort: boolean = isFolder && items.length > 0;
+    const showSort: boolean = isFolder && hasItems;
     const showAdd: boolean = (!!canUpload || !!canCreateNewFolder) && isFolder;
+    const showGridButton: boolean = isFolder && hasItems;
 
     return (
         <div className="be-sub-header-right">
+            <FeatureFlag feature="contentExplorer.gridView.enabled">
+                {showGridButton && (
+                    <ViewModeChangeButton
+                        className="bdl-SubHeaderRight-ViewModeChangeButton"
+                        viewMode={viewMode}
+                        onViewModeChange={onViewModeChange}
+                    />
+                )}
+            </FeatureFlag>
+
             {showSort && !!sortBy && !!sortDirection && (
                 <Sort onSortChange={onSortChange} sortBy={sortBy} sortDirection={sortDirection} />
             )}
--- a/src/elements/common/sub-header/ViewModeChangeButton.js
+++ b/src/elements/common/sub-header/ViewModeChangeButton.js
@@ -33,6 +33,7 @@ const ViewModeChangeButton = ({ className = '', onViewModeChange, intl, viewMode
             }
         >
             <Button
+                data-testid="view-mode-change-button"
                 className={classNames('bdl-ViewModeChangeButton', className)}
                 type="button"
                 onClick={onClick}
--- a/src/elements/common/sub-header/ViewModeChangeButton.js
+++ b/src/elements/common/sub-header/ViewModeChangeButton.js
@@ -1,0 +1,47 @@
+// @flow
+
+import React from 'react';
+import { injectIntl, FormattedMessage } from 'react-intl';
+import classNames from 'classnames';
+
+import Button from '../../../components/button';
+import IconGridViewInverted from '../../../icons/general/IconGridViewInverted';
+import IconListView from '../../../icons/general/IconListView';
+import messages from '../messages';
+import Tooltip from '../Tooltip';
+import type { ViewMode } from '../flowTypes';
+import { VIEW_MODE_GRID, VIEW_MODE_LIST } from '../../../constants';
+import './ViewModeChangeButton.scss';
+
+type Props = {
+    className?: string,
+    onViewModeChange: (viewMode: ViewMode) => void,
+    viewMode: ViewMode,
+} & InjectIntlProvidedProps;
+
+const ViewModeChangeButton = ({ className = '', onViewModeChange, intl, viewMode, ...rest }: Props) => {
+    const isGridView = viewMode === VIEW_MODE_GRID;
+
+    const onClick = () => {
+        onViewModeChange(isGridView ? VIEW_MODE_LIST : VIEW_MODE_GRID);
+    };
+
+    return (
+        <Tooltip
+            text={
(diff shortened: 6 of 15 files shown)


boxcla:

src/elements/common/sub-header/GridViewButton.js
```diff
@@ -0,0 +1,29 @@
...
+import Tooltip from '../Tooltip';
+import './GridViewButton.scss';
+
+type Props = {
+    isGridView: boolean,
+};
+
+const GridViewButton = (props: Props) => {
+    const { isGridView, ...rest } = props;
+    const icon = isGridView ? <IconGridView width={17} height={17} /> : <IconGridViewInverted width={17} height={17} />;
```
jstoffan: Seems like this can be moved inside the button, since the variable is only used once.

src/elements/common/sub-header/GridViewButton.js
```diff
@@ -0,0 +1,29 @@
...
+
+    return (
+        <Tooltip text={<FormattedMessage {...messages.gridView} />}>
+            <Button className="be-btn-grid-view" aria-label={messages.gridView.defaultMessage} type="button" {...rest}>
```
jstoffan: We should be passing each message to `react-intl` so it can be rendered in the user's selected locale. Referencing the `defaultMessage` directly willy likely only show the English copy.

src/elements/common/sub-header/GridViewButton.scss
```diff
@@ -0,0 +1,8 @@
+.btn.be-btn-grid-view {
```
jstoffan: New components should use the SUIT naming convention.

src/elements/common/sub-header/SubHeader.js
```diff
@@ -50,7 +54,9 @@ const SubHeader = ({
             canCreateNewFolder={canCreateNewFolder}
             canUpload={canUpload}
             currentCollection={currentCollection}
+            isGridView={isGridView}
             onCreate={onCreate}
+            onGridViewSwitch={onGridViewSwitch}
```
jstoffan: We typically use `toggle` for switching between states, so this could be `onGridViewToggle`. However, it could also be modified to support multiple other "views", in case we add more in the future. In that case, something like `onViewChange` might make sense.

src/elements/common/sub-header/SubHeaderRight.js
```diff
@@ -7,35 +7,46 @@
...
+    onGridViewSwitch,
     onSortChange,
+    onUpload,
+    view,
 }: Props) => {
     const { sortBy, sortDirection, items = [] }: Collection = currentCollection;
     const isFolder: boolean = view === VIEW_FOLDER;
     const showSort: boolean = isFolder && items.length > 0;
     const showAdd: boolean = (!!canUpload || !!canCreateNewFolder) && isFolder;
+    const showGridButton: boolean = isFolder && items.length > 0;
```
jstoffan: Should the `items.length > 0` check move to its own local variable, `hasItems`, since it's used more than once?

src/elements/content-explorer/Content.js
```diff
@@ -49,62 +50,73 @@ type Props = {
...
-            />
-        )}
-    </div>
-);
+    rootElement,
+    rootId,
+    tableRef,
+    view,
+}: Props) => {
+    const listView = (
```
jstoffan: We should try to avoid rendering non-dynamic components into variables, when possible. If a render method is getting too complicated, it should probably be broken into sub-components.

src/elements/content-explorer/__tests__/ContentExplorer-test.js
```diff
@@ -39,4 +39,25 @@ describe('elements/content-explorer/ContentExplorer', () => {
...
+                canDownload: false,
+                canDelete: false,
+                canUpload: false,
+                canRename: false,
+                canShare: false,
+                canPreview: false,
+                canSetShareAccess: false,
+                canCreateNewFolder: false,
+            });
+            expect(wrapper).toMatchSnapshot();
```
jstoffan: Changes to snapshots at such a high level are often ignored, reducing their value. Ideally, tests should assert specific expectations and only use snapshots where they can't be avoided.

test/integration/content-explorer/ContentExplorer.int.test.js
```diff
@@ -195,4 +212,27 @@ describe('ContentExplorer', () => {
             helpers.checkRowSelections(4);
         });
     });
+
+    describe('Grid View', () => {
+        it('Should initially show list view', () => {
+            helpers.load({ features: gridViewOn });
```
jstoffan: It looks like this is duplicated. Can it be moved to a `beforeEach`?

examples/src/features.js
```diff
@@ -12,4 +12,9 @@ export default {
             enabled: false,
         },
     },
+    contentExplorer: {
+        gridView: {
+            buttonVisible: false,
```
ConradJChan: maybe just `enabled`?

src/elements/common/messages.js
```diff
@@ -122,6 +122,11 @@ const messages = defineMessages({
         description: 'Label for add action',
         defaultMessage: 'Add',
     },
+    gridView: {
+        id: 'be.gridView',
+        description: 'Label for grid view action',
+        defaultMessage: 'Grid View',
```
ConradJChan: Should it be "Switch to Grid View"?

src/elements/common/sub-header/GridViewButton.js
ConradJChan: Do we need a corresponding message for switching back to list view once grid view is toggled?

test/explorer.html
```diff
@@ -74,6 +74,13 @@ <h1>Content Explorer (no folder upload)</h1>
                 const { ContentExplorer } = Box;
                 const token = document.querySelector('.token').value || localStorage.getItem('token');
                 const folderId = document.querySelector('.folder').value || localStorage.getItem('folder');
+                const features = {
```
ConradJChan: Is the goal to keep the grid view hidden? If that's the case, then I think leaving the features object out entirely should keep it hidden. Though maybe you could add a separate explorer example with the feature turned on?

src/elements/common/sub-header/GridViewButton.js
```diff
@@ -0,0 +1,52 @@
...
+    viewMode: ViewMode,
+} & InjectIntlProvidedProps;
+
+const GridViewButton = (props: Props) => {
+    const { onViewModeChange, intl, viewMode, ...rest } = props;
+
+    const isGridView = viewMode === VIEW_MODE_GRID;
+
+    const onClick = () => {
+        if (isGridView) {
```
ConradJChan: can convert this to a one-liner: `onViewModeChange(isGridView ? VIEW_MODE_LIST : VIEW_MODE_GRID`

src/elements/common/sub-header/GridViewButton.scss
```diff
@@ -0,0 +1,8 @@
+.bdl-GridViewButton {
+    margin-left: 7px;
```
ConradJChan: Does this margin need to be baked into the Button component? Seems like it would be a style more suited for the consumer of this button

src/elements/content-explorer/Content.js
```diff
@@ -46,65 +47,79 @@ type Props = {
...
-    </div>
-);
+    rootElement,
+    rootId,
+    tableRef,
+    view,
+    viewMode = VIEW_MODE_LIST,
+}: Props) => {
+    let content;
+    switch (viewMode) {
```
ConradJChan: The switch statement seems a bit unnecessary, since we're only dealing with two cases, could we do an inline ternary operator instead on L119?

src/elements/common/sub-header/GridViewButton.js
```diff
@@ -0,0 +1,52 @@
...
+import { VIEW_MODE_GRID, VIEW_MODE_LIST } from '../../../constants';
+import './GridViewButton.scss';
+
+type Props = {
+    onViewModeChange: Function,
+    viewMode: ViewMode,
+} & InjectIntlProvidedProps;
+
+const GridViewButton = (props: Props) => {
+    const { onViewModeChange, intl, viewMode, ...rest } = props;
```
jstoffan: This can be done inline with the function declaration on line 20.

src/elements/content-explorer/Content.js
jstoffan: That would end up being a nested ternary, which is restricted by our linter, iirc. I think this is usually a sign that a component is doing too much and should probably be broken up. If that's too big a change, though, we could also create some simple state helper variables and use logical AND (`&&`), instead:

```
const isEmpty = isEmpty(view, currentcollation);
const isListView = viewMode === VIEW_MODE_LIST;
...
{isEmpty && <EmptyState view={view} isLoading={currentCollection.percentLoaded !== 100} />}
{!isEmpty && isListView && <ItemList .../>
{!isEmpty && !isListView && <GridView .../>
```

src/elements/common/flowTypes.js
```diff
@@ -0,0 +1,4 @@
+// @flow
+import { VIEW_MODE_GRID, VIEW_MODE_LIST } from '../../constants';
+
+export type ViewMode = typeof VIEW_MODE_GRID | typeof VIEW_MODE_LIST;
```
jstoffan: Should this be in `/elements/common`? It seems to be used only by ContentExplorer at the moment.

src/elements/content-explorer/Content.js
```diff
@@ -46,65 +47,79 @@ type Props = {
...
+        case VIEW_MODE_GRID:
+            content = <div className="bdl-GridView">Placeholder for grid view</div>;
```
jstoffan: It seems like the props passed to ItemList and GridView may end up being very similar. Would something like this work?

```
const View = viewMode === VIEW_MODE_LIST ? ItemList : GridView;
...
<View 
    isSmall={isSmall}
    isMedium={isMedium}
    ...etc
/>
```

src/elements/common/flowTypes.js
rickstefanik: There has been some discussion if the gallery/grid view will ever make its way to the ContentPicker, so I put this in `/elements/common` for that reason.  Currently it is only being used in ContentExplorer, so does it make sense to move it to the `/elements/content-explorer` folder?

src/elements/content-explorer/__tests__/ContentExplorer-test.js
```diff
@@ -39,4 +39,11 @@ describe('elements/content-explorer/ContentExplorer', () => {
             expect(instance.fetchFolder).toHaveBeenCalledWith('123', false);
         });
     });
+
+    describe('render()', () => {
+        test('should render correctly with default props', () => {
+            const wrapper = getWrapper({});
+            expect(wrapper).toMatchSnapshot();
```
jstoffan: Reposting my prior comment:

>Changes to snapshots at such a high level are often ignored, reducing their value. Ideally, tests should assert specific expectations and only use snapshots where they can't be avoided.

src/elements/common/flowTypes.js
jstoffan: Fair enough!

src/elements/content-explorer/__tests__/ContentExplorer-test.js
rickstefanik: Apologies, I thought your prior comment only pertained to the snapshot test with specific props. I'll take out this snapshot too. I'll also read up more on using jest and look deeper into our codebase to better understand how we test higher level components.

src/elements/content-explorer/Content.js
rickstefanik: That would work, but GridView (and perhaps some additional view) may need to be wrapped in an `AutoSizer` or some other component.  I think to build for the future and not let things get too hairy, I'll opt for doing a `ContentView` component.

src/elements/common/sub-header/GridViewButton.js
```diff
@@ -0,0 +1,52 @@
...
+import type { ViewMode } from '../flowTypes';
+import { VIEW_MODE_GRID, VIEW_MODE_LIST } from '../../../constants';
+import './GridViewButton.scss';
+
+type Props = {
+    onViewModeChange: Function,
+    viewMode: ViewMode,
+} & InjectIntlProvidedProps;
+
+const GridViewButton = (props: Props) => {
```
jeremypress: Rename to something more generic since this handles both modes of viewing content in the explorer/picker?

src/elements/common/sub-header/ViewModeChangeButton.js
```diff
@@ -0,0 +1,48 @@
...
+import IconListView from '../../../icons/general/IconListView';
+import messages from '../messages';
+import Tooltip from '../Tooltip';
+import type { ViewMode } from '../flowTypes';
+import { VIEW_MODE_GRID, VIEW_MODE_LIST } from '../../../constants';
+import './ViewModeChangeButton.scss';
+
+type Props = {
+    className?: string,
+    onViewModeChange: Function,
```
jstoffan: We usually want to specify the types of the parameters and return value of the function.

src/elements/common/sub-header/ViewModeChangeButton.js
```diff
@@ -0,0 +1,48 @@
...
+
+    return (
+        <Tooltip
+            text={
+                isGridView ? <FormattedMessage {...messages.listView} /> : <FormattedMessage {...messages.gridView} />
+            }
+        >
+            <Button
+                className={classNames('bdl-ViewModeChangeButton', className)}
+                aria-label={isGridView ? intl.formatMessage(messages.listView) : intl.formatMessage(messages.gridView)}
```
jstoffan: It looks like `Tooltip` adds `aria-describedby` to its child component, so this is probably not needed.

src/elements/content-explorer/ContentView.js
```diff
@@ -0,0 +1,119 @@
...
+    onItemRename,
+    onItemSelect,
+    onItemShare,
+    onSortChange,
+    rootElement,
+    rootId,
+    slotRenderer,
+    tableRef,
+    view,
+    viewMode,
```
jstoffan: Consider making use of `{...rest}` to avoid repeating common/shared passed-through props both here and in the parent.

src/elements/content-explorer/__tests__/ContentExplorer-test.js
```diff
@@ -39,4 +40,18 @@ describe('elements/content-explorer/ContentExplorer', () => {
             expect(instance.fetchFolder).toHaveBeenCalledWith('123', false);
         });
     });
+
+    describe('changeViewMode()', () => {
+        test('should initially be list view', () => {
+            const wrapper = getWrapper({});
```
jstoffan: You could update `getWrapper`'s `props` param to default to an empty object to avoid passing one each time.

src/elements/common/sub-header/SubHeaderRight.js
```diff
@@ -17,25 +20,41 @@ type Props = {
...
+    const showSort: boolean = isFolder && hasItems;
     const showAdd: boolean = (!!canUpload || !!canCreateNewFolder) && isFolder;
+    const showGridButton: boolean = isFolder && hasItems;
 
     return (
         <div className="be-sub-header-right">
+            <FeatureFlag feature="contentExplorer.gridView.enabled">
+                {showGridButton && (
+                    <ViewModeChangeButton
+                        className="bdl-SubHeaderRight-ViewModeChangeButton"
```
ConradJChan: Is this className used for any styling?

src/elements/content-explorer/Content.js
```diff
@@ -44,67 +48,46 @@ type Props = {
...
-                onItemDownload={onItemDownload}
-                onItemRename={onItemRename}
-                onItemShare={onItemShare}
-                onItemPreview={onItemPreview}
-                onSortChange={onSortChange}
-            />
-        )}
-    </div>
-);
+    slotRenderer = index => <div>{index}</div>,
```
ConradJChan: Is this a placeholder or the real default for the `slotRenderer` of content explorer?

test/explorer.html
```diff
@@ -147,11 +151,26 @@ <h1>Content Explorer (no folder upload)</h1>
 
                 const explorer3 = new ContentExplorer();
                 document.querySelector('.explorer3').innerHTML = '';
-                explorer2.show(folderId, token, {
+                explorer3.show(folderId, token, {
```
ConradJChan: nice good catch

src/elements/content-explorer/Content.js
rickstefanik: I was just using that as a placeholder so something would render to the screen for this PR.  That default will not be there in the final version of the code.  Is that ok to keep, or should I remove it?

src/elements/common/sub-header/ViewModeChangeButton.js
rickstefanik: I looked a little further into this, and `Tooltip` only adds `aria-describedby` to its child component when the tooltip is being shown.  (It also only adds a generic label like `aria-describedby="tooltip206"`) . At least for Cypress testing, I think we might need a specific aria-label.

jstoffan: The tooltip should be shown on focus and hover, which is also when it'll be read for screen readers. The `aria-describedby` attribute is meant to associate the element with another element using the latter's unique `id`, so it seems like it's working as we intend.

PR #1350, opened 2019-06-29
fix(pill-selector): Allow copy pasting line separated values

jpfranco:
This update will change the `PillSelector` element from an `input` to a `textarea`. This is needed in order to be able to process newline characters which in turn allows the user to paste a list of line-separated values and get them parsed into pills.

This PR also includes a fix for an issue in which you had to `tab` twice in order to get the input to focus.


boxcla:

src/components/pill-selector-dropdown/PillSelector.js
```diff
@@ -181,14 +181,13 @@ class PillSelector extends React.Component<Props, State> {
...
             <Tooltip isShown={!!error} text={error || ''} position="middle-right" theme="error">
+                {/* eslint-disable-next-line jsx-a11y/no-static-element-interactions */}
                 <span
                     className={classes}
                     onBlur={this.handleBlur}
                     onClick={this.handleClick}
                     onFocus={this.handleFocus}
                     onKeyDown={this.handleKeyDown}
-                    role="button"
-                    tabIndex={0}
```
jpfranco: This was causing tabbing to require a second keystroke before the input got focus. I did some digging and it seems like the most likely reason that this was added was to avoid the eslint warning above.

src/components/pill-selector-dropdown/PillSelector.scss
```diff
@@ -91,8 +91,7 @@
             }
         }
 
-        .pill-selector-input,
-        input[type='text'].pill-selector-input {
```
jpfranco: Couldn't find a reason why `input[type='text']` was added here, looks like `.pill-selector-input` should've been enough

src/components/pill-selector-dropdown/PillSelector.scss
```diff
@@ -104,6 +103,8 @@
             min-width: 0;
             overflow: hidden;
             padding: 5px 3px 3px;
+            resize: none;
+            white-space: nowrap;
```
jpfranco: This will prevent the input from wrapping to multiple lines so the behavior will be identical to the text input.

PR #1351, opened 2019-06-29
fix(versions): Restrict preview for watermarked files

jstoffan:
This is partial revert of #1342 due to another tweak to the requirements for watermark previews. I also took the opportunity to centralize the logic we use to determine what actions can be performed on a given version.


boxcla:

src/elements/content-sidebar/versions/VersionsItem.js
```diff
@@ -118,27 +131,28 @@ const VersionsItem = ({
                         {!!size && <span className="bcs-VersionsItem-size">{versionSize}</span>}
                     </div>
 
-                    {isLimited && (
+                    {isLimited && hasActions && (
```
ConradJChan: Is this additional check for `hasActions` necessary? Would we want to show the message even if there are no actions?

jstoffan: Good question, but no. We only want to show the message if the user would have had actions otherwise. I've validated this with product.

ConradJChan: 👍

PR #1353, opened 2019-07-01
fix(task-form): Update error message when leaving invalid required field

booc0mtaco:


boxcla:

PR #1354, opened 2019-07-01
fix(usm): Do not pass itemTypedID to getContactsByEmail

baipenghan:


boxcla:

PR #1357, opened 2019-07-02
fix(pill-selector-dropdown): disable pills when PillSelector is disabled

tjgupta:
Screen Shot 2019-07-02 at 10 19 33 AM

### DIFF
--- a/src/components/pill-selector-dropdown/Pill.js
+++ b/src/components/pill-selector-dropdown/Pill.js
@@ -11,19 +11,17 @@ type Props = {
 };
 
 const Pill = ({ isDisabled = false, isSelected = false, isValid = true, onRemove, text }: Props) => {
-    const removeFn = isDisabled ? undefined : onRemove;
+    const styles = classNames('pill', {
+        'is-selected': isSelected && !isDisabled,
+        'is-invalid': !isValid,
+        'is-disabled': isDisabled,
+    });
     return (
-        <span
-            className={classNames('pill', {
-                'is-selected': isSelected,
-                'is-invalid': !isValid,
-                'is-disabled': isDisabled,
-            })}
-        >
+        <span className={styles}>
             <span className="pill-text">{text}</span>
-            <span aria-hidden="true" className="close-btn" onClick={removeFn}>
+            <button aria-hidden="true" className="close-btn" disabled={isDisabled} onClick={onRemove} type="button">
                 ✕
-            </span>
+            </button>
         </span>
     );
 };
--- a/src/components/pill-selector-dropdown/Pill.js
+++ b/src/components/pill-selector-dropdown/Pill.js
@@ -3,19 +3,29 @@ import React from 'react';
 import classNames from 'classnames';
 
 type Props = {
+    isDisabled?: boolean,
     isSelected?: boolean,
     isValid?: boolean,
     onRemove: Function,
     text: string,
 };
 
-const Pill = ({ isSelected = false, isValid = true, onRemove, text }: Props) => (
-    <span className={classNames('pill', { 'is-selected': isSelected, 'is-invalid': !isValid })}>
-        <span className="pill-text">{text}</span>
-        <span aria-hidden="true" className="close-btn" onClick={onRemove}>
-            ✕
+const Pill = ({ isDisabled = false, isSelected = false, isValid = true, onRemove, text }: Props) => {
+    const removeFn = isDisabled ? undefined : onRemove;
+    return (
+        <span
+            className={classNames('pill', {
+                'is-selected': isSelected,
+                'is-invalid': !isValid,
+                'is-disabled': isDisabled,
+            })}
+        >
+            <span className="pill-text">{text}</span>
+            <span aria-hidden="true" className="close-btn" onClick={removeFn}>
+                ✕
+            </span>
         </span>
-    </span>
-);
+    );
+};
 
 export default Pill;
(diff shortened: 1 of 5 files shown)


boxcla:

src/components/pill-selector-dropdown/Pill.js
```diff
@@ -3,19 +3,29 @@ import React from 'react';
...
+        <span
+            className={classNames('pill', {
+                'is-selected': isSelected,
+                'is-invalid': !isValid,
+                'is-disabled': isDisabled,
+            })}
+        >
+            <span className="pill-text">{text}</span>
+            <span aria-hidden="true" className="close-btn" onClick={removeFn}>
+                ✕
```
booc0mtaco: Wondering if we should use the proper [HTML Entity](https://dev.w3.org/html5/html-author/charref) here to avoid encoding issues.

`{'&cross;'}` or however you do it in react?

src/components/pill-selector-dropdown/Pill.js
```diff
@@ -3,19 +3,29 @@ import React from 'react';
...
-const Pill = ({ isSelected = false, isValid = true, onRemove, text }: Props) => (
-    <span className={classNames('pill', { 'is-selected': isSelected, 'is-invalid': !isValid })}>
-        <span className="pill-text">{text}</span>
-        <span aria-hidden="true" className="close-btn" onClick={onRemove}>
-            ✕
+const Pill = ({ isDisabled = false, isSelected = false, isValid = true, onRemove, text }: Props) => {
+    const removeFn = isDisabled ? undefined : onRemove;
+    return (
+        <span
+            className={classNames('pill', {
```
priyajeet: Let's move this out now to a const now for breakpoints

src/components/pill-selector-dropdown/Pill.js
```diff
@@ -3,19 +3,29 @@ import React from 'react';
...
+                'is-selected': isSelected,
```
jstoffan: Can a pill be both disabled and selected at the same time or should they be mutually-exclusive?

src/components/pill-selector-dropdown/Pill.js
```diff
@@ -3,19 +3,29 @@ import React from 'react';
...
+                'is-invalid': !isValid,
+                'is-disabled': isDisabled,
+            })}
+        >
+            <span className="pill-text">{text}</span>
+            <span aria-hidden="true" className="close-btn" onClick={removeFn}>
```
jstoffan: I realize this was already here, but using a span here seems a little sketchy on the semantics side of things.

src/components/pill-selector-dropdown/Pill.js
tjgupta: Hmm, `&cross;` doesn't look very good, and neither does `&times;`. In our modals we use an icon, but I'm not sure if that's what we want here too?

This is with `&cross;`:
Screen Shot 2019-07-02 at 10 42 33 AM

src/components/pill-selector-dropdown/Pill.js
tjgupta: I'm not sure I follow what you're asking...

src/components/pill-selector-dropdown/Pill.js
tjgupta: True, I'll update.

src/components/pill-selector-dropdown/Pill.js
tjgupta: It looks like `is-selected` is used for keyboard accessibility of the pills, so if you for example hit the left arrow it will select the previous pill. However, when the PillSelector is disabled, these keyboard interactions won't work because the whole textarea is disabled, so by nature of the UI they seem to be mutually exclusive. I'm not sure we gain much by adding that complexity into the code though. It will not set `is-selected` if it's disabled because there is no way to select.

jstoffan: I see. I think each component should make sure its internal state (in this case expressed via CSS classes) is consistent and correct, though. I think we basically just need to add `'is-selected': isSelected && !isDisabled,`, which I'm not sure adds much in the way of complexity. It's up to you, though.

src/components/pill-selector-dropdown/Pill.js
priyajeet: yeah I would leave it at X for now

src/components/pill-selector-dropdown/Pill.js
priyajeet: const styles = classNames()
<span className={styles} >

src/components/pill-selector-dropdown/Pill.js
```diff
@@ -3,19 +3,27 @@ import React from 'react';
...
+const Pill = ({ isDisabled = false, isSelected = false, isValid = true, onRemove, text }: Props) => {
+    const styles = classNames('pill', {
+        'is-selected': isSelected && !isDisabled,
+        'is-invalid': !isValid,
+        'is-disabled': isDisabled,
+    });
+    return (
+        <span className={styles}>
+            <span className="pill-text">{text}</span>
+            <button aria-hidden="true" className="close-btn" disabled={isDisabled} onClick={onRemove} type="button">
```
jstoffan: Just a heads-up: I think `PlainButton` handles disabled behavior a bit differently, via `event.preventDefault()`, though I'm not 100% sure why. Browsers also may style this button differently when it's disabled.

tjgupta: Yep I tried to use `PlainButton` at first, but it required a lot more CSS overrides to get it to look visually the same as it currently does. I tested this with `disabled={true}` and it seemed to work as expected (prevented removal of the pill).

PR #1358, opened 2019-07-02
fix(comment): padding and a11y adjustments for comment UI

booc0mtaco:
- set default button to "cancel" on confirm mini-modal is open
- allow for escape to close when confirm mini-modal is open
- use consistent button padding/margin so if buttons stack they have space between (and sync up spacing with other buttons)

### DIFF
--- a/src/elements/content-sidebar/activity-feed/comment/Comment.scss
+++ b/src/elements/content-sidebar/activity-feed/comment/Comment.scss
@@ -18,11 +18,11 @@
 
     .bcs-comment-confirm-delete,
     .bcs-comment-confirm-cancel {
-        margin: 0;
+        margin-left: 0;
     }
 
     .bcs-comment-confirm-cancel {
-        margin-right: 8px;
+        margin-right: 10px;
     }
 }
 
--- a/src/elements/content-sidebar/activity-feed/comment/Comment.js
+++ b/src/elements/content-sidebar/activity-feed/comment/Comment.js
@@ -206,6 +206,7 @@ class Comment extends React.Component<Props, State> {
                                             onKeyDown={this.onKeyDown}
                                             shouldOutlineFocus={false}
                                             shouldDefaultFocus
+                                            role="dialog"
                                         >
                                             <div className="bcs-comment-confirm-prompt">
                                                 <FormattedMessage {...deleteConfirmMessage} />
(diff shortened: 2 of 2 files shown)


boxcla:

booc0mtaco:
demo

src/elements/content-sidebar/activity-feed/comment/Comment.scss
```diff
@@ -18,11 +18,11 @@
 
     .bcs-comment-confirm-delete,
     .bcs-comment-confirm-cancel {
-        margin: 0;
+        margin-left: 0;
```
alexkrolick: Side note: we really need a ButtonGroup component that does this automatically

jstoffan:
Is the content of the flyout read aloud to users using screenreaders with this change? Or without?

src/elements/content-sidebar/activity-feed/comment/Comment.scss
jstoffan: s it still be considered a group if the buttons are visually separated? If so, you're welcome to extend the [ButtonGroup component](http://opensource.box.com/box-ui-elements/#/Components?id=buttongroup) to support it.

booc0mtaco: This isn't really a fix because the buttons aren't grouped. The design called for a narrow window, and the button margins pushed them to flow. A design without this outer constraint would allow us to remove this `margin` override entirely.

[`ButtonGroup`](https://opensource.box.com/box-ui-elements/#!/ButtonGroup) makes sense currently, and these two are associated with each other in the same way two words in a sentence are. A `FlexLayout` component of some sort might be a better direction to go in (sorta like the customize toolbar option in macOS). we could compose the layout with buttons together that way, and repurpose the layout if desired. The first consumer of that would be our other modal components, header bars, etc.

alexkrolick: Good point, it's more like this: https://bulma.io/documentation/elements/button/#list-of-buttons

(also Bootstrap/Zurb seem to have removed default horizontal margins from their buttons, so they are left-aligned by default)

booc0mtaco:
@jstoffan we didn't handle this before. I'll add something to treat it as an alert.

PR #1359, opened 2019-07-02
fix(sidebar): Add flex-shrink 0 to comment input

ConradJChan:


boxcla:

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
```diff
@@ -86,6 +86,7 @@
 
     .bcs-activity-feed-comment-input {
         background-color: $white;
+        flex-shrink: 0;
```
booc0mtaco: What are are the other two `flex` values here? does it inherit `flex-grow` and `flex-basis` from somewhere else already? Should we use the shorthand syntax to be more deliberate?

ConradJChan: Good point, let me fix it to be more clear

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
```diff
@@ -86,7 +86,7 @@
 
     .bcs-activity-feed-comment-input {
         background-color: $white;
-        flex-shrink: 0;
+        flex: 0 0 auto;
```
jstoffan: Does the comment box still grow/shrink as expected given min/max height set below? It definitely looks like it's meant to change size.

ConradJChan: Yeah it does, the flex-grow: 0 specifies that it won't take up any of the remaining space in the flex container but will still reflect it's natural height

PR #1362, opened 2019-07-03
fix(button): remove horizontal margin from buttons

alexkrolick:
Changes buttons to not have a default horizontal margin, except between other buttons. The current 5px margin is almost always overridden when used.

### DIFF
--- a/src/styles/common/_buttons.scss
+++ b/src/styles/common/_buttons.scss
@@ -82,7 +82,7 @@
     }
 
     & + & {
-        margin-left: 5px;
+        margin-left: 10px;
     }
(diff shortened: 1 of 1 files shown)


boxcla:

src/styles/common/_buttons.scss
```diff
@@ -80,6 +80,10 @@
         box-shadow: none;
         opacity: .4;
     }
+
+    & + & {
+        margin-left: 5px;
```
jstoffan: It seems like this should be `10px`, since the left/right margins were additive.

PR #1364, opened 2019-07-03
fix(pills): Fix vertical centering of pill selector input text

jstoffan:
Before
tasks_before

After
tasks_after


boxcla:

PR #1365, opened 2019-07-04
fix(versions): Fix version groups collapsing prior months/years

jstoffan:
Currently, prior months of the current year (June, May, etc.) collapse into the most recent prior month (June, at the moment); likewise for prior years. This change makes it so each prior month and year has its own group and heading while retaining the order of the versions array.


boxcla:

PR #1366, opened 2019-07-04
fix(sidebar): Set dirty state if the sidebar was forced open

jstoffan:


boxcla:

PR #1367, opened 2019-07-05
fix(pill-selector-dropdown): Fix for pills getting removed on click

jpfranco:
pill selector bug

Switching the `Pill` close button from a `span` to a `button` was causing the close button on the first pill to be automatically clicked when clicking on the parent container:

https://github.com/box/box-ui-elements/pull/1357/files#diff-5fa22640b1541d06041959dd58e56b3cR22

I've not been able to figure out why this happens yet, but this is a workaround that we can use while the root cause is identified. Switching back to `span` prevents the issue from happening.


boxcla:

src/components/pill-selector-dropdown/Pill.js
```diff
@@ -16,12 +17,14 @@ const Pill = ({ isDisabled = false, isSelected = false, isValid = true, onRemove
         'is-invalid': !isValid,
         'is-disabled': isDisabled,
     });
+    const onClick = isDisabled ? noop : onRemove;
+
     return (
         <span className={styles}>
             <span className="pill-text">{text}</span>
-            <button aria-hidden="true" className="close-btn" disabled={isDisabled} onClick={onRemove} type="button">
+            <span aria-hidden="true" className="close-btn" onClick={onClick}>
```
jstoffan: This element should use something other than a `span`, as they're not intended to be interactive and carry little semantic meaning.

jpfranco: Agreed but I think the best candidate is `button` which is the element that introduced the bug. There's a deeper issue that's causing this element to get clicked when the parent is clicked, but I haven't been able to figure it out yet. This update is just meant to be a temporary workaround for now

jstoffan: Fair enough. Let's come back to it in the future.

PR #1371, opened 2019-07-08
build: perf improvements and fixes

priyajeet:
1. Only builds the explorer entry point that includes explorer, preview, sidebar and open with. Does not include content picker so we may need another job for that but that can be added later.
2. Fixes an issue in circle config that was preventing PR builds
3. Combines babel and webpack build for circle
4. Removes a couple of steps from pre-push hook

Should show circle under checks tab now

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "scripts": {
         "build": "yarn build:dev:dist",
         "build:ci:es": "yarn build:prod:es",
-        "build:ci:dist": "LANGUAGE=en-US REACT=true yarn build:prod:dist",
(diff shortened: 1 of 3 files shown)


boxcla:

package.json
```diff
@@ -38,7 +38,7 @@
     "scripts": {
         "build": "yarn build:dev:dist",
         "build:ci:es": "yarn build:prod:es",
-        "build:ci:dist": "LANGUAGE=en-US REACT=true yarn build:prod:dist",
+        "build:ci:dist": "LANGUAGE=en-US ENTRY=explorer REACT=true yarn build:prod:dist",
```
j3tan: revert?

priyajeet: nope

PR #1372, opened 2019-07-08
feat(content-explorer): Show grid with icons

rickstefanik:
Now displaying grid with functionality and generic icons. (Thumbnails of files will come in later PR.)


boxcla:

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -1235,6 +1307,24 @@ class ContentExplorer extends Component<Props, State> {
...
+    /**
+     * Determine the maximum number of columns based on the
+     * current width of the ContentExplorer.
+     *
+     * @private
+     * @return {number} - max number of columns
+     */
+    getMaxColumns = (): number => {
+        const clientWidth = getProp(this.contentExplorerRef, 'current.clientWidth', 0);
+        if (clientWidth < 700) {
```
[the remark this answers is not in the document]
rickstefanik: Do you mean should the user be able to configure the number of columns displayed? A slider to set the number of columns is a feature coming in the future.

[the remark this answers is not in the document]
rickstefanik: Or do you mean the breakpoints themselves should be configurable?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -1225,6 +1240,63 @@ class ContentExplorer extends Component<Props, State> {
...
+            this.select,
+            canPreview,
+            isSmall, // shows details if false
+            isTouch,
+        );
+
+        const dateCell = dateCellRenderer();
+
+        return (
+            <div>
```
ConradJChan: This seems like a good candidate for its own component

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -140,6 +150,8 @@ class ContentExplorer extends Component<Props, State> {
 
     table: any;
 
+    contentExplorerRef: refType<HTMLElement>;
```
ConradJChan: I see in other places we define the flow type inline such as https://github.com/box/box-ui-elements/blob/15326176688e562e3bb4e7c3077f7d66be40f348/src/features/metadata-instance-editor/Instance.js#L427

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -220,7 +232,10 @@ class ContentExplorer extends Component<Props, State> {
 
         this.id = uniqueid('bce_');
 
+        this.contentExplorerRef = React.createRef();
+
         this.state = {
+            columnCount: 5,
```
ConradJChan: Try to avoid magic numbers and instead create a meaningful constant, something like `DEFAULT_COLUMN_COUNT`

src/components/grid-view/GridView.scss
```diff
@@ -18,16 +17,18 @@
     }
 
     &.bdl-GridView--columns-3 .bdl-GridViewSlot {
-        $columns3prct: calc(33.3333% - 6.6666px); /* 100%/3 - 20px/3 */
+        $columns3prct: calc(33.3333% - 10px); /* 100%/3 - 20px/3 */
 
         flex-basis: $columns3prct;
```
ConradJChan: Can we use the `flex` rule shorthand so that we can be explicit about what the flex-grow, flex-shrink values are as well? We've had flexbox issues related to that bite us in the past

src/components/grid-view/GridViewSlot.js
```diff
@@ -1,16 +1,13 @@
...
 import * as React from 'react';
+import './GridViewSlot.scss';
 
 type Props = {
     slotIndex: number,
-    slotRenderer: ?(slotIndex: number) => React.Element<any>,
+    slotRenderer: (slotIndex: number) => React.Element<any>,
 };
 
 function GridViewSlot({ slotIndex, slotRenderer }: Props) {
```
ConradJChan: Perhaps default the `slotRenderer` to lodash/noop to be safe?

src/constants.js
```diff
@@ -318,6 +318,10 @@ export const SIDEBAR_VIEW_METADATA: 'metadata' = 'metadata';
 export const SIDEBAR_VIEW_ACTIVITY: 'activity' = 'activity';
 export const SIDEBAR_VIEW_VERSIONS: 'versions' = 'versions';
 
+/* ------------------ Content Explorer Grid View ---------------------- */
```
jstoffan: We haven't been great about it historically, but we should try to keep constants, types, and utils as close to their components as possible. It doesn't seem like this is meant to be used by any other component/element.

src/elements/content-explorer/GridViewItem.scss
```diff
@@ -0,0 +1,19 @@
+.bdl-GridViewItem-itemThumbnail {
+    margin-bottom: 10px;
+    padding-bottom: 65%;
```
jstoffan: We should avoid relative values for spacing, as they will not be consistent across viewport sizes.

src/elements/content-explorer/GridViewItem.scss
```diff
@@ -0,0 +1,19 @@
...
+    position: relative;
+
+    .bdl-GridViewItem-itemIcon {
+        align-items: center;
+        bottom: 0;
+        display: flex;
+        height: 50%;
```
jstoffan: We should avoid using relative dimensions when the parent doesn't have set dimensions. This icon should have a fixed size and be positioned via flexbox.

src/elements/content-explorer/GridViewItem.js
```diff
@@ -0,0 +1,94 @@
...
+
+    const dateCell = dateCellRenderer();
+
+    return (
+        <div>
+            <div className="bdl-GridViewItem-itemThumbnail">
+                <div className="bdl-GridViewItem-itemIcon"> {getIcon(128, item)} </div>
+            </div>
+            <div>
+                {nameCell({ rowData: item })}
```
jstoffan: Can these be invoked as regular components rather than as functions, like `<NameCell />`?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -1225,6 +1237,40 @@ class ContentExplorer extends Component<Props, State> {
...
+            <GridViewItem
+                canDelete={canDelete}
+                canDownload={canDownload}
+                canPreview={canPreview}
+                canRename={canRename}
+                canShare={canShare}
+                isSmall={isSmall}
+                isTouch={isTouch}
+                item={item}
+                onItemClick={this.onItemClick}
```
jstoffan: It looks like some of these props are duplicated from those passed into `Content`. Can `Content` pass the props down to its children directly?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -1225,6 +1237,40 @@ class ContentExplorer extends Component<Props, State> {
         this.setState({ currentOffset: newOffset }, this.refreshCollection);
     };
 
+    /**
+     * Renderer used for cards in grid view
+     *
+     * @param {number} slotIndex - index of item in currentCollection.items
+     * @return {React.Element} - Element to display in card
+     */
+    slotRenderer = (slotIndex: number) => {
```
jstoffan: This should probably live in a GridView-specific component.

src/components/grid-view/GridView.scss
```diff
@@ -2,32 +2,41 @@
...
     }
 
-    &.bdl-GridView--columns-3 .bdl-GridViewSlot {
-        $columns3prct: calc(33.3333% - 6.6666px); /* 100%/3 - 20px/3 */
+    &.bdl-GridView--columns-3 .bdl-GridView-slot {
+        $columns3prct: calc(33.3333% - 10px); /* 100%/3 - 10px */
 
-        flex-basis: $columns3prct;
+        flex: 0 1 $columns3prct;
+        margin-right: 10px;
```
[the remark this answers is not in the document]
rickstefanik: I had to include this as a separate rule, not inside `.bdl-GridView`, which may or may not have been what you intended.  If I included inside the `bdl-GridView` like this:
<pre>
.bdl-GridView {
    ...
    .bdl-GridView-slot {
        ...
        & + & {
            margin-left: 10px;
        } 
    }
}
</pre>
The rule would be translated to `.bdl-GridView .bdl-GridView-slot + .bdl-GridView .bdl-GridView-slot`.  Since the css is read from right to left, this would not properly select the GridViewSlots as desired.

src/components/grid-view/GridViewWrapper.js
```diff
@@ -0,0 +1,39 @@
...
+
+const GridViewWrapper = ({ width, ...rest }: Props) => {
+    /**
+     * Determine the maximum number of columns based on the
+     * current width of the ContentExplorer.
+     *
+     * @private
+     * @return {number} - max number of columns
+     */
+    const getMaxColumns = (): number => {
```
jstoffan: This probably doesn't need to be a separate function anymore; it can just be calculated directly and passed to `Math.min` below.

src/elements/content-explorer/ItemGrid.js
```diff
@@ -0,0 +1,94 @@
...
+    onItemRename: Function,
+    onItemSelect: Function,
+    onItemShare: Function,
+    rootElement?: HTMLElement,
+    rootId: string,
+    view: View,
+};
+
+const ItemGrid = ({
+    canDelete,
```
jstoffan: You should be able to make use of `...rest` here, since the props are all defined in the Props type.

src/elements/content-explorer/ItemGridCell.js
```diff
@@ -0,0 +1,94 @@
...
+    const DateCell = dateCellRenderer();
+
+    return (
+        <figure>
+            <div className="bce-ItemGridCell-itemThumbnail">
+                <div className="bce-ItemGridCell-itemIcon"> {getIcon(128, item)} </div>
+            </div>
+            <figcaption>
+                <NameCell rowData={item} />
+                <div> {getSize(item.size)} </div>
```
jstoffan: Are the non-breaking spaces used intentionally here for visual spacing? If so, they should be replaced with CSS padding. If not, they should still be removed to avoid rendering issues; same with `getIcon` above.

rickstefanik:
Still investigating a better way to solve the double-click button bug.  Would like to use `PureComponents` over `shouldComponentUpdate`, but that has been difficult since `item` still needs to be passed into `ItemGridCell` for the callbacks.

examples/src/features.js
```diff
@@ -14,7 +14,7 @@ export default {
     },
     contentExplorer: {
         gridView: {
-            enabled: false,
+            enabled: true,
```
jstoffan: Is this ready to be enabled on the examples site?

src/components/grid-view/GridViewWrapper.js
```diff
@@ -0,0 +1,31 @@
...
+    let maxColumns;
+    if (width < ONE_COLUMN_BREAKPOINT) {
+        maxColumns = 1;
+    } else if (width < THREE_COLUMN_BREAKPOINT) {
+        maxColumns = 3;
+    } else {
+        maxColumns = MAX_COLUMN_COUNT;
+    }
+
+    return <GridView columnCount={Math.min(DEFAULT_COLUMN_COUNT, maxColumns)} width={width} {...rest} />;
```
jstoffan: Is the `Math.min` call or `DEFAULT_COLUMN_COUNT` needed? It looks like `maxColumns` is always a number and less than or equal to `DEFAULT_COLUMN_COUNT`.

src/components/grid-view/GridViewWrapper.js
```diff
@@ -0,0 +1,31 @@
...
+};
+
+const DEFAULT_COLUMN_COUNT = 5;
+const MAX_COLUMN_COUNT = 5;
+
+const ONE_COLUMN_BREAKPOINT = 700;
+const THREE_COLUMN_BREAKPOINT = 1400;
+
+const GridViewWrapper = ({ width, ...rest }: Props) => {
+    let maxColumns;
```
jstoffan: Does this number reflect `maxColumns` or just `columnCount`? Can the number of columns be changed in some other way, such as via CSS media queries or something?

src/elements/content-explorer/ItemGrid.js
```diff
@@ -0,0 +1,56 @@
...
+type Props = {
+    canDelete: boolean,
+    canDownload: boolean,
+    canPreview: boolean,
+    canRename: boolean,
+    canShare: boolean,
+    currentCollection: Collection,
+    isSmall: boolean,
+    isTouch: boolean,
+    onItemClick: Function,
```
jstoffan: It looks like these types are different between components. Do they need to be specified at each level? If so, should they be shared?

src/components/grid-view/GridViewWrapper.js
rickstefanik: My thinking on this is that in the future, `DEFAULT_COLUMN_COUNT` will become a prop (named something like `columnCount`) that is passed in from `GridViewSlider`.  In that scenario, the `Math.min` call would be needed.

src/components/grid-view/GridViewWrapper.js
rickstefanik: For now, `maxColumns` === `columnCount` and this is the only way the number of columns can be changed, but in the future, `columnCount` will be a prop passed in from `GridViewSlider`.  Does it make sense to keep this code that is anticipating the `GridViewSlider` or should I remove it?

src/elements/content-explorer/flowTypes.js
```diff
@@ -0,0 +1,23 @@
...
+    onItemPreview: (item: BoxItem) => void,
+    onItemRename: (item: BoxItem) => void,
+    onItemSelect: (item: BoxItem, callback: Function) => void,
+    onItemShare: (item: BoxItem) => void,
+    rootId: string,
+    view: View,
+};
+
+// eslint-disable-next-line import/prefer-default-export
+export type { ItemGridProps };
```
rickstefanik: I couldn't seem to find the right syntax for doing a default export of a flowType.  All our other `flowTypes` files export more than one type, so they don't use a default export.

src/elements/content-explorer/ItemGrid.js
rickstefanik: I addressed this, but not sure if I did exactly what you wanted.  I took the common props used in `ItemGrid` and `ItemGridCell`, put them in `/src/elements/content-explorer/flowTypes.js`, and then imported that type in `ItemGrid` and `ItemGridCell`.

src/elements/content-explorer/flowTypes.js
rickstefanik: I noticed this previous commit: `feat(preview): better message when blocked by a policy`, ran into same issue, and used the same approach of disabling eslint for that line.

src/components/grid-view/GridView.js
```diff
@@ -54,33 +50,24 @@ class GridView extends React.Component<Props> {
...
             contents.push(
-                <GridViewSlot
-                    key={key}
-                    slotIndex={slotIndex}
-                    slotRenderer={slotIndex < count ? slotRenderer : null}
-                    item={item}
-                    onItemSelect={onItemSelect}
-                />,
+                <div
+                    key={getProp(currentCollection, `items[${slotIndex}].id`, uniqueId('bdl-GridView-slot'))}
```
ConradJChan: Minor, but maybe it would be better to wrap the default value in a anonymous function so that `uniqueId` isn't called each time even if it's not being used

src/components/grid-view/GridView.js
```diff
@@ -112,6 +98,7 @@ class GridView extends React.Component<Props> {
                 width={width}
                 gridClassName="bdl-GridView-body"
                 rowClassName="bdl-GridView-tableRow"
+                scrollToIndex={0}
```
ConradJChan: It'd be an interesting enhancement in the future if we could maybe cache this value when they've scrolled down the gridview so that when they leave that folder and come back, they're at the same position in the grid

src/components/grid-view/GridView.scss
```diff
@@ -34,8 +28,26 @@
...
+}
+
+.bdl-GridView-row {
+    display: flex;
+    justify-content: flex-start;
+}
+
+.bdl-GridView-slot {
+    border: 1px solid $bdl-gray-10;
+    flex: 0 1;
```
ConradJChan: what should the flex-basis value be?

test/integration/content-explorer/ContentExplorer.int.test.js
```diff
@@ -220,15 +218,49 @@ describe('ContentExplorer', () => {
...
+                .click();
+            helpers.getRenameButton().click();
+            cy.getByAriaLabel(localize('be.rename')).should('have.length', 1);
+            helpers.getCancelButton().click();
+            cy.getByAriaLabel(localize('be.rename')).should('have.length', 0);
+        });
+
+        it('Should open and close preview', () => {
+            cy.getByAriaLabel(localize('be.preview')).should('have.length', 0);
+            cy.contains('Sample 3D.box3d').click();
```
jstoffan: Can we instead click the first file in the list to avoid selecting on a specific file name?

test/integration/content-explorer/ContentExplorer.int.test.js
```diff
@@ -220,15 +218,49 @@ describe('ContentExplorer', () => {
...
+                .eq(0)
+                .click();
+            helpers.getRenameButton().click();
+            cy.getByAriaLabel(localize('be.rename')).should('have.length', 1);
+            helpers.getCancelButton().click();
+            cy.getByAriaLabel(localize('be.rename')).should('have.length', 0);
+        });
+
+        it('Should open and close preview', () => {
+            cy.getByAriaLabel(localize('be.preview')).should('have.length', 0);
```
jstoffan: Can we use `.should('not.exist')`, instead of length checks? Also, these seem long enough to deserve an alias:

```
cy.getByAriaLabel(localize('be.preview')).as('previewButton').should('not.exist').
cy.get('@previewButton').should('exist'); // Not strictly necessary, as .get carries an implicit exist assertion
```

src/elements/common/item/Name.js
```diff
@@ -0,0 +1,38 @@
...
+import ItemDetails from './ItemDetails';
+import { VIEW_SEARCH } from '../../../constants';
+
+import './NameCell.scss';
+
+type Props = {
+    canPreview: boolean,
+    isTouch: boolean,
+    item: BoxItem,
+    onItemClick: Function,
```
jstoffan: Can these be specified?

src/elements/content-explorer/MoreOptions.js
```diff
@@ -0,0 +1,131 @@
...
+
+type Props = {
+    canDelete: boolean,
+    canDownload: boolean,
+    canPreview: boolean,
+    canRename: boolean,
+    canShare: boolean,
+    isSmall: boolean,
+    item: BoxItem,
+    onItemDelete: Function,
```
jstoffan: These types seem duplicated and less specific than in other components.

src/elements/content-explorer/MoreOptions.js
```diff
@@ -0,0 +1,131 @@
...
+                <Button
+                    className="bce-btn-more-options"
+                    data-testid="bce-btn-more-options"
+                    onFocus={onFocus}
+                    type="button"
+                >
+                    ···
+                </Button>
+                <Menu>
+                    {allowPreview ? (
```
jstoffan: These can be simplified to use logical and instead of ternaries, such as `allowPreview && <MenuItem>`

test/integration/content-explorer/ContentExplorer.int.test.js
```diff
@@ -220,15 +218,49 @@ describe('ContentExplorer', () => {
...
+        it('Should open and close share modal', () => {
+            cy.getByAriaLabel(localize('be.share')).should('have.length', 0);
+            cy.contains(localize('be.share')).click();
+            cy.getByAriaLabel(localize('be.share')).should('have.length', 1);
+            helpers.getCloseButton().click();
+            cy.getByAriaLabel(localize('be.share')).should('have.length', 0);
+        });
+
+        it('Should open and close rename modal', () => {
+            cy.getByAriaLabel(localize('be.rename')).should('have.length', 0);
```
jstoffan: It's not super clear to me that we're trying to test a modal rather than a button, since the names on everything are so similar. Can we select the modal or modal header itself to make it more clear?

src/components/grid-view/GridView.js
rickstefanik: I think so too. It seems like ListView also always returns to the top of the list, so the feature could be added there as well.

PR #1373, opened 2019-07-08
build(ci): check for generated styles and properties

priyajeet:


boxcla:

PR #1376, opened 2019-07-09
build: prepush should try running tests since master revision

priyajeet:


boxcla:

scripts/prepush.sh
```diff
@@ -45,7 +45,7 @@ prepush() {
     printf "${blue}-------------------------------------------------------------${end}"
     printf "${blue}Testing${end}"
     printf "${blue}-------------------------------------------------------------${end}"
-    yarn test --lastCommit || exit 1
+    yarn test --changedSince=master || exit 1
```
jstoffan: Jest is so good.

priyajeet: unsure tho this is correct

scripts/prepush.sh
```diff
@@ -45,7 +45,7 @@ prepush() {
     printf "${blue}-------------------------------------------------------------${end}"
     printf "${blue}Testing${end}"
     printf "${blue}-------------------------------------------------------------${end}"
-    yarn test --lastCommit || exit 1
+    yarn test --changedSince=upstream/master || exit 1
```
j3tan: add a `git fetch upstream` above this

PR #1377, opened 2019-07-09
 build: fix caching for circle ci

priyajeet:
Fixes an issue where yarn cache was getting bust due to branch/PR name, which was adding 1-2m to the build time.

Also adds a badge


boxcla:

PR #1378, opened 2019-07-09
fix(classification): Classification badge UX tweaks

jpfranco:
Just a few small UI updates for the classification badge:

- Changed "Add Classification" text to "Classify"
- Updated "Classify" button text color for better accessibility
- Capitalized "Classify" for better i18n
- Removed tag icon
- Alignment adjustment

### DIFF
--- a/src/features/classification/ClassifiedBadge.scss
+++ b/src/features/classification/ClassifiedBadge.scss
@@ -13,4 +13,5 @@
     @include bdl-ClassificationBadge-name;
 
     color: $bdl-gray !important; // To inhibit overriding
+    text-transform: uppercase;
 }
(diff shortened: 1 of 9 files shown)


boxcla:

src/features/classification/ClassifiedBadge.scss
```diff
@@ -13,4 +13,5 @@
     @include bdl-ClassificationBadge-name;
 
     color: $bdl-gray !important; // To inhibit overriding
+    text-transform: uppercase;
```
jstoffan: I was under the impression we wanted to avoid this due to accessibility issues.

jpfranco: Yes, this was moved from here:
https://github.com/box/box-ui-elements/pull/1378/files#diff-ee5711bf1ff3f3fcd869e4de1f5c6419L17

It will now only apply to the dynamic text. When the localized message is displayed this style will not be in effect.

jstoffan: Cool, sounds good. Thanks for confirming!

PR #1380, opened 2019-07-09
fix(collapsible): remove react-animate-height dependency

### LINKED ISSUE #1379: Collapsible component unnecessarily calculates height when collapsed
As it stands right now the `Collapsible` component will run an expensive height calculate for it's children even when the component is collapsed and the height is predictably 0.

BLamy:
fixes #1379

I'm using the `Collapsible` component in the sidebar of an internal application I made. I've noticed that it will attempt to calculate the height of the collapsible even if it is closed. This operation is pretty expensive and leads to janky rendering.

# before
sidebar takes 211ms to render
image

# after
sidebar takes 12ms to render with all Collapsible components closed. 
image


boxcla:
Hi @BLamy, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

BLamy:
CLA signed

src/components/collapsible/Collapsible.js
```diff
@@ -119,9 +119,11 @@ class Collapsible extends React.Component<Props, State> {
                     </PlainButton>
                     {isOpen && headerActionItems}
                 </div>
-                <AnimateHeight duration={animationDuration} height={isOpen ? 'auto' : 0}>
-                    <div className="collapsible-card-content">{children}</div>
-                </AnimateHeight>
+                {isOpen && (
+                    <AnimateHeight duration={animationDuration} height="auto">
```
jstoffan: Does animation still work properly with this change? React can sometimes struggle with animating a "new" item in the DOM.

BLamy: Yeah it seems to work fine.

jstoffan: I'm seeing the cards fail to animate if I make the same change locally, likely due to the class being applied after the element has been added to the DOM at full height. Is it possible the react-animate-height library supports this already in a different way?

BLamy:
@jstoffan I updated this PR with a new approach. Would you mind checking it out and seeing how you feel about the animation? 

This will hide the children when it is closed. But then it is opened, the children will render and immediately afterward the scroll height of the div will be set to the height of the animated div.

src/components/collapsible/Collapsible.js
```diff
@@ -119,9 +141,15 @@ class Collapsible extends React.Component<Props, State> {
                     </PlainButton>
                     {isOpen && headerActionItems}
                 </div>
-                <AnimateHeight duration={animationDuration} height={isOpen ? 'auto' : 0}>
-                    <div className="collapsible-card-content">{children}</div>
-                </AnimateHeight>
+                <div
+                    ref={this.ref}
+                    style={{
```
jstoffan: Ideally, the non-dynamic portions of these styles should live as part of a class in the stylesheet rather than as inline styles.

src/components/collapsible/Collapsible.js
```diff
@@ -40,7 +39,13 @@ type State = {
     isOpen: boolean,
 };
 
+type Ref = {
+    current: null | HTMLElement,
+};
+
 class Collapsible extends React.Component<Props, State> {
+    ref: Ref;
```
jstoffan: Minor, but this could have a slightly more descriptive name and its type could be declared inline: `transitionElement: { current: null | HTMLElement } = React.createRef();`

src/components/collapsible/Collapsible.js
```diff
@@ -50,11 +55,18 @@ class Collapsible extends React.Component<Props, State> {
...
         super(props);
+        this.ref = React.createRef();
         this.state = {
             isOpen: props.isOpen,
         };
     }
 
+    componentDidMount() {
+        if (this.ref.current) {
+            this.ref.current.style.height = `0px`;
```
jstoffan: Could we instead we set the height value in state on `componentDidMount` when `isOpen` changed, then pass the height style to the wrapping div in render?

src/components/collapsible/Collapsible.js
```diff
@@ -63,10 +75,20 @@ class Collapsible extends React.Component<Props, State> {
...
+                if (isOpen) {
+                    if (this.ref.current) {
+                        this.ref.current.style.height = `${this.ref.current.scrollHeight}px`;
+                    }
+                    if (onOpen) {
+                        onOpen(this);
+                    }
+                } else {
+                    if (this.ref.current) {
+                        this.ref.current.style.height = '0px';
```
jstoffan: I think this can just be `0`.

src/components/collapsible/Collapsible.js
```diff
@@ -63,10 +75,20 @@ class Collapsible extends React.Component<Props, State> {
             }),
             () => {
                 const { isOpen } = this.state;
-                if (isOpen && onOpen) {
-                    onOpen(this);
-                } else if (!isOpen && onClose) {
-                    onClose(this);
+                if (isOpen) {
```
priyajeet: Few things can be cleaned up a bit.

1. Since this function is doing more now, can extract it out at the class level instead of being inlined. Will make it easier to test.
2. onOpen and onClose can be defaulted to lodash/noop, which would reduce the need to test 2 branch statements.
3. this.ref.current is repeated, can be saved as a const/variable.

src/components/collapsible/Collapsible.js
priyajeet: +2

Just need to have transition-duration here

package.json
```diff
@@ -226,7 +226,6 @@
         "query-string": "5.1.1",
         "raf": "^3.4.0",
         "react": "^16.8.0",
-        "react-animate-height": "^2.0.8",
```
priyajeet: Also nuke flow-typed/npm/react-animate-height_vx.x.x.js

src/components/collapsible/Collapsible.js
```diff
@@ -75,7 +97,7 @@ class Collapsible extends React.Component<Props, State> {
     render() {
         const { isOpen }: State = this.state;
         const {
-            animationDuration,
+            animationDuration = 100,
```
priyajeet: Already defaulted in defaultProps above

src/components/collapsible/Collapsible.js
```diff
@@ -63,10 +75,20 @@ class Collapsible extends React.Component<Props, State> {
             }),
             () => {
                 const { isOpen } = this.state;
-                if (isOpen && onOpen) {
-                    onOpen(this);
-                } else if (!isOpen && onClose) {
-                    onClose(this);
+                if (isOpen) {
+                    if (this.ref.current) {
```
priyajeet: The if() check needed?

priyajeet: I guess flow would complain otherwise :/

src/components/collapsible/Collapsible.js
```diff
@@ -40,7 +39,13 @@ type State = {
     isOpen: boolean,
 };
 
+type Ref = {
+    current: null | HTMLElement,
+};
+
 class Collapsible extends React.Component<Props, State> {
```
priyajeet: Might be fine changing this to a React.PureComponent too, since props/state is primitives mostly

src/components/collapsible/Collapsible.js
```diff
@@ -119,9 +141,15 @@ class Collapsible extends React.Component<Props, State> {
...
                     {isOpen && headerActionItems}
                 </div>
-                <AnimateHeight duration={animationDuration} height={isOpen ? 'auto' : 0}>
-                    <div className="collapsible-card-content">{children}</div>
-                </AnimateHeight>
+                <div
+                    ref={this.ref}
+                    style={{
+                        overflow: 'hidden',
+                        transition: `height ${animationDuration}ms cubic-bezier(0.4, 0, 0.2, 1) 0ms`,
```
priyajeet: Probably briefly test IE11 too, so that things arent janky

src/components/collapsible/Collapsible.js
BLamy: I did this to make flow happy. Flow doesn't seem to understand the `static defaultProps`.

Therefore, it warned that I could get `transaction-duration: undefinedms`.

src/components/collapsible/Collapsible.js
BLamy: Yeah this was causing were you couldn't default a collapsible to open.

BLamy:
> Does the `AnimateHeight` library do anything else in terms of behavior that we'd lose by removing it?

AnimateHeight does have some additional properties such as the ability to animate opacity. However, we don't expose that to the consumer of `Collapsible`. 

Animate height also adds classes to the element while it is animating. It's possible a User of BUIE is doing some style overrides on these. 
```
{
  animating:                  'rah-animating',
  animatingUp:                'rah-animating--up',
  animatingDown:              'rah-animating--down',
  static:                     'rah-static',
  animatingToHeightZero:      'rah-animating--to-height-zero',
  animatingToHeightAuto:      'rah-animating--to-height-auto',
  animatingToHeightSpecific:  'rah-animating--to-height-specific',
  staticHeightZero:           'rah-static--height-zero',
  staticHeightAuto:           'rah-static--height-auto',
  staticHeightSpecific:       'rah-static--height-specific',
}
```

src/components/collapsible/Collapsible.js
BLamy: https://cloud.box.com/s/z38hzyup9xmxcpls3f5fec7oarbk2ubg

BLamy:
@priyajeet @jstoffan Updated

src/components/collapsible/Collapsible.js
```diff
@@ -50,18 +55,36 @@ class Collapsible extends React.Component<Props, State> {
...
     }
 
+    componentDidMount() {
+        this.updateTransitionElementHeight();
+    }
+
+    updateTransitionElementHeight = () => {
+        const { isOpen } = this.state;
+        const { current } = this.transitionElement;
+        if (!current) return;
```
jstoffan: We usually try to avoid this style of inline return.

```
if (!current) {
    return;
}
```

src/components/collapsible/Collapsible.js
```diff
@@ -50,18 +55,36 @@ class Collapsible extends React.Component<Props, State> {
...
+
+        if (isOpen) {
+            current.style.height = `${current.scrollHeight}px`;
```
jstoffan: Seems like this could be a ternary to avoid repeating the property access?

BLamy:
Updated again:

https://cloud.box.com/s/0pcl8ilchrynnotr32k4tbgiqrv906mw

src/components/collapsible/Collapsible.js
```diff
@@ -50,11 +58,26 @@ class Collapsible extends React.Component<Props, State> {
...
     }
 
+    componentDidMount() {
+        this.updateTransitionElementHeight();
+    }
+
+    updateTransitionElementHeight = () => {
+        const { current } = this.transitionElement;
+        if (current) {
+            this.setState(prevState => ({
```
jstoffan: Can this property be set at the same time as the `this.setState` call in `toggleVisibility`?

src/components/collapsible/Collapsible.js
```diff
@@ -50,11 +58,26 @@ class Collapsible extends React.Component<Props, State> {
...
         super(props);
+        this.transitionElement = React.createRef();
         this.state = {
             isOpen: props.isOpen,
+            height: '0',
         };
     }
 
+    componentDidMount() {
+        this.updateTransitionElementHeight();
```
jstoffan: It seems like this would cause the panel to animate open if `this.props.isOpen` is true initially.

src/components/collapsible/Collapsible.js
```diff
@@ -50,11 +58,26 @@ class Collapsible extends React.Component<Props, State> {
 
     constructor(props: Props) {
         super(props);
+        this.transitionElement = React.createRef();
         this.state = {
             isOpen: props.isOpen,
+            height: '0',
```
jstoffan: It seems like this should be conditional based on `props.isOpen`.

src/components/collapsible/Collapsible.js
```diff
@@ -68,12 +91,13 @@ class Collapsible extends React.Component<Props, State> {
                 } else if (!isOpen && onClose) {
                     onClose(this);
                 }
+                this.updateTransitionElementHeight();
```
jstoffan: Does this need to happen after the previous setState call or can they be combined?

src/components/collapsible/Collapsible.js
BLamy: At this point `this.transitionElement.current` would be null so you can't do a conditional. We can set to `auto` by default then use `componentDidMount` to update from there. There won't be any animation because you can't animate from auto.

src/components/collapsible/Collapsible.js
```diff
@@ -37,10 +36,19 @@ type Props = {
...
 type State = {
+    /** This is the height of the animated div */
+    height: string,
+    /** This value represents if the collapsible is open or not */
     isOpen: boolean,
 };
 
-class Collapsible extends React.Component<Props, State> {
+type Ref = {
+    current: null | HTMLElement,
```
priyajeet: I believe this can just be ?HTMLDivElement and you can inline it within the class if you want since its used at one place only.

src/components/collapsible/Collapsible.js
```diff
@@ -50,30 +58,45 @@ class Collapsible extends React.Component<Props, State> {
...
-                const { isOpen } = this.state;
-                if (isOpen && onOpen) {
-                    onOpen(this);
-                } else if (!isOpen && onClose) {
-                    onClose(this);
-                }
-            },
-        );
+        const { current } = this.transitionElement;
+        if (current) {
```
priyajeet: Minor stylistic comment:
```
if (!current) {
    return;
}
```
In this way majority of your function is less indented, aka less arrow-ness.

PR #1381, opened 2019-07-09
docs(styleguide): create usable import paths

alexkrolick:
Adds a config option that formats component paths in the documentation in a way that can be copy-pasted into consuming apps

Screen Shot 2019-07-09 at 3 12 30 PM


boxcla:

PR #1383, opened 2019-07-09
feat(media): create media object component

alexkrolick:


boxcla:

src/components/media/MediaMenu.js
```diff
@@ -0,0 +1,33 @@
...
+    className?: string,
+    /** is the dropdown menu button disabled */
+    isDisabled: boolean,
+};
+
+function MediaMenu({ className, children, isDisabled = false, ...rest }: Props) {
+    return (
+        <div className={classnames('bdl-Media-menu', className)} {...rest}>
+            <DropdownMenu constrainToScrollParent isRightAligned>
+                <PlainButton isDisabled={isDisabled} type="button">
```
jstoffan: The toggle button should have hidden text to support accessibility tools.

src/components/media/Media.js
```diff
@@ -0,0 +1,34 @@
...
+function Media({ as: Wrapper, children, className, ...rest }: Props) {
+    return (
+        <Wrapper className={classnames('bdl-Media', className)} {...rest}>
+            {children}
+        </Wrapper>
+    );
+}
+
+Media.defaultProps = {
+    as: 'div',
```
jstoffan: Can we add semantic meaning to the markup? Since the intended use is for an image with related content, would a `figure/figcaption` pairing make sense?

src/components/media/Media.js
```diff
@@ -0,0 +1,34 @@
+// @flow
+import * as React from 'react';
+import classnames from 'classnames';
+import MediaImg from './MediaImg';
```
jstoffan: We've generally tried to avoid abbreviations. Would `MediaImage` work?

src/components/media/Media.scss
```diff
@@ -0,0 +1,28 @@
+@import '../../styles/variables';
+
+// the outermost element
+.bdl-Media {
+    align-items: flex-start;
+    display: flex;
+}
+
+// the media content, ie, an avatar
+.bdl-Media-img {
```
jstoffan: Each sub-component should have its own top-level class name, such as `.bdl-MediaImage`. It's probably fine for them to live in the same stylesheet, though, as long as re-importing it in each component doesn't create duplicate styles in the output CSS.

src/components/media/Media.js
alexkrolick: Does that make sense if it's a text-only avatar?

src/components/media/Media.scss
alexkrolick: I was following this 

https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md#ComponentName-descendentName

jstoffan: That applies when different elements live in the same file, but since these are declared as separate components, they should be named as such.

src/components/media/Media.js
jstoffan: Perhaps the naming is confusing, as `MediaImage` also doesn't fit text content very well.

src/components/media/Media.scss
alexkrolick: The sub-elements are meant to be used like `Media.Body` not as separate imports.

These are sub-elements almost exactly like the Tweet example in the SUIT docs.

src/components/media/Media.js
alexkrolick: In the Bootstrap implementation, the first child is just the media content, it doesn't have a wrapper. We could do that, but I'm not a huge fan because you have to use .cloneElement to add vertical alignment options, etc.

Some ideas:

- `Media.Content` 
- `Media.Left`
- `Media.Object`

Prior Art:

- https://react-bootstrap.github.io/layout/media/
- https://foundation.zurb.com/sites/docs/media-object.html
- https://bulma.io/documentation/layout/media-object/

alexkrolick: MDN says other flow content is inside figure seems OK so maybe we _should_ default to that. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figure#Usage_notes

alexkrolick: Going with `<Media.Figure />` and defaulting to `figure` within

priyajeet:
Probably should check if https://github.com/box/box-ui-elements/blob/master/src/elements/content-sidebar/versions/VersionsItem.js can be refactored to use these components/design if possible

src/components/media/MediaFigure.js
```diff
@@ -0,0 +1,27 @@
...
+
+function MediaFigure({ as: Wrapper, className, children, ...rest }: Props) {
+    return (
+        <Wrapper className={classnames('bdl-Media-figure', className)} {...rest}>
+            {children}
+        </Wrapper>
+    );
+}
+
+MediaFigure.defaultProps = {
```
priyajeet: Not a big deal, but doing so I believe prevents uglifyJS from cleaning up MediaFigure. Since MediaFigure.defaultProps counts as a usage as such it cannot determine if MediaFigure was actually referenced.

alexkrolick: I'll see if using default values for the function shows up in styleguidist. I'd prefer that anyways.

src/components/media/MediaBody.js
```diff
@@ -0,0 +1,22 @@
...
+import './Media.scss';
+
+type Props = {
+    /** Child elements */
+    children: React.Node,
+    /** Additional class names */
+    className?: string,
+};
+
+function MediaBody(props: Props) {
```
priyajeet: For code consistency I would do

```jsx
const MediaBody = ({ className, children, ...rest }: Props) => (
        <div className={classnames('bdl-Media-body', className)} {...rest}>
            {children}
        </div>
);
```

For example even within this PR two functional components seem to have their props extracted differently. And then most of our functional components normally use the const foo = () style.

src/components/media/MediaFigure.js
alexkrolick: Oh it was because of this: https://github.com/facebook/flow/issues/5461

alexkrolick: I can either use `defaultProps` or use this syntax that doesn't report defaultValues to Styleguidist

```js
const MediaFigure = (props: Props) => {
    const { as: Wrapper = 'figure', className, children, ...rest } = props;
```

priyajeet: Its fine I think to start using defaultProps for functional components. I am thinking we should make it a norm (so that flow is happy) and eat the theoretical minor bundle size increase. Maybe we can chat about it in #frontend

alexkrolick:
I tested this out with the Activity Feed Comment/Task container and it seems to work. I'll put that out as a separate PR once this is in.

src/components/media/MediaMenu.js
```diff
@@ -0,0 +1,37 @@
...
+    /** is the dropdown menu button disabled */
+    isDisabled: boolean,
+    /** aria-label prop for button icon */
+    label?: string,
+};
+
+const MediaMenu = ({ className, children, isDisabled, label, ...rest }: Props) => (
+    <div className={classnames('bdl-Media-menu', className)} {...rest}>
+        <DropdownMenu constrainToScrollParent isRightAligned>
+            <PlainButton isDisabled={isDisabled} type="button" aria-label={label}>
```
jstoffan: It looks like this currently doesn't support adding className or resin attributes to the button/menu, which we use in VersionsSidebar, for example.

src/components/media/Media.scss
```diff
@@ -0,0 +1,30 @@
...
+    display: flex;
+}
+
+// the media content, ie, an avatar
+.bdl-Media-figure {
+    align-self: flex-start;
+    flex: 0 0 auto;
+    justify-self: flex-start;
+    margin: 0;
+    margin-right: 10px;
```
jstoffan: I thought we had a SCSS lint rule against using shorthand and long-hand properties in the same style. It's usually discouraged. We should also consider moving the margin to the `-body`, so everything is pushing itself left rather than pushing its neighbor right.

src/components/dropdown-menu/DropdownMenu.js
```diff
@@ -171,6 +171,7 @@ class DropdownMenu extends React.Component<Props, State> {
             constrainToScrollParent,
             constrainToWindow,
             className,
+            ...rest
```
jstoffan: This seems like a risky change, given DropdownMenu's widespread use.

src/components/dropdown-menu/DropdownMenu.js
```diff
@@ -184,6 +185,7 @@ class DropdownMenu extends React.Component<Props, State> {
         const menu = elements[1];
 
         const menuButtonProps: Object = {
+            ...rest,
```
jstoffan: This seems like a risky change, given DropdownMenu's widespread use.

src/components/media/Media.scss
```diff
@@ -0,0 +1,40 @@
...
+    padding: 0;
+}
+
+// the main content, e.g. title and message
+.bdl-Media-body {
+    flex: 1 1 auto;
+    margin-left: 10px;
+    // handle wrapping long words in text content without breaking
+    // float position of menu
+    word-break: break-word;
```
jstoffan: Were you able to get this to work in IE11?

src/components/media/Media.scss
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// the "..." menu button, meant to be used inside Media-body
+// float is used to reflow content instead of pushing it to the left
+.bdl-Media-menu {
+    float: right;
+}
+
+// overrides for .btn-plain's margin resets on focus states
+.bdl-Media-menu,
```
jstoffan: Minor, but this can be modified to avoid the repetition of the class name:

```
.bdl-Media-menu {
    &,
    &:active,
    &:hover,
    &:focus {
        margin-bottom: 5px;
        margin-left: 5px;
    }
}
```

src/components/media/Media.scss
alexkrolick: yes

Screen Shot 2019-07-16 at 4 43 51 PM

src/components/media/Media.scss
```diff
@@ -0,0 +1,45 @@
...
+    flex-grow: 0;
+    flex-shrink: 0;
+    justify-self: flex-start;
+    margin: 0;
+    padding: 0;
+}
+
+// the main content, e.g. title and message
+.bdl-Media-body {
+    flex-basis: 100%;
```
jstoffan: Does shorthand `flex: 1 1 100%` work here?

src/components/media/Media.scss
```diff
@@ -0,0 +1,45 @@
...
+// the outermost element
+.bdl-Media {
+    align-items: flex-start;
+    display: flex;
+}
+
+// the media content, ie, an avatar
+.bdl-Media-figure {
+    align-self: flex-start;
+    flex-grow: 0;
```
jstoffan: We usually want to define all three properties, since otherwise we're leaving the basis/width up to the various browsers. Is this `flex: 0 0 auto;`?

alexkrolick: The intersection of the shorthand with IE and minifiers is kind of ambiguous but I will give the shorthand a shot

https://developer.mozilla.org/en-US/docs/Web/CSS/flex#Syntax

alexkrolick:
Thanks for your feedback everyone

PR #1384, opened 2019-07-09
feat(uploads): add resin tracking for uploads manager

rosepierce:
Adding resin tracking for retrying failed uploads and expanding/collapsing the uploads manager.

### DIFF
--- a/src/elements/content-uploader/ItemAction.js
+++ b/src/elements/content-uploader/ItemAction.js
@@ -26,6 +26,7 @@ type Props = {
 
 const ItemAction = ({ status, onClick, intl, isFolder = false }: Props) => {
     let icon = <IconClose />;
+    let retry = false;
     let tooltip = intl.formatMessage(messages.uploadsCancelButtonTooltip);
 
     if (isFolder && status !== STATUS_PENDING) {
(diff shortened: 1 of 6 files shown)


boxcla:

src/elements/content-uploader/OverallUploadsProgressBar.js
```diff
@@ -73,6 +75,10 @@ const OverallUploadsProgressBar = ({ percent, view, onClick, onKeyDown, isDraggi
     return (
         <div
             className="bcu-overall-progress-bar"
+            data-resin-target={classNames({
```
jstoffan: Should we use `classNames` for non-CSS purposes? I had to look at this a few times to realize it wasn't setting actual class names.

src/elements/content-uploader/__tests__/__snapshots__/ItemAction-test.js.snap
```diff
@@ -18,6 +18,7 @@ exports[`elements/content-uploader/ItemAction should render correctly with STATU
     theme="default"
   >
     <PlainButton
+      data-resin-target=""
```
jstoffan: It seems like the attribute shouldn't be applied if the value is empty.

src/elements/content-uploader/ItemAction.js
```diff
@@ -52,7 +54,7 @@ const ItemAction = ({ status, onClick, intl, isFolder = false }: Props) => {
     return (
         <div className="bcu-item-action">
             <Tooltip position="top-left" text={tooltip}>
-                <PlainButton onClick={onClick} type="button">
+                <PlainButton data-resin-target={target} onClick={onClick} type="button">
```
mickr: Should this be used to track other states as well? If it isn't retrying should that data attribute be on the button?

PR #1387, opened 2019-07-10
build: remove lint and test step temporarily

priyajeet:
so that build reaches release step faster.
Adds GIT_BRANCH env variable for semantic release
Adds github token to the git repo url


boxcla:

PR #1388, opened 2019-07-10
docs(styleguide): apply box theme

alexkrolick:
## After

2019-07-09 18 43 33 localhost c1224d448569

## Before

2019-07-09 18 54 30 opensource box com e6b052385a9d

### DIFF
--- a/scripts/styleguide.config.js
+++ b/scripts/styleguide.config.js
@@ -3,6 +3,9 @@ const webpackConf = require('./webpack.config.js');
 
 const webpackConfig = Array.isArray(webpackConf) ? webpackConf[0] : webpackConf;
 
+// theme variables
+const vars = require('../src/styles/variables.json');
+
 const allSections = [
     {
         name: 'Elements',
@@ -97,6 +100,9 @@ const allSections = [
             '../src/components/toggle/Toggle.js',
             '../src/components/tooltip/Tooltip.js',
         ],
+        description: 'Box UI Elements components implement the reusable building blocks of the Box Design Language',
+        sectionDepth: 2,
+        usageMode: 'expand',
     },
     {
         name: 'Icons',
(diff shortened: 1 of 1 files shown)


boxcla:

scripts/styleguide.config.js
```diff
@@ -97,6 +100,9 @@ const allSections = [
             '../src/components/toggle/Toggle.js',
             '../src/components/tooltip/Tooltip.js',
         ],
+        description: 'Box UI Elements components implement the reusable building blocks of the Box Design Language',
+        sectionDepth: 2,
```
alexkrolick: This makes each component have its own route instead of one huge page with every component

scripts/styleguide.config.js
```diff
@@ -97,6 +100,9 @@ const allSections = [
...
+        usageMode: 'expand',
```
alexkrolick: This expands the props table by default for the components table only

scripts/styleguide.config.js
```diff
@@ -270,21 +276,89 @@ module.exports = {
...
+            sidebarBackground: vars['bdl-gray-02'],
+            light: vars['bdl-gray-80'],
+            lightest: vars['bdl-gray-20'],
+            focus: vars['bdl-light-blue-50'],
+            codeBase: vars['bdl-gray'],
         },
         fontFamily: {
             base: 'Lato, "Helvetica Neue", Helvetica, Arial, sans-serif',
         },
+        fontSize: {
```
alexkrolick: The import of the box-ui css causes some of the element styles to be applied, except things with explicit overrides in the styleguide theme. This harmonizes the values for the font-sizes between the box theme and the styleguide.

scripts/styleguide.config.js
```diff
@@ -270,21 +276,89 @@ module.exports = {
...
+            code: {
+                fontSize: 12,
+            },
+        },
+        Type: {
+            type: {
+                fontSize: 12,
+            },
+        },
+        Name: {
```
alexkrolick: Props Table > prop names

scripts/styleguide.config.js
```diff
@@ -270,21 +276,89 @@ module.exports = {
...
+                fontSize: 12,
+                padding: 10,
+            },
+        },
+        Code: {
+            code: {
+                fontSize: 12,
+            },
+        },
+        Type: {
```
alexkrolick: Props Table > prop types

scripts/styleguide.config.js
```diff
@@ -270,21 +276,89 @@ module.exports = {
     styles: {
         Heading: {
             heading: {
-                lineHeight: 1,
+                lineHeight: 2,
+            },
+        },
+        Pathline: {
```
alexkrolick: This is the import path block

scripts/styleguide.config.js
```diff
@@ -270,21 +276,89 @@ module.exports = {
...
+                borderRadius: 3,
+                color: vars['bdl-gray'],
+                display: 'inline',
+                fontSize: 12,
+                padding: 10,
+            },
+        },
+        Code: {
+            code: {
+                fontSize: 12,
```
jstoffan: It looks like this is set quite a few times. Is there a more central option, such as `theme.fontSize.base` or `theme.fontSize.small`? I'm not certain if the theme affects the layout-level elements.

alexkrolick: They use the base fontSize which is too large with the default monospace font

PR #1391, opened 2019-07-10
chore(deps): update jest

alexkrolick:


boxcla:

priyajeet:
yarn lock diff looks more than it should?

src/__mocks__/react-intl.js
```diff
@@ -0,0 +1 @@
+export * from '../../scripts/jest/mocks/react-intl-mocks';
```
priyajeet: exclude mocks folder from babel compile, within package.json scripts, similar to tests

src/__mocks__/react-intl.js
```diff
@@ -31,7 +33,7 @@ type Props = {
     locale: string,
 };
 
-export const injectIntl = Component => {
+export const injectIntl = (Component: React.ComponentType<any>) => {
```
priyajeet: Probably needs the import * as React on top

PR #1395, opened 2019-07-11
fix(versions): Reload the page when a version is promoted (temporary)

jstoffan:
Currently, the host application isn't informed when a version is promoted, which can lead to stale data and user confusion. This is a temporary workaround to avoid a launch blocker; I intend to follow-up with a better solution soon.


boxcla:

PR #1397, opened 2019-07-11
 fix(uploads): resin for canceling uploads

rosepierce:
Adding target for resin for canceling uploads.

### DIFF
--- a/src/elements/content-uploader/ItemAction.js
+++ b/src/elements/content-uploader/ItemAction.js
@@ -26,9 +26,7 @@ type Props = {
 
 const ItemAction = ({ status, onClick, intl, isFolder = false }: Props) => {
     let icon = <IconClose />;
-    let retry = false;
-    let cancel = false;
-    let buttonProps = {};
+    let target = null;
     let tooltip = intl.formatMessage(messages.uploadsCancelButtonTooltip);
 
     if (isFolder && status !== STATUS_PENDING) {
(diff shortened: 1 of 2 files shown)


boxcla:

src/elements/content-uploader/ItemAction.js
```diff
@@ -27,6 +27,8 @@ type Props = {
 const ItemAction = ({ status, onClick, intl, isFolder = false }: Props) => {
     let icon = <IconClose />;
     let retry = false;
+    let cancel = false;
```
mickr: Couldn't these be just a single status that is populated with the `data-resin-target` attribute?

src/elements/content-uploader/ItemAction.js
```diff
@@ -54,7 +55,7 @@ const ItemAction = ({ status, onClick, intl, isFolder = false }: Props) => {
     return (
         <div className="bcu-item-action">
             <Tooltip position="top-left" text={tooltip}>
-                <PlainButton onClick={onClick} type="button" {...(retry ? { 'data-resin-target': 'uploadretry' } : {})}>
+                <PlainButton onClick={onClick} type="button" {...(target ? { 'data-resin-target': target } : {})}>
```
jstoffan: To avoid the ternary, you could go with something like this:

```
let resin = {};

if (whatever) {
   resin.target = 'mytarget';
}

<PlainButton onClick={onClick} type="button" {...resin}>
```

PR #1398, opened 2019-07-11
fix(button): don't override margin, padding for :hover state

alexkrolick:
BREAKING: apply margin, padding overrides for .btn-plain only to base class, not to pseudoselectors, so that buttons do not reposition on focus/hover
BREAKING: add override to prevent margin being added between .btn-plain buttons

### DIFF
--- a/src/styles/common/_buttons.scss
+++ b/src/styles/common/_buttons.scss
@@ -86,14 +86,22 @@
     }
 }
 
+.btn-plain {
+    margin: 0;
+    padding: 0;
+
+    // prevent default .btn behavior of adding space between buttons
+    & + & {
+        margin-left: 0;
+    }
+}
+
 .btn-plain,
 .btn-plain:hover,
 .btn-plain:active {
     border: none;
     cursor: pointer;
     font-weight: normal;
-    margin: 0;
-    padding: 0;
 }
 
 .btn-plain,


boxcla:

src/styles/common/_buttons.scss
```diff
@@ -86,14 +86,22 @@
     }
 }
 
+.btn-plain {
+    margin: 0;
+    padding: 0;
+
+    // prevent default .btn behavior of adding space between buttons
+    & + & {
+        margin: 0;
```
jstoffan: We should try to limit overrides as much as possible. Can this just override `margin-left`?

src/styles/common/_buttons.scss
```diff
@@ -86,14 +86,22 @@
...
+    }
+}
+
 .btn-plain,
 .btn-plain:hover,
 .btn-plain:active {
     border: none;
```
jstoffan: Can these live in the base class as well? Or do they get set by the browser on hover/active?

alexkrolick: I don't know and I don't know why :focus doesn't get it, so I am assuming there is a reason to limit the scope of changes

jstoffan:
What kind of manual testing did you run on this change? Do the various top-level Elements look the same? Internal apps?

alexkrolick:
The place where the issue was reported was the tab bar, which adds the .btn-plain class to the tab link .btn components. They were showing up with padding.

## After fix

Screen Shot 2019-07-11 at 4 49 27 PM
Screen Shot 2019-07-11 at 4 49 22 PM

## Before fix

Screen Shot 2019-07-11 at 4 51 08 PM
Screen Shot 2019-07-11 at 4 51 01 PM

alexkrolick:
Linking and testing other components now

alexkrolick:
Internal apps look good

alexkrolick:
The `+` combinator creates increased specificity for the override, which makes buttons harder to work with. Reverting the fix instead in #1400 and coming back with a better solution later.

PR #1401, opened 2019-07-12
chore(list-view): Delete List View from BUIE

Visylvius:
Deletes List View from BUIE.


boxcla:

PR #1404, opened 2019-07-15
feat(versions): Add client-side error messages for version actions

jstoffan:


boxcla:

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -75,17 +77,17 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
             this.deleteVersion(versionId)
                 .then(this.fetchData)
                 .then(() => this.handleDeleteSuccess(versionId))
-                .catch(this.handleActionError);
+                .catch(() => this.handleActionError(messages.versionActionDeleteError));
```
ConradJChan: Is there any issue where `fetchData` is the one with a promise rejection triggering a fetch error as well as delete error handling?

jstoffan: I'm not re-throwing in the `fetchData` catch, so I don't think we'd see two errors. We could potentially see an issue where a subsequent `.then` wouldn't receive file/version data, but the instances we have like that have guard checks in them already.

jstoffan: I moved the `.catch(this.handleFetchError)` to `componentDidMount` to avoid clearing the list if delete, promote, or restore fail.

PR #1405, opened 2019-07-15
fix(versions): Disable prior version preview for watermarked files

jstoffan:
Watermarked files currently do not support preview for older versions. I hope to be able to revert this change in the future, but the timeline is uncertain.

### DIFF
--- a/src/elements/content-sidebar/versions/VersionsItem.js
+++ b/src/elements/content-sidebar/versions/VersionsItem.js
@@ -77,7 +77,7 @@ const VersionsItem = ({
     const isDeleted = action === VERSION_DELETE_ACTION;
     const isDownloadable = !!is_download_available;
     const isLimited = versionCount - versionInteger >= versionLimit;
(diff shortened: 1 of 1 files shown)


boxcla:

src/elements/content-sidebar/versions/VersionsItem.js
```diff
@@ -77,7 +77,7 @@ const VersionsItem = ({
     const isDeleted = action === VERSION_DELETE_ACTION;
     const isDownloadable = !!is_download_available;
     const isLimited = versionCount - versionInteger >= versionLimit;
-    const isRestricted = isWatermarked && !isCurrent && !can_download; // Watermarked files use can_download for preview
+    const isRestricted = isWatermarked && !isCurrent; // Watermarked files do not support prior version preview
```
priyajeet: Didnt break test?

jstoffan: The test is _also_ checking for the `can_download` permission, so it still passes. I'm hoping this change is temporary.

PR #1408, opened 2019-07-16
build: make bundles analysis on demand only

priyajeet:


boxcla:

package.json
```diff
@@ -36,6 +36,7 @@
         "url": "https://github.com/box/box-ui-elements/issues"
     },
     "scripts": {
+        "analyze:bundles": "BUNDLE_ANALYSIS=true npm-run-all setup build:prod:npm",
```
jstoffan: Is a separate script needed at this point?

PR #1409, opened 2019-07-16
feat(content-explorer): show cards with thumbnails

rickstefanik:


boxcla:

src/api/File.js
```diff
@@ -202,6 +204,85 @@ class File extends Item {
...
+        const access_token = this.xhr.token;
+        // no need to make api call since folders do not have thumbnails
+        if (item.type === 'folder' || !access_token) {
+            this.successHandler(null);
+            return;
+        }
+
+        const { id } = item;
+        const cache: APICache = this.getCache();
+        const key: string = this.getCacheKey(`_thumbnail_${id}`);
```
rickstefanik: Using same cache as `getFile` above.  Prefixing with `_thumbnail_` to prevent collisions.  Is this ok, or should a separate cache be used?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -411,6 +443,9 @@ class ContentExplorer extends Component<Props, State> {
         } else {
             this.setState({ rootName });
         }
+        if (viewMode === VIEW_MODE_GRID && items) {
```
rickstefanik: This logic is used multiple times, so I thought about making a `shouldFetchThumbnails` function, but flow would complain about `items` potentially being undefined below.  Would a `shouldFetchThumbnails` function be worth it?

src/api/File.js
```diff
@@ -202,6 +204,85 @@ class File extends Item {
...
+     * @return {void}
+     */
+    getFileThumbnail(
+        item: BoxItem,
+        dimensions: string,
+        successCallback: Function,
+        errorCallback: ElementsErrorCallback,
+    ): void {
+        if (this.isDestroyed()) {
+            this.successHandler(null);
```
ConradJChan: I'm seeing that in other api files, we return without calling the `successHandler`. Is there a reason why we want this for `getFileThumbnail`?

src/elements/content-explorer/ItemGridCell.js
```diff
@@ -24,10 +26,20 @@ const ItemGridCell = ({
...
 }: Props) => {
+    const url = getProp(item, `thumbnailUrl`);
+    const itemClassName = classNames(
+        'bce-ItemGridCell-item',
+        { 'bce-ItemGridCell-item--postLoadThumbnail': url },
+        { 'bce-ItemGridCell-item--itemIcon': !url },
+    );
     return (
-        <figure>
+        <figure className="bce-itemGridCell-figure">
```
rickstefanik: `figure` was adding 40px margins on the right and left sides of the content.  This made the thumbnails noticeably smaller.

src/api/File.js
```diff
@@ -202,6 +204,85 @@ class File extends Item {
...
+        this.errorCallback = errorCallback;
+
+        const access_token = this.xhr.token;
+        // no need to make api call since folders do not have thumbnails
+        if (item.type === 'folder' || !access_token) {
+            this.successHandler(null);
+            return;
+        }
+
+        const { id } = item;
```
ConradJChan: Should we also check to see if the representations are already available on the item?

src/api/File.js
```diff
@@ -202,6 +204,85 @@ class File extends Item {
...
+                    const entries = response.data.representations.entries;
+                    if (getProp(entries, '[0].status.state') !== 'success') {
+                        return null;
+                    }
+
+                    // if unable to fetch jpg thumbnail, grab png rep of first page.
+                    // (Asset path for thumbnail is simply empty string.)
+                    const asset_path = entries[0].representation === 'jpg' ? '' : '1.png';
+                    const thumbnailBaseUrl = entries[0].content.url_template.replace('{+asset_path}', asset_path);
+                    return `${thumbnailBaseUrl}?access_token=${access_token}`;
```
ConradJChan: Are we able to pass the `access_token` via Authorization header instead? Also would help considering that tokens expire so if we cache the token with the url, that may get stale

src/api/File.js
ConradJChan: Should we do `_thumbnail_url` to be more clear?

src/api/File.js
```diff
@@ -202,6 +204,85 @@ class File extends Item {
...
+            url: this.getUrl(id),
+            headers: {
+                // API will return first representation it finds, so 1024x1024 png is fallback.
+                'X-Rep-Hints': `[jpg?dimensions=${dimensions},png?dimensions=1024x1024]`,
+            },
+            params: {
+                fields: FIELD_REPRESENTATIONS,
+            },
+        };
+        try {
```
ConradJChan: Is this try-catch necessary? Since the logic is done in the promise chain, would a promise `catch` work instead?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -383,6 +386,35 @@ class ContentExplorer extends Component<Props, State> {
...
+    fetchThumbnailUrls(items: Array<BoxItem>, dimensions: string): void {
+        const fileAPI = this.api.getFileAPI();
+
+        // not using Promise.all since thumbnails should load one at a time
+        items.forEach((item, index) => {
+            new Promise(resolve => {
+                // getFileThumbnail resolves with thumbnailUrl if one is found, null otherwise.
+                fileAPI.getFileThumbnail(item, dimensions, resolve, this.errorCallback);
+            }).then(thumbnailUrl => {
+                const currentCollection = { ...this.state.currentCollection };
```
ConradJChan: This makes a shallow copy of the currentCollection but the underlying items array is sharing the reference with state.currentCollection so this modification is still a mutation. This is a helpful reference for updating immutable state https://redux.js.org/recipes/structuring-reducers/immutable-update-patterns

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -383,6 +386,40 @@ class ContentExplorer extends Component<Props, State> {
...
+     * The thumbnailUrls array in state will be populated with urls to thumbnails.
+     * If a thumbnail could not be found, that item's entry in thumbnailUrls will be null.
+     *
+     * @private
+     * @param {Array<BoxItem>} items - item collection object
+     * @param {string} dimensions - desired dimensions of thumbnail. Acceptable dimensions
+     * for a jpg are: "32x32", "94x94", "160x160", "320x320", "1024x1024", "2048x2048".
+     * @return {void}
+     */
+    fetchThumbnailUrls(items: Array<BoxItem> = [], dimensions: string): void {
```
ConradJChan: Could we include the `representations` field in the initial folder fetch request?

src/elements/content-explorer/ItemGridCell.js
ConradJChan: Is that from user-agent styles? We might want to add that _reset.scss

ConradJChan: Should this be `bce-ItemGridCell`?

src/elements/content-explorer/ItemGridCell.js
```diff
@@ -24,10 +26,20 @@ const ItemGridCell = ({
...
             <div className="bce-ItemGridCell-itemThumbnail">
-                <div className="bce-ItemGridCell-itemIcon">{getIcon(128, item)}</div>
+                {url ? (
```
ConradJChan: There's enough to consider making the thumbnail its own component?

src/elements/content-explorer/ItemGridCell.js
rickstefanik: Yes it is.  I'll make that change in `_reset.scss`.  This appears to be the only place we use `figure` in elements, so the change shouldn't affect anything else.

src/api/File.js
```diff
@@ -202,6 +204,85 @@ class File extends Item {
...
+                .then(response => {
+                    const entries = response.data.representations.entries;
+                    if (getProp(entries, '[0].status.state') !== 'success') {
+                        return null;
+                    }
+
+                    // if unable to fetch jpg thumbnail, grab png rep of first page.
+                    // (Asset path for thumbnail is simply empty string.)
+                    const asset_path = entries[0].representation === 'jpg' ? '' : '1.png';
+                    const thumbnailBaseUrl = entries[0].content.url_template.replace('{+asset_path}', asset_path);
```
jstoffan: The property accessors here are difficult to read and seem like they'd be brittle; if any of the objects are missing, a generic "property is undefined" error will be thrown. Can we use `lodash/get` for some of these?

src/api/File.js
```diff
@@ -202,6 +204,85 @@ class File extends Item {
...
+                .get(xhrOptions)
+                .then(response => {
+                    const entries = response.data.representations.entries;
+                    if (getProp(entries, '[0].status.state') !== 'success') {
+                        return null;
+                    }
+
+                    // if unable to fetch jpg thumbnail, grab png rep of first page.
+                    // (Asset path for thumbnail is simply empty string.)
+                    const asset_path = entries[0].representation === 'jpg' ? '' : '1.png';
```
jstoffan: Should `entries[0]` be assigned to a variable, since it's accessed multiple times?

src/api/File.js
```diff
@@ -202,6 +204,85 @@ class File extends Item {
...
+        if (this.isDestroyed()) {
+            this.successHandler(null);
+            return;
+        }
+        this.successCallback = successCallback;
+        this.errorCallback = errorCallback;
+
+        const access_token = this.xhr.token;
+        // no need to make api call since folders do not have thumbnails
+        if (item.type === 'folder' || !access_token) {
```
jstoffan: These conditions should be combined with the `this.isDestroyed` check and should probably not call the success handler.

src/api/File.js
```diff
@@ -202,6 +204,85 @@ class File extends Item {
...
+        errorCallback: ElementsErrorCallback,
+    ): void {
+        if (this.isDestroyed()) {
+            this.successHandler(null);
+            return;
+        }
+        this.successCallback = successCallback;
+        this.errorCallback = errorCallback;
+
+        const access_token = this.xhr.token;
```
jstoffan: I don't think we access the token in such a way elsewhere. Is it available in `this.options.token`?

src/api/File.js
```diff
@@ -202,6 +204,85 @@ class File extends Item {
...
+     * @param {string} dimensions - desired dimensions of jpg thumbnail. Acceptable dimensions
+     * for the jpg are: "32x32", "94x94", "160x160", "320x320", "1024x1024", "2048x2048".
+     * @param {Function} successCallback - function to call with the thumbnail url. The function
+     *  will recieve null as its argument if a thumbnail could not be fetched.
+     * @param {Function} errorCallback - Function to call with errors
+     * @return {void}
+     */
+    getFileThumbnail(
+        item: BoxItem,
+        dimensions: string,
```
jstoffan: Do we plan to request different dimensions when the grid slider is implemented? If so, it seems like it'd be more efficient to re-use the 1024x1024 images and scale them up/down. In that case, the additional complexity here can be avoided.

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -383,6 +386,40 @@ class ContentExplorer extends Component<Props, State> {
...
+        const fileAPI = this.api.getFileAPI();
+
+        // not using Promise.all since thumbnails should load one at a time
+        items.forEach((item, index) => {
```
jstoffan: This seems like it'll take a while if each image is loaded serially. Can these representations be requested as a field on the `/folders/folder_item` call? If not, should we try loading them in batches?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -411,6 +448,9 @@ class ContentExplorer extends Component<Props, State> {
         } else {
             this.setState({ rootName });
         }
+        if (this.shouldFetchThumbnails(items)) {
```
jstoffan: It looks like this is repeated a few times. Can it live in `componentDidUpdate` and re-fetch based on when `currentCollection.items` changes?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -383,6 +386,40 @@ class ContentExplorer extends Component<Props, State> {
         }
     };
 
+    shouldFetchThumbnails(items: ?Array<BoxItem>) {
```
jstoffan: Can this pull `items` from state rather than having it passed in?

src/elements/content-explorer/ContentExplorer.js
jstoffan: It seems like the first part of this method should live in either the `Folder` or `File` API helper.

src/elements/content-explorer/ItemGridCell.js
```diff
@@ -24,10 +26,20 @@ const ItemGridCell = ({
     view,
     ...rest
 }: Props) => {
+    const url = getProp(item, `thumbnailUrl`);
```
jstoffan: Is the `getProp` usage here meant to address a concern that `item` is undefined? Does that affect anything else?

src/elements/content-explorer/ItemGridCell.js
```diff
@@ -24,10 +26,20 @@ const ItemGridCell = ({
...
+    const itemClassName = classNames(
+        'bce-ItemGridCell-item',
+        { 'bce-ItemGridCell-item--postLoadThumbnail': url },
```
jstoffan: Modifiers (`--modifier`) should be used for variants of a component, but these look more like descendent component names. You could also consider casting `url` to a boolean via `!!`.

src/elements/content-explorer/ItemGridCell.js
jstoffan: This is also the root element in `ItemGridCell`, so it's class name should be `.bce-ItemGridCell`.

src/elements/content-explorer/ItemGridCell.js
```diff
@@ -24,10 +26,20 @@ const ItemGridCell = ({
...
+    );
     return (
-        <figure>
+        <figure className="bce-itemGridCell-figure">
             <div className="bce-ItemGridCell-itemThumbnail">
-                <div className="bce-ItemGridCell-itemIcon">{getIcon(128, item)}</div>
+                {url ? (
+                    <div className={itemClassName} style={{ backgroundImage: `url("${url}")` }} />
+                ) : (
+                    <div className={itemClassName}> {getIcon(128, item)} </div>
```
jstoffan: Please avoid single non-breaking spaces in markup, as they can change the way it's rendered in some browsers.

src/elements/content-explorer/ItemGridCell.js
rickstefanik: `item` should always be defined. The `getProp` here is because `thumbnailUrl` may not be defined.  I check later on whether this url value is defined or not.

rickstefanik:
Continuing to address unresolved comments.

test/integration/content-explorer/ContentExplorer.int.test.js
```diff
@@ -102,6 +102,19 @@ describe('ContentExplorer', () => {
     beforeEach(() => {
         cy.server();
         cy.route('GET', '**/folders/*', 'fixture:content-explorer/root-folder.json');
+        cy.route('GET', '**/files/*', {
```
rickstefanik: The API call for thumbnails was failing, since this Cypress test is using a fixture (line above).  If the API call for thumbnails fails, an error screen appears in the BCE saying a network error occurred.  Stubbing this call seemed to be the best solution.

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -383,6 +393,44 @@ class ContentExplorer extends Component<Props, State> {
...
+     * @param {string} thumbnailUrl - url pointing towards thumbnail of item
+     * @param {number} index - index of item in currentCollection.items
+     * @return {void}
+     */
+    getFileThumbnailSuccessCallback = (thumbnailUrl: ?string, index: number) => {
+        const currentCollection = { ...this.state.currentCollection };
+        if (getProp(currentCollection, `items[${index}]`)) {
+            currentCollection.items = currentCollection.items.map((item, idx) => {
+                if (idx === index && thumbnailUrl) {
+                    return { ...item, thumbnailUrl };
```
rickstefanik: Realizing now that `item` will run into the same issue with modifying state.  This function will likely be removed soon though since the thumbnails will be fetched with the folder.

rickstefanik:
Going to make thumbnail calls along with folder.  Substantial amount of code will go away, so probably best to hold off on reviewing until then.

rickstefanik:
Made the API call change.  Code is good to be reviewed again.

src/api/Folder.js
```diff
@@ -11,7 +11,15 @@ import { getBadItemError } from '../utils/error';
...
-import { CACHE_PREFIX_FOLDER, ERROR_CODE_FETCH_FOLDER, ERROR_CODE_CREATE_FOLDER } from '../constants';
+import {
+    CACHE_PREFIX_FOLDER,
+    ERROR_CODE_FETCH_FOLDER,
+    ERROR_CODE_CREATE_FOLDER,
+    FIELD_REPRESENTATIONS,
+} from '../constants';
+
+const DEFAULT_THUMBNAIL_DIMENSIONS = '1024x1024';
+const SHOULD_FETCH_THUMBNAILS = false;
```
rickstefanik: I'm still investigating how I can use `feature-checking` to determine this value.  This should not be a constant!

flow-typed/box-ui-elements.js
```diff
@@ -472,6 +473,25 @@ type Collection = {
     totalCount?: number,
 };
 
+type RepresentationObject = {
+    entries: Array<EntryObject>,
+};
+
+type EntryObject = {
```
rickstefanik: Would `RepresentationEntryObject` be a better name?

src/api/Folder.js
```diff
@@ -207,19 +206,19 @@ class Folder extends Item {
...
     ): Promise<any> {
         if (this.isDestroyed()) {
             return Promise.reject();
         }
 
         const requestFields = fields || FOLDER_FIELDS_TO_FETCH;
-
-        this.errorCode = ERROR_CODE_FETCH_FOLDER;
-        if (SHOULD_FETCH_THUMBNAILS) {
+        if (shouldFetchThumbnails) {
```
ConradJChan: I wonder how necessary this approach is vs just adding the `FIELD_REPRESENTATIONS` to the `FOLDER_FIELDS_TO_FETCH`. Could consider doing some comparison between calls with and without representations to see how measurable this performance impact is

src/elements/content-explorer/ItemGridCell.scss
```diff
@@ -1,18 +1,3 @@
...
-    align-items: center;
-    bottom: 0;
-    display: flex;
-    justify-content: center;
-    left: 0;
-    margin: auto;
-    position: absolute;
-    right: 0;
-    top: 0;
+.bce-ItemGridCell {
```
ConradJChan: Is this still needed if the addition to `_reset.scss` was made?

src/api/Folder.js
```diff
@@ -226,6 +234,13 @@ class Folder extends Item {
             .get({
                 url: this.getUrl(this.id),
                 params,
+                headers: shouldFetchThumbnails
+                    ? {
+                          // if unable to fetch jpg thumbnail, grab png rep of first page. Certain file types do
+                          // not have a thumbnail rep but do have a first page rep.
+                          'X-Rep-Hints': `[jpg?dimensions=${DEFAULT_THUMBNAIL_DIMENSIONS},png?dimensions=1024x1024]`,
```
jstoffan: Dimensions seem to be duplicated.

src/elements/content-explorer/ItemGridThumbnail.js
```diff
@@ -0,0 +1,42 @@
...
+    item: BoxItem,
+    token: Token,
+};
+
+const ItemGridThumbnail = ({ item, token }: Props) => {
+    let url;
+    const entries = getProp(item, 'representations.entries');
+    if (getProp(entries, '[0].status.state') === 'success') {
+        const entry = getProp(entries, '[0]');
+        const rep = getProp(entry, 'representation');
```
ConradJChan: In general, we like to avoid abbreviations. Increased readability is the priority, it all gets minified by webpack anyway

src/api/Folder.js
```diff
@@ -305,7 +320,7 @@ class Folder extends Item {
         }
 
         // Make the XHR request
-        this.folderRequest();
+        this.folderRequest(options.shouldFetchThumbnails ? { shouldFetchThumbnails: true } : {});
```
jstoffan: Looks like we can do something like `this.folderRequest(options)`.

src/elements/content-explorer/ItemGridThumbnail.scss
```diff
@@ -0,0 +1,27 @@
...
+    margin-bottom: 10px;
+    // padding is used to maintain cell's aspect ratio on resize of container
+    padding-bottom: 65%;
+    position: relative;
+}
+
+.bce-ItemGridThumbnail-item {
+    bottom: 0;
+    left: 0;
+    margin: auto;
```
ConradJChan: Is this margin needed?

src/elements/content-explorer/ItemGridThumbnail.js
```diff
@@ -0,0 +1,42 @@
+// @flow
+import * as React from 'react';
+import classNames from 'classnames';
+import getProp from 'lodash/get';
+import { getIcon } from '../common/item/iconCellRenderer';
+import './ItemGridThumbnail.scss';
+
+type Props = {
+    item: BoxItem,
+    token: Token,
```
jstoffan: `Token` can be a string or a function, which the latter seems like it isn't handled below.

src/elements/content-explorer/ItemGridThumbnail.js
```diff
@@ -0,0 +1,42 @@
...
+    let url;
+    const entries = getProp(item, 'representations.entries');
+    if (getProp(entries, '[0].status.state') === 'success') {
+        const entry = getProp(entries, '[0]');
+        const rep = getProp(entry, 'representation');
+        const template = getProp(entry, 'content.url_template');
+        if (token && template && rep) {
+            // Asset path for thumbnail is simply empty string.
+            const thumbnailBaseUrl = template.replace('{+asset_path}', rep === 'jpg' ? '' : '1.png');
+            url = `${thumbnailBaseUrl}?access_token=${token}`;
```
jstoffan: It doesn't seem like `token` should be passed this deeply, since it's meant to be used for data access. Can the URL be passed in from a higher level component?

src/elements/content-explorer/ItemGridThumbnail.js
```diff
@@ -0,0 +1,42 @@
...
+        }
+    }
+
+    const className = classNames(
+        'bce-ItemGridThumbnail-item',
+        { 'bce-ItemGridThumbnail-postLoadThumbnail': !!url },
```
jstoffan: These should probably live directly on their respective divs, since we already have a `url` condition there.

src/elements/content-explorer/ItemGridThumbnail.scss
```diff
@@ -0,0 +1,27 @@
...
+    position: absolute;
+    right: 0;
+    top: 0;
+}
+
+.bce-ItemGridThumbnail-itemIcon {
```
jstoffan: Do these styles need to be separate or can they all simply live on `-item` together?

rickstefanik:
Still investigating removing `ContentExplorer.contextType = FeatureContext` and unresolved comments.

src/api/Folder.js
rickstefanik: I ran 10 trials with and without the reps field.  Average time without the reps field was `2.00s`, average time with the reps field was `2.52s`.

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -393,11 +434,11 @@ class ContentExplorer extends Component<Props, State> {
      */
     fetchFolderSuccessCallback(collection: Collection, triggerNavigationEvent: boolean): void {
         const { onNavigate, rootFolderId }: Props = this.props;
-        const { id, name, boxItem }: Collection = collection;
+        const { boxItem, id, name }: Collection = collection;
         const { selected }: State = this.state;
         const rootName = id === rootFolderId ? name : '';
 
-        this.updateCollection(collection, selected);
+        this.updateCollection(collection, selected, () => this.assignThumbnailUrls());
```
jstoffan: The anonymous function wrapper isn't necessary here.

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -435,6 +476,7 @@ class ContentExplorer extends Component<Props, State> {
         const hasFolderChanged = currentId && currentId !== folderId;
         const hasSearchQuery = !!searchQuery.trim().length;
         const offset = hasFolderChanged || hasSearchQuery ? 0 : currentOffset; // Reset offset on folder or mode change
+        const shouldFetchThumbnails = getProp(features, 'contentExplorer.gridView.enabled');
```
jstoffan: I think there is a `isFeatureEnabled` helper available in `feature-checking/util`.

src/styles/_reset.scss
```diff
@@ -47,6 +47,10 @@ td {
     padding: 0;
 }
 
+figure {
```
jstoffan: This may lead to some unexpected styling in our consuming apps, as this stylesheet is widely used. Can we move the override to just where it's needed?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -383,6 +392,48 @@ class ContentExplorer extends Component<Props, State> {
...
+                    return newItem;
+                });
+                this.setState({ currentCollection: newCollection });
+            }
+        });
+    }
+
+    assignCollectionThumbnailUrls(): void {
+        const { currentCollection } = this.state;
+        if (currentCollection.items) {
```
ConradJChan: Instead of iterating over the collection.items and for each item creating a newCollection and setting that to the state, could we instead `.map` the currentCollection to add the thumbnailUrl and then finally set to state in `assignCollectionThumbnailUrls`?

src/api/Folder.js
```diff
@@ -90,6 +104,64 @@ class Folder extends Item {
...
+     * Takes representations fields from items in collection and assigns thumbnailUrls
+     * to items, with the use of the callback
+     *
+     * @param {Collection} collection - collection with items that have representations field populated
+     * @param {Token} token - access_token
+     * @param {(item: BoxItem, index: number, thumbnailUrl: string) => void} callback - Function to call with generated thumbnailUrl
+     * @return {void}
+     */
+    formatRepresentations(
+        collection: Collection,
```
ConradJChan: could destructure the items out of the collection since we're not using collection anywhere

src/api/Folder.js
```diff
@@ -90,6 +104,64 @@ class Folder extends Item {
...
+        token: Token,
+        callback: (item: BoxItem, index: number, thumbnailUrl: string) => void,
+    ): void {
+        if (collection.items) {
+            collection.items.forEach(async (item, index) => {
+                const entries = getProp(item, 'representations.entries');
```
ConradJChan: Might be helpful to store the state in a const for debugging purposes

src/api/Folder.js
```diff
@@ -90,6 +104,64 @@ class Folder extends Item {
...
+                if (getProp(entries, '[0].status.state') === 'success') {
+                    const { representation, template } = this.getDataFromEntries(entries, 0);
+                    if (!!template && !!representation) {
+                        const tokenString = await TokenService.getToken(getTypedFileId(item.id), token);
```
ConradJChan: I think it's possible for the token to come back as an object as well, see https://github.com/box/box-ui-elements/blob/e282d5f53e3d0e3c15a728e9b113888269acfea6/src/utils/TokenService.js#L64

src/api/Folder.js
```diff
@@ -90,6 +104,64 @@ class Folder extends Item {
...
+     *
+     * @param {Collection} collection - collection with items that have representations field populated
+     * @param {Token} token - access_token
+     * @param {(item: BoxItem, index: number, thumbnailUrl: string) => void} callback - Function to call with generated thumbnailUrl
+     * @return {void}
+     */
+    formatRepresentations(
+        collection: Collection,
+        token: Token,
+        callback: (item: BoxItem, index: number, thumbnailUrl: string) => void,
```
ConradJChan: Default to `noop`?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -383,6 +389,27 @@ class ContentExplorer extends Component<Props, State> {
...
 
+    /**
+     * Updates item in currentCollection with the given thumbnailUrl
+     *
+     * @param {BoxItem} item - item that should have thumbnailUrl field set
+     * @param {number} index - index of item in this.state.currentCollection
+     * @param {string} thumbnailUrl - thumbnailUrl to populate field with
+     * @return {void}
+     */
+    assignItemThumbnailUrl = (item: BoxItem, index: number, thumbnailUrl: string): void => {
```
ConradJChan: Is the `index` necessary? Could we use the `item.id` instead?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -383,6 +389,27 @@ class ContentExplorer extends Component<Props, State> {
...
+        const newCollection = { ...this.state.currentCollection };
+        newCollection.items = newCollection.items.map((newItem, newIndex) => {
```
ConradJChan: Could also do this in one line `(newIndex === index) ? {...cloneDeep(item), thumbnailUrl} : newItem`

src/api/Folder.js
```diff
@@ -108,28 +106,22 @@ class Folder extends Item {
...
-        collection: Collection,
-        token: Token,
-        callback: (item: BoxItem, index: number, thumbnailUrl: string) => void,
+        items: ?Array<BoxItem>,
+        callback: (item: BoxItem, representation: ?string, template: ?string) => Promise<void> = () =>
+            Promise.resolve(),
     ): void {
-        if (collection.items) {
-            collection.items.forEach(async (item, index) => {
+        const itemsToBeFormatted = items;
```
jstoffan: Please avoid rename-only reassignments.

i18n/en-US.properties
```diff
@@ -78,6 +78,16 @@ be.contentSidebar.activityFeed.comment.taskDeleteMenuItem = Delete task
 be.contentSidebar.activityFeed.comment.taskDeletePrompt = Are you sure you want to permanently delete this task?
 # Text to show on menu item to edit task
 be.contentSidebar.activityFeed.comment.taskEditMenuItem = Modify task
+# Title for x icon indicating someone rejected a task
```
rickstefanik: trying to remove this from the PR.  It got added after squashing for some reason.

rickstefanik: removed and pushed with `--no-verify`.  Think that worked.

src/elements/content-explorer/__tests__/ContentExplorer-test.js
```diff
@@ -42,16 +43,97 @@ describe('elements/content-explorer/ContentExplorer', () => {
...
+    describe('makeThumbnailUrl()', () => {
+        const template = 'baseUrl/{+asset_path}';
+        const token = 'token';
+        let instance;
+        beforeEach(() => {
+            instance = getWrapper().instance();
+        });
+
+        test('should properly make thumbnail url with png', () => {
+            expect(instance.makeThumbnailUrl('png', template, token)).toBe('baseUrl/1.png?access_token=token');
```
rickstefanik: Not sure if the `toBe` should be hardcoded like it is, or if it should be something like <pre>.toBe(\`${baseUrl}/1.png?access_token=${token}\`)</pre> 
Not sure which is the right philosophy for this type of test.

ConradJChan: Both are acceptable but I personally would prefer to make that connection between the input and the expectation so that if the inputs change later the tests properly reflect it

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -383,6 +391,52 @@ class ContentExplorer extends Component<Props, State> {
...
+     * @param {BoxItem} item - item that should have thumbnailUrl field set
+     * @param {?string} representation - representation from item's representations field
+     * @param {?string} template -  template from item's representations field
+     * @return {Promise<void>}
+     */
+    assignItemThumbnailUrl = async (item: BoxItem, representation: ?string, template: ?string): Promise<void> => {
+        const { token } = this.props;
+        if (!!template && !!representation) {
+            const tokenResult = await TokenService.getToken(getTypedFileId(item.id), token);
+            if (
```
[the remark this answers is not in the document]
rickstefanik: I don't think just checking `!tokenResult` would be sufficient, since `getToken` can return an object without a `.read` field.  (If the return token object only has a `.write` field, I'm not sure that we can use that to get the thumbnails.)  I was able to simplify the logic though.

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -383,6 +392,46 @@ class ContentExplorer extends Component<Props, State> {
...
+     * @param {BoxItem} item - item that should have thumbnailUrl field set
+     * @param {?string} representation - representation from item's representations field
+     * @param {?string} template -  template from item's representations field
+     * @return {Promise<void>}
+     */
+    assignItemThumbnailUrl = async (item: BoxItem, representation: ?string, template: ?string): Promise<void> => {
+        const { token } = this.props;
+        if (!!template && !!representation) {
+            const tokenResult = await TokenService.getToken(getTypedFileId(item.id), token);
+            const tokenString = typeof tokenResult === 'string' ? tokenResult : getProp(tokenResult, 'read');
```
jstoffan: It looks like this logic is already present in `TokenService.getReadToken`. Can it be reused here?

src/api/Folder.js
```diff
@@ -305,7 +320,7 @@ class Folder extends Item {
         }
 
         // Make the XHR request
-        this.folderRequest();
+        this.folderRequest(options);
```
jstoffan: Looks like this is no longer needed?

src/api/Folder.js
```diff
@@ -205,13 +216,10 @@ class Folder extends Item {
         if (this.isDestroyed()) {
             return Promise.reject();
         }
-
```
jstoffan: More space between unrelated statements/blocks is usually better than less.

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -391,12 +399,24 @@ class ContentExplorer extends Component<Props, State> {
...
         const { selected }: State = this.state;
         const rootName = id === rootFolderId ? name : '';
 
+        if (isFeatureEnabled(features, 'contentExplorer.gridView.enabled')) {
+            const { items = [] } = collection;
+            const fileAPI = this.api.getFileAPI();
+            const thumbnailUrlPromises = items.map(item => fileAPI.getThumbnailUrl(item));
+
+            await Promise.all(thumbnailUrlPromises).then(thumbnailUrls => {
+                collection.items = items.map((item, index) => {
```
jstoffan: We lose a lot of the benefit of async/await and Promises by mixing them; refactoring to use one or the other should remove the need to modify `collection` directly. Here's one potential solution using await/async:

```
const { items } = collection;
const itemThumbnails = await Promise.all(thumbnailUrlPromises);
const itemsWithThumbnails = items.map((item, index) => {
    const thumbnailUrl = thumbnailUrls[index];
    return thumbnailUrl ? { ...cloneDeep(item), thumbnailUrl } : item;
});
this.updateCollection({ ...collection, items: itemsWithThumbnails }, selected);
```

src/api/File.js
```diff
@@ -77,6 +78,20 @@ class File extends Item {
         this.successHandler(`${downloadBaseUrl}?${downloadUrlQuery}`);
     }
 
+    async getThumbnailUrl(item: BoxItem): Promise<?string> {
+        const entry = getProp(item, 'representations.entries[0]');
+        const extension = getProp(entry, 'representation');
+        const template = getProp(entry, 'content.url_template');
+        const token = await TokenService.getReadToken(getTypedFileId(item.id), this.options.token);
```
ConradJChan: Should this be moved after the if check on L87?

src/api/File.js
```diff
@@ -77,6 +78,20 @@ class File extends Item {
         this.successHandler(`${downloadBaseUrl}?${downloadUrlQuery}`);
     }
 
+    async getThumbnailUrl(item: BoxItem): Promise<?string> {
```
ConradJChan: Should the return type be updated?

src/api/__tests__/Folder-test.js
```diff
@@ -51,120 +51,120 @@ describe('api/Folder', () => {
...
-                folder.formatRepresentations([item1, item2], callback);
-                expect(folder.getDataFromEntries.mock.calls).toEqual([[entries1, 0], [entries2, 0]]);
-                expect(callback.mock.calls).toEqual([
-                    [item1, 'representation1', 'template1'],
-                    [item2, 'representation2', 'template2'],
-                ]);
-            });
-        });
-    });
+    // describe('getDataFromEntries()', () => {
```
ConradJChan: Clean up?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -440,18 +399,25 @@ class ContentExplorer extends Component<Props, State> {
      * @param {Boolean|void} triggerNavigationEvent - To trigger navigate event and focus grid
      * @return {void}
      */
-    fetchFolderSuccessCallback(collection: Collection, triggerNavigationEvent: boolean): void {
+    async fetchFolderSuccessCallback(collection: Collection, triggerNavigationEvent: boolean): Promise<void> {
```
ConradJChan: Still returns void right?

src/elements/content-explorer/__tests__/ContentExplorer-test.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import { mount } from 'enzyme';
 import { ContentExplorerComponent as ContentExplorer } from '../ContentExplorer';
-import TokenService from '../../../utils/TokenService';
+// import TokenService from '../../../utils/TokenService';
```
ConradJChan: Cleanup

src/api/__tests__/File-test.js
```diff
@@ -95,6 +97,76 @@ describe('api/File', () => {
...
+                            url_template,
+                        },
+                    },
+                ],
+            },
+        };
+
+        let item;
+        beforeEach(() => {
+            item = cloneDeep(baseItem);
```
rickstefanik: Is this the best way to do this? The properties I need to change in `item` are deeply nested, and I thought this would be cleaner than declaring a new item for every case.

src/elements/content-explorer/__tests__/ContentExplorer-test.js
```diff
@@ -42,16 +43,113 @@ describe('elements/content-explorer/ContentExplorer', () => {
...
+
+        let wrapper;
+        let instance;
+
+        test('thumbnail url should not be assigned to item if grid view is not enabled', () => {
+            wrapper = getWrapper();
+            instance = wrapper.instance();
+            instance.api = { getFileAPI };
+            instance.setState = jest.fn();
+            instance.closeModals = jest.fn();
```
rickstefanik: Is it useful it to check functions like this even though they are not specific to what is being tested?

src/api/__tests__/File-test.js
ConradJChan: Another approach would be to have a `createTestItem` method that takes in arguments for what you want to change

src/api/File.js
```diff
@@ -77,6 +78,26 @@ class File extends Item {
...
+        const extension = getProp(entry, 'representation');
+        const template = getProp(entry, 'content.url_template');
+        const token = await TokenService.getReadToken(getTypedFileId(item.id), this.options.token);
+
+        if (getProp(entry, 'status.state') !== 'success' || !extension || !template || !token) {
+            return null;
+        }
+
+        const thumbnailBaseUrl = template.replace('{+asset_path}', extension === 'jpg' ? '' : '1.png');
+        return `${thumbnailBaseUrl}?access_token=${token}`;
```
jstoffan: Consider parsing the original tempalte url using `queryString.parseUrl` to retain any existing query parameters and also `queryString.stringify` to produce the final string. Take a look at `getDownloadUrl` for an example.

src/api/File.js
```diff
@@ -77,6 +78,26 @@ class File extends Item {
...
+     * @return {Promise<?string>} - the url for the item's thumbnail, or null
+     */
+    async getThumbnailUrl(item: BoxItem): Promise<?string> {
+        const entry = getProp(item, 'representations.entries[0]');
+        const extension = getProp(entry, 'representation');
+        const template = getProp(entry, 'content.url_template');
+        const token = await TokenService.getReadToken(getTypedFileId(item.id), this.options.token);
+
+        if (getProp(entry, 'status.state') !== 'success' || !extension || !template || !token) {
+            return null;
```
jstoffan: Can probably just `return;` here unless `null` is meaningful as a result.

src/elements/content-explorer/__tests__/ContentExplorer-test.js
ConradJChan: It can be, as an indication of the code paths which are being executed under test. In this case, maybe not so much since it's always called no matter what. But stubbing I would still recommend because we wouldn't want any side effects of calling that method to affect our test

src/api/Folder.js
rickstefanik: This actually is still needed in the current version of the code, since `ContentExplorer` passes `fields` into `options`.  (Previously, `fields` was not defined in `options` and defaulted to `FOLDER_FIELDS_TO_FETCH`, but now, since `ContentExplorer` determines the fields based on `features`, I think this is needed.)

PR #1415, opened 2019-07-16
build(scripts): Build and update all files prior to push

booc0mtaco:
small change to the prepush script which prevents building against stale files that are `.gitignore`d. (e.g., i18n/json and i18n/*.js)

- run all the bundle builds prior to tests and localization updates
- remove duplicate call to `yarn build:i18n` (which would build translations against the contents of es/ *before* updating it)

Tested via:

- populating en-US.js, messages.json, and es/**/messages.js with dummy data to make sure they were updated prior to the uncommited file check
- same as № 1 but adding a new message
- same as № 1 and № 2, but also removing the newly-added message to make sure .properties got updated properly


boxcla:

PR #1419, opened 2019-07-16
refactor: move from BOX_BLUE to bdlBoxBlue

booc0mtaco:
Add initial conversion script, and use it to batch update colors away from hard-coded box blue values

`find . -name "*.js" -exec ../scripts/utils/swapColors.sh {} BOX_BLUE bdlBoxBlue \;`

### DIFF
--- a/scripts/utils/swapColor.sh
+++ b/scripts/utils/swapColor.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+###
+# Simple script to replace a single color value across multiple files
+###
+
+red=$"\n\e[1;31m(✖) "
+end=$"\e[0m\n"
+
+if [ ! $# -eq 3 ]; then
+    echo "${red}You must specify the file, fromColor and toColor value${end}"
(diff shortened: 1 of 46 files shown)


boxcla:

scripts/utils/swapColor.sh
```diff
@@ -0,0 +1,17 @@
...
+
+red=$"\n\e[1;31m(✖) "
+end=$"\e[0m\n"
+
+if [ ! $# -eq 3 ]; then
+    echo "${red}You must specify the file, fromColor and toColor value${end}"
+    exit 1
+fi
+
+sed -i '' -e "s/$2/$3/g" $1
```
jstoffan: Is there anything color-specific in here or is this a more generic find/replace script?

booc0mtaco: right now it's just find/replace. I do have a list of conversion values that could be embedded here, but waiting to incorporate that until i get the usage patterns just right

PR #1422, opened 2019-07-17
refactor(versions): Move version api helper methods to a separate class

jstoffan:
The versions container was getting a bit large, so it seemed like a good idea to extract the data access layer to its own class.

### DIFF
--- a/src/api/Versions.js
+++ b/src/api/Versions.js
@@ -338,25 +338,6 @@ class Versions extends OffsetBasedAPI {
             errorCallback,
         });
     }
-
-    /**
-     * Helper to sort versions by their created date
-     *
-     * @param {FileVersions} versions - API returned file versions for this file
-     * @return {FileVersions} sorted versions array
-     */
-    sortVersions(versions: ?FileVersions): ?FileVersions {
-        if (!versions) {
-            return versions;
-        }
-
-        const { entries, total_count } = versions;
-
-        return {
-            entries: [...entries].sort((a, b) => Date.parse(b.created_at) - Date.parse(a.created_at)),
-            total_count,
-        };
-    }
 }
 
 export default Versions;
--- a/src/elements/content-sidebar/versions/VersionsSidebarAPI.js
+++ b/src/elements/content-sidebar/versions/VersionsSidebarAPI.js
@@ -70,12 +70,12 @@ export default class VersionsSidebarAPI {
     deleteVersion = (version: ?BoxItemVersion): Promise<null> => {
         const { id: versionId, permissions = {} } = version || {};
 
-        return new Promise((successCallback, errorCallback) =>
+        return new Promise((resolve, reject) =>
             this.api.getVersionsAPI(false).deleteVersion({
                 fileId: this.fileId,
                 permissions,
-                successCallback,
-                errorCallback,
+                successCallback: resolve,
+                errorCallback: reject,
                 versionId,
             }),
         );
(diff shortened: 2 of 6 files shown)


boxcla:

src/elements/content-sidebar/versions/VersionsSidebarAPI.js
```diff
@@ -0,0 +1,111 @@
...
+        this.api = api;
+        this.fileId = fileId;
+    }
+
+    fetchData = (): Promise<fetchPayload> => {
+        return Promise.all([this.fetchFile(), this.fetchVersions()]).then(this.fetchVersionCurrent);
+    };
+
+    fetchDownloadUrl = (version: ?BoxItemVersion): Promise<string> => {
+        return new Promise((resolve, reject) => {
```
ConradJChan: Noticed below you're using `successCallback` and `errorCallback` in lieu of `resolve/reject`, should we make them consistent?

src/elements/content-sidebar/versions/VersionsSidebarAPI.js
```diff
@@ -0,0 +1,111 @@
...
+                return reject(new Error('Could not find requested version'));
+            }
+
+            return this.api.getFileAPI().getDownloadUrl(this.fileId, version, resolve, reject);
+        });
+    };
+
+    fetchFile = (): Promise<BoxItem> => {
+        return new Promise((resolve, reject) =>
+            this.api.getFileAPI().getFile(this.fileId, resolve, reject, {
```
ConradJChan: Should we update `getFileAPI` to also take the `shouldDestroy` param? As is, would this potentially cause any issues like we saw previously where promises would get unresolved?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -67,11 +69,17 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
...
+        this.fetchData();
     }
 
-    componentDidUpdate({ versionId: prevVersionId }: Props) {
-        const { versionId } = this.props;
+    componentDidUpdate({ fileId: prevFileId, versionId: prevVersionId }: Props) {
+        const { fileId, versionId } = this.props;
+
+        if (fileId !== prevFileId) {
+            this.initialize();
```
ConradJChan: This should be the same instance of the APIFactory so there isn't any cleanup required right?

src/elements/content-sidebar/versions/VersionsSidebarAPI.js
jstoffan: Possibly, but that seems a bit beyond the scope of this PR.

src/elements/content-sidebar/versions/VersionsSidebarAPI.js
jstoffan: Sure, I can update these to `successCallback` and `errorCallback`.

src/api/Versions.js
```diff
@@ -338,25 +338,6 @@ class Versions extends OffsetBasedAPI {
             errorCallback,
         });
     }
-
```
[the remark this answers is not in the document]
jstoffan: The `format` helper is part of the `/api/Base` class and is meant to format the API response. I think `addPermissions` is a candidate, but it's also not hurting anything staying put.

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
jstoffan: Yeah, it's just meant to update the helper's `fileId` property. I could make a setter for it rather than re-instantiating the class. Thoughts?

src/elements/content-sidebar/versions/VersionsSidebarAPI.js
ConradJChan: Yeah definitely 👍

PR #1423, opened 2019-07-17
chore(bdl-colors): add script to detect and convert non-BDL color values

booc0mtaco:
Adding better script to help detect and replace obsolete colors in JS and SCSS files. *note*: postponing using a CLI parser for now

```
Usage: bdlColorManager operation fileName [--verbose]

Supported operations:
        check           check to see if a file contains deprecated values (returns error if it does)
        swap            replace known-deprecated variable names in a JS or SCSS file
```


boxcla:

booc0mtaco:
I attempted to write in a way so we could potentially incorporate into our linter or elsewhere (a prelude to fully deprecating the old values)

booc0mtaco:
Need to add a few pieces I missed for one use case

booc0mtaco:
@jstoffan can you re-review, please? I forgot a few pieces to handle SCSS, so added those in.

Issue #1412, opened 2019-07-16
View classification label in sharing modal


feixuanli:
**Describe the solution you'd like**
As an end user, I would like to view the classification for a file or folder when I am trying to share it using the Unified Sharing Model so that is obvious to me I am sharing a sensitive filter.

sahiga:
Hi @feixuanli, thanks for the PR. Can we close this issue?

PR #1425, opened 2019-07-18
feat(unified-share-modal): show classification for the file

feixuanli:
issue #1412

### DIFF
--- a/src/features/unified-share-modal/UnifiedShareModal.js
+++ b/src/features/unified-share-modal/UnifiedShareModal.js
@@ -52,6 +52,7 @@ type Props = {
     changeSharedLinkPermissionLevel: (
         newPermissionLevel: permissionLevelType,
     ) => Promise<{ permissionLevel: permissionLevelType }>,
+    children: React.Node,
     /** If item is classified this property contains the classification name */
     classificationName?: string,
     /** Message warning about restrictions regarding inviting collaborators to the item */
@@ -685,6 +686,7 @@ class UnifiedShareModal extends React.Component<Props, State> {
         const {
             changeSharedLinkAccessLevel,
             changeSharedLinkPermissionLevel,
+            children,
             classificationName,
             focusSharedLinkOnLoad,
             item,
@@ -707,6 +709,7 @@ class UnifiedShareModal extends React.Component<Props, State> {
             sendSharedLinkError,
             trackingProps,
         } = rest;
+        const classificationBtn = children;
         const {
             modalTracking,
             sharedLinkTracking,
@@ -736,7 +739,12 @@ class UnifiedShareModal extends React.Component<Props, State> {
                     className="unified-share-modal"
                     isOpen={isConfirmModalOpen ? false : isOpen}
                     onRequestClose={submitting ? undefined : onRequestClose}
-                    title={<span>{this.renderModalTitle()}</span>}
+                    title={
+                        <span className="modal-title-label">
+                            {this.renderModalTitle()}
+                            {classificationBtn}
+                        </span>
+                    }
                     {...extendedModalProps}
                 >
                     <LoadingIndicatorWrapper isLoading={isFetching} hideContent>
--- a/src/features/unified-share-modal/UnifiedShareModal.js
+++ b/src/features/unified-share-modal/UnifiedShareModal.js
@@ -15,6 +15,7 @@ import { UpgradeBadge } from '../../components/badge';
 import { ITEM_TYPE_WEBLINK } from '../../common/constants';
 import Tooltip from '../../components/tooltip';
 import { CollaboratorAvatars, CollaboratorList } from '../collaborator-avatars';
+import Classification from '../classification';
 
 import InviteePermissionsMenu from './InviteePermissionsMenu';
 import messages from './messages';
@@ -52,7 +53,7 @@ type Props = {
     changeSharedLinkPermissionLevel: (
         newPermissionLevel: permissionLevelType,
     ) => Promise<{ permissionLevel: permissionLevelType }>,
-    children: React.Node,
+    classification: ClassificationInfo,
     /** If item is classified this property contains the classification name */
     classificationName?: string,
     /** Message warning about restrictions regarding inviting collaborators to the item */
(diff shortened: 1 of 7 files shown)


boxcla:

src/features/unified-share-modal/UnifiedShareModal.js
```diff
@@ -52,6 +52,7 @@ type Props = {
     changeSharedLinkPermissionLevel: (
         newPermissionLevel: permissionLevelType,
     ) => Promise<{ permissionLevel: permissionLevelType }>,
+    children: React.Node,
```
priyajeet: I would not call it children, since thats a generic for all/any children.
Probably should be `classification?: ClassificationInfo` and should use https://github.com/box/box-ui-elements/tree/master/src/features/classification

src/features/unified-share-modal/UnifiedShareModal.js
```diff
@@ -734,7 +737,12 @@ class UnifiedShareModal extends React.Component<Props, State> {
                     className="unified-share-modal"
                     isOpen={isConfirmModalOpen ? false : isOpen}
                     onRequestClose={submitting ? undefined : onRequestClose}
-                    title={<span>{this.renderModalTitle()}</span>}
+                    title={
```
Sunyang730: Might also be good to pull this out so it doesn't add more clutter to the title prop

priyajeet: Maybe a new title component in its own file

src/features/unified-share-modal/UnifiedShareModal.js
feixuanli: updated branch! yet I didn't put the classification as optional, because it's passed in as object from upstream

src/features/unified-share-modal/UnifiedShareModal.js
feixuanli: The reason for putting them together is to get the label rendered aligned, and avoid updating ModalDialog component in the even downstream. I put aside a method before passing them in the Modal. 
@priyajeet is that what you mean by "a new title component in its own file" ?

priyajeet: A separate component in its own file if possible. The whole renderModalTitle() maybe can be moved there?

feixuanli:
updated

src/features/unified-share-modal/HeaderTitleSection.js
```diff
@@ -0,0 +1,68 @@
...
+                    itemName: name,
+                }}
+            />
+        );
+    }
+
+    return title;
+};
+
+const HeaderTitleSection = ({ classification, isEmailLinkSectionExpanded, showCollaboratorList, item }: Props) => {
```
priyajeet: Can we just call this (and file name) `HeaderTitle` since suffixing it with Section doesn't add much value.

src/features/unified-share-modal/HeaderTitleSection.js
```diff
@@ -0,0 +1,68 @@
...
+import type { item as ItemType } from './flowTypes';
+
+type Props = {
+    classification: ClassificationInfo,
+    isEmailLinkSectionExpanded: boolean,
+    item: ItemType,
+    showCollaboratorList: boolean,
+};
+
+const renderModalTitle = (isEmailLinkSectionExpanded, showCollaboratorList, item) => {
```
priyajeet: maybe just getTitle()

priyajeet: Can also be a simple non arrow function

src/features/unified-share-modal/HeaderTitleSection.js
```diff
@@ -0,0 +1,68 @@
...
+
+const HeaderTitleSection = ({ classification, isEmailLinkSectionExpanded, showCollaboratorList, item }: Props) => {
+    const title = renderModalTitle(isEmailLinkSectionExpanded, showCollaboratorList, item);
+    const { advisoryMessage, name } = classification;
+
+    return (
+        <span className="modal-title-label">
+            {title}
+            {advisoryMessage && (
+                <span className="EditClassificationButton">
```
priyajeet: Is this advisoryMessage check and wrapping span needed?
Can you just have this below the title

`<Classification advisoryMessage={advisoryMessage} messageStyle="tooltip" name={name} />`

feixuanli: I put this advisoryMessage check here is to avoid rendering the box if there is no classification assigned, and there is a margin-left style set on "EditClassificationButton" span.

src/features/unified-share-modal/HeaderTitleSection.js
feixuanli: thanks, will update!

src/features/unified-share-modal/HeaderTitleSection.js
priyajeet: advisory message is not the classification though. Its the classification `name` that holds that value. The component internally takes care of what to show or not show, so check is not really that necessary. If you remove it, it should still work or at least have the name check. Or you can use the ClassifiedBadge component instead, that just shows the badge.

If you want styling, then use a class specific to this component. We are also not editing classification in this case.

feixuanli: Thanks a lot! branch updated.

src/features/unified-share-modal/HeaderTitle.js
```diff
@@ -0,0 +1,69 @@
...
+
+    return title;
+}
+
+const HeaderTitle = ({ classification, isEmailLinkSectionExpanded, showCollaboratorList, item }: Props) => {
+    const title = getTitle(isEmailLinkSectionExpanded, showCollaboratorList, item);
+    const { advisoryMessage, name } = classification;
+
+    return (
+        <span className="modal-title-label">
```
priyajeet: Follow the SUIT css naming convention for this file, so `<span className="bdl-HeaderTitle">` and the classification section below `className="bdl-HeaderTitle-classification"`

Though thinking about it, bdl-HeaderTitle seems to give no information about it being part of usm.

feixuanli: can we have the usm as midifierName `bdl-HeaderTitle--usm` and `bdl-HeaderTitle-classification--usm` ?

priyajeet: Thats generally not the use case for modifiers. Modifiers are more like `button` to `button--plain`, that just modify the styles a bit. Options for this would be

1. usm-HeaderTitle, usm-HeaderTitle-classification
2. bdl-UnifiedSharedModal-headerTitle, bdl-UnifiedSharedModal-headerTitleClassification
3. bdl-UnifiedSharedModalHeaderTitle, bdl-UnifiedSharedModalHeaderTitle-classification

For [3] also rename the file name to match.

src/features/unified-share-modal/UnifiedShareModal.scss
```diff
@@ -50,6 +50,16 @@
 }
 
 .unified-share-modal {
+    /** title and classifiction label */
+    .bdl-UnifiedShareModalTitle {
+        display: flex;
+        flex-flow: row wrap;
+
+        .bdl-UnifiedShareModalTitle-classification {
```
priyajeet: With suit you dont have to nest that much btw, you can have this one one level above like

.bdl-UnifiedShareModalTitle {
    display: flex;
    flex-flow: row wrap;
}

.bdl-UnifiedShareModalTitle-classification {
    margin-left: 6px;
}

PR #1426, opened 2019-07-18
docs(colors): add a11y guidelines

alexkrolick:


boxcla:

booc0mtaco:
At a glance, i don't know what that information means or why it's true. If anything, I'd prefer we link to something [like the guidelines themselves](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html) since we didn't define the ratios.

alexkrolick:
@booc0mtaco is this more clear?

Screen Shot 2019-07-18 at 1 08 37 PM

PR #1432, opened 2019-07-19
build: cleanup and optimizations

priyajeet:
- Uses the same cypress docker image for all jobs since it has all the necessary tools needed for all jobs. So cache can get re-used. Can be seen here https://github.com/cypress-io/cypress-docker-images/blob/master/base/10.16.0/Dockerfile
- Combines the unit tests and build jobs since they combined still take less than the e2e job which is the bottleneck. Frees up a container.
- Switches commit lint script to use node instead of JQ for json parsing as the latter may not be available on all docker images
- Gives names to job tasks

### DIFF
--- a/scripts/commitlint.sh
+++ b/scripts/commitlint.sh
@@ -7,9 +7,11 @@ set -o errexit
 if [ -n "${CIRCLE_PULL_REQUEST}" ]
 then
   PULL_REQUEST_DETAILS=$(curl https://api.github.com/repos/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}/pulls/${CIRCLE_PR_NUMBER})
-  BASE_SHA1=$(echo "${PULL_REQUEST_DETAILS}" | jq -r .base.sha)
-  HEAD_SHA1=$(echo "${PULL_REQUEST_DETAILS}" | jq -r .head.sha)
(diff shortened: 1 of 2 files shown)


boxcla:

scripts/commitlint.sh
```diff
@@ -7,9 +7,11 @@ set -o errexit
 if [ -n "${CIRCLE_PULL_REQUEST}" ]
 then
   PULL_REQUEST_DETAILS=$(curl https://api.github.com/repos/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}/pulls/${CIRCLE_PR_NUMBER})
-  BASE_SHA1=$(echo "${PULL_REQUEST_DETAILS}" | jq -r .base.sha)
-  HEAD_SHA1=$(echo "${PULL_REQUEST_DETAILS}" | jq -r .head.sha)
+  BASE_SHA1=$(node -pe 'JSON.parse(process.argv[1]).base.sha' "${PULL_REQUEST_DETAILS}")
+  echo "Base SHA1: $BASE_SHA1"
```
ConradJChan: extra `echo`'s?

priyajeet: Just for some info to be listed while it runs

ConradJChan: Ok just wanted to check because it should be printed out on L15 as part of the `COMMIT_RANGE` too right?

PR #1434, opened 2019-07-19
fix(mention-selector): Fix hasReceivedFirstInteraction logic

ConradJChan:
The main changes here:
* Detect when we transition to a new empty `EditorState` which indicates that we should reset the `hasReceivedFirstInteraction` state so that error validation should not be applied
* Change the `hasReceivedFirstInteraction` to `true` when we transition from empty EditorState to having some text instead of `onBlur`

DraftJSMentionSelector

TODO:
 - [x] Unit tests
 - [x] e2e tests


boxcla:

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelector.js
```diff
@@ -87,15 +87,47 @@ class DraftJSMentionSelector extends React.Component<Props, State> {
      * @returns {void}
      */
     componentWillReceiveProps(nextProps: Props) {
```
priyajeet: Change this to componentDidMount while you are at it, unless it causes obvious probs.

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelector.js
```diff
@@ -87,15 +87,47 @@ class DraftJSMentionSelector extends React.Component<Props, State> {
...
-        const { contacts } = nextProps;
+        const { editorState } = this.props;
+        const { contacts, editorState: nextEditorState } = nextProps;
+        let nextState = {};
 
         if (contacts) {
-            this.setState({ contacts }, () => {
+            nextState = { ...nextState, contacts };
```
priyajeet: ...nextState isn't doing anything since its defined couple of lines above.
nextState = { contacts };

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelector.js
```diff
@@ -87,15 +87,47 @@ class DraftJSMentionSelector extends React.Component<Props, State> {
...
+            } else if (this.isEditorStateEmpty(editorState) && !this.isEditorStateEmpty(nextEditorState)) {
+                // Detect case where controlled EditorState has been made dirty
+                // If the current editorState is empty and the next editorState is not
+                // empty then this is the first interaction so mark this component dirty
+                nextState = { ...nextState, hasReceivedFirstInteraction: true };
+            }
+        }
+
+        if (Object.keys(nextState).length !== 0) {
+            this.setState(nextState, () => {
```
priyajeet: this.setState(nextState, this.checkValidity)

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelector.js
```diff
@@ -87,15 +87,47 @@ class DraftJSMentionSelector extends React.Component<Props, State> {
...
+            nextState = { ...nextState, contacts };
+        }
+
+        // Only check if operating in the mode where EditorState is received as a
+        // prop vs internalEditorState stored in state
+        if (editorState) {
+            // Detect case where controlled EditorState is created anew and empty.
+            // If next editorState is empty and the current editorState is not empty
+            // that means it is a new empty state and this component should not be marked dirty
+            if (this.isEditorStateEmpty(nextEditorState) && !this.isEditorStateEmpty(editorState)) {
```
priyajeet: Maybe make these constants, might be a wee bit more readable.

isCurrentEditorStateEmpty
isNextEditorStateEmpty

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelector.js
```diff
@@ -87,15 +87,47 @@ class DraftJSMentionSelector extends React.Component<Props, State> {
...
+        }
+
+        if (Object.keys(nextState).length !== 0) {
+            this.setState(nextState, () => {
                 this.checkValidity();
             });
         }
     }
 
+    isEditorStateEmpty = (editorState: EditorState): boolean => {
```
priyajeet: Since this isn't called with callback, can be a simple function isEditorStateEmpty() {...} instead of arrow.

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelector.js
```diff
@@ -174,11 +204,8 @@ class DraftJSMentionSelector extends React.Component<Props, State> {
                     this.checkValidity();
```
priyajeet: is this one still needed?

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelector.js
ConradJChan: Not sure if componentDidMount will be enough since we need to inspect the `editorState` coming in from `ApprovalCommentForm` each time a change is made to the DraftJS component. Maybe in combination with componentDidUpdate but leaving it as is should save extra renders right?

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelector.js
ConradJChan: This is still needed for the case where there is `internalStateEditor` -- at least that is unless the consumer of the component uses it in the same way as `ApprovalCommentForm` which takes the `editorState` returned from `onChange` and sets it back as the prop on `DraftJSMentionSelector`

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelector.js
```diff
@@ -87,15 +87,50 @@ class DraftJSMentionSelector extends React.Component<Props, State> {
...
+            const isCurrentEditorStateEmpty = this.isEditorStateEmpty(editorState);
+            const isNextEditorStateEmpty = this.isEditorStateEmpty(nextEditorState);
+            const isNewEditorState = isNextEditorStateEmpty && !isCurrentEditorStateEmpty;
+            const isEditorStateDirty = isCurrentEditorStateEmpty && !isNextEditorStateEmpty;
+
+            // Detect case where controlled EditorState is created anew and empty.
+            // If next editorState is empty and the current editorState is not empty
+            // that means it is a new empty state and this component should not be marked dirty
+            if (isNewEditorState) {
+                nextState = { ...nextState, hasReceivedFirstInteraction: false, error: null };
```
jstoffan: I know this code was already here, but can we change `hasReceivedFirstInteraction` to `isTouched` or something?

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelector.js
jstoffan: It looks like we'd need a combination of `getDerivedStateFromProps` to handle what `state` should become and `componentDidMount` to determine whether or not to run `this.checkValidity()`, depending on what changed. I think doing so would also eliminate much of the code in `this.handleChange`, as well.

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelector.js
```diff
@@ -75,27 +75,83 @@ class DraftJSMentionSelector extends React.Component<Props, State> {
...
+
+        const { editorState: prevEditorStateFromProps } = prevProps;
+        const { editorState } = this.props;
+
+        let prevEditorState;
+        let currentEditorState;
+
+        // Determine whether we're working with the internal editor state or
+        // external editor state passed in from props
+        if (prevInternalEditorState) {
```
jstoffan: It looks like `this.state.internalEditorState` is `null` if the component is controlled. Would something like the below work?

```
const prevEditorState = prevInternalEditorState || prevEditorStateFromProps;
const currentEditorState = internalEditorState || editorState;
```

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelector.js
```diff
@@ -163,27 +217,14 @@ class DraftJSMentionSelector extends React.Component<Props, State> {
...
-                    }
-                    this.checkValidity();
-                },
-            );
-        } else {
-            if (onChange) {
-                onChange(nextEditorState);
-            }
-            this.checkValidity();
+            this.setState({ internalEditorState: nextEditorState }, () => onChange && onChange(nextEditorState));
```
jstoffan: Can we default `onChange` to `noop` while we're at it? Then it can just be called at the very bottom of the method outside of the conditions.

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelector.js
```diff
@@ -75,25 +77,87 @@ class DraftJSMentionSelector extends React.Component<Props, State> {
...
+        let nextState = null;
 
         if (contacts) {
-            this.setState({ contacts }, () => {
-                this.checkValidity();
-            });
+            nextState = { contacts };
+        }
+
+        return nextState;
```
jstoffan: Minor, but can be simplified a bit:
```
const { contacts } = this.props;
return contacts ? { contacts } : null;
```

test/integration/content-sidebar/ContentSidebar.e2e.test.js
```diff
@@ -96,4 +97,57 @@ describe('ContentSidebar', () => {
...
+        });
+
+        it('should reset validation errors if comment is cancelled', () => {
+            getDraftJSEditor()
+                .click()
+                .type('qwerty')
+                .clear();
+
+            getTooltip()
+                .should('exist')
```
jstoffan: This is checked implicitly by the `.contains` check below.

test/integration/content-sidebar/ContentSidebar.e2e.test.js
```diff
@@ -96,4 +97,57 @@ describe('ContentSidebar', () => {
             cy.get('@versionHistory').should('not.exist');
         });
     });
+
+    describe('activity feed comments', () => {
+        const getDraftJSEditor = () => cy.get('[contenteditable]');
+        const getTooltip = () => cy.get('[role="tooltip"]');
```
jstoffan: Is it worth scoping these down to the comment components?

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelector.js
```diff
@@ -159,31 +215,18 @@ class DraftJSMentionSelector extends React.Component<Props, State> {
...
-                onChange(nextEditorState);
-            }
-            this.checkValidity();
+            this.setState({ internalEditorState: nextEditorState });
         }
     };
 
     handleValidityStateUpdateHandler = () => {
-        if (!this.state.hasReceivedFirstInteraction) {
+        if (!this.state.isTouched) {
```
priyajeet: destruct

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelector.js
```diff
@@ -75,27 +77,83 @@ class DraftJSMentionSelector extends React.Component<Props, State> {
...
+            // If the current editorState is empty and the next editorState is not
+            // empty then this is the first interaction so mark this component dirty
+            newState = { isTouched: true };
+        }
+
+        return newState;
+    }
+
+    checkValidityIfAllowed() {
+        const { validateOnBlur }: Props = this.props;
```
priyajeet: can default it to noop to reduce a branch

ConradJChan: In this case, I want the branch because I only want to `checkValidity` if `validateOnBlur` if false. If it's true, then on through `handleBlur` will `checkValidity` be called

test/integration/content-sidebar/ContentSidebar.e2e.test.js
ConradJChan: I scoped down the `getDraftJSEditor` but tooltip I've left for now since Tether just appends it to the body and I couldn't figure a good way to be more specific

src/components/form-elements/draft-js-mention-selector/__tests__/DraftJSMentionSelector-test.js
```diff
@@ -28,22 +28,141 @@ describe('bcomponents/form-elements/draft-js-mention-selector/DraftJSMentionSele
...
+            expect(DraftJSMentionSelector.getDerivedStateFromProps({})).toEqual(null);
+        });
+    });
+
+    describe('componentDidUpdate()', () => {
+        let mockGetDerivedStateFromEditorState;
+        let mockCheckValidityIfAllowed;
+        let spySetState;
+
+        const setupInstance = props => {
```
mickr: Should these be wrapped in a `beforeEach` 🍹

test/integration/content-sidebar/ContentSidebar.e2e.test.js
```diff
@@ -96,4 +97,50 @@ describe('ContentSidebar', () => {
...
+        it('should show required error if type and then delete text', () => {
+            getDraftJSEditor()
+                .click()
+                .type('qwerty')
+                .clear();
+
+            getTooltip().contains(localize('boxui.validation.requiredError'));
+        });
+
+        it('should reset validation errors if comment is cancelled', () => {
```
jstoffan: Consider combining some of these test cases to save on setup overhead.

PR #1437, opened 2019-07-19
chore(tasks): delete old task ui from activity feed

alexkrolick:
#1438 comes after this step


boxcla:

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -135,113 +135,53 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         this.fetchFeedItems();
     };
 
-    tasksApiNew = {
```
alexkrolick: delete the "tasksApiNew" object and promote the functions to top-level scope in the component

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -398,7 +310,7 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
      */
     fetchFeedItems(shouldRefreshCache: boolean = false, shouldDestroy: boolean = false) {
         const { file, api, features } = this.props;
-        const shouldShowNewTasks = isFeatureEnabled(features, 'activityFeed.tasks.newApi');
+        const shouldShowNewTasks = true;
```
shikharmohan: Can we just get rid of this?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -253,32 +193,34 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+            this.feedSuccessCallback();
+
+            if (onSuccess) {
+                onSuccess();
+            }
+
+            onTaskUpdate();
+        };
+
+        api.getFeedAPI(false).updateTaskNew(file, task, successCallback, errorCallback);
```
shikharmohan: can we just get rid of *TaskNew and just call it

alexkrolick:
I'll do that rename as a separate step - it's very confusing to replace updateTask (old) with updateTask (new) at the same time.

src/elements/content-sidebar/ActivitySidebar.js
alexkrolick: It's a positional arg, so until feed api is updated, it needs to be there because it's not the last flag

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -253,32 +193,34 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
-     * @param {string} taskId - ID of task to be updated
-     * @param {string} taskAssignmentId - Task assignment ID
-     * @param {TaskAssignmentStatus} status - New task assignment status
-     * @return void
-     */
     updateTaskAssignment = (taskId: string, taskAssignmentId: string, status: TaskAssignmentStatus): void => {
         const { file, api } = this.props;
 
-        api.getFeedAPI(false).updateTaskAssignment(
+        api.getFeedAPI(false).updateTaskCollaborator(
```
jstoffan: Is this intentional? Is it addressing a bug?

jstoffan: I think I see what happened now. It looks like this is how `tasksNew` already worked, which was very similar to the prior flow, except for this one line.

alexkrolick: Yep

PR #1440, opened 2019-07-22
 feat(typography): Add new heading components

### LINKED ISSUE #1435: Update the typography to keep consistent with box design language (bdl)

Sunyang730:
- Add H1 ~ H6 that will have the bdl style

Screen Shot 2019-07-21 at 5 13 21 PM
Screen Shot 2019-07-21 at 5 13 32 PM

fixes #1435

### DIFF
--- a/scripts/styleguide.config.js
+++ b/scripts/styleguide.config.js
@@ -115,6 +115,10 @@ const allSections = [
                 description: 'Implements the "media object" layout',
                 usageMode: 'expand',
             },
+            {
+                name: 'Headings',
+                content: '../src/components/headings/README.md',
+            },
         ],
     },
     {
--- a/src/components/headings/H1.js
+++ b/src/components/headings/H1.js
@@ -1,0 +1,14 @@
+// @flow
+import * as React from 'react';
+import classNames from 'classnames';
+
+import './H1.scss';
+
+type Props = {
+    children: React.Node,
+    className?: string,
+};
+
+const H1 = ({ children, className }: Props) => <h1 className={classNames('bdl-H1', className)}>{children}</h1>;
+
+export default H1;
--- a/src/components/headings/H1.scss
+++ b/src/components/headings/H1.scss
@@ -1,0 +1,11 @@
+@import '../../styles/variables';
+
+h1.bdl-H1 {
+    @include font-family;
+
+    color: $bdl-gray;
+    font-size: 22px;
+    font-weight: bold;
+    letter-spacing: .3px;
+    line-height: 28px;
+}
--- a/src/components/headings/H5.scss
+++ b/src/components/headings/H5.scss
@@ -1,0 +1,12 @@
+@import '../../styles/variables';
+
+h5.bdl-H5 {
+    @include font-family;
+
+    color: $bdl-gray-62;
+    font-size: 11px;
+    font-weight: normal;
+    letter-spacing: .6px;
+    line-height: 16px;
+    text-transform: uppercase;
+}
--- a/src/components/headings/H6.scss
+++ b/src/components/headings/H6.scss
@@ -1,0 +1,11 @@
+@import '../../styles/variables';
+
+h6.bdl-H6 {
+    @include font-family;
+
+    color: $bdl-gray-62;
+    font-size: 13px;
+    font-weight: normal;
+    letter-spacing: .3px;
+    line-height: 20px;
+}
--- a/src/features/quick-search/QuickSearch.scss
+++ b/src/features/quick-search/QuickSearch.scss
@@ -1,3 +1,5 @@
+@import '../../styles/variables';
+
 .quick-search {
     display: inline-block;
     transition: width .25s linear;
@@ -8,7 +10,7 @@
     }
 
     .overlay {
-        border-radius: 2px;
+        border-radius: $bdl-border-radius-size;
     }
 
     ul.overlay {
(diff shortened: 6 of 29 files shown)


boxcla:

src/components/headings/H5.scss
```diff
@@ -0,0 +1,12 @@
...
+
+h5.bdl-H5 {
+    @include font-family;
+
+    color: $bdl-gray-62;
+    font-size: 11px;
+    font-weight: normal;
+    letter-spacing: .6px;
+    line-height: 16px;
+    text-transform: uppercase;
```
priyajeet: Get rid of this uppercase, headings should not dictate uppercasing, that will be in the localized text if needed.

src/features/quick-search/QuickSearch.scss
```diff
@@ -8,7 +10,7 @@
     }
 
     .overlay {
-        border-radius: 2px;
+        border-radius: $bdl-border-radius-size;
```
priyajeet: revert from this PR?

src/components/headings/H6.scss
```diff
@@ -0,0 +1,11 @@
+@import '../../styles/variables';
+
+h6.bdl-H6 {
+    @include font-family;
+
+    color: $bdl-gray-62;
+    font-size: 13px;
+    font-weight: normal;
+    letter-spacing: .3px;
+    line-height: 20px;
```
priyajeet: H6 supposed to have taller line height than H5?

src/components/headings/H1.scss
```diff
@@ -0,0 +1,11 @@
+@import '../../styles/variables';
+
+h1.bdl-H1 {
+    @include font-family;
+
+    color: $bdl-gray;
+    font-size: 22px;
+    font-weight: bold;
+    letter-spacing: .3px;
```
priyajeet: I am wondering if one should create a common mixin for
- the font family
- the bdl gray color
- the letter spacing

And override the one that has it different

priyajeet: Each should include overrides for https://github.com/box/box-ui-elements/blob/master/src/styles/_reset.scss#L92
As well should include https://github.com/box/box-ui-elements/blob/master/src/styles/_variables.scss#L182 because any of the Hs can be portaled/tethered outside the parent container

priyajeet: Also make sure that all the rules here like margin https://github.com/box/box-ui-elements/blob/master/src/styles/common/_typography.scss#L32 are overriden or reset in these new stylesheets. Because we want to make that typography.scss irrelevant. In other words if its included on the page it should have no effect on these new components.

scripts/styleguide.config.js
```diff
@@ -115,6 +115,10 @@ const allSections = [
                 description: 'Implements the "media object" layout',
                 usageMode: 'expand',
             },
+            {
+                name: 'Headings',
```
booc0mtaco: I think we should replace the existing Typography section instead of creating a new one for these header components, since they are essentially redundant. That would also be a good place for showing more complex combinations of the headers and body text.

Sunyang730: I was thinking of replacing but we still have the style though. So I created the new sections. I can update that instead and make the old one says deprecated or something

src/components/headings/H1.scss
Sunyang730: Re: better-black style: I have a separate PR for the better black since I consider that as a breaking change 🤷‍♂. 
Re: Overrides: I tested on the style sheet which include all the resets and it works. 
Re: commonmixin, was thinking of that too but the common styles are very mix.. so I didn't create the mixin. One thing I do think it might be helpful is creating the H1 ~ H6 mixin so others might want to mix that in with their style?

src/components/headings/H5.scss
Sunyang730: Yeah, debating on that too and remember about the locale thing but it's also part of the BDL, so I am keeping it here.

src/components/headings/H6.scss
Sunyang730: according to the BDL, yes

src/components/headings/H1.scss
priyajeet: Because of uppercasing, it looks like only H5 is the outliner. For example letter spacing is the same in all, which is the same as common-typography which you need to include here. Likewise missing margin will be common. bdl-gray color is also common to all other than H5 H6

So can be

```scss
@mixin bdl-H { 
    @include common-typography;
    color: bdl-gray;
    margin: 8px 0;
    text-rendering: optimizeLegibility;
}

.bdl-H1 {
    @include bdl-H;
    font-size: 22px;
    font-weight: bold;
    line-height: 28px;
}

.bdl-H5 {
    @include bdl-H;
    color: $bdl-gray-62;
    font-size: 11px;
    letter-spacing: .6px;
    line-height: 16px;
    text-transform: uppercase;
}
```

src/components/headings/H1.js
```diff
@@ -0,0 +1,14 @@
...
+import classNames from 'classnames';
+
+import './H1.scss';
+
+type Props = {
+    children: React.Node,
+    className?: string,
+};
+
+const H1 = ({ children, className }: Props) => <h1 className={classNames('bdl-H1', className)}>{children}</h1>;
```
alexkrolick: All of these component should spread props onto the underlying dom node

alexkrolick: You can also get rid of the type for children when you spread the props

src/components/headings/H1.scss
```diff
@@ -0,0 +1,11 @@
+@import '../../styles/variables';
+
+h1.bdl-H1 {
```
alexkrolick: Is the `h1` specificity needed? That makes a modifier class require equivalent specificity to override any of the default properties

PR #1442, opened 2019-07-22
feat(typography): update body typography

Sunyang730:
BREAKING CHANGE: The color of the body will be default to bdl-gray. The letter spacing for common typography will now be px instead of em. Please check scaling.

fixes #1442


boxcla:

src/features/quick-search/QuickSearch.scss
```diff
@@ -8,7 +10,7 @@
     }
 
     .overlay {
-        border-radius: 2px;
+        border-radius: $bdl-border-radius-size;
```
Sunyang730: not sure why this is here, will remove

src/styles/_variables.scss
```diff
@@ -195,7 +195,7 @@ $bdl-line-height: 20px;
 @mixin common-typography {
     @include fonts;
 
-    color: $better-black;
+    color: $bdl-gray;
```
booc0mtaco: Here we specify the `letter-spacing` in `em`s but the others are in pixels. I think we should adopt pixels for all of them since `em`s can behave differently based on nested values. Also, I think we should probably drop usage of `$bdl-line-height` here so the values match what's in design specifically.

Sunyang730: I was thinking about that too... but decided to leave it as em since it might break scale. Do you know if there's any plan for bdl to support scaling?

booc0mtaco: not certain, but I'd guess it's not going to in the near future. My guess is that the design listed this specific value when it was originally added, without intending to imply scaling requirement.

alexkrolick: `rem` is based on the root element instead of cascading like `em`, we could use that?

priyajeet: rem generally will require that we set a % at the body style, that might be a bigger breaking change.

@Sunyang730 can you change font-weight: 400; on line 178 to font-weight: normal; so that we are consistent.

Sunyang730: Will make that change in the other PR since that one will add something to the typography file too.

booc0mtaco: without 62.5%, the *em value will be based on 16px font-size. that trick gets it so that 1em = 10px

overall, looks like we have around 8 usages of `em` and exactly one use of `rem`. My thought is that we can move away from both since i doubt we're needing the behaviors of either right now (separate piece of work to consider)

PR #1445, opened 2019-07-22
chore(util): create suit component helper

alexkrolick:
- [x] ~~TODO: this could also implement ref forwarding by default https://reactjs.org/docs/forwarding-refs.html~~


boxcla:

PR #1420, opened 2019-07-17
refactor(comment): port to Media layout

alexkrolick:

PR #1446, opened 2019-07-23
fix(menu): Fix padding for menu item buttons

jstoffan:
It looks like the change from #1420 inadvertently caused an issue in our compiled stylesheets. The `.menu-*` styles now appear before our `.btn-*` styles, rather than vice-versa. Unfortunately, this causes menu-item/button hybrid elements, such as those in VersionHistoryAction, to lose their padding.

version-padding


boxcla:

PR #1447, opened 2019-07-23
feat: add prop for resumable uploads feature flip

rosepierce:
Adding prop in ContentUploader to be passed from the parent application to allow use of a feature flip for resumable uploads.


boxcla:

src/elements/content-uploader/ContentUploader.js
```diff
@@ -131,6 +132,7 @@ class ContentUploader extends Component<Props, State> {
         onMinimize: noop,
         onCancel: noop,
         isFolderUploadEnabled: false,
+        isResumableUploadsEnabled: true,
```
jstoffan: Is this meant to be enabled by default? If so, is it needed?

rosepierce: It's set to be enabled by default right now for testing purposes, but it doesn't necessarily have to be enabled by default.

PR #1448, opened 2019-07-23
fix(radio-group): pass rest of props to radio input

mkazimahmed:
The rest of the props are not being passed from radio group to the radio button input. Fixed that in this PR.

### DIFF
--- a/src/components/radio/RadioGroup.js
+++ b/src/components/radio/RadioGroup.js
@@ -47,20 +47,15 @@ class RadioGroup extends React.Component<Props, State> {
 
     render() {
         const { children, className, name } = this.props;
-        const { value } = this.state;
+        const { value: stateValue } = this.state;
 
         return (
             <div className={`radio-group ${className}`} onChange={this.onChangeHandler}>
-                {React.Children.map(children, radio => (
-                    <RadioButton
-                        description={radio.props.description}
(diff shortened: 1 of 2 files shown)


boxcla:
Verified that @mkazimahmed has signed the CLA. Thanks for the pull request!

src/components/radio/RadioGroup.js
```diff
@@ -47,20 +47,24 @@ class RadioGroup extends React.Component<Props, State> {
...
-                        name={name}
-                        value={radio.props.value}
-                    />
-                ))}
+                {React.Children.map(children, radio => {
+                    const { description, isDisabled, label, value, ...rest } = radio.props;
+
+                    return (
+                        <RadioButton
+                            description={description}
```
priyajeet: If you are using ...rest, you dont need to explicitly add anything but name and isSelected

src/components/radio/RadioGroup.js
```diff
@@ -47,20 +47,24 @@ class RadioGroup extends React.Component<Props, State> {
 
     render() {
         const { children, className, name } = this.props;
-        const { value } = this.state;
```
priyajeet: Can also do `const { value: stateValue } = this.state;`

PR #1450, opened 2019-07-23
fix(pill-selector): expose onInput to trap input

priyajeet:
Also fixes a few flow types and defaulting

### DIFF
--- a/src/components/pill-selector-dropdown/PillSelectorDropdownField.js
+++ b/src/components/pill-selector-dropdown/PillSelectorDropdownField.js
@@ -14,19 +14,21 @@ type Props = FieldProps & {
     /** CSS class for the component. */
     className?: string,
     /** Given selected values and input text, returns a list of filtered options. Defaults to defaultDropdownFilter. */
-    dropdownFilter?: (options: Array<Option>, selectedValues: Array<Option>, inputText: string) => Array<Option>,
+    dropdownFilter: (options: Array<Option>, selectedValues: Array<Option>, inputText: string) => Array<Option>,
     /** Given options, renders the dropdown list. Defaults to defaultDropdownRenderer. */
     dropdownRenderer: (options: Array<Option>) => React.Node,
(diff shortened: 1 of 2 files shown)


boxcla:

src/components/pill-selector-dropdown/PillSelectorDropdownField.js
```diff
@@ -14,19 +14,21 @@ type Props = FieldProps & {
     /** CSS class for the component. */
     className?: string,
     /** Given selected values and input text, returns a list of filtered options. Defaults to defaultDropdownFilter. */
-    dropdownFilter?: (options: Array<Option>, selectedValues: Array<Option>, inputText: string) => Array<Option>,
+    dropdownFilter: (options: Array<Option>, selectedValues: Array<Option>, inputText: string) => Array<Option>,
```
jpfranco: I think removing the `?` might make flow complain about a missing prop if the consumer doesn't provide a filter.

src/components/pill-selector-dropdown/PillSelectorDropdownField.js
```diff
@@ -43,10 +45,12 @@ type State = {
 
 class PillSelectorDropdownField extends React.PureComponent<Props, State> {
     static defaultProps = {
+        dropdownFilter: defaultDropdownFilter,
```
jpfranco: Ah yes, I thought I had updated this. Probably just set `defaultDropdownRenderer` and forgot about this one.

src/components/pill-selector-dropdown/PillSelectorDropdownField.js
priyajeet: Its part of defaultProps, so flow should be happy

jpfranco: Oh ok, cool. Didn't know that flow used the `defaultProps`

PR #1452, opened 2019-07-24
fix(radio-group): fix radio group by passing the value prop

mkazimahmed:
Had to add the `value` prop explicitly as it was extracted from the props and was not part of the `...rest`

### DIFF
--- a/src/components/radio/RadioGroup.js
+++ b/src/components/radio/RadioGroup.js
@@ -54,7 +54,7 @@ class RadioGroup extends React.Component<Props, State> {
                 {React.Children.map(children, radio => {
                     const { value, ...rest } = radio.props;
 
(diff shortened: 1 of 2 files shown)


boxcla:
Verified that @mkazimahmed has signed the CLA. Thanks for the pull request!

i18n/en-US.properties
```diff
@@ -202,6 +202,8 @@ be.deleteDialogFileText = Are you sure you want to delete {name}?
 be.deleteDialogFolderText = Are you sure you want to delete {name} and all its contents?
 # Label for delete confirmation dialog
 be.deleteDialogLabel = Confirm Delete
+# Aria label for button to delete a comment, task, or app activity
+be.deleteLabel = Delete
```
priyajeet: Can you rebase? This file should not be there

PR #1459, opened 2019-07-25
build: add clean command to prepush

booc0mtaco:
- Pros: this prevents all flows that might use a dirty i18n directory from producing inaccurate changes to .properties files
- Cons: users who push for review while running development servers will have the running instance get corrupted

- Workaround suggestion (1): prepush should work on a temp directory to avoid any discrepancies, and make different git workflows more predictable

Discuss? Other alternatives?

### DIFF
--- a/scripts/prepush.sh
+++ b/scripts/prepush.sh
@@ -37,6 +37,7 @@ check_and_commit_updated_translations() {
 
 # lint, test, and build assets to update translations
 prepush() {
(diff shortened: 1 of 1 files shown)


boxcla:

jstoffan:
I'm going to close this for now in the hopes that #1574 will help address these cases. Let's revisit if they start to crop up again.

PR #1460, opened 2019-07-25
refactor(comment): extract shared code between tasks and comments

alexkrolick:


boxcla:

src/elements/content-sidebar/activity-feed/comment/Comment.js
```diff
@@ -142,9 +142,9 @@ class Comment extends React.Component<Props, State> {
         const isMenuVisible = (canDelete || canEdit) && !isPending;
 
         return (
-            <div className="bcs-comment-container">
+            <div className="bcs-Comment">
                 <Media
-                    className={classNames('bcs-comment', {
+                    className={classNames('bcs-Comment-content', {
```
jstoffan: Minor, but how about `bcs-Comment-media`?

src/elements/content-sidebar/activity-feed/common/activity-error/ActivityError.scss
```diff
@@ -0,0 +1,13 @@
+@import '../../../../../styles/variables';
+
+.bcs-ActivityError {
+    padding: 14px 15px;
+}
+
+.bcs-ActivityError-action,
+button.bcs-ActivityError-action {
```
jstoffan: You can use `&` to avoid repeating the class name both here and in UserLink. Also consider using `.btn-plain` rather than an element selector.

src/elements/content-sidebar/activity-feed/common/activity-message/ActivityMessage.scss
```diff
@@ -0,0 +1,28 @@
+@import '../../../../../styles/variables';
+
+.bcs-ActivityMessage {
+    white-space: pre-wrap;
+
+    .link {
+        color: $bdl-box-blue;
+        white-space: normal;
+
+        &:hover {
```
jstoffan: Is `&:focus` also needed here?

alexkrolick: This is what was there, just moved to the new file. Possibly?

src/elements/content-sidebar/activity-feed/common/activity-error/ActivityError.scss
alexkrolick: The `button` part is just to match specificity; ideally the classname itself would be enough. Since we have a lot of issues with `.btn-plain` I decided to just avoid it for now since we know there has to be an underlying button element.

src/elements/content-sidebar/activity-feed/comment/Comment.js
alexkrolick: Maybe; I'll think about in the next PR extracting the tasks code.

PR #1463, opened 2019-07-25
fix(media): increase specificity of Media-button css to avoid .btn-plain

alexkrolick:


boxcla:

src/components/media/Media.scss
```diff
@@ -27,7 +27,8 @@
 }
 
 // the "..." menu button, meant to be used inside Media-body
-.bdl-Media-menu {
+.bdl-Media-menu,
+button.bdl-Media-menu {
```
alexkrolick: - & doesn't seem to parse unless you are in the `{}` block
- could put `.btn-plain` inside but we might have a different class soon

src/components/media/Media.scss
```diff
@@ -27,7 +27,8 @@
 }
 
 // the "..." menu button, meant to be used inside Media-body
-.bdl-Media-menu {
+.bdl-Media-menu,
```
jstoffan: Does `.bdl-Menu-menu.btn-plain` work on this line? It avoids doubling-up the output CSS and the extra qualifier can be removed once `ButtonPlain` is no more.

alexkrolick: I'd prefer to keep the base class without the qualifier so it doesn't break if we forget to remove this reference.

alexkrolick: Could do this?

```scss
// the "..." menu button, meant to be used inside Media-body
.bdl-Media-menu {
    // float is used to reflow content instead of pushing it to the left
    float: right;

    // overrides for .btn-plain's margin resets on focus states
    &,
    &.btn-plain:active,
    &.btn-plain:hover,
    &.btn-plain:focus {
        margin-bottom: 5px;
        margin-left: 10px;
    }
}
```

alexkrolick: That seems best actually, it only overrides the `.btn-plain` stuff and doesn't propagate the issue further.

PR #1464, opened 2019-07-25
feat(content-explorer): grid view styling

rickstefanik:


boxcla:

src/components/grid-view/GridView.js
```diff
@@ -38,33 +39,43 @@ class GridView extends React.Component<Props> {
...
             contents.push(
                 <div
-                    key={getProp(currentCollection, `items[${slotIndex}].id`) || uniqueId('bdl-GridView-slot')}
-                    className="bdl-GridView-slot"
+                    key={getProp(item, `id`) || uniqueId('bdl-GridView-slot')}
+                    className={classNames('bdl-GridView-slot', {
+                        'bdl-GridView-slot--selected': getProp(item, 'selected'),
+                    })}
+                    onClick={() => onItemSelect(item)}
+                    role="presentation"
```
rickstefanik: A few times I ran into the es-lint rule: `Visible, non-interactive elements with click handlers must have at least one keyboard listener.` . Putting `role="presentation` "fixed" the problem.  Is this ok, or am I dodging the problem?

src/components/grid-view/GridView.scss
```diff
@@ -5,29 +5,65 @@
...
+
+        flex-basis: $columns7prct;
+        max-width: $columns7prct;
+    }
+
     .ReactVirtualized__Table__rowColumn:first-of-type {
         margin: 0;
     }
+
+    & .bce-btn-more-options {
```
rickstefanik: Buttons seemed very large without reduced padding

src/elements/common/item/iconCellRenderer.js
```diff
@@ -12,15 +12,19 @@ import { TYPE_FOLDER, TYPE_FILE, TYPE_WEBLINK } from '../../../constants';
...
+export function getIcon(dimension: number, rowData: BoxItem, changeWeblinkHeight: boolean = false) {
     const { type, extension, has_collaborations, is_externally_owned }: BoxItem = rowData;
     switch (type) {
         case TYPE_FOLDER:
             return <FolderIcon dimension={dimension} isCollab={has_collaborations} isExternal={is_externally_owned} />;
         case TYPE_FILE:
             return <FileIcon dimension={dimension} extension={extension} />;
         case TYPE_WEBLINK:
-            return <BookmarkIcon width={dimension} />;
+            return changeWeblinkHeight ? (
```
rickstefanik: Is there a better way to do this?  Not sure why this was only changing `width` and not `height`, but I felt this approach was safer and wouldn't change behavior in other areas where we use `getIcon`.

src/elements/common/progress-bar/ProgressBar.js
```diff
@@ -112,6 +113,7 @@ class ProgressBar extends PureComponent<Props, State> {
     render() {
         const { percent }: State = this.state;
         const containerStyle = {
+            zIndex: overlayZIndex,
```
rickstefanik: Progress bar could get hidden behind grid cards when scrolling.  Is this a fix or a hack?

src/elements/content-explorer/ItemGridThumbnail.scss
```diff
@@ -11,10 +15,16 @@
...
     display: flex;
     justify-content: center;
     left: 0;
     position: absolute;
     right: 0;
     top: 0;
+
+    & svg {
+        height: 60%;
+        width: 60%;
```
rickstefanik: This lets generic BoxIcons scale as the cards change size.  The height and width should be bounded by the grid card itself, so is using `%'s` here ok?

src/components/grid-view/GridView.js
```diff
@@ -38,33 +39,43 @@ class GridView extends React.Component<Props> {
...
+            const item = getProp(currentCollection, `items[${slotIndex}]`);
+
             // using item's id as key is important for renrendering.  React Virtualized Table rerenders
             // on every 1px scroll, so using improper key would lead to image flickering in each
             // card of the grid view when scrolling.
             contents.push(
                 <div
-                    key={getProp(currentCollection, `items[${slotIndex}].id`) || uniqueId('bdl-GridView-slot')}
-                    className="bdl-GridView-slot"
+                    key={getProp(item, `id`) || uniqueId('bdl-GridView-slot')}
```
ConradJChan: many of these calls to `getProp` seem to be for a property directly on the item itself, could we destructure them out `const { id, selected } = item` instead? IMO, it clutters up the code less

src/components/grid-view/GridView.js
ConradJChan: Yeah that seems like dodging the problem. Seems like `role="presentation"` is to remove semantic meaning from the element (https://www.accessibility-developer-guide.com/examples/sensible-aria-usage/presentation/). It makes more sense to allow keyboard users to hit "Enter" and "Space" to trigger the same action as click

src/elements/common/item/iconCellRenderer.js
ConradJChan: Looks like this is only used by ContentExplorer iconCellRenderer so it seems like it should be ok to use height and width

src/elements/content-explorer/ItemGridCell.js
```diff
@@ -26,26 +24,20 @@ const ItemGridCell = ({
 }: Props) => {
     return (
         <figure className="bce-ItemGridCell">
-            <ItemGridThumbnail item={item} />
-            <figcaption>
+            <ItemGridThumbnail item={item} onItemClick={onItemClick} />
```
ConradJChan: Seems like this would only allow you to click on the thumbnail? Would we instead want to click the whole grid item to select?

src/components/grid-view/GridView.scss
```diff
@@ -5,29 +5,65 @@
...
 
     &.bdl-GridView--columns-1 .bdl-GridView-slot {
-        $columns1prct: 100%;
+        $columns1prct: calc(100% - 10px);
 
         flex-basis: $columns1prct;
         max-width: $columns1prct;
     }
 
+    &.bdl-GridView--columns-2 .bdl-GridView-slot {
```
jstoffan: Can we use the `columns` [style rules](https://developer.mozilla.org/en-US/docs/Web/CSS/columns) to eliminate all this math? It appears to be supported in IE11.

src/components/grid-view/GridView.js
```diff
@@ -38,36 +39,47 @@ class GridView extends React.Component<Props> {
...
+
             // using item's id as key is important for renrendering.  React Virtualized Table rerenders
             // on every 1px scroll, so using improper key would lead to image flickering in each
             // card of the grid view when scrolling.
             contents.push(
-                <div
-                    key={getProp(currentCollection, `items[${slotIndex}].id`) || uniqueId('bdl-GridView-slot')}
-                    className="bdl-GridView-slot"
+                <button
+                    key={id || uniqueId('bdl-GridView-slot')}
```
jstoffan: Is there a situation where `id` is undefined in item?

src/components/grid-view/GridView.js
```diff
@@ -38,36 +39,47 @@ class GridView extends React.Component<Props> {
...
+            const { id, selected } = item;
+
             // using item's id as key is important for renrendering.  React Virtualized Table rerenders
             // on every 1px scroll, so using improper key would lead to image flickering in each
             // card of the grid view when scrolling.
             contents.push(
-                <div
-                    key={getProp(currentCollection, `items[${slotIndex}].id`) || uniqueId('bdl-GridView-slot')}
-                    className="bdl-GridView-slot"
+                <button
```
jstoffan: If I remember correctly, there can be a dropdown menu toggle button inside of each cell. This will lead to buttons inside of buttons, which isn't valid markup. It seems like the cell itself should remain a div and the clickable elements inside it should be buttons.

src/components/grid-view/GridView.scss
```diff
@@ -37,17 +73,41 @@
...
+
     & + & {
         margin-left: 10px;
     }
 }
 
 .bdl-GridView-tableRow {
     border: none;
+
+    &:first-child .bdl-GridView-slot {
```
jstoffan: This can probably be simplified with sibling selectors:

```
.bdl-GridView-tableRow {
    & + & {
        margin-top: 10px;
    }
}
```

src/elements/common/progress-bar/ProgressBar.js
jstoffan: Can the z-index be set on the `.be-progress-container` class, instead?

src/elements/common/sub-header/ViewModeChangeButton.js
```diff
@@ -21,6 +22,7 @@ type Props = {
 
 const ViewModeChangeButton = ({ className = '', onViewModeChange, intl, viewMode, ...rest }: Props) => {
     const isGridView = viewMode === VIEW_MODE_GRID;
+    const iconColor = bdlNeutral03;
```
jstoffan: Please avoid rename-only reassignments; we can pass the color value directly. Also update to use new color palette (bdlGray*)

src/components/grid-view/GridView.js
```diff
@@ -38,36 +39,47 @@ class GridView extends React.Component<Props> {
...
     componentDidUpdate(prevProps: Props) {
-        const { columnCount, width } = this.props;
+        const { columnCount, currentCollection, width } = this.props;
 
         // The React Virtualized Table must be notified when either the cached
         // row sizes or the parent width change. If omitted, rows are sized
         // incorrectly resulting in gaps or content overlap.
-        if (columnCount !== prevProps.columnCount || width !== prevProps.width) {
+        if (
+            columnCount !== prevProps.columnCount ||
```
jstoffan: Previous values should be destructured in the function signature as `prevColumnCount`, etc.

jstoffan: Previous values should be destructured in the function signature as `prevColumnCount`, etc.

src/elements/common/sub-header/ViewModeChangeButton.js
rickstefanik: Made the change.  Want to ask though, is it ever a good idea to have rename-only reassignments? My thinking here was if there were more icons, and if someone wanted to change the color of all icons in the future, having `iconColor` would reduce the number of lines that would need to be changed.

rickstefanik:
Still working on unresolved comments.

src/elements/common/sub-header/ViewModeChangeButton.js
jstoffan: Renaming can almost always be accomplished during destructuring or at import. Changing the icon color in one place versus two seems like a minor optimization.

```
import { bdlGray01 as iconColor } from 'constants';
```

src/components/grid-view/GridView.scss
rickstefanik: Was able to remove this classes and put the width in the javascript.  Still looking into using padding instead of margins.  (Have it working, just a little difficult to get everything perfect.)

src/components/grid-view/GridView.scss
rickstefanik: Still looking into this.  React Virtualized seems to place the rows improperly if the margins aren't on the GridViewSlots.

src/elements/content-explorer/ItemGridThumbnail.scss
```diff
@@ -17,4 +21,9 @@
     position: absolute;
     right: 0;
     top: 0;
+
+    & svg {
```
jstoffan: Is the `&` needed here? It should mainly be used to create compound selectors, but this looks like a descendant selector.

src/components/grid-view/GridView.scss
```diff
@@ -4,30 +4,17 @@
 .bdl-GridView {
     padding: 0 5px;
 
-    &.bdl-GridView--columns-1 .bdl-GridView-slot {
-        $columns1prct: 100%;
-
-        flex-basis: $columns1prct;
-        max-width: $columns1prct;
+    .ReactVirtualized__Table__rowColumn:first-of-type {
```
jstoffan: It's a little confusing, but :first-of-type refers to the element rather than the class. Can you use :first-child, instead?

src/components/grid-view/GridView.scss
```diff
@@ -2,32 +2,19 @@
 @import '../../styles/variables';
 
 .bdl-GridView {
-    padding: 0 5px;
```
rickstefanik: This was not working as expected.  This div appears entirely to the left of the actual grid view content, so in reality, it was adding 10px of padding to the left of the content, and 0 px of padding to the right of the content.

src/components/grid-view/GridView.scss
```diff
@@ -2,32 +2,19 @@
...
     }
+}
 
-    .ReactVirtualized__Table__rowColumn:first-of-type {
-        margin: 0;
-    }
+.bdl-GridView-body {
+    outline: none;
+    padding-left: 5px;
+    padding-right: 5px;
```
rickstefanik: Putting the padding here (as opposed to putting it on `.bdl-GridView`) does work as expected.

src/components/grid-view/GridView.js
```diff
@@ -98,6 +113,7 @@ class GridView extends React.Component<Props> {
                 width={width}
                 gridClassName="bdl-GridView-body"
                 rowClassName="bdl-GridView-tableRow"
+                rowStyle={{ paddingRight: 0, width: '100%' }}
```
rickstefanik: Setting these styles seems to be necessary since we are altering `.bdl-GridView-body`.  (These style attributes can't be put on the class, because ReactVirtualized puts its own styles on the row elements, which would override what was set on `bdl-GridView-tableRow`.)

rickstefanik: This line is unnecessary if we don't alter `.bdl-GridView-body`, and just put all the margins/padding on the GridViewSlots themselves.

src/components/grid-view/GridView.scss
```diff
@@ -2,49 +2,45 @@
...
+        padding: 4px 10px;
     }
+}
 
-    .ReactVirtualized__Table__rowColumn:first-of-type {
-        margin: 0;
-    }
+.bdl-GridView-body {
+    outline: none;
+    padding: 5px;
```
rickstefanik: Putting the padding here (as opposed to putting it on .bdl-GridView) does work as expected.

src/components/grid-view/GridView.scss
rickstefanik: Have it working using the padding approach.

rickstefanik: Screen Shot 2019-07-29 at 10 34 09 AM

src/elements/common/sub-header/SubHeaderRight.js
```diff
@@ -39,9 +39,9 @@ const SubHeaderRight = ({
     const { sortBy, sortDirection, items = [] }: Collection = currentCollection;
     const hasItems: boolean = items.length > 0;
     const isFolder: boolean = view === VIEW_FOLDER;
-    const showSort: boolean = isFolder && hasItems;
+    const showSort: boolean = hasItems;
     const showAdd: boolean = (!!canUpload || !!canCreateNewFolder) && isFolder;
-    const showGridButton: boolean = isFolder && hasItems;
+    const showGridButton: boolean = hasItems;
```
rickstefanik: Fixing this!

src/components/grid-view/GridViewSlot.js
```diff
@@ -0,0 +1,32 @@
...
+};
+
+const GridViewSlot = ({ item, slotIndex, slotRenderer, slotWidth }: Props) => {
+    const { id, selected } = item;
+
+    // using item's id as key is important for renrendering.  React Virtualized Table rerenders
+    // on every 1px scroll, so using improper key would lead to image flickering in each
+    // card of the grid view when scrolling.
+    return (
+        <div key={id} className="bdl-GridViewSlot" style={{ maxWidth: slotWidth, flexBasis: slotWidth }}>
```
ConradJChan: Should the key be moved back to where the list of `GridViewSlot` is created? Since there is no list within this component, I'm not sure of the benefits of adding `key` here

ConradJChan: It might make sense to have `selected` be a prop as well and if we do that we don't even need to pass in the item if id is used for the key on the GridViewSlot itself

rickstefanik:
Screen Shot 2019-07-30 at 9 27 36 AM

src/elements/common/breadcrumbs/InlineBreadcrumbs.js
```diff
@@ -12,20 +12,29 @@ import { DELIMITER_SLASH } from '../../../constants';
...
+                <FormattedMessage {...messages.in} />
+                &nbsp;
+                <Breadcrumbs
+                    crumbs={breadcrumbs}
+                    delimiter={DELIMITER_SLASH}
+                    onCrumbClick={onItemClick}
+                    rootId={rootId}
+                />
+            </span>
+            {isGridView && <div className="bdl-InlineBreadcrumbs-spacer" />}
```
ConradJChan: Is this something we can defer to `GridView` to override in it's css instead of mixing GridView concerns into the common breadcrumb component?

src/elements/common/breadcrumbs/InlineBreadcrumbs.js
```diff
@@ -12,20 +12,29 @@ import { DELIMITER_SLASH } from '../../../constants';
...
     return (
-        <span className="be-inline-breadcrumbs">
-            <FormattedMessage {...messages.in} />
-            &nbsp;
-            <Breadcrumbs crumbs={breadcrumbs} delimiter={DELIMITER_SLASH} onCrumbClick={onItemClick} rootId={rootId} />
-        </span>
+        <React.Fragment>
+            <span className="be-inline-breadcrumbs">
+                <FormattedMessage {...messages.in} />
+                &nbsp;
```
mickr: Is this something that needs to be done through an html entity and not css spacing? 🍹

src/components/grid-view/GridViewSlot.scss
```diff
@@ -0,0 +1,25 @@
...
+    box-shadow: 0 1px 5px rgba(50, 50, 50, .1);
+    height: 100%;
+
+    &:hover,
+    &:active,
+    &:focus {
+        border-color: $bdl-box-blue-50;
+    }
+
+    &.bdl-GridViewSlot-content--selected {
```
rickstefanik: Is it better to keep `&.bdl-GridViewSlot-content--selected` separate from `&:hover, &:active, &:focus` above since `bdl-GridViewSlot-content--selected` is a class, or should I include them all in the same rule?

src/components/grid-view/GridView.js
```diff
@@ -58,16 +57,20 @@ class GridView extends React.Component<Props> {
         const maxSlotIndex = Math.min(startingIndex + columnCount, count);
 
         for (let slotIndex = startingIndex; slotIndex < maxSlotIndex; slotIndex += 1) {
+            const item = getProp(currentCollection, `items[${slotIndex}]`);
```
jstoffan: Minor, but this can be inlined to:

```
const { id, selected } = getProp(currentCollection, `items[${slotIndex}]`);
```

src/components/grid-view/GridView.scss
```diff
@@ -1,53 +1,31 @@
...
+    & .bce-more-options {
+        align-self: flex-start;
     }
 
-    &.bdl-GridView--columns-5 .bdl-GridView-slot {
-        $columns5prct: calc(20% - 10px); /* 100%/5 - 10px */
-
-        flex-basis: $columns5prct;
-        max-width: $columns5prct;
+    & .bce-btn-more-options {
```
jstoffan: I don't think the `&` is needed here.

src/components/grid-view/GridView.js
jstoffan: Can we move it to CSS if we use higher specificity or `!important`? If not, this seems okay, but you could make it a constant if you really want.

src/components/grid-view/GridView.scss
```diff
@@ -1,53 +1,31 @@
...
+        margin-top: 0;
+        padding-bottom: 5px;
+        padding-left: 10px;
+        padding-right: 10px;
+        padding-top: 4px;
```
jstoffan: This looks like a magic number. Can we keep the padding consistent and use flexbox to vertically center the content?

src/elements/common/item/ItemSubDetails.js
```diff
@@ -33,15 +35,17 @@ const ItemSubDetails = ({ view, item }: Props) => {
...
-            <FormattedMessage
-                {...message}
-                values={{
-                    date: DateValue,
-                    name: modifiedBy,
-                }}
-            />
-            <span>
-                &nbsp;-&nbsp;
+            <span className={isGridView ? 'bdl-ItemSubDetails-span-modifiedBy' : ''}>
```
jstoffan: Class names should avoid mention of visual treatments or element names (span). Also, can we apply the class always and then override it based on the presence (or absence) class from a higher level, like `bdl-GridView`?

src/elements/common/item/ItemSubDetails.js
```diff
@@ -33,15 +35,17 @@ const ItemSubDetails = ({ view, item }: Props) => {
...
+                <FormattedMessage
+                    {...message}
+                    values={{
+                        date: DateValue,
+                        name: modifiedBy,
+                    }}
+                />
+            </span>
+            <span className={isGridView ? 'bdl-ItemSubDetails-span-size' : ''}>
+                {!isGridView && <React.Fragment>&nbsp;-&nbsp;</React.Fragment>}
```
jstoffan: Please use CSS for spacing rather than non-breaking spaces.

src/elements/common/item/ItemSubDetails.scss
```diff
@@ -0,0 +1,7 @@
+.bdl-ItemSubDetails-span-modifiedBy,
```
jstoffan: Does it hurt for these to be added in list view?

rickstefanik:
Screen Shot 2019-07-30 at 2 28 04 PM

rickstefanik:
Screen Shot 2019-07-30 at 2 45 03 PM

PR #1468, opened 2019-07-27
feat(comments): implement comment edit

alexkrolick:

PR #1469, opened 2019-07-29
fix(comments): format existing mentions in editor

alexkrolick:
Followup to #1468 - fixes issue with the draftjs comment editor where text already containing mentions was turned back into plaintext, which destroys the references on save.

### DIFF
--- a/flow/DraftJSFlowStub.js.flow
+++ b/flow/DraftJSFlowStub.js.flow
@@ -4,4 +4,5 @@ declare export var ContentState: any
 declare export var Editor: any
 declare export var CompositeDecorator: any
 declare export var Modifier: any
+declare export var SelectionState: any
 declare export default {}
--- a/src/components/form-elements/draft-js-mention-selector/createMentionSelectorState.js
+++ b/src/components/form-elements/draft-js-mention-selector/createMentionSelectorState.js
@@ -5,9 +5,11 @@ import DraftMentionDecorator from './DraftMentionDecorator';
 // returns data for first mention in a string
 const getMentionFromText = (text: string) => {
     // RegEx.exec() is stateful, so we create a new regex instance each time
-    const MENTION_REGEX = /([@＠﹫])\[([0-9]+):([^\]]+)]/gi;
-    const matchArray = MENTION_REGEX.exec(text);
-    if (!matchArray) return null;
+    const mentionRegex = /([@＠﹫])\[([0-9]+):([^\]]+)]/gi;
+    const matchArray = mentionRegex.exec(text);
+    if (!matchArray) {
+        return null;
+    }
     const [fullMatch, triggerSymbol, id, name] = matchArray;
     const start = matchArray.index;
     const end = start + fullMatch.length;
@@ -29,8 +31,7 @@ const createMentionSelectorState = (message: string = '') => {
             const { data, start, end } = mention;
             contentState.createEntity('MENTION', 'IMMUTABLE', data);
             const mentionEntityKey = contentState.getLastCreatedEntityKey();
-            let mentionRange = SelectionState.createEmpty(contentBlock.getKey());
-            mentionRange = mentionRange.merge({
+            const mentionRange = SelectionState.createEmpty(contentBlock.getKey()).merge({
                 anchorOffset: start,
                 focusOffset: end,
             });
@@ -38,8 +39,7 @@ const createMentionSelectorState = (message: string = '') => {
             contentBlock = contentState.getBlockForKey(contentBlock.getKey());
         }
     }
-    const editorState = EditorState.createWithContent(contentState, DraftMentionDecorator);
-    return editorState;
+    return EditorState.createWithContent(contentState, DraftMentionDecorator);
 };
 
 export default createMentionSelectorState;
--- a/src/components/form-elements/draft-js-mention-selector/createMentionSelectorState.js
+++ b/src/components/form-elements/draft-js-mention-selector/createMentionSelectorState.js
@@ -1,0 +1,45 @@
+// @flow
+import { ContentState, EditorState, Modifier, SelectionState } from 'draft-js';
+import DraftMentionDecorator from './DraftMentionDecorator';
+
+// returns data for first mention in a string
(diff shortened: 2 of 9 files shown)


boxcla:

flow/DraftJSFlowStub.js.flow
```diff
@@ -4,4 +4,5 @@ declare export var ContentState: any
 declare export var Editor: any
 declare export var CompositeDecorator: any
 declare export var Modifier: any
+declare export var SelectionState: any
```
alexkrolick: One thing to note is that this stub is extremely limited to the Flow types that ship with Draft by default. We should probably switch to the real types and fix the corresponding errors.

src/components/form-elements/draft-js-mention-selector/createMentionSelectorState.js
```diff
@@ -0,0 +1,45 @@
+// @flow
+import { ContentState, EditorState, Modifier, SelectionState } from 'draft-js';
+import DraftMentionDecorator from './DraftMentionDecorator';
+
+// returns data for first mention in a string
+const getMentionFromText = (text: string) => {
+    // RegEx.exec() is stateful, so we create a new regex instance each time
+    const MENTION_REGEX = /([@＠﹫])\[([0-9]+):([^\]]+)]/gi;
+    const matchArray = MENTION_REGEX.exec(text);
+    if (!matchArray) return null;
```
jstoffan: I think we usually prefer full block formatting, even for guard clauses:

```
if (!matchArray) {
    return null;
}
```

src/components/form-elements/draft-js-mention-selector/createMentionSelectorState.js
```diff
@@ -0,0 +1,45 @@
+// @flow
+import { ContentState, EditorState, Modifier, SelectionState } from 'draft-js';
+import DraftMentionDecorator from './DraftMentionDecorator';
+
+// returns data for first mention in a string
+const getMentionFromText = (text: string) => {
+    // RegEx.exec() is stateful, so we create a new regex instance each time
+    const MENTION_REGEX = /([@＠﹫])\[([0-9]+):([^\]]+)]/gi;
```
jstoffan: It seems like we should use camelCase rather all-caps here, since we are intentionally reintantiating this regex.

src/components/form-elements/draft-js-mention-selector/createMentionSelectorState.js
```diff
@@ -0,0 +1,45 @@
...
+            let mentionRange = SelectionState.createEmpty(contentBlock.getKey());
+            mentionRange = mentionRange.merge({
+                anchorOffset: start,
+                focusOffset: end,
+            });
+            contentState = Modifier.replaceText(contentState, mentionRange, data.content, null, mentionEntityKey);
+            contentBlock = contentState.getBlockForKey(contentBlock.getKey());
+        }
+    }
+    const editorState = EditorState.createWithContent(contentState, DraftMentionDecorator);
```
jstoffan: Minor, but it looks like this can be done inline, since `editorState` is only used once.

src/components/form-elements/draft-js-mention-selector/createMentionSelectorState.js
```diff
@@ -0,0 +1,45 @@
...
+        const text = contentBlock.getText();
+        const mention = text ? getMentionFromText(text) : null;
+        if (mention == null) {
+            contentBlock = contentState.getBlockAfter(contentBlock.getKey());
+        } else {
+            const { data, start, end } = mention;
+            contentState.createEntity('MENTION', 'IMMUTABLE', data);
+            const mentionEntityKey = contentState.getLastCreatedEntityKey();
+            let mentionRange = SelectionState.createEmpty(contentBlock.getKey());
+            mentionRange = mentionRange.merge({
```
jstoffan: Could this be chained to avoid the variable reassignment? 
```
const mentionRange = SelectionState.createEmpty(contentBlock.getKey()).merge({ anchorOffset: start, focusOffset: end });

Comment (alexkrolick) [reply src/components/form-elements/draft-js-mention-selector/createMentionSelectorState.js]: For some reason that didn't work when I tried it, but I didn't have good tests at the time. I can try again.

File: src/components/form-elements/draft-js-mention-selector/createMentionSelectorState.js [hunk 6]
```diff
@@ -0,0 +1,45 @@
...
+
+// creates draftjs state with mentions parsed into entities
+const createMentionSelectorState = (message: string = '') => {
+    let contentState = ContentState.createFromText(message);
+    let contentBlock = contentState.getFirstBlock();
+
+    while (contentBlock != null) {
+        const text = contentBlock.getText();
+        const mention = text ? getMentionFromText(text) : null;
+        if (mention == null) {
```
booc0mtaco: why truthy/falsy? what kinds of things can `mention` and `contentBlock` be?

src/components/form-elements/draft-js-mention-selector/createMentionSelectorState.js
```diff
@@ -0,0 +1,45 @@
+// @flow
+import { ContentState, EditorState, Modifier, SelectionState } from 'draft-js';
+import DraftMentionDecorator from './DraftMentionDecorator';
+
+// returns data for first mention in a string
+const getMentionFromText = (text: string) => {
+    // RegEx.exec() is stateful, so we create a new regex instance each time
+    const mentionRegex = /([@＠﹫])\[([0-9]+):([^\]]+)]/gi;
```
booc0mtaco: nit: could we pull this out of the function so that we don't repeatedly compile the regex. Luckily, i don't think this will run all that often, so not necessary if possible

alexkrolick: No, see comment

src/components/form-elements/draft-js-mention-selector/createMentionSelectorState.js
alexkrolick: `== null` is only undefined/null, not falsy

PR #1470, opened 2019-07-29
feat(unified-share-modal): invite section ux changes

jasminevle:
Changes invite section to expand depending on whether there are initiallySelectedContacts


boxcla:
Verified that @jasminevle has signed the CLA. Thanks for the pull request!

src/features/unified-share-modal/UnifiedShareModal.js
```diff
@@ -161,15 +161,16 @@ class UnifiedShareModal extends React.Component<Props, State> {
 
     constructor(props: Props) {
         super(props);
+        const inviteCollabsContacts = props.initiallySelectedContacts || [];
```
priyajeet: Would it be better to add `initiallySelectedContacts` to defaultProps above and default it to []

src/features/unified-share-modal/UnifiedShareModal.js
```diff
@@ -164,12 +165,12 @@ class UnifiedShareModal extends React.Component<Props, State> {
 
         this.state = {
             emailSharedLinkContacts: [],
-            inviteCollabsContacts: props.initiallySelectedContacts || [],
+            inviteCollabsContacts: props.initiallySelectedContacts,
```
priyajeet: Also, shouldn't there be a unit test, to test that initial state is as expected?
If there is none, should add one. You can disable lifecycle methods so that the constructor is the only one executed besides render.

PR #1472, opened 2019-07-29
feat(sidebar): programmatic refresh

ConradJChan:
Provides mechanism to trigger a refresh of the ContentSidebar active panel

### DIFF
--- a/src/elements/content-preview/ContentPreview.js
+++ b/src/elements/content-preview/ContentPreview.js
@@ -1195,7 +1195,7 @@ class ContentPreview extends React.PureComponent<Props, State> {
                                 getViewer={this.getViewer}
                                 history={history}
                                 language={language}
-                                componentRef={this.setSidebarRef}
+                                ref={this.setSidebarRef}
                                 sharedLink={sharedLink}
                                 sharedLinkPassword={sharedLinkPassword}
                                 requestInterceptor={requestInterceptor}
--- a/src/elements/content-preview/ContentPreview.js
+++ b/src/elements/content-preview/ContentPreview.js
@@ -62,6 +62,7 @@ type Props = {
     collection: Array<string | BoxItem>,
     contentOpenWithProps: ContentOpenWithProps,
     contentSidebarProps: ContentSidebarProps,
+    contentSidebarRef: React.ref,
     enableThumbnailsSidebar: boolean,
     features?: FeatureConfig,
     fileId?: string,
@@ -161,8 +162,6 @@ class ContentPreview extends React.PureComponent<Props, State> {
     mouseMoveTimeoutID: TimeoutID;
 
     rootElement: HTMLElement;
-
-    sidebarRef: ?React.Node;
 
     stagedFile: ?BoxItem;
 
@@ -1083,12 +1082,6 @@ class ContentPreview extends React.PureComponent<Props, State> {
         this.previewContainer = container;
     };
 
-    getSidebarRef = (): ?React.Node => this.sidebarRef;
-
-    setSidebarRef = (sidebar: ?React.Node) => {
-        this.sidebarRef = sidebar;
-    };
-
     /**
      * Renders the file preview
      *
@@ -1104,6 +1097,7 @@ class ContentPreview extends React.PureComponent<Props, State> {
             messages,
             className,
             contentSidebarProps,
+            contentSidebarRef,
             contentOpenWithProps,
             hasHeader,
             history,
@@ -1195,7 +1189,7 @@ class ContentPreview extends React.PureComponent<Props, State> {
                                 getViewer={this.getViewer}
                                 history={history}
                                 language={language}
-                                ref={this.setSidebarRef}
+                                ref={contentSidebarRef}
                                 sharedLink={sharedLink}
                                 sharedLinkPassword={sharedLinkPassword}
                                 requestInterceptor={requestInterceptor}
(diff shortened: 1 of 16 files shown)


boxcla:

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -91,6 +92,13 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         this.fetchCurrentUser(currentUser);
     }
 
+    componentDidUpdate({ refresh: prevRefresh }: Props) {
+        const { refresh } = this.props;
+        if (refresh !== prevRefresh) {
```
alexkrolick: This will get recalled if you toggle back to false - how can this work more than once? It sounds like you want a ref, `key`, or move state up a level. Or perhaps a random value?

- https://reactjs.org/docs/refs-and-the-dom.html#adding-a-ref-to-a-class-component
- https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#recommendation-fully-uncontrolled-component-with-a-key

ConradJChan: As we discussed offline, the value itself is not important, just a trigger to detect when refresh should occur. Will change the `setState` in `ContentSidebar` to use the updater function instead so async setState's batched together won't fail to trigger an update

src/elements/common/async-load/AsyncLoad.js
```diff
@@ -42,13 +42,14 @@ const AsyncLoad = (asyncProps: Props = {}) => {
...
             const { error } = this.state;
+            const { componentRef, ...rest } = this.props;
             if (error) {
                 return <ErrorComponent error={error} />;
             }
 
             return (
                 <React.Suspense fallback={fallback}>
-                    <LazyComponent {...this.props} />
+                    <LazyComponent ref={componentRef} {...rest} />
```
jstoffan: Should we use `React.forwardRef`, instead?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -91,6 +92,13 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         this.fetchCurrentUser(currentUser);
     }
 
+    componentDidUpdate({ refreshIdentity: prevRefreshIdentity }: Props) {
+        const { refreshIdentity } = this.props;
+        if (refreshIdentity !== prevRefreshIdentity) {
```
jstoffan: Can we expose a similar `refresh` method from here that the parent can call?

ConradJChan: Sounds reasonable, let me work on that

src/elements/content-sidebar/__tests__/ContentSidebar-test.js
```diff
@@ -243,4 +243,25 @@ describe('elements/content-sidebar/ContentSidebar', () => {
...
+    describe('refresh()', () => {
+        let wrapper;
+        let instance;
+
+        beforeEach(() => {
+            wrapper = getWrapper();
+            instance = wrapper.instance();
+        });
+
+        test.each`
```
mickr: That's cool, I haven't used this before 👏

src/elements/content-sidebar/DetailsSidebar.js
```diff
@@ -48,6 +48,7 @@ type ExternalProps = {
     onClassificationClick?: (e: SyntheticEvent<HTMLButtonElement>) => void,
     onRetentionPolicyExtendClick?: Function,
     onVersionHistoryClick?: Function,
+    refreshIdentity?: boolean,
```
mickr: Might be cool if this were a function or boolean, so the parent component could pass in a callback for on refresh, maybe in a follow up if useful.

src/elements/content-sidebar/SidebarPanels.js
```diff
@@ -114,6 +116,7 @@ const SidebarPanels = ({
                             onVersionHistoryClick={onVersionHistoryClick}
                             startMarkName={MARK_NAME_JS_LOADING_ACTIVITY}
                             {...activitySidebarProps}
+                            refreshIdentity={refreshIdentity}
```
mickr: Should these go after the `...activitySidebarProps` for precedence (overwritten by a parent)

mickr:
Just a few small comments but looks good

src/elements/content-preview/ContentPreview.js
```diff
@@ -1081,6 +1083,12 @@ class ContentPreview extends PureComponent<Props, State> {
         this.previewContainer = container;
     };
 
+    getSidebarRef = (): ?React.Node => this.sidebarRef;
```
jstoffan: Can we make it so that there's a `contentSidebarRef` prop that we forward to `ContentSidebar`? That way we don't have to worry about setting it up ourselves.

src/elements/content-sidebar/DetailsSidebar.js
ConradJChan: yeah will revisit this to see if I can get a programmatic refresh on each Panel itself instead of relying on this toggling refresh prop

src/elements/content-sidebar/SidebarPanels.js
ConradJChan: I can see what you're saying but I think `refreshIdentity` is a reflection of `ContentSidebar` internal state rather than a prop passed in from the top level user of the element so I'd rather not let them override it

PR #1474, opened 2019-07-29
fix(comment): Add height to comment area to support multiple mentions

jstoffan:
This change is meant to allow more than one mention to be shown at a time in the dropdown selector by padding out the bottom of the comment form. I also took the opportunity to overhaul the styles and migrate to SUIT naming conventions.

**Before**
current-type

**After**
new-type


boxcla:

src/components/media/Media.js
```diff
@@ -10,7 +10,7 @@ type Props = {
     /** Component to use as outermost element, e.g., 'li' */
     as: React.ElementType,
     /** Child elements */
-    children: React.ChildrenArray<React.Element<typeof MediaFigure | typeof MediaBody | typeof MediaMenu>>,
+    children: React.ChildrenArray<React.Element<typeof MediaFigure | typeof MediaBody | typeof MediaMenu> | boolean>,
```
alexkrolick: Technically this should be `false | null` for the `&&` and `? :` use cases

src/components/media/Media.scss
```diff
@@ -11,14 +11,13 @@
     align-self: flex-start;
     flex: 0 0 auto;
     justify-self: flex-start;
-    margin: 0;
+    margin: 0 10px 0 0;
```
ConradJChan: What is the convention for this, would it make sense to move this layout styling to `.bdl-Media` instead of baking it into the `.bdl-Media-figure`?

alexkrolick: No, this is correct.
- If the figure is present, it should push over the content (body).
- If not present, the body is full-width.
- The specificity for overriding the margin should be (0,0,1)

ConradJChan: Ah that makes sense, didn't realize figure could be absent

PR #1476, opened 2019-07-30
feat(content-explorer): local storage for view mode preference

rickstefanik:


boxcla:

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -290,6 +296,13 @@ class ContentExplorer extends Component<Props, State> {
         } else {
             this.fetchFolder(currentFolderId);
         }
+
+        if (isFeatureEnabled(features, 'contentExplorer.gridView.enabled')) {
+            const pref = this.store.getItem(localStoreViewMode);
+            if (pref && (pref === VIEW_MODE_GRID || pref === VIEW_MODE_LIST)) {
```
rickstefanik: Initial `pref` check is to pass `flow`.  Also, not sure if it's possible to dynamically check if `pref` is of type `ViewMode`, so had to do the two equality checks.

jstoffan: Is the initial `pref` identity check needed? Seems like the two equality checks should be enough, since null/undefined won't result in `true`.

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -1256,6 +1269,7 @@ class ContentExplorer extends Component<Props, State> {
      * @return {void}
      */
     changeViewMode = (viewMode: ViewMode): void => {
+        this.store.setItem(localStoreViewMode, viewMode);
         this.setState({ viewMode });
```
jstoffan: Ideally, we only want to track state in one place to avoid them getting out of sync. Can we eliminate `this.state.viewMode` entirely and just call `this.store.getItem` each time?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -290,6 +296,13 @@ class ContentExplorer extends Component<Props, State> {
         } else {
             this.fetchFolder(currentFolderId);
         }
+
+        if (isFeatureEnabled(features, 'contentExplorer.gridView.enabled')) {
+            const pref = this.store.getItem(localStoreViewMode);
+            if (pref && (pref === VIEW_MODE_GRID || pref === VIEW_MODE_LIST)) {
+                this.changeViewMode(pref);
```
jstoffan: This will call `setState` after the component is mounted, which will result in a double render. We may be able to eliminate this by calling `getItem` during render.

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -1325,6 +1328,11 @@ class ContentExplorer extends Component<Props, State> {
         const allowUpload: boolean = canUpload && !!can_upload;
         const allowCreate: boolean = canCreateNewFolder && !!can_upload;
 
+        let viewMode = this.store.getItem(localStoreViewMode);
+        if (!viewMode) {
+            viewMode = VIEW_MODE_LIST;
+        }
```
rickstefanik: cleaning up this logic

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -292,6 +297,15 @@ class ContentExplorer extends Component<Props, State> {
         }
     }
 
+    componentDidUpdate() {
+        const { features }: Props = this.props;
+        const { viewMode }: State = this.state;
+
+        if (isFeatureEnabled(features, 'contentExplorer.gridView.enabled')) {
```
jstoffan: Probably want to check if the new `viewMode` is different from `prevState.viewMode`.

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -1249,6 +1263,14 @@ class ContentExplorer extends Component<Props, State> {
         this.setState({ currentOffset: newOffset }, this.refreshCollection);
     };
 
+    getViewMode = (): ViewMode => {
+        const { features }: Props = this.props;
+        const viewModePreference = this.store.getItem(localStoreViewMode);
+        return viewModePreference && isFeatureEnabled(features, 'contentExplorer.gridView.enabled')
```
jstoffan: This may be a bit easier to read if we break the `isFeatureEnabled` result into its own variable.

PR #1479, opened 2019-07-31
fix(explorer): fetch shared link on demand

ConradJChan:
todo:
- [x] unit tests
- [x] e2e tests

### DIFF
--- a/src/elements/content-explorer/ContentExplorer.js
+++ b/src/elements/content-explorer/ContentExplorer.js
@@ -405,7 +405,7 @@ class ContentExplorer extends Component<Props, State> {
     async fetchFolderSuccessCallback(collection: Collection, triggerNavigationEvent: boolean): Promise<void> {
         const { features, onNavigate, rootFolderId }: Props = this.props;
         const { boxItem, id, items = [], name }: Collection = collection;
-        const { selected }: State = this.state;
+        const { selected = {} }: State = this.state;
         const rootName = id === rootFolderId ? name : '';
 
         if (isFeatureEnabled(features, 'contentExplorer.gridView.enabled')) {
@@ -415,9 +415,9 @@ class ContentExplorer extends Component<Props, State> {
                 const thumbnailUrl = itemThumbnails[index];
                 return thumbnailUrl ? { ...item, thumbnailUrl } : item;
             });
-            this.updateCollection({ ...collection, items: itemsWithThumbnails }, selected);
+            this.updateCollection({ ...collection, items: itemsWithThumbnails }, selected.id);
         } else {
-            this.updateCollection(collection, selected);
+            this.updateCollection(collection, selected.id);
         }
 
         // Close any open modals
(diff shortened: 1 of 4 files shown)


boxcla:

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -1105,6 +1106,50 @@ class ContentExplorer extends Component<Props, State> {
...
+     *
+     * @param {Object} item file or folder
+     * @returns {void}
+     */
+    handleSharedLinkSuccess = (newItem: BoxItem) => {
+        const { currentCollection } = this.state;
+        const { items = [] } = currentCollection;
+
+        // Update item in collection
+        const newItems = items.map(item => (item.id === newItem.id ? newItem : item));
```
jstoffan: Should we create a merge helper to avoid duplicating this logic?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -809,7 +809,8 @@ class ContentExplorer extends Component<Props, State> {
 
         const selectedItem: BoxItem = { ...item, selected: true };
 
-        this.updateCollection(currentCollection, selectedItem, () => {
+        const newItems = items.map(currentItem => (currentItem.id === item.id ? item : currentItem));
```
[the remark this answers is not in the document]
ConradJChan: Opted to create a helper function for updating an item in the collection instead of having `updateCollection` do too much since it's scope currently is to update the select state of all the items in the collection and call `setState`

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -751,21 +751,20 @@ class ContentExplorer extends Component<Props, State> {
      *
      * @private
      * @param {Collection} collection - collection that needs to be updated
-     * @param {?BoxItem} selected - item that should be selected in that collection (if present)
+     * @param {string} [selectedId] - ID of the item that should be selected in that collection (if present)
      * @param {Function} [callback] - callback function that should be called after setState occurs
      * @return {void}
      */
-    updateCollection(collection: Collection, selected: ?BoxItem, callback: Function = noop): Object {
+    updateCollection(collection: Collection, selectedId: ?string, callback: Function = noop): Object {
```
ConradJChan: Changed this method to just take in the id of the selected item, to remove the confusion about which is the source of truth

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -1105,6 +1109,46 @@ class ContentExplorer extends Component<Props, State> {
...
+     * it does not already exist, otherwise update the item in the state currentCollection.
+     *
+     * @param {Object} item file or folder
+     * @returns {void}
+     */
+    handleSharedLinkSuccess = (newItem: BoxItem) => {
+        const { currentCollection } = this.state;
+
+        // Update item in collection
+        this.updateCollection(currentCollection, newItem, () => this.setState({ isShareModalOpen: true }));
```
jstoffan: Any way we can avoid calling `setState` twice?

ConradJChan: Not sure... the way to avoid it would be to have `updateCollection` not set state and defer to the caller but then that introduces more duplication throughout the code.

PR #1481, opened 2019-07-31
feat: add metadata_queries api layer

manishPh:


boxcla:

src/api/MetadataQuery.js
```diff
@@ -0,0 +1,137 @@
...
+    errorCallback: ElementsErrorCallback;
+
+    /**
+     * Creates a key for the metadata cache
+     *
+     * @param {string} id - metadata template
+     * @return {string} key
+     */
+    getCacheKey(id: string): string {
+        return `${CACHE_PREFIX_METADATA}query_${id}`;
```
ConradJChan: Should we just keep them separate and create a new constant for the metadata query cache prefix?

src/api/MetadataQuery.js
```diff
@@ -0,0 +1,137 @@
...
+    /**
+     * API for querying enterprise metadata
+     * @param {Object} query - metadata query object
+     * @param {Function} successCallback - Function to call with results
+     * @param {Function} errorCallback - Function to call with errors
+     * @param {boolean|void} [options.forceFetch] - Bypasses the cache
+     * @return {void}
+     */
+    queryMetadata(
+        query: Object,
```
jstoffan: Can these parameters be more strongly-typed?

src/api/MetadataQuery.js
```diff
@@ -0,0 +1,137 @@
...
+    };
+
+    /**
+     * Does the network request to metadata query API
+     * @param {Object} query query object with SQL Clauses like properties
+     * @return {Promise}
+     */
+    queryMetadataRequest(query: Object): Promise<void> {
+        if (this.isDestroyed()) {
+            return Promise.reject();
```
jstoffan: I think we either `return` or call `successCallback` in most other API modules. What error would be shown to the user in this case?

src/api/MetadataQuery.js
```diff
@@ -0,0 +1,137 @@
...
+     *
+     * @return {boolean} if query results are loaded
+     */
+    isLoaded(): boolean {
+        const cache: APICache = this.getCache();
+        return cache.has(this.key);
+    }
+
+    /**
+     * Returns the results
```
jstoffan: It looks like the method doesn't currently return anything.

src/api/MetadataQuery.js
```diff
@@ -0,0 +1,137 @@
...
+        cache.set(this.key, data);
+        this.finish();
+    };
+
+    /**
+     * Does the network request to metadata query API
+     * @param {Object} query query object with SQL Clauses like properties
+     * @return {Promise}
+     */
+    queryMetadataRequest(query: Object): Promise<void> {
```
jstoffan: Can both this and the success handler be grouped more closely with its caller, `queryMetadata`?

src/api/MetadataQuery.js
```diff
@@ -0,0 +1,137 @@
...
+        successCallback: Function,
+        errorCallback: ElementsErrorCallback,
+        options: Object = {},
+    ): void {
+        if (this.isDestroyed()) {
+            return;
+        }
+
+        const { context = {} } = options;
+        this.key = this.getCacheKey(context.id);
```
ConradJChan: What is the `context` and `id` represent here? Does it make sense to do something similar to `Search.js` where the key is the encoded query?

src/api/MetadataQuery.js
manishPh: Both `Folder.js` and `Recents.js` we `return Promise.reject();`. Other files, we simply `return`

src/api/MetadataQuery.js
manishPh: The idea is `context` object will have some useful info about the query thats hard to extract from query object. e.g.enterprise_id, template_key, ancestor_folder_id etc. This is different from the `Search.js` where the query is a simple `string`. 
Component using this API should do the extraction and pass down this info for context. `id` could be generated as a unique cache property for each query.

flow-typed/box-ui-elements.js
```diff
@@ -892,3 +892,35 @@ type AdditionalVersionInfo = {
     currentVersionId?: ?string,
     updateVersionToCurrent: () => void,
 };
+
+type MetadataQeuryResponseEntryEnterprise = {
```
jstoffan: We're trying to eliminate global types. Can these types live in something like `src/common/types/metadataQueries.js`?

manishPh: Will make that change.

PR #1482, opened 2019-08-01
feat(content-explorer): grid view slider

rickstefanik:


boxcla:

src/components/grid-view/GridViewSlider.js
```diff
@@ -0,0 +1,62 @@
+// @flow
```
rickstefanik: Noticed that there isn't a `Slider` component in ui-elements.  Would it be useful to make a generic `Slider` component and then have `GridViewSlider` consume `Slider`?

src/components/grid-view/GridViewSlider.scss
```diff
@@ -0,0 +1,103 @@
+@import '../../elements/common/variables';
```
rickstefanik: Most of these styles were taken from the slider in EUA.

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -240,6 +245,8 @@ class ContentExplorer extends Component<Props, State> {
             currentPageSize: initialPageSize,
             errorCode: '',
             focusedRow: 0,
+            gridColumnCount: DEFAULT_GRID_VIEW_COLUMNS,
+            maxGridColumnCountForWidth: GRID_VIEW_MAX_COLUMNS,
```
rickstefanik: There are two different maximums at play here.  `GRID_VIEW_MAX_COLUMNS` is the maximum number of columns we want to show in the GridView, regardless of the width.  `maxGridColumnCountForWidth` is the maximum number of columns that can be displayed based on the given width.  `maxGridColumnCountForWidth` gets updated based on the width of the `AutoSizer` containing `GridView`.

src/components/grid-view/GridViewSlider.scss
```diff
@@ -0,0 +1,103 @@
...
+        background: $bdl-gray-20;
+    }
+}
+
+.bdl-GridViewSlider-PlainButton {
+    align-items: center;
+    display: flex;
+    justify-content: center;
+
+    &.bdl-GridViewSlider-PlainButton--minus {
```
jstoffan: This can live at the top level, as the extra specificity shouldn't be necessary if it appears below the class it's overriding.

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -122,13 +125,15 @@ type State = {
...
     focusedRow: number,
+    gridColumnCount: number,
     isCreateFolderModalOpen: boolean,
     isDeleteModalOpen: boolean,
     isLoading: boolean,
     isPreviewModalOpen: boolean,
     isRenameModalOpen: boolean,
     isShareModalOpen: boolean,
     isUploadModalOpen: boolean,
+    maxGridColumnCountForWidth: number,
```
jstoffan: Do these need to be public props?

rickstefanik:
Still working on the unresolved comments.

src/components/grid-view/GridViewSlider.js
ConradJChan: That's always a good thing to do, you might need to generalize it when you do so. You can always do a follow up PR to split it out

src/components/grid-view/GridViewSlider.scss
```diff
@@ -0,0 +1,96 @@
...
+    &:focus::-ms-fill-lower {
+        background: $bdl-gray-50;
+    }
+
+    &:focus::-ms-fill-upper {
+        background: $bdl-gray-20;
+    }
+}
+
+.bdl-GridViewSlider-PlainButton {
```
ConradJChan: Is this still used somewhere?

src/elements/common/messages.js
```diff
@@ -132,6 +132,16 @@ const messages = defineMessages({
         description: 'Label for switching to grid view',
         defaultMessage: 'Switch to Grid View',
     },
+    gridViewIncreaseColumnSize: {
+        id: 'be.gridViewIncreaseColumnSize',
+        description: 'Label for increasing the size of columns in grid view',
+        defaultMessage: 'Increase size of columns',
```
ConradJChan: Is it size of columns or number of columns?

src/elements/content-explorer/ContentExplorer.js
ConradJChan: minor but could we make the naming more consistent?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -1422,6 +1461,9 @@ class ContentExplorer extends Component<Props, State> {
                             onItemShare={this.share}
                             onItemPreview={this.preview}
                             onSortChange={this.sort}
+                            gridColumnCount={Math.min(gridColumnCount, maxGridColumnCountForWidth)}
```
ConradJChan: Alphabetize the props?

src/elements/content-explorer/ItemGrid.js
```diff
@@ -21,13 +27,33 @@ const ItemGrid = ({ currentCollection, rootId, ...rest }: Props) => {
...
+    };
+
+    /**
+     * Calls onGridViewResize with the max number of columns that can be displayed given
+     * the current width of the GridView.
+     *
+     * @param {height: number, width: number} dimensions - current dimensions of GridView, as given by AutoSizer
+     * @return {void}
+     */
+    const onResize = ({ width }): void => {
```
ConradJChan: I wonder if this logic could be somehow reconciled with the `makeResponse` on the `ContentExplorer` component. There we determine whether the size is small, med, or large and pass that as a prop to `ContentExplorer`. Could we somehow utilize that to determine the maxColumns instead having this duplication?

src/components/grid-view/GridViewSlider.js
```diff
@@ -0,0 +1,65 @@
...
+    return (
+        GRID_VIEW_MIN_COLUMNS < maxColumnCount && (
+            <div className="bdl-GridViewSlider">
+                <PlainButton
+                    className="bdl-GridViewSlider-button"
+                    onClick={() => {
+                        onChange(Math.max(RANGE_MIN, sliderValue - RANGE_STEP));
+                    }}
+                    type="button"
+                    aria-label={messages.gridViewIncreaseNumberOfColumns.defaultMessage}
```
rickstefanik: The minus button increases the number of columns (same behavior as webapp), so this isn't a typo.

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -240,6 +245,7 @@ class ContentExplorer extends Component<Props, State> {
             currentPageSize: initialPageSize,
             errorCode: '',
             focusedRow: 0,
+            gridColumnCount: GRID_VIEW_DEFAULT_COLUMNS,
```
jstoffan: The value can probably just be inlined here.

src/elements/common/messages.js
```diff
@@ -132,6 +132,16 @@ const messages = defineMessages({
         description: 'Label for switching to grid view',
         defaultMessage: 'Switch to Grid View',
     },
+    gridViewIncreaseNumberOfColumns: {
+        id: 'be.gridViewIncreaseNumberOfColumns',
```
jstoffan: Should we start namespacing these ids to group them together, like `be.gridView.increaseColumnCount`?

src/components/grid-view/GridViewSlider.js
jstoffan: It seems like the message should reflect the icon, though Should the text refer to something like `Decrease column size`, instead?

src/elements/content-explorer/ItemGrid.js
```diff
@@ -21,13 +22,14 @@ const ItemGrid = ({ currentCollection, rootId, ...rest }: Props) => {
     const slotRenderer = (slotIndex: number): ?React.Element<any> => {
         const item: ?BoxItem = getProp(currentCollection, `items[${slotIndex}]`);
 
-        return item ? <ItemGridCell item={item} rootId={rootId} {...rest} /> : null;
+        return item ? <ItemGridCell item={item} onItemSelect={onItemSelect} rootId={rootId} {...rest} /> : null;
```
jstoffan: Does `onItemSelect` need to be defined explicitly or would it be passed in as part of `...rest`?

rickstefanik:
Squashed and added tests.

rickstefanik:
Squashed and added tests.

test/integration/content-explorer/ContentExplorer.int.test.js
```diff
@@ -1,5 +1,6 @@
 import localize from '../../support/i18n';
 import utils from '../../support/utils';
+import { GRID_VIEW_MAX_COLUMNS, GRID_VIEW_MIN_COLUMNS } from '../../../src/constants';
```
jstoffan: We should try to avoid importing source code into our integration tests.

test/integration/content-explorer/ContentExplorer.int.test.js
```diff
@@ -85,6 +86,19 @@ const helpers = {
...
     },
+    getSliderPlusButton() {
+        return cy.getByAriaLabel(localize('be.gridView.increaseColumnSize'));
+    },
+    getSliderMinusButton() {
+        return cy.getByAriaLabel(localize('be.gridView.decreaseColumnSize'));
+    },
+    getGridViewColumnCount() {
+        return cy
+            .get('.bdl-GridView-row')
```
jstoffan: We should try to avoid using CSS classes in E2E tests.

test/integration/content-explorer/ContentExplorer.int.test.js
```diff
@@ -276,5 +292,41 @@ describe('ContentExplorer', () => {
...
+        });
+
+        it('Should decrease number of columns', () => {
+            helpers.getGridViewColumnCount().then(numColumns => {
+                helpers.getSliderPlusButton().click();
+                helpers.getGridViewColumnCount().should('equal', numColumns - 1);
+            });
+        });
+
+        it('Should cycle through all slider values, always staying within column restraints', () => {
```
jstoffan: It seems like this may be more appropriate for a unit test.

test/integration/content-explorer/ContentExplorer.int.test.js
```diff
@@ -276,5 +292,41 @@ describe('ContentExplorer', () => {
...
+
+        it('Should increase number of columns', () => {
+            helpers.getGridViewColumnCount().then(numColumns => {
+                helpers.getSliderMinusButton().click();
+                helpers.getGridViewColumnCount().should('equal', numColumns + 1);
+            });
+        });
+
+        it('Should decrease number of columns', () => {
+            helpers.getGridViewColumnCount().then(numColumns => {
```
jstoffan: We should favor larger, more complex cases for our E2E tests, due to the overhead in setting up and tearing down each case. Should these be unit tests, instead?

src/components/grid-view/GridViewSlider.js
```diff
@@ -0,0 +1,65 @@
...
+                    type="range"
+                    value={sliderValue}
+                />
+                <PlainButton
+                    className="bdl-GridViewSlider-button"
+                    onClick={() => {
+                        onChange(Math.min(RANGE_MAX, sliderValue + RANGE_STEP));
+                    }}
+                    type="button"
+                    aria-label={messages.gridViewIncreaseColumnSize.defaultMessage}
```
jstoffan: We shouldn't be using `defaultMessage` directly, as it won't be translated. Use `injectIntl` from `react-intl`, if needed.

test/integration/content-explorer/ContentExplorer.int.test.js
```diff
@@ -85,6 +85,12 @@ const helpers = {
     getClosePreviewButton() {
         return cy.getByAriaLabel(localize('be.close'));
     },
+    getSliderPlusButton() {
```
jstoffan: Unused?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -1280,13 +1284,37 @@ class ContentExplorer extends Component<Props, State> {
...
         return isGridViewEnabled && viewModePreference ? viewModePreference : VIEW_MODE_LIST;
     };
 
+    /**
+     * Get the maximum number of grid view columns based on the current width of the
+     * content explorer.
+     *
+     * @return {number}
+     */
+    getMaxNumberOfGridViewColumnsForWidth = (): number => {
```
jstoffan: Can we group this with the `getGridColumnCount` helper? Maybe just `getGridColumnMax`?

src/constants.js
```diff
@@ -330,6 +331,10 @@ export const SIDEBAR_VIEW_METADATA: 'metadata' = 'metadata';
 export const SIDEBAR_VIEW_ACTIVITY: 'activity' = 'activity';
 export const SIDEBAR_VIEW_VERSIONS: 'versions' = 'versions';
 
+/* ------------------ Grid View ---------------------- */
+export const GRID_VIEW_MAX_COLUMNS: 7 = 7;
```
jstoffan: It seems odd that this needs to be shared. It seems like it should either be passed in from the parent component (controlled) or it should live in the slider component (uncontrolled).

src/elements/content-explorer/ContentExplorer.js
rickstefanik: The result of `getMaxNumberOfGridViewColumnsForWidth` is used in `GridViewSlider`, while the result of `getContentGridColumnCount` is used in `Content`, so I don't think they can be grouped into one.  I'll refactor the code though to eliminate `getContentGridColumnCount`.

src/elements/common/sub-header/ViewModeChangeButton.scss
```diff
@@ -1,4 +1,5 @@
 .bdl-ViewModeChangeButton {
+    margin-left: 7px;
```
ConradJChan: Should this layout style be moved to `SubHeaderRight.scss`?

src/elements/content-explorer/Content.js
```diff
@@ -54,6 +55,7 @@ type Props = {
 const Content = ({
     currentCollection,
     focusedRow,
+    gridColumnCount = 0,
```
ConradJChan: What would the `ItemGrid` display if `gridColumnCount` actually was 0? Should it default to 1 instead?

PR #1484, opened 2019-08-01
fix(preview): Clean up preivew header styles and remove extra borders

jstoffan:


boxcla:

src/elements/content-preview/preview-header/FileInfo.scss
```diff
@@ -0,0 +1,9 @@
+.bcpr-FileInfo {
+    align-items: center;
+    display: flex;
+}
+
+.bcpr-FileInfo-name {
+    font-weight: 600;
```
booc0mtaco: we're almost evenly split on using `bold` (equivalent to `700`), and `600` directly. we can probably consolidate these but will need to check the BDL and possibly use a variable to describe it

src/elements/content-preview/preview-header/PreviewHeader.js
```diff
@@ -64,82 +64,82 @@ const PreviewHeader = ({
...
+        <header
+            className={classNames('bcpr-PreviewHeader', {
+                'bcpr-PreviewHeader--basic': !isPreviewingCurrentVersion,
+            })}
+        >
+            <FileInfo file={file} version={selectedVersion} />
+
+            <div className="bcpr-PreviewHeader-controls">
+                {isPreviewingCurrentVersion && (
+                    <React.Fragment>
```
booc0mtaco: thoughts about `<React.Fragment>` versus using `<>`? I prefer the former, personally.

src/elements/content-preview/preview-header/PreviewHeader.scss
```diff
@@ -0,0 +1,109 @@
...
+                fill: $bdl-gray-10;
+                vertical-align: middle;
+            }
+        }
+    }
+}
+
+.bcpr-PreviewHeader--basic {
+    background: $bdl-gray;
+    border-bottom: 1px solid $bdl-gray;
```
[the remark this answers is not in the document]
jstoffan: It's meant to match the background color, as otherwise it creates an unpleasant break between the header and the content.

src/elements/content-preview/preview-header/FileInfo.scss
jstoffan: Okay. Is there something I need to change?

src/elements/content-preview/preview-header/PreviewHeader.js
jstoffan: I prefer the former, too. I don't think we have a standard for it.

src/elements/content-preview/preview-header/FileInfo.scss
booc0mtaco: From looking at src/elements/common/fonts.scss, looks like we only set the face and weight for `normal` and `bold`. My guess is we're meant to use those two, and not the number values

PR #1485, opened 2019-08-02
refactor(async): Update async load helper to forward refs

jstoffan:
Needed for #1472.

### DIFF
--- a/src/elements/common/async-load/AsyncError.js
+++ b/src/elements/common/async-load/AsyncError.js
@@ -1,0 +1,32 @@
+// @flow
+import * as React from 'react';
+
+type Props = {
+    children: React.Node,
+    component: React.ComponentType<any>,
+};
+
+type State = {
+    error: ?Error,
+};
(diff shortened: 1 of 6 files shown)


boxcla:

src/elements/common/async-load/AsyncError.js
```diff
@@ -0,0 +1,32 @@
...
+type Props = {
+    children: React.Node,
+    component: React.ComponentType<any>,
+};
+
+type State = {
+    error: ?Error,
+};
+
+class AsyncError extends React.Component<Props, State> {
```
ConradJChan: Could this somehow be combined with ErrorBoundary?

jstoffan: Probably, but it'll take more time. ErrorBoundary has extra logic for logging and prop forwarding, which doesn't translate directly to this simpler use case. We could tackle it in a follow-up?

PR #1486, opened 2019-08-02
feat(content-explorer) - grid view generate thumbnails

rickstefanik:
In testing, I noticed some file types that should have thumbnails were showing the generic Box Icons instead.  In some of these cases, the status of the response of the Representations API was `none`.  In looking at the docs for the Representations API, I found that `none` means the thumbnail has not been generated, but can be generated by polling the item's `info.url`.  This PR polls those urls to generate thumbnails that have not been generated yet.


boxcla:

src/api/File.js
```diff
@@ -78,29 +79,71 @@ class File extends Item {
...
+        const { id } = item;
+        const infoUrl = getProp(item, 'representations.entries[0].info.url');
+
+        if (!infoUrl) {
+            return;
+        }
+
+        let status;
+        do {
+            // eslint-disable-next-line no-await-in-loop
```
rickstefanik: From [https://eslint.org/docs/rules/no-await-in-loop](url) on when not to use `no-await-in-loop` rule:
>loops may be used to retry asynchronous operations that were unsuccessful

I think this scenario fits that case.

src/api/File.js
```diff
@@ -78,29 +79,71 @@ class File extends Item {
...
+            status = getProp(await this.xhr.get({ url: infoUrl }), 'data.status.state');
```
rickstefanik: Would it be worth it to put in some kind of delay here rather than constantly polling the url to see if the thumbnail is generated?

src/api/File.js
```diff
@@ -78,29 +79,71 @@ class File extends Item {
...
         const { query, url: thumbnailBaseUrl } = queryString.parseUrl(thumbnailUrl);
         const thumbnailUrlParams = { ...query, access_token: token };
         const thumbnailUrlQuery = queryString.stringify(thumbnailUrlParams);
+        const result = `${thumbnailBaseUrl}?${thumbnailUrlQuery}`;
+
+        if (status !== 'success') {
+            if (callback) {
+                this.pollInfoUrl(item, result, callback);
+            }
+            return null;
```
rickstefanik: Idea is that generating thumbnails could take time, so better to return null now and let the callback of `pollInfoUrl` update state when the thumbnail is generated.

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -780,6 +782,21 @@ class ContentExplorer extends Component<Props, State> {
...
+     * @param {string} id - id of BoxItem to update
+     * @param {string} thumbnailUrl - thumbnailUrl to update BoxItem with
+     * @return {void}
+     */
+    updateItemThumbnailUrl = (id: string, thumbnailUrl: string): void => {
+        const { currentCollection } = this.state;
+        const { items = [] } = currentCollection;
+        const newCollection = { ...currentCollection };
+        newCollection.items = items.map(item => (item.id === id ? { ...cloneDeep(item), thumbnailUrl } : item));
+        this.setState({ currentCollection: newCollection });
```
rickstefanik: This function will be called for each thumbnail that needs to be generated, which also means that `setState` will be called more often.  I think this may be worth it, since generating the thumbnails could vary between files, and it's better to display them one by one than wait for all of them to finish generating.

src/api/File.js
```diff
@@ -78,29 +79,71 @@ class File extends Item {
...
+    /**
+     * Polls an item's infoUrl until it's representation is generated or an error occurrs.
+     * Calls callback if representation was successfully generated.
+     *
+     * @param {BoxItem} item - BoxItem that needs infoUrl polled
+     * @param {string} thumbanilUrl - thumbanil url of item
+     * @param {string, string) => void} callback - function to call if polling results in success
+     * @return {Promise<void>}
+     */
+    async pollInfoUrl(
```
ConradJChan: Might make sense to generalize this a bit more. Maybe something like `getRepresentation(item, type, polling)` that will get the representation type and do some sort of polling if the rep is not ready and polling is desired

src/api/File.js
ConradJChan: As discussed, check out `retryNumOfTimes` from `utils/function`

rickstefanik: Looking into using `retryNumOfTimes` in `utils/function` to do an exponential backoff on requests.

src/api/File.js
```diff
@@ -78,29 +80,81 @@ class File extends Item {
...
+        item: BoxItem,
+        thumbnailUrl: string,
+        callback: (id: string, thumbnailUrl: string) => void = noop,
+    ): Promise<void> {
+        const { id } = item;
+        const infoUrl = getProp(item, 'representations.entries[0].info.url');
+
+        const numOfTries = 8;
+        const initialTimeout = 1000;
+        const backoffFactor = 2;
```
rickstefanik: I think this should equate to waiting a maximum of approximately `2^8 = 256` seconds for a representation to generate.

rickstefanik:
Just implemented some architecture changes with regards to Conrad's comment.  I might have missed the mark though, so it may be worth it to hold off on reviewing until I go over it with Conrad.

src/api/File.js
```diff
@@ -78,6 +82,44 @@ class File extends Item {
...
+
+        const numOfTries = 3;
+        const initialTimeout = 3000;
+        const backoffFactor = 2;
+
+        if (!infoUrl) {
+            return representation;
+        }
+
+        const pollInfoUrl = async (resolve: Function, reject: Function): Promise<?string> => {
```
ConradJChan: This might be better served as a normal promise chain callback like:
```
    const isStatusSuccess = response => {
        const status = getProp(response, 'data.status.state');
        return !status || status === REPRESENTATIONS_RESPONSE_SUCCESS || status === REPRESENTATIONS_RESPONSE_ERROR);
    }
    return this.xhr.get({ url: infoUrl }).catch(() =>
        retryNumOfTimes(
            (successCallback, errorCallback) =>
                this.xhr
                    .get({ url: infoUrl })
                    .then(response => (isStatusSuccess(response) ? successCallback(response.data) : errorCallback())
                    .catch(errorCallback),
            numOfTries,
            initialTimeout,
            backoffFactor,
        ),
    );
```

rickstefanik:
Architecture changes have been made.

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -780,6 +787,44 @@ class ContentExplorer extends Component<Props, State> {
...
+        const { representations } = item;
+        const representation = getProp(representations, 'entries[0]');
+        const fileAPI = this.api.getFileAPI(false);
+
+        if (isFeatureEnabled(features, 'contentExplorer.gridView.enabled')) {
+            const updatedRepresentation = await fileAPI.generateRepresentation(representation);
+            if (updatedRepresentation) {
+                this.updateItemInCollection({
+                    ...cloneDeep(item),
+                    representations: { entries: [updatedRepresentation] },
```
ConradJChan: Will this erase the other representations that were there on `item`?

src/api/File.js
```diff
@@ -80,38 +83,41 @@ class File extends Item {
...
      */
-    async generateRepresentation(item: BoxItem): Promise<boolean> {
-        const infoUrl = getProp(item, 'representations.entries[0].info.url');
+    async generateRepresentation(representation: ?FileRepresentation): Promise<?FileRepresentation> {
+        const infoUrl = getProp(representation, 'info.url');
 
-        const numOfTries = 8;
-        const initialTimeout = 1000;
+        const numOfTries = 3;
+        const initialTimeout = 3000;
```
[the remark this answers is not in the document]
rickstefanik: If we set `initialTimeout` to 500 that will mean we wait a total of 3.5 seconds for a thumbnail to generate.  In testing I've seen some thumbnails take longer than 3.5 seconds to generate.   Should I bump up the `backoffFactor`?  Or should we just decide that if a thumbnail takes more than 3.5 seconds we won't display it?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -796,16 +796,17 @@ class ContentExplorer extends Component<Props, State> {
...
         const { features } = this.props;
-        const { representations } = item;
-        const representation = getProp(representations, 'entries[0]');
+        const entries = getProp(item, 'representations.entries');
         const fileAPI = this.api.getFileAPI(false);
 
-        if (isFeatureEnabled(features, 'contentExplorer.gridView.enabled')) {
-            const updatedRepresentation = await fileAPI.generateRepresentation(representation);
+        if (isFeatureEnabled(features, 'contentExplorer.gridView.enabled') && entries) {
+            const updatedRepresentation = await fileAPI.generateRepresentation(entries[0]);
```
ConradJChan: It seems like doing the truthy check for `updatedRepresentation` is unnecessary since it always returns at least the same representation that was passed in. What we may want to to though is to consider doing an equality check to avoid an unnecessary `setState` call if nothing has changed. That or change `generateRepresentation` to return null in case the generation failed

src/api/File.js
```diff
@@ -92,32 +91,35 @@ class File extends Item {
         const infoUrl = getProp(representation, 'info.url');
 
         const numOfTries = 3;
-        const initialTimeout = 3000;
-        const backoffFactor = 2;
+        const initialTimeout = 500;
```
ConradJChan: Perhaps it would be worth checking with conversion to team to see what they recommend for a polling interval that would make sense?

src/api/File.js
```diff
@@ -78,6 +81,47 @@ class File extends Item {
...
+
+        return this.xhr
+            .get({ url: infoUrl })
+            .then(() =>
+                retryNumOfTimes(
+                    (successCallback, errorCallback) =>
+                        this.xhr
+                            .get({ url: infoUrl })
+                            .then(response =>
+                                isStatusSuccess(response) ? successCallback(response.data) : errorCallback(),
```
mickr: Should the error be surfaced somewhere? 🍹

rickstefanik:
Added tests and squashed.

src/api/File.js
```diff
@@ -78,6 +82,56 @@ class File extends Item {
         this.successHandler(`${downloadBaseUrl}?${downloadUrlQuery}`);
     }
 
+    /**
+     * Polls a representation's infoUrl, attempting to generate a representation
+     *
+     * @param {FileRepresentation} representation - reprsentation that should have it's info.url polled
```
ConradJChan: `reprsentation`

src/api/File.js
```diff
@@ -78,6 +82,56 @@ class File extends Item {
...
+                            )
+                            .catch(e => {
+                                errorCallback(e);
+                            }),
+                    numOfTries,
+                    initialTimeout,
+                    backoffFactor,
+                ),
+            )
+            .catch(() => representation);
```
ConradJChan: Should this return the response from the xhr call? i.e. if the info.url returned rep status of error?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -760,16 +762,21 @@ class ContentExplorer extends Component<Props, State> {
             itemThumbnails = await Promise.all(items.map(item => fileAPI.getThumbnailUrl(item)));
```
ConradJChan: Should we account for folders?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -760,16 +762,21 @@ class ContentExplorer extends Component<Props, State> {
...
+            const currentItem = isSelected ? selectedItem : item;
             const thumbnailUrl = isGridViewEnabled ? itemThumbnails[index] : null;
+
             const newItem = {
                 ...currentItem,
                 selected: isSelected,
                 thumbnailUrl,
             };
 
+            if (thumbnailUrl && !isThumbnailReady(newItem)) {
```
ConradJChan: Should we do this only for files?

src/api/File.js
```diff
@@ -78,6 +82,56 @@ class File extends Item {
...
+                            )
+                            .catch(e => {
+                                errorCallback(e);
+                            }),
+                    numOfTries,
+                    initialTimeout,
+                    backoffFactor,
+                ),
+            )
+            .catch(e => e);
```
ConradJChan: I think I was mistaken here, we should probably just let the promise rejection return as normal instead of returning it as a promise

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -786,6 +797,45 @@ class ContentExplorer extends Component<Props, State> {
...
+     * @param {BoxItem} item - item to generate thumbnail for
+     * @return {Promise<void>}
+     */
+    attemptThumbnailGeneration = async (item: BoxItem): Promise<void> => {
+        const { features } = this.props;
+        const entries = getProp(item, 'representations.entries');
+        const representation = getProp(entries, '[0]');
+
+        if (isFeatureEnabled(features, 'contentExplorer.gridView.enabled') && representation) {
+            const updatedRepresentation = await this.api.getFileAPI(false).generateRepresentation(representation);
```
ConradJChan: Then we can get rid of the instanceof check on L814

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -763,19 +765,28 @@ class ContentExplorer extends Component<Props, State> {
...
             const thumbnailUrl = isGridViewEnabled ? itemThumbnails[index] : null;
+
             const newItem = {
                 ...currentItem,
                 selected: isSelected,
                 thumbnailUrl,
             };
 
+            if (item.type === TYPE_FILE && thumbnailUrl && !isThumbnailReady(newItem)) {
+                this.attemptThumbnailGeneration(newItem);
```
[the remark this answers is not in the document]
rickstefanik: In this case I don't think so.  Some items may have their thumbnails already generated, so we want to set those immediately.  Items that don't already have their thumbnails generated will fall into this if statement.  Since thumbnail generation could take some time, I think we want to do this in the background, without `await`.  Otherwise, all thumbnails would have to wait for all other thumbnails to be generated.

rickstefanik:
Still happy to make any changes if there are more comments and get this PR merged.

sahiga:
Hey @rickstefanik, sorry it's taken us so long to merge this PR. It looks great, and I'd like to merge it this week if possible. I'll take care of the merge conflicts, but could you sign the CLA agreement? 🙏

rickstefanik:
Screen Shot 2020-04-01 at 2 49 23 PM
@sahiga Just signed the CLA! The status still says pending on GitHub, but I'm guessing it might take a little bit for that to update?

sahiga:
Thanks @rickstefanik! I'll add another commit to fix the merge conflicts. The CLA check should run again after that.

boxcla:
Manually verified that rickstefanik signed CLA at Apr 1, 2020 11:29:40 AM

sahiga:
@rickstefanik Looks like the CLA check is still pending. Could you try adding the email associated with your commits (rstefanik@box.com) to your GitHub account?

src/api/File.js
```diff
@@ -81,6 +85,51 @@ class File extends Item {
...
+    /**
+     * Polls a representation's infoUrl, attempting to generate a representation
+     *
+     * @param {FileRepresentation} representation - representation that should have it's info.url polled
+     * @return {Promise<FileRepresentation>} - representation updated with most current status
+     */
+    async generateRepresentation(representation: FileRepresentation): Promise<FileRepresentation> {
+        const infoUrl = getProp(representation, 'info.url');
+
+        const numOfTries = 4;
```
jstoffan, line 97: Do these values need to be different than the default values of `retryNumOfTimes`? If so, do they need to be declared as separate variables? If so, should they be declared outside of the function as constants?

src/api/File.js
```diff
@@ -81,6 +85,51 @@ class File extends Item {
...
+            const status = getProp(response, 'data.status.state');
+            return (
+                !status ||
+                status === REPRESENTATIONS_RESPONSE_ERROR ||
+                status === REPRESENTATIONS_RESPONSE_SUCCESS ||
+                status === REPRESENTATIONS_RESPONSE_VIEWABLE
+            );
+        };
+
+        return this.xhr.get({ url: infoUrl }).then(() =>
```
jstoffan, line 115: Does this first call need to be duplicated or can we return `retryNumOfTimes` directly?

src/api/File.js
```diff
@@ -81,6 +85,51 @@ class File extends Item {
...
+            retryNumOfTimes(
+                (successCallback, errorCallback) =>
+                    this.xhr
```
jstoffan, line 118: Should we be using `this.get` instead of `this.xhr.get`?

src/api/File.js
```diff
@@ -93,16 +142,14 @@ class File extends Item {
         const template = getProp(entry, 'content.url_template');
         const token = await TokenService.getReadToken(getTypedFileId(item.id), this.options.token);
 
-        if (getProp(entry, 'status.state') !== 'success' || !extension || !template || !token) {
```
jstoffan, line 96: Does this have any unintended consequences for other thumbnail usages?

src/api/File.js
```diff
@@ -81,6 +85,51 @@ class File extends Item {
...
+
+        const numOfTries = 4;
+        const initialTimeout = 2000;
+        const backoffFactor = 2;
+
+        if (!infoUrl) {
+            return representation;
+        }
+
+        const isStatusSuccess = response => {
```
jstoffan, line 105: Does this need to be redeclared every time the method is called?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -840,22 +841,37 @@ class ContentExplorer extends Component<Props, State> {
      * @return {void}
      */
     async updateCollection(collection: Collection, selectedItem: ?BoxItem, callback: Function = noop): Object {
+        // const { features } = this.props;
```
swfree, line 844: should this line be removed?

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -840,22 +841,37 @@ class ContentExplorer extends Component<Props, State> {
...
         const { items = [] } = collection;
         const fileAPI = this.api.getFileAPI(false);
         const newCollection: Collection = { ...collection };
         const selectedId = selectedItem ? selectedItem.id : null;
-        const thumbnails = await Promise.all(items.map(item => fileAPI.getThumbnailUrl(item)));
+        // const thumbnails = await Promise.all(items.map(item => fileAPI.getThumbnailUrl(item)));
```
swfree, line 849: same here?

src/api/File.js
sahiga: Yup, these values don't need to be declared as separate variables. I'll move them into the function call.

There are a couple of discussions about the polling values above: https://github.com/box/box-ui-elements/pull/1486#discussion_r310803697 and https://github.com/box/box-ui-elements/pull/1486#discussion_r311165030. Given that some representations take longer to generate, I think it makes sense to set a value higher than the default `initialTimeout` specified in `retryNumberOfTimes()`.

src/api/File.js
sahiga: Looks like this function only works with `this.xhr.get`, because of the custom success/error callback behavior. Is there a strong preference for `this.get` over `this.xhr.get`?

src/api/File.js
sahiga: So far, I haven't found any. This function is only called within the `updateCollection` function in the ContentExplorer.

sahiga:
@jstoffan @swfree I updated the code based on your suggestions. Could you review again? 🙏

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -844,18 +845,29 @@ class ContentExplorer extends Component<Props, State> {
...
+            const thumbnailUrl = itemThumbnails[index];
+
             const newItem = {
                 ...currentItem,
                 selected: isSelected,
-                thumbnailUrl: thumbnails[index],
+                thumbnailUrl,
             };
 
+            if (item.type === TYPE_FILE && thumbnailUrl && !isThumbnailReady(newItem)) {
```
ConradJChan, line 867: Seems like `updateCollection` gets called in various ways, after view change (search, recents, folder) as well as page operations (select, sharedlink). Is it possible that the thumbnail generation could get duplicated for instance after each check selection user interaction?

sahiga: Just tested these flows. `attemptThumbnailGeneration()` only gets called when `isThumbnailReady()` returns false, and that result only occurs if the representation status is neither "viewable" nor "success." So, thumbnails that have already been generated will not be generated again.

ConradJChan: 👍 thanks for checking

PR #1490, opened 2019-08-05
fix(i18n): Fix bug where messages starting with HTML cause an exception

ehoogerbeets:
A message like this:

```jsx
<FormattedCompMessage
  description="Body for tooltip pointing to add user button"
    id="adminConsole.components.Onboarding.addUserButtonTooltipBodyText"
>
    <b>Managed users</b> are employees in your Box Enterprise. They often share your company domain(s) and require a greater amount of control, while <b>External users</b> are clients or partners who have been invited to collaborate on specific content in your enterprise.
</FormattedCompMessage>
```

Caused an exception even though it is perfectly valid.

### DIFF
--- a/i18n/en-US.properties
+++ b/i18n/en-US.properties
@@ -208,6 +208,8 @@ be.deleteDialogFileText = Are you sure you want to delete {name}?
 be.deleteDialogFolderText = Are you sure you want to delete {name} and all its contents?
 # Label for delete confirmation dialog
 be.deleteDialogLabel = Confirm Delete
+# Aria label for button to delete a comment, task, or app activity
+be.deleteLabel = Delete
 # Label for the description field in the preview sidebar.
 be.description = Description
 # Placeholder for file description in preview sidebar.
@@ -256,6 +258,10 @@ be.getVersionInfo = Get version information
 be.gridView = Switch to Grid View
 # Label for in action.
 be.in = In
+# Button text to cancel inline item deletion
+be.inlineDeleteCancel = No
+# Button text to confirm inline item deletion
+be.inlineDeleteConfirm = Yes
 # Text for last accessed date with last access prefix.
 be.interactedDate = Last accessed on {date}
 # Label for item created date.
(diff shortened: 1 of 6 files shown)


boxcla:

i18n/en-US.properties
```diff
@@ -208,6 +208,8 @@ be.deleteDialogFileText = Are you sure you want to delete {name}?
 be.deleteDialogFolderText = Are you sure you want to delete {name} and all its contents?
 # Label for delete confirmation dialog
 be.deleteDialogLabel = Confirm Delete
+# Aria label for button to delete a comment, task, or app activity
+be.deleteLabel = Delete
```
jstoffan: Are these changes related to this PR?

src/components/i18n/Composition.js
```diff
@@ -64,6 +65,15 @@ class Composition {
         return this.ma.getMinimalString();
     }
 
+    /**
+     * @private
+     */
+    nextKey() {
+        const ret = `key${this.keyIndex}`;
```
jstoffan: We usually try to avoid abbreviations, where possible, as they can hinder readability.

src/components/i18n/FormattedCompMessage.js
```diff
@@ -150,7 +150,7 @@ class FormattedCompMessage extends React.Component<Props, State> {
     }
 
     render() {
-        const { count, tagName, intl, description, id, ...rest } = this.props;
+        const { count, tagName, intl, description, id, defaultMessage, ...rest } = this.props;
```
jstoffan: It looks like `defaultMessage` is unused. Is that intentional?

i18n/en-US.properties
ehoogerbeets: No, git push automatically put them there

src/components/i18n/FormattedCompMessage.js
ehoogerbeets: Yes, it was getting into the "rest" which made its way into the output. Anything that is not a prop we use directly should go onto the output span element like className for example.

i18n/en-US.properties
ehoogerbeets: I just did a rebase onto the latest from master, so I am not sure why that is happening...

src/components/i18n/Composition.js
ehoogerbeets: You would prefer something like "returnValue" perhaps?

jstoffan: There's always good ol' `result`.

i18n/en-US.properties
jstoffan: Try deleting your `/es` directory, then run `./scripts/prepush.sh`.

ehoogerbeets: Yeah, talked with Andrew H and he essentially said the same thing... someone deleted some strings that still existed in my es dir. He recommended to do "yarn clean" and "yarn build" after every rebase. Now the en-US.properties is gone from the PR.

src/components/i18n/Composition.js
ehoogerbeets: Changed to "result"

i18n/en-US.properties
booc0mtaco: to be clear, this can occur if there are stale values in `i18n/json` and a rebased commit contains a deletion to some content referenced there

PR #1491, opened 2019-08-05
feat(tasks): add any task icon + tooltip

shikharmohan:
image


boxcla:

shikharmohan:
@alexkrolick Fair point that it can be moved around independent of status. I'll update this PR to reflect that

src/icons/general/IconAnyTask.js
```diff
@@ -0,0 +1,35 @@
...
+    className?: string,
+    color?: string,
+    height?: number,
+    /** A text-only string describing the icon if it is not purely decorative for accessibility */
+    title?: string,
+    width?: number,
+};
+
+const IconAnyTask = ({ className = '', color = bdlGray62, height = 10, title, width = 11 }: Props) => (
+    <AccessibleSVG className={className} height={height} title={title} viewBox="0 0 10 11" width={width}>
```
booc0mtaco: Hmm, not a good viewbox. We were working toward them being consistently exported, and they should be 16x16 / 32x32.

src/elements/content-sidebar/activity-feed/task-new/Task.scss
```diff
@@ -27,6 +27,10 @@ $bcs-task-left-offset: 10 * $bcs-task-grid + 2px;
 
 .bcs-Task-statusContainer {
     margin-top: 2 * $bcs-task-grid;
+
+    .bcs-TaskCompletionRuleIcon {
```
alexkrolick: I think if you are using this classname it can live in a file imported by the icon itself. If it is passed in from the parent `.bcs-Task-completionRule` would probably be the preferred way of using the naming convention. In both cases it doesn't need nesting; that would create unnecessary specificity.

src/icons/general/IconAnyTask.js
priyajeet: Probably preferred 32x32 if shrinking it to desired size doesn't cause artifacts.

src/elements/content-sidebar/activity-feed/task-new/TaskCompletionRuleIcon.js
```diff
@@ -0,0 +1,28 @@
...
+import type { TaskCompletionRule } from '../../../../common/types/tasks';
+
+import './TaskCompletionRuleIcon.scss';
+
+type Props = {|
+    completionRule: ?TaskCompletionRule,
+|};
+
+const TaskCompletionRuleIcon = ({ completionRule }: Props): React.Node => (
+    <React.Fragment>
```
jstoffan: It seems like the condition should live in the parent, but if not, this can still be simplified to something like the following:

```
const TaskCompletionRuleIcon = ({ completionRule }: Props): React.Node => ( 
    completionRule === TASK_COMPLETION_RULE_ANY &&  (
        <Tooltip position="top-center" text={<FormattedMessage {...messages.taskAnyAffordanceTooltip} />}>
            <IconAnyTask height={11} width={11} className="bcs-TaskCompletionRuleIcon" />
        </Tooltip>
    )
);
```

src/elements/content-sidebar/activity-feed/task-new/TaskCompletionRuleIcon.js
```diff
@@ -0,0 +1,28 @@
...
+        {completionRule === TASK_COMPLETION_RULE_ANY ? (
+            <Tooltip position="top-center" text={<FormattedMessage {...messages.taskAnyAffordanceTooltip} />}>
+                <span>
```
jstoffan: Is this span needed?

src/elements/content-sidebar/activity-feed/task-new/TaskCompletionRuleIcon.js
shikharmohan: When we add new completion rules in the future, then all the logic to decide what icon to render can live in this tooltip

src/elements/content-sidebar/activity-feed/task-new/TaskCompletionRuleIcon.js
shikharmohan: Not needed updating.

shikharmohan: Actually @jstoffan looks like the span is needed...not sure why exactly. But the tooltip fails to render with only an SVG as it's child. The span is used by the tether component to render the tooltip and apply all the tooltip classes. 

All usages of tooltips in BUIE have a div or span as a child. `InfoIconWithTooltip` as an example.

alexkrolick: The SVG might need to forward refs or spread props

shikharmohan: I think it's because of this line in Tooltip.js 

`        const bodyEl = bodyElement instanceof HTMLElement ? bodyElement : document.body;`

SVG's aren't instances of HTMLElement

PR #1497, opened 2019-08-06
 feat(uploads): resume failed uploads from uploads manager

rosepierce:
Adding functionality to resume uploads for files of size >100 MB that have failed from where in the upload the file failed, rather than restarting the upload from the beginning. Additionally adding a button on the Uploads Manager header which will resume any items in the Uploads Manager that have failed.

### DIFF
--- a/src/api/uploads/MultiputUpload.js
+++ b/src/api/uploads/MultiputUpload.js
@@ -14,6 +14,7 @@ import {
     DEFAULT_RETRY_DELAY_MS,
     ERROR_CODE_UPLOAD_STORAGE_LIMIT_EXCEEDED,
     HTTP_STATUS_CODE_FORBIDDEN,
+    MS_IN_S,
 } from '../../constants';
 import MultiputPart, {
     PART_STATE_UPLOADED,
@@ -77,6 +78,8 @@ class MultiputUpload extends BaseMultiput {
     numPartsUploaded: number;
 
     numPartsUploading: number;
+
+    numResumeRetries: number;
 
     sessionEndpoints: Object;
 
@@ -387,7 +390,7 @@ class MultiputUpload extends BaseMultiput {
         this.fileId = fileId;
         this.sessionId = sessionId;
 
-        this.getSessionInfo(sessionId);
+        this.getSessionInfo();
     }
 
     /**
@@ -397,9 +400,9 @@ class MultiputUpload extends BaseMultiput {
      * @private
      * @return {void}
      */
-    getSessionInfo = async (session: string): Promise<any> => {
+    getSessionInfo = async (): Promise<any> => {
         const uploadUrl = this.getBaseUploadUrl();
-        const sessionUrl = `${uploadUrl}/files/upload_sessions/${session}`;
+        const sessionUrl = `${uploadUrl}/files/upload_sessions/${this.sessionId}`;
         try {
             const response = await this.xhr.get({ url: sessionUrl });
             this.getSessionSuccessHandler(response.data);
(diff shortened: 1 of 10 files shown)


boxcla:

src/elements/content-uploader/UploadsManagerAction.js
```diff
@@ -0,0 +1,34 @@
...
+    onClick: Function,
+};
+
+const UploadsManagerAction = ({ onClick, numFailedUploads }: Props) => {
+    const handleResumeClick = event => {
+        event.stopPropagation();
+        onClick(STATUS_ERROR);
+    };
+
+    const resumeMessage =
```
priyajeet: Similar to a prior comment:

const resumeMessage = numFailedUploads > 1 ? message.resumeAll : messages.resume;
...
<FormattedMessage {...resumeMessage} />

priyajeet: And additionally, does numFailedUploads need to be passed in as-is, or can we have a boolean prop `hasMultipleUploadsFailed` which abstracts out how its calculated from the perspective of this function

src/api/uploads/MultiputUpload.js
```diff
@@ -329,6 +341,198 @@ class MultiputUpload extends BaseMultiput {
...
+        }
+
+        if (errorData && errorData.status === 429) {
+            let retryAfterMs = DEFAULT_RETRY_DELAY_MS;
+            if (errorData.headers) {
+                const retryAfterSec = parseInt(
+                    errorData.headers['retry-after'] || errorData.headers.get('Retry-After'),
+                    10,
+                );
+                if (!Number.isNaN(retryAfterSec)) {
```
priyajeet: FYI, Number.isNaN doesn't work in IE. Likely works fine in webapp due to polyfill.
But can use `import isNaN from 'lodash/isNaN';` which we do in other files.

src/api/uploads/MultiputUpload.js
```diff
@@ -329,6 +341,198 @@ class MultiputUpload extends BaseMultiput {
...
+        if (this.numResumeRetries > this.config.retries) {
+            this.errorCallback(errorData);
+            return;
+        }
+
+        if (errorData && errorData.status === 429) {
+            let retryAfterMs = DEFAULT_RETRY_DELAY_MS;
+            if (errorData.headers) {
+                const retryAfterSec = parseInt(
+                    errorData.headers['retry-after'] || errorData.headers.get('Retry-After'),
```
priyajeet: I think this code is probably repeated from somewhere in the inheritance, can make it a function and just call that to get retury after

src/api/uploads/MultiputUpload.js
```diff
@@ -329,6 +341,198 @@ class MultiputUpload extends BaseMultiput {
...
+            // Restart upload process for errors resulting from invalid session
+            this.parts.forEach(part => {
+                part.cancel();
+            });
+            this.parts = [];
+            // Reset information about uploaded parts
+            this.fileSha1 = null;
+            this.totalUploadedBytes = 0;
+            this.numPartsNotStarted = 0;
+            this.numPartsDigestComputing = 0;
```
priyajeet: Isn't some/most of this also happening in the constructor? Maybe should be extracted out as a reset method?

Likewise for anything else thats being duplicated in bulk

src/elements/content-uploader/ContentUploader.js
```diff
@@ -746,7 +786,7 @@ class ContentUploader extends Component<Props, State> {
         const { onUpload, useUploadsManager } = this.props;
 
         item.progress = 100;
-        if (!item.error) {
+        if (item.status !== STATUS_ERROR) {
```
danubevictoria: Why are we changing this to check status? Should it be that we only mark the status as complete if there is no error?

src/elements/content-uploader/ContentUploader.js
```diff
@@ -917,14 +958,36 @@ class ContentUploader extends Component<Props, State> {
                 this.removeFileFromUploadQueue(item);
                 break;
             case STATUS_ERROR:
-                this.resetFile(item);
-                this.uploadFile(item);
+                if (isResumableUploadsEnabled && chunked && item.api.sessionId) {
```
danubevictoria: Are there any clients where chunked upload is not supported?

src/api/uploads/__tests__/MultiputUpload-test.js
```diff
@@ -418,6 +419,183 @@ describe('api/uploads/MultiputUpload', () => {
         });
     });
 
+    describe('getSessionSuccessHandler()', () => {
+        test('should commit if all parts have been uploaded', () => {
+            // Setup
+            multiputUploadTest.numPartsUploaded = 20;
+            const response = {
```
mickr: Can this be moved to a `beforeEach`?

src/api/uploads/MultiputUpload.js
```diff
@@ -340,7 +523,9 @@ class MultiputUpload extends BaseMultiput {
      * @return {Promise}
      */
     async sessionErrorHandler(error: ?Error, logEventType: string, logMessage?: string): Promise<any> {
-        this.destroy();
+        if (!this.isResumableUploadsEnabled) {
```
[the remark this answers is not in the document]
rosepierce: MultiputUpload is only used for the upload process after a check for both `chunked` and if multiput is supported in ContentUploader, and if these aren't supported for the given item and client then the upload process goes through PlainUpload instead, so I don't think those checks need to be performed here or at line 546.

src/api/uploads/MultiputUpload.js
```diff
@@ -329,6 +341,198 @@ class MultiputUpload extends BaseMultiput {
...
+    }
+
+    /**
+     * Handle error from getting upload session.
+     * Restart uploads without valid sessions from the beginning of the upload process.
+     *
+     * @param error
+     * @return {void}
+     */
+    getSessionErrorHandler(error: Object): void {
```
mickr: Is there a more descriptive type that can be used for the error object?

src/api/uploads/MultiputUpload.js
```diff
@@ -329,6 +381,171 @@ class MultiputUpload extends BaseMultiput {
...
+
+    /**
+     * Get session information from API.
+     * Uses session info to commit a complete session or continue an in-progress session.
+     *
+     * @private
+     * @return {void}
+     */
+    getSessionInfo = async (): Promise<any> => {
+        const uploadUrl = this.getBaseUploadUrl();
```
Alhalalsheh: would the uploadUrl include the /api/2.0 or 2.1?

src/api/uploads/MultiputUpload.js
```diff
@@ -329,6 +381,171 @@ class MultiputUpload extends BaseMultiput {
...
+        const sessionUrl = `${uploadUrl}/files/upload_sessions/${this.sessionId}`;
```
Alhalalsheh: make sure this is a GET ... just confirming

src/api/uploads/MultiputUpload.js
```diff
@@ -329,6 +381,171 @@ class MultiputUpload extends BaseMultiput {
...
+
+    /**
+     * Handles a getSessionInfo success and either commits the session or continues to process
+     * the parts that still need to be uploaded.
+     *
+     * @param response
+     * @return {void}
+     */
+    getSessionSuccessHandler(data: any): void {
+        const { total_parts, part_size, session_endpoints } = data;
```
Alhalalsheh: it worth checking "session_expires_at". if the session is already expired, you just can not resume the download

src/api/uploads/MultiputUpload.js
```diff
@@ -329,6 +381,171 @@ class MultiputUpload extends BaseMultiput {
...
+                this.numPartsNotStarted += 1;
+            }
+            nextUploadIndex += 1;
+        }
+
+        if (this.numPartsUploaded === total_parts) {
+            // Commit sessions that have all the parts of the file uploaded
+            this.commitSession();
+        } else {
+            this.processNextParts();
```
Alhalalsheh: im assuming processNextParts will eventually direct you to the commitSession flow ... and im assuming commitSession will perform that sha calculation to insure file integrity

src/api/uploads/MultiputUpload.js
rosepierce: This uploadUrl uses /api/2.0, which is also being used for the rest of the upload process - is 2.0 or 2.1 what should be used to use chunked uploads?

src/api/uploads/MultiputUpload.js
rosepierce: yes, when all the parts have been successfully uploaded processNextParts calls commitSession, which sends the sha calculation that has been performed during the file upload with the POST to commit the session

src/api/uploads/MultiputUpload.js
rosepierce: From what I saw in the documentation for the chunked upload API, I thought that any request made with a session ID for a session that has expired would return a error, and so the scenario where the session has expired is being handled in the getSessionErrorHandler. Is this not necessarily the case that an error would be returned?

src/api/uploads/MultiputUpload.js
Alhalalsheh: i see the browser today is also using 2.0, while we recommend using the latest version, i have no strong opinion to push you towards 2.1. 2.0 is fine by me

src/api/uploads/MultiputUpload.js
Alhalalsheh: cool ... do we have a unit test for this? sorry i did not go through unit tests ... but just wanted to confirm we have a unit test to exercise/confirm sha calculation at client side on multiput (chunked) upload scenario

src/api/uploads/MultiputUpload.js
Alhalalsheh: if you know that your session is expired why not to optimize and fail (or restart the upload) immediately? yet if you feel adding an extra check for session_expires_at will hurt the code readability i would skip it!

it is up to you to decide, yet if you decided to add session_expires_at check, make sure NOT to remove the "scenario where the session has expired is being handled in the getSessionErrorHandler"

PR #1500, opened 2019-08-07
feat(shared-link-settings-modal): Disable direct downloads if classified

feixuanli:
Screen Shot 2019-08-12 at 2 47 58 PM
Screen Shot 2019-08-12 at 2 48 11 PM


boxcla:

src/features/shared-link-settings-modal/AllowDownloadSection.js
```diff
@@ -35,23 +38,38 @@ const AllowDownloadSection = ({
         </div>
     );
 
+    const label = classification ? `"${classification}"` : null;
```
priyajeet: Is this variable needed? Can below be label: classification || null. Also if null, then the display message on tooltip will say null. In other places we just say that "disabled due to an applied access policy" since having a classification doesn't do the disabling, but its the access policy. So maybe for consistency we should match what we say in other tooltips.

src/features/shared-link-settings-modal/AllowDownloadSection.js
```diff
@@ -35,23 +38,38 @@ const AllowDownloadSection = ({
...
+    const tooltipMessage = { ...messages.directDownloadBlockedByAccessPolicy, values: { label } };
+    const downloadWrapperClass = isDirectLinkUnavailableDueToAccessPolicy
+        ? 'bdl-AllowDownloadSection-wrapper--disabled'
```
priyajeet: We normally try to avoid words like "wrapper" or "container" unless absolutely needed as they don't add any additional value to the CSS class.

Also in this case, you want to use `bdl-is-disabled`, as this is a disabled state vs -- which is for modifiers. One way to distinguish is when class is being added based on a condition, then its likely state. If you are permanently changing some CSS rule, then its a modifier.

src/features/shared-link-settings-modal/SharedLinkSettingsModal.scss
```diff
@@ -1,6 +1,16 @@
 @import '../../styles/variables';
 
 .shared-link-settings-modal {
+    /** title and classifiction label */
+    .bdl-SharedLinkSettingsModalTitle {
```
priyajeet: These new classes don't need to be nested, you can have all of them at the top level.
Only the bdl-is-disabled (and other similar states or modifiers) should be &.nested since they should never exist on their own all alone.

src/features/shared-link-settings-modal/SharedLinkSettingsModal.js
```diff
@@ -310,30 +311,52 @@ class SharedLinkSettingsModal extends Component {
...
         );
     }
 
+    renderModalTitle() {
+        const { item } = this.props;
+        const { bannerPolicy, classification } = item;
+        const title = <FormattedMessage {...messages.modalTitle} />;
+
+        return (
+            <span className="bdl-SharedLinkSettingsModalTitle">
```
priyajeet: `bdl-SharedLinkSettingsModal-title`, since SharedLinkSettingsModal  is the component's name, and you are decorating the title section of it.

src/features/shared-link-settings-modal/SharedLinkSettingsModal.js
```diff
@@ -310,30 +311,52 @@ class SharedLinkSettingsModal extends Component {
...
                 isDownloadEnabled={isDownloadEnabled}
                 onChange={this.onAllowDownloadChange}
             />
         );
     }
 
+    renderModalTitle() {
+        const { item } = this.props;
+        const { bannerPolicy, classification } = item;
+        const title = <FormattedMessage {...messages.modalTitle} />;
```
priyajeet: Unnecessary variable

src/features/shared-link-settings-modal/SharedLinkSettingsModal.js
```diff
@@ -310,30 +311,52 @@ class SharedLinkSettingsModal extends Component {
...
+
+        return (
+            <span className="bdl-SharedLinkSettingsModalTitle">
+                {title}
+                <Classification
+                    advisoryMessage={bannerPolicy ? bannerPolicy.body : undefined}
+                    messageStyle="tooltip"
+                    name={classification}
+                    className="bdl-SharedLinkSettingsModalTitle-classification"
```
priyajeet: In a similar manner, `bdl-SharedLinkSettingsModal-titleClassification` or perhaps just `bdl-SharedLinkSettingsModal-classification`

src/features/shared-link-settings-modal/AllowDownloadSection.js
Sunyang730: Nit: I would also recommend using `classnames` library.

src/features/shared-link-settings-modal/AllowDownloadSection.js
```diff
@@ -35,23 +38,38 @@ const AllowDownloadSection = ({
...
+                    >
+                        <Checkbox
+                            isChecked={isDownloadEnabled}
+                            isDisabled={!canChangeDownload || isDirectLinkUnavailableDueToAccessPolicy}
+                            label={<FormattedMessage {...messages.allowDownloadLabel} />}
+                            name="isDownloadEnabled"
+                            onChange={onChange}
+                            {...downloadCheckboxProps}
+                        />
+                        {(isDirectLinkAvailable || isDirectLinkUnavailableDueToDownloadSettings) && isDownloadEnabled
```
Sunyang730: can you set this as a constant? getting long here

src/features/shared-link-settings-modal/AllowDownloadSection.js
```diff
@@ -35,23 +38,38 @@ const AllowDownloadSection = ({
         </div>
     );
 
+    const label = classification ? `"${classification}"` : null;
+    const tooltipMessage = { ...messages.directDownloadBlockedByAccessPolicy, values: { label } };
```
Sunyang730: If label can be null, then that means there will be a situation where the message will be `Download for this content is disabled due to the classification`. Do we want to show the tooltip if label is null?

src/features/shared-link-settings-modal/AllowDownloadSection.js
```diff
@@ -5,16 +5,19 @@ import { FormattedMessage } from 'react-intl';
...
 
 const AllowDownloadSection = ({
     canChangeDownload,
+    classification,
     directLink,
     directLinkInputProps = {},
     downloadCheckboxProps = {},
     isDirectLinkAvailable,
     isDirectLinkUnavailableDueToDownloadSettings,
+    isDirectLinkUnavailableDueToAccessPolicy = false,
```
Sunyang730: @priyajeet is this component suppose to be so specific? I am curious if this could just be `isDirectLinkDisabled` and then have a prop of `reason`

src/features/shared-link-settings-modal/AllowDownloadSection.js
feixuanli: Thanks!

src/features/shared-link-settings-modal/AllowDownloadSection.js
priyajeet: Also once made into a const, you can remove the ?:

src/features/shared-link-settings-modal/AllowDownloadSection.js
```diff
@@ -35,23 +37,35 @@ const AllowDownloadSection = ({
         </div>
     );
 
+    const tooltipMessage = { ...messages.directDownloadBlockedByAccessPolicy };
+    const downloadWrapperClass = isDirectLinkUnavailableDueToAccessPolicy ? 'bdl-is-disabled' : null;
```
priyajeet: So state should not exist on its own. So you need to have a parent class.

downloadWrapperClass = classnames('bdl-AllowDownloadSection', { bdl-is-disabled: isDirectLinkUnavailableDueToAccessPolicy});

and in CSS

.bdl-AllowDownloadSection {
    &.bdl-is-disabled { ... }
}

src/features/shared-link-settings-modal/AllowDownloadSection.js
priyajeet: Its an older component :/ Could use some refactoring, also see all those proptypes at the bottom. But making all those changes is going way beyond the scope

src/features/shared-link-settings-modal/AllowDownloadSection.js
feixuanli: currently `.bdl-is-disabled` is nested within `.shared-link-settings-modal` in css,
should we still add `bdl-AllowDownloadSection` here ?

priyajeet: you can have that outside

priyajeet: but inside AllowDownloadSection.jsAllowDownloadSection.scss instead

src/features/shared-link-settings-modal/AllowDownloadSection.js
```diff
@@ -35,23 +37,35 @@ const AllowDownloadSection = ({
...
+                    isDisabled={!isDirectLinkUnavailableDueToAccessPolicy}
+                    text={<FormattedMessage {...tooltipMessage} />}
+                    position="middle-left"
+                >
+                    <Fieldset
+                        disabled={isDirectLinkUnavailableDueToAccessPolicy}
+                        title={<FormattedMessage {...messages.allowDownloadTitle} />}
+                    >
+                        <Checkbox
+                            isChecked={isDownloadEnabled}
```
mickr: I think this is confusing, if the the download is restricted, should it still show as checked?

feixuanli: `isDownloadEnbaled`is received as false if restricted.

feixuanli:
updated

src/features/shared-link-settings-modal/AllowDownloadSection.scss
```diff
@@ -0,0 +1,5 @@
+.bdl-AllowDownloadSectio {
```
tjgupta: Typo.

src/features/shared-link-settings-modal/__tests__/SharedLinkSettingsModal-test.js
```diff
@@ -261,10 +277,18 @@ describe('features/shared-link-settings-modal/SharedLinkSettingsModal', () => {
...
+            expect(section.prop('isDirectLinkUnavailableDueToAccessPolicy')).toEqual(
+                isDirectLinkUnavailableDueToAccessPolicy,
+            );
             expect(section.prop('onChange')).toEqual(wrapper.instance().onAllowDownloadChange);
         });
     });
 
+    describe('renderModalTitle()', () => {
+        const wrapper = getWrapper();
+        expect(wrapper).toMatchSnapshot();
```
tjgupta: Probably would be good to narrow the snapshot to just the title here since that's what this test is for, something like
```
expect(wrapper.find('.bdl-SharedLinkSettingsModal-title').toMatchSnapshot();
```

src/features/shared-link-settings-modal/messages.js
```diff
@@ -16,6 +16,18 @@ const messages = defineMessages({
...
         id: 'boxui.share.sharedLinkSettings.directLinkLabel',
     },
+    directDownloadBlockedByAccessPolicyWithClassification: {
+        defaultMessage: 'Download has been disabled for content due to the classification',
+        description:
+            'Text to show that direct link download is disabled due to applied shield access policy with classification',
+        id: 'boxui.share.sharedLinkSettings.directDownloadBlockedByAccessPolicyWithClassification',
+    },
+    directDownloadBlockedByAccessPolicyWithoutClassification: {
+        defaultMessage: 'Download has been disabled for content without Classification.',
```
priyajeet: Should lowercase Classification to match the other message.

src/features/shared-link-settings-modal/__tests__/SharedLinkSettingsModal-test.js
priyajeet: You can also just shallow mount the response of renderModalTitle() and match that snapshot.

feixuanli:
updated

PR #1507, opened 2019-08-12
feat(leftsidebar): Show drop zones only on hover

shahzadaziz:
This PR allows you to set `showDropZoneOnHover` for menu items in left sidebar. 

`showDropZoneOnHover`  when true it will only show drop veil on hover. 
`showDropZoneOnHover`  when false it will show drop veil for the duration of items being dragged.

Updated: GIF to show behavior on favorites 

favoritesDragDrop2

### DIFF
--- a/src/features/left-sidebar/LeftSidebarDropWrapper.js
+++ b/src/features/left-sidebar/LeftSidebarDropWrapper.js
@@ -1,5 +1,6 @@
 // @flow
 import * as React from 'react';
+import classNames from 'classnames';
 
 import './styles/LeftSidebarDropWrapper.scss';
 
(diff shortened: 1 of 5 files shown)


boxcla:

booc0mtaco:
more of a UX question; we show that you can't drag to other files by dimming them out. could we do the same for the left sidebar items that can't be dragged to?

src/features/left-sidebar/LeftSidebarDropWrapper.js
```diff
@@ -9,24 +9,67 @@ type Props = {
...
+            className = '',
+            isDragging = false,
+            showDropZoneOnHover = false,
+            message = '',
+            dropTargetRef,
+            ...rest
+        } = this.props;
+        const { dropZoneHover } = this.state;
+        const shouldShowVeil = isDragging && (showDropZoneOnHover ? dropZoneHover : true);
+        const hoverEventHandlers =
```
booc0mtaco: do we handle escape when dragging?

shahzadaziz: We do. `isDragging` is set false.

shahzadaziz:
> more of a UX question; we show that you can't drag to other files by dimming them out. could we do the same for the left sidebar items that can't be dragged to?

I discussed this with design. We might lean towards not showing any visual indicators at all when dragging and move towards drop zones on hover only.  Design wants it to behave similar to how some operating systems e.g Mac OS does it.

booc0mtaco:
@shahzadaziz yeah, the on-drag-start indicators is more of an older web convention. if we aim to move away from that or not is cool.

But, if we're aiming to masquerade as having macOS-like behavior, worth nothing what the HIG recommends, RE: the question about dimming out stuff

[User Interaction: Drag and Drop](https://developer.apple.com/design/human-interface-guidelines/macos/user-interaction/drag-and-drop/)

> **Clearly indicate when dropping isn’t possible**. During a drag, users should know when they can and can’t drop content. Some apps communicate this by displaying an insertion point, highlighting, or acceptance pointer badge only when dropping is possible—in other words, the lack of visual feedback indicates when dropping isn’t possible. Other apps display an explicit not allowed badge, such as a circle with a slash.

this would dim out the things that can't be dragged to in the left side bar, or change the cursor, etc.

> **Highlight a new destination when content is dragged above it**. When there are multiple destinations, highlight one at a time. When the source and destination containers are the same, highlighting is unnecessary. For example, in Finder, moving a document icon to a different location in the same folder window doesn’t highlight the folder window because doing so would cause confusion. In some cases, highlighting may occur if content is dragged completely out of a container and then reenters it. Highlighting should be removed when content is no longer positioned above a destination.

we're doing this stuff with your PR, which is nice

shahzadaziz:
@booc0mtaco Thanks for sharing this article. I will run the question of dimming out things that are not drop zones by design team. We can determine what kind of visual treatment we want to have for these and it should be consistent between Item list and Left sidebar. We can track this as an incremental improvement separately.

priyajeet:
Make sure you are sending PRs from your fork and not pushing branches to main repo

PR #1508, opened 2019-08-12
fix(sidebar): Fix sidebar not auto-opening for large viewports

jstoffan:
I also added some browser-based tests for the auto-open feature to hopefully avoid similar issues in the future.


boxcla:

PR #1495, opened 2019-08-06
feat(leftsidebar): LeftSidebar supports customLinkRenderer

krapspark:
LeftSidebar menu items support for `customLinkRenderer`.
This prop allows for changing Left Nav menu items with UI treatment or wrapping the `defaultNavLinkRenderer` with another component or both!

The PR isn't ready to be merged yet. I'm pushing it as a hand-off.

PR #1510, opened 2019-08-12
feat(left-sidebar): LeftSidebar supports custom navLinkRenderer

viodragon2:
Follow up to #1495 

This prop allows for changing Left Nav menu items with UI treatment or wrapping the `defaultNavLinkRenderer` with another component or both!

### DIFF
--- a/src/features/left-sidebar/LeftSidebar.js
+++ b/src/features/left-sidebar/LeftSidebar.js
@@ -16,6 +16,7 @@ import LeftSidebarIconWrapper from './LeftSidebarIconWrapper';
 import NewItemsIndicator from './NewItemsIndicator';
 import defaultNavLinkRenderer from './defaultNavLinkRenderer';
 
+import type { Props as LeftSidebarLinkProps } from './LeftSidebarLink';
 import type { Callout } from './Callout';
 
 import './styles/LeftSidebar.scss';
@@ -40,7 +41,7 @@ type SubMenuItem = {
     /** Localized text string to use for individual menu items */
     message: string,
     /** Optional left side bar link renderer. Defaults to defaultNavLinkRenderer */
-    navLinkRenderer?: Function,
+    navLinkRenderer?: (props: LeftSidebarLinkProps) => React.Node,
     /** Whether we should show a badge marking new item content */
     newItemBadge?: boolean,
     /** Optional remove link click handler */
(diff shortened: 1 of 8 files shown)


boxcla:
Verified that @viodragon2 has signed the CLA. Thanks for the pull request!

src/features/left-sidebar/LeftSidebar.js
```diff
@@ -78,6 +80,8 @@ type MenuItem = {
     menuItems?: Array<SubMenuItem>,
     /** Localized text string to use for individual menu items */
     message: string,
+    /** Optional left side bar link renderer. Defaults to defaultNavLinkRenderer */
+    navLinkRenderer?: Function,
```
priyajeet: can you change Function to actual function signature () => React.Node or whatever it might be, here and above

src/features/left-sidebar/defaultNavLinkRenderer.js
```diff
@@ -0,0 +1,62 @@
...
+    newItemBadge?: React.Element<any> | null,
+    onClickRemove?: Function,
+    removeButtonHtmlAttributes?: Object,
+    routerLink?: React.ComponentType<any>,
+    routerProps?: Object,
+    selected?: boolean,
+    showTooltip?: boolean,
+};
+
+function defaultNavLinkRenderer(props: Props): React.Node {
```
priyajeet: Probably just add a simple snapshot unit test for this

src/features/left-sidebar/defaultNavLinkRenderer.js
```diff
@@ -0,0 +1,62 @@
...
+import * as React from 'react';
+import LeftSidebarLink from './LeftSidebarLink';
+
+import type { Callout } from './Callout';
+
+type Props = {
+    callout?: Callout,
+    className?: string,
+    customTheme?: Object,
+    htmlAttributes?: Object,
```
priyajeet: Would be nice if we can change these Objects to actual { something } shape

viodragon2: How do you suggest declaring html attributes as type? Not sure if you meant to list out everything in https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes.

PR #1511, opened 2019-08-13
 fix(content-uploader): Disable cancel button when 100% is reached

mxiao6:


boxcla:

src/elements/content-uploader/ContentUploader.js
```diff
@@ -1049,6 +1049,7 @@ class ContentUploader extends Component<Props, State> {
 
         const hasFiles = items.length !== 0;
         const isLoading = items.some(item => item.status === STATUS_IN_PROGRESS);
+        const allDone = items.every(item => item.progress === 100);
```
mickr: Maybe should be `isDone`?

src/elements/content-uploader/__tests__/__snapshots__/ItemAction-test.js.snap
```diff
@@ -18,6 +18,7 @@ exports[`elements/content-uploader/ItemAction should render correctly with STATU
     theme="default"
   >
     <PlainButton
+      isDisabled={false}
```
mickr: You shouldn't need to specify false

src/elements/content-uploader/ContentUploader.js
jstoffan: It seems like we should be using `item.status` rather than checking the progress percentage. Can we use `!isLoading` or do we need to introduce a new status enum value?

src/elements/content-uploader/__tests__/__snapshots__/ItemAction-test.js.snap
mxiao6: This is the auto-generated snapshot. Should I edit it?

mickr: Ah, never mind!

src/elements/content-uploader/ContentUploader.js
mxiao6: The problem here is that, when the progress is 100, the status remains in IN_PROGRESS for a few seconds, which in this time user can click 'cancel' button but the file has been already uploaded.

mxiao6: > Maybe should be `isDone`?

Sure

jstoffan: I may be missing something. It sounds like we'd definitely want to use `!isLoading` if the progress property can stay at 100 even while it's not finished. Let's sync up tomorrow.

src/elements/content-uploader/ContentUploader.js
```diff
@@ -893,7 +894,11 @@ class ContentUploader extends Component<Props, State> {
         }
 
         item.progress = Math.min(Math.round((event.loaded / event.total) * 100), 100);
-        item.status = STATUS_IN_PROGRESS;
+        if (item.progress === 100) {
+            item.status = STATUS_STAGED;
```
jstoffan: It looks like this can be simplified to a ternary. Should this be status be added to the guard check on line 892?

src/elements/content-uploader/__tests__/ItemAction-test.js
```diff
@@ -31,6 +31,14 @@ describe('elements/content-uploader/ItemAction', () => {
         expect(wrapper).toMatchSnapshot();
     });
 
+    test('should render correctly with STATUS_STAGED', () => {
+        const wrapper = getWrapper({
+            status: STATUS_STAGED,
+        });
+
+        expect(wrapper).toMatchSnapshot();
```
jstoffan: Ideally, unit tests should have explicit assertions in addition to (or in place of) snapshots.

src/elements/content-uploader/ContentUploader.js
```diff
@@ -1049,7 +1055,7 @@ class ContentUploader extends Component<Props, State> {
 
         const hasFiles = items.length !== 0;
         const isLoading = items.some(item => item.status === STATUS_IN_PROGRESS);
-        const isDone = items.every(item => item.progress === 100);
+        const isDone = items.every(item => item.status === STATUS_COMPLETE || item.status === STATUS_STAGED);
```
jstoffan: Tests needed?

src/elements/content-uploader/ContentUploader.js
mxiao6: you're right, thanks!

PR #1513, opened 2019-08-13
fix(tasks): add missing resin component

realmorganw:


boxcla:
Hi @realmorganw, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

realmorganw:
CLA Signed

src/elements/content-sidebar/activity-feed/task-new/Task.js
```diff
@@ -288,6 +288,7 @@ class Task extends React.Component<Props, State> {
                                     {permissions.can_update && (
                                         <MenuItem
                                             data-resin-target={ACTIVITY_TARGETS.INLINE_EDIT}
+                                            data-resin-component="tasks"
```
jstoffan: Does resin work for menu items? I recall the consuming code looking for specific element types, like anchors and buttons. If it does work, can this property live on `Media.Menu` so it's not duplicated and will propagate to any actions we add going forward?

realmorganw: @booc0mtaco @alexkrolick what do you think?

alexkrolick: The reason we are adding it here is because adding the attribute onto the Media.Menu or even the wrapped Tether component would pass the prop onto an element that isn't inside the Portal. Resin's event bubbling follows the DOM, not React's synthetic event propagation.

jstoffan: Ah, gotcha. I implemented something similar for version history actions, but I'm using `Menu` directly there. I also double-checked the resin library and it supports selecting components by role as well as tag, so you should be good to go.

jstoffan:
It looks like the build is failing due to outdated unit test snapshots.

alexkrolick:
Since we have to touch the snapshots anyways, I'll add that extra prop and fix the tracking for comments, too.

src/elements/content-sidebar/activity-feed/common/delete-confirmation/DeleteConfirmation.js
```diff
@@ -50,15 +51,26 @@ class DeleteConfirmation extends React.Component<Props> {
...
+                        data-resin-target={ACTIVITY_TARGETS.INLINE_DELETE_CANCEL}
+                    >
                         <FormattedMessage {...commonMessages.cancel} />
                     </Button>
-                    <PrimaryButton className="bcs-DeleteConfirmation-delete" onClick={onDeleteConfirm} type="button">
+                    <PrimaryButton
+                        className="bcs-DeleteConfirmation-delete"
+                        onClick={onDeleteConfirm}
+                        type="button"
+                        data-resin-target={ACTIVITY_TARGETS.INLINE_DELETE_CONFIRM}
```
alexkrolick: Since this is shared across tasks and comments, events can be distinguished by adding the `data-resin-component` prop to the component (it gets passed through the Overlay and FocusTrap).

src/elements/content-sidebar/activity-feed/common/delete-confirmation/DeleteConfirmation.js
```diff
@@ -50,15 +51,26 @@ class DeleteConfirmation extends React.Component<Props> {
                 role="dialog"
                 shouldDefaultFocus
                 shouldOutlineFocus={false}
+                {...rest}
```
alexkrolick: Since this is shared across tasks and comments, `data-resin-component` can be added here to distinguish events.

src/elements/common/interactionTargets.js
```diff
@@ -28,8 +28,8 @@ export const DETAILS_TARGETS = {
 };
 
 export const ACTIVITY_TARGETS = {
-    INLINE_EDIT: 'activityfeed-inlineedit',
-    INLINE_DELETE: 'activityfeed-inlinedelete',
+    INLINE_DELETE_CONFIRM: 'activityfeed-delete-confirm',
```
jstoffan: Will this cause any issues with existing event records or were they broken previously due to the missing component attribute?

src/elements/common/interactionTargets.js
```diff
@@ -40,6 +40,12 @@ export const ACTIVITY_TARGETS = {
     TASK_APPROVE: 'activityfeed-taskapprove',
     TASK_COMPLETE: 'activityfeed-taskcomplete',
     TASK_REJECT: 'activityfeed-taskreject',
+    TASK_OPTIONS: 'activityfeedtaskoptions', // component
+    TASK_OPTIONS_DELETE: 'activityfeed-taskoptions-delete',
+    TASK_OPTIONS_EDIT: 'activityfeed-taskoptions-edit',
+    COMMENT_OPTIONS: 'activityfeedcommentoptions', // component
```
jstoffan: Is there a typo? `activityfeed-commentoptions`? Should we also pull it out into something like `ACTIVITY_COMPONENTS`?

src/elements/content-sidebar/activity-feed/task-new/Task.js
```diff
@@ -297,7 +303,8 @@ class Task extends React.Component<Props, State> {
                                     )}
                                     {permissions.can_delete && (
                                         <MenuItem
-                                            data-resin-target={ACTIVITY_TARGETS.INLINE_DELETE}
+                                            data-resin-target={ACTIVITY_TARGETS.TASK_OPTIONS_DELETE}
+                                            data-resin-component="tasks"
```
jstoffan: Still needed?

src/elements/common/interactionTargets.js
```diff
@@ -40,6 +40,12 @@ export const ACTIVITY_TARGETS = {
     TASK_APPROVE: 'activityfeed-taskapprove',
     TASK_COMPLETE: 'activityfeed-taskcomplete',
     TASK_REJECT: 'activityfeed-taskreject',
+    TASK_OPTIONS: 'activityfeedtaskoptions', // component
```
jstoffan: These all get built down into strings following the format (usually) of `page-component-feature-target`, so there may end up being a lot of duplicated info in these strings, such as `files-activityfeedtaskoptions--activityfeed-taskoptions-delete`

src/elements/common/interactionTargets.js
alexkrolick: The old variables were not referenced anywhere and the new ones were not reporting in

src/elements/common/interactionTargets.js
alexkrolick: not allowed to have `-` in `component`

src/elements/content-sidebar/activity-feed/task-new/Task.js
alexkrolick: (no, fixed in push)

PR #1514, opened 2019-08-13
fix(bdl): Add SCSS check for color variable name

booc0mtaco:
add additional parsing dollar-sign check to make sure partial color names aren't found and replaced.

### DIFF
--- a/scripts/utils/bdlColorManager
+++ b/scripts/utils/bdlColorManager
@@ -96,8 +96,11 @@ async function main() {
     const fileType = fileName.split('.').pop();
     const verboseMode = process.argv[4] === '--verbose';
 
-    const converter = fileType === 'js' ? camelCase : identity;
-    const reBadName = new RegExp(Object.keys(conversionMap).map(entry => converter(entry)).join('|'), 'g');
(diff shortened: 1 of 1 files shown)


boxcla:

scripts/utils/bdlColorManager
```diff
@@ -96,8 +96,11 @@ async function main() {
     const fileType = fileName.split('.').pop();
     const verboseMode = process.argv[4] === '--verbose';
 
-    const converter = fileType === 'js' ? camelCase : identity;
-    const reBadName = new RegExp(Object.keys(conversionMap).map(entry => converter(entry)).join('|'), 'g');
+    const toRegExp = fileType === 'js' ? identity : entity => `\\\$${entity}(?!-)`;
```
priyajeet: fileType === 'js' repeated 4 times :) but since this is non prod stuff its fine haha

booc0mtaco: I'll probably go in here 1-2 more times to tighten this up, and find a way to test it too. still need to do something about the `toRegExp(caseConverter())` combo too :) ideally, anyone needing to move off old values has easy paths forward, incl. using this if needed

PR #1515, opened 2019-08-14
 fix(content-explorer): ContentExplorer and ContentPicker can handle sharing Weblinks now

mxiao6:


boxcla:

src/api/WebLink.js
```diff
@@ -28,6 +28,44 @@ class WebLink extends Item {
...
+        }
+
+        const cacheKey = this.getCacheKey(id);
+        const params = { fields: fields.toString() };
+
+        try {
+            const { data } = await this.xhr.get({
+                url: this.getUrl(id),
+                params,
+                headers: { 'X-Rep-Hints': X_REP_HINTS },
```
jstoffan: Are the rep hints needed for web links?

src/api/WebLink.js
```diff
@@ -28,6 +28,44 @@ class WebLink extends Item {
...
+     *
+     * @param {string} id - Weblink id
+     * @param {Function} successCallback - Function to call with results
+     * @param {Function} errorCallback - Function to call with errors
+     * @param {Array<String>} fields - Array of field strings
+     * @returns {Promise}
+     */
+    async getWeblink(
+        id: string,
+        successCallback: Function,
```
jstoffan: Can we use stronger types for the success and error callbacks?

src/api/WebLink.js
```diff
@@ -28,6 +28,44 @@ class WebLink extends Item {
         const suffix: string = id ? `/${id}` : '';
         return `${this.getBaseApiUrl()}/web_links${suffix}`;
     }
+
+    /**
+     * Gets a box weblink.
```
jstoffan: Capital `B` in Box.

src/api/WebLink.js
```diff
@@ -28,6 +28,44 @@ class WebLink extends Item {
...
+        successCallback: Function,
+        errorCallback: Function,
+        { fields }: FetchOptions = {},
+    ): Promise<void> {
+        if (this.isDestroyed()) {
+            return;
+        }
+
+        const cacheKey = this.getCacheKey(id);
+        const params = { fields: fields.toString() };
```
jstoffan: What happens if `fields` is undefined?

src/api/WebLink.js
```diff
@@ -28,6 +28,44 @@ class WebLink extends Item {
...
+
+        try {
+            const { data } = await this.xhr.get({
+                url: this.getUrl(id),
+                params,
+                headers: { 'X-Rep-Hints': X_REP_HINTS },
+            });
+
+            const cachedEntry = this.getCache().get(cacheKey);
+            const updatedCacheEntry = { ...cachedEntry, ...data };
```
jstoffan: Should this do a deep merge rather than a shallow assignment?

src/api/WebLink.js
```diff
@@ -28,6 +28,44 @@ class WebLink extends Item {
...
+    ): Promise<void> {
+        if (this.isDestroyed()) {
+            return;
+        }
+
+        const cacheKey = this.getCacheKey(id);
+        const params = { fields: fields.toString() };
+
+        try {
+            const { data } = await this.xhr.get({
```
mickr: Should this check the cache before initiating the api call?

src/api/WebLink.js
mxiao6: not needed... removed

src/api/WebLink.js
mxiao6: sure

src/api/WebLink.js
mxiao6: right

src/api/WebLink.js
mxiao6: right

src/api/WebLink.js
mxiao6: checked

mxiao6:
> Please make sure to add unit tests.

Sorry, will add.

src/api/WebLink.js
```diff
@@ -28,6 +29,70 @@ class WebLink extends Item {
...
+     * @param {Array<String>} fields - Array of field strings
+     * @returns {Promise}
+     */
+    async getWeblink(
+        id: string,
+        successCallback: (newItem: BoxItem) => void,
+        errorCallback: Function,
+        { fields }: FetchOptions = {},
+    ): Promise<void> {
+        if (this.isDestroyed()) {
```
jstoffan: Is this check needed? I don't think it's possible for the API to be destroyed as soon as it's invoked.

src/api/WebLink.js
```diff
@@ -28,6 +29,70 @@ class WebLink extends Item {
...
+
+        try {
+            const { data } = await this.xhr.get(xhrOptions);
+
+            if (this.isDestroyed()) {
+                return;
+            }
+
+            // Cache check is again done since this code is executed async
+            if (cache.has(key)) {
```
jstoffan: This code looks familiar. I'm wondering if we can update `cache.merge` to support unset keys to avoid this check. @priyajeet, do you have a preference?

src/api/WebLink.js
```diff
@@ -28,6 +29,70 @@ class WebLink extends Item {
...
+            return;
+        }
+
+        const cache: APICache = this.getCache();
+        const key: string = this.getCacheKey(id);
+        this.errorCode = ERROR_CODE_FETCH_WEBLINK;
+        this.successCallback = successCallback;
+        this.errorCallback = errorCallback;
+
+        if (cache.has(key)) {
```
jstoffan: I don't think this check is needed, as the `cache.get` return value is also checked in `findMissingProperties`.

src/api/WebLink.js
```diff
@@ -28,6 +29,70 @@ class WebLink extends Item {
...
+            // Cache check is again done since this code is executed async
+            if (cache.has(key)) {
+                cache.merge(key, data);
+            } else {
+                // If there was nothing in the cache
+                cache.set(key, data);
+            }
+
+            this.successHandler(cache.get(key));
+        } catch (e) {
```
jstoffan: Try to avoid abbreviations when possible, as well.

src/api/WebLink.js
mxiao6: Copied from File.js, trying to be consistent 😂

src/api/WebLink.js
mxiao6: It is needed, because if it's `undefined`, `findMissingProperties` will return `[]` which length is `0`, then `successCallback` will execute.

src/api/WebLink.js
mxiao6: Copied from File.js  😂

PR #1516, opened 2019-08-14
feat(select-field): Allow rendering custom options

jpfranco:
This update will allow passing a `optionRenderer` function to `BaseSelectField`, which can be used to render custom select options. 

This PR also contains the following fixes:

- Avoid scrolling options into view when hovering with the mouse (will still do this for keyboard actions).
- Clicking on the scrollbar (when overlay is set to `overflow: auto` and has max height) no longer closes the dropdown.
- Changed `scrollIntoView` mode in order to make the transition less jarring.

### DIFF
--- a/src/components/select-field/BaseSelectField.js
+++ b/src/components/select-field/BaseSelectField.js
@@ -43,13 +43,15 @@ type Props = {
     onChange: Function,
     /** Function will be called with the user selected option (even on deselect or when the option was previously selected) */
     onOptionSelect?: Function,
+    /** Function that allows custom rendering of select field options. When not provided the component will only render the option's displayText by default */
+    optionRenderer?: (option: SelectOptionProp) => React.Node,
     /** List of options (displayText, value) */
     options: Array<SelectOptionProp>,
     /** The select button text shown when no options are selected. */
     placeholder?: string | React.Element<any>,
     /** The currently selected option values (can be empty) */
     selectedValues: Array<SelectOptionValueProp>,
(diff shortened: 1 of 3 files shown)


boxcla:

src/components/select-field/BaseSelectField.js
```diff
@@ -321,12 +330,8 @@ class BaseSelectField extends React.Component<Props, State> {
 
                     this.selectOption(index);
                 },
-                /* preventDefault on mousedown so blur doesn't happen before click */
-                onMouseDown: event => {
```
jpfranco: This was causing the issue where clicking on the scrollbar (when overlay has max height) would close the dropdown. Setting the mouse down event on the overlay itself achieves the same result without causing the issue

src/components/select-field/BaseSelectField.js
```diff
@@ -80,20 +83,26 @@ class BaseSelectField extends React.Component<Props, State> {
...
 
     setActiveItemID = (id: ?string) => {
+        const { shouldScrollIntoView } = this.state;
         const itemEl = id ? document.getElementById(id) : null;
-        this.setState({ activeItemID: id });
-        scrollIntoView(itemEl);
+
+        this.setState({ activeItemID: id, shouldScrollIntoView: false });
+
+        if (shouldScrollIntoView) {
```
priyajeet: Should this be part of the setState callback?

jpfranco: Yeah, I think so. Will update this

PR #1517, opened 2019-08-14
fix(bdl): adjust font weights to match normal and bold lato

booc0mtaco:


boxcla:

PR #1520, opened 2019-08-14
feat(classification): change icon for classify/classification action

feixuanli:
Screen Shot 2019-08-14 at 11 36 08 AM

### DIFF
--- a/src/features/classification/AddClassificationBadge.js
+++ b/src/features/classification/AddClassificationBadge.js
@@ -2,14 +2,14 @@
 import * as React from 'react';
 import { FormattedMessage } from 'react-intl';
 
-import IconAddTags from '../../icons/general/IconAddTags';
+import IconSecurityClassification from '../../icons/general/IconSecurityClassification';
 import { bdlGray62 } from '../../styles/variables';
 import messages from './messages';
 import './AddClassificationBadge.scss';
 
 const AddClassificationBadge = () => (
     <h1 className="bdl-AddClassificationBadge">
-        <IconAddTags color={bdlGray62} height={10} width={10} strokeWidth={3} />
+        <IconSecurityClassification color={bdlGray62} height={10} width={10} />
         <FormattedMessage {...messages.addClassification}>
             {value => <span className="bdl-AddClassificationBadge-name">{value}</span>}
         </FormattedMessage>
--- a/src/features/classification/ClassifiedBadge.js
+++ b/src/features/classification/ClassifiedBadge.js
@@ -2,7 +2,7 @@
 import * as React from 'react';
 
 import Tooltip from '../../components/tooltip';
-import IconAddTags from '../../icons/general/IconAddTags';
+import IconSecurityClassification from '../../icons/general/IconSecurityClassification';
 import { bdlYellorange } from '../../styles/variables';
 import type { Position } from '../../components/tooltip';
 import './ClassifiedBadge.scss';
@@ -16,7 +16,7 @@ type Props = {
 const ClassifiedBadge = ({ name, tooltipPosition = 'bottom-center', tooltipText }: Props) => (
     <Tooltip isDisabled={!tooltipText} position={tooltipPosition} text={tooltipText}>
         <h1 className="bdl-ClassifiedBadge">
-            <IconAddTags color={bdlYellorange} height={10} width={10} strokeWidth={3} />
+            <IconSecurityClassification color={bdlYellorange} height={10} width={10} />
             <span className="bdl-ClassifiedBadge-name">{name}</span>
         </h1>
     </Tooltip>
--- a/src/icons/general/IconSecurityClassification.js
+++ b/src/icons/general/IconSecurityClassification.js
@@ -1,0 +1,48 @@
+// @flow
+import * as React from 'react';
+import classNames from 'classnames';
+
+import AccessibleSVG from '../accessible-svg';
+import { bdlGray } from '../../styles/variables';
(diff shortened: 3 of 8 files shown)


boxcla:

src/features/classification/AddClassificationBadge.js
```diff
@@ -2,14 +2,14 @@
...
-import IconAddTags from '../../icons/general/IconAddTags';
+import IconSecurityClassification from '../../icons/general/IconSecurityClassification';
 import { bdlGray62 } from '../../styles/variables';
 import messages from './messages';
 import './AddClassificationBadge.scss';
 
 const AddClassificationBadge = () => (
     <h1 className="bdl-AddClassificationBadge">
-        <IconAddTags color={bdlGray62} height={10} width={10} strokeWidth={3} />
+        <IconSecurityClassification color={bdlGray62} height={13} width={11} strokeWidth={1} />
```
sahiga: The values for `height`, `width`, and `strokeWidth` match the default values, so you can remove the attributes here.

src/features/classification/ClassifiedBadge.js
```diff
@@ -16,7 +16,7 @@ type Props = {
 const ClassifiedBadge = ({ name, tooltipPosition = 'bottom-center', tooltipText }: Props) => (
     <Tooltip isDisabled={!tooltipText} position={tooltipPosition} text={tooltipText}>
         <h1 className="bdl-ClassifiedBadge">
-            <IconAddTags color={bdlYellorange} height={10} width={10} strokeWidth={3} />
+            <IconSecurityClassification color={bdlYellorange} height={13} width={11} strokeWidth={1} />
```
sahiga: Same comment about removing `height`, `width`, and `strokeWidth`

src/icons/general/__tests__/IconSecurityClassification-test.js
```diff
@@ -0,0 +1,25 @@
...
+        expect(wrapper).toMatchSnapshot();
+    });
+
+    test('should correctly render icon with all props specified', () => {
+        const wrapper = getWrapper({
+            className: 'test',
+            color: '#987654',
+            height: 100,
+            title: 'title',
+            width: 200,
```
sahiga: `strokeWidth` should also be set, since this test checks for all props on the component.

feixuanli:
updated

src/icons/general/IconSecurityClassification.js
```diff
@@ -0,0 +1,44 @@
...
+    color = bdlGray,
+    title,
+    strokeWidth = 1,
+    width = 11,
+}: Props) => (
+    <AccessibleSVG
+        className={`bdl-IconSecurityClassification ${className}`}
+        height={height}
+        title={title}
+        viewBox="0 0 11 13"
```
booc0mtaco: we should use an icon from design that has been exported at the appropriate viewbox size (32/32 or 16/16). With that, we can set the default props for width and height to be the default size (16x16 for badges, or 32x32 for our regular icons).

src/icons/general/IconSecurityClassification.js
```diff
@@ -0,0 +1,44 @@
...
+const IconSecurityClassification = ({
+    className = '',
+    height = 13,
+    color = bdlGray,
+    title,
+    strokeWidth = 1,
+    width = 11,
+}: Props) => (
+    <AccessibleSVG
+        className={`bdl-IconSecurityClassification ${className}`}
```
sahiga: I recommend using `classnames` instead, to prevent the addition of extra whitespace.

feixuanli: Hey @sahiga , what extra whitespace are you referring here ?

src/icons/general/IconSecurityClassification.js
feixuanli: Thanks @booc0mtaco 
The icon is updated, I leave the size as 10*10 since 16 looks a bit off here. 

Screen Shot 2019-08-14 at 5 57 40 PM

src/icons/general/IconSecurityClassification.js
sahiga: @feixuanli In the event that this component doesn't have a `className` prop, the computed class name on the SVG will be "bdl-IconSecurityClassfication " instead of "bdl-IconSecurityClassification." Other components in BUIE use `classnames` to avoid this issue. This is pretty minor, though.

feixuanli: Thanks, updated.

src/icons/general/IconSecurityClassification.js
```diff
@@ -0,0 +1,41 @@
...
+    color = bdlGray,
+    title,
+    strokeWidth = 2,
+    width = 32,
+}: Props) => {
+    const classes = classNames('bdl-IconSecurityClassification', className);
+
+    return (
+        <AccessibleSVG className={classes} height={height} title={title} viewBox="0 0 32 32" width={width}>
+            <g strokeWidth={strokeWidth} fill="none" fillRule="evenodd">
```
priyajeet: You may be able to move the strokeWidth to the 2 paths below and get rid of the `<g>` wrapper completely

feixuanli: `fill="none"` would also be needed by the 2 paths, would we avoid the `<g>` in the case ?

priyajeet: sure, why not, less DOM

PR #1544, opened 2019-08-15
feat(modal): prop to override X dismiss icon with text

ittaishay:
### Existing default close icon
Screen Shot 2019-08-15 at 11 11 12 AM

### With button text override
Screen Shot 2019-08-15 at 11 11 06 AM

### DIFF
--- a/src/components/modal/Modal.js
+++ b/src/components/modal/Modal.js
@@ -15,6 +15,8 @@ type Props = {
     children: React.Node,
     /** Additional CSS classname of the `.modal` element */
     className?: string,
+    /** If provided, will render a button with this text instead of the 'X' to dismiss the modal (onRequestClose must also be provided) */
+    closeButtonText?: string,
     focusElementSelector?: string,
     isLoading?: boolean,
     isOpen?: boolean,
@@ -130,7 +132,16 @@ class Modal extends React.Component<Props> {
     };
 
     render() {
-        const { className, isLoading, isOpen, onRequestClose, shouldNotUsePortal, style, ...rest } = this.props;
+        const {
+            className,
+            closeButtonText,
+            isLoading,
+            isOpen,
+            onRequestClose,
+            shouldNotUsePortal,
+            style,
+            ...rest
+        } = this.props;
 
         if (!isOpen) {
             return null;
@@ -156,6 +167,7 @@ class Modal extends React.Component<Props> {
                         <LoadingIndicator size="large" />
                     ) : (
                         <ModalDialog
+                            closeButtonText={closeButtonText}
                             modalRef={modalEl => {
                                 // This callback gets passed through as a regular prop since
                                 // ModalDialog is wrapped in a HOC
--- a/src/components/modal/Modal.scss
+++ b/src/components/modal/Modal.scss
@@ -93,6 +93,12 @@
     top: 20px;
 }
 
+.bdl-ModalCloseButton--with-custom-text {
+    position: absolute;
+    right: 20px;
+    top: 20px;
+}
+
 .modal-backdrop {
     background: fade-out($black, .25);
     bottom: 0;
--- a/src/components/modal/ModalDialog.js
+++ b/src/components/modal/ModalDialog.js
@@ -5,6 +5,7 @@ import omit from 'lodash/omit';
 import uniqueId from 'lodash/uniqueId';
 import { defineMessages, injectIntl } from 'react-intl';
 
+import Button from '../button';
 import IconClose from '../../icons/general/IconClose';
 
 const ALERT_TYPE = 'alert';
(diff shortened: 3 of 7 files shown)


boxcla:

src/components/modal/Modal.js
```diff
@@ -15,6 +15,8 @@ type Props = {
     children: React.Node,
     /** Additional CSS classname of the `.modal` element */
     className?: string,
+    /** If provided, will render a button with this text instead of the 'X' to dismiss the modal (onRequestClose must also be provided) */
+    closeButtonText?: string,
```
DanDeMicco: This seems fine, but at minimum i would change to React.Node instead of string.

Can you foresee something different being there besides an x or a white button with a label? Instead of passing in a string, what about passing in a render prop? e.g. `renderCloseButton: (closeButtonProps) => React.Node`

src/components/modal/Modal.scss
```diff
@@ -93,6 +93,12 @@
     top: 20px;
 }
 
+.bdl-ModalCloseButton--with-custom-text {
```
priyajeet: withCustomText

src/components/modal/Modal.js
priyajeet: +2

src/components/modal/ModalDialog.js
```diff
@@ -58,20 +60,25 @@ class ModalDialog extends React.Component<Props> {
      * @return {ReactElement|null} - Returns the button, or null if the button shouldn't be rendered
      */
     renderCloseButton() {
-        const { closeButtonProps, onRequestClose, intl } = this.props;
+        const { closeButtonProps, closeButtonText, onRequestClose, intl } = this.props;
```
priyajeet: can closeButtonText be part of closeButtonProps?

src/components/modal/ModalDialog.js
```diff
@@ -58,20 +60,25 @@ class ModalDialog extends React.Component<Props> {
...
+            </Button>
+        ) : (
             // eslint-disable-next-line react/button-has-type
-            <button
-                {...closeButtonProps}
-                aria-label={formatMessage(messages.closeModalText)}
-                className="modal-close-button"
-                onClick={this.onCloseButtonClick}
-            >
+            <button className="modal-close-button" {...sharedProps}>
```
priyajeet: Can you change the icon color below to match the closest new bdl color also?

src/components/modal/ModalDialog.js
```diff
@@ -58,20 +60,25 @@ class ModalDialog extends React.Component<Props> {
...
 
-        return (
+        const sharedProps = {
+            ...closeButtonProps,
+            'aria-label': formatMessage(messages.closeModalText),
+            onClick: this.onCloseButtonClick,
+        };
+
+        return closeButtonText ? (
+            <Button className="bdl-ModalCloseButton--with-custom-text" {...sharedProps}>
```
priyajeet: Modifiers should ideally be with their parent

"bdl-ModalCloseButton  bdl-ModalCloseButton--withCustomText"

And in css it should be

.bdl-ModalCloseButton {
    &.bdl-ModalCloseButton--withCustomText { ... }
}

priyajeet: also missing type="button"

src/components/modal/ModalDialog.js
```diff
@@ -58,20 +60,25 @@ class ModalDialog extends React.Component<Props> {
...
+        const { closeButtonProps, closeButtonText, onRequestClose, intl } = this.props;
         const { formatMessage } = intl;
         if (!onRequestClose) {
             return null;
         }
 
-        return (
+        const sharedProps = {
+            ...closeButtonProps,
+            'aria-label': formatMessage(messages.closeModalText),
```
priyajeet: not sure if aria label is needed since button has text visible and screen reader should read it fine

priyajeet: I guess you are using it for the X version, nvm

src/components/modal/ModalDialog.js
```diff
@@ -58,20 +60,25 @@ class ModalDialog extends React.Component<Props> {
...
+            onClick: this.onCloseButtonClick,
+        };
+
+        return closeButtonText ? (
+            <Button className="bdl-ModalCloseButton--with-custom-text" {...sharedProps}>
+                {closeButtonText}
+            </Button>
+        ) : (
```
priyajeet: 1 thing to test, if button text is too long or of modal title is too long or both are too long...overflows. Maybe ellipsis might be needed

ittaishay:
after comments, re-dicussed with design and we're going to change the UI so this change won't be required, thanks for all the feedback @priyajeet!

PR #1545, opened 2019-08-15
fix(left-sidebar): Adjust margins for loading indicator

viodragon2:
When `NavList` is wrapped in `LoadingIndicatorWrapper`, it appends a `LoadingIndicator` as a child which `:last-child` pseudo class rule does not account for.

Added additional rules when loading indicator is present.


boxcla:
Verified that @viodragon2 has signed the CLA. Thanks for the pull request!

PR #1546, opened 2019-08-15
feat(bdl): Add script for SCSS bdl scoping

swfree:
Add script to update bdl namespace to classes in JS and SCSS files. See `scripts/utils/bdlClassnameManager` for usage guide. See https://github.com/swfree/box-ui-elements/pull/2 to take a look at the script's results.

### DIFF
--- a/scripts/utils/bdlClassnameManager
+++ b/scripts/utils/bdlClassnameManager
@@ -1,13 +1,10 @@
 #!/usr/bin/env node
 
-/* eslint-disable */
 const fs = require('fs');
 const colors = require('colors/safe');
 const { promisify } = require('util');
-const camelCase = require('lodash/camelCase');
-const kebabCase = require('lodash/kebabCase');
-const identity = require('lodash/identity');
 const invert = require('lodash/invert');
+const trimStart = require('lodash/trimStart');
 const readFile = promisify(fs.readFile);
 const writeFile = promisify(fs.writeFile);
 
@@ -43,17 +40,28 @@ const conversionMap = {
     // 'tooltip': 'bdl-Tooltip',
 };
 const scssPrefix = '.';
-const jsPrefixes = ['\'', '"', '`'];
+const jsPrefixes = ["'", '"', '`'];
 
-function lengthSorter (a, b) {
+function lengthSorter(a, b) {
     return a.length > b.length ? -1 : 1;
 }
 
+function findAndReplaceSCSS(toReplace, match, replaceWith) {
+    return toReplace.replace(new RegExp(`\\${match}(?!-)`, 'g'), replaceWith);
+}
+
+function findAndReplaceJS(toReplace, match, replaceWith) {
+    return toReplace.replace(new RegExp(`(?<!-)${match}(?!-)`, 'g'), replaceWith);
+}
+
+/* eslint-disable no-console */
 function printHelp() {
     console.log(`Usage: ${process.argv[1].split('/').pop()} operation fileName [--verbose]`);
     console.log('\nSupported operations:');
     console.log('\tcheck\t\tcheck to see if a file contains deprecated values (returns error if it does)');
-    console.log('\tswap\t\treplace known-deprecated variable names in a JS or SCSS file');
+    console.log('\tswap\t\treplace known-deprecated classes in a JS or SCSS file');
+    console.log('\tappend\t\tappend new namespaced clases to known-deprecated classes in a JS or SCSS file');
+    console.log('\textend\t\tadd @extend to SCSS files with new namespaced classes');
 }
 
 function printVerboseOperationMessage(fileName, matches) {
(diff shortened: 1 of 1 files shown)


boxcla:

scripts/utils/bdlClassnameManager
```diff
@@ -0,0 +1,225 @@
...
+                    // Exit with an error code: we should not find bad names
+                    if (verboseMode) {
+                        console.error(colors.red('Bad BDL class name found in', fileName));
+                    }
+                    process.exit(1);
+                }
+
+                break;
+            case 'append': {
+                const matches = [...new Set(fileToParse.match(reBadName))];
```
jstoffan: It looks like there's a bit of duplication in the setup for each mode. Can the shared logic be extracted and reused?

scripts/utils/bdlClassnameManager
```diff
@@ -0,0 +1,225 @@
...
+
+                if (verboseMode) {
+                    printVerboseOperationMessage(fileName, matches);
+                }
+
+                matches
+                    .sort(lengthSorter) // sort prevents root classnames ('btn') from preemptively changing longer classnames ('btn-primary')
+                    .forEach((match) => {
+                        if (fileType === 'scss') {
+                            fileToParse = fileToParse.replace(new RegExp(`\\${match}(?!-)`, 'g'), `${scssPrefix}${currentConversionMap[kebabCase(match)]}`);
```
jstoffan: Is the `kebab-case` necessary here? It seems like the source class name is likely already in that format (`btn-primary`, for example).

scripts/utils/bdlClassnameManager
```diff
@@ -0,0 +1,225 @@
...
+    const extendMode = operation === 'extend'
+    const currentConversionMap = extendMode ? invert(conversionMap) : conversionMap;
+
+    let reBadName;
+    if (fileType === 'scss') {
+        const converter = entry => `\\${scssPrefix}${entry}(?!-)`;
+        reBadName = new RegExp(
+            Object.keys(currentConversionMap)
+                .map(converter)
+                .sort(lengthSorter) // sort order matters here for regex
```
jstoffan: Comments are super helpful at providing context. Why does sort matter here?

scripts/utils/bdlClassnameManager
```diff
@@ -0,0 +1,225 @@
...
+    if (!Object.keys(conversionMap).length) {
+        console.error(colors.red('Missing values in: '), colors.white('conversionMap'));
+        process.exit(1);
+    }
+    const fileType = fileName.split('.').pop();
+    const verboseMode = process.argv[4] === '--verbose';
+    const extendMode = operation === 'extend'
+    const currentConversionMap = extendMode ? invert(conversionMap) : conversionMap;
+
+    let reBadName;
```
jstoffan: We usually try to avoid abbreviations. Maybe something like `sourceNameRegex`?

scripts/utils/bdlClassnameManager
```diff
@@ -0,0 +1,225 @@
...
+    if (!fileName) {
+        console.error(colors.red('Missing parameter:'), colors.white('fileName'));
+        process.exit(1);
+    }
+    if (!Object.keys(conversionMap).length) {
+        console.error(colors.red('Missing values in: '), colors.white('conversionMap'));
+        process.exit(1);
+    }
+    const fileType = fileName.split('.').pop();
+    const verboseMode = process.argv[4] === '--verbose';
```
jstoffan: We usually prefix boolean variables with `is` or `has`, so `isVerbose`, etc.

scripts/utils/bdlClassnameManager
```diff
@@ -0,0 +1,225 @@
...
+        // Restrict to those prefixes to avoid overrwriting variable names
+        const converter = entry => jsPrefixes.map(prefix => `${prefix}${entry}`).join('|');
+        reBadName = new RegExp(Object.keys(currentConversionMap).map(converter).join('|'), 'g');
+    } else {
+        console.error(colors.yellow('Unrecognized file type for this tool. skipping', fileName, '...'));
+        process.exit(0);
+    }
+
+    try {
+        let fileToParse = await readFile(fileName, { encoding: 'utf8' });
```
jstoffan: I had to double-check that this was the file's contents rather than the file itself due to the name.

scripts/utils/bdlClassnameManager
```diff
@@ -0,0 +1,225 @@
...
+
+        switch (operation) {
+            case 'check':
+                // determine if the file is using one of the legacy variable names
+                const foundMatch = reBadName.test(fileToParse);
```
jstoffan: May want to consider pulling the logic for each operation into its own function.

scripts/utils/bdlClassnameManager
```diff
@@ -0,0 +1,225 @@
...
+                break;
+            }
+            case 'extend': {
+                const matches = [...new Set(fileToParse.match(reBadName))];
+
+                if (verboseMode) {
+                    printVerboseOperationMessage(fileName, matches);
+                }
+
+                matches
```
jstoffan: You may want to throw in a few lines of whitespace and some recognizable heading to visually separate the two sections of the stylesheet. It may make it easier to clean up the file later.

scripts/utils/bdlClassnameManager
swfree: Good point, will try to clean that up

scripts/utils/bdlClassnameManager
swfree: True `kebab-case` isn't needed, I'll replace it with lodash's `trimStart`. The issue here is that `match` is `.tooltip` instead of `tooltip`, which is the key in the conversionMap.

scripts/utils/bdlClassnameManager
swfree: I'll make the comment more clear -- the sort order means that the regex will find matches for longer classnames first, before root classname matches. I couldn't find another way to have multiple options for exact matches in a single regex.

For example:
```
// bad
'.btn .btn-primary'.match(/\.btn|\.btn-primary/g)
=> [".btn", ".btn"]

// good
'.btn .btn-primary'.match(/\.btn-primary|\.btn/g)
=> [".btn", ".btn-primary"]
```

PR #1550, opened 2019-08-16
feat(icons): Add CollectionItemLink icon

shahzadaziz:
Screen Shot 2019-08-16 at 1 37 33 PM


boxcla:

src/icons/two-toned/IconCollectionItemLink.js
```diff
@@ -0,0 +1,32 @@
...
+
+const IconCollectionItemLink = ({ className = '', height = 24, title, width = 24 }: TwoTonedIcon) => (
+    <AccessibleSVG
+        className={`bdl-IconCollectionItemLink ${className}`}
+        height={height}
+        title={title}
+        viewBox="0 0 32 32"
+        width={width}
+    >
+        <g fill="none" fillRule="evenodd">
```
priyajeet: <g> wrapper likely unnecessary

src/icons/two-toned/IconCollectionItemLink.js
```diff
@@ -0,0 +1,32 @@
+// @flow
+import * as React from 'react';
+
+import AccessibleSVG from '../accessible-svg';
+import type { TwoTonedIcon } from '../flowTypes';
+import { bdlGray, white } from '../../styles/variables';
+
+const IconCollectionItemLink = ({ className = '', height = 24, title, width = 24 }: TwoTonedIcon) => (
```
priyajeet: I would start defaulting new icons to 32 32 also, @booc0mtaco ?

j3tan: most of the two-toned default to 24 right now

j3tan: but yeah defaulting to 32px seems reasonable seeing as that's the viewbox

shahzadaziz:
@priyajeet addressed your comments.

src/icons/two-toned/IconCollectionItemLink.js
booc0mtaco: i think using default values equal to the viewbox could work. getting into the habit of specifying the width/height where the icons are used would make future refactoring easier, as well

booc0mtaco:
let's get this icons such that it doesn't have embedded padding. that allows us to specify a `border-radius` circle around it, if needed

shahzadaziz:
@booc0mtaco removed icon padding.

PR #1553, opened 2019-08-16
feat(uploads): resume multiple failed uploads

rosepierce:
Adding ability to resume one or multiple files at once through a button in the uploads manager header.

**Note:** Depends on the changes being made [in this PR](https://github.com/box/box-ui-elements/pull/1552) for resuming the upload of a single file.

PR #1555, opened 2019-08-19
 feat(uploads): uploads manager contents for resumable uploads

rosepierce:
Updates to the contents of the uploads manager to show appropriate UI for the resumable uploads MVP. Includes added remove icon present on each item, different icon to show progress or completed upload or resume when failed, and additional badge alert for failed uploads.

**Note:** Depends on the changes being made in [this PR](https://github.com/box/box-ui-elements/pull/1552) for resuming a single file and in [this PR](https://github.com/box/box-ui-elements/pull/1553) for resuming multiple files at once.


boxcla:

src/elements/content-uploader/ItemList.scss
```diff
@@ -5,3 +5,7 @@
     border-bottom: 1px solid $lighter-grey-blue;
     outline: none;
 }
+
+.bcu-item-list-actionremovecell-column {
```
ConradJChan: Another approach to this to make it more generic (in case we add more actions in the future) would be to name it `bcu-item-list-action-column` and do a CSS selector with :last-child like `.bcu-item-list-action-column:last-child { margin-right: 18px }` to make sure it only applies to the last action column

src/elements/content-uploader/__tests__/ItemResume-test.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import noop from 'lodash/noop';
+import { shallow } from 'enzyme';
+import ItemResume from '../ItemResume';
+import { STATUS_PENDING, STATUS_IN_PROGRESS, STATUS_COMPLETE, STATUS_ERROR } from '../../../constants';
+
+describe('elements/content-uploader/ItemResume', () => {
+    const getWrapper = props => shallow(<ItemResume onClick={noop} status={STATUS_PENDING} {...props} />);
+
+    test('should render correctly with STATUS_COMPLETE', () => {
```
jstoffan: Consider using `jest.each` for tests where there's only one or two inputs and outputs. We should also tend toward writing tests with specific expectations rather than snapshots, as changes to the latter tend to be ignored.

src/elements/content-uploader/removeCellRenderer.js
```diff
@@ -0,0 +1,21 @@
...
+
+import React from 'react';
+import ItemRemove from './ItemRemove';
+import { STATUS_ERROR, STATUS_IN_PROGRESS } from '../../constants';
+
+type Props = {
+    rowData: UploadItem,
+};
+
+export default (onClick: Function) => ({ rowData }: Props) => {
```
jstoffan: Can we add more specific types for all new code? We generally try to avoid Function/Object/any when possible.

src/elements/content-uploader/removeCellRenderer.js
```diff
@@ -0,0 +1,21 @@
...
+    const isUploading = rowData.status === STATUS_IN_PROGRESS;
+    const isFailed = rowData.status === STATUS_ERROR;
+    if (!rowData.isFolder) {
+        return <ItemRemove isFailed={isFailed} isUploading={isUploading} onClick={() => onClick(rowData)} />;
+    }
+    return null;
```
jstoffan: This should be moved up to the first line inside a guard clause.

src/elements/content-uploader/ItemResume.js
```diff
@@ -0,0 +1,64 @@
...
+type Props = {
+    isFolder: boolean,
+    onClick: Function,
+    size: number,
+    status: UploadStatus,
+};
+
+const ItemResume = ({ status, onClick, isFolder = false }: Props) => {
+    let icon;
+    const resin = {};
```
jstoffan: The most common convention is to put `const` before `let`.

src/elements/content-uploader/ItemResume.js
```diff
@@ -0,0 +1,64 @@
...
+            break;
+        case STATUS_IN_PROGRESS:
+            icon = <LoadingIndicator />;
+            break;
+        case STATUS_PENDING:
+        default:
+        // empty
+    }
+
+    const shownIcon =
```
jstoffan: This variable doesn't seem necessary. It's generally preferred to keep the JSX declarations together, when possible.

src/elements/content-uploader/ItemRemove.js
```diff
@@ -0,0 +1,45 @@
...
+type Props = {
+    isFailed: boolean,
+    isUploading: boolean,
+    onClick: Function,
+};
+
+const ItemRemove = ({ isFailed, isUploading, onClick }: Props) => {
+    const resin = {};
+    let target = null;
+    const tooltip = <FormattedMessage {...messages.remove} />;
```
jstoffan: Is this variable necessary if the value is only used once?

src/elements/content-uploader/ItemList.scss
```diff
@@ -5,3 +5,7 @@
     border-bottom: 1px solid $lighter-grey-blue;
     outline: none;
 }
+
+.bcu-item-list-action-column:last-child {
```
jstoffan: What is the `:last-child` in this case? Can you attach a screenshot of the new experience?

src/elements/content-uploader/ItemResume.js
```diff
@@ -0,0 +1,64 @@
+/**
+ * @flow
+ * @file Item resume component
+ */
+
+import React from 'react';
+import { FormattedMessage } from 'react-intl';
+import PlainButton from '../../components/plain-button/PlainButton';
```
swfree: Minor suggestion - we often alphabetize imports in these groupings, so you could alphabetize the relative imports if you want:
- third party imports
- relative imports
- scss

src/elements/content-uploader/ItemResume.js
```diff
@@ -0,0 +1,64 @@
...
+import messages from '../common/messages';
+
+type Props = {
+    isFolder: boolean,
+    onClick: Function,
+    size: number,
+    status: UploadStatus,
+};
+
+const ItemResume = ({ status, onClick, isFolder = false }: Props) => {
```
jstoffan: This file looks very similar to `IconAction`. We should avoid duplicating components when possible.

src/elements/content-uploader/ItemList.js
```diff
@@ -45,12 +51,21 @@ const ItemList = ({ items, onClick }: Props) => (
                         width={300}
                     />
                     <Column
-                        cellRenderer={actionCell}
+                        className={isResumableUploadsEnabled ? '' : 'bcu-item-list-action-column'}
+                        cellRenderer={isResumableUploadsEnabled ? resumeCell : actionCell}
```
jstoffan: Can the new behaviors in `resumeCell` instead live inside `actionCell` to avoid duplication?

src/elements/content-uploader/ItemList.scss
rosepierce: With the new UI the `:last-child` would be the remove button that is newly separated from the in-progress/retry/completed icon.
Screen Shot 2019-08-30 at 9 48 14 AM

jstoffan: The cell spacing, content alignment, and icon sizes look a bit off in this screenshot. Each cell, including those in header, should follow the same vertical lines. I think we may get a better result if we make each action cell a certain size, center its icon, and apply a consistent height/width/margin-left.

src/elements/content-uploader/ContentUploader.js
```diff
@@ -1133,6 +1133,7 @@ class ContentUploader extends Component<Props, State> {
                             isVisible={isVisible}
                             items={items}
                             onItemActionClick={this.onClick}
+                            onRemoveActionClick={this.removeFileFromUploadQueue.bind(this)}
```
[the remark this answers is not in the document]
mickr: Should be bound in the constructor or arrow function. Something something prototypes 😃

src/elements/content-uploader/resumeCellRenderer.js
```diff
@@ -0,0 +1,15 @@
...
+
+import React from 'react';
+import ItemResume from './ItemResume';
+
+type Props = {
+    rowData: UploadItem,
+};
+
+export default (onClick: Function) => ({ rowData }: Props) => (
+    <ItemResume {...rowData} onClick={() => onClick(rowData)} />
```
mickr: Minor suggestion: Should pass a reference to a function as to not create a new function on each render.

src/elements/content-uploader/ItemList.scss
rosepierce: Screen Shot 2019-09-03 at 10 28 16 AM
@jstoffan Does this alignment make a bit more sense? I shifted the Resume button in the header over, standardized the size of each icon in the item rows, removed the extra padding to the right of the percentage progress, and slightly shifted the icon in the header for collapsing/expanding the uploads manager to the right to make it more evenly placed with the remove icons below it.

jstoffan: Looks much better, @rosepierce! Thanks for taking the time to polish it up so nicely.

src/elements/content-uploader/ItemRemove.js
```diff
@@ -0,0 +1,44 @@
...
+import { FormattedMessage } from 'react-intl';
+import PlainButton from '../../components/plain-button/PlainButton';
+import Tooltip from '../../components/tooltip';
+import IconClose from '../../icons/general/IconClose';
+import messages from '../common/messages';
+
+type Props = {
+    isFailed: boolean,
+    isUploading: boolean,
+    onClick: Function,
```
jstoffan: Can we include a stronger type?

src/elements/content-uploader/ItemRemove.js
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    if (target) {
+        resin['data-resin-target'] = target;
+    }
+
+    return (
+        <div className="bcu-item-action">
+            <Tooltip position="top-left" text={<FormattedMessage {...messages.remove} />}>
+                <PlainButton onClick={onClick} type="button" {...resin}>
```
jstoffan: I think we recently added logic around disabling this button if the item was in a certain status. Has that been lost?

PR #1556, opened 2019-08-20
fix(versions): Update version actions to avoid unnecessary buttons

jstoffan:
I've confirmed that Resin events are still firing as expected with this change; the buttons were never necessary.


boxcla:

PR #1557, opened 2019-08-20
fix(sidebar): Limit maximum height of additional tabs in sidebar nav

jstoffan:
Currently, previewing a file with a high number of recommended apps with a short (>500px) viewport causes the preview header to overflow off the top of the screen. This change resolves the issue by allowing the recommended apps section to shrink and scroll.

### DIFF
--- a/src/elements/content-sidebar/SidebarNav.js
+++ b/src/elements/content-sidebar/SidebarNav.js
@@ -49,55 +49,62 @@ const SidebarNav = ({
 }: Props) => (
     <div className="bcs-SidebarNav" aria-label={intl.formatMessage(messages.sidebarNavLabel)} role="tablist">
         <div className="bcs-SidebarNav-tabs">
-            {hasActivity && (
-                <SidebarNavButton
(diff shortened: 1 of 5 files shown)


boxcla:

src/elements/content-sidebar/SidebarNav.js
```diff
@@ -49,55 +49,62 @@ const SidebarNav = ({
...
-                    onNavigate={onNavigate}
-                    tooltip={<FormattedMessage {...messages.sidebarMetadataTitle} />}
-                >
-                    <IconMetadataThick />
-                </SidebarNavButton>
+            <div className="bcs-SidebarNav-main">
+                {hasActivity && (
+                    <SidebarNavButton
+                        data-resin-target={SIDEBAR_NAV_TARGETS.ACTIVITY}
+                        data-testid="sidebaractivity"
```
sahiga: Should we also have constants for the `data-testid` attributes?

jstoffan: We haven't been doing that so far, from what I recall. This is also just an indentation change, so it seems unrelated.

PR #1561, opened 2019-08-21
feat(content-sidebar): programmatic refresh for sidebar panels

swfree:


boxcla:

src/elements/content-sidebar/ContentSidebar.js
```diff
@@ -350,8 +357,10 @@ class ContentSidebar extends React.Component<Props, State> {
                             metadataSidebarProps={metadataSidebarProps}
                             onVersionChange={onVersionChange}
                             onVersionHistoryClick={onVersionHistoryClick}
-                            refreshIdentity={refreshIdentity}
                             versionsSidebarProps={versionsSidebarProps}
+                            wrappedComponentRef={ref => {
```
swfree: `<Sidebar />` is wrapped in withRouter HOC. The `wrappedComponentRef` prop is available to pass the ref through to the wrapped component. It expects a function: https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/withRouter.md#wrappedcomponentref-func

src/elements/content-sidebar/SidebarPanels.js
```diff
@@ -68,123 +71,166 @@ const LoadableVersionsSidebar = SidebarUtils.getAsyncSidebarContent(
...
-                    exact
-                    path={`/${SIDEBAR_VIEW_SKILLS}`}
-                    render={() => (
-                        <LoadableSkillsSidebar
-                            key={file.id}
-                            file={file}
-                            getPreview={getPreview}
-                            getViewer={getViewer}
-                            startMarkName={MARK_NAME_JS_LOADING_SKILLS}
+class SidebarPanels extends React.Component<Props> {
```
swfree: Converted to a class in order to use `createRef`

src/elements/content-sidebar/SidebarPanels.js
```diff
@@ -19,9 +19,13 @@ import {
     SIDEBAR_VIEW_SKILLS,
     SIDEBAR_VIEW_VERSIONS,
 } from '../../constants';
+import DetailsSidebar from './DetailsSidebar';
```
jstoffan: You can simplify these a bit: `import DetailsSidebar, { type DetailsSidebarProps } from './DetailsSidebar';`

src/elements/content-preview/ContentPreview.js
```diff
@@ -158,6 +158,8 @@ class ContentPreview extends React.PureComponent<Props, State> {
 
     api: API;
 
+    contentSidebarRef: { current: null | ContentSidebar } = React.createRef();
```
jstoffan: Should this also be just `contentSidebar` rather than `contentSidebarRef`?

src/elements/content-preview/ContentPreview.js
```diff
@@ -62,7 +63,6 @@ type Props = {
     collection: Array<string | BoxItem>,
     contentOpenWithProps: ContentOpenWithProps,
     contentSidebarProps: ContentSidebarProps,
-    contentSidebarRef: React.Ref<any>,
```
jstoffan: We'll definitely want to let our partners team know about this change.

src/elements/wrappers/ContentPreview.js
```diff
@@ -10,6 +10,14 @@ import ES6Wrapper from './ES6Wrapper';
 import ContentPreviewResponsive from '../content-preview';
 
 class ContentPreview extends ES6Wrapper {
+    /**
+     * Helper to programmatically refresh the preview's sidebar panel
+     * @returns {void}
+     */
+    refreshSidebar(): void {
```
jstoffan: Do we also need to expose `refresh` in the ContentSidebar ES6 wrapper?

swfree: That one already exists: https://github.com/box/box-ui-elements/blob/master/src/elements/wrappers/ContentSidebar.js#L17

src/elements/content-preview/ContentPreview.js
swfree: Good catch, I'll also update the sidebarPanelsRef

src/elements/content-preview/ContentPreview.js
swfree: Sure thing! Can you point me in the right direction so I can let them know?

src/elements/content-preview/ContentPreview.js
```diff
@@ -47,6 +47,7 @@ import {
     ERROR_CODE_UNKNOWN,
 } from '../../constants';
 import type { ErrorType } from '../common/flowTypes';
+import ContentSidebar from '../content-sidebar';
```
jstoffan: I think this may actually cause issues with how we're doing code splitting. Can we use a more generic type, instead?

swfree: Ah thanks for pointing this out! Will do 👍

src/elements/content-sidebar/SidebarPanels.js
```diff
@@ -68,124 +71,167 @@ const LoadableVersionsSidebar = SidebarUtils.getAsyncSidebarContent(
...
+            hasSkills,
+            hasVersions,
+            isOpen,
+            metadataSidebarProps,
+            onVersionChange,
+            onVersionHistoryClick,
+            versionsSidebarProps,
+        }: Props = this.props;
+
+        if (isOpen) {
```
jstoffan: Can we make this a guard clause to avoid the level of indentation it introduces?

```
if (!isOpen) {
    return null;
}
```

src/elements/content-sidebar/SidebarPanels.js
```diff
@@ -41,10 +41,13 @@ type Props = {
     metadataSidebarProps: MetadataSidebarProps,
     onVersionChange?: Function,
     onVersionHistoryClick?: Function,
-    refreshIdentity?: boolean,
     versionsSidebarProps: VersionsSidebarProps,
 };
 
+type ElementRefType = {
+    current: null | Object,
```
jstoffan: Can we specify `{ refresh: () => {} }` here, as we did in ContentPreview?

swfree: Unfortunately flow throws an error with that structure, thats why I just defined a type here instead of in the elements flowType file, since the ContentPreview example doesn't share the same type shape. Its likely due to the async loading of these sidebar components: https://github.com/box/box-ui-elements/pull/1561/files#diff-ea8b257187d0274f39a99eef07fc779bR59

PR #1563, opened 2019-08-21
fix(sidebar): add markup and accessibility to tab list

mxiao6:


boxcla:

mxiao6:
> This is open source, so it should not have the ticket reference in the commits or title. Also, can you add a little more detail to the PR description?

sorry, fixed it now.

src/components/plain-button/PlainButton.js
```diff
@@ -22,7 +31,7 @@ const PlainButton = ({ children, className = '', getDOMRef, isDisabled = false,
 
     return (
         // eslint-disable-next-line react/button-has-type
-        <button className={`btn-plain ${className}`} ref={getDOMRef} type={type} {...rest} {...buttonProps}>
+        <button className={`btn-plain ${className}`} ref={getDOMRef} type={type} id={id} {...rest} {...buttonProps}>
```
alexkrolick: `id` would already be passed by the `...rest` here, no need to add it

src/elements/common/nav-button/NavButton.js
```diff
@@ -26,6 +27,7 @@ type Props = {
 
 const NavButton = React.forwardRef<Props, React.Ref<any>>((props: Props, ref: React.Ref<any>) => {
     const {
+        'data-testid': dataTestId,
```
alexkrolick: Here as well, the `..rest` captures all the props that aren't explicitly listed.

src/components/plain-button/PlainButton.js
mxiao6: Fixed.

src/elements/common/nav-button/NavButton.js
mxiao6: Have to explicitly list 'data-testid' here because it's a string, which needs to be converted to a valid variable name.

mxiao6: Never mind, I noticed that I can totally remove it from this layer.

src/elements/content-sidebar/DetailsSidebar.js
```diff
@@ -329,7 +329,12 @@ class DetailsSidebar extends React.PureComponent<Props, State> {
 
         // TODO: Add loading indicator and handle errors once file call is split out
         return (
-            <SidebarContent className="bcs-details" title={SidebarUtils.getTitleForView(SIDEBAR_VIEW_DETAILS)}>
+            <SidebarContent
+                className="bcs-details"
+                title={SidebarUtils.getTitleForView(SIDEBAR_VIEW_DETAILS)}
+                aria-labelledby="sidebardetails"
+                id="sidebardetails-content"
```
jstoffan: We technically support having multiple of the same types of Element on the page at one time. I think we'll need to add some form of unique id generation to the mix here.

src/elements/content-sidebar/Sidebar.js
```diff
@@ -175,6 +182,46 @@ class Sidebar extends React.Component<Props, State> {
         }
     }
 
+    /**
+     * Handle keyboard interaction
+     *
+     * @param {Array<string>} tablist - Contains the pathnames of tabs
+     * @param {SyntheticKeyboardEvent} event - Keyboard event
+     */
+    onKeyDown = (tablist: Array<string>) => (event: SyntheticKeyboardEvent<>): void => {
```
jstoffan: The most common convention uses the prefix `on` for event-oriented prop names and `handle` for their values:

```
onKeyDown={handleKeyDown}
```

src/elements/content-sidebar/Sidebar.js
```diff
@@ -175,6 +182,46 @@ class Sidebar extends React.Component<Props, State> {
...
+        const { history } = this.props;
+        const { nativeEvent } = event;
```
jstoffan: This can be destructured inline in the function signature.

src/elements/content-sidebar/Sidebar.js
```diff
@@ -206,6 +253,20 @@ class Sidebar extends React.Component<Props, State> {
             'bcs-is-open': isOpen,
         });
 
+        const tablist = [];
```
jstoffan: Can we pull this information from React Router, instead? The routes should be registered after the initial render. If not, does this need to be redefined in every render call? Should it have its own method?

src/elements/content-sidebar/SidebarNavButton.js
```diff
@@ -53,6 +54,7 @@ const SidebarNavButton = ({
                                 state: sidebarState,
                             }}
                             type="button"
+                            tabIndex={isActive() ? '0' : '-1'}
```
jstoffan: Can we pull the return value of `isActive()` into its own variable to avoid calling it twice?

src/elements/content-sidebar/__tests__/Sidebar-test.js
```diff
@@ -128,6 +136,91 @@ describe('elements/content-sidebar/Sidebar', () => {
...
 
+    describe('onKeyDown', () => {
+        const tabList = [
+            `/${SIDEBAR_VIEW_ACTIVITY}`,
+            `/${SIDEBAR_VIEW_DETAILS}`,
+            `/${SIDEBAR_VIEW_SKILLS}`,
+            `/${SIDEBAR_VIEW_METADATA}`,
+        ];
+
+        test('should naviagate to previous tab', () => {
```
sahiga: Typo: "navigate"

src/elements/content-sidebar/__tests__/Sidebar-test.js
```diff
@@ -128,6 +136,91 @@ describe('elements/content-sidebar/Sidebar', () => {
...
+                key: KEYS.arrowUp,
+                stopPropagation: jest.fn(),
+                preventDefault: jest.fn(),
+            };
+            onKeyDownWithList(event);
+
+            expect(instance.props.history.location.pathname).toBe(`/${SIDEBAR_VIEW_ACTIVITY}`);
+        });
+
+        test('should naviagate to next tab', () => {
```
sahiga: "navigate"

src/elements/content-sidebar/__tests__/Sidebar-test.js
```diff
@@ -128,6 +136,91 @@ describe('elements/content-sidebar/Sidebar', () => {
...
+                key: KEYS.arrowDown,
+                stopPropagation: jest.fn(),
+                preventDefault: jest.fn(),
+            };
+            onKeyDownWithList(event);
+
+            expect(instance.props.history.location.pathname).toBe(`/${SIDEBAR_VIEW_SKILLS}`);
+        });
+
+        test('should stay in the same tab if key is not up nor down', () => {
```
sahiga: Minor: This should say "**neither** up nor down"

src/elements/content-sidebar/__tests__/Sidebar-test.js
```diff
@@ -128,6 +136,91 @@ describe('elements/content-sidebar/Sidebar', () => {
...
+                    stopImmediatePropagation: jest.fn(),
+                },
+                key: KEYS.arrowLeft,
+            };
+            onKeyDownWithList(event);
+
+            expect(instance.props.history.location.pathname).toBe(`/${SIDEBAR_VIEW_DETAILS}`);
+        });
+
+        test('should stay in the same tab if no more prev/next tab', () => {
```
sahiga: This test appears to be checking for an `arrowUp` key event on the first tab. Should we also check for an `arrowDown` key event on the last tab (`SIDEBAR_VIEW_METADATA`)?

src/elements/content-sidebar/MetadataSidebar.js
```diff
@@ -405,6 +405,8 @@ class MetadataSidebar extends React.PureComponent<Props, State> {
                 }
                 className="bcs-metadata"
                 title={SidebarUtils.getTitleForView(SIDEBAR_VIEW_METADATA)}
+                aria-labelledby="sidebarmetadata"
+                id="sidebarmetadata-content"
```
priyajeet: You will have to use lodash/uniqueid to generate one, or pass it down from ContentSidebar.js which generates an ID and use that in concatenation or something

src/elements/content-sidebar/Sidebar.js
```diff
@@ -175,6 +182,46 @@ class Sidebar extends React.Component<Props, State> {
...
+     * Handle keyboard interaction
+     *
+     * @param {Array<string>} tablist - Contains the pathnames of tabs
+     * @param {SyntheticKeyboardEvent} event - Keyboard event
+     */
+    onKeyDown = (tablist: Array<string>) => (event: SyntheticKeyboardEvent<>): void => {
+        const { history } = this.props;
+        const { nativeEvent } = event;
+
+        nativeEvent.stopImmediatePropagation();
```
priyajeet: Is native stuff needed? Dont see use of addeventlistener

src/elements/content-sidebar/Sidebar.js
```diff
@@ -175,6 +182,46 @@ class Sidebar extends React.Component<Props, State> {
...
+            case KEYS.arrowUp:
+                if (currIndex === 0) {
+                    return;
+                }
+
+                history.replace(tablist[currIndex - 1]);
+
+                event.stopPropagation();
+                event.preventDefault();
+                break;
```
priyajeet: history.replace(tablist[XYZ])
event.stopPropagation();
event.preventDefault();

can probably be made into a function

src/elements/content-sidebar/SidebarNavButton.js
```diff
@@ -38,13 +38,15 @@ const SidebarNavButton = ({
...
-                    <Tooltip position="middle-left" text={tooltip}>
+                    <Tooltip position="middle-left" text={tooltip} isTabbable={false}>
                         <NavButton
                             aria-selected={isActive()}
+                            aria-controls={`${dataTestId}-content`}
                             activeClassName="bcs-is-selected"
                             className="bcs-NavButton"
                             data-resin-target={dataResinTarget}
                             data-testid={dataTestId}
+                            id={dataTestId}
```
priyajeet: Feels a bit weird that we are mixing something that is used for testing with things needed for accessibility.

src/elements/content-sidebar/__tests__/Sidebar-test.js
mxiao6: fixed, thank you

src/elements/content-sidebar/__tests__/Sidebar-test.js
mxiao6: Thanks 😂

src/elements/content-sidebar/__tests__/Sidebar-test.js
mxiao6: Sure

src/elements/content-sidebar/Sidebar.js
mxiao6: not needed...

src/elements/content-sidebar/Sidebar.js
mxiao6: the unit test is not complicated 😂

src/elements/content-sidebar/Sidebar.js
```diff
@@ -76,8 +83,32 @@ class Sidebar extends React.Component<Props, State> {
     constructor(props: Props) {
         super(props);
 
+        const hasActivity = SidebarUtils.canHaveActivitySidebar(props);
```
jstoffan: We have very similar code to register routes for each sidebar in `SidebarPanels`. I think we'll need to rethink our approach to keep the logic for routing in one place.

src/elements/content-sidebar/Sidebar.js
```diff
@@ -76,8 +83,32 @@ class Sidebar extends React.Component<Props, State> {
...
+        if (hasSkills) {
+            tablist.push(`/${SIDEBAR_VIEW_SKILLS}`);
+        }
+        if (hasMetadata) {
+            tablist.push(`/${SIDEBAR_VIEW_METADATA}`);
+        }
+
         this.state = {
             isDirty: this.getLocationState('open') || false,
+            hasActivity,
```
jstoffan: These values are currently calculated in `render` because their values are derived exclusively from props; they shouldn't live in state at all.

src/elements/common/nav-button/NavButton.js
```diff
@@ -48,6 +48,7 @@ const NavButton = React.forwardRef<Props, React.Ref<any>>((props: Props, ref: Re
                 return (
                     <Component
                         className={classNames(className, { [activeClassName]: isActiveValue })}
+                        getDOMRef={ref}
```
jstoffan: This will only work for certain component types, like PlainButton, but not for native elements. Is this necessary with the `{...rest}` below?

src/elements/content-sidebar/SidebarNavTablist.js
```diff
@@ -0,0 +1,79 @@
...
+                history.push(tablist[nextIndex]);
+
+                event.stopPropagation();
+                event.preventDefault();
+                break;
+            default:
+                break;
+        }
+        if (refs.length > 0) {
+            refs[nextIndex].focus();
```
jstoffan: Will this cause the first tab to be focused on all keystrokes, even other than up/down?

src/elements/content-sidebar/SidebarNavTablist.js
```diff
@@ -0,0 +1,79 @@
...
+                if (!tab) {
+                    return null;
+                }
+
+                return React.cloneElement(tab, {
+                    elementId,
+                    isOpen,
+                    onNavigate,
+                    ref: ref => {
+                        refs.push(ref);
```
jstoffan: Do we need to track the `refs` in a separate array or can we access them from `.map` above, similar to how we're getting the paths.

src/elements/content-sidebar/__tests__/SidebarNavTablist-test.js
```diff
@@ -0,0 +1,85 @@
...
+
+        const wrapper = shallow(
+            <SidebarNavTablist>
+                <MockChildren />
+            </SidebarNavTablist>,
+        );
+
+        expect(wrapper.type()).toEqual('div');
+        expect(wrapper.hasClass('bcs-SidebarNav-main')).toBe(true);
+        expect(wrapper.find(MockChildren).length).toEqual(1);
```
jstoffan: This can be simplified to `expect(wrapper.exists(MockChildren)).toBe(true)`.

test/integration/content-sidebar/ContentSidebar.e2e.test.js
```diff
@@ -44,6 +44,26 @@ describe('ContentSidebar', () => {
...
+            cy.getByTestId('bcs-content').should('exist');
+            cy.getByTestId('sidebarskills').should('have.class', 'bcs-is-selected');
+
+            cy.getByTestId('sidebarskills').trigger('keydown', { key: 'ArrowDown' });
+            cy.getByTestId('sidebarskills').should('not.have.class', 'bcs-is-selected');
+            cy.getByTestId('sidebarmetadata').should('have.class', 'bcs-is-selected');
+            cy.focused().should('have.attr', 'data-testid', 'sidebarmetadata');
+        });
+
+        it('should switch to the previous panel when a user presses up arrow ', () => {
```
jstoffan: These can be combined into a single test to save on the overhead of starting a new test case.

src/elements/content-sidebar/__tests__/SidebarNavTablist-test.js
```diff
@@ -0,0 +1,85 @@
...
+            const wrapper = shallow(
+                <SidebarNavTablist history={history}>
+                    {viewList.map(view => {
+                        return <div sidebarView={view} key={view} />;
+                    })}
+                </SidebarNavTablist>,
+            );
+
+            const event = {
+                nativeEvent: {
```
jstoffan: Is this needed?

src/elements/content-sidebar/__tests__/SidebarNavTablist-test.js
```diff
@@ -0,0 +1,85 @@
...
+                    stopImmediatePropagation: jest.fn(),
+                },
+                key: KEYS.arrowUp,
+                stopPropagation: jest.fn(),
```
jstoffan: Should we confirm that these are being called?

src/elements/content-sidebar/__tests__/SidebarNavTablist-test.js
```diff
@@ -0,0 +1,85 @@
...
+        expect(wrapper.hasClass('bcs-SidebarNav-main')).toBe(true);
+        expect(wrapper.find(MockChildren).length).toEqual(1);
+    });
+
+    describe('handleKeyDown', () => {
+        const viewList = [SIDEBAR_VIEW_ACTIVITY, SIDEBAR_VIEW_DETAILS, SIDEBAR_VIEW_SKILLS, SIDEBAR_VIEW_METADATA];
+
+        test('should navigate to previous tab', () => {
+            const history = createMemoryHistory();
+            history.push(`/${SIDEBAR_VIEW_DETAILS}`);
```
jstoffan: You should be able to pass `initialEntries` as an option to `createMemoryHistory` to avoid needing to push here.

src/elements/content-sidebar/SidebarNavTablist.js
mxiao6: genius

src/elements/content-sidebar/SidebarNavTablist.js
mxiao6: no

PR #1564, opened 2019-08-21
feat(tasks): handle new error message with title and description

booc0mtaco:
add support for handling race conditions that can occur when altering a task in a certain way. 

- if a task is being edited to follow the "any" completion rule, any existing approvers/completers could trigger the rule stating that the task is done.
- we need to make sure that we perform the checks in a better sequence to preserve changes to tasks, collaborators, and status in the safest order possible
- re-sequencing the work such that task details are updated before removals
- adding new message for any attempts to remove assignees **after** a task is updated to "any" status

### DIFF
--- a/src/api/Feed.js
+++ b/src/api/Feed.js
@@ -535,7 +535,7 @@ class Feed extends Base {
      * @param {Function} errorCallback - the function which will be called on error
      * @return {void}
      */
-    updateTaskNew = (
+    updateTaskNew = async (
         file: BoxItem,
         task: TaskUpdatePayload,
         successCallback: () => void = noop,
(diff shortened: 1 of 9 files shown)


boxcla:

priyajeet:
Can you edit the PR title to be correct format

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
```diff
@@ -357,7 +384,7 @@ class TaskForm extends React.Component<Props, State> {
                         <PillSelectorDropdown
                             className={pillSelectorOverlayClasses}
                             error={this.getErrorByFieldname('taskAssignees')}
-                            disabled={isLoading}
+                            disabled={isLoading || isForbiddenErrorOnEdit}
```
priyajeet: This logic is super repeated. Should be made a new const above, isDisabled= and used wherever applicable

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
```diff
@@ -332,23 +367,15 @@ class TaskForm extends React.Component<Props, State> {
...
             : messages.tasksEditTaskFormSubmitLabel;
-
-        const taskErrorMessage = isCreateEditMode
-            ? apiMessages.taskCreateErrorMessage
-            : messages.taskUpdateErrorMessage;
-
         const shouldShowCompletionRule = approvers.length > 0;
         const isCompletionRuleCheckboxDisabled = approvers.length <= 1;
         const isCompletionRuleCheckboxChecked = completionRule === TASK_COMPLETION_RULE_ANY;
+        const isForbiddenErrorOnEdit = !!(error && error.status && error.status === 403 && !isCreateEditMode);
```
priyajeet: Unnecessary error.status check

Also can use lodash/get

getProp(error, 'status') === 403 &&

src/api/Feed.js
```diff
@@ -551,33 +551,50 @@ class Feed extends Base {
         this.updateFeedItem({ isPending: true }, task.id);
 
         Promise.all(task.addedAssignees.map(assignee => this.createTaskCollaborator(file, task, assignee)))
+            .then(() => {
```
priyajeet: Should async/await to get our of this nesting

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
```diff
@@ -102,6 +102,41 @@ function convertAssigneesToSelectorItems(approvers: Array<TaskCollabAssignee>):
     });
 }
 
+function getEditErrorMessage(taskType: TaskType, editMode?: TaskEditMode, error?: { status: number }): React.Node {
```
priyajeet: This should ideally be its own component `<TaskError>`

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
```diff
@@ -102,6 +102,41 @@ function convertAssigneesToSelectorItems(approvers: Array<TaskCollabAssignee>):
...
+                <InlineError title={<FormattedMessage {...messages.taskEditWarningTitle} />}>
+                    <FormattedMessage {...messages.taskGeneralAssigneeRemovalWarningMessage} />
+                </InlineError>
+            ) : (
+                <InlineError title={<FormattedMessage {...messages.taskCreateErrorTitle} />}>
+                    <FormattedMessage {...taskErrorMessage} />
+                </InlineError>
+            );
+        case 'APPROVAL':
+            return isForbiddenErrorOnEdit ? (
```
priyajeet: Inline error component is repeated. You can just set the title message and tile body conditionally in the switch/blocks, and use 1 component at the ned

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
booc0mtaco: yah could do. at first this was simply to get the message object, but there were changes to both messages used. can separate this out.

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
booc0mtaco: Personally, i find it easier to reason about why it's disabled when the evaluation has parts like this, but to your point, there are several that need to match, and duplicating these two is a potential source of future bugs

PR #1566, opened 2019-08-22
feat(classification): updated ui for classification

jpfranco:
BREAKING CHANGE: Also changes advisoryMessage prop to definition to keep terminology consistent

UI will now show a "Definition" label on top of the definition text:

Screen Shot 2019-08-22 at 2 30 07 PM


boxcla:

PR #1568, opened 2019-08-23
 fix(versions): Display date and name for most relevant version action

jstoffan:
I'm not attached to the naming of the helper methods, so suggestions are welcome. I wanted to avoid duplicating the logic of how a version's action is determined without adding more derived state to the underlying model. I also took the opportunity to update the Activity Sidebar's version item to SUIT.


boxcla:

flow-typed/box-ui-elements.js
```diff
@@ -352,20 +352,22 @@ type JSONPatch = {
...
     extension?: string,
     id: string,
     is_download_available?: boolean,
     modified_at?: string,
-    modified_by: User,
+    modified_by: ?User,
     name?: string,
     permissions?: BoxItemVersionPermission,
+    restored_at?: string,
+    restored_by?: ?User,
```
mickr: Doesn't this have a default value of the placeholder user? If so, this shouldn't be a maybe type? 🍹

jstoffan: It doesn't, unfortunately. Its value can be explicitly `null` in the endpoint response.

PR #1572, opened 2019-08-26
fix(datepicker): use locale preference for week start date

alexkrolick:
NOTE: do not merge unless root level react-intl providers are set up to provide full locale string (`en-US` not `en`).


mxiao6:
@alexkrolick You can merge it now!

PR #1579, opened 2019-08-28
refactor(css): extract deprecated variables

alexkrolick:


boxcla:

booc0mtaco:
Let's check for whether the generator for variables.js/json is still doing what it should. expected some changes here

jstoffan:
@alexkrolick, I think the `style` commit prefix usually refers to coding style rather than CSS styles. Should this be a `fix`, instead?

alexkrolick:
@jstoffan Isn't style for code style? Like naming conventions and stuff? I guess `refactor` is the other good candidate?

scripts/utils/bdlColorManager
```diff
@@ -55,9 +57,10 @@ const conversionMap = {
     'better-black': 'bdl-gray-80',
     'downtown-grey': 'bdl-gray-50',
     'twos': 'bdl-gray',
+    'threes': 'bdl-gray',
     'fours': 'bdl-gray-80',
     'sevens': 'bdl-gray-62',
-    'eights': 'bdl-gray-50',
+    'eights': 'bdl-gray-50', // false positives
```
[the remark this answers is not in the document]
jstoffan: May be worthwhile to update the comment with that context.

package.json
```diff
@@ -49,7 +49,7 @@
         "build:prod:dist": "NODE_ENV=production webpack --config scripts/webpack.config.js --mode production",
         "build:prod:examples": "LANGUAGE=en-US REACT=true NODE_ENV=production node --max_old_space_size=8192 node_modules/react-styleguidist/bin/styleguidist.js build --config scripts/styleguide.config.js --mode production",
         "build:prod:npm": "BABEL_ENV=production OUTPUT=dist LANGUAGE=en-US REACT=true yarn build:prod:dist",
-        "build:variables": "node ./scripts/build-style-vars.js ./src/styles/_variables.scss ./src/styles/variables.js ./src/styles/variables.json",
```
jstoffan: I think we'd still want people to import `variables.json` rather than each individual component of it.

alexkrolick: So far it's only been used in the `styleguide.config.js` file due to ES module import issues. The issue is that the build-style-vars script can't parse transient variables, so a root-level scss file that reexports other files gets converted to an empty JS/JSON file. I worked around it for the ES module by converting to a manual reexport of the submodules.

alexkrolick: 🤔 

```
cat `find src/styles/constants -name '*.scss'` > src/styles/variables.scss | node ./scripts/build-style-vars.js src/styles/variables.scss && rm src/styles/variables.scss
```

src/styles/variables.json
```diff
@@ -60,74 +60,12 @@
...
     "bdl-watermelon-red-05": "#fef5f6",
     "bdl-watermelon-red-02": "#fffbfc",
-    "bdl-secondary-blue": "#003c84",
-    "bdl-neutral-01": "#222",
-    "bdl-neutral-02": "#767676",
-    "bdl-neutral-03": "#909090",
-    "bdl-neutral-04": "#e8e8e8",
-    "bdl-neutral-05": "#f4f4f4",
-    "bdl-neutral-06": "#fbfbfb",
-    "box-blue": "#0061d5",
```
jstoffan: We're sure that these values aren't being used anywhere? Are we expecting consumers to import `deprecated.*` files directly?

alexkrolick: - `variables.json` is only used in `styleguide.config.js`
- `variables.js` is used by js code but Flow can tell us what is used. This PR fixes anything in BUIE affected by the change, and EUA already has removed the deprecated values.
- `_variables.scss` continues to re-export the deprecated values for the few places that haven't been ported over.

PR #1582, opened 2019-08-30
Creating task empty icon

il10:
creating task empty state icon


boxcla:

PR #1585, opened 2019-08-30
feat(avatarinitials): move backgroundColor into scss

cweeii:


boxcla:

booc0mtaco:
What's the use case here? I'm guessing you don't actually want any/all style attributes to be set inline for this component.

cweeii:
@booc0mtaco You're right, it's mainly to overwrite backgroundColor and add add color.  For some context:  in accessibility mode, we'd like to have two possible versions, white initials on black background or black initials on white background.  If I were to account for these two versions in props, I might have to do something like `shouldHaveHighContrastWithDarkBackground` or `shouldHaveHighContrastWithWhiteBackground`.  Or maybe even do a merge of two props, `isHighContrast`, and `hasDarkBackground` or `hasLightBackground`.  Instead of playing around with the props, I felt it was a little more generic by just allowing the overriding of AvatarInitials style.

cweeii:
After some in-person discussion, we came to a conclusion of having a Contrast mode type somewhere and using that type to define style overrides in AvatarInitials

alexkrolick:
It seems like it would be easier to define a `className` that can be targeted with an injected stylesheet than to override the style prop everywhere an Avatar needs to comply with this mode

cweeii:
@alexkrolick since the background color is calculated via the id, do you suggest making a class for each color (in order to move it out from the style prop)?

src/components/avatar/AvatarInitials.js
```diff
@@ -12,6 +12,7 @@ import {
     bdlWatermelonRed,
 } from '../../styles/variables';
 
+// NOTE: The same list exists in Avatar.scss
```
jstoffan: This isn't ideal. It seems like all we need is the length of the list? Can we declare the list as a variable in `src/styles/constants/_colors.scss`, then import it here and in the Avatar.scss file?

src/components/avatar/AvatarInitials.js
```diff
@@ -38,9 +39,10 @@ type Props = {
 
 const AvatarInitials = ({ className = '', id, name }: Props) => {
     const avatarColorSelector = parseInt(id, 10) || 0;
-    const backgroundColor = AVATAR_COLORS[avatarColorSelector % AVATAR_COLORS.length];
+    const backgroundColorIndex = avatarColorSelector % AVATAR_COLORS.length;
```
jstoffan: Can we do this math in the SCSS file to avoid needing the list length entirely? We could set the id itself as a `data-` attribute, instead.

src/components/avatar/AvatarInitials.js
alexkrolick: Maybe not there exactly - that should be a pretty pure file with just the palette definitions

alexkrolick: I was thinking of adding `src/styles/constants/_theme.scss` for aliases and derived values

src/components/avatar/AvatarInitials.js
cweeii: I tried going down this route but I would then need to add some untestable(?) sass logic to parse the id since attributes are strings.

src/styles/variables.js
```diff
@@ -64,6 +64,7 @@ export const bdlWatermelonRed10 = "#fdebee"; // bdl-watermelon-red-10
 export const bdlWatermelonRed05 = "#fef5f6"; // bdl-watermelon-red-05
 export const bdlWatermelonRed02 = "#fffbfc"; // bdl-watermelon-red-02
 export const bdlLineHeight = "20px"; // bdl-line-height
+export const avatarColors = "#2486fc #0061d5 #003c84 #767676 #26c281 #f5b31b #4826c2 #9f3fed #ed3757"; // avatar-colors
```
alexkrolick: this doesn't seem that useful as a string...

cweeii: Yeah I'm looking into seeing if I could make it an array since by default it looks like the parser just makes it a large string.

src/components/avatar/Avatar.scss
```diff
@@ -38,6 +37,14 @@
         line-height: 1;
     }
 
+    @for $i from 1 through length($avatar-colors) {
+        $backgroundColor: nth($avatar-colors, $i);
```
jstoffan: This probably doesn't need to be a variable.

src/styles/variables.js
jstoffan: It looks like it's a limitation of the variable parser? Should be okay as-is.

src/components/avatar/Avatar.scss
cweeii: Removed

PR #1597, opened 2019-09-04
fix(pillselectordropdown): support paste from context menu

booc0mtaco:
Due to how certain versions of IE11 handle paste and input events, sometimes the UI for `PillSelectorDropdown` instances would handle the pill creation **before** parsing the input. this would cause the field to be cleared before pills were created and would also destroy the user input. 

This change uses the [`Clipboard API`](https://caniuse.com/#search=clipboarddata) synthetic event to get the input data on paste, and refactors the pill parsing methods to use a parameter, so that it always uses the proper context and input value based on the current user workflow.


boxcla:

src/components/pill-selector-dropdown/PillSelectorDropdown.js
```diff
@@ -177,21 +176,24 @@ class PillSelectorDropdown extends React.Component<Props, State> {
...
 
-    handlePaste = () => {
+    handlePaste = (event: SyntheticClipboardEvent<HTMLInputElement>) => {
         /**
          * NOTE (ishay): setTimeout is necessary because
          * otherwise addPillsFromInput gets triggered as soon
          * as the user "paste's", but before the inputValue
          * is actually updated.
          */
-        setTimeout(this.addPillsFromInput, 0);
```
alexkrolick: 👀 this is the problem

src/components/pill-selector-dropdown/__tests__/PillSelectorDropdown-test.js
```diff
@@ -435,19 +441,68 @@ describe('components/pill-selector-dropdown/PillSelectorDropdown', () => {
...
-                .mock(instance)
-                .expects('addPillsFromInput')
-                .once();
+            instance.handlePaste(mockEvent);
+
+            expect(mockOnInput).toHaveBeenCalledWith(mockPastedValue, mockEvent);
+            expect(mockOnPillCreate).not.toHaveBeenCalled();
+        });
+
+        test('should call onPillCreate prop method with valid input', () => {
```
alexkrolick: 👍

src/features/unified-share-modal/UnifiedShareModal.js
```diff
@@ -363,6 +363,11 @@ class UnifiedShareModal extends React.Component<Props, State> {
                 }
             });
         }
+
+        this.setState({
```
jstoffan: This seems unrelated?

src/components/pill-selector-dropdown/PillSelectorDropdown.js
```diff
@@ -177,21 +176,20 @@ class PillSelectorDropdown extends React.Component<Props, State> {
...
-    handlePaste = () => {
-        /**
-         * NOTE (ishay): setTimeout is necessary because
-         * otherwise addPillsFromInput gets triggered as soon
-         * as the user "paste's", but before the inputValue
-         * is actually updated.
-         */
-        setTimeout(this.addPillsFromInput, 0);
+    handlePaste = (event: SyntheticClipboardEvent<HTMLInputElement>) => {
+        event.nativeEvent.preventDefault();
```
jstoffan: Do we need to access `nativeEvent` here or does `event.preventDefault()` work?

src/components/pill-selector-dropdown/PillSelectorDropdown.js
```diff
@@ -177,21 +176,20 @@ class PillSelectorDropdown extends React.Component<Props, State> {
...
+
+        const inputValue: string = event.clipboardData.getData('Text');
+        this.setState({ inputValue });
+        this.props.onInput(inputValue, event);
```
jstoffan: Is this going to cause issues with consuming expecting a `SyntheticInputEvent` rather than a `SyntheticClipboardEvent`? The flow type currently doesn't specify the parameters.

src/components/pill-selector-dropdown/PillSelectorDropdown.js
```diff
@@ -177,21 +176,20 @@ class PillSelectorDropdown extends React.Component<Props, State> {
...
-         * NOTE (ishay): setTimeout is necessary because
-         * otherwise addPillsFromInput gets triggered as soon
-         * as the user "paste's", but before the inputValue
-         * is actually updated.
-         */
-        setTimeout(this.addPillsFromInput, 0);
+    handlePaste = (event: SyntheticClipboardEvent<HTMLInputElement>) => {
+        event.nativeEvent.preventDefault();
+
+        const inputValue: string = event.clipboardData.getData('Text');
```
jstoffan: Does this work in all browsers? Should it be `clipboardData.getData('text')` or `clipboardData.getData('text/plain')`, per [MDN](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/getData)?

src/features/unified-share-modal/UnifiedShareModal.js
booc0mtaco: true, will remove (came up in testing)

src/components/pill-selector-dropdown/PillSelectorDropdown.js
booc0mtaco: It appears that the synthetic event works just as well, so i can switch to that

src/components/pill-selector-dropdown/PillSelectorDropdown.js
booc0mtaco: The types around this are unfortunate :( BUT, the type for `handleInput` specifies an event or some object with `target`, which `SyntheticClipboarEvent` – and by extension its native cousin – has. So the minimum is checked here.

src/components/pill-selector-dropdown/PillSelectorDropdown.js
booc0mtaco: React appears to normalize the parts of the event specific to the Clipboard API. There used to be more details about the `.types` documented, but they took those off. I'll double-check if a non-legacy MIME type works here (e.g., `text/plain`)

booc0mtaco: https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/types

booc0mtaco: `Text` and `text` work in all browsers, but IE11 doesn't support the full MIME types. Will lower-case this

PR #1598, opened 2019-09-04
feat(metadata): Enable metadata template filtering

sahiga:
This feature enables template filtering in the Metadata Sidebar, which consumes two new props:
```
selectedTemplateKey?: string,
templateFilters?: Array<string> or string
```
If the `selectedTemplateKey` prop is provided, the underlying `TemplateInstance` will only show the template that corresponds to the key. If the `templateFilters` prop is provided, the `TemplateInstance` will only display (and allow edits for) the fields that match the IDs in the array (or the ID, if `templateFilters` is a string). Otherwise, the `TemplateInstance` will display all fields from the template.


boxcla:

src/features/metadata-instance-editor/Instances.js
```diff
@@ -14,27 +14,50 @@ type Props = {
...
-            />
-        ),
+        ({ isDirty = false, instance, hasError = false, template }: MetadataEditor) => {
+            const { templateKey } = template;
+            let isOpen = false;
+            let includedFieldIds;
+            // Open the included template by default, and only display the included fields in the template
+            if (templateFilters) {
+                isOpen = templateKey === templateFilters.includedTemplateKey;
+                if (isOpen && templateFilters.includedFieldIds) {
```
jstoffan: Could we move this check to a `.filter` call on line 20? It would remove the instance from the output array and avoid the need to `includedFieldIds`, I think.

src/features/metadata-instance-editor/Instances.js
```diff
@@ -14,27 +14,50 @@ type Props = {
...
+                    includedFieldIds = new Set(templateFilters.includedFieldIds);
```
jstoffan: Should this de-duping be done at a higher level?

src/features/metadata-instance-editor/__tests__/TemplatedInstance-test.js
```diff
@@ -160,4 +162,19 @@ describe('features/metadata-instance-editor/fields/TemplatedInstance', () => {
...
+                data={data}
+                dataValue="value"
+                errors={{}}
+                includedFieldIds={includedFieldIds}
+                template={{
+                    fields,
+                }}
+            />,
+        );
+        expect(wrapper).toMatchSnapshot();
```
jstoffan: We should try to avoid snapshots, especially on such large components. Can we add more explicit test cases?

src/features/metadata-instance-editor/Instances.js
```diff
@@ -14,27 +14,50 @@ type Props = {
...
-                template={template}
-            />
-        ),
+        ({ isDirty = false, instance, hasError = false, template }: MetadataEditor) => {
+            const { templateKey } = template;
+            let isOpen = false;
+            let includedFieldIds;
+            // Open the included template by default, and only display the included fields in the template
+            if (templateFilters) {
+                isOpen = templateKey === templateFilters.includedTemplateKey;
```
jstoffan: It looks like `includedTemplateKey` only controls whether the template is open by default. Should it be a separate prop rather than part of `templateFilters`?

src/features/metadata-instance-editor/MetadataInstanceEditor.js
```diff
@@ -57,6 +62,8 @@ const MetadataInstanceEditor = ({
                     onModification={onModification}
                     onRemove={onRemove}
                     onSave={onSave}
+                    selectedTemplateKey={selectedTemplateKey}
+                    templateFilters={templateFilters ? convertTemplateFilters(templateFilters) : undefined}
```
jstoffan: Is this needed? It looks like we're doing the same in MetadataSidebar.

src/features/metadata-instance-editor/Instances.js
```diff
@@ -14,27 +14,43 @@ type Props = {
...
-                onModification={onModification}
-                onSave={onSave}
-                onRemove={onRemove}
-                template={template}
-            />
-        ),
+        ({ isDirty = false, instance, hasError = false, template }: MetadataEditor) => {
+            const { templateKey } = template;
+            // Open the included template by default, and only display the included fields in the template
+            const isOpen = templateKey === selectedTemplateKey || (!selectedTemplateKey && editors.length === 1);
```
jstoffan: Can we retain the prior `isOpen` behavior, even if a `selectedTemplateKey` is provided?

```
isOpen={editors.length === 1 || templateKey === selectedTemplateKey}
```

src/features/metadata-instance-editor/Instance.js
```diff
@@ -642,6 +650,7 @@ class Instance extends React.PureComponent<Props, State> {
                                         canEdit={isEditing}
                                         data={data}
                                         errors={errors}
+                                        templateFilters={templateFilters}
```
jstoffan: Minor, but can we alphabetize these props? Same for the destructured object above?

src/features/metadata-instance-editor/__tests__/metadataUtil-test.js
```diff
@@ -1,102 +1,125 @@
...
+            filters: singleFilter,
+            description: 'should convert a single filter into a Set object with one item',
+            expected: new Set([singleFilter]),
+        },
+        {
+            filters: multipleFilters,
+            description: 'should convert multiple template filters into a Set object',
+            expected: new Set(multipleFilters),
+        },
+    ].forEach(({ description, filters, expected }) => {
```
jstoffan: We don't really use this style of testing as much now that we've moved away from Karma. Can we convert this to use `jest.each`?

src/features/metadata-instance-editor/metadataUtil.js
```diff
@@ -1,7 +1,16 @@
 // @flow
 
+/**
+ * Utility function for converting a string or array of strings into a Set object
+ * @param templateFilters - Array<string> | string
+ * @returns {Set<T>}
+ */
+const convertTemplateFilters = (templateFilters: Array<string> | string): Set<string> => {
```
jstoffan: `normalizeTemplateFilters`?

src/features/metadata-instance-editor/MetadataInstanceEditor.js
sahiga: I added the ternary statement so that we would only pass the normalized `templateFilters` prop in the event that it is defined. I thought it would be cleaner to not pass `templateFilters` at all than to pass `templateFilters={null}`, but I can change this. What do you think?

src/features/metadata-instance-editor/metadataUtil.js
```diff
@@ -4,4 +4,38 @@ const isHidden = (obj: MetadataTemplate | MetadataTemplateField): boolean => {
...
+    templates: Array<MetadataTemplate>,
+    selectedTemplateKey?: string,
+    templateFilters?: Array<string> | string,
+): Array<MetadataTemplate> => {
+    let clonedTemplates = templates.slice(0); // cloned for editing purposes
+    if (!selectedTemplateKey) {
+        return clonedTemplates;
+    }
+    clonedTemplates = clonedTemplates.filter(template => template.templateKey === selectedTemplateKey);
+    const fields = clonedTemplates[0] ? clonedTemplates[0].fields : null;
```
manishPh: Can we do:
```
if (!selectedTemplateKey) {
    return [...templates];
}
const clonedTemplates = templates.filter(template => template.templateKey === selectedTemplateKey);
...  // the rest of the code
```
?

src/elements/content-sidebar/MetadataSidebar.js
```diff
@@ -416,6 +420,8 @@ class MetadataSidebar extends React.PureComponent<Props, State> {
                                 onModification={this.onModification}
                                 onRemove={this.onRemove}
                                 onSave={this.onSave}
+                                selectedTemplateKey={selectedTemplateKey}
+                                templates={templates}
```
jstoffan: Does this need to be passed in? It doesn't seem to be used.

src/elements/content-sidebar/README.md
```diff
@@ -61,7 +61,12 @@ var ContentSidebar = require('./ContentSidebar').default;
 #### MetadataSidebarProps
 | Prop | Type | Default | Description |
 | --- | --- | --- | --- |
-| *tbd* | - | - | - |
+| selectedTemplateKey | `string`         |  | `templateKey` for the template to show. If provided, |
+|                     |                  |  | no other templates will be visible. |
+| templateFilters     | `Array<string>`  |  | Template field IDs for fields that should be visible and editable. |
```
jstoffan: Could also be simplified to just accept an array of strings, if that makes things easier.

src/elements/content-sidebar/MetadataSidebar.js
sahiga: Good catch!

PR #1600, opened 2019-09-05
fix(uploads): reset item error after retry

rosepierce:
After an upload fails and has an error tied to the item, this error is currently never removed when retrying an upload. If this error exists when the retried upload finishes, the status of the item in the uploads manager is not set to completed, and so the item will continue to show the loading bar in the uploads manager (and because it is in a staged state, cannot be removed from the uploads manager due to the remove button being disabled). This PR removes the error after an upload is retried, so that the uploads manager will show the upload as completed after it finishes.

### DIFF
--- a/src/elements/content-uploader/ContentUploader.js
+++ b/src/elements/content-uploader/ContentUploader.js
@@ -768,6 +768,7 @@ class ContentUploader extends Component<Props, State> {
         item.api = this.getUploadAPI(file, options);
         item.progress = 0;
         item.status = STATUS_PENDING;
+        delete item.error;
 
         const { items } = this.state;
         items[items.indexOf(item)] = item;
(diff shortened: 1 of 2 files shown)


boxcla:

mxiao6:
I found there is no unit test for `resetFile`. Can you add one?

src/elements/content-uploader/ContentUploader.js
```diff
@@ -768,6 +768,7 @@ class ContentUploader extends Component<Props, State> {
         item.api = this.getUploadAPI(file, options);
         item.progress = 0;
         item.status = STATUS_PENDING;
+        delete item.error;
```
jstoffan: Should we have a unit test to go along with this to avoid a regression in the future?

PR #1602, opened 2019-09-09
fix(user-link): allow wrapping of long user names

tjgupta:
Allow long user names to wrap in comments and tasks.

Before:
Screen Shot 2019-09-09 at 10 47 48 AM

After:
Screen Shot 2019-09-09 at 10 48 03 AM


boxcla:

PR #1604, opened 2019-09-10
fix(activity-feed): fix long user names in tasks/comments in Edge

tjgupta:
Follow-up to https://github.com/box/box-ui-elements/pull/1602 that fixes long username wrapping in Edge. Sorry in advance for the browser targeting, but I tried and tried and couldn't find a better way that didn't break other browsers and fixed Edge... :(

### DIFF
--- a/src/elements/content-sidebar/activity-feed/comment/Comment.scss
+++ b/src/elements/content-sidebar/activity-feed/comment/Comment.scss
@@ -16,6 +16,14 @@
     }
 }
 
+.bcs-Comment-headline {
+    // target only Edge because this breaks other browsers
+    // but fixes Edge
+    @supports (-ms-ime-align: auto) {
+        word-break: break-all;
+    }
+}
+
 // tethered styles for delete confirmation modal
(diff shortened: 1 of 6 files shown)


boxcla:

src/elements/content-sidebar/activity-feed/comment/Comment.scss
```diff
@@ -16,6 +16,14 @@
     }
 }
 
+.bcs-Comment-headline {
```
jstoffan: Should we apply this to the UserLink component directly, so it's fixed for all usages?

tjgupta: @jstoffan for some reason that didn't work. It only worked on the `<div>` above it.

tjgupta: Actually it also worked if I did it on the entire `<Media.Body>` too, but that seems worse with potentially more awkward side effects for Edge.

PR #1605, opened 2019-09-11
fix(task-form): clear task assignee autocomplete on form submit

tjgupta:
This fixes an issue where:
1. user creates a new task (from Add Task button)
2. user enters name for task assignee field and gets autocomplete results
3. user selects an assignee (or multiple assignees) and saves task
4. user goes to create another new task, but previous autocomplete results weren't being cleared

This wasn't an issue when entering the task form by editing an existing task, because that was already clearing autocomplete results on form submit properly.


boxcla:

src/elements/content-sidebar/AddTaskButton.js
```diff
@@ -45,7 +45,11 @@ class AddTaskButton extends React.Component<Props, State> {
         this.setState({ isTaskFormOpen: false, error: null });
     };
 
-    handleSubmitSuccess = () => this.setState({ isTaskFormOpen: false, error: null });
+    handleSubmitSuccess = () => {
```
jstoffan: Can this be combined with, replaced by, or call the `handleModalClose` method?

tjgupta: Hmm, so with this change `onTaskModalClose` is called for both Cancel and Submit (success), and it clears the autocomplete state stored up in `<ActivitySidebar>`. This handler is just closing the modal and resetting the error (and now calling to clear the autocomplete state).

I'm not sure of a way to combine since it needs to reset state at two different component levels. See https://github.com/box/box-ui-elements/blob/master/src/elements/content-sidebar/activity-feed/task-new/Task.js#L141-L148 for the other place that we're doing this same logic (for opening the same form for editing an existing task).

PR #1606, opened 2019-09-11
chore(deps): update react peer/dev dep

alexkrolick:
- update react to 16.9.0
- update react-dom to 16.9.0
- update enzyme deps to update react-test-renderer to 16.9.0


j3tan:
Must be declared as a breaking change due to peerDep update. Make sure the `react-test-renderer` in `yarn.lock` gets updated too

j3tan:
I would also include a warning in the PR that `react-tether` will start throwing warnings and we don't have any immediate plans to address it

yarn.lock
```diff
@@ -1237,64 +783,66 @@
...
+    browserslist "^4.3.4"
     invariant "^2.2.2"
     js-levenshtein "^1.1.3"
     semver "^5.3.0"
 
 "@babel/preset-env@^7.5.5":
-  version "7.5.5"
-  resolved "https://registry.yarnpkg.com/@babel/preset-env/-/preset-env-7.5.5.tgz#bc470b53acaa48df4b8db24a570d6da1fef53c9a"
-  integrity sha512-GMZQka/+INwsMz1A5UEql8tG015h5j/qjptpKY2gJ7giy8ohzU710YciJB5rcKsWGWHiW3RUnHib0E5/m3Tp3A==
+  version "7.6.0"
```
jstoffan: Is upgrading babel and other build dependencies needed for React 16.9?

package.json
```diff
@@ -225,10 +225,10 @@
         "properties-parser": "^0.3.1",
         "query-string": "5.1.1",
         "raf": "^3.4.0",
-        "react": "^16.8.0",
+        "react": "16.9.0",
```
jstoffan: Can we leave the `^` here or is the lockdown intentional?

yarn.lock
alexkrolick: It happened through `yarn upgrade --dev enzyme`

jstoffan: I think `yarn upgrade` always upgrades all packages, even if you pass it a specific package name. Can you reset yarn.lock and run the following, instead?

```
yarn add react@^16.9.0 react-dom@^16.9.0
```

alexkrolick: That doesn't update the transient dependency react-test-renderer

alexkrolick: > Optionally, one or more package names can be specified. When package names are specified, only those packages will be upgraded. When no package names are specified, all dependencies will be upgraded.

https://yarnpkg.com/lang/en/docs/cli/upgrade/#toc-yarn-upgrade-package-package-tag-package-version-scope-ignore-engines-pattern

🤷‍♂

package.json
alexkrolick: leaving it

PR #1617, opened 2019-09-12
perf(sidebar): merge response to avoid redundant api calls

mxiao6:


src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -217,6 +217,30 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
         return versions[0] ? versions[0].id : null;
     };
 
+    mergeVersions = (newVersion: BoxItemVersion) => {
+        const { versions } = this.state;
+
+        const newVersions = cloneDeep(versions).map(version => {
+            if (version.id !== newVersion.id) {
```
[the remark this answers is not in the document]
mxiao6: good idea

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -217,6 +217,26 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
         return versions[0] ? versions[0].id : null;
     };
 
+    mergeVersions = (newVersion: BoxItemVersion) => {
+        const { versions } = this.state;
+        const newVersions = cloneDeep(versions).map(version =>
```
jstoffan: I think you can use array destructuring rather than `cloneDeep` here. You may also want to use `merge` for the `version` objects themselves.

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -217,6 +217,26 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
         return versions[0] ? versions[0].id : null;
     };
 
+    mergeVersions = (newVersion: BoxItemVersion) => {
```
jstoffan: Can we add flow types for the return values of the new methods?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -217,6 +217,26 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
...
+        const { versions } = this.state;
+        const newVersions = cloneDeep(versions).map(version =>
+            version.id !== newVersion.id ? version : { ...version, ...newVersion },
```
jstoffan: We usually want conditions to assert `is` before or instead of `is not` when possible. Can we switch this to `===`?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -217,6 +217,26 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
...
+        );
+        return newVersions;
```
jstoffan: The `newVersions` variable isn't needed, as the `.map` result can be returned directly.

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -95,8 +96,8 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
 
         return this.api
             .deleteVersion(this.findVersion(versionId))
-            .then(this.api.fetchData)
-            .then(this.handleFetchSuccess)
+            .then(() => this.api.fetchVersion(versionId))
+            .then(this.mergeResponse)
```
jstoffan: Should this call be moved to `handleDeleteSuccess`?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -132,8 +133,7 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
 
         return this.api
             .restoreVersion(this.findVersion(versionId))
-            .then(this.api.fetchData)
-            .then(this.handleFetchSuccess)
+            .then(this.mergeResponse)
```
jstoffan: Can we introduce a `handleRestoreSuccess` helper that calls `mergeVersions` and then sets state?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -217,6 +223,23 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
...
+    mergeResponse = (data: BoxItemVersion): Promise<Array<BoxItemVersion>> => {
+        const newVersions = this.mergeVersions(data);
+
+        this.setState({
+            error: undefined,
+            isLoading: false,
+            versions: newVersions,
+        });
+
+        return Promise.resolve(newVersions);
```
jstoffan: I don't think `Promise.resolve` is needed here.

src/api/Versions.js
```diff
@@ -58,7 +64,7 @@ class Versions extends OffsetBasedAPI {
             return;
         }
 
-        this.successCallback(data);
+        this.successCallback(data || { id: this.versionId });
```
jstoffan: It seems a little odd to always retain the original parameter and pass it back if the data payload is undefined. Can we use the original version id in the consuming call, instead?

PR #1622, opened 2019-09-18
fix(classification): badge should not have a tabindex

priyajeet:


boxcla:

PR #1623, opened 2019-09-18
feat(activity-feed): add support for deeplinking tasks and comments

tjgupta:


boxcla:

src/elements/content-sidebar/Sidebar.js
```diff
@@ -112,11 +126,13 @@ class Sidebar extends React.Component<Props, State> {
...
         const fileVersionSlug = currentVersion ? `/${currentVersion.id}` : '';
 
+        const urlPrefix = this.getUrlPrefix(history.location.pathname);
+
         if (event.preventDefault) {
             event.preventDefault();
         }
 
-        history.push(`${history.location.pathname}/versions${fileVersionSlug}`);
+        history.push(`/${urlPrefix}/versions${fileVersionSlug}`);
```
tjgupta: This was previously assuming that the path had no deeplink, so for example `/activity` or `/details`, but now with these deeplinks this was breaking by appending the pathname with `versions`. For example, when clicking from a deeplinked comment the pathname is `/activity/comments/12345` so this was resulting in `/activity/comments/12345/versions/6789`, which is obviously not what we want here.

src/elements/content-sidebar/SidebarPanels.js
```diff
@@ -218,6 +205,35 @@ class SidebarPanels extends React.Component<Props> {
                         )}
                     />
                 )}
+                {/* This handles both the default activity sidebar and the activity sidebar with a
+                comment or task deeplink.  */}
+                {hasActivity && (
+                    <Route
+                        exact
+                        path={[
+                            `/${SIDEBAR_VIEW_ACTIVITY}`,
```
jstoffan: Can we avoid this array by using optional parameters syntax? Can we use parameter regex to match only comment/task?

```
`/${SIDEBAR_VIEW_ACTIVITY}/:activeFeedItemType(comments|tasks)?/:activeFeedItemId?`
```

src/elements/content-sidebar/SidebarPanels.js
```diff
@@ -218,6 +205,35 @@ class SidebarPanels extends React.Component<Props> {
...
+                            `/${SIDEBAR_VIEW_ACTIVITY}/:activeFeedItemType/:activeFeedItemId`,
+                        ]}
+                        render={({ match }) => {
+                            const activeFeedItemType = match.params.activeFeedItemType
```
jstoffan: "Item" usually refers to a file or folder. Can the variable names be more generic, such as `activityId` and `activityType`?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
```diff
@@ -70,6 +75,19 @@ class ActivityFeed extends React.Component<Props, State> {
         if (hasLoaded || hasMoreItems || hasNewItems || hasInputOpened) {
             this.resetFeedScroll();
         }
+
+        // do the scroll only once after first fetch of feed items
+        if (prevFeedItems === undefined && currFeedItems !== undefined) {
```
jstoffan: Can we use `isEmpty`/`wasEmpty` or `hasLoaded` here?

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -77,7 +91,12 @@ const ActiveState = ({
                     );
                 case 'task':
                     return (
-                        <li key={type + id} className="bcs-activity-feed-task-new" data-testid="task">
+                        <li
+                            key={type + id}
+                            className={classNames('bcs-activity-feed-task-new', { 'bcs-is-focused': isFocused })}
+                            data-testid="task"
+                            ref={refValue}
```
jstoffan: Rather than passing refs up to the parent, can we pass something like `isActive` to the child components and allow them to scroll themselves into view? Similar to how [version items](https://github.com/box/box-ui-elements/blob/master/src/elements/content-sidebar/versions/VersionsItemButton.js) work?

src/elements/content-sidebar/Sidebar.js
jstoffan: It seems like we're going through a lot of trouble to have the `/tasks` and `/comments` subpaths. Would it be simpler to make it a compound parameter, such as `/activity/task_12345`? React Router supports regexes for parameters.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
tjgupta: We decided to defer this change to a followup PR. May involve creating a reusable component like `<ScrollFocusable>` or similar as well.

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
```diff
@@ -177,9 +202,16 @@ class ActivityFeed extends React.Component<Props, State> {
                     }}
                     className="bcs-activity-feed-items-container"
                 >
-                    {isEmpty ? (
-                        <EmptyState isLoading={isLoading} showCommentMessage={showCommentForm} />
-                    ) : (
+                    {isLoading && (
```
tjgupta: We ended up pulling this loader out of `<EmptyState>` because the logic for empty and loaded was so conflated and confusing (and buggy).

src/elements/content-sidebar/activity-feed/activity-feed/EmptyState.js
```diff
@@ -6,30 +6,26 @@
 import * as React from 'react';
 import { FormattedMessage } from 'react-intl';
 
-import LoadingIndicator from '../../../../components/loading-indicator/LoadingIndicator';
+// import LoadingIndicator from '../../../../components/loading-indicator/LoadingIndicator';
```
tjgupta: Oops, need to remove.

src/elements/content-sidebar/Sidebar.js
```diff
@@ -101,6 +102,17 @@ class Sidebar extends React.Component<Props, State> {
...
 
+    getUrlPrefix = (pathname: string) => {
+        const pathNameParts = pathname.substring(1).split('/');
+        const basePath = pathNameParts[0];
+        const pathNameMap = {
+            activity: SIDEBAR_VIEW_ACTIVITY,
+            details: SIDEBAR_VIEW_DETAILS,
+            metadata: SIDEBAR_VIEW_METADATA,
+        };
+        return pathNameMap[basePath] || SIDEBAR_VIEW_ACTIVITY;
```
jstoffan: Is this map necessary or can we use the `basePath` value directly?

src/elements/content-sidebar/SidebarPanels.js
```diff
@@ -218,6 +205,32 @@ class SidebarPanels extends React.Component<Props> {
...
                     />
                 )}
+                {/* This handles both the default activity sidebar and the activity sidebar with a
+                comment or task deeplink.  */}
+                {hasActivity && (
+                    <Route
+                        exact
+                        path={`/${SIDEBAR_VIEW_ACTIVITY}/:activeFeedEntryType(comments|tasks)?/:activeFeedEntryId?`}
+                        render={({ match }) => {
+                            const activeFeedEntryType = match.params.activeFeedEntryType
```
jstoffan: Can we destructure `activeFeedEntryType` to avoid duplicating its access?

src/elements/content-sidebar/SidebarPanels.js
```diff
@@ -202,6 +187,8 @@ class SidebarPanels extends React.Component<Props> {
                         )}
                     />
                 )}
+                {/* This is before the activity feed subroutes because otherwise it doesn't get
```
jstoffan: Is this still needed with the addition of the `(comments|tasks)` regex below? If not, can we also move the main Activity route back to its original position?

src/elements/content-sidebar/SidebarPanels.js
```diff
@@ -202,6 +187,8 @@ class SidebarPanels extends React.Component<Props> {
...
+                    matched properly due to the :sidebar wildcard. */}
                 {hasVersions && (
                     <Route
                         path="/:sidebar/versions/:versionId?"
```
jstoffan: Can we change this to `path="/:sidebar(activity|details)/versions/:versionId?"` to clarify where it's meant to be used?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
```diff
@@ -55,33 +61,50 @@ class ActivityFeed extends React.Component<Props, State> {
...
+            this.scrollToActiveFeedItem();
+        }
+    }
+
+    scrollToActiveFeedItem() {
+        const { activeFeedEntryId } = this.props;
+        if (!activeFeedEntryId) {
+            return;
+        }
+        if (activeFeedEntryId && this.activeFeedItemRef.current) {
```
jstoffan: Is this condition duplicated in the guard clause? Is it needed or can we check the existence of the ref itself? Can we destructure the ref variable, such as:

```
const { current: activeFeedItemRef } = this.activeFeedItemRef;

if (!activeFeedItemRef) {
    return;
}

scrollIntoView(activeFeedItemRef);
```

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
```diff
@@ -177,9 +202,16 @@ class ActivityFeed extends React.Component<Props, State> {
                     }}
                     className="bcs-activity-feed-items-container"
                 >
-                    {isEmpty ? (
-                        <EmptyState isLoading={isLoading} showCommentMessage={showCommentForm} />
-                    ) : (
+                    {isLoading && (
+                        <div className="bcs-activity-feed-loading-state">
+                            {' '}
```
jstoffan: We should avoid using spaces for visual spacing. Can we use CSS, instead?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
```diff
@@ -177,9 +202,16 @@ class ActivityFeed extends React.Component<Props, State> {
                     }}
                     className="bcs-activity-feed-items-container"
                 >
-                    {isEmpty ? (
-                        <EmptyState isLoading={isLoading} showCommentMessage={showCommentForm} />
-                    ) : (
+                    {isLoading && (
+                        <div className="bcs-activity-feed-loading-state">
```
jstoffan: Time for SUIT?

src/elements/content-sidebar/activity-feed/activity-feed/__tests__/ActivityFeed-test.js
```diff
@@ -2,10 +2,12 @@ import * as React from 'react';
 import { shallow } from 'enzyme';
 
 import ActivityFeed from '../ActivityFeed';
+import { scrollIntoView } from '../../../../../utils/dom';
 
 jest.mock('../../Avatar', () => 'Avatar');
 jest.mock('../ActiveState', () => 'ActiveState');
 jest.mock('lodash/uniqueId', () => () => 'uniqueId');
```
jstoffan: I realize this is existing code, but can we mock in the same order we import, with third-party libraries first?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
tjgupta: If you don't mind, I'd prefer to make that change separately. It's likely to be pretty large. Unless you mean changing this one selector to SUIT?

src/elements/content-sidebar/Sidebar.js
jstoffan: Too paranoid, I think. 😄

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
jstoffan: I don't mind at all. It's more a statement that, if we're going to be mucking around in here more now, we should take the time to convert it.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -557,7 +559,15 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
     };
 
     render() {
-        const { elementId, file, isDisabled = false, onVersionHistoryClick, getUserProfileUrl } = this.props;
+        const {
+            elementId,
+            file,
+            isDisabled = false,
```
booc0mtaco: I think this default is handled by `defaultProps` above, right?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
```diff
@@ -38,13 +38,44 @@
...
+                right: 0;
+                bottom: 0;
+                left: 0;
+                margin-right: -25px;
+                margin-left: -25px;
+                background-color: transparent;
+                animation: fade-out 6s ease-out;
+                content: '';
+
+                @keyframes fade-out {
```
booc0mtaco: We should use `$bdl-box-blue-05` here, since we don't want the focus color to change if the background changes, right?

src/elements/content-sidebar/activity-feed/activity-feed/__tests__/ActivityFeed-test.js
```diff
@@ -207,6 +199,30 @@ describe('elements/content-sidebar/ActivityFeed/activity-feed/ActivityFeed', ()
...
+    test('should scroll to active feed item', () => {
+        const wrapper = shallow(
+            <ActivityFeed
+                currentUser={currentUser}
+                feedItems={[{ type: 'comment' }]}
+                activeFeedEntryId={comments.entries[0].id}
+            />,
+        );
+        const instance = wrapper.instance();
+        const li = document.createElement('li');
```
booc0mtaco: I suppose we could suggest `mount`ing an `<li/>` instead of calling document directly, but i don't think this matters much in practice.

src/elements/content-sidebar/Sidebar.js
```diff
@@ -112,11 +119,13 @@ class Sidebar extends React.Component<Props, State> {
         const { file_version: currentVersion } = file;
         const fileVersionSlug = currentVersion ? `/${currentVersion.id}` : '';
 
+        const urlPrefix = this.getUrlPrefix(history.location.pathname);
```
jstoffan: Minor, but this is probably simple enough to inline: `const basePath = pathname.substring(1).split('/')[0];`

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
```diff
@@ -55,33 +61,49 @@ class ActivityFeed extends React.Component<Props, State> {
...
-        feedItems.length === 0 ||
-        (feedItems.length === 1 && feedItems[0].type === ItemTypes.fileVersion);
+    isEmpty = ({ feedItems }: Props = this.props): boolean => {
+        if (feedItems === undefined) {
+            return false;
+        }
+        return feedItems.length === 0 || (feedItems.length === 1 && feedItems[0].type === ItemTypes.fileVersion);
+    };
+
+    hasLoaded = ({ currentUser, feedItems }: Props = this.props): boolean =>
```
jstoffan: Minor: Should we make this `isLoading`, since that's how it seems to be used most often (in render)?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
```diff
@@ -1,4 +1,5 @@
 @import '../../../common/variables';
+$container-padding: 25px;
```
jstoffan: This will be declared as a global variable. Please move it inside its most relevant style rule, such as `.bcs-activity-feed-app-activity`, to make it local instead.

src/elements/content-sidebar/activity-feed/activity-feed/__tests__/ActiveState-test.js
```diff
@@ -120,4 +121,17 @@ describe('elements/content-sidebar/ActiveState/activity-feed/ActiveState', () =>
...
+        const wrapper = mount(
+            <ActiveState
+                items={[comment, fileVersion, taskWithAssignment, appActivity]}
+                currentUser={currentUser}
+                activeFeedEntryId={comment.id}
+                activeFeedEntryType={comment.type}
+            />,
+        );
+
+        expect(wrapper.find('.bcs-activity-feed-comment.bcs-is-focused')).toHaveLength(1);
```
jstoffan: Can we use a combination of data-testid and `hasClass` here?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
il10: @tjgupta  and I discussed it and we think sticking to having both hasLoaded and isLoading is ideal since we are using both within the file

src/elements/content-sidebar/activity-feed/activity-feed/__tests__/ActiveState-test.js
```diff
@@ -120,4 +121,18 @@ describe('elements/content-sidebar/ActiveState/activity-feed/ActiveState', () =>
...
+        const wrapper = mount(
+            <ActiveState
+                items={[comment, fileVersion, taskWithAssignment, appActivity]}
+                currentUser={currentUser}
+                activeFeedEntryId={comment.id}
+                activeFeedEntryType={comment.type}
+            />,
+        );
+
+        // expect(wrapper.find('.bcs-activity-feed-comment.bcs-is-focused')).toHaveLength(1);
```
jstoffan: We want to avoid commented code when possible.

src/elements/content-sidebar/ActivitySidebar.js
il10: It is defined by defaultProps, but when removed from render, ActivitySidebar fails to load. Not why this happens

PR #1626, opened 2019-09-20
chore(lint): unignore fixed rules

alexkrolick:
~~Ignored a few places that wanted to do reassignment through destructure, which looks like this:~~ changed the base config to ignore reassignment

```js
let { bar } = foo;
({ bar } = newFoo);
```


boxcla:

.eslintrc.js
```diff
@@ -7,7 +7,6 @@ module.exports = {
         'class-methods-use-this': 'off',
         'jsx-a11y/label-has-associated-control': 'off',
         'import/no-extraneous-dependencies': 'off', // fixme
-        'prefer-destructuring': 'off', // fixme
```
jstoffan: Can we limit it to enforcing object destructuring only? That way we can continue using `items[0]` notation.

alexkrolick: We can do that in `@box/frontend`, which this inherits from.

package.json
```diff
@@ -201,7 +201,7 @@
         "ilib-tree-node": "^1.2.0",
         "immutable": "^3.7.4",
         "intl": "^1.2.5",
-        "intl-messageformat": "^2.2.0",
+        "intl-messageformat": "^7.2.1",
```
jstoffan: Probably want to revert this line. I saw that this version was throwing jest errors previously, which is why I left it at 2.2.0.

alexkrolick: will do

PR #1629, opened 2019-09-23
fix(metadata): Show autocomplete for metadata input fields

ConradJChan:
By wrapping the input fields with a form and providing a name for each input field, we can take advantage of the browser autocomplete

Screen Shot 2019-10-01 at 1 54 38 PM


boxcla:

src/features/metadata-instance-editor/Footer.js
```diff
@@ -12,10 +12,10 @@ import './Footer.scss';
 type Props = {
     onCancel: () => void,
     onRemove: () => void,
-    onSave?: () => void,
+    showSave: Boolean,
```
sahiga: `boolean`?

PR #1631, opened 2019-09-23
fix(versions): Improve experience for restored vs promoted versions

jstoffan:
**Activity Sidebar**
activity-sidebar

**Versions Sidebar**
versions-sidebar


boxcla:

src/elements/content-sidebar/versions/VersionsItem.js
```diff
@@ -35,6 +40,7 @@ type Props = {
 const ACTION_MAP = {
```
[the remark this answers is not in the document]
jstoffan: I don't think so. Each component maps each action to a different message, in most cases.

src/elements/content-sidebar/versions/__tests__/__snapshots__/VersionsItem-test.js.snap
```diff
@@ -32,6 +32,7 @@ exports[`elements/content-sidebar/versions/VersionsItem render should show actio
           values={
             Object {
               "name": "Test User",
+              "versionPromoted": undefined,
```
mickr: Should these be in the snap / dom if they are undefined?

jstoffan: Yeah, since I'm not filtering them out of the `values` object. It should be functionally equivalent, though.

PR #1633, opened 2019-09-24
chore(version): Remove version history modal

mxiao6:


boxcla:

ConradJChan:
Instead of `perf` should it be `chore`?

PR #1635, opened 2019-09-24
refactor(avatar): Convert Avatar to hooks implementation

j3tan:
Refactoring `Avatar` to use hooks (take 2). The first iteration worked fine but due to a few libraries using their own version of React (and also during node_module linking), this was reverted.

Now with `yarn resolutions`, we can avoid having duplicate React instances and multiple versions of react.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -28,15 +28,17 @@ A published version of the style guide can be seen at https://opensource.box.com
 
 ### Webpack Setup
 
-`box-ui-elements` must use the same `react` and `react-dom` instances as the parent application for React hooks to work properly. Application repositories must add the following webpack `resolve` configuration to satify this requirement:
+`box-ui-elements` must use the same `react` and `react-dom` instances as the parent application for React hooks to work properly. Application repositories must add the following webpack `resolve` alias configuration to satify this requirement:
 
 ```js
 // webpack.config.js
 {
     ...
     resolve: {
-        'react': path.resolve('node_modules/react'),
-        'react-dom': path.resolve('node_modules/react-dom'),
+        alias: {
+            'react': path.resolve('node_modules/react'),
+            'react-dom': path.resolve('node_modules/react-dom'),
+        }
     }
     ...
 }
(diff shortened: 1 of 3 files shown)


boxcla:

src/components/avatar/Avatar.js
```diff
@@ -1,5 +1,5 @@
 // @flow
-import * as React from 'react';
+import React, { useState, useEffect } from 'react';
```
jstoffan: I think we've been tending toward the `*` import style and referencing things directly off the `React` alias.

j3tan: Hmm, would that mean React.useState and React.useEffect? Personally, I prefer it without the React identifier. I forget, but is there a reason we do `import * as React` instead of just `import React`? They seem equivalent based on this file (https://github.com/facebook/react/blob/master/packages/react/src/React.js)

alexkrolick: `import * as React` is to bring in all the Flow types

j3tan:
Usage should be relatively straight forward since `react` and `react-dom` are peer dependencies but I will put a recommendation into the README to use `yarn resolutions` in applications. There is no built-in equivalent for native `npm` users though

j3tan:
I've also verified that this components works when `box-ui-elements` is `yarn link`ed in an application repository with `yarn resolutions`

alexkrolick:
@j3tan probably as important as the yarn resolution is a webpack resolver alias, which would work with npm. The `yarn resolution` just makes that eventual webpack resolution less surprising.

DEVELOPING.md
```diff
@@ -26,6 +26,40 @@ A published version of the style guide can be seen at https://opensource.box.com
...
 
+### Webpack Setup
+
+`box-ui-elements` must use the same `react` and `react-dom` instances as the parent application for React hooks to work properly. Application repositories must add the following webpack `resolve` configuration to satify this requirement:
+
+```js
+// webpack.config.js
+{
+    ...
+    resolve: {
```
alexkrolick: this should be 

```js
resolve: {
  alias: {
    'react': path.resolve('node_modules/react'),
    'react-dom': path.resolve('node_modules/react-dom'),
  }
}
```

PR #1636, opened 2019-09-24
feat(icon): Create users empty state icon

indiramuru:
Screen Shot 2019-09-23 at 11 47 50 PM


boxcla:

booc0mtaco:
We don't normally specify an opacity as a prop to the icon, because of some artifacts that can appear (note how the overlapping strokes in the screenshot above appear darker in some instances). instead, we control this via CSS in the consuming app, so the whole svg gets the same opacity for all strokes and fills.

I'd remove that prop and the usages. Otherwise looks good. Might need one of those optimizers to clean out the ID properties as well.

PR #1637, opened 2019-09-24
fix(annotation): Fix header by adding selector used by box-annotations

mxiao6:


boxcla:

jstoffan:
Hm, this was my bad. This is partial revert of some of my changes in #1484.

src/elements/content-preview/preview-header/PreviewHeader.js
```diff
@@ -70,74 +70,80 @@ const PreviewHeader = ({
                 'bcpr-PreviewHeader--basic': !isPreviewingCurrentVersion,
             })}
         >
-            <FileInfo file={file} version={selectedVersion} />
+            <div className="bp-header bp-base-header">
```
jstoffan: Can we add a JSX comment here about this class/element being used by box-annotations? Hopefully that will help avoid this issue from reoccurring.

src/elements/content-preview/preview-header/PreviewHeader.scss
```diff
@@ -5,11 +5,21 @@
...
     background-color: $bdl-gray-02;
     border-bottom: 1px solid $bdl-gray-10;
 
+    // Box Annotations selector
+    .bp-header {
+        flex: 1;
+        height: 60px;
+    }
+
+    .bp-base-header {
```
jstoffan: This isn't a flex container like its parent, which seems like it will cause some visual issues. Can you test this with the OpenWith button and post screenshots of before/after?

src/elements/content-preview/preview-header/PreviewHeader.js
jstoffan: Let's also add a unit test for it so we can make the expectation explicit.

src/elements/content-preview/preview-header/PreviewHeader.scss
```diff
@@ -3,12 +3,23 @@
...
+        align-items: center;
+        justify-content: space-between;
+        padding: 0 15px;
+        background-color: $bdl-gray-02;
+        border-bottom: 1px solid $bdl-gray-10;
+    }
+
+    // Box Annotations selector
+    .bp-header {
+        flex: 1;
```
jstoffan: Make sure to test IE11, Edge, and Safari in particular with these changes.

PR #1638, opened 2019-09-24
feat(icon): create add users empty state icon

indiramuru:
Screen Shot 2019-09-24 at 7 18 27 PM


boxcla:

src/icons/states/UsersEmptyState.js
```diff
@@ -0,0 +1,125 @@
...
+
+const UsersEmptyState = ({ className, color = bdlBoxBlue, height = 135, opacity, title, width = 130 }: Props) => (
+    <AccessibleSVG
+        className={classNames('users-empty-state', className)}
+        height={height}
+        title={title}
+        viewBox="0 0 130 135"
+        width={width}
+    >
+        <g fill="none" fillRule="evenodd">
```
priyajeet: This `<g>` wrapper is likely redundant.

indiramuru:
Review comments from @booc0mtaco 
We don't normally specify an opacity as a prop to the icon, because of some artifacts that can appear (note how the overlapping strokes in the screenshot above appear darker in some instances). instead, we control this via CSS in the consuming app, so the whole svg gets the same opacity for all strokes and fills.
I'd remove that prop and the usages. Otherwise looks good. Might need one of those optimizers to clean out the ID properties as well.

src/icons/states/UsersEmptyState.js
```diff
@@ -0,0 +1,114 @@
...
+                cy="70"
+                fill={white}
+                r="8"
+                stroke={color}
+                strokeLinecap="round"
+                strokeLinejoin="round"
+                strokeWidth="2"
+            />
+            <path
+                className="fill"
```
priyajeet: fill-color

src/icons/states/UsersEmptyState.js
indiramuru: removing <g>  adds a weird shade to the icon
Screen Shot 2019-09-24 at 3 12 06 PM

PR #1639, opened 2019-09-24
feat(icon): create users empty state icon

indiramuru:


boxcla:

PR #1643, opened 2019-09-25
fix(avatar): use derived state instead of effect for url change

alexkrolick:
- https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops
- changing state as a result of props changed should follow the
  derived state pattern instead of enqueing an update after rendering with
  the old state + new props


boxcla:

j3tan:
looks like there's a flow error with undefined. Also, if this is the case, do we need to use `act()` in the test?

alexkrolick:
@jtan 

- [x] flow
- [x] without `act` there is this console warning

Screen Shot 2019-09-25 at 13 53 02

PR #1646, opened 2019-09-26
fix(uploads): Fix contentUploader and metadataInstance

mickr:
Removed logic gate from newly implemented componentDidUpdate in contentUploader
Fix comparison in metadata instance


boxcla:

PR #1647, opened 2019-09-26
fix(unfied-share-modal): shared link access levels

priyajeet:
When creating shared links, this changes the access level drop down to show menu items in a disabled state if disabled by an access policy. Before such happened just based on the fact that some classification existed or not, which is incorrect since you can just have classifications without any applicable policies.

Additionally a small cleanup/refactor of the menu generation.

When access levels are disabled due to a policy
Screen Shot 2019-09-26 at 1 33 43 PM

When access levels are disabled but due to other reasons
Screen Shot 2019-09-26 at 1 31 05 PM

### DIFF
--- a/src/features/unified-share-modal/__tests__/SharedLinkAccessMenu.test.js
+++ b/src/features/unified-share-modal/__tests__/SharedLinkAccessMenu.test.js
@@ -31,7 +31,7 @@ describe('features/unified-share-modal/SharedLinkAccessMenu', () => {
                 submitting: false,
             },
         ].forEach(({ submitting }) => {
-            test('should render correct menu', () => {
+            test(`should render correct menu when submitting is ${submitting}`, () => {
                 const sharedLinkAccessMenu = getWrapper({ submitting });
                 expect(sharedLinkAccessMenu).toMatchSnapshot();
             });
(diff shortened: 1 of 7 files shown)


boxcla:

src/features/unified-share-modal/__tests__/SharedLinkAccessMenu-test.js
```diff
@@ -31,7 +31,7 @@ describe('features/unified-share-modal/SharedLinkAccessMenu', () => {
                 submitting: false,
             },
         ].forEach(({ submitting }) => {
-            test('should render correct menu', () => {
+            test(`should render correct menu when submitting is ${submitting}`, () => {
```
Sunyang730: Can we not use test.each here?

src/features/unified-share-modal/flowTypes.js
```diff
@@ -29,6 +29,12 @@ export type allowedAccessLevelsType = {
     peopleWithTheLink?: boolean,
 };
 
+export type accessLevelsDisabledReasonType = {
+    peopleInThisItem?: 'access_policy' | null,
```
Sunyang730: Is this not the same as `peopleInThisItem?: ?'access_policy'`

priyajeet: Yeah I guess I can change to that style

Sunyang730: That would be the same as `?('foo' | 'bar')` no?
If we want to do null, it will be changing the style and I would prefer to keep things consistent. Unless `?` is not something people commonly use?

src/features/unified-share-modal/__tests__/SharedLinkAccessMenu.test.js
```diff
@@ -31,7 +31,7 @@ describe('features/unified-share-modal/SharedLinkAccessMenu', () => {
                 submitting: false,
             },
         ].forEach(({ submitting }) => {
-            test('should render correct menu', () => {
+            test(`should render correct menu when submitting is ${submitting}`, () => {
```
Sunyang730: Still thinks this is weird to not use `test.each`.

src/features/unified-share-modal/__tests__/SharedLinkAccessMenu.test.js
```diff
@@ -23,24 +23,45 @@ describe('features/unified-share-modal/SharedLinkAccessMenu', () => {
...
+            const sharedLinkAccessMenu = getWrapper({ submitting });
+            expect(sharedLinkAccessMenu).toMatchSnapshot();
         });
 
         test('should render tooltipContent if provided', () => {
             const sharedLinkAccessMenu = getWrapper({ tooltipContent: 'Hello, world!' });
             expect(sharedLinkAccessMenu).toMatchSnapshot();
         });
+
+        test('should render no access level menu items if disabled by something other than access policy', () => {
```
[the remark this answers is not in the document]
priyajeet: The 1st one above

PR #1649, opened 2019-09-27
fix(sidebar): Fix deeplinking for version history sidebar

jstoffan:
I guess regex + optional parameters don't work.


boxcla:

PR #1652, opened 2019-09-30
fix(content-uploader): Fix duplicate files upload at the same time

### LINKED ISSUE #1433: Duplicate file upload in ContentUploader
Please fill out the following template so we can reproduce and fix your issue as quickly as possible!

**Note:** if your issue includes a potential security vulnerability, please do not file it here. Instead, email the issue to security@box.com for support.

**Environment:**
Included 'box-ui-element' in React application and added ContentUploader to upload files to box cloud.

- Elements version:

**Desktop (please complete the following information):**

- OS: Windows 10
- Browser: All browsers
- Version: latest

**Steps to reproduce the problem:**
Select a file with browse functionality and and then Drag-Drop the same file in Content Uploader, both files can be seen in queue.

**What is the expected behavior? (Screenshots can be helpful here)**
If user selects a file with browse functionality and then drag-drops the same file, upload should be restricted, it will be helpful if user is shown a info message "File in Queue."  

**What went wrong? (Screenshots, console logs, or HAR files can be helpful here)**

uploader-issue

mxiao6:
Fix #1433


boxcla:

jstoffan:
This is a critical path of a heavily used component. Are we confident there are no behavioral regressions? Can we add tests that fail prior to this change and then pass after it, validating the new behavior?

src/elements/content-uploader/ContentUploader.js
```diff
@@ -304,26 +304,27 @@ class ContentUploader extends Component<Props, State> {
...
+                    ...state.itemIds,
+                    ...newItemIds,
+                },
+            }),
+            () => {
+                if (firstFile.webkitRelativePath && !isRelativePathIgnored) {
+                    // webkitRelativePath should be ignored when the upload destination folder is known
+                    this.addFilesWithRelativePathToQueue(newFiles, itemUpdateCallback);
+                } else {
+                    onBeforeUpload(newFiles);
```
mickr: Doesn't this function needed to be executed in either case, since it is a hook or renderProp that is passed into this component?

mxiao6: I agree your idea but the logic of the original code is like my code now. I think it might be a mistake in the original code.

PR #1653, opened 2019-10-01
fix(content-uploader): Fix cancelling task triggers the others to start

mxiao6:


boxcla:

src/elements/content-uploader/ContentUploader.js
```diff
@@ -628,6 +628,7 @@ class ContentUploader extends Component<Props, State> {
      */
     removeFileFromUploadQueue = (item: UploadItem) => {
         const { onCancel, useUploadsManager } = this.props;
+        const { view } = this.state;
```
ConradJChan: Can this be combined with L638?

src/elements/content-uploader/ContentUploader.js
```diff
@@ -628,21 +628,24 @@ class ContentUploader extends Component<Props, State> {
...
         // Minimize uploads manager if there are no more items
         const callback = useUploadsManager && !items.length ? this.minimizeUploadsManager : noop;
 
         onCancel([item]);
         this.updateViewAndCollection(items, callback);
-        this.upload();
+
+        if (view === VIEW_UPLOAD_IN_PROGRESS) {
+            this.upload();
+        }
```
ConradJChan: would it make sense to add this guard check for other calls to `this.upload()`? Specifically, `handleUploadSuccess` and `handleUploadError`?

src/elements/content-uploader/ContentUploader.js
```diff
@@ -580,12 +580,16 @@ class ContentUploader extends Component<Props, State> {
...
+            if (itemUpdateCallback) {
+                itemUpdateCallback();
+            }
 
-        // Automatically start upload if other files are being uploaded
-        if (view === VIEW_UPLOAD_IN_PROGRESS) {
-            this.upload();
-        }
+            // Automatically start upload if other files are being uploaded
+            if (view === VIEW_UPLOAD_IN_PROGRESS) {
```
ConradJChan: In these cases, where we have a callback to setState, do we need to reference the new `view` value instead?

mxiao6: I think we need to, because for example, at the moment a file is added to the queue, another task failed, then we want to pause the uploading instead of continuing.

src/elements/content-uploader/ContentUploader.js
```diff
@@ -628,21 +632,26 @@ class ContentUploader extends Component<Props, State> {
...
         onCancel([item]);
-        this.updateViewAndCollection(items, callback);
-        this.upload();
+        this.updateViewAndCollection(items, () => {
+            // Minimize uploads manager if there are no more items
+            if (useUploadsManager && !items.length) {
+                this.minimizeUploadsManager();
+            }
+
+            if (view === VIEW_UPLOAD_IN_PROGRESS) {
```
mickr: It's not clear to me which state `view` should be referencing since this callback is passed as the callback function to `setState()`

src/elements/content-uploader/ContentUploader.js
```diff
@@ -808,8 +817,11 @@ class ContentUploader extends Component<Props, State> {
             onUpload(item.boxFile);
         }
 
-        this.updateViewAndCollection(items);
-        this.upload();
+        this.updateViewAndCollection(items, () => {
+            if (view === VIEW_UPLOAD_IN_PROGRESS) {
+                this.upload();
+            }
```
mickr: Same here, would this view be the most current view since it is being called AFTER the render?

src/elements/content-uploader/__tests__/ContentUploader-test.js
```diff
@@ -81,26 +84,47 @@ describe('elements/content-uploader/ContentUploader', () => {
...
+        test('should cancel and remove item from uploading queue', () => {
             instance.removeFileFromUploadQueue(item);
 
             expect(item.api.cancel).toBeCalled();
             expect(wrapper.state().items.length).toBe(0);
-            expect(instance.upload).toBeCalled();
         });
+
+        test.each([VIEW_ERROR, VIEW_UPLOAD_EMPTY, VIEW_UPLOAD_IN_PROGRESS, VIEW_UPLOAD_SUCCESS])(
+            'should continue uploading if other files are being uploaded',
```
mickr: Can you use interpolation here to make the test cases unique while still being generated from `test.each`

src/elements/content-uploader/ContentUploader.js
```diff
@@ -580,12 +580,16 @@ class ContentUploader extends Component<Props, State> {
             }
         }
 
-        this.updateViewAndCollection(updatedItems, itemUpdateCallback);
+        this.updateViewAndCollection(updatedItems, () => {
+            if (itemUpdateCallback) {
+                itemUpdateCallback();
```
mickr: I'm not sure if this is needed, but if we are being more defensive around `itemUpdateCallback` you may also want to ensure that it is a function as well before calling it.

mxiao6: The flow type of it is `Function` so I think it can only be a real function or `undefined`.

src/elements/content-uploader/ContentUploader.js
mxiao6: Good question... I'm going to make it clearer

src/elements/content-uploader/ContentUploader.js
mickr: See my other comment about the context.

PR #1654, opened 2019-10-01
fix(arrow-stepper): Fix arrow stepper in KeyBinding component

mickr:
Move from `componentDidUpdate` to use the  `getDerivedStateFromProps` function. This also adds tests for the arrow keys and prop updates.

### DIFF
--- a/src/elements/common/__tests__/KeyBinder.js
+++ b/src/elements/common/__tests__/KeyBinder.js
@@ -1,0 +1,66 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import KeyBinder from '../KeyBinder';
+
+describe('KeyBinder', () => {
+    let onScrollToChangeMock;
+    let getEvent;
+    let wrapper;
+
+    beforeEach(() => {
+        onScrollToChangeMock = jest.fn();
+        wrapper = shallow(
+            <KeyBinder
+                id="123"
+                onScrollToChange={onScrollToChangeMock}
+                rowCount={10}
+                items={[]}
+                columnCount={10}
+                scrollToRow={0}
+                scrollToColumn={0}
+            >
+                {() => {}}
+            </KeyBinder>,
+        );
+        getEvent = type => {
+            return {
+                key: type,
+                stopPropagation: jest.fn(),
+                preventDefault: jest.fn(),
+            };
+        };
+    });
+
+    test('it should update scrollToRow when props change', () => {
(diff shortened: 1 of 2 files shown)


boxcla:

src/elements/common/__tests__/KeyBinder.js
```diff
@@ -0,0 +1,66 @@
...
+        ${'ArrowDown'}  | ${'scrollToRow'}    | ${8}
+        ${'ArrowLeft'}  | ${'scrollToColumn'} | ${2}
+        ${'ArrowRight'} | ${'scrollToColumn'} | ${8}
+    `('Arrow Handlers', ({ eventType, scrollType, result }) => {
+        const event = getEvent(eventType);
+
+        wrapper.setProps({ [`${scrollType}`]: 5 });
+        wrapper.simulate('keyDown', event);
+        wrapper.simulate('keyDown', event);
+        wrapper.simulate('keyDown', event);
```
jstoffan: I think the current guidance is to avoid `.simulate` and just call the prop callback directly.

src/elements/common/__tests__/KeyBinder.js
```diff
@@ -0,0 +1,66 @@
...
+        expect(wrapper.state('scrollToRow')).toEqual(5);
+    });
+
+    test.each`
+        eventType       | scrollType          | result
+        ${'ArrowUp'}    | ${'scrollToRow'}    | ${2}
+        ${'ArrowDown'}  | ${'scrollToRow'}    | ${8}
+        ${'ArrowLeft'}  | ${'scrollToColumn'} | ${2}
+        ${'ArrowRight'} | ${'scrollToColumn'} | ${8}
+    `('Arrow Handlers', ({ eventType, scrollType, result }) => {
```
jstoffan: Can we improve this test description and make it more like a sentence?

src/elements/common/__tests__/KeyBinder.js
```diff
@@ -0,0 +1,66 @@
...
+import KeyBinder from '../KeyBinder';
+
+describe('KeyBinder', () => {
+    let onScrollToChangeMock;
+    let getEvent;
+    let wrapper;
+
+    beforeEach(() => {
+        onScrollToChangeMock = jest.fn();
+        wrapper = shallow(
```
jstoffan: We usually use a `getWrapper` helper and invoke it per test rather than relying on a global mutable variable.

PR #1657, opened 2019-10-02
fix(breadcrumb): add missing import

frogzsj:
-Fix flow issue of missing import


boxcla:

src/components/breadcrumb/Breadcrumb.js
```diff
@@ -1,6 +1,7 @@
 // @flow
 import * as React from 'react';
 import { injectIntl } from 'react-intl';
```
jstoffan: I think this can be simplified: `import { injectIntl, type InjectIntlProvidedProps } from 'react-intl';`

PR #1659, opened 2019-10-03
fix(subheader): Add logic around grid view

mickr:
This change removes the view change (grid view) button unless there is a provided `gridColumnCount` prop. Content Picker was showing the grid view button when it doesn't support grid view.


boxcla:

mickr:
Adding it now

PR #1662, opened 2019-10-04
fix(versions): Update version sidebar to use latest retention schema

jstoffan:
Note that there are some outstanding questions around indefinite retention and how it will be surfaced from the API. We'll be following-up with the backend team next week to hammer out the final details.


boxcla:

mxiao6:
image
image
image

flow-typed/box-ui-elements.js
```diff
@@ -178,10 +181,10 @@ type BoxItemVersionRetention = {
 };
 
 type BoxItemVersionRetentionPolicy = {
-    disposition_action: 'permanently_delete' | 'remove_retention',
+    disposition_action: VERSION_RETENTION_DELETE_ACTION | VERSION_RETENTION_REMOVE_ACTION,
```
ConradJChan: I think these should be `typeof VERSION_RETENTION_DELETE_ACTION` etc...

PR #1663, opened 2019-10-07
feat(security): adding SecurityBadge component

jpfranco:
This update will move out the actual badge markup from the `ClassifiedBadge` to a slightly more re-usable `SecurityBadge` component. This will allow us to use the same look and feel of classification with other security-related badges.

Screen Shot 2019-10-07 at 4 39 50 PM

Screen Shot 2019-10-07 at 4 41 02 PM

### DIFF
--- a/src/features/classification/ClassifiedBadge.scss
+++ b/src/features/classification/ClassifiedBadge.scss
@@ -1,5 +1,0 @@
-.bdl-ClassifiedBadge {
-    .bdl-SecurityBadge-name {
-        margin-left: 5px;
-    }
-}
--- a/src/features/classification/ClassifiedBadge.scss
+++ b/src/features/classification/ClassifiedBadge.scss
@@ -1,17 +1,5 @@
-@import '../../styles/variables';
-@import './mixins';
-
-.bdl-ClassifiedBadge,
-.bdl-ClassifiedBadge:hover {
-    @include bdl-ClassificationBadge;
-
-    background-color: $bdl-yellorange-10;
-    border-color: $bdl-yellorange;
-}
-
-.bdl-ClassifiedBadge-name {
-    @include bdl-ClassificationBadge-name;
-
-    color: $bdl-gray !important; // To inhibit overriding
-    text-transform: uppercase;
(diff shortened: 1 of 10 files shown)


boxcla:

src/features/security/index.js
```diff
@@ -0,0 +1,3 @@
+// @flow
+// eslint-disable-next-line import/prefer-default-export
+export { default as SecurityBadge } from './SecurityBadge';
```
priyajeet: export { default } from './SecurityBadge';

src/features/classification/ClassifiedBadge.scss
```diff
@@ -1,17 +1,5 @@
...
-}
-
-.bdl-ClassifiedBadge-name {
-    @include bdl-ClassificationBadge-name;
-
-    color: $bdl-gray !important; // To inhibit overriding
-    text-transform: uppercase;
+.bdl-ClassifiedBadge {
+    .bdl-SecurityBadge-name {
+        margin-left: 5px;
```
priyajeet: Is this margin not needed for space between icon and name?

src/features/security/SecurityBadge.js
```diff
@@ -0,0 +1,26 @@
...
+import './SecurityBadge.scss';
+
+type Props = {
+    className?: string,
+    icon?: React.Node,
+    message: string,
+};
+
+const SecurityBadge = ({ className, icon, message, ...rest }: Props) => {
+    const iconElement = icon || <IconAlertDefault color={bdlYellorange} height={22} width={22} strokeWidth={3} />;
```
priyajeet: maybe add it as defaultProp instead?

src/features/classification/ClassifiedBadge.scss
jpfranco: Couldn't find a good solution for this. The icons seems to have inconsistent whitespace around them (`IconAlertDefault` has a lot of it), so you basically have to tweak the margin manually for each. An alternative  would be to set a negative margin on `IconAlertDefault` so that all badges have that 5px spacing by default.

PR #1665, opened 2019-10-08
fix(content-preview): remove deeplink path upon navigation

il10:


boxcla:

src/elements/content-preview/PreviewNavigation.js
```diff
@@ -14,11 +17,19 @@ import messages from '../common/messages';
 type Props = {
     collection: Array<string | BoxItem>,
     currentIndex: number,
+    history: RouterHistory,
     onNavigateLeft: Function,
     onNavigateRight: Function,
 } & InjectIntlProvidedProps;
```
il10: Observation: Since InjectIntlProvidedProps is not imported in this file then it resolves to an `any(implicit)` type.  When it is imported, flow resolves the type properly which causes flow errors for the history prop in the ContentPreview component.

alexkrolick: We need to fix this in this PR, the current `any` type is just hiding the issue

src/elements/content-preview/PreviewNavigation.js
```diff
@@ -27,28 +38,45 @@ const PreviewNavigation = ({ collection = [], currentIndex, intl, onNavigateLeft
...
-                    title={intl.formatMessage(messages.nextFile)}
-                    type="button"
-                >
-                    <IconNavigateRight />
-                </PlainButton>
-            )}
-        </>
+        <NavRouter history={history}>
+            <Route path={['/:activeTab/:deeplink', '/']}>
+                {({ match }) => (
```
alexkrolick: I think what we can do here is extract the `history` prop from the `Route` component:

https://reacttraining.com/react-router/web/api/history

```suggestion
{({ match, history }) => (
```

That will let us use the fresh history object created by the NavRouter's MemoryRouter when the `history` prop is `undefined`

src/elements/content-preview/PreviewNavigation.js
```diff
@@ -27,28 +38,46 @@ const PreviewNavigation = ({ collection = [], currentIndex, intl, onNavigateLeft
...
+                    <>
+                        {/* {console.log(match)} */}
```
mxiao6: Can you remove this `console.log`? Do we still need the Fragments here?

src/elements/content-preview/ContentPreview.js
```diff
@@ -1191,6 +1191,7 @@ class ContentPreview extends React.PureComponent<Props, State> {
                                 currentIndex={this.getFileIndex()}
                                 onNavigateLeft={this.navigateLeft}
                                 onNavigateRight={this.navigateRight}
+                                history={history}
```
mxiao6: Can we keep these props alphabetized?

src/elements/content-preview/PreviewNavigation.js
```diff
@@ -27,28 +38,45 @@ const PreviewNavigation = ({ collection = [], currentIndex, intl, onNavigateLeft
...
+            <Route path={['/:activeTab/:deeplink', '/']}>
+                {({ match }) => (
+                    <>
+                        {hasLeftNavigation && (
+                            <PlainButton
+                                className="bcpr-navigate-left"
+                                onClick={() => {
+                                    if (match.params.deeplink) {
+                                        history.push(`/${match.params.activeTab}`);
+                                    }
```
mxiao6: Due to the added complexity of deeplink, you can extract the `PlainButton` to form a new component like `NavigationButton`.

src/elements/content-preview/PreviewNavigation.js
```diff
@@ -27,28 +38,45 @@ const PreviewNavigation = ({ collection = [], currentIndex, intl, onNavigateLeft
...
+                                    onNavigateLeft();
```
ConradJChan: Could this logic be reused for both `onNavigateLeft` and `onNavigateRight`?

src/elements/content-preview/PreviewNavigation.js
```diff
@@ -27,28 +38,45 @@ const PreviewNavigation = ({ collection = [], currentIndex, intl, onNavigateLeft
...
-                    onClick={onNavigateRight}
-                    title={intl.formatMessage(messages.nextFile)}
-                    type="button"
-                >
-                    <IconNavigateRight />
-                </PlainButton>
-            )}
-        </>
+        <NavRouter history={history}>
+            <Route path={['/:activeTab/:deeplink', '/']}>
```
ConradJChan: Are these paths comprehensive enough to ensure that these navigation buttons always render? Should `'/'` be `'*'`?

ConradJChan: Nvm, just read that `'/'` will always match the URL

src/elements/content-preview/PreviewNavigation.js
il10: I discussed it with the team and since the only piece really being added is the `if`, we think it's simple enough to leave it as is.

ConradJChan:
In a follow up PR, could you add a E2E test that captures this functionality?

il10:
> In a follow up PR, could you add a E2E test that captures this functionality?

Ok, we can add a test to the parent app

PR #1666, opened 2019-10-08
feat(metadata-views): Update metadata columns prop type in metadata view

manishPh:


boxcla:

src/features/metadata-based-view/MetadataBasedItemList.js
```diff
@@ -58,6 +60,10 @@ class MetadataBasedItemList extends React.Component<Props, State> {
         };
     }
 
+    getMetadataColumnName(column: MetadataColumnConfig | string): string {
+        return typeof column === 'string' ? column : getProp(column, 'name');
```
mxiao6: Do we need to add a `defaultValue` for `getProp`?

manishPh: These are columns passed by the parent app. We don't know what could be the default value.

PR #1667, opened 2019-10-11
feat(metadata-views): Show edit icon in metadata view

manishPh:


boxcla:

src/features/metadata-based-view/MetadataBasedItemList.js
```diff
@@ -117,10 +128,19 @@ class MetadataBasedItemList extends React.Component<Props, State> {
                 break;
             default: {
                 const data = getProp(item, 'metadata.data', {});
-                const mdFieldName = this.getMetadataColumnName(
-                    metadataColumnsToShow[columnIndex - FIXED_COLUMNS_NUMBER],
+                const mdFieldName = this.getMetadataColumnName(metadataColumn);
```
ConradJChan: Could we get the cell data in one shot using `getProp` and string interpolation with the `mFieldName`?

src/features/metadata-based-view/MetadataBasedItemList.js
```diff
@@ -143,9 +163,17 @@ class MetadataBasedItemList extends React.Component<Props, State> {
...
 
-    handleMouseEnter = (rowIndex: number): void => this.setState({ hoveredRowIndex: rowIndex });
+    handleMouseEnter = (columnIndex: number, rowIndex: number): void =>
+        this.setState({
+            hoveredColumnIndex: columnIndex,
+            hoveredRowIndex: rowIndex,
+        });
 
-    handleMouseLeave = (): void => this.setState({ hoveredRowIndex: -1 });
+    handleMouseLeave = (): void =>
```
ConradJChan: Would it make sense to debounce this leave handler or is React's setState efficient enough that it will batch the changes to state based on leave and enter events so that no extra render happens?

src/features/metadata-based-view/MetadataBasedItemList.js
```diff
@@ -168,7 +196,7 @@ class MetadataBasedItemList extends React.Component<Props, State> {
                 className={classes}
```
ConradJChan: Unrelated to the changeset, but do we want to set the `'bdl-MetadataBasedItemList-cell--hover'` class even if it's a header row?

src/features/metadata-based-view/MetadataBasedItemList.js
mickr: React event handlers are batched by default - https://stackoverflow.com/questions/48563650/does-react-keep-the-order-for-state-updates/48610973#48610973

src/features/metadata-based-view/MetadataBasedItemList.js
manishPh: no. So far, this class is only used for adding background color, and we don't need bgColor for header row.

src/features/metadata-based-view/MetadataBasedItemList.js
manishPh: As discussed on Slack, `mdFieldName` is column name. The same function is used to get header value so returning cell data from the function using `getProp` is not useful here.

PR #1671, opened 2019-10-16
fix(select-field): Enable arrow scrolling in select fields

sahiga:
This PR sets a maximum height for the `SelectField` component, which enables arrow key scrolling in dropdown menus.

**Chrome**
buie-arrow-scroll-in-select-fields-chrome

**Firefox**
buie-arrow-scroll-in-select-fields-firefox

**Safari**
buie-arrow-scroll-in-select-fields-safari

**IE11**
buie-arrow-scroll-in-select-fields-ie11
 
**In a large viewport**
buie-arrow-scroll-in-select-fields-large-viewport

**In a small viewport**
buie-arrow-scroll-in-select-fields-small-viewport

**With fields above the expanded field**
buie-arrow-scroll-in-select-fields-third-field

### DIFF
--- a/src/components/select-field/BaseSelectField.js
+++ b/src/components/select-field/BaseSelectField.js
@@ -38,6 +38,8 @@ type Props = {
     error?: React.Node,
     /** The select button is disabled if true */
     isDisabled?: boolean,
+    /** The select field overlay (dropdown) will have a scrollbar and max-height if true * */
+    isScrollable?: boolean,
     multiple: boolean,
     /** Function will be called with an array of all selected options after user selects a new option */
     onChange: Function,
@@ -49,8 +51,6 @@ type Props = {
     options: Array<SelectOptionProp>,
     /** The select button text shown when no options are selected. */
     placeholder?: string | React.Element<any>,
-    /** The select field overlay (dropdown) will have a scrollbar and max-height if true * */
-    scrollable?: boolean,
     /** The currently selected option values (can be empty) */
     selectedValues: Array<SelectOptionValueProp>,
     /** Array of ordered indices indicating where to insert separators (ex. index 2 means insert a separator after option 2) */
@@ -72,9 +72,9 @@ class BaseSelectField extends React.Component<Props, State> {
     static defaultProps = {
         buttonProps: {},
         isDisabled: false,
+        isScrollable: false,
         multiple: false,
         options: [],
-        scrollable: false,
         selectedValues: [],
         separatorIndices: [],
     };
@@ -366,7 +366,7 @@ class BaseSelectField extends React.Component<Props, State> {
     };
 
     render() {
-        const { className, multiple, scrollable } = this.props;
+        const { className, multiple, isScrollable } = this.props;
         const { isOpen } = this.state;
 
         // @TODO: Need invariants on specific conditions.
@@ -397,7 +397,7 @@ class BaseSelectField extends React.Component<Props, State> {
                     >
                         <ul
                             className={classNames('overlay', {
-                                [OVERLAY_SCROLLABLE_CLASS]: scrollable,
+                                [OVERLAY_SCROLLABLE_CLASS]: isScrollable,
                             })}
                             id={this.selectFieldID}
                             role="listbox"
--- a/src/components/select-field/BaseSelectField.js
+++ b/src/components/select-field/BaseSelectField.js
@@ -8,7 +8,7 @@ import IconCheck from '../../icons/general/IconCheck';
 import SelectButton from '../select-button';
 import DatalistItem from '../datalist-item';
 import type { SelectOptionValueProp, SelectOptionProp } from './props';
-import { OVERLAY_WRAPPER_CLASS } from '../../constants';
+import { BASE_SELECT_FIELD_OFFSET_TOP_PADDING, OVERLAY_WRAPPER_CLASS } from '../../constants';
 
 import './SelectField.scss';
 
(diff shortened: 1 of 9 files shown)


boxcla:

src/components/select-field/BaseSelectField.js
```diff
@@ -85,6 +85,24 @@ class BaseSelectField extends React.Component<Props, State> {
             isOpen: false,
             shouldScrollIntoView: false,
         };
+
+        this.fieldRef = React.createRef();
```
jstoffan: This field should be defined on the class along with an accompanying flow type.

src/components/select-field/BaseSelectField.js
```diff
@@ -85,6 +85,24 @@ class BaseSelectField extends React.Component<Props, State> {
...
+
+    /**
+     * Calculate this field's distance from the top.
+     * This allows us to create an overflow container for the field's dropdown menu that is navigable using both
+     * the mouse and arrow keys and will always fit within the current viewport.
+     */
+    calculateOffsetTopValue() {
+        this.offsetTopValue =
+            window.pageYOffset +
+            this.fieldRef.current.getBoundingClientRect().top +
```
jstoffan: We should be checking to make sure that `this.fieldRef.current` exists before operating on it. You can use destructured renaming to do so:

```
const { current: fieldRef } = this.fieldRef;

if (!fieldRef) {
    return 0;
}

return window.pageYOffset + fieldRef.getBoundingClientRect().top + 75;
```

src/constants.js
```diff
@@ -420,3 +420,4 @@ export const KEYS = {
 
 /* ----------------- Other ----------------------- */
 export const ONE_HOUR_MS = 3600000; // 60 * 60 * 1000
+export const BASE_SELECT_FIELD_OFFSET_TOP_PADDING = 75;
```
jstoffan: Can we add a comment explaining the significance of the magic number of `75`? Also, this is a very specific and narrowly-used constant, so it should probably live in the component itself rather than in the global constants file.

src/components/select-field/SelectField.scss
```diff
@@ -3,6 +3,8 @@
 .select-field {
     .overlay {
         min-width: 100%;
+        min-height: 34px;
```
jstoffan: What is the significance of the magic number of `34`? Is it meant to display a minimum number of lines of text?

src/components/select-field/BaseSelectField.js
```diff
@@ -396,6 +415,7 @@ class BaseSelectField extends React.Component<Props, State> {
                             role="listbox"
                             // preventDefault on mousedown so blur doesn't happen before click
                             onMouseDown={event => event.preventDefault()}
+                            style={{ maxHeight: `calc(100vh - ${this.offsetTopValue}px)` }}
```
jstoffan: I suspect some instances of these select fields may need to be constrained to one or their parent elements rather than to the document via `vh`.

src/components/select-field/BaseSelectField.js
```diff
@@ -85,6 +85,24 @@ class BaseSelectField extends React.Component<Props, State> {
...
+    /**
+     * Calculate this field's distance from the top.
+     * This allows us to create an overflow container for the field's dropdown menu that is navigable using both
+     * the mouse and arrow keys and will always fit within the current viewport.
+     */
+    calculateOffsetTopValue() {
+        this.offsetTopValue =
+            window.pageYOffset +
+            this.fieldRef.current.getBoundingClientRect().top +
+            BASE_SELECT_FIELD_OFFSET_TOP_PADDING;
```
jstoffan: Can we add this via CSS as padding or margin rather than including it in the height calculation?

src/components/select-field/BaseSelectField.js
```diff
@@ -85,6 +85,24 @@ class BaseSelectField extends React.Component<Props, State> {
...
+        this.calculateOffsetTopValue();
+    }
+
+    /**
+     * Calculate this field's distance from the top.
+     * This allows us to create an overflow container for the field's dropdown menu that is navigable using both
+     * the mouse and arrow keys and will always fit within the current viewport.
+     */
+    calculateOffsetTopValue() {
+        this.offsetTopValue =
```
jstoffan: Minor, but I'm not sure this method is necessary. It seems like its logic could be inlined in `componentDidMount`. If we decide to keep it, I think it should just calculate and return the value rather than assigning it to the class instance, to reduce side effects and ease testing.

src/components/select-field/BaseSelectField.js
```diff
@@ -382,6 +400,7 @@ class BaseSelectField extends React.Component<Props, State> {
                 className={classNames(className, 'select-container')}
                 onBlur={this.handleBlur}
                 onKeyDown={this.handleKeyDown}
+                ref={this.fieldRef}
```
jstoffan: Minor, but should this be `containerRef`, instead?

src/components/select-field/BaseSelectField.js
sahiga: In that case, it might be better to set a defined `max-height` via CSS and remove the JS calculations in this component. Thoughts?

jstoffan: Yeah, possibly. It seems like a reasonable compromise on the behavior and would be a much simpler implementation.

src/components/select-field/SelectField.scss
```diff
@@ -3,6 +3,8 @@
 .select-field {
     .overlay {
         min-width: 100%;
+        max-height: 340px; // show at most 12 items in the dropdown menu
```
jstoffan: This component is probably used in lots of places. Would it be worthwhile to add this behavior to a modifier class that can be activated via a prop, defaulted to false?

src/components/select-field/BaseSelectField.js
```diff
@@ -49,6 +49,8 @@ type Props = {
     options: Array<SelectOptionProp>,
     /** The select button text shown when no options are selected. */
     placeholder?: string | React.Element<any>,
+    /** The select field overlay (dropdown) will have a scrollbar and max-height if true * */
+    scrollable?: boolean,
```
jstoffan: Should this be `isScrollable`?

sahiga: Yes, will change this name

PR #1672, opened 2019-10-17
chore(feed): clean up feed

alexkrolick:
- remove dead features:
  - keywords feature
  - "old" tasks api
    - note: some places will still refer to "TaskNew" until followup
- fix flow types for feed items
- refactor feed.js feature flip to use options instead of positional args,
  so that removing flips does not cause a shift in other flags


boxcla:

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -74,16 +73,15 @@ const ActiveState = ({
...
     return (
         <ul className="bcs-activity-feed-active-state">
-            {items.map((item: any) => {
-                const { type, id, versions, permissions } = item;
+            {items.map((item: FeedItem) => {
                 const isFocused = item === activeEntry;
                 const refValue = isFocused ? activeFeedItemRef : undefined;
 
-                switch (type) {
+                switch (item.type) {
```
alexkrolick: destructuring changes the type from an enum of literals to an enum including `string`, which doesn't allow the case statements to be refined for each item type

flow-typed/box-ui-elements.js
```diff
@@ -658,31 +670,18 @@ type FileAccessStats = {
...
-    message: string,
-    task_assignment_collection: TaskAssignments,
+    permissions: BoxTaskPermission,
     type: 'task',
 };
 
 type Tasks = {
     entries: Array<Task>,
-    total_count: number,
+    next_marker: 'string' | null,
```
sahiga: Are the quotes around "string" intentional?

src/api/__tests__/Feed-test.js
```diff
@@ -476,25 +383,16 @@ describe('api/Feed', () => {
...
+            feed.feedItems(file, false, successCb, errorCb, errorCb, { shouldShowAppActivity: true });
             setImmediate(() => {
                 expect(feed.fetchAppActivity).toHaveBeenCalled();
                 done();
             });
         });
 
-        test('should note use the app activity api if the seventh arg is false', done => {
-            feed.feedItems(file, false, successCb, errorCb, errorCb, false, false);
+        test('should note use the app activity api if the { shouldShowAppActivity } option in the is false', done => {
```
sahiga: "note" -> "not"?
"option in the" -> "option in the last argument"?

flow-typed/box-ui-elements.js
alexkrolick: copied from "Collaborators" object but I doubt it should be a literal, fixing

tjgupta:
🔥

PR #1675, opened 2019-10-18
feat(metadata-views): Data types for inline edit in metadata view

manishPh:


boxcla:

src/features/metadata-based-view/MetadataBasedItemList.js
```diff
@@ -127,11 +129,11 @@ class MetadataBasedItemList extends React.Component<Props, State> {
                 );
                 break;
             default: {
-                const data = getProp(item, 'metadata.data', {});
                 const mdFieldName = this.getMetadataColumnName(metadataColumn);
+                const data = getProp(find(fields, ['name', mdFieldName]), 'value'); // find field and get value
```
mxiao6: Should we have a default value `{}` just like the origin code?

src/api/MetadataQuery.js
```diff
@@ -98,32 +123,34 @@ class MetadataQuery extends Base {
...
+            .filter(key => !key.startsWith('$'))
+            .map(key => {
+                const type = getProp(find(templateFields, ['key', key]), 'type'); // get data type
+                const field: FlattenedMetadataQueryResponseEntryMetadataField = {
+                    name: key,
+                    value: instance[key],
+                    type,
                 };
+
+                if ([FIELD_TYPE_ENUM, FIELD_TYPE_MULTISELECT].includes(type)) {
```
sahiga: IE doesn't support `array.includes`. Can you change this to an object or set?

src/api/MetadataQuery.js
```diff
@@ -147,24 +174,48 @@ class MetadataQuery extends Base {
...
             nextMarker: next_marker,
         };
     };
 
     /**
-     * @param {Object} response
+     * @param {Object} template metadata template object
      */
-    queryMetadataSuccessHandler = ({ data }: { data: MetadataQueryResponse }): void => {
+    successHandler = (template: MetadataTemplate | void): void => {
```
sahiga: You could shorten this signature to `(template: ?MetadataTemplate): void`

src/api/MetadataQuery.js
```diff
@@ -98,32 +123,34 @@ class MetadataQuery extends Base {
...
-                    },
+        const templateFields = getProp(this.templateInfo, 'fields', []);
+        const instance = getProp(metadata, `${this.templateScope}.${this.templateKey}`);
+
+        if (!instance) {
+            return {};
+        }
+
+        const fields = Object.keys(instance)
+            .filter(key => !key.startsWith('$'))
```
sahiga: Previously, this code filtered out only six keys. Do we want to filter out all keys that start with `$`?

src/api/MetadataQuery.js
```diff
@@ -147,24 +174,48 @@ class MetadataQuery extends Base {
...
+        const { entries, next_marker }: MetadataQueryResponse = this.metadataQueryResponse;
         return {
             items: entries.map<FlattenedMetadataQueryResponseEntry>(this.flattenResponseEntry),
             nextMarker: next_marker,
         };
     };
 
     /**
-     * @param {Object} response
+     * @param {Object} template metadata template object
```
sahiga: Should this be just "metadata template object"?

src/api/MetadataQuery.js
```diff
@@ -147,24 +174,48 @@ class MetadataQuery extends Base {
...
+    getTemplateSchemaInfo = (metadataQueryResponse: MetadataQueryResponse): Promise<MetadataTemplate | void> => {
+        this.metadataQueryResponse = this.filterMetdataQueryResponse(metadataQueryResponse);
+        const { entries } = metadataQueryResponse;
+
+        if (!entries || entries.length === 0) {
+            // if no items found for provided query, don't make API call to get template info
+            return Promise.resolve();
+        }
+
+        const metadata = getProp(entries, '[0].metadata');
```
sahiga: Can you add a constant for this string?

src/features/metadata-based-view/MetadataBasedItemList.js
```diff
@@ -113,6 +114,7 @@ class MetadataBasedItemList extends React.Component<Props, State> {
         const isCellEditable = isCellHovered && !!getProp(metadataColumn, 'canEdit', false);
         const item = items[rowIndex - 1];
         const { name } = item;
+        const fields = getProp(item, 'metadata.fields', []);
```
sahiga: Can you add a constant for this string as well?

src/api/MetadataQuery.js
```diff
@@ -5,26 +5,51 @@
...
     FlattenedMetadataQueryResponseEntry,
     FlattenedMetadataQueryResponseEntryMetadata,
+    FlattenedMetadataQueryResponseEntryMetadataField,
     MetadataQueryResponse,
     MetadataQueryResponseEntry,
     MetadataQueryResponseEntryMetadata,
 } from '../common/types/metadataQueries';
 
+const FIELD_TYPE_MULTISELECT: 'multiSelect' = 'multiSelect';
+const FIELD_TYPE_ENUM: 'enum' = 'enum';
```
mxiao6: Should we define these types inside `'../common/constants'`?

src/api/MetadataQuery.js
mxiao6: Should this be `@param {Object} template - metadata template object`?

src/api/Metadata.js
```diff
@@ -179,6 +189,17 @@ class Metadata extends File {
         return getProp(templates, 'data.entries', []);
     }
 
+    /**
+     * Gets metadata template schema by template key
+     *
+     * @param {string} templateKey - template key
+     * @return {Object} metadata template
```
mxiao6: Here should be a Promise, right?

src/api/MetadataQuery.js
manishPh: Yes

src/api/MetadataQuery.js
manishPh: Will use `lodash/includes`

src/features/metadata-based-view/MetadataBasedItemList.js
manishPh: Default value of `getProp()` is `undefined` which is fine in this case.

src/api/Metadata.js
```diff
@@ -94,6 +94,16 @@ class Metadata extends File {
         return `${this.getMetadataTemplateUrl()}?metadata_instance_id=${id}`;
     }
 
+    /**
+     * API URL for getting metadata template schema by template key
+     *
+     * @param {string} templateKey - metadata template key
+     * @return {string} API url for getting template schema by template key
+     */
+    getSchemaByTemplateKeyUrl(templateKey: string): string {
```
ConradJChan: Should this be named `getMetadataTemplateSchemaUrl`?

src/api/MetadataQuery.js
```diff
@@ -5,26 +5,51 @@
...
+    /**
+     * @property {object}
+     */
+    templateInfo: MetadataTemplate;
+
+    /**
+     * @property {object}
+     */
+    metadataQueryResponse: MetadataQueryResponse;
+
```
ConradJChan: Can we add property descriptions? Or just remove it since the flow type somewhat describes it.

src/api/MetadataQuery.js
ConradJChan: Could this instead be a constant defined above like `SELECT_TYPES` and referenced here? `SELECT_TYPES.includes(type)`

src/api/MetadataQuery.js
```diff
@@ -98,32 +123,34 @@ class MetadataQuery extends Base {
...
+            .map(key => {
+                const type = getProp(find(templateFields, ['key', key]), 'type'); // get data type
+                const field: FlattenedMetadataQueryResponseEntryMetadataField = {
+                    name: key,
+                    value: instance[key],
+                    type,
                 };
+
+                if ([FIELD_TYPE_ENUM, FIELD_TYPE_MULTISELECT].includes(type)) {
+                    field.options = getProp(find(templateFields, ['key', key]), 'options'); // get "options"
```
ConradJChan: Could we store the reference to the found template field in L136 and re-use that here?

src/api/MetadataQuery.js
ConradJChan: Function description in js doc?

src/features/metadata-based-view/MetadataBasedItemList.js
manishPh: This is object path `item.metadata.fields`. Don't think we should create constant out of this as this is not going to be used elsewhere?

src/api/MetadataQuery.js
```diff
@@ -147,24 +174,48 @@ class MetadataQuery extends Base {
...
+        if (!entries || entries.length === 0) {
+            // if no items found for provided query, don't make API call to get template info
+            return Promise.resolve();
+        }
+
+        const metadata = getProp(entries, '[0].metadata');
+        this.templateScope = Object.keys(metadata)[0];
+        const instance = metadata[this.templateScope];
+        this.templateKey = Object.keys(instance)[0];
+        const metadataAPI = new MetadataAPI(this.options);
```
ConradJChan: Should this api be stored on the class and cleaned up in a `destroy()` method?

src/api/MetadataQuery.js
manishPh: This is object path `entries[0].metadata`.  Don't think we should create constant out of this as this is not going to be used elsewhere?

src/api/MetadataQuery.js
```diff
@@ -147,54 +158,84 @@ class MetadataQuery extends Base {
...
+            // if no items found for provided query, don't make API call to get template info
+            return Promise.resolve();
+        }
+
+        const metadata = getProp(entries, '[0].metadata');
+        this.templateScope = Object.keys(metadata)[0];
+        const instance = metadata[this.templateScope];
+        this.templateKey = Object.keys(instance)[0];
+
+        return this.metadataAPI.getSchemaByTemplateKey(this.templateKey);
```
jstoffan: I think we should probably avoid having API classes call each others methods directly. Can the calling component handle this as part of its promise chain? Or should the `MetadataQuery` class be combined with or extend the `Metadata` class?

src/api/MetadataQuery.js
```diff
@@ -147,54 +158,84 @@ class MetadataQuery extends Base {
...
     };
 
     /**
-     * @param {Object} response
+     * Success handler function storing flattened response in Cache
+     * @param {Object} templateSchemaResponse - metadata template schema response
      */
-    queryMetadataSuccessHandler = ({ data }: { data: MetadataQueryResponse }): void => {
+    successHandler = (templateSchemaResponse: ?MetadataTemplateSchemaResponse): void => {
+        this.metadataTemplate = getProp(templateSchemaResponse, 'data');
```
jstoffan: I think our eventual goal is for our API layer to be as stateless as possible. Can we avoid storing the response payload in the API class? Also, do we need `getProp` here or can we destructure this value in the function signature?

src/api/MetadataQuery.js
```diff
@@ -147,54 +158,84 @@ class MetadataQuery extends Base {
...
     /**
      * Does the network request to metadata query API
      * @param {Object} query query object with SQL Clauses like properties
      * @return {void}
      */
-    queryMetadataRequest(query: MetadataQueryType): void {
+    queryMetadataRequest(query: MetadataQueryType): Promise<void> {
         if (this.isDestroyed()) {
-            return;
+            Promise.resolve();
```
jstoffan: What purpose is this serving if this Promise isn't returned?

src/api/MetadataQuery.js
```diff
@@ -207,6 +248,7 @@ class MetadataQuery extends Base {
         this.key = this.getCacheKey(context.id);
         this.successCallback = successCallback;
         this.errorCallback = errorCallback;
+        this.metadataAPI = metadataAPI;
```
jstoffan: I think we should reconsider the architecture of these classes to avoid passing class instances around as function parameters.

src/api/MetadataQuery.js
manishPh: Doesn't matter. Undoing this change.

src/api/MetadataQuery.js
manishPh: There is not really a parent-child relationship between Metadata API and MetadataQuery API. So this entire logic merging metadata and template data type in `<ContentExplorer>` as this is the calling component. If we really don't want API classes calling each others methods (although there are examples of that currently), then I suggest adding the `flattenMetadata` util back in `src/utils` and `<ContentExplorer />` calling both APIs in its promise chain to gather the required data & types and merging & flattening using the`flattenMetadata` util.

src/api/MetadataQuery.js
manishPh: We always store references to success and error callbacks, error codes at class level. This is to avoid passing/relaying arguments between methods. `getProp` is used to gracefully handle the absence of template (when `response.entries = [], we don't need to get template info`). Destructuring data in function signature won't work when templateSchemaResponse is `undefined`.

src/api/MetadataQuery.js
manishPh: This is to manage `destroy()` better as `componentWillUnmount()` will destroy and delete all active xhrs and APIs respectively in `<ContentExplorer />`.

src/api/MetadataQuery.js
jstoffan: We currently store callbacks and error codes because they tie into underlying functionality in the API base classes. Even in those cases we should aim to move away from it in the future. Adding _more_ state to the class in the form of API responses seems counterproductive.

src/api/__tests__/MetadataQuery.test.js
```diff
@@ -204,42 +257,64 @@ describe('api/MetadataQuery', () => {
...
+            expect(metadataQuery.metadataAPI.getSchemaByTemplateKey).toHaveBeenCalledWith(templateKey);
+            expect(response).toBe(metadataTemplate);
+        });
+    });
+
     describe('queryMetadataRequest()', () => {
         beforeEach(() => {
-            metadataQuery.queryMetadataSuccessHandler = jest.fn();
+            metadataQuery.successHandler = jest.fn();
+            // metadataQuery.getTemplateSchemaInfo = jest.fn();
```
jstoffan: Typo? We usually want to avoid committing commented code.

src/api/MetadataQuery.js
jstoffan: I think the main API class that calls other API classes is Feed, which has been pretty difficult to work with. This type of coordination would normally be done by a container invoking an asynchronous action, which most of our top-level Elements currently lack. If you're concerned about polluting ContentExplorer, you could consider writing a component-oriented helper class, similar to the [version history sidebar's](https://github.com/box/box-ui-elements/blob/master/src/elements/content-sidebar/versions/VersionsSidebarAPI.js).

src/api/MetadataQuery.js
jstoffan: I see. Will the calling code need to check for the existence of the promise before adding to its chain?

src/api/MetadataQuery.js
jstoffan: I didn't see any other instances of this in our code. If it's meant to address a component-related issue, it should be handled in the component layer rather than the API layer.

PR #1677, opened 2019-10-21
style(uploadmanager): deja blue reskin for upload manager

il10:
**BEFORE: **
Screen Shot 2019-10-21 at 4 18 56 PM
Screen Shot 2019-10-21 at 4 20 33 PM

**AFTER: **
Screen Shot 2019-10-21 at 4 19 44 PM
Screen Shot 2019-10-21 at 4 20 05 PM

https://app.zeplin.io/project/5d813c56b87bf954d7b9074e/screen/5daa5ed27746247997b10d41

### DIFF
--- a/src/elements/content-uploader/UploadsManager.scss
+++ b/src/elements/content-uploader/UploadsManager.scss
@@ -2,9 +2,14 @@
 
 .be.bcu-uploads-manager-container {
     position: fixed;
+    right: 56px;
     bottom: 0;
-    width: 100%;
+    left: 288px;
+    width: unset;
     max-height: 0;
+    border-top-left-radius: 12px;
+    border-top-right-radius: 12px;
+    box-shadow: 0 2px 25px 0 rgba(0, 0, 0, .1);
     transition: max-height .5s;
 
     .bcu-progress-container {
--- a/src/elements/content-uploader/OverallUploadsProgressBar.scss
+++ b/src/elements/content-uploader/OverallUploadsProgressBar.scss
@@ -28,6 +28,6 @@
     .bcu-upload-status {
         flex: none;
         margin-right: 20px;
-        font-size: 16px;
+        font-size: 14px;
     }
 }


boxcla:

priyajeet:
Could you add a before/after screenshot?

src/elements/content-uploader/UploadsManager.scss
```diff
@@ -2,9 +2,14 @@
 
 .be.bcu-uploads-manager-container {
```
shikharmohan: So you're going to want to scope these changes for deja blue specifically and shouldn't affect non-deja blue webapp. 

So something like this is what you're going to need to do at the end of this file: 

```
.Body.is-deja-blue {
  .be.bcu-uploads-manager-container  {
 // add new styles here
}
}
```

src/elements/content-uploader/UploadsManager.scss
```diff
@@ -2,9 +2,14 @@
...
     position: fixed;
+    right: 56px;
     bottom: 0;
-    width: 100%;
+    left: 288px;
```
shikharmohan: This needs to accomodate the scenario when the left nav is collapsed in a responsive mode. So the actual left positioning of the uploads manager should be $left-nav-width + 32px

$left-nav-width can either be 256px or the smaller/collapsed state (don't know the px off the top of my head)

You'll probably need to import the media queries and variables for left nav collapsed state and set the styles accordingly

src/elements/content-uploader/UploadsManager.scss
```diff
@@ -2,9 +2,14 @@
 
 .be.bcu-uploads-manager-container {
     position: fixed;
+    right: 56px;
```
priyajeet: Are these positioning styles supposed to be here or in the parent app? Only generic styles should be here ideally.
And for the box-shadow, can you use the $black variable instead. rgba( $black, .1 )

src/elements/content-uploader/UploadsManager.scss
jstoffan: These styles should live in the parent application, as this component is used by multiple parties, many of whom do not have a left nav at all.

src/elements/content-uploader/UploadsManager.scss
il10: My bad. All changes go under a deja blue class.

src/elements/content-uploader/UploadsManager.scss
```diff
@@ -2,9 +2,14 @@
...
     bottom: 0;
-    width: 100%;
+    left: 288px;
+    width: unset;
     max-height: 0;
+    border-top-left-radius: 12px;
```
mickr: Can these also use a variable as I can imagine that `border-radius` should be consistent ?

il10: The radius is only applied to this specific case - it won't be used anywhere else

PR #1678, opened 2019-10-22
fix: slight adjust in check mark alignment

priyajeet:
Does not change menu item padding

**Before**
Screen Shot 2019-10-21 at 4 43 32 PMScreen Shot 2019-10-21 at 4 43 23 PMScreen Shot 2019-10-21 at 4 47 44 PM

**After**
Screen Shot 2019-10-21 at 5 08 17 PMScreen Shot 2019-10-21 at 5 08 25 PMScreen Shot 2019-10-21 at 4 47 27 PM

### DIFF
--- a/src/components/select-field/SelectField.scss
+++ b/src/components/select-field/SelectField.scss
@@ -18,9 +18,9 @@
 }
 
 .select-option-check-icon {
-    position: relative;
-    left: -5px;
     flex: none;
-    width: 16px;
+    width: 20px;
     height: 16px;
+    margin-left: -5px;
+    text-align: left;
 }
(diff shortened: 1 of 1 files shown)


boxcla:

src/components/select-field/SelectField.scss
```diff
@@ -18,7 +18,9 @@
 }
 
 .select-option-check-icon {
+    position: relative;
```
jstoffan: Would `margin-left: -5px` work? It might allow you to avoid the position style.

priyajeet: Yeah tried that 1st, didn't work because it moved the label along:

Screen Shot 2019-10-21 at 5 52 55 PM

let me double check

priyajeet: I can make that though will have to keep icon width to 20px

PR #1681, opened 2019-10-22
fix(pill-selector-dropdown): fix aria attributes

tjgupta:
Recommend ignoring whitespace to review. The main change is to move the `<Label>` outside of the `<SelectorDropdown>` so that the `<PillSelector>` is the root of the `selector` prop. This is needed in order to properly pass the `aria-*` attributes to the `textfield`.

### DIFF
--- a/src/components/pill-selector-dropdown/PillSelectorDropdown.js
+++ b/src/components/pill-selector-dropdown/PillSelectorDropdown.js
@@ -243,16 +243,16 @@ class PillSelectorDropdown extends React.Component<Props, State> {
         } = this.props;
 
         return (
-            <SelectorDropdown
-                className={classNames('pill-selector-wrapper', className)}
-                dividerIndex={dividerIndex}
-                onEnter={this.handleEnter}
-                onSelect={this.handleSelect}
-                overlayTitle={overlayTitle}
-                scrollBoundarySelector={dropdownScrollBoundarySelector}
-                shouldSetActiveItemOnOpen={shouldSetActiveItemOnOpen}
-                selector={
-                    <Label text={label}>
+            <Label text={label}>
+                <SelectorDropdown
+                    className={classNames('pill-selector-wrapper', className)}
+                    dividerIndex={dividerIndex}
+                    onEnter={this.handleEnter}
+                    onSelect={this.handleSelect}
+                    overlayTitle={overlayTitle}
+                    scrollBoundarySelector={dropdownScrollBoundarySelector}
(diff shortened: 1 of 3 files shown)


boxcla:

src/components/pill-selector-dropdown/PillSelectorDropdown.js
```diff
@@ -243,16 +243,16 @@ class PillSelectorDropdown extends React.Component<Props, State> {
...
-            <SelectorDropdown
-                className={classNames('pill-selector-wrapper', className)}
-                dividerIndex={dividerIndex}
-                onEnter={this.handleEnter}
-                onSelect={this.handleSelect}
-                overlayTitle={overlayTitle}
-                scrollBoundarySelector={dropdownScrollBoundarySelector}
-                shouldSetActiveItemOnOpen={shouldSetActiveItemOnOpen}
-                selector={
-                    <Label text={label}>
```
tjgupta: This is the main change. I moved this to outside of the `<SelectorDropdown>`.

src/components/pill-selector-dropdown/__tests__/PillSelectorDropdown.test.js
```diff
@@ -1,115 +1,83 @@
 import React from 'react';
 import { List, Record } from 'immutable';
-import sinon from 'sinon';
```
tjgupta: I refactored this file to remove `sinon` and to make use of a `getWrapper` function.

src/components/pill-selector-dropdown/PillSelectorDropdown.js
```diff
@@ -243,16 +243,16 @@ class PillSelectorDropdown extends React.Component<Props, State> {
...
+            <Label text={label}>
```
jstoffan: It looks like Label uses Tooltip, which uses TetherComponent. Will there be any impact on styling or behavior if the DOM elements of SelectorDropdown are now attached to the body rather than their parent?

tjgupta: I think it still works ok. Here's an example with the field disabled:

Screen Shot 2019-10-22 at 3 02 10 PM

jstoffan: Looking again, Label only wraps its children with a Tooltip if the `tooltip` prop is defined (go figure), which isn’t the case here. Sorry for the false alarm!

PR #1682, opened 2019-10-22
fix(notifications): add accessibility attributes

shahzadaziz:


boxcla:

src/components/notification/NotificationsWrapper.js
```diff
@@ -1,12 +1,20 @@
...
     children?: React.Node,
 };
 
-const NotificationsWrapper = ({ children }: Props) => <Portal className="notifications-wrapper">{children}</Portal>;
+const NotificationsWrapper = ({ children }: Props) => (
+    <FocusTrap>
+        <Portal className="notifications-wrapper" aria-live="polite">
+            {children}
+        </Portal>
+        ;
```
jstoffan: Typo?

src/components/notification/NotificationsWrapper.js
```diff
@@ -1,12 +1,19 @@
...
+import FocusTrap from '../focus-trap';
 import Portal from '../portal';
 
 type Props = {
     children?: React.Node,
 };
 
-const NotificationsWrapper = ({ children }: Props) => <Portal className="notifications-wrapper">{children}</Portal>;
+const NotificationsWrapper = ({ children }: Props) => (
+    <FocusTrap>
```
jstoffan: I'm not super familiar with these components. Does the FocusTrap render around the correct part of the DOM, given that its children are rendered in a portal? Or should it be more like:

```
 <Portal className="notifications-wrapper" aria-live="polite">
    <FocusTrap>
        {children}
    </FocusTrap>
</Portal>
```

shahzadaziz: Yes , that is correct. I just moved it inside. You just saved me some extra debugging time.

src/components/notification/__tests__/NotificationsWrapper.test.js
```diff
@@ -7,7 +7,25 @@ describe('components/notification/NotificationsWrapper', () => {
...
+            </NotificationsWrapper>,
+        );
+        const focusTrap = wrapper.find('FocusTrap');
+        expect(focusTrap.length).toEqual(1);
+    });
+
+    test('should not render focusTrap if there are no children', () => {
+        const wrapper = shallow(<NotificationsWrapper />);
+        const focusTrap = wrapper.find('FocusTrap');
+        expect(focusTrap.length).toEqual(0);
```
jstoffan: You can also use `expect(wrapper.exists('FocusTrap')).toBe(false);`, if you prefer. It was improved in a recent version of Enzyme.

src/components/notification/NotificationsWrapper.js
```diff
@@ -1,12 +1,17 @@
...
 import Portal from '../portal';
 
 type Props = {
     children?: React.Node,
 };
 
-const NotificationsWrapper = ({ children }: Props) => <Portal className="notifications-wrapper">{children}</Portal>;
+const NotificationsWrapper = ({ children }: Props) => (
+    <Portal className="notifications-wrapper" aria-live="polite">
+        {children ? <FocusTrap>{children}</FocusTrap> : null}
```
priyajeet: Children should be required, no? Otherwise what is this component rendering?

jstoffan: It was already nullable, so seems like we're just maintaining the existing API?

priyajeet: maybe should fix the flow type (though unsure what other component flow might require fixing)

src/components/notification/__tests__/NotificationsWrapper.test.js
```diff
@@ -7,7 +7,25 @@ describe('components/notification/NotificationsWrapper', () => {
     test('should render a Portal with the notifications-wrapper class', () => {
         const wrapper = shallow(<NotificationsWrapper />);
         expect(wrapper.is('Portal')).toBeTruthy();
+        const portal = wrapper.find('Portal');
```
cweeii: Is this needed if the wrapper is Portal?

src/components/notification/NotificationsWrapper.js
shahzadaziz: For now goal is to maintain existing API. I agree that children should be required.

src/components/notification/NotificationsWrapper.js
```diff
@@ -1,12 +1,17 @@
...
+import FocusTrap from '../focus-trap';
 import Portal from '../portal';
 
 type Props = {
     children?: React.Node,
 };
 
-const NotificationsWrapper = ({ children }: Props) => <Portal className="notifications-wrapper">{children}</Portal>;
+const NotificationsWrapper = ({ children }: Props) => (
+    <Portal className="notifications-wrapper" aria-live="polite">
```
tjgupta: It seems from the docs for `aria-live` that the content has to always be there in the DOM for it to read it out. Since this appearing in a transient way, does this work with screen readers?

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions

"Always make sure that the live region is present in the document first, and only then dynamically add/change any content."

shahzadaziz: <NotificationsWrapper> is always present and aria attributes are applied to it so this works fine. 
I did experience weird behavior when I applied live region to individual notification. Which makes sense now given the docs you shared.

shahzadaziz:
@jstoffan addressed comments.

PR #1684, opened 2019-10-22
fix(modal): move close button after header for better screen reader a11y

tjgupta:
This looks visually the same (due to absolute positioning), but results in the screen reader reading the title of the modal before saying "Close button" (for browsers like Chrome that read the contents of the modal... it seems Safari doesn't read it all out).


boxcla:

PR #1686, opened 2019-10-23
feat(dropdown-menu): allow esc to propagate if dropdown is closed

cweeii:


boxcla:

src/components/dropdown-menu/DropdownMenu.js
```diff
@@ -125,7 +127,10 @@ class DropdownMenu extends React.Component<Props, State> {
                 break;
 
             case 'Escape':
```
mxiao6: Why didn't we use `KEYS.escape` etc. from `constants.js` here? Can you correct them?

cweeii: Yep!  I'll update them, good idea.

PR #1687, opened 2019-10-23
fix(notifications): add support for keyboard and closeOnClick

shahzadaziz:


boxcla:

src/components/notification/Notification.js
```diff
@@ -49,6 +50,8 @@ type Props = {
      * - Notification buttons must be the `<Button />` component.
      */
     children: React.Node,
+    /** If set it forces the notification to stay open and only close when clicked on buttons inside it */
+    closeOnClick?: boolean,
```
priyajeet: prefix with is/has/should for booleans

src/components/notification/Notification.js
```diff
@@ -8,6 +8,7 @@ import IconBell from '../../icons/general/IconBell';
 import IconClose from '../../icons/general/IconClose';
 import IconInfoThin from '../../icons/general/IconInfoThin';
 import IconSync from '../../icons/general/IconSync';
+import { KEYS } from '../../constants';
```
priyajeet: If you dont mind, while you are at it, change #FFF to white imported from styles.js
line 123, 134

src/components/notification/Notification.js
```diff
@@ -103,7 +118,7 @@ class Notification extends React.Component<Props> {
         const classes = classNames('notification', type, overflow);
 
         return (
-            <div className={classes}>
+            <div className={classes} onKeyDown={this.handleKeyDown} role="presentation">
```
cweeii: I'm curious, why did you decide to add a role of presentation to the div?

shahzadaziz: This needed to be assigned a role to be treated as an interactive banner. The closest role I could find for this was "presentation".

alexkrolick: What about "alertdialog"

src/components/notification/Notification.js
```diff
@@ -89,6 +96,14 @@ class Notification extends React.Component<Props> {
         }
     };
 
+    handleKeyDown = (event: SyntheticKeyboardEvent<>) => {
+        if (event.key === KEYS.escape) {
+            event.stopPropagation();
+            event.preventDefault();
+            this.onClose();
```
mickr: Should this be called with the `event`?

src/components/notification/__tests__/Notification.test.js
```diff
@@ -75,101 +77,87 @@ describe('components/notification/Notification', () => {
...
+    test('should call onClose when Escape key is pressed', () => {
+        const mockEvent = {
+            key: 'Escape',
+            preventDefault: jest.fn(),
+            stopPropagation: jest.fn(),
+        };
+        const closeMock = jest.fn();
+        const component = getWrapper({ onClose: closeMock });
+
+        component.simulate('keyDown', mockEvent);
```
mickr: I've been told that we don't usually `simulate` events. Not sure if that wants to be applied here as well.

PR #1688, opened 2019-10-23
fix: various form accessibility improvements

tjgupta:


boxcla:

examples/src/SelectorDropdownExamples.js
```diff
@@ -56,7 +56,7 @@ class SelectorDropdownContainer extends Component {
         const dropdownTitle = <div>This is a Title</div>;
 
         return (
-            <div>
+            <div style={{ paddingBottom: '330px' }}>
```
tjgupta: Fixes this overlap:
Screen Shot 2019-10-23 at 3 40 52 PM

src/components/pill-selector-dropdown/PillSelector.js
```diff
@@ -207,8 +214,10 @@ class PillSelector extends React.Component<Props, State> {
                         onBlur={this.resetSelectedIndex}
                         ref={this.hiddenRef}
                         tabIndex={-1}
+                        data-testid="pill-selection-helper"
                     />
                     <textarea
+                        {...ariaAttrs}
```
tjgupta: Unfortunately `PillSelector` isn't currently using the `TextArea` component, so I had to do this separately. TODO: Move to `<TextArea>`.

src/components/pill-selector-dropdown/PillSelector.js
```diff
@@ -227,6 +236,9 @@ class PillSelector extends React.Component<Props, State> {
                         suggestedPillsData={suggestedPillsData}
                         title={suggestedPillsTitle}
                     />
+                    <span id={this.errorMessageID} className="accessibility-hidden" role="alert">
```
tjgupta: This will allow the error message to get read with a screen reader (in many cases, though unfortunately not all cases... but better than current anyway).

src/components/pill-selector-dropdown/PillSelector.js
```diff
@@ -178,6 +181,10 @@ class PillSelector extends React.Component<Props, State> {
             'show-error': !!error,
             'pill-selector-suggestions-enabled': suggestedPillsEnabled,
         });
+        const ariaAttrs = {
+            'aria-invalid': !!error,
+            'aria-errormessage': this.errorMessageID,
+        };
 
         return (
             <Tooltip isShown={!!error} text={error || ''} position="middle-right" theme="error">
```
priyajeet: Looks like we are doing !!error in a few places, perhaps should make a new const hasError

src/features/unified-share-modal/EmailForm.js
```diff
@@ -338,7 +338,7 @@ class EmailForm extends React.Component<Props, State> {
                             <FormattedMessage {...commonMessages.cancel} />
                         </Button>
                         <PrimaryButton
-                            isDisabled={submitting || selectedContacts.length === 0 || contactsFieldError} // Check selectedContacts.length === 0 for initial render when contactsFieldError is empty
```
tjgupta: The new directive is to keep form submit buttons enabled, even if form is not yet valid. This makes it easier to discover where the errors in the form are (because once you submit an invalid form you then see the errors inline on invalid fields), and this also improves keyboard a11y because you can focus on the button and activate it (and then get read the invalid field in some cases).

src/components/pill-selector-dropdown/PillSelector.js
```diff
@@ -207,8 +214,10 @@ class PillSelector extends React.Component<Props, State> {
                         onBlur={this.resetSelectedIndex}
                         ref={this.hiddenRef}
                         tabIndex={-1}
+                        data-testid="pill-selection-helper"
```
cweeii: If this attribute is used for the sole purpose for finding in tests, can you use some other attribute on this span?

tjgupta: Actually I think we've been moving _toward_ adding these test IDs in favor of more brittle selectors like CSS classes or other various attributes. This is also the Cypress recommendation as well: https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements

The previous class based brittle selector is why the current tests gave a false positive. Any particular reason to avoid adding a `data-testid`?

src/components/pill-selector-dropdown/PillSelector.js
tjgupta: Done @priyajeet

src/components/pill-selector-dropdown/PillSelector.js
priyajeet: It's fine, it will get stripped out in the release builds

src/components/text-input/TextInput.js
```diff
@@ -49,10 +50,18 @@ const TextInput = ({
     labelTooltip,
     ...rest
 }: Props) => {
+    const hasError = !!error;
     const classes = classNames(className, 'text-input-container', {
-        'show-error': !!error,
+        'show-error': hasError,
     });
 
+    const errorMessageID = React.useRef(uniqueId('errorMessage')).current;
```
alexkrolick: To avoid bumping that counter too much we can do this 

```jsx
// above the comp
const createInstanceId = memoize(uniqueId)

// the comp
const X = (props) => {
  const instance = useRef()
  const id = createInstanceId(instance)
}
```

tjgupta: I can set that up in a followup PR if it becomes a problem, but honestly if it comes to that, I think it's more directly understandable as a class component with an instance property.

PR #1689, opened 2019-10-23
chore(content-preview): Pass token generator to preview library

mxiao6:
The related preview change: https://github.com/box/box-content-preview/pull/1087


boxcla:

src/elements/content-preview/ContentPreview.js
```diff
@@ -700,8 +700,7 @@ class ContentPreview extends React.PureComponent<Props, State> {
         }
 
         const fileOpts = { ...fileOptions };
-        const typedId: string = getTypedFileId(fileId);
-        const token: TokenLiteral = await TokenService.getReadToken(typedId, tokenOrTokenFunction);
+        const token = typedId => TokenService.getReadTokens(typedId, tokenOrTokenFunction);
```
ConradJChan: Do we even need the argument in the token function? Since we know what file we're generating the token for, can we instead do `const token = () => TokenService.getReadToken(typedId, tokenOrTokenFunction)`?

mxiao6: If I get rid of the argument, I have to change the preview lib's `getToken` function, which I don't think we should touch.

PR #1690, opened 2019-10-24
style: global font size change to 14px

shikharmohan:
Adding new mixins for upcoming font changes

### DIFF
--- a/src/styles/common/_typography.scss
+++ b/src/styles/common/_typography.scss
@@ -20,7 +20,7 @@ body {
 }
 
 body.is-deja-blue {
-    @include common-typography-dejablue;
+    @include bdl-commonTypography-dejaBlue;
 }
 
 p {
--- a/src/styles/common/_typography.scss
+++ b/src/styles/common/_typography.scss
@@ -20,7 +20,7 @@ body {
 }
 
 body.is-deja-blue {
-    @include common-typography-deja-blue;
+    @include common-typography-dejablue;
 }
 
 p {
--- a/src/styles/common/_typography.scss
+++ b/src/styles/common/_typography.scss
@@ -19,6 +19,10 @@ body {
     @include common-typography;
 }
 
+body.is-deja-blue {
+    @include common-typography-deja-blue;
+}
+
 p {
     margin: 0 0 20px;
 }
--- a/src/styles/constants/_typography.scss
+++ b/src/styles/constants/_typography.scss
@@ -1,5 +1,5 @@
 $bdl-line-height: 20px;
 
 // font sizes
-$bdl-font-size: 13px;
-$bdl-font-size--dejablue: 14px;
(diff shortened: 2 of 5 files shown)


boxcla:

src/styles/common/_typography.scss
```diff
@@ -19,6 +19,10 @@ body {
     @include common-typography;
 }
 
+.Body.is-deja-blue {
```
alexkrolick: Check that this also affects modals and other detached roots

shikharmohan: modals/portalled elements have this new font size

src/styles/mixins/_typography.scss
```diff
@@ -19,10 +19,26 @@
...
     color: $bdl-gray;
     line-height: $bdl-line-height;
     letter-spacing: .3px;
 }
+
+@mixin common-typography-deja-blue {
+    @include fonts-deja-blue;
+
+    color: $bdl-gray;
+    line-height: $bdl-line-height;
```
tjgupta: Is the line height supposed to be the same with the larger font?

shikharmohan: same line height: 20px

src/styles/common/_typography.scss
```diff
@@ -19,6 +19,10 @@ body {
     @include common-typography;
 }
 
+body.is-deja-blue {
```
shikharmohan: Not going to change this top level class, requires monolith changes. Considering it out of scope right now.

src/styles/constants/_typography.scss
```diff
@@ -1 +1,5 @@
 $bdl-line-height: 20px;
+
+// font sizes
+$bdl-font-size: 13px;
+$bdl-font-size--dejablue: 14px;
```
priyajeet: fontSize, likewise camel casing other places too

src/styles/mixins/_typography.scss
priyajeet: I would prefix all these with bdl-

bdl-typography
bdl-commonTypography

src/styles/constants/_typography.scss
alexkrolick: None of the other variables are like that

shikharmohan: I'm cool w/ whatever we go with, lmk!

src/styles/mixins/_typography.scss
```diff
@@ -16,7 +16,15 @@
     @include font-aliasing;
 
     font-weight: normal;
-    font-size: 13px;
+    font-size: $bdl-fontSize;
+}
+
+@mixin bdl-fonts-dejaBlue {
```
priyajeet: --dejaBlue if you want to keep it as a modifier pattern like you did with the variable name

PR #1691, opened 2019-10-24
feat(select-field): Select field opens away from viewport boundary

ConradJChan:
Added an `Overlay` component which is viewport aware, i.e. if the select field is near the bottom of the viewport, the select options will appear above the button, or if the select field is near the right of the viewport, the select options will appear right aligned to the button.

Before:
Screen Shot 2019-10-24 at 12 40 04 PM

After:
Screen Shot 2019-10-24 at 12 39 39 PM

Animated gif:
context aware dropdown

TODO:
- [ ] fix unit tests

### DIFF
--- a/src/components/select-field/Overlay.js
+++ b/src/components/select-field/Overlay.js
@@ -1,7 +1,7 @@
 // @flow
 import * as React from 'react';
 import classNames from 'classnames';
-import { getDimensions, getViewportIntersections } from '../../utils/dom';
+import { getDimensions, getViewportIntersections, type ElementDimensions } from '../../utils/dom';
 import { OVERLAY_WRAPPER_CLASS, VIEWPORT_BORDERS } from '../../constants';
 
 type Props = {
@@ -10,27 +10,10 @@ type Props = {
     isOpen?: boolean,
 };
 
-type Dimensions = {
-    height?: number,
-    width?: number,
-};
-
 type Translations = {
     x?: number,
     y?: number,
 };
-
-function getDimensionsWhileHidden(element: HTMLElement): Dimensions {
-    if (!element) {
-        return {};
-    }
-
-    element.classList.add('offscreen');
-    const dimensions = getDimensions(element);
-    element.classList.remove('offscreen');
-    return dimensions;
-}
-
 class Overlay extends React.Component<Props, State> {
     overlayRef: ?HTMLDivElement;
 
@@ -42,24 +25,24 @@ class Overlay extends React.Component<Props, State> {
         this.overlayRef = React.createRef();
     }
 
-    getOverlayContentDimensions = (): Dimensions => {
+    getOverlayContentDimensions = (): ElementDimensions => {
         const node = this.overlayRef.current;
         const content = node && node.firstChild;
 
-        const dimensions = content.clientHeight ? getDimensions(content) : getDimensionsWhileHidden(content);
+        const dimensions = getDimensions(content);
 
         return dimensions;
     };
 
     getOverlayIntersections = (): Array<string> => {
         const { isOpen } = this.props;
-        const node = this.overlayRef.current;
 
         if (!isOpen) {
-            return false;
+            return [];
         }
 
         const dimensions = this.getOverlayContentDimensions();
+        const node = this.overlayRef.current;
         const boundingClientRect = node && node.getBoundingClientRect();
         const bounding = {
             top: boundingClientRect.top,
@@ -76,7 +59,9 @@ class Overlay extends React.Component<Props, State> {
         const { height: buttonHeight, width: buttonWidth } = this.buttonRef.current.getBoundingClientRect();
         const translations = { x: 0, y: 0 };
 
-        if (intersections.length >= 3) {
+        // If there are more than 2 intersections then it is likely that the overlay will extend beyond
+        // the viewport in at least one direction, therefore, there is no need to reposition the overlay
+        if (intersections.length > 2) {
             return translations;
         }
 
@@ -108,7 +93,7 @@ class Overlay extends React.Component<Props, State> {
 
         const intersections = this.getOverlayIntersections();
         const translations = intersections.length ? this.getOverlayTranslations(intersections) : null;
-        const inlineStyle = translations && { top: `${translations.y}px`, left: `${translations.x}px` };
+        const inlineStyle = translations ? { top: `${translations.y}px`, left: `${translations.x}px` } : null;
 
         return (
             <>
--- a/src/components/select-field/Overlay.js
+++ b/src/components/select-field/Overlay.js
@@ -1,8 +1,11 @@
 // @flow
 import * as React from 'react';
+import getProp from 'lodash/get';
 import classNames from 'classnames';
 import { getDimensions, getViewportIntersections, type ElementDimensions } from '../../utils/dom';
-import { OVERLAY_WRAPPER_CLASS, VIEWPORT_BORDERS } from '../../constants';
+import { VIEWPORT_BORDERS } from '../../constants';
+
+import './Overlay.scss';
 
 type Props = {
     children: React.Node,
@@ -71,7 +74,7 @@ class Overlay extends React.Component<Props, State> {
                     translations.y = -(buttonHeight + overlayHeight);
                     break;
                 case VIEWPORT_BORDERS.right:
-                    translations.x = -(overlayWidth - buttonWidth);
+                    translations.x = -Math.floor(overlayWidth - buttonWidth);
                     break;
                 default:
             }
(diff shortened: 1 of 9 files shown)


boxcla:

src/components/select-field/Overlay.js
```diff
@@ -0,0 +1,127 @@
...
+};
+
+function getDimensionsWhileHidden(element: HTMLElement): Dimensions {
+    if (!element) {
+        return {};
+    }
+
+    element.classList.add('offscreen');
+    const dimensions = getDimensions(element);
+    element.classList.remove('offscreen');
```
manishPh: What is the purpose of adding and removing class `offscreen` on either side of getting element's dimensions?

src/components/select-field/Overlay.js
```diff
@@ -0,0 +1,127 @@
...
+
+        return dimensions;
+    };
+
+    getOverlayIntersections = (): Array<string> => {
+        const { isOpen } = this.props;
+        const node = this.overlayRef.current;
+
+        if (!isOpen) {
+            return false;
```
manishPh: should return `[]`?

src/components/select-field/Overlay.js
```diff
@@ -0,0 +1,127 @@
...
+
+        return getViewportIntersections(bounding);
+    };
+
+    getOverlayTranslations = (intersections: Array<string>): Translations => {
+        const { height: overlayHeight, width: overlayWidth } = this.getOverlayContentDimensions();
+        const { height: buttonHeight, width: buttonWidth } = this.buttonRef.current.getBoundingClientRect();
+        const translations = { x: 0, y: 0 };
+
+        if (intersections.length >= 3) {
```
manishPh: Can you add comment why returning from here? May be add a constant for value `3`? Not sure why we return if more than 2 sides of rectangle intersect with viewport.

src/utils/dom.js
```diff
@@ -120,3 +120,49 @@ export function scrollIntoView(itemEl: ?HTMLElement, options?: Object = {}): voi
...
     }
 }
+
+/**
+ * Gets the viewport intersections.
+ *
+ * @param {Object} bounding - Object describing the position relative to the viewport. Typically retrieved from getBoundingClientRect()
+ * @return {Array} Returns an array of the boundary intersections ['top', 'left', 'bottom', 'right']
+ */
+export function getViewportIntersections(bounding) {
```
manishPh: add param and return types?

src/utils/dom.js
```diff
@@ -120,3 +120,49 @@ export function scrollIntoView(itemEl: ?HTMLElement, options?: Object = {}): voi
...
+    return boundaryIntersections;
+}
+
+/**
+ * Gets an HTMLElement's dimensions (width and height)
+ *
+ * @param {HTMLElement} element - The HTML element
+ * @return {Object} Object containing the width and height of the element
+ */
+export function getDimensions(element) {
```
manishPh: add param and return types?

src/components/select-field/Overlay.js
ConradJChan: I want to get the dimensions of the element while not visible, so adding the `offscreen` class allows me to do so. But I think I am changing up the CSS so that I can avoid this

src/components/select-field/Overlay.js
ConradJChan: I'll add a comment, but basically my thinking is that if there are more than 2 intersections with the viewport, part of the dropdown will appear outside the viewport regardless, so there is no need to move the overlay around

src/components/select-field/Overlay.js
```diff
@@ -0,0 +1,116 @@
...
+        super(props);
+        this.buttonRef = React.createRef();
+        this.overlayRef = React.createRef();
+    }
+
+    getOverlayContentDimensions = (): ElementDimensions => {
+        const node = this.overlayRef.current;
+        const content = node && node.firstChild;
+
+        const dimensions = getDimensions(content);
```
mickr: Can directly return `getDimensions()`

src/components/select-field/Overlay.js
```diff
@@ -0,0 +1,116 @@
...
+    render() {
+        const { children, className, isOpen } = this.props;
+        const elements = React.Children.toArray(children);
+
+        if (elements.length !== 2) {
+            throw new Error('Overlay must have exactly two children: A button component and the overlay content');
+        }
+
+        const button = elements[0];
+        const overlayContent = elements[1];
```
mickr: `const [button, overlayContent] = elements; `

src/components/select-field/Overlay.js
```diff
@@ -0,0 +1,116 @@
...
+        const intersections = this.getOverlayIntersections();
+        const translations = intersections.length ? this.getOverlayTranslations(intersections) : null;
+        const inlineStyle = translations ? { top: `${translations.y}px`, left: `${translations.x}px` } : null;
+        const overlayContentClassnames = classNames('bdl-Overlay-content', overlayContentExistingClassnames, {
+            'bdl-Overlay-content--visible': isOpen,
+        });
+
+        return (
+            <>
+                {React.cloneElement(button, { ref: this.buttonRef })}
```
mickr: Do you need to clone the element since you are already extracting them out as elements? Just call them like `<Button  ref={this.buttonRef} />`?

ConradJChan: I think because I modify it with `ref` I need to clone it

src/components/select-field/Overlay.js
```diff
@@ -0,0 +1,116 @@
...
+            bottom: boundingClientRect.top + dimensions.height,
+            right: boundingClientRect.left + dimensions.width,
+        };
+
+        return getViewportIntersections(bounding);
+    };
+
+    getOverlayTranslations = (intersections: Array<string>): Translations => {
+        const { height: overlayHeight, width: overlayWidth } = this.getOverlayContentDimensions();
+        const { height: buttonHeight, width: buttonWidth } = this.buttonRef.current.getBoundingClientRect();
```
mxiao6: Do we need to check the existence of `buttonRef` just like checking `overlayRef` above?

src/components/select-field/Overlay.js
```diff
@@ -0,0 +1,116 @@
...
+
+    buttonRef: ?HTMLButtonElement;
+
+    constructor(props) {
+        super(props);
+        this.buttonRef = React.createRef();
+        this.overlayRef = React.createRef();
+    }
+
+    getOverlayContentDimensions = (): ElementDimensions => {
```
mxiao6: Should we add JSDoc to these new functions?

src/components/select-field/Overlay.js
```diff
@@ -0,0 +1,116 @@
+// @flow
```
alexkrolick: Not sure if you already considered this, but Isn't this pretty similar to what Tether does for is in the other dropdown-style components? Could we use that for this intersection logic?

ConradJChan: Yeah, my initial implementation was to use the `Dropdown` component, but it turns out Tether will re-render each time the child changes causing a flicker in the dropdown which only really affects the `MultiSelectField` component where the dropdown remains open after selection in order to select other options.

ConradJChan: I had initially tried using `Dropdown` the component but found that Tether does a re-render when a child updates which causes a problem for the `MultiSelectField` because it remains open for multiple selections and causes a flicker

src/utils/dom.js
```diff
@@ -120,3 +120,54 @@ export function scrollIntoView(itemEl: ?HTMLElement, options?: Object = {}): voi
...
     }
 }
+
+/**
+ * Gets the viewport intersections.
+ *
+ * @param {Object} bounding - Object describing the position relative to the viewport. Typically retrieved from getBoundingClientRect()
+ * @return {Array} Returns an array of the boundary intersections ['top', 'left', 'bottom', 'right']
+ */
+export function getViewportIntersections(bounding: DOMRectReadOnly): Array<string> {
```
priyajeet: Move export to the bottom

priyajeet: Have you looked at / tried intersection observer

src/components/select-field/Overlay.js
priyajeet: I think he meant did you try using tether for this implementation, not drop-down. Like surrounding the ul with tether (which I am suggesting use popper instead)

ConradJChan: I didn't use Tether directly, but via `Dropdown`, but while debugging the flicker issue found that it's a limitation of Tether. I'll look into whether popper suffers from the same issue or not

PR #1692, opened 2019-10-24
fix(tooltip): color contrast and layout fixes

tjgupta:
Old error tooltip with low color contrast text and misaligned close button (when shown):
Screen Shot 2019-10-24 at 2 32 29 PM

New error tooltip with better color contrast and improved layout of close button (when shown):
Screen Shot 2019-10-24 at 2 22 24 PM

### DIFF
--- a/src/components/tooltip/Tooltip.scss
+++ b/src/components/tooltip/Tooltip.scss
@@ -61,7 +61,7 @@ $tooltip-offset: 4px; /* amount of space between target and tooltip arrow */
 
     .tooltip-close-button {
         position: absolute;
-        top: 10px;
+        top: 9px;
         right: 3px;
         cursor: pointer;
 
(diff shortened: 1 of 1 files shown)


boxcla:

src/components/tooltip/Tooltip.scss
```diff
@@ -56,13 +56,13 @@ $tooltip-offset: 4px; /* amount of space between target and tooltip arrow */
...
     &.with-close-button {
-        padding-right: 24px;
+        padding-right: 28px;
     }
 
     .tooltip-close-button {
         position: absolute;
-        top: 4px;
-        right: 0;
+        top: 10px;
```
alexkrolick: this doesn't look centered in the screenshot - would it make sense to use flex here to get it exactly right and handle multiple lines better?

tjgupta: I'm not sure if flexbox will help here. The close button needs to be centered with the first line of text, whether it's the only line or one of many in the tooltip.

The other issue is that depending on how the text is capitalized it looks less or more centered. I added more realistic text (text we actually show) to test the alignment, and I moved the close button up one pixel. Looks like this now:

Screen Shot 2019-10-25 at 9 44 19 AM

PR #1694, opened 2019-10-25
fix(feed): Task assignee names not truncating IE11

ConradJChan:
Before:
Screen Shot 2019-10-25 at 1 52 25 PM

After:
Screen Shot 2019-10-25 at 1 53 39 PM

### DIFF
--- a/src/components/media/Media.scss
+++ b/src/components/media/Media.scss
@@ -18,6 +18,7 @@
 // the main content, e.g. title and message
 .bdl-Media-body {
     flex: 1 1 100%;
+    overflow: hidden;
     word-wrap: break-word; // old IE
     word-break: break-word; // CJK scripts
     // handle wrapping long words in text content without breaking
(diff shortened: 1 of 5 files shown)


boxcla:

src/components/media/Media.scss
```diff
@@ -18,6 +18,7 @@
 // the main content, e.g. title and message
 .bdl-Media-body {
     flex: 1 1 100%;
+    overflow: hidden;
```
alexkrolick: Is this always true for every Media object?

ConradJChan: That's a good point, let me put this style under `.bcs-Task`

src/elements/content-sidebar/activity-feed/task-new/Task.scss
```diff
@@ -11,6 +11,10 @@ $bcs-task-left-offset: 10 * $bcs-task-grid + 2px;
     &.bcs-is-pending {
         opacity: .4;
     }
+
+    .bdl-Media-body {
```
alexkrolick: you could also add a class to the component instead of adding overloads to the name here

PR #1695, opened 2019-10-25
fix(uploads): Fix hanging on resuming uploads

stevenzyang:
* Fixes bug with subsequent uploadFile calls in Xhr.js failing
automatically after the original times out.
* Fixes hanging during logging of part failures by not blocking on it
* Fixes bug with listParts always returning undefined, causing the
client reupload a part it has already uploaded. Remove the byte check in
determining whether or not the file has been uploaded, because it has
shown to be unreliable.
* Removes PART_STATE_COMPUTING_DIGEST state as it is never used in the
code, and is not a useful state to keep track of.

Fixes most of WEBEN-1098, though I have still seen it hang from time to
time.


boxcla:

src/api/uploads/MultiputPart.js
```diff
@@ -265,7 +262,7 @@ class MultiputPart extends BaseMultiput {
                 throw new Error('logEvent endpoint not found');
             }
 
-            await retryNumOfTimes(
+            retryNumOfTimes(
```
mickr: Does this still belong in the `try/catch`?

src/api/uploads/MultiputUpload.js
```diff
@@ -496,13 +495,6 @@ class MultiputUpload extends BaseMultiput {
```
mickr: Noticed on L488 a reference to DigestComputing, is that still relevant?

src/api/uploads/MultiputPart.js
stevenzyang: Good point, I'll refactor this to attach a log on error instead

src/api/uploads/MultiputUpload.js
stevenzyang: Good catch, will remove

stevenzyang:
@mickr Could you take another look?

PR #1698, opened 2019-10-29
chore(formik): upgrade to v2

priyajeet:
- Formik hooks should be available with this upgrade.
- Fixes value and rest prop passthorugh for Toggle.
- Fixes a couple of null exceptions for pill selector.

BREAKING CHANGE: Drops vastly unused inputProps prop from Toggle component. Formik has been upgraded to V2. [Upgrade guide](https://github.com/jaredpalmer/formik/blob/master/docs/migrating-v2.md).


boxcla:

PR #1701, opened 2019-10-29
docs(icons): add prop types to two-toned icon docs

tjgupta:
This PR allows the ability to add props docs to each icon using the `IconsExample` component. This should unblock converting the "General" icons to using the `IconsExample` grid.


boxcla:

PR #1703, opened 2019-10-30
feat(metadata-views): Handle edit and cancel edit actions

manishPh:


src/features/metadata-based-view/MetadataBasedItemList.js
```diff
@@ -101,18 +111,50 @@ class MetadataBasedItemList extends React.Component<Props, State> {
         onItemClick(itemWithPreviewPermission);
     }
 
-    getGridCellData(columnIndex: number, rowIndex: number): GridCellData {
+    handleEditIconClick(columnIndex: number, rowIndex: number): void {
```
sahiga: Should this be an arrow function as well?

src/features/metadata-based-view/MetadataBasedItemList.js
```diff
@@ -101,18 +111,50 @@ class MetadataBasedItemList extends React.Component<Props, State> {
...
+            hoveredRowIndex: rowIndex,
+        });
+
+    handleMouseLeave = (): void =>
+        this.setState({
+            hoveredRowIndex: -1,
+            hoveredColumnIndex: -1,
+        });
+
+    getGridCellData(columnIndex: number, rowIndex: number): GridCellData | void {
```
manishPh: moving all event handlers together.

src/features/metadata-based-view/MetadataBasedItemList.js
```diff
@@ -101,18 +111,50 @@ class MetadataBasedItemList extends React.Component<Props, State> {
...
+
+    handleCancelEdit = (): void => {
+        this.setState({
+            editedColumnIndex: -1,
+            editedRowIndex: -1,
+        });
+    };
+
+    handleSave = (): void => {
+        /* Implement me */
```
manishPh: Capturing user input, validating, and saving data will be covered by subsequent PR.

src/features/metadata-based-view/MetadataBasedItemList.js
```diff
@@ -121,23 +163,57 @@ class MetadataBasedItemList extends React.Component<Props, State> {
...
+                            />
+                        )}
+                        {value && isCellBeingEdited && (
+                            <div className="bdl-MetadataBasedItemList-cell--edit">
+                                <Field
+                                    canEdit
+                                    dataKey={value}
+                                    dataValue={value}
+                                    displayName=""
+                                    type={type}
```
sahiga: Minor: can we alphabetize these props, as well as the props for the icons?

src/features/metadata-based-view/__tests__/IconWithTooltip.test.js
```diff
@@ -0,0 +1,43 @@
+import * as React from 'react';
+
+import { IconWithTooltipComponent as IconWithTooltip } from '../IconWithTooltip';
+
+import messages from '../../../elements/common/messages';
+import { CANCEL_ICON_TYPE, EDIT_ICON_TYPE, SAVE_ICON_TYPE } from '../constants';
+
+describe('features/metadata-based-view/IconWithTooltip', () => {
+    const getWrapper = props => mount(<IconWithTooltip {...props} />);
```
sahiga: Since these tests aren't dependent on component lifecycle methods, can we use `shallow` here instead?

src/features/metadata-based-view/__tests__/IconWithTooltip.test.js
```diff
@@ -0,0 +1,43 @@
...
+    const onClick = () => {};
+
+    test('Get IconClose with Tooltip to cancel edited value', () => {
```
sahiga: Minor: the tests in this repo seem to follow a naming convention of "(lowercase) should _verb_..."

src/features/metadata-based-view/__tests__/MetadataBasedItemList.test.js
```diff
@@ -17,16 +17,44 @@ describe('features/metadata-based-view/MetadataBasedItemList', () => {
...
+                    fields: [
+                        {
+                            name: 'type',
+                            type: 'string',
+                            value: 'receipt',
+                        },
+                        {
+                            name: 'amount',
+                            type: 'float',
+                            value: 200,
```
sahiga: Since the type of the value is `float`, should we test with a float instead of an integer?

src/features/metadata-based-view/MetadataBasedItemList.js
manishPh: No. This works because on filename.onClick is an arrow function.
```
filename.onClick = () => this.handleEditIconClick(columnIndex, rowIndex);
```
Reference to `this` is maintained when the onClick handler is called.

src/features/metadata-based-view/IconWithTooltip.js
```diff
@@ -0,0 +1,51 @@
...
+
+const IconWithTooltip = ({ className, onClick, tooltipText, type }: Props): Element<typeof Tooltip> | null => {
+    let icon;
+
+    switch (type) {
+        case CANCEL_ICON_TYPE:
+            icon = <IconClose color={bdlGray62} width={16} height={16} />;
+            break;
+        case EDIT_ICON_TYPE:
+            icon = <IconPencil />;
```
ConradJChan: Does the pencil icon not need the color and dimensions specified?

src/features/metadata-based-view/IconWithTooltip.js
```diff
@@ -0,0 +1,51 @@
...
+import { bdlGray62 } from '../../styles/variables';
+
+import { CANCEL_ICON_TYPE, EDIT_ICON_TYPE, SAVE_ICON_TYPE } from './constants';
+
+type IconType = typeof CANCEL_ICON_TYPE | typeof EDIT_ICON_TYPE | typeof SAVE_ICON_TYPE;
+type Props = {
+    className?: string,
+    onClick: void => void,
+    tooltipText: String,
+    type?: IconType,
```
ConradJChan: Instead of type type prop, what if we just rendered the `children` to keep the `IconWithTooltip` as reusable as possible? That way, you can pass whatever icon you want

src/features/metadata-based-view/MetadataBasedItemList.js
```diff
@@ -21,17 +22,24 @@ import type {
...
+    FILE_NAME_COLUMN_INDEX,
+    FILE_NAME_COLUMN_WIDTH,
+    FIXED_COLUMNS_NUMBER,
+    FIXED_ROW_NUMBER,
+    HEADER_ROW_INDEX,
+    CANCEL_ICON_TYPE,
+    EDIT_ICON_TYPE,
+    SAVE_ICON_TYPE,
+    MIN_METADATA_COLUMN_WIDTH,
+} from './constants';
```
ConradJChan: minor: alphabetize?

src/features/metadata-based-view/MetadataBasedItemList.js
```diff
@@ -121,23 +163,57 @@ class MetadataBasedItemList extends React.Component<Props, State> {
...
+                                onClick={() => this.handleEditIconClick(columnIndex, rowIndex)}
+                            />
+                        )}
+                        {value && isCellBeingEdited && (
+                            <div className="bdl-MetadataBasedItemList-cell--edit">
+                                <Field
+                                    canEdit
+                                    dataKey={value}
+                                    dataValue={value}
+                                    displayName=""
```
ConradJChan: Do we have a value that we can pass in here? This propagates to the text input label which is helpful for leveraging browser autofill capability

src/features/metadata-based-view/MetadataBasedItemList.js
```diff
@@ -121,23 +163,57 @@ class MetadataBasedItemList extends React.Component<Props, State> {
...
+                                type={EDIT_ICON_TYPE}
+                                tooltipText={intl.formatMessage(messages.editLabel)}
+                                onClick={() => this.handleEditIconClick(columnIndex, rowIndex)}
+                            />
+                        )}
+                        {value && isCellBeingEdited && (
+                            <div className="bdl-MetadataBasedItemList-cell--edit">
+                                <Field
+                                    canEdit
+                                    dataKey={value}
```
ConradJChan: Is there a field.key that can be used here?

src/features/metadata-based-view/constants.js
```diff
@@ -0,0 +1,15 @@
+// @flow strict
+
+export const CANCEL_ICON_TYPE = 'cancel';
+export const EDIT_ICON_TYPE = 'edit';
+export const SAVE_ICON_TYPE = 'save';
```
ConradJChan: Should these define their literal flow type? i.e. `export const CANCEL_ICON_TYPE: 'cancel' = 'cancel'`

src/features/metadata-based-view/__tests__/IconWithTooltip.test.js
```diff
@@ -0,0 +1,43 @@
...
+        const props = {
+            className: 'bdl-MetadataBasedItemList-cell--cancelIcon',
+            onClick,
+            tooltipText: messages.cancel,
+            type: CANCEL_ICON_TYPE,
+        };
+        const wrapper = getWrapper(props);
+        expect(wrapper).toMatchSnapshot();
```
ConradJChan: Are there specific assertions you want to make about the mounted (or shallow mounted) component? Seems like snapshots generally get ignored (especially fully mounted ones)

src/features/metadata-based-view/IconWithTooltip.js
manishPh: The default props for size and color look fine for this use case.

src/features/metadata-based-view/__tests__/IconWithTooltip.test.js
manishPh: Because this is a stateless functional component, there are no lifecycle methods and their behaviors to test. The snapshot testing in this case makes sense to me.

src/features/metadata-based-view/MetadataBasedItemList.js
manishPh: There is no key in fields.

src/features/metadata-based-view/MetadataBasedItemList.js
manishPh: This scenario occurs because we are using `<Fields />` component from "features/metadata-instance-editor". If you pass any display name, it breaks the table layout we need. This is something we can tackle when eventually `<Fields />` components will . be moved in its own feature and this prop can be made optional.

manishPh:
cancelEdit
The misalignment of field component is known. @priyajeet , this is what I was talking about. Extracting `<Fields />` into "features/metadata-fields" without styles will solve the issue.

src/features/metadata-based-view/MetadataBasedItemList.scss
```diff
@@ -14,8 +14,26 @@
...
+
+.bdl-MetadataBasedItemList-cell--edit {
+    display: flex;
+    height: 32px;
+    background-color: $white;
+    border: 1px solid $bdl-gray-20;
+    border-radius: 3px;
+
+    .bdl-MetadataBasedItemList-cell--cancelIcon,
+    .bdl-MetadataBasedItemList-cell--saveIcon {
```
sahiga: Since these classes are all modifiers of `.bdl-MetadataBasedItemList-cell`, should they be at the same level, rather than nested?

src/features/metadata-based-view/__tests__/IconWithTooltip.test.js
```diff
@@ -0,0 +1,43 @@
...
+import IconWithTooltip from '../IconWithTooltip';
+
+import messages from '../../../elements/common/messages';
+import { CANCEL_ICON_TYPE, EDIT_ICON_TYPE, SAVE_ICON_TYPE } from '../constants';
+
+describe('features/metadata-based-view/IconWithTooltip', () => {
+    const getWrapper = props => shallow(<IconWithTooltip {...props} />);
+    const onClick = () => {};
+
+    test('get IconClose with Tooltip to cancel edited value', () => {
```
sahiga: Minor: all the other test names in this repo begin with "should"

src/features/metadata-based-view/MetadataBasedItemList.scss
manishPh: Need this added specificity to get over https://github.com/box/box-ui-elements/blob/master/src/styles/common/_buttons.scss#L91

PR #1706, opened 2019-11-05
fix(contentpicker): correctly clear old item when maxSelectable=1 and parent reloaded

### LINKED ISSUE #1705: Folder Picker `maxSelectable: 1` gives the appearance of allowing multiple selections
**Environment:**

- Elements version: latest

**Desktop (please complete the following information):**

- OS: [e.g. Windows 10] macOS 10.14
- Browser [e.g. chrome, safari] Chrome and Safari

**Steps to reproduce the problem:**

Forked version of the official folder picker codepen with `maxSelectable` set to `1`: https://codepen.io/box-platform/pen/WRQLey

1) select `A Sample Folder`
2) navigate to `An Ordered Folder`
3) navigate back to the parent folder
4) select `An Ordered Folder`

Check marks appear next to both `A Sample Folder` and `An Ordered Folder` but Selected Items shows 1 selection. When the selection is saved, the last folder selected is passed to the client (i.e. the behavior is correct, but appearance of the picker is out of sync). 

**What is the expected behavior? (Screenshots can be helpful here)**

When `maxSelectable = 1`, a check mark only appears next to one folder at a time. 

**What went wrong? (Screenshots, console logs, or HAR files can be helpful here)**

Nothing logged in console.

**Additional context**

We are using the bundled JS/CSS assets, rather than the react components directly. 

Animated GIF of the issue on codepen: 

bug

cgrdavies:
This change fixes #1705, which was preventing the prior selection from being properly cleared when the content picker was in single file selection mode and the folder enclosing the previously selected item had been reloaded before changing the selection. 

### Problem
Because the content picker relied on a mutating a reference to the previously selected item, when the items were reloaded from the server, that reference was no longer valid and the item within `currentCollection` was not updated. 

### Solution
Scan the `currentCollection` for an item that matches the prior selection based on the item's id. If found, remove the `selected` keyword from that item directly.


boxcla:
Hi @cgrdavies, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

cgrdavies:
CLA signed

src/elements/content-picker/ContentPicker.js
```diff
@@ -814,9 +814,12 @@ class ContentPicker extends Component<Props, State> {
...
             // Clear out the prior item for single file selection mode
             if (selectedCount > 0 && isSingleFileSelection) {
-                const prior = selectedKeys[0]; // only one item
-                delete selected[prior].selected;
-                delete selected[prior];
+                const priorSelectionIndex = selectedKeys[0]; // only one item
+                const priorSelection = selected[priorSelectionIndex];
+                const matchingItem = items.find(potentialMatch => potentialMatch.id === priorSelection.id);
+
+                if (matchingItem) delete matchingItem.selected;
```
priyajeet: if (matchingItem) {
    ...
}

priyajeet:
BTW there is a possibility this issue applies to multi-selection. Trying to see when it broke, likely when we switched to paging.

priyajeet:
In order to properly fix this, and frankly make it less confusing, we may have to get rid of the reference mutations and keep selected isolated from currentCollection.items. I am trying a few things, will report back.

priyajeet:
Roughly something along these lines might be better https://github.com/box/box-ui-elements/pull/1709/files. We still can use a map to prevent multiple find()s so that its a O(1) lookup or we can just keep the finds.

cgrdavies:
> Roughly something along these lines might be better https://github.com/box/box-ui-elements/pull/1709/files. We still can use a map to prevent multiple find()s so that its a O(1) lookup or we can just keep the find()s as in practice `selected` will always be a small-ish array. Do you want to update your PR following this appraoch?

Thanks for taking a look at this. I agree that removing the reference mutations is a better approach, but I was originally hesitant to do so given the magnitude of the change that would be required. With your outline though I'm happy to update this PR.

cgrdavies:
@priyajeet: should be all set for another look on this!

src/elements/content-picker/itemSelectionHelper.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * @flow
+ * @file Functions to determine whether a given item is selected.
+ * @author Box
+ */
+
+export function containsItem(item: BoxItem) {
```
priyajeet: Can you move these exports to the bottom. It's our new convention. So

```
function foo() {}

function bar() {}

export { foo, bar };
```

priyajeet: Also if you have time can you add a couple of unit tests for this file for these new functions

src/elements/content-picker/itemSelectionHelper.js
```diff
@@ -0,0 +1,15 @@
...
+    const { id, type } = item;
+
+    return (selected: BoxItem) => selected.id === id && selected.type === type;
+}
+
+export function isSelected(item: BoxItem, selectedItems: Array<BoxItem>) {
```
priyajeet: For both of these functions probably should add a return type, so `: boolean` here and `: (item: BoxItem) => boolean` above

src/elements/content-picker/selectionCellRenderer.js
```diff
@@ -8,18 +8,21 @@ import React from 'react';
...
     selectableType: string,
     extensionsWhitelist: string[],
     hasHitSelectionLimit: boolean,
     isRadio: boolean,
+    selected: Array<BoxItem>,
 ): (({ rowData: BoxItem }) => {}) => ({ rowData }: { rowData: BoxItem }) => {
-    const { name = '', selected = false } = rowData;
+    const { name = '' } = rowData;
     const Component = isRadio ? RadioButton : Checkbox;
+    const isSelected = !!isRowSelected(rowData, selected);
```
priyajeet: !! is probably unnecessary since the function returns a boolean

src/elements/content-picker/ItemList.js
```diff
@@ -80,10 +85,18 @@ const ItemList = ({
             return '';
         }
 
-        const { selected, type } = items[index];
-        const isSelectable = isRowSelectable(selectableType, extensionsWhitelist, hasHitSelectionLimit, items[index]);
+        const { type } = items[index];
+        const isSelected = !!isRowSelected(items[index], selected);
```
priyajeet: !! not needed here either

src/elements/content-picker/ItemList.js
```diff
@@ -97,9 +110,11 @@ const ItemList = ({
         index: number,
         rowData: BoxItem,
     }) => {
+        const isSelected = !!isRowSelected(rowData, selected);
```
priyajeet: likewise

src/elements/content-picker/ContentPicker.js
```diff
@@ -881,14 +855,8 @@ class ContentPicker extends Component<Props, State> {
         if (!item[FIELD_SHARED_LINK]) {
             this.changeShareAccess(null, item);
         } else {
-            const { selected } = this.state;
-            const { id, type } = item;
-            const cacheKey = this.api.getAPI(type).getCacheKey(id);
             // if shared link already exists, update the collection in state
             this.updateItemInCollection(item);
-            if (item.selected && item !== selected[cacheKey]) {
```
priyajeet: Actually removing this part may cause a problem, since the item object in question has changed, it needs to be the one that is returned, so one will have to change the choose() function to return items from the updated collection and not whats inside selected.

Further simplification would be to make selected array just an array of IDs and Types, since thats what is really used everywhere to check if a row is selected or not. So that when choose button is pressed, then we build the returned array with latest item objects.

priyajeet:
BTW, if you dont have time to make more changes, we can take over the PR...

src/elements/content-picker/ContentPicker.js
priyajeet: Alternatively, updateItemInCollection() function can be changed to also update `selected` array with the updated item.

cgrdavies:
@priyajeet should be all set from a code perspective. I changed `updateItemInCollection` to replace the matching item in the `selected` array. 

I do have a failing test though that seems unrelated, and I can't immediately figure out what's going on. Any chance someone could take a look at that? I don't have enough overall context to be able to fix it expediently.

sahiga:
Hey @cgrdavies, sorry we've taken so long to merge this PR. I'm looking into the failing tests right now. Is it okay with you if I add commits to your PR? Also, could you sign the new CLA? Thanks!

PR #1709, opened 2019-11-05
fix(content-picker): selection bug

priyajeet:


boxcla:

src/elements/content-picker/ContentPicker.js
```diff
@@ -102,7 +102,7 @@ type State = {
     isUploadModalOpen: boolean,
     rootName: string,
     searchQuery: string,
-    selected: { [string]: BoxItem },
+    selected: Array<BoxItem>,
```
jstoffan: Ideally, we should have a single source of truth for the actual item objects. Can this be an array of item ids that we use to look up the item object from `currentCollection`, instead?

src/elements/content-picker/ContentPicker.js
```diff
@@ -1152,7 +1137,7 @@ class ContentPicker extends Component<Props, State> {
         }: State = this.state;
         const { id, offset, permissions, totalCount }: Collection = currentCollection;
         const { can_upload }: BoxItemPermission = permissions || {};
-        const selectedCount: number = Object.keys(selected).length;
+        const selectedCount: number = selected.length;
```
jstoffan: Should this use destructuring?

src/elements/content-picker/ItemList.js
```diff
@@ -80,10 +84,17 @@ const ItemList = ({
             return '';
         }
 
-        const { selected, type } = items[index];
-        const isSelectable = isRowSelectable(selectableType, extensionsWhitelist, hasHitSelectionLimit, items[index]);
+        const { id: itemId, type } = items[index];
+        const isSelected = !!selected.find(sel => sel.id === itemId && sel.type === type);
```
jstoffan: This logic seem to be duplicated in `onRowClick`

src/elements/content-picker/selectionCellRenderer.js
```diff
@@ -15,11 +15,13 @@ export default (
     extensionsWhitelist: string[],
     hasHitSelectionLimit: boolean,
     isRadio: boolean,
+    selected: Array<BoxItem>,
```
jstoffan: Should we be passing the entire selected collection into each row? Can the parent pass in an `isSelected` value, instead?

src/elements/content-picker/selectionCellRenderer.js
```diff
@@ -15,11 +15,13 @@ export default (
...
 ): (({ rowData: BoxItem }) => {}) => ({ rowData }: { rowData: BoxItem }) => {
-    const { name = '', selected = false } = rowData;
+    const { name = '', id, type } = rowData;
     const Component = isRadio ? RadioButton : Checkbox;
+    const isSelected = !!selected.find(sel => sel.id === id && sel.type === type);
```
jstoffan: Duplicated logic from the parent component?

PR #1710, opened 2019-11-06
fix(classification): classification tooltip not showing in IE11

priyajeet:
An issue with certain versions of IE11, where rendering a tooltip inside the button breaks. So making the tooltip be the wrapper for button.

### DIFF
--- a/src/features/classification/ClassifiedBadge.js
+++ b/src/features/classification/ClassifiedBadge.js
@@ -1,26 +1,47 @@
 // @flow
 import * as React from 'react';
 
+import PlainButton from '../../components/plain-button';
 import Tooltip from '../../components/tooltip';
 import IconSecurityClassification from '../../icons/general/IconSecurityClassification';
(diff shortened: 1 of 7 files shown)


boxcla:

src/features/classification/ClassifiedBadge.js
```diff
@@ -1,26 +1,47 @@
...
         />
-    </Tooltip>
-);
+    );
+    return (
+        <Tooltip isDisabled={isTooltipDisabled} isTabbable={false} position={tooltipPosition} text={tooltipText}>
+            {isClickable ? (
+                <PlainButton
+                    className="bdl-ClassifiedBadge-editButton"
+                    data-resin-target="editclassification"
```
jpfranco: I thought we were using camel case for these

priyajeet: Hmm, unsure but just using what we had before

PR #1711, opened 2019-11-06
style(media): Correctly size Media body flex container for MS Edge

ConradJChan:
Before:
Screen Shot 2019-11-06 at 10 06 57 AM

After:
Screen Shot 2019-11-06 at 10 07 11 AM


boxcla:

PR #1713, opened 2019-11-07
fix(metadata-sidebar): bind onApiError properly

mickr:
onApiError wasn't bound properly to the class so calls to extract onError would fail.


boxcla:

PR #1715, opened 2019-11-08
feat(select-field): Use Popper to implement dropdown

ConradJChan:
Uses react-popper to implement dropdown for `BaseSelectField` which affects `SingleSelectField` and `MultiSelectField`.

TODO:
- [x] Fix unit tests

BREAKING CHANGE


boxcla:

package.json
```diff
@@ -234,6 +234,7 @@
         "react-intl": "^2.8.0",
         "react-measure": "^2.3.0",
         "react-modal": "^3.8.1",
+        "react-popper": "^1.3.4",
```
jstoffan: We'll need to add this to our peer dependencies list, as well.

src/components/select-field/BaseSelectField.js
```diff
@@ -381,34 +399,47 @@ class BaseSelectField extends React.Component<Props, State> {
             listboxProps['aria-multiselectable'] = true;
```
jstoffan: Is `listboxProps` still used in this class? It looks like it was moved into SelectDropdownList.

src/components/select-field/BaseSelectField.js
```diff
@@ -365,8 +368,23 @@ class BaseSelectField extends React.Component<Props, State> {
...
+        if (!this.selectButtonRef) {
+            return 0;
+        }
+
+        const { width } = this.selectButtonRef.getBoundingClientRect();
+
+        return width;
+    };
+
+    setSelectButtonRef = setReferenceNode => node => {
```
jstoffan: I think we're moving toward `React.createRef` rather than a callback style. Can we store `setReferenceNode` on the instance and call it in `componentDidMount`, instead? Or is this the prescribed way to get the reference ref per the Popper docs?

src/components/select-field/SelectDropdownList.js
```diff
@@ -0,0 +1,57 @@
...
+        const { children, popperRef, style, placement, isScrollable, multiple, selectFieldID } = this.props;
+
+        const listboxProps = {};
+        if (multiple) {
+            listboxProps['aria-multiselectable'] = true;
+        }
+
+        return (
+            <ul
+                ref={popperRef}
```
jstoffan: Can we use `React.forwardRef` in this class, as well, to avoid this renaming?

src/components/select-field/SelectDropdownList.js
```diff
@@ -0,0 +1,57 @@
...
+
+import type { SelectOptionValueProp } from './props';
+
+export const OVERLAY_SCROLLABLE_CLASS = 'bdl-SelectField-overlay--scrollable';
+
+type Props = {
+    children: React.Node,
+    isScrollable?: boolean,
+    multiple?: boolean,
+    placement: string,
```
jstoffan: Seems like this could/should be optional with a default value?

src/components/select-field/BaseSelectField.js
```diff
@@ -381,34 +399,47 @@ class BaseSelectField extends React.Component<Props, State> {
...
-                            {this.renderSelectOptions()}
-                        </ul>
-                    </div>
-                </div>
+                <Manager>
+                    <Reference>
+                        {({ ref }) =>
+                            // Need to store the select button reference so we can calculate the button width
+                            // in order to set it as the min width of the dropdown list
+                            React.cloneElement(this.renderSelectButton(), { ref: this.setSelectButtonRef(ref) })
```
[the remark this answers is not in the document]
priyajeet: Or can we just do this in the renderSelectButton()
```jsx
<SelectButton {...buttonProps} error={error} ref={this.selectButtonRef}>
    {buttonText}
</SelectButton>
```

src/components/select-field/BaseSelectField.js
```diff
@@ -381,34 +399,47 @@ class BaseSelectField extends React.Component<Props, State> {
             listboxProps['aria-multiselectable'] = true;
         }
 
+        const dropdownPlacement = isRightAligned ? 'bottom-end' : 'bottom-start';
```
priyajeet: We'll eventually want these as importable constants, since they will be shared with all pop-ing things like drop downs and tooltips etc.

src/components/select-field/SelectDropdownList.js
```diff
@@ -0,0 +1,57 @@
...
+    popperRef: (?HTMLElement) => void,
```
priyajeet: HTMLUlElement?

src/components/select-field/SelectDropdownList.js
```diff
@@ -0,0 +1,57 @@
...
+    scheduleUpdate?: () => void,
```
priyajeet: Will there be a case when this is undefined?

src/components/select-field/BaseSelectField.js
```diff
@@ -381,34 +399,47 @@ class BaseSelectField extends React.Component<Props, State> {
...
+                                const buttonWidth = this.getSelectButtonWidth();
+                                const updatedStyle = { ...style, minWidth: buttonWidth };
+
+                                return (
+                                    <SelectDropdownList
+                                        popperRef={ref}
+                                        style={updatedStyle}
+                                        data-placement={placement}
+                                        isScrollable={isScrollable}
+                                        multiple
```
priyajeet: multiple={multiple}

src/components/select-field/SelectDropdownList.js
```diff
@@ -0,0 +1,57 @@
...
+    selectFieldID: String,
+    selectedValues: Array<SelectOptionValueProp>,
+    style: Object,
+};
+
+class SelectDropdownList extends React.Component<Props> {
+    componentDidUpdate({ selectedValues: prevSelectedValues }) {
+        const { scheduleUpdate, selectedValues } = this.props;
+        if (scheduleUpdate && prevSelectedValues !== selectedValues) {
```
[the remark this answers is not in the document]
priyajeet: or maybe you just need a length check, since the button only changes width, when selected values change in number

src/components/select-field/SelectDropdownList.js
priyajeet: I kinda like using named ref (innerRef etc...) more than forwardRef because the flow types are so hard to read/understand and people just end up dumping <any>

src/components/select-button/SelectButton.js
```diff
@@ -12,19 +12,24 @@ type Props = {
...
-            })}
-            disabled={isDisabled}
-            type="button"
-            {...rest}
-        >
-            {children}
-        </button>
-    </Tooltip>
-);
+const SelectButton = React.forwardRef<Props, React.Ref<any>>((props: Props, ref: React.Ref<any>) => {
```
priyajeet: React.Ref<HTMLButtonElement>>?

src/components/select-field/BaseSelectField.js
```diff
@@ -381,34 +399,47 @@ class BaseSelectField extends React.Component<Props, State> {
...
+                        <Popper placement={dropdownPlacement} className={OVERLAY_WRAPPER_CLASS}>
+                            {({ ref, style, placement, scheduleUpdate }) => {
+                                const buttonWidth = this.getSelectButtonWidth();
+                                const updatedStyle = { ...style, minWidth: buttonWidth };
+
+                                return (
+                                    <SelectDropdownList
+                                        popperRef={ref}
+                                        style={updatedStyle}
+                                        data-placement={placement}
```
priyajeet: placement={placement}

src/components/select-field/SelectDropdownList.js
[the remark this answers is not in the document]
priyajeet: Unsure about this - you could also change this component to a PureComponent and rely on the auto-shallow-prop equality to render. You only want to re-render when style prop changes, and you can pass in the unmodified style prop. If no props changed, then componentDidUpdate() should not be called.

src/components/select-field/BaseSelectField.js
ConradJChan: I did it this way because Popper's `Reference` uses a render prop in which it passes in its own `refHandler` (https://github.com/FezVrasta/react-popper/blob/master/src/Reference.js#L35). Is there a way in which I can use `React.createRef` and also not have it overwritten by the `refHandler` given to me by Popper? @priyajeet @jstoffan

src/components/select-field/BaseSelectField.js
ConradJChan: I'll create this in  `/common/constants.js` for now, unless there's a better place you had in mind?

src/components/select-field/BaseSelectField.js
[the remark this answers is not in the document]
ConradJChan: I'll try passing it into `renderSelectButton`, I'm not convinced that I can use `React.createRef` for selectButtonRef because I need to invoke Popper's refHandler to associate the popper with the button, see my reply here: https://github.com/box/box-ui-elements/pull/1715/files#r344312163

src/components/select-field/SelectDropdownList.js
ConradJChan: This `placement` is just a description used to fill the `data-placement` attribute of the `<ul>`, so it seems like if it's optional without a default value that should be fine. Thoughts?

src/components/select-field/SelectDropdownList.js
ConradJChan: Is that a thing? I can't find a flow type for that

src/components/select-field/SelectDropdownList.js
ConradJChan: Shouldn't be, I'll remove the `?`

src/components/select-field/BaseSelectField.js
jstoffan: Should we have a PopperComponent, similar to TetherComponent, that can encapsulate the options and their string values?

src/components/select-field/BaseSelectField.js
```diff
@@ -352,7 +357,13 @@ class BaseSelectField extends React.Component<Props, State> {
                     <div className="select-option-check-icon">
                         {isSelected ? <IconCheck height={16} width={16} /> : null}
                     </div>
-                    {optionRenderer ? optionRenderer(item) : displayText}
+                    {optionRenderer ? (
+                        optionRenderer(item)
+                    ) : (
+                        <span className="select-option-text" title={displayText}>
```
jstoffan: Can this be a the `defaultProp` value for `optionRenderer`?

src/components/select-field/BaseSelectField.js
```diff
@@ -7,8 +7,10 @@ import { scrollIntoView } from '../../utils/dom';
 import IconCheck from '../../icons/general/IconCheck';
 import SelectButton from '../select-button';
 import DatalistItem from '../datalist-item';
+import PopperComponent from '../popper';
+import SelectFieldDropdown from './SelectFieldDropdown';
 import type { SelectOptionValueProp, SelectOptionProp } from './props';
-import { OVERLAY_WRAPPER_CLASS } from '../../constants';
+import { PLACEMENT_BOTTOM_END, PLACEMENT_BOTTOM_START } from '../popper/constants';
```
jstoffan: Can we hide these details inside the PopperComponent?

src/components/select-field/SelectField.scss
```diff
@@ -29,3 +29,17 @@
     max-height: 340px; // show at most 12 items in the dropdown menu
     overflow-y: auto;
 }
+
+.bdl-SelectField {
+    .bdl-SelectFieldDropdown {
```
jstoffan: Is this nesting needed? It'll increase specificity and shouldn't be required since we're using SUIT.

jstoffan:
I think this needs to be marked as a `BREAKING CHANGE` due to the peerDeps being updated. @priyajeet, thoughts?

src/components/select-field/BaseSelectField.js
ConradJChan: Makes sense!

src/components/select-field/BaseSelectField.js
ConradJChan: I think the PopperComponent should be pretty general use, exposing something like a  `isRightAligned` prop to PopperComponent feels a bit too dropdown-y

scripts/styleguide.styles.scss
```diff
@@ -42,3 +42,9 @@
         height: 600px;
     }
 }
+
+.flex-center-center {
```
jstoffan: I realize this is just the example styles, but ideally this should be a mixin rather than a class, or at least a utility class, like `.bdl-u-flexCenter`.

src/components/popper/PopperComponent.js
```diff
@@ -27,9 +27,11 @@ const PopperComponent = (props: Props) => {
             {isOpen && (
                 <Popper placement={popperPlacement} modifiers={modifiers}>
                     {({ ref, style, placement, scheduleUpdate }) => {
+                        const { style: contentStyles } = popperContent.props;
+                        const updatedStyles = { ...contentStyles, ...style };
```
jstoffan: Minor, but not sure this variable is needed, since its value is used immediately and only once.

src/components/popper/README.md
```diff
@@ -0,0 +1,40 @@
...
+It expects two children where the first child is the reference element and the second child is the popper content.
+
+### Examples
+
+**auto placement**
+
+```js
+<div style={{position: 'relative'}}>
+    <PopperComponent isOpen>
+        <div className="flex-center-center" style={{border: "1px solid black", width: '200px', height: '100px', cursor: 'pointer'}}>Reference Element</div>
```
jstoffan: I think the inline styles here make this a bit difficult to read. Can these examples share a single JS block so they can share a style variable?

src/components/popper/__tests__/PopperComponent.test.js
```diff
@@ -0,0 +1,53 @@
...
+
+    test('should apply the placement to the Popper', () => {
+        const wrapper = shallow(
+            <PopperComponent isOpen placement="bottom-end">
+                <div>Reference element</div>
+                <div>Popper content</div>
+            </PopperComponent>,
+        );
+
+        expect(wrapper.find(Reference).length).toBe(1);
```
jstoffan: You can also write this as `expect(wrapper.exists(Reference)).toBe(true)`.

src/components/popper/PopperComponent.js
```diff
@@ -0,0 +1,45 @@
...
+
+type Props = {
+    children: React.Node,
+    isOpen?: boolean,
+    modifiers?: Modifiers,
+    placement?: Placement,
+};
+
+const PopperComponent = (props: Props) => {
+    const { children, isOpen, modifiers, placement: popperPlacement = PLACEMENT_AUTO } = props;
```
priyajeet: Move the placement defaulting to PopperComponent.defaultProps below and remove the ? From it's type

src/components/popper/README.md
```diff
@@ -0,0 +1,46 @@
+The `PopperComponent` component is used to append an arbitrary popover panel to an on-screen element. It relies
```
priyajeet: Optional: convert to storybook  story instead if you want to

src/components/popper/PopperComponent.js
priyajeet: Definitely test a popper-ed component within something like a react-virtualized table  row that sets overflow hidden. The pop-up may get cut off. In that case you need to pass in an additional prop to this component like isPortaled. And if true, you portal the content.

priyajeet: For reference see https://github.com/FezVrasta/react-popper/issues/79 and https://github.com/FezVrasta/popper.js/issues/567

src/components/select-field/SelectFieldDropdown.js
```diff
@@ -0,0 +1,57 @@
...
+import classNames from 'classnames';
+
+import type { SelectOptionValueProp } from './props';
+import type { PopperChildrenProps } from '../popper/props';
+
+export const OVERLAY_SCROLLABLE_CLASS = 'bdl-SelectField-overlay--scrollable';
+
+type Props = {
+    children: React.Node,
+    innerRef?: React.Ref<any>,
```
priyajeet: Should this be UL element? Instead of any

src/components/select-field/SelectField.scss
```diff
@@ -29,3 +29,15 @@
     max-height: 340px; // show at most 12 items in the dropdown menu
     overflow-y: auto;
 }
+
+.bdl-SelectFieldDropdown {
```
priyajeet: You can choose to have this as bdl-SelectField-dropdown too considering the drop-down component is not to be used on its own. We do it for other components too that span file boundaries.

src/components/select-field/BaseSelectField.js
```diff
@@ -66,14 +70,22 @@ type State = {
     shouldScrollIntoView: boolean,
 };
 
-export const OVERLAY_SCROLLABLE_CLASS = 'bdl-SelectField-overlay--scrollable';
+function defaultOptionRenderer({ displayText }: SelectOptionProp) {
```
priyajeet: Maybe move this out to its own file with its own mini test

src/components/select-field/SelectField.scss
mickr: I think I'll keep the same as I'm not entirely sure of @ConradJChan's intent

src/components/select-field/BaseSelectField.js
mickr: It would make sense to move it out, but not sure if it is going to be used anywhere else.

src/components/popper/PopperComponent.js
mickr: Can this be added in a small follow up PR so I can merge this larger PR?

src/components/select-field/SelectFieldDropdown.js
mickr: `HTMLUListElement` is incompatible with `React.Component` so it fails the flow-type.

PR #1722, opened 2019-11-13
docs(icons): switch general icons to icon example grid

tjgupta:
This ports the general icons to the icon example grid.

Screen Shot 2019-11-12 at 9 47 28 PM

It also normalizes the icon flow types where possible to share the common `Icon` type.


boxcla:

PR #1725, opened 2019-11-13
build: simplify dist tag and branch determination

priyajeet:
- Simplifies dist and branch determination because now its provided explicitly via package.json
- Adding a new type of release should now just involve touching package.json by utilizing DIST and BRANCH env variables
- Gets rid of the need of having a release branch


boxcla:

PR #1726, opened 2019-11-13
fix(uploads): resuming an upload while disconnected

stevegcoder:
Resuming an upload while disconnected will no longer reset the upload from the beginning.  Only 4xx statuses that indicate a problem with session retrieval will reset the upload


boxcla:

src/api/uploads/MultiputUpload.js
```diff
@@ -536,8 +538,8 @@ class MultiputUpload extends BaseMultiput {
         } else if (errorData && errorData.status >= 500) {
             this.retryTimeout = setTimeout(this.getSessionInfo, 2 ** this.numResumeRetries * MS_IN_S);
             this.numResumeRetries += 1;
-        } else {
-            // Restart upload process for errors resulting from invalid session
+        } else if (errorData && errorData.status >= 400 && errorData.status < 500) {
+            // Restart upload process for errors resulting from invalid/expired session or no permission
```
borajimin: The ticket says that we should not retry resume if (and only if) the session is expired but the comment says something different. Error code 410 is supposed to be returned when the upload session has expired according to this: https://developer.box.com/en/guides/api-calls/permissions-and-errors/common-errors/#410-gone. Shouldn't we just look out for the error code 410? Could you explain more on the ticket?

stevegcoder: Yeah I took a look at all the 4xx errors and it seems like they all warrant restarting the session.  429 is already handled and doesn't restart the session.  What do you think @danubevictoria?

danubevictoria: I think we should handle for all 4xx errors because we don't want the user to keep attempting to resume if they can't and we want to give them a way to still upload their file, so restarting seems appropriate. The 4xx's that I could see being exceptions could be 409 and 429. 429 is handled above. Did we review any others? 415 might also seem weird to restart if its unsupported file type.

Also, this is assuming we will get back 500 when we disconnect from the internet, correct? We don't want to restart in that case but attempt to resume.

stevegcoder: Yeah, I reviewed them all.  I think we can define different behaviors for each but figured it would be easier to understand if all 4xx errors were handled the same way.

Yeah 500 just resumes.  A client side internet disconnect is handled in the else clause since the error response is undefined so there is no status.

src/api/uploads/MultiputUpload.js
```diff
@@ -536,8 +538,8 @@ class MultiputUpload extends BaseMultiput {
         } else if (errorData && errorData.status >= 500) {
```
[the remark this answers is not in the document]
stevegcoder: Yeah that works, I can make the change.

src/api/uploads/MultiputUpload.js
```diff
@@ -558,6 +557,11 @@ class MultiputUpload extends BaseMultiput {
                 fileId: this.fileId,
             };
             this.upload(uploadOptions);
+        } else {
+            // Handle internet disconnects (error.request && !error.response) and (!error.request)
+            // Also handle any 500 error messages
+            this.retryTimeout = setTimeout(this.getSessionInfo, 2 ** this.numResumeRetries * MS_IN_S);
```
mickr: Should this be handled by the `retryNumOfTimes` util in the `utils/function.js` file?

stevegcoder: I think the code could be refactored to use it, but since it's already been tested and works, I think it may be better to keep it.  It was also existing code in box-ui-elements before my changes.

PR #1729, opened 2019-11-14
feat(icons): integrate new file and folder icons

booc0mtaco:
Screen Shot 2019-11-13 at 9 56 46 PM
update icons and examples in documentation. also update snapshots accordingly


boxcla:

src/icons/file/README.md
```diff
@@ -21,111 +21,107 @@ const IconsExample = require('../../../examples/src/IconsExample').default;
...
-    },
-    {
-        name: 'IconFileDocument',
-        component: require('./IconFileDocument').default,
+        component: require('../imported/Icon/IconFileDefault32').default,
     },
     {
         name: 'IconFileDwg',
-        component: require('./IconFileDwg').default,
+        component: require('../imported/Icon/IconFIleDwg32').default,
```
jstoffan: Typo?

src/icons/file/README.md
```diff
@@ -21,111 +21,107 @@ const IconsExample = require('../../../examples/src/IconsExample').default;
 const icons = [
     {
         name: 'IconFileAudio',
-        component: require('./IconFileAudio').default,
+        component: require('../imported/Icon/IconFileAudio32').default,
```
jstoffan: We don't currently differentiate imported icons, nor do we use PascalCase for directory names. It seems like these new components should live directly in `src/icons/file` rather than `src/icons/imported/Icon`.

src/icons/imported/Icon/BoxLogo.js
```diff
@@ -0,0 +1,22 @@
+// @flow
+/* eslint-disable react/jsx-sort-props */
+import * as React from 'react';
+import AccessibleSVG from '../../accessible-svg';
+
+type SvgProps = {
```
jstoffan: Can this type be shared between the components?

src/icons/imported/Icon/IconBoxApp512.js
```diff
@@ -0,0 +1,24 @@
...
+import AccessibleSVG from '../../accessible-svg';
+
+type SvgProps = {
+    height?: number,
+    title?: string | React.Element<any>,
+    width?: number,
+};
+
+const IconBoxApp512 = (props: SvgProps) => (
+    <AccessibleSVG width={32} height={32} viewBox="0 0 512 512" {...props}>
```
jstoffan: The component file name has `512` in its name, though it seems like it's 32x32px. Do we need the dimensions in the file name?

src/icons/imported/Icon/IconBoxApp512.js
```diff
@@ -0,0 +1,24 @@
...
+        <g fill="none" fillRule="evenodd">
+            <rect fill="#0061D5" width={512} height={512} rx={118} />
```
jstoffan: Can we use predefined variables for the color values?

src/icons/imported/Icon/IconBoxApp512.js
```diff
@@ -0,0 +1,24 @@
+// @flow
+/* eslint-disable react/jsx-sort-props */
```
jstoffan: Is this necessary?

src/icons/file-icon/FileIcon.js
```diff
@@ -1,40 +1,39 @@
...
-import IconFilePowerpointPresentation from '../file/IconFilePowerpointPresentation';
-import IconFilePresentation from '../file/IconFilePresentation';
-import IconFileSpreadsheet from '../file/IconFileSpreadsheet';
-import IconFileText from '../file/IconFileText';
-import IconFileThreeD from '../file/IconFileThreeD';
-import IconFileVector from '../file/IconFileVector';
-import IconFileVideo from '../file/IconFileVideo';
-import IconFileWordDocument from '../file/IconFileWordDocument';
-import IconFileZip from '../file/IconFileZip';
+import IconFileAudio from '../imported/Icon/IconFileAudio32';
```
priyajeet: I think file icons should remain logically under "file"

src/icons/file-icon/FileIcon.js
```diff
@@ -1,40 +1,39 @@
...
+import IconFileWordDocument from '../imported/Icon/IconFileWord32';
+import IconFileZip from '../imported/Icon/IconFileZip32';
 
 const Components = {
     IconFileAudio,
     IconFileBoxNote,
     IconFileCode,
     IconFileDefault,
-    IconFileDocument,
+    IconFileDocument: IconFileText,
```
priyajeet: Can we just rename the original file to say document?

src/icons/imported/Icon/IconBoxApp512.js
priyajeet: Yeah we should use the imported vars. If these are getting auto-generated, then something like this can work

```jsx
import * as colors from '../../styles/variables';

<rect fill={colors.bdlBoxBlue} width={512} height={512} rx={118} />
```

src/icons/file/README.md
booc0mtaco: the originating file had this in, so had to push an update then re-import. should be fixed now

src/icons/imported/Icon/BoxLogo.js
booc0mtaco: Possibly, although we haven't done that yet for many of the other icons. I was considering introducing a type alongside `AccessibleSVG` and having the importer reference that type. Something like that?

src/icons/imported/Icon/IconBoxApp512.js
booc0mtaco: Added this to account for the props in the raw SVGs not being in any specific order. From design's (and their tool's) standpoint, I don't think they have as much control over the order of SVG attributes. And one of the goals in the importing was to keep as much the same as possible, so I introduced this exception.

Related question to me is whether we should annotate these as "auto-generated" somehow? so that people don't go in and tweak them.

src/icons/file-icon/FileIcon.js
booc0mtaco: Looks like the original Icons included a duplicate, so there was no export for `IconFileDocument32` in the new set. I remapped this one to keep the current behavior in any consumers. Lemme know if I misunderstood

src/icons/imported/Icon/BoxLogo.js
jstoffan: I think @tjgupta just did [something similar](https://github.com/box/box-ui-elements/pull/1722/files#diff-0289c5bf2a1b179fed23b0670fc2bcaeR7) for our general icons.

src/icons/imported/Icon/IconBoxApp512.js
booc0mtaco: From design's perspective, the number in the name is just to mark the resolution, not the default size it should use. The import script is currently naïve in this regard (it always sets the default to 32x32), tho it does have the ability to account for per-asset details (and viewbox metrics could be part of that).

Thoughts?

src/icons/imported/Icon/IconBoxApp512.js
booc0mtaco: I was going back and forth on that for a bit actually. these icons have some colors that are defined, and others that aren't (due to branding restrictions, etc.) I err'd on the side of simplicity since not every one of these icons will include colors that are part of BDL and we don't want to import the color module(s) if it won't get used.

That said, these imports will eventually contain other icons that **should** have swapped values for sure, so I planned to add that in once we get got the first set of non-file/folder icons.

Let me know if you'd rather see that now

src/icons/file/README.md
booc0mtaco: There will eventually be other assets being imported that aren't specifically file or folder icons, so I wanted to sequester the new ones from all the others for that and a few other reasons:

- since the organization of the source asset repository might change, the import clears the contents then rebuilds it
- the directory structure (including the naming) is copied from the source repository as-is. File names are "PascalCase"d for consistency with component naming.

If we object specifically to `Icon`, I could request that change, and the imports would become `../imported/another-path/`... to whatever depth was suggested.

src/icons/imported/Icon/BoxLogo.js
booc0mtaco: Oh this is *very* new. Thanks for pointing out.

booc0mtaco: actually, looks like I added the `Icon` flow type earlier this year, but it wasn't used across the board yet. I remember Tim and I working thru all that back then.

I'll update this to import that type instead of in-lining them, which should be fine.

src/icons/imported/Icon/IconBoxApp512.js
booc0mtaco: went ahead and added ability to read/use the viewBox values for the default prop values to the importer

booc0mtaco:
@priyajeet @jstoffan any thoughts on those comments above?

src/icons/file/README.md
booc0mtaco: @jstoffan perhaps using something other than "imported" is better here? Other ideas (considering the details I mentioned above)

src/icons/imported/Icon/IconBoxApp512.js
jstoffan: Is this eslint-disable still needed with the external type declaration?

src/icons/file/README.md
jstoffan: So the tool we're using to import images clears the output directory each time, basically? That's not ideal, but if it's unavoidable, we should still try to respect the organization of the icon directories. Can we use `src/icons/file/imported` for now?

src/icons/file-icon/FileIcon.js
booc0mtaco: I removed the extra "Icon" path from the path it deposits the assets, but didn't move where it goes since there will be non-file/-folder icons that get imported

src/icons/imported/Icon/IconBoxApp512.js
booc0mtaco: Yah, currently we do not have a way to coerce the ordering of the props in the exported SVGs, and this rule is complaining about "width" coming before "height" in the converted `AccessibleSVG` prop list.

booc0mtaco:
@jstoffan @priyajeet FYI i'm creating follow-up work to annotate these as auto-generated, and will work with design on annotating icons with appropriate color variables & class names. will keep you posted

PR #1736, opened 2019-11-18
fix(i18n): formatted comp message

ehoogerbeets:
Fixed a bug where a Param component inside of another component or HTML inside of the children of a FormattedCompMessage would not have its value substituted into the string for translated versions of that string.

Had to unmock the IntlProvider class from react-intl in the FormattedCompMessage tests so that we could use the real react-intl code to do translations. This bug didn't show up if we only used the English string, and the mocked version of the react-intl stuff doesn't support real translations. This also meant we had to do real tests for the Russian plurals tests using react-intl, which is better anyways because it works correctly and we're testing the real thing.

It also meant we had to do a little fancy footwork to get enzyme to work correctly because when you test with real react-intl, it wraps other classes and injects the "intl" prop into them by searching for a IntlProvider context somewhere up the element tree and finding the locale that way. That context doesn't exist when you are testing a little snippet of code (as opposed to a whole app), so when you try to "mount" your code that with enzyme, it ends up not not finding the context and therefore not having the correct locale. To get around that, we inject the intl prop manually instead so that we don't need that context to be available in the tree. We do that with the call "mountWithIntl".

### DIFF
--- a/src/components/i18n/Composition.js
+++ b/src/components/i18n/Composition.js
@@ -27,7 +27,7 @@ class Composition {
                 if (Array.isArray(element)) {
                     element.forEach(subelement => this.recompose(subelement));
                 } else if (element) {
-                    if (element.type === 'Param') {
+                    if (element.type === 'Param' || element.type.name === 'Param') {
                         this.ma.addParam(element);
                     } else {
                         this.ma.push(element);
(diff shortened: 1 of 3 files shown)


boxcla:

PR #1738, opened 2019-11-19
feat(classification): Add Classification Security Controls

jpfranco:
This PR will add the `SecurityControls` component, which will display a summary of restrictions applied to a specific classification.

The component can be displayed as a short, one-sentence summary, a bullet list of restrictions, and a short summary that displays the list of restrictions on a tooltip.

This PR is getting a bit large, so I will address these pending items on a separate one:
- Integrating with the `Classification` component so that the `SecurityControls` are displayed automatically when restrictions are passed in the props
- `Tooltip` has an issue with scrollable containers so I'll need to find a workaround or change the tooltip implementation

Screen Shot 2019-11-19 at 1 12 48 PM
Screen Shot 2019-11-19 at 1 17 15 PM
Screen Shot 2019-11-19 at 1 19 14 PM


boxcla:

i18n/en-US.properties
```diff
@@ -1002,6 +1002,64 @@ boxui.searchForm.searchLabel = Search query
 boxui.securityCloudGame.instructions = For security purposes, please drag the white cloud into the dark cloud.
 # Success message shown when a user successfully drags the cloud into position.
 boxui.securityCloudGame.success = Success!
+# Bullet point that summarizes application download blocked restriction applied to classification
+boxui.securityControls.appDownloadBlock = Some application restrictions apply.
+# Bullet point that summarizes application download restriction applied to classification
+boxui.securityControls.appDownloadList = Some applications will be restricted; {appNames}
+# Bullet point that summarizes application download restriction applied to classification. This variation is used when the list of applications is longer than the configured threshold
+boxui.securityControls.appDownloadListOverflow = Only select applications are allowed; {appNames} +{remainingAppCount} more
```
priyajeet: ; expected?

i18n/en-US.properties
```diff
@@ -1002,6 +1002,64 @@ boxui.searchForm.searchLabel = Search query
...
+# Bullet point that summarizes mobile download restrictions applied to classification, when restriction applies to managed users except Owners/Co-Owners/Editors
+boxui.securityControls.mobileDownloadOwnersEditors = Download restricted on mobile, except Owners/Co-Owners/Editors.
+# Bullet point that summarizes collaborators shared link restriction applied to classification
+boxui.securityControls.sharingCollabAndCompanyOnly = Shared links allowed for collaborators only.
+# Bullet point that summarizes shared link restriction applied to classification
+boxui.securityControls.sharingCollabOnly = Shared links cannot be made publicly accessible.
+# Short summary displayed for classification when sharing, download and app download restrictions are applied to it
+boxui.securityControls.shortAllRestrictions = Sharing, download and app and restrictions apply
+# Short summary displayed for classification when an application download restriction is applied to it
+boxui.securityControls.shortApp = Application restrictions apply
```
priyajeet: application --> app to match other places?

i18n/en-US.properties
priyajeet: some places we use `app`, other places `applications`. Should one make them consistent?

src/features/classification/security-controls/SecurityControlsItem.scss
```diff
@@ -0,0 +1,33 @@
+@import '../../../styles/variables';
+
+.bdl-SecurityControlsItem {
+    padding-left: 1em;
```
priyajeet: em on purpose?

src/features/classification/security-controls/messages.js
```diff
@@ -0,0 +1,177 @@
...
+        description: 'Bullet point that summarizes application download restriction applied to classification',
+        id: 'boxui.securityControls.appDownloadList',
+    },
+    appDownloadListOverflow: {
+        defaultMessage: 'Only select applications are allowed; {appNames} +{remainingAppCount} more',
+        description:
+            'Bullet point that summarizes application download restriction applied to classification. This variation is used when the list of applications is longer than the configured threshold',
+        id: 'boxui.securityControls.appDownloadListOverflow',
+    },
+    // Web Download Restrictions
```
jpfranco: Confirmed that it could potentially cause translation issues to translate the platform (e.g. web, desktop, mobile) string separately and pass it as a parameter, hence the need to duplicate these for all three platforms

src/features/classification/security-controls/utils.js
```diff
@@ -0,0 +1,178 @@
...
+import { ACCESS_POLICY_RESTRICTION, DOWNLOAD_CONTROL, LIST_ACCESS_LEVEL, SHARED_LINK_ACCESS_LEVEL } from './constants';
+
+const { SHARED_LINK, DOWNLOAD, EXTERNAL_COLLAB, APP } = ACCESS_POLICY_RESTRICTION;
+const { DESKTOP, MOBILE, WEB } = DOWNLOAD_CONTROL;
+const { BLOCK, WHITELIST, BLACKLIST } = LIST_ACCESS_LEVEL;
+const { COLLAB_ONLY, COLLAB_AND_COMPANY_ONLY } = SHARED_LINK_ACCESS_LEVEL;
+
+const getShortSecurityControlsMessage = (accessPolicyRestrictions: AccessPolicyRestrictions): ?MessageDescriptor => {
+    const { sharedLink, download, externalCollab, app } = accessPolicyRestrictions;
+    // Shared link and external collab restrictions are grouped
```
priyajeet: `const sharing = sharedLink || externalCollab` for reuse

i18n/en-US.properties
jpfranco: Yes, the semicolon here is deliberate. I think it was suggested as a grammar improvement.

I'm now using `app` consistently in the code to keep the variable names short, but I will confirm with design if we want to update the copies. I think some of these were optimized for length, but I'll update them in the next PR if they prefer to keep it consistent.

i18n/en-US.properties
jpfranco: Same here, will check with product/design

src/features/classification/security-controls/SecurityControlsItem.scss
jpfranco: Yes this is meant to keep the gap size for the bullet relative to the font size

src/features/classification/security-controls/utils.js
jpfranco: Updated

PR #1743, opened 2019-11-20
feat(classification): Integrating Security Controls with Classification

jpfranco:
Just a quick update to bundle the `SecurityControls` with the `Classification` component since they will be used together most of the time.

Also renamed the main prop for `SecurityControls` to `controls` (instead of `accessPolicyRestrictions` in order to keep it more generic and allow future use cases.

Screen Shot 2019-11-19 at 1 12 48 PM


boxcla:

src/features/classification/security-controls/SecurityControls.js
```diff
@@ -14,24 +14,24 @@ import './SecurityControls.scss';
 const { FULL, SHORT, SHORT_WITH_TOOLTIP } = SECURITY_CONTROLS_FORMAT;
 
 type Props = {
-    accessPolicyRestrictions: AccessPolicyRestrictions,
+    controls: AccessPolicyRestrictions,
```
priyajeet: should type also be renamed?

jpfranco: Yes, updated

src/features/classification/Classification.js
```diff
@@ -44,6 +61,9 @@ const Classification = ({ definition, className = '', messageStyle, name, onClic
                     <FormattedMessage {...messages.missing} />
                 </span>
             )}
+            {controls && isSecurityControlsEnabled && (
```
jpfranco: Checking both vars mostly for flow's sake

PR #1747, opened 2019-11-21
fix(classification): Added missing application restriction messages

jpfranco:
`SecurityControls` component was using the same message for application restrictions regardless of whether the list of apps was allowed or disallowed. We will now use a separate message for each case.


boxcla:

PR #1749, opened 2019-11-21
fix(classification): Updated Security Controls messages

jpfranco:
Changing from `;` to `:` in app restriction messages.


boxcla:

PR #1745, opened 2019-11-20
fix(css): Update pill components in modals

sahiga:
This PR updates CSS rules for all BUIE features that depend on `PillSelectorDropdown`, fixing the regressions introduced in https://github.com/box/box-ui-elements/pull/1734.

bugfix-pill-newfoldermodal
bugfix-pill-sharedlinkmodal
bugfix-pill-taskmodal
bugfix-pill-usm

PR #1734, opened 2019-11-15
refactor(css): Add backwards-compatible prefixed class names for pills

sahiga:
This PR adds backwards-compatible prefixed class names for pill components (`PillCloud` and `PillSelectorDropdown`).

PR #1750, opened 2019-11-21
fix(css): Add legacy class names to pill components

sahiga:
This PR brings back legacy class names for pill components, to maintain backwards compatibility in apps that consume BUIE. Related to #1734 and #1745.


boxcla:

Issue #1693, opened 2019-10-25
Content Explorer not AODA compliant


dan12t3:
Please fill out the following template so we can reproduce and fix your issue as quickly as possible!

**Note:** if your issue includes a potential security vulnerability, please do not file it here. Instead, email the issue to security@box.com for support.

**Note:** if your issue involves previewing files, please report the issue [here](https://github.com/box/box-content-preview/issues/new/choose).

**Environment:**

- Elements version: Not sure, latest version?

**Desktop (please complete the following information):**

- OS: Windows 10
- Browser : Chrome
- Version : 77.0.3865.120

**Steps to reproduce the problem:**

issues on Content Explorer when testing for AODA
 
1.  <button> doesn't have any content, so it gets flagged as an "Empty" button, when it must have discernible text 

2. Missing label(aria-labels) on Search
  
3. id must be unique, Content Explorer is rendered on a page multiple times; which makes the id non-unique

**If relevant, link to file (or attach file here)**
 
![pic3](https://user-images.githubusercontent.com/11617645/67584744-43ca7f00-f71c-11e9-8384-4e8089c77557.PNG)
![pic4](https://user-images.githubusercontent.com/11617645/67584745-43ca7f00-f71c-11e9-9f45-aa2aebc0661d.PNG)
![pic5](https://user-images.githubusercontent.com/11617645/67584746-43ca7f00-f71c-11e9-9d93-320e1d4bd374.PNG)

PR #1752, opened 2019-11-22
fix(accessibility): add labels and remove un-needed ids

mickr:
This is meant to address #1693 which points out some issues with the `ContentExplorer` not being compatible with the [AODA](https://www.aoda.ca/)


boxcla:

src/elements/common/sub-header/ViewModeChangeButton.js
```diff
@@ -22,18 +22,15 @@ type Props = {
 
 const ViewModeChangeButton = ({ className = '', onViewModeChange, intl, viewMode, ...rest }: Props) => {
     const isGridView = viewMode === VIEW_MODE_GRID;
-
+    const viewMessage = isGridView ? intl.formatMessage(messages.listView) : intl.formatMessage(messages.gridView);
```
sahiga: Can we use `<FormattedMessage />` instead, to avoid the `injectIntl` HOC?

mickr: The HOC is already included and the Formatted Message component wont work with the ARIA labels.

PR #1760, opened 2019-11-24
feat(classification): Adding loading indicator for SecurityControls

jpfranco:
The information for the `SecurityControls` will normally be fetched separately from the item's own classification data and will be subsequently cached. We're adding a loading indicator in order to allow displaying classification while this additional data is being retrieved.

This PR also includes some style tweaks for the `SecurityControlsItem` tooltip styles.

Loading Indicator

### DIFF
--- a/src/features/classification/Classification.js
+++ b/src/features/classification/Classification.js
@@ -43,6 +43,7 @@ const Classification = ({
     const isTooltipMessageEnabled = isClassified && hasDefinition && messageStyle === STYLE_TOOLTIP;
     const isInlineMessageEnabled = isClassified && hasDefinition && messageStyle === STYLE_INLINE;
     const isNotClassifiedMessageVisible = !isClassified && messageStyle === STYLE_INLINE;
+    const isControlsIndicatorEnabled = isClassified && isLoadingControls && messageStyle === STYLE_INLINE;
     const isSecurityControlsEnabled =
         isClassified && !isLoadingControls && hasSecurityControls && messageStyle === STYLE_INLINE;
 
@@ -65,10 +66,10 @@ const Classification = ({
                     <FormattedMessage {...messages.missing} />
                 </span>
             )}
-            {controls && isSecurityControlsEnabled && (
-                <SecurityControls controls={controls} controlsFormat={controlsFormat} maxAppCount={maxAppCount} />
+            {isSecurityControlsEnabled && (
+                <SecurityControls controls={controls || {}} controlsFormat={controlsFormat} maxAppCount={maxAppCount} />
             )}
-            {isLoadingControls && <LoadingIndicator />}
+            {isControlsIndicatorEnabled && <LoadingIndicator />}
         </article>
     );
 };
--- a/src/features/classification/Classification.js
+++ b/src/features/classification/Classification.js
@@ -67,7 +67,7 @@ const Classification = ({
                 </span>
             )}
             {isSecurityControlsEnabled && (
-                <SecurityControls controls={controls || {}} controlsFormat={controlsFormat} maxAppCount={maxAppCount} />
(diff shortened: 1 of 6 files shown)


boxcla:

src/features/classification/Classification.js
```diff
@@ -61,9 +66,10 @@ const Classification = ({
                     <FormattedMessage {...messages.missing} />
                 </span>
             )}
-            {controls && isSecurityControlsEnabled && (
-                <SecurityControls controls={controls} controlsFormat={controlsFormat} maxAppCount={maxAppCount} />
+            {isSecurityControlsEnabled && (
```
Sunyang730: The loading indicator is just for the security controls message? If so, it would be good if we do a ? : or have the SecurityControls handle the loading state?

jpfranco: Yes, only for the security controls since this data will be loaded separately. The loading indicator doesn't necessarily show when `isSecurityControlsEnabled` so a `? :` wouldn't work here. Also think it makes sense for `Classification` to handle the loading state since it's only needed when `SecurityControls` is used in conjunction with `Classification`

src/features/classification/Classification.js
```diff
@@ -61,9 +66,10 @@ const Classification = ({
...
+                <SecurityControls controls={controls || {}} controlsFormat={controlsFormat} maxAppCount={maxAppCount} />
             )}
```
priyajeet: Should the defaulting to {} be inside security controls?

src/features/classification/Classification.js
Sunyang730: The way the code is structure confused me a little. The loading indicator looks like it's on its own but it should be for security controls only. Can we have isSecurityControlsEnabled and then a showSecurityControl/isLoadingSecurityControl flag after. Also, are we saying the SecurityControls won't have loading state in other usage? I will assume if we need to fetch the data, it should still have a loading state. no?

src/features/classification/__tests__/Classification.test.js
```diff
@@ -86,4 +87,47 @@ describe('features/classification/Classification', () => {
...
+        expect(wrapper.find(LoadingIndicator)).toHaveLength(1);
+        expect(wrapper.find(SecurityControls)).toHaveLength(0);
+    });
+
+    test('should not render loading indicator when item is not classified', () => {
+        const wrapper = getWrapper({
+            messageStyle: 'inline',
+            isLoadingControls: true,
+        });
+        expect(wrapper.find(LoadingIndicator)).toHaveLength(0);
```
Sunyang730: Can you also add to make sure securityControls is also having a length(0)?

src/features/classification/Classification.js
jpfranco: I think that's what I'm doing unless I'm misunderstanding your point. I have a `isControlsIndicatorEnabled` variable, just dropped the word `security` in order to keep the variable name short. In the other use cases we do not load the security controls data separately, so there is no need for a loading state. In general I think it the parent component should be the one handling the loading state.

Sunyang730: ok, I think my concern is if SecurityControls need to be used some where else and also need the loading state, then all the parent components will have to implement that too. From our conversation, sounds like the isLoadingIndicator is the main flag to display the loading indicator during the network request. 

Would be good if we could have it like the following
```
<Classification>
    <SecurityControls isLoadingIndicator />
</Classification>
```
And SecurityControls will have it's own loading state and it will be the component that makes the call to determine if it should be in loading state or not. That been said, this also highly tied to how the user will be using this component. It sounds like we won't have none Box use case at the moment. So spending more time on this discussion doesn't seems to be a good use of the time. To be continue in the future...

PR #1762, opened 2019-11-25
chore: upgrades react-intl dev dependency

mxiao6:
BREAKING CHANGE: Peer dependency upgrade


boxcla:

package.json
```diff
@@ -232,7 +232,7 @@
         "react-dom": "^16.9.0",
         "react-draggable": "^3.3.0",
         "react-immutable-proptypes": "^2.1.0",
-        "react-intl": "^2.8.0",
+        "react-intl": "2.9.0",
```
jstoffan: Seems like we should retain the `^` unless there's a reason to pin to this specific version.

package.json
```diff
@@ -286,7 +286,7 @@
...
```
jstoffan: Peer dependency upgrades should be marked as a breaking change via the commit message.

package.json
mxiao6: > Seems like we should retain the `^` unless there's a reason to pin to this specific version.

The EUA and admin-console is using this specific version. Should we be consistent with them?

jstoffan: I think we want to be a bit more permissive with supported peer dependency versions, since this is a shared library. If a consuming application wants to use 2.9.1, it seems like they should be able to do so without getting a yarn warning.

PR #1764, opened 2019-11-26
refactor(types): port some types to explicit imports

tjgupta:


boxcla:

.vscode/settings.json
```diff
@@ -5,7 +5,6 @@
     "flow.useNPMPackagedFlow": true,
     "flow.enabled": true,
     "javascript.validate.enable": false,
-    "prettier.eslintIntegration": true,
```
tjgupta: My editor was complaining about this setting. I can remove if we prefer, but vscode doesn't like this anymore.

src/common/types/api.js
```diff
@@ -0,0 +1,45 @@
...
+type ElementsError = {
+    code: string,
+    context_info: Object,
+    message: string,
+    origin: ElementOrigin,
+    type: 'error',
+};
+
+type ErrorContextProps = {
+    onError: (error: ElementsXhrError | Error, code: string, contextInfo?: Object, origin?: ElementOrigin) => void,
```
tjgupta: I had to make `origin` an optional type because we're not always passing this in.

src/elements/common/flowTypes.js
```diff
@@ -10,5 +21,23 @@ type ErrorType = {
...
+
+type ElementOrigin =
+    | typeof ORIGIN_CONTENT_SIDEBAR
+    | typeof ORIGIN_CONTENT_PREVIEW
+    | typeof ORIGIN_PREVIEW
+    | typeof ORIGIN_DETAILS_SIDEBAR
+    | typeof ORIGIN_ACTIVITY_SIDEBAR
+    | typeof ORIGIN_SKILLS_SIDEBAR
+    | typeof ORIGIN_METADATA_SIDEBAR
+    | typeof ORIGIN_OPEN_WITH;
```
tjgupta: I had to add this as it was expected in the `ContentOpenWith` component.

.vscode/settings.json
priyajeet: yeah it may have been removed?

src/common/types/api.js
```diff
@@ -0,0 +1,45 @@
...
+};
+
+type ElementsErrorCallback = (e: ElementsXhrError, code: string, contextInfo?: Object) => void;
+
+export type {
```
booc0mtaco: Curious if its better to `export` on each line? if possible, that's a small amount less code to maintain.

tjgupta: Yeah, I have a slight preference to that for type files personally, but I think we collectively landed on this style instead. I suppose an argument could be made to do these type files the same as other JS files...?

alexkrolick: If there are a lot of internal types that aren't exported, having them on the bottom seems a bit easier for identifying what is actually exported. But when everything is exported, seems like the opposite. 🤔

PR #1766, opened 2019-11-26
refactor(types): port some types to explicit imports

tjgupta:
Some cleanup from a few imports I missed related to the API types earlier. Adding these explicit imports causes a lot of flow errors that I'm not quite sure how to resolve though. May need some help on this.


boxcla:

PR #1771, opened 2019-11-26
refactor(types): port integration types etc to explicit imports

tjgupta:


boxcla:

src/common/types/integrations.js
```diff
@@ -0,0 +1,55 @@
+// @flow
+import { FormattedMessage } from 'react-intl';
+import { APP_INTEGRATION } from '../../constants';
+
+type AppIntegrationAPIMiniItem = {
+    id: string,
+    type: typeof APP_INTEGRATION,
```
tjgupta: This `typeof` is the only real change in this PR (besides explicit imports). Same thing on line 52, plus a corresponding change int the constants file.

src/common/types/integrations.js
```diff
@@ -0,0 +1,55 @@
...
+};
+
+type ExecuteAPIParam = {
+    key: string,
+    value: string,
+};
+
+type ExecuteAPI = {
+    integration_type: string,
+    method: HTTP_POST | HTTP_GET,
```
alexkrolick: these might need a `typeof`?

tjgupta: Good catch, done @alexkrolick

src/common/types/integrations.js
```diff
@@ -0,0 +1,55 @@
...
+};
+
+type ExecuteAPIParam = {
+    key: string,
+    value: string,
+};
+
+type ExecuteAPI = {
+    integration_type: string,
+    method: typeof HTTP_POST | typeof HTTP_GET,
```
tjgupta: Actually, added `typeof` here too after @alexkrolick mentioned it.

PR #1775, opened 2019-11-27
refactor(types): port all remaining types to explicit imports

tjgupta:
BREAKING CHANGE: This ports all remaining implicit global types to explicit imports. This may cause flow type check errors where there were none previously if code was using these implicit global types. Note, also that these global types were previously resolving to `any` types, but now should resolve accurately.

To fix, import the type before using:

### Before
```
type Props = {
  item: BoxItem
}
```

### After
```
import type { BoxItem } from 'box-ui-elements/es/common/types/core';

type Props = {
  item: BoxItem
}
```


boxcla:

src/common/types/api.js
```diff
@@ -34,12 +57,34 @@ type ErrorContextProps = {
 
 type ElementsErrorCallback = (e: ElementsXhrError, code: string, contextInfo?: Object) => void;
 
+type XHROptions = {
```
jstoffan: Can we name these `APIOptions`? Some of them are used to create XHRs, but not all.

src/common/types/api.js
```diff
@@ -34,12 +57,34 @@ type ErrorContextProps = {
...
+    sharedLink?: string,
+    sharedLinkPassword?: string,
+    shouldRetry?: boolean,
+    token: Token,
+    uploadHost?: string,
+    version?: string,
+};
+
 export type {
+    Method,
```
jstoffan: Can/should we alphabetize declarations and exports?

src/common/types/core.js
```diff
@@ -1,5 +1,77 @@
...
+    FIELD_DATE,
+    FIELD_NAME,
+    FIELD_SIZE,
+    FIELD_RELEVANCE,
+    DEFAULT_VIEW_RECENTS,
+    DEFAULT_VIEW_FILES,
+    VERSION_RETENTION_DELETE_ACTION,
+    VERSION_RETENTION_REMOVE_ACTION,
+    VERSION_RETENTION_INDEFINITE,
+} from '../../constants';
```
jstoffan: Should type files import source files? Or should the types for these constants live in their own type file and be imported into the source files?

src/common/types/core.js
```diff
@@ -26,6 +98,26 @@ type UserMini = {
     type: 'user',
 };
 
+type User = {
```
jstoffan: Do we want to dump all of these in `core` or should they each get their own domain-specific file?

src/common/types/metadata.js
```diff
@@ -1,8 +1,123 @@
-// @flow strict
```
jstoffan: Any way we can retain the strict checks?

src/utils/sorter.js
```diff
@@ -6,6 +6,15 @@
 
 import comparator from './comparator';
 import { getBadItemError } from './error';
+import type { Comments, Tasks, AppActivityItems, FeedItems } from '../common/types/feed';
```
jstoffan: Should we create `FeedItem/FeedItems` types to abstract this away from the consumers?

src/common/types/api.js
tjgupta: Yep

src/common/types/api.js
tjgupta: The declarations have to be in order of usage to appease Eslint. In fact, I had to disable the rule on one line in types/core.js due to one type using the other and vice versa... no way to do that.

Do you prefer them in alpha with that rule disabled?

jstoffan: That makes sense. I guess I'm thinking about named functions, where the declaration order doesn't matter. Feel free to leave them as-is.

src/common/types/core.js
tjgupta: Most of these constants are used elsewhere besides these types... I'm inclined to have it as is personally.

I was kind of thinking to try to combine `src/common/constants.js` with `src/constants.js` at some point in a followup though. Seems redundant to me.

src/common/types/core.js
tjgupta: I think they could probably get split up. I tried to do some splitting, but it's somewhat hard to see the divisions until it's done in some cases.

I was thinking perhaps a file related to the various `BoxItem` related stuff would be good. Any other divisions?

Good with doing this in a followup though?

src/common/types/metadata.js
tjgupta: I can look and see. I didn't know the code well enough to understand the proper type, but let me try again.

src/utils/sorter.js
tjgupta: Sure, though I'd prefer to avoid changing the types until a followup.

src/common/types/core.js
alexkrolick: most of the constants are defined like `const X: 'x': 'x'` so they can be used with `typeof X` in types and `X` in source files and get strict constant values

jstoffan: I like the inline type declarations for constants. The main concern I have with this change is the possible circular dependencies it could introduce. @tjgupta mentioned he already ran into such an issue during his work on this PR. If that's not something worth worrying about further, though, then feel free to disregard.

src/common/types/core.js
jstoffan: It looks like item, metadata, and user might deserve their own type files, but I think doing it in a follow-up is fine, yeah.

src/common/types/api.js
tjgupta: Done

src/common/types/metadata.js
tjgupta: Added this back, with a $FlowFixMe on the one line that was problematic.

src/common/types/core.js
tjgupta: Yep it caused only one issue, and that was here:
https://github.com/box/box-ui-elements/pull/1775/files#diff-a7748843fbda5167d065231e3a2e7a1fL375

Arguably though, that isn't really a constant in my opinion. Probably doesn't really belong in that file.

src/elements/common/sub-header/Sort.js
```diff
@@ -21,7 +22,9 @@ type Props = {
     sortDirection: SortDirection,
 };
 
-const SORT_ITEMS: Array<Array<SortBy>> = [
+type SortItem = [SortBy, SortDirection];
```
tjgupta: This was incorrect before, but was working because of the `any` type that it was getting set to.

src/features/metadata-instance-editor/constants.js
```diff
@@ -7,7 +7,3 @@ export const FIELD_TYPE_FLOAT: 'float' = 'float';
 export const FIELD_TYPE_MULTISELECT: 'multiSelect' = 'multiSelect';
 export const FIELD_TYPE_STRING: 'string' = 'string';
 export const TEMPLATE_CUSTOM_PROPERTIES: 'properties' = 'properties';
-export const JSON_PATCH_OP_ADD: 'add' = 'add';
```
tjgupta: Consolidated these with the main constants file.

src/common/types/core.js
```diff
@@ -43,14 +135,252 @@ type SelectorItem = {
...
+    shared_link?: SharedLink,
+    size?: number,
+    thumbnailUrl?: ?string,
+    type?: ItemType,
+    url?: string,
+    version_limit?: ?number,
+    version_number?: string,
+};
+
+type BoxItemCollection = {
```
booc0mtaco: We can probably move this above `BoxItem` and remove the comment on line 244

src/common/types/core.js
```diff
@@ -1,5 +1,77 @@
...
+} from '../../constants';
+import type { MetadataType } from './metadata';
+
+type Token = null | typeof undefined | string | Function;
+type TokenReadWrite = { read: string, write?: string };
+type TokenLiteral = null | typeof undefined | string | TokenReadWrite;
+
+type ClassComponent<P, S> = Class<React$Component<P, S>>;
+
+type StringMap = { [string]: string };
```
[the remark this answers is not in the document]
alexkrolick: I think using these types may actually loosen the guarantees about what you can do with the result later in terms of $Keys and $Values, typeof, etc - would have to check in a sandbox though

src/common/types/core.js
tjgupta: @booc0mtaco they each depend on each other, so neither order works entirely.

PR #1777, opened 2019-12-02
refactor(types): remove global type imports

tjgupta:


boxcla:

flow-typed/box-ui-elements.js
```diff
@@ -6,17 +6,13 @@
 /* eslint-disable no-unused-vars */
 // NOTE: all of these imports resolve to `any`
 // see https://github.com/facebook/flow/issues/7574
-import * as React from 'react';
-import { FormattedMessage } from 'react-intl';
-import type { MessageDescriptor, InjectIntlProvidedProps } from 'react-intl';
+import type { InjectIntlProvidedProps } from 'react-intl';
```
tjgupta: The rest of these left here are going to require more work and maybe more familiarity with the code, because removing them results in heavier flow errors.

Also note, this diff makes it look like I'm adding this import here, but it was already there. Just removed the other `MessageDescriptor` import along with it (plus it's on a new line), so Github is making it look like an addition.

PR #1778, opened 2019-12-03
fix(classification): Fixing typo on security controls message

jpfranco:
Just removing an extra "and" from one of the messages


boxcla:

PR #1780, opened 2019-12-03
fix(css): Replace tilde imports

kimd-box:
- that breaks module resolution in parceljs based apps


boxcla:
Verified that @kimd-box has signed the CLA. Thanks for the pull request!

mickr:
Will these changes have any effect on load order and precedence?

kimd-box:
@mickr Based on my research, webpack loads the css and scss imported in JS in the same order as presented. 

I verified using in ContentUploader in EUA that the load order/precedence remains the same.  (also tested with ContentExplorer in Relay). 

Screen Shot 2019-12-07 at 9 11 32 AM

kimd-box:
Sorry @mickr. I had to rebase post your approval.. one more thumbs up would be much appreciated. Sorry fo the trouble.

PR #1782, opened 2019-12-03
feat(context-menu): Pass class name to target div

macrotim:


boxcla:

src/components/context-menu/__tests__/ContextMenu.test.js
```diff
@@ -30,6 +30,18 @@ describe('components/context-menu/ContextMenu', () => {
     });
 
     describe('render()', () => {
+        test('should pass className to the target div', () => {
+            const className = 'context-menu';
+            const wrapper = shallow(
+                <ContextMenu className={className}>
+                    <FakeButton />
```
swfree: Just curious - since the classname is being applied to the FakeButton target component, would the user not add the classname directly as `<FakeButton className={className} />`? Maybe there's a use case I'm not thinking of?

macrotim: You're right. Thanks for the pointer!

I'll go ahead and close this out.

PR #1785, opened 2019-12-05
fix(uploads): fix auto retry after timeout

danubevictoria:
Adds pause state and resets uploading process for all parts on error. Updates state and retries upload on unpausing the parts. This change stops all auto-resuming after we display an error message to the user.

### DIFF
--- a/src/api/uploads/MultiputPart.js
+++ b/src/api/uploads/MultiputPart.js
@@ -250,7 +250,7 @@ class MultiputPart extends BaseMultiput {
         const xhr_ready_state = getProp(this.xhr, 'xhr.readyState', null);
         const xhr_status_text = getProp(this.xhr, 'xhr.statusText', '');
 
-        console.log(`Upload failure ${error.message} for part ${this.toJSON()}. XHR state: ${xhr_ready_state}.`);
+        this.consoleLog(`Upload failure ${error.message} for part ${this.toJSON()}. XHR state: ${xhr_ready_state}.`);
 
         const eventInfo = {
             message: error.message,
@@ -293,7 +293,6 @@ class MultiputPart extends BaseMultiput {
 
         this.numUploadRetriesPerformed += 1;
         this.consoleLog(`Retrying uploading part ${this.toJSON()} in ${retryDelayMs} ms`);
-        console.log(`Retrying uploading part ${this.toJSON()} in ${retryDelayMs} ms`);
         this.retryTimeout = setTimeout(this.retryUpload, retryDelayMs);
     };
 
@@ -311,7 +310,7 @@ class MultiputPart extends BaseMultiput {
             const parts = await this.listParts(this.index, 1);
 
             if (parts && parts.length === 1 && parts[0].offset === this.offset && parts[0].part_id) {
-                console.log(`Part ${this.toJSON()} is available on server. Not re-uploading.`);
+                this.consoleLog(`Part ${this.toJSON()} is available on server. Not re-uploading.`);
                 this.id = parts[0].part_id;
                 this.uploadSuccessHandler({
                     data: {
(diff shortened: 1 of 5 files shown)


boxcla:
Verified that @danubevictoria has signed the CLA. Thanks for the pull request!

src/api/uploads/MultiputUpload.js
```diff
@@ -655,6 +639,23 @@ class MultiputUpload extends BaseMultiput {
      */
     partUploadErrorHandler = (error: Error, eventInfo: string): void => {
         this.sessionErrorHandler(error, LOG_EVENT_TYPE_PART_UPLOAD_RETRIES_EXCEEDED, eventInfo);
```
cweeii: It seems like if we pause, we shouldn't be calling this, right?

src/api/uploads/MultiputPart.js
```diff
@@ -339,6 +345,48 @@ class MultiputPart extends BaseMultiput {
...
 
+    /**
+     * Pauses upload for this Part.
+     *
+     * @return {void}
+     */
+    pause(): void {
+        this.isPaused = true;
+        this.state = PART_STATE_DIGEST_READY;
+        this.xhr.abort(); //  This calls the error handler.
```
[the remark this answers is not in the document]
danubevictoria: @cweeii When we go to upload a file, we use `xhr.uploadFile`. When user disconnects from the internet mid-upload, we still attempt to upload the parts for 5 min. Once we reach the timeout, in `xhr.uploadFile`, there is an `idleTimeoutFunc` that calls the first `abort` for one of the parts that was still uploading. `xhr.abort` results in cancelling the request (`axios.cancel`) for the part which results in an error. The error then gets handled by `partUploadErrorHandler` because when we first populate the parts in `MultiputUpload.populateParts` we set that to be the error handler. 

We want to abort on pause because if one part is aborted, we shouldn't let the rest of the parts just hang around or be cancelled because cancelling destroys the parts and we need to reuse the parts and their xhrs to resume. We have to reuse the parts because the API validates the parts using a unique sha.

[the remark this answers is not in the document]
danubevictoria: And yes u are right that it is an intended action when the user pauses their upload, which is why `isDestroyedOrPaused()` was created so that we ignore the errors that were caused from aborting through pause.

src/api/uploads/MultiputUpload.js
danubevictoria: Not sure about this one, but I can play around with it. I think we still need to call `sessionErrorHandler` because it calls the errorCallback which is set to be `ContentUploader.handleUploadError` from `resume` in `ContentUploader.resumeFile`. The error callback updates the Uploads Manager in EUA to show uploads error.

danubevictoria: Ok tried it and we definitely need it otherwise it gets into an infinite loading state.

src/api/uploads/__tests__/MultiputPart.test.js
```diff
@@ -27,8 +27,29 @@ describe('api/uploads/MultiputPart', () => {
...
+            MultiputPartTest.destroyed = true;
+            MultiputPartTest.isPaused = false;
+            MultiputPartTest.xhr.uploadFile = jest.fn();
+
+            MultiputPartTest.upload();
+
+            expect(MultiputPartTest.xhr.uploadFile).not.toHaveBeenCalled();
+        });
+
+        test('should noop if paused', () => {
```
sahiga: I recommend using `test.each` for these tests and others that have a similar structure

src/api/uploads/__tests__/MultiputUpload.test.js
```diff
@@ -120,6 +120,46 @@ describe('api/uploads/MultiputUpload', () => {
             expect(multiputUploadTest.numPartsUploading).toBe(1);
             expect(multiputUploadTest.parts[0].upload).toHaveBeenCalled();
         });
+
+        test('should call unpause when part is paused and in digest ready state', () => {
```
sahiga: You could also use `test.each` for these tests

danubevictoria: I was not able to pass in the different function calls to `test.each` so left it as is.

PR #1786, opened 2019-12-05
feat(content-uploader): batch onComplete and add onResume prop

borajimin:
Partially reverting the revert. The regression was a false positive.


boxcla:
Verified that @borajimin has signed the CLA. Thanks for the pull request!

PR #1788, opened 2019-12-06
feat(base-select): Add ability to use first letter for select menu traversal

mickr:


boxcla:

src/components/select-field/BaseSelectField.js
```diff
@@ -205,7 +206,17 @@ class BaseSelectField extends React.Component<Props, State> {
                     this.closeDropdown();
                 }
                 break;
-            // no default
+            default: {
+                stopDefaultEvent(event);
+                const lowerCaseKey = key.toLowerCase();
+                const optionIndex = options.findIndex(
```
sahiga: IE doesn't support `array.findIndex`. Can we use the lodash `_.findIndex` method instead?

PR #1792, opened 2019-12-10
feat(assets): Import latest assets from design

booc0mtaco:
Update imports from the design team to latest version. This also reorganizes the content to reflect asset hierarchy.

BREAKING CHANGE: Import paths for previous file/folder icons has changed, and will need to be
updated to support the new structure. Users of the ItemIcon and FileIcon components will not need to make any changes.


boxcla:

src/icons/imported/Icon/Content/FileBookmark32.js
```diff
@@ -1,10 +1,10 @@
 // @flow
```
priyajeet: Can you make folder path lowercase.
Also, can "imported" be dropped now from the path, since we have new top level folders now

src/icon/fill/FileCode16.js
```diff
@@ -0,0 +1,16 @@
+// @flow
+/* eslint-disable react/jsx-sort-props */
+import * as React from 'react';
+import AccessibleSVG from '../../icons/accessible-svg';
+import type { Icon } from '../../icons/flowTypes';
+
+const FileCode16 = (props: Icon) => (
+    <AccessibleSVG width={16} height={16} viewBox="0 0 16 16" {...props}>
```
tjgupta: Are we intentionally removing the classNames? Like in this case it would have been `icon-file-code` previously?

src/icons/file/README.md
```diff
@@ -21,107 +21,107 @@ const IconsExample = require('../../../examples/src/IconsExample').default;
 const icons = [
     {
         name: 'IconFileAudio',
-        component: require('../imported/IconFileAudio32').default,
+        component: require('../../icon/content/FileAudio32').default,
```
tjgupta: Is this accurate, that using `<IconFileAudio>` would render `<FileAudio32>`?

booc0mtaco: oh good catch, the names should be updated

src/icon/fill/FileCode16.js
booc0mtaco: I have a follow up step to add in class names per component, among some other changes (to handle color variables, and better documentation generation).

PR #1794, opened 2019-12-10
feat(icons): add Relay flag icon

frogzsj:
relayflag


boxcla:

PR #1797, opened 2019-12-11
feat: typescript infrastructure and few components conversion

priyajeet:
This PR introduces typescript into BUIE. It includes:

- Infra for TS, configs etc
- PrimaryButton, Button, RadarAnimation, LoadingIndicator are migrated to TSX

Parent projects doing local symlink-ing with BUIE will need to update their configs to allow TS.


boxcla:

.eslintrc.js
```diff
@@ -18,16 +18,44 @@ module.exports = {
         'react/no-access-state-in-setstate': 'off', // fixme
         'react/no-array-index-key': 'off', // fixme
         'react/no-this-in-sfc': 'off',
+        '@typescript-eslint/explicit-function-return-type': 'off', // fixme
```
priyajeet: remove

.vscode/settings.json
```diff
@@ -1,11 +1,13 @@
 // Place your settings in this file to overwrite default and user settings.
 {
-    "typescript.validate.enable": false,
-    "typescript.format.enable": false,
     "flow.useNPMPackagedFlow": true,
     "flow.enabled": true,
+    "eslint.autoFixOnSave": true,
     "javascript.validate.enable": false,
```
priyajeet: Enabling this to true breaks hell loose, with TSC trying to check .js files ignoring tsconfig.json

examples/src/ContentExplorerModalContainerExamples.js
```diff
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
 
-import Button from '../../src/components/button';
+import Button from '../../src/components/button'; // eslint-disable-line
```
priyajeet: A .js file can no longer import .ts file.
The compiled versions will be fine, but uncompiled this will be an issue, as such you will see this (and flow suppression of the same) in many files.

i18n/en-US.properties
```diff
@@ -1010,10 +1010,6 @@ boxui.searchForm.clearButtonTitle = Clear
 boxui.searchForm.searchButtonTitle = Search
 # Label for a search input
 boxui.searchForm.searchLabel = Search query
-# Instructional message displayed on the embed widget security drag-drop game
-boxui.securityCloudGame.instructions = For security purposes, please drag the white cloud into the dark cloud.
```
priyajeet: revert

package.json
```diff
@@ -8,6 +8,7 @@
     "main": "es/index.js",
     "module": "es/index.js",
     "jsnext:main": "es/index.js",
+    "types": "src/components/index.d.ts",
```
priyajeet: should point to es

package.json
```diff
@@ -42,28 +43,30 @@
...
         "build:prod:npm": "BABEL_ENV=production OUTPUT=dist LANGUAGE=en-US REACT=true yarn build:prod:dist",
         "build:prod:storybook": "LANGUAGE=en-US REACT=true BROWSERSLIST_ENV=production BABEL_ENV=development NODE_ENV=development build-storybook -c .storybook -o styleguide/storybook",
         "build:sync": "LANGUAGE=en-US BABEL_ENV=development NODE_ENV=development RSYNC=true webpack --config scripts/webpack.config.js --mode development",
         "build:variables": "cat `find src/styles/constants -name '*.scss'` > src/styles/variables.scss | node ./scripts/build-style-vars.js src/styles/variables.scss && rm src/styles/variables.scss",
         "clean": "rm -rf dist es i18n/json i18n/*.js reports styleguide",
+        "copy:flow": "flow-copy-source --ignore \"**/__tests__/**\" --ignore \"**/__mocks__/**\" --ignore \"**/*.ts\" ./src ./es",
         "copy:styles": "copyfiles './src/**/*.scss' es --up 1",
-        "copy:flow": "flow-copy-source --ignore \"**/__tests__/**\" --ignore \"**/__mocks__/**\" ./src ./es",
+        "copy:types": "tsc --declaration --emitDeclarationOnly --declarationDir es --noEmit false",
```
priyajeet: rename to ts:definitions or something since its not a copy

scripts/styleguide.config.js
```diff
@@ -1,4 +1,6 @@
 const path = require('path');
+// const typescriptDocGen = require('react-docgen-typescript');
```
priyajeet: revert, but without this styleguide cannot detect props iirc. But thats fine, storybook will be taking care of this now on.

src/components/button/Button.tsx
```diff
@@ -1,43 +1,45 @@
-// @flow
 import * as React from 'react';
 import classNames from 'classnames';
-import omit from 'lodash/omit';
-
 import LoadingIndicator from '../loading-indicator';
 import RadarAnimation from '../radar';
 
-type Props = {
+export enum ButtonType {
```
priyajeet: 😇

src/components/button/Button.tsx
```diff
@@ -1,43 +1,45 @@
...
+    BUTTON = 'button',
+    RESET = 'reset',
+    SUBMIT = 'submit',
+}
+export interface ButtonProps {
```
priyajeet: Based on the interwebs, not much realistic difference between using `type` or `interface` for the purpose of react props. But since the latter is more extensible, might as well use it

src/components/button/Button.tsx
```diff
@@ -1,43 +1,45 @@
...
-    btnElement: ?HTMLButtonElement;
+class Button extends React.Component<ButtonProps> {
+    btnElement: HTMLButtonElement | null = null;
 
     static defaultProps = {
         className: '',
         isLoading: false,
         showRadar: false,
-        type: 'submit',
+        type: ButtonType.SUBMIT,
```
priyajeet: 😇

src/components/button/__tests__/Button.test.tsx
```diff
@@ -45,11 +42,7 @@ describe('components/button/Button', () => {
         const preventDefault = sinon.spy();
         const stopPropagation = sinon.spy();
 
-        const wrapper = shallow(<Button isDisabled onClick={onClickHandler} />);
-
-        const contains = sinon.stub();
-        contains.withArgs('is-disabled').returns(true);
-        wrapper.instance().btnElement = { classList: { contains } };
+        const wrapper = shallow<Button>(<Button isDisabled onClick={onClickHandler} />);
```
priyajeet: shallow<TYPE> indicates that the wrapper.instance() will return that TYPE and you can call functions or methods directly on that instance. Think of it syntactical sugar over casting.

src/components/button/__tests__/__snapshots__/Button.test.tsx.snap
```diff
@@ -4,6 +4,7 @@ exports[`components/button/Button should not render a RadarAnimation if showRada
 
 exports[`components/button/Button should render a RadarAnimation if showRadar is true 1`] = `
 <RadarAnimation
+  className=""
```
priyajeet: Such changes in this file are due to the use of defaultProps over defaults in args

src/components/index.ts
```diff
@@ -0,0 +1,5 @@
+export { default as RadarAnimation } from './radar';
+export * from './radar';
```
priyajeet: You not only export the component default as named, but also all types exports. This will be useful when generating .d.ts files at build time, the root index.d.ts will be pointed to within package.json

src/components/loading-indicator/LoadingIndicatorWrapper.tsx
```diff
@@ -0,0 +1,56 @@
+import * as React from 'react';
```
priyajeet: I am guessing the diff/rename was too big so git added this as a new file. Compare to older .js version

src/components/loading-indicator/__tests__/LoadingIndicatorWrapper.test.tsx
```diff
@@ -48,14 +55,20 @@ describe('components/loading-indicator/LoadingIndicatorWrapper', () => {
 
     test('should pass thru any rest properties', () => {
         const wrapper = shallow(
-            <LoadingIndicatorWrapper className="test-name" data-resin-target="test" isLoading="true" />,
+            <LoadingIndicatorWrapper className="test-name" data-resin-target="test" isLoading>
```
priyajeet: This line throws a tsc error that has been a bit hard to fix
TODO

src/components/radar/RadarAnimation.tsx
```diff
@@ -1,88 +1,89 @@
...
-const BOTTOM_CENTER = 'bottom-center';
-const BOTTOM_LEFT = 'bottom-left';
-const BOTTOM_RIGHT = 'bottom-right';
-const MIDDLE_CENTER = 'middle-center';
-const MIDDLE_LEFT = 'middle-left';
-const MIDDLE_RIGHT = 'middle-right';
-const TOP_CENTER = 'top-center';
-const TOP_LEFT = 'top-left';
-const TOP_RIGHT = 'top-right';
+export enum RadarAnimationPosition {
```
priyajeet: A change like this will not break a parent project, since the compiled down es5 version is still string based. However symlinking and using "src" directly will likely no longer work and/or parent projects typescript check will complain even though at runtime it is fine.

src/components/radar/__tests__/RadarAnimation.test.tsx
```diff
@@ -13,81 +13,81 @@ describe('components/radar/RadarAnimation', () => {
...
         });
     });
 
     describe('position instance method', () => {
-        test.each([true, false])(`should only position the tether when shown`, isShown => {
+        test.each([true, false])('should only position the tether when shown', isShown => {
             const positionTetherMock = jest.fn();
 
-            const wrapper = getWrapper({ isShown });
+            const wrapper = getWrapper({ isShown } as RadarAnimationProps);
```
priyajeet: eslint prettier/prettier rule has been randomly complaining about use of "as" at random (but not all) places. This is one of the things that needs to be figured out why. Could just be a bug in the rule and we may need to disable it.

tsconfig.json
```diff
@@ -0,0 +1,22 @@
+{
+    "compilerOptions": {
```
priyajeet: This will evolve over time

examples/src/ContentExplorerModalContainerExamples.js
davidjohnburrowes: would like more info where that does not work...

src/components/button/Button.tsx
davidjohnburrowes: neither here nor there. I prefer to use interface only when planning to work with classes as a signal that I'm planning on this participating in OO

src/components/button/__tests__/Button.test.tsx
```diff
@@ -30,11 +31,7 @@ describe('components/button/Button', () => {
     test('simulates click events', () => {
         const onClickHandler = sinon.spy();
 
-        const wrapper = shallow(<Button onClick={onClickHandler} />);
-
-        const contains = sinon.stub();
-        contains.withArgs('is-disabled').returns(false);
-        wrapper.instance().btnElement = { classList: { contains } };
+        const wrapper = shallow<Button>(<Button onClick={onClickHandler} />);
```
tjgupta: Is there a lot of value in adding types in our test files? Or is it purely opt-in? I feel like this will slow productivity in these types of files personally.

src/components/button/__tests__/Button.test.tsx
alexkrolick: A lot of our tests are not currently type-safe due to missing required props, stubbed objects, etc. Will take some effort to migrate them.

src/components/button/__tests__/__snapshots__/Button.test.tsx.snap
alexkrolick: Worth pointing out `defaultProps` is on the deprecation path, so it might be worth it to go the other way if there are no issues with it.

https://twitter.com/dan_abramov/status/1133878326358171650

src/components/index.ts
```diff
@@ -1,3 +1,9 @@
+export { default as Button } from './button';
+export * from './button';
+
+export { default as PrimaryButton } from './button';
```
priyajeet: from primary-button

src/components/button/Button.tsx
alexkrolick: surprising there isn't a TS global for this

src/components/button/Button.tsx
alexkrolick: It looks like `interface` is used for structural sub-typing which is what you want with props usually

https://www.typescriptlang.org/docs/handbook/interfaces.html

src/components/primary-button/__tests__/PrimaryButton.test.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
-
+import { shallow } from 'enzyme';
```
priyajeet: less overall globals keeps typescript happy

src/components/button/__tests__/Button.test.tsx
alexkrolick: I think there is value in it in terms of autocomplete, not creating invalid test definitions, etc. but many of our tests will fail because of incomplete mock data. You can see this already if you add `// @flow` to a file.

src/components/button/__tests__/Button.test.tsx
priyajeet: indeed

src/components/button/__tests__/__snapshots__/Button.test.tsx.snap
priyajeet: interesting, I guess saves some time then

.eslintrc.js
```diff
@@ -18,16 +18,44 @@ module.exports = {
         'react/no-access-state-in-setstate': 'off', // fixme
         'react/no-array-index-key': 'off', // fixme
         'react/no-this-in-sfc': 'off',
+        'import/no-unresolved': 'off', // fixme
```
priyajeet: This is to keep eslint happy from when a JS file tries to import a TS module

.eslintrc.js
```diff
@@ -18,16 +18,44 @@ module.exports = {
...
+                "typescript": {
+                    "directory": "."
+                }
+                }
+            },
+            overrides: [
+                {
+                    files: ['*.stories.tsx', '*.test.tsx'],
+                    rules: {
+                        'import/no-extraneous-dependencies': 'off'
```
priyajeet: we dont want to consider stories as something needed at runtime. So imports dont need to be in deps section and can remain in dev deps.

src/components/loading-indicator/__tests__/__snapshots__/LoadingIndicatorWrapper.test.tsx.snap
```diff
@@ -21,6 +21,7 @@ exports[`components/loading-indicator/LoadingIndicatorWrapper should hide conten
 <div
   className="loading-indicator-wrapper test-name"
 >
+  <div />
```
priyajeet: Check why these extra divs are there

examples/src/ContentExplorerModalContainerExamples.js
priyajeet: Had to disable this eslint rule. Alternative is we add it to each line of import that causes the problem.

.eslintrc.js
```diff
@@ -18,16 +18,45 @@ module.exports = {
...
                 shallow: true,
                 mount: true,
                 takeScreenshot: true,
             },
+        },
+        {
+            files: ['*.ts', '*.tsx'],
+            rules: {
+                '@typescript-eslint/explicit-function-return-type': 'off', // fixme
+                'import/extensions': 'off',
```
priyajeet: This rule forces extension on imports ==> `import Button from '../button.js'`

src/components/loading-indicator/__tests__/__snapshots__/LoadingIndicatorWrapper.test.tsx.snap
priyajeet: This is there because the wrapper requires a child. Before it was incorrect.

src/components/loading-indicator/__tests__/makeLoadable.test.tsx
```diff
@@ -35,7 +40,8 @@ describe('components/loading-indicator/makeLoadable', () => {
 
     test('should pass loadingIndicatorPorps to LoadingIndicator', () => {
         const LoadableComponent = makeLoadable(TestComponent);
-        const wrapper = shallow(<LoadableComponent isLoading loadingIndicatorProps={{ className: 'foobar' }} />);
+        // eslint-disable-next-line prettier/prettier
```
priyajeet: Not yet sure how to fix this lint error

src/components/loading-indicator/__tests__/makeLoadable.test.tsx
```diff
@@ -19,7 +20,11 @@ describe('components/loading-indicator/makeLoadable', () => {
 
     test('should pass props down to BaseComponent', () => {
         const LoadableComponent = makeLoadable(TestComponent);
-        const wrapper = shallow(<LoadableComponent className="foo" hello="123" isLoading={false} />);
+        const props = {
+            className: 'foo',
+            hello: '123',
+        };
+        const wrapper = shallow(<LoadableComponent isLoading={false} {...props} />);
```
priyajeet: putting className and hello directly as props causes TS error. Something with the makeLoadable HOC.

src/components/loading-indicator/makeLoadable.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react';
+import LoadingIndicator, { LoadingIndicatorProps } from './LoadingIndicator';
+
+export interface MakeLoadableProps {
+    isLoading?: boolean,
+    loadingIndicatorProps?: LoadingIndicatorProps,
+};
+
+// eslint-disable-next-line prettier/prettier
```
priyajeet: Not yet sure how to fix this lint error

src/components/radar/RadarAnimation.tsx
```diff
@@ -0,0 +1,139 @@
...
+        if (constrainToWindow) {
+            constraints.push({
+                to: 'window',
+                attachment: 'together',
+            });
+        }
+
+        const { attachment, targetAttachment } = positions[position];
+        const child = React.Children.only(children);
+        // eslint-disable-next-line prettier/prettier
```
priyajeet: Not yet sure how to fix this lint error

src/components/radar/__tests__/RadarAnimation.test.tsx
```diff
@@ -13,81 +13,84 @@ describe('components/radar/RadarAnimation', () => {
...
         },
         {
             // description: 'should render correctly with top-right positioning',
-            position: 'top-right',
+            position: RadarAnimationPosition.TOP_RIGHT,
         },
     ].forEach(({ position }) => {
         test(`should render correctly with ${position} positioning`, () => {
-            const wrapper = getWrapper({ position });
+            // eslint-disable-next-line prettier/prettier
```
priyajeet: Not yet sure how to fix this lint error

src/components/radar/__tests__/RadarAnimation.test.tsx
```diff
@@ -13,81 +13,84 @@ describe('components/radar/RadarAnimation', () => {
...
 
     describe('position instance method', () => {
-        test.each([true, false])(`should only position the tether when shown`, isShown => {
+        test.each([true, false])('should only position the tether when shown', isShown => {
             const positionTetherMock = jest.fn();
 
-            const wrapper = getWrapper({ isShown });
+            const wrapper = getWrapper({ isShown } as RadarAnimationProps);
+            // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+            // @ts-ignore: react-tether shenanigans
```
priyajeet: Not yet sure how to fix this lint error, something tether related

src/components/index.ts
```diff
@@ -0,0 +1,11 @@
+export { default as Button } from './button';
```
jstoffan: Is this needed if the component's index file includes `export { default as Button } from './Button';`? Or can this file just be a conglomeration of `export * from './whatever';` statements?

src/components/button/Button.tsx
```diff
@@ -0,0 +1,93 @@
...
+import classNames from 'classnames';
+import LoadingIndicator from '../loading-indicator';
+import RadarAnimation from '../radar';
+
+export enum ButtonType {
+    BUTTON = 'button',
+    RESET = 'reset',
+    SUBMIT = 'submit',
+}
+export interface ButtonProps {
```
jstoffan: Does this leave the interface open to having other properties [merged](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-interfaces) into it? Is that desirable? If not, should we use `type`, instead?

src/components/button/Button.tsx
```diff
@@ -0,0 +1,93 @@
...
+                ref={element => {
+                    this.btnElement = element;
+                    if (setRef) {
+                        setRef(element);
+                    }
+                }}
+                className={styleClassName}
+                type={type}
+                {...buttonProps}
+                onClick={this.handleClick} // Use internal onClick
```
jstoffan: We should consider moving this and any other functional change to a separate changeset to avoid potential bugs.

src/components/primary-button/PrimaryButton.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react';
+import Button, { ButtonProps } from '../button';
+
+const PrimaryButton = ({ children, className = '', ...rest }: Partial<ButtonProps>) => (
```
jstoffan: Partial makes all the props optional, right? Is it needed here?

src/components/loading-indicator/__tests__/makeLoadable.test.tsx
alexkrolick: May need to change the parser option to `typescript` https://prettier.io/docs/en/options.html#parser

src/components/index.ts
priyajeet: Actually I misspoke before. This is correct as is, unless of course we want to refactor existing imports. 

Within src/components/button
`export { default } from './Button';`

So that consumers like src/components/primary-button or parent apps will do (what they already do)
`import Button from '../button';`
`import Button from 'buie/es/components/button';`

If instead inside src/components/button you do `export { default as Button } from './Button';` then consumers will have to do `import { Button } from '../button';`

Now in src/components/index.ts nothing can be default, so default from "button" folder is exported as named. Going up, src/index.ts will just re-export that named export using *. So the parent consumer will then do `import { Button } from 'buie';`

src/components/primary-button/PrimaryButton.tsx
priyajeet: Without partial, TS errors arise when doing `<PrimaryButton className="foo" >` because it complains about missing "type" (and everything else) on it, which is a required prop of "Button". Even though within Button its part of defaultProps, I am guessing typescript cannot infer that from the view point of PrimaryButton.

src/components/button/Button.tsx
priyajeet: The changes all within this function are to fix TS/lint errors

tjgupta: +1

priyajeet: Let me try Omit<T,K>

src/components/loading-indicator/__tests__/makeLoadable.test.tsx
priyajeet: Good call, that worked

.prettierrc.js
```diff
@@ -1 +1,16 @@
-module.exports = require('@box/frontend/prettier/prettierrc.js');
+const prettierrc = require('@box/frontend/prettier/prettierrc.js');
+module.exports = {
+    ...prettierrc,
+    parser: undefined,
```
priyajeet: This will ideally end up in frontend repo

src/components/accessible-svg/AccessibleSVG.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+class AccessibleSVG extends React.Component<AccessibleSVGProps & SVGProps> {
+    id: string = uniqueId('icon');
+
+    render() {
+        const { children, title, ...rest } = this.props;
+        const titleID = `${this.id}-title`;
+
+        // Make sure parent doesn't accidentally override these values
+        const svgProps: Record<string, string | number | React.ReactNode> = omit(rest, ['role', 'aria-labelledby']);
```
priyajeet: Dealing with rest and omit is tricky.

src/components/accessible-svg/AccessibleSVG.js.flow
```diff
@@ -0,0 +1,41 @@
+// @flow
```
alexkrolick: I don't think these .flow files were checked in before

scripts/build-style-vars.js
```diff
@@ -48,7 +53,9 @@ async function main() {
     }
 
     if (!isEqual(priorJson, newJson)) {
-        await writeFile(outputJs, `${moduleHeader}${moduleString}`);
+        await writeFile(outputJs, `${moduleHeader}${moduleString}`); // deprecate eventually
+        await writeFile(outputTs, `${moduleTSHeader}${moduleString}`);
+        execSync(`yarn eslint --fix ${outputTs}`);
```
priyajeet: Only linting ts for now. Helps not needing eslint disable directives.

src/components/accessible-svg/AccessibleSVG.tsx
alexkrolick: where does the Record type come from?

src/components/avatar/__tests__/Avatar.test.tsx
```diff
@@ -1,12 +1,8 @@
 import React from 'react';
+import { shallow, mount } from 'enzyme';
 import { act } from 'react-dom/test-utils';
 import Avatar from '../Avatar';
-
-function MockAvatarImage() {
-    return <div className="avatar-image" />;
-}
-
-jest.mock('../AvatarImage', () => MockAvatarImage);
```
alexkrolick: I think this was here so it didn't make an actual network call for the invalid images

src/styles/variables.ts
```diff
@@ -0,0 +1,83 @@
+/* File auto-generated */
```
alexkrolick: Is this the strictest possible constant definition in TS? In flow you can also do this to make it a constant string instead of regular string:

```
const black: '#000' = '#000'
```

src/components/accessible-svg/AccessibleSVG.js.flow
priyajeet: These flow files are to keep VSCode happy with flow+ts together while you are working within BUIE. Its similar to the ones we generated at publish time because at that time it was only for consuming projects. These ones are for BUIE now.

src/components/accessible-svg/AccessibleSVG.tsx
priyajeet: Its native to TypeScript

src/styles/variables.ts
priyajeet: Good question, I can make it generate that. Ideally this should be an enum like this:

enum bdlVars {
    white = 'white',
    black = 'black',
    ...
}

.eslintrc.js
```diff
@@ -5,7 +5,6 @@ module.exports = {
     rules: {
         camelcase: 'off',
         'class-methods-use-this': 'off',
-        'formatjs/enforce-description': 'error', // fixme
```
priyajeet: moved to @box/platform

.eslintrc.js
```diff
@@ -18,17 +17,23 @@ module.exports = {
         'react/no-access-state-in-setstate': 'off', // fixme
         'react/no-array-index-key': 'off', // fixme
         'react/no-this-in-sfc': 'off',
+        'import/no-unresolved': 'off', // fixme
     },
-    plugins: ['formatjs'],
```
priyajeet: moved to @box/frontend

scripts/styleguide.config.js
jstoffan: Do we need these commented lines? Same for the exports below.

priyajeet: They are currently left there for reference/documentation. Will be nuked eventually.

src/components/button/Button.tsx
jstoffan: Did we figure this out, @priyajeet?

tsconfig.json
```diff
@@ -1,7 +1,11 @@
 {
+    "extends": "@box/frontend/ts/tsconfig",
     "compilerOptions": {
-        "outDir": "./es"
+        "baseUrl": ".",
+        "outDir": "es",
+        "rootDir": "src"
     },
-    "extends": "@box/frontend/ts/tsconfig",
```
priyajeet: If paths came directly from the extended config, that will be relative to that config and not this project. tsconfig is missing tokens like <rootDir> that can be used in the path to avoid ambiguity. See https://github.com/microsoft/TypeScript/issues/25430

tsconfig.json
```diff
@@ -1,7 +1,11 @@
...
-    "files": ["./.storybook/typings.d.ts"]
+    "exclude": ["src/**/*.js", "node_modules", "es", "dist"],
+    "files": [".storybook/typings.d.ts"],
+    "include": ["src/**/*"]
```
jstoffan: Rather than using both `exclude` and `include`, could we do `"include": ["src/**/*.ts", "src/**/*.tsx"]`?

PR #1793, opened 2019-12-10
fix(metadata): Fix misaligned multiselect dropdown

swfree:
Issue: misaligned dropdown in Chrome (Mac), IE11 & Edge

**Mac Chrome - before**
mac_chrome_before

**Mac Chrome - after**
mac_chrome_after

after_mac_chrome_1

**Edge - before**
before_win_edge_1

**Edge - after**
after_win_edge_2

**IE11 - before**
before_ie11

**IE11 - after**
after_ie11

**Safari - after**
after_safari_1

**Firefox - after**
after_mac_firefox_2

PR #1801, opened 2019-12-13
fix(metadata): Fix misaligned singleselect dropdown

swfree:
Same fix as https://github.com/box/box-ui-elements/pull/1793, I had just missed the EnumField component.

**Win Chrome - before**

win-chrome-single-select-before

**Win Chrome - after**

win-chrome-single-select-after

**Mac Chrome - before**

mac-chrome-before

**Mac Chrome - after**

mac-chrome-after

**Safari - before**

mac-safari-before

**Safari - after**

mac-safari-after

**Mac Firefox - before**

mac-firefox-before

**Mac Firefox - after**

mac-firefox-after

IE11 has no change, it wasn't misaligned and looks the same before & after:
ie11-after


boxcla:

PR #1803, opened 2019-12-16
fix(select-field): Reintroduce separators

sahiga:
This PR reintroduces separators into `BaseSelectField` components.

A `SingleSelectField` without the intended separators:
buie-missing-separators

The same `SingleSelectField`, fixed:
buie-bugfix-separators

### DIFF
--- a/src/components/select-field/SelectField.scss
+++ b/src/components/select-field/SelectField.scss
@@ -1,5 +1,6 @@
 @import '../../styles/variables';
 
+.bdl-SelectField,
 .select-field {
     .overlay {
         min-width: 100%;


boxcla:

src/components/select-field/SelectField.scss
```diff
@@ -1,5 +1,6 @@
 @import '../../styles/variables';
 
+.bdl-SelectField,
 .select-field {
```
mxiao6: Where is `.select-field` used? Should we delete it?

sahiga: We're not using `.select-field` in BUIE, but I found overrides for this class name in a couple of other projects.

PR #1807, opened 2019-12-19
fix(activity-feed): Check for component changes before scrolling

sahiga:
This PR fixes a bug in which the ActivityFeed would automatically scroll to the newest comment on every call to componentDidUpdate(), regardless of changes to props or state. The hasLoaded() function will now return true **only** if currentUser or feedItems were previously undefined.


boxcla:

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
```diff
@@ -118,8 +118,16 @@ class ActivityFeed extends React.Component<Props, State> {
...
 
-    hasLoaded = ({ currentUser, feedItems }: Props = this.props): boolean =>
-        currentUser !== undefined && feedItems !== undefined;
+    /**
+     * Determines whether currentUser and feedItems have loaded.
+     * @param prevCurrentUser - The previous value of the currentUser prop
+     * @param prevFeedItems - The previous value of the feedItems prop
+     * @returns {boolean}
+     */
+    hasLoaded = (prevCurrentUser?: User, prevFeedItems?: FeedItems): boolean => {
```
tjgupta: This changes the logic to be more of a `hasJustRecentlyLoaded()` (as in, in the previous cycle `currentUser` and `feedItems` wasn't set, but now it is or isn't), which if I recall we moved away from to simplify other logic. Does this change still work with the loading state in the `render()`? Just double checking, because we ran into a lot of problems with getting this logic working.

sahiga: @tjgupta The loading indicator described in this block shows up while the ActivityFeed is loading items:
```
{isLoading && (
   <div className="bcs-activity-feed-loading-state">
       <LoadingIndicator />
   </div>
)}
```
Is there anything else I should check?

PR #1809, opened 2019-12-21
feat(tasks): using expand groups api endpoint

il10:
Using expand groups api endpoint for assign tasks to groups

Related to #1810 

- [x] handle creating tasks
- [x] handle updating tasks


boxcla:

src/api/Feed.js
```diff
@@ -645,10 +647,15 @@ class Feed extends Base {
...
         try {
             const taskLink = await this.createTaskLink(file, task);
-            const taskAssignments: Array<TaskCollabAssignee> = await Promise.all(
-                assignees.map((assignee: SelectorItem) => {
-                    return this.createTaskCollaborator(file, task, assignee);
-                }),
+            const taskAssignments: Array<TaskCollabAssignee> = flatten<TaskCollabAssignee, TaskCollabAssignee>(
+                await Promise.all(
+                    assignees.map((assignee: SelectorItem): Promise<Array<TaskCollabAssignee> | TaskCollabAssignee> => {
+                        if (assignee.item.type === 'group') {
```
il10: `item` does not contain `type for existing task collaborator selector items, but `assignee` contains `type`. 
Should be `assignee.type`

src/api/Feed.js
```diff
@@ -645,10 +647,15 @@ class Feed extends Base {
 
         try {
             const taskLink = await this.createTaskLink(file, task);
-            const taskAssignments: Array<TaskCollabAssignee> = await Promise.all(
-                assignees.map((assignee: SelectorItem) => {
-                    return this.createTaskCollaborator(file, task, assignee);
-                }),
+            const taskAssignments: Array<TaskCollabAssignee> = flatten<TaskCollabAssignee, TaskCollabAssignee>(
+                await Promise.all(
```
il10: Rather than doing these in parallel, we should do something like Promise.all filter for individual collaborators first and then do Promise.all filter down to groups. That way the group endpoints de-dupes

src/api/tasks/TaskCollaborators.js
```diff
@@ -66,6 +70,43 @@ class TaskCollaborators extends TasksBase {
...
+        group: { id: string },
+        successCallback: Function,
+        task: { id: string },
+    }): void {
+        this.errorCode = ERROR_CODE_CREATE_TASK_COLLABORATOR;
+
+        const requestData = {
+            data: {
+                task: {
+                    type: 'group',
```
il10: Need to change `type: 'group'` to `type: 'task'`

src/api/__tests__/Feed.test.js
```diff
@@ -534,6 +569,58 @@ describe('api/Feed', () => {
...
+
+            const task = {
+                id: '1',
+                description: 'updated description',
+                assignees: [
+                    {
+                        id: '3086276240',
+                        name: 'Test User',
+                        login: 'testuser@foo.com',
+                        item: {
```
il10: Need to remove `type` out of `item` and move it up one level under `assignees`

src/api/__tests__/Feed.test.js
```diff
@@ -534,6 +569,58 @@ describe('api/Feed', () => {
...
+                            name: 'Test User',
+                            login: 'testuser@foo.com',
+                        },
+                    },
+                    {
+                        id: '89321113453',
+                        name: 'Test Group',
+                        login: null,
+                        item: {
+                            type: 'group',
```
il10: Need to remove `type` out of `item` and move it up one level under `assignees`

src/api/tasks/TaskCollaborators.js
```diff
@@ -66,6 +70,43 @@ class TaskCollaborators extends TasksBase {
...
+        errorCallback,
+        file,
+        successCallback,
+        task,
+        group,
+    }: {
+        errorCallback: ElementsErrorCallback,
+        file: BoxItem,
+        group: { id: string },
+        successCallback: Function,
```
booc0mtaco: We can probably type this more specifically. What is the signature and return value for `successCallback`?

il10: Its type is `Function` thoughout the file so I think it makes sense to keep it uniform for consistency until we can track down what the return value/signature is of `successCallback`

src/api/Feed.js
```diff
@@ -401,7 +401,13 @@ class Feed extends Base {
         this.updateFeedItem({ isPending: true }, task.id);
 
         try {
-            await Promise.all(task.addedAssignees.map(assignee => this.createTaskCollaborator(file, task, assignee)));
+            await Promise.all(
+                task.addedAssignees.map(assignee =>
```
booc0mtaco: See https://github.com/box/box-ui-elements/pull/1809/commits/6f02ea3c97c8ccd4886516017e51b5a9be20e241#diff-96176db00fd4ba294b0258debdbd3a25R665 Last week, we chose to check specifically for the two types and reject if some unknown type is encountered.

Since these two should be identical (correct?) I think one of them needs changing. I lean toward the above, as it'll fail in a useful way. We could also `.reject` early to avoid attempting to write known-bad collaborators.

src/api/Feed.js
```diff
@@ -645,12 +654,23 @@ class Feed extends Base {
...
+            /* TODO: change block to process users in parallel and process groups sequentially
+            -check if length of users + groups = assignees (if not throw an error)
+            -filter out the users from assignees
+            -use the result of filter to store the task collaborators in parallel
+            -filter out the groups from assignees
+            -use result of filter to process each group sequentially */
+            const taskAssignments: Array<TaskCollabAssignee> = flatten<TaskCollabAssignee, TaskCollabAssignee>(
+                await Promise.all(
+                    assignees.map((assignee: SelectorItem): Promise<Array<TaskCollabAssignee> | TaskCollabAssignee> =>
+                        // $FlowFixMe
```
il10: $FlowFixMe will be removed in a followup PR where the flow types get resolved @alexkrolick @jstoffan

src/api/tasks/TaskCollaborators.js
```diff
@@ -66,6 +70,43 @@ class TaskCollaborators extends TasksBase {
...
+        task: { id: string },
+    }): void {
+        this.errorCode = ERROR_CODE_CREATE_TASK_COLLABORATOR;
```
jstoffan: Should this have its own error code, since it's a different endpoint?

src/api/tasks/TaskCollaborators.js
```diff
@@ -66,6 +70,43 @@ class TaskCollaborators extends TasksBase {
...
+
+        const requestData = {
```
jstoffan: This variable doesn't seem necessary, as it's just spread again on line 104.

src/api/tasks/TaskCollaborators.js
il10: Yes! That will be coming up in a follow up PR

PR #1810, opened 2019-12-23
feat(tasks): allow selecting groups for task assignment

alexkrolick:
This is behind a feature flip

Related to #1809


boxcla:

src/common/types/core.js
```diff
@@ -131,14 +130,14 @@ type MarkerPaginatedCollection<T> = {
     next_marker: ?string,
 };
 
-type SelectorItem = {
+type SelectorItem<T: any = any> = {
```
alexkrolick: Changed this to a generic so that we can tell what the type of the items are, as we will be using that information to change the display of the selector UI

src/common/types/core.js
```diff
@@ -112,6 +109,8 @@ type User = {
     type: 'user',
 };
 
+type User = UserMini;
```
alexkrolick: There were more "mini definitions" than not, so I changed this to an alias. The definitions were exactly the same.

src/api/MarkerBasedAPI.js
```diff
@@ -89,14 +89,14 @@ class MarkerBasedApi extends Base {
...
      * Marker based API get
-     *
-     * @param {string} id the file id
-     * @param {Function} successCallback the success callback
-     * @param {Function} errorCallback the error callback
-     * @param {string} marker the marker from the start to start fetching at
-     * @param {number} limit the number of items to fetch
-     * @param {Object} params the request query params
-     * @param {boolean} shouldFetchAll true if should get all the pages before calling the sucessCallback
+     * @param {Object} options
```
alexkrolick: This JSDoc was just wrong, the function takes an object

src/api/FileCollaborators.js
```diff
@@ -24,26 +30,34 @@ class FileCollaborators extends MarkerBasedAPI {
...
-        });
+        // Transform into "mention selector" format:
+        const collaboratorSelectorItems: SelectorItems<UserMini | GroupMini> = data.entries.map(
+            (collab: UserMini | GroupMini) => {
+                let item;
+                if (collab.type === 'group') {
+                    item = collab; // flow needs assignment to happen after type refinement
+                } else {
+                    item = collab;
+                    item.email = item.login; // transform user object
```
alexkrolick: Difference: this leaves "login" in the user object as well as aliasing it to "email"; may be useful if we move more towards using the API response format without transforming this field everywhere.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -490,9 +494,16 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
      * @param {Function} successCallback - the success callback
      * @param {Function} errorCallback - the error callback
      * @param {string} searchStr - the search string
+     * @param {Object} [options]
```
alexkrolick: This is the important change - we added a query param for fetching file collaborators that are groups. `approverSelector` uses the query param behind a feature flip; `mentionSelector` does not use it.

src/common/types/core.js
booc0mtaco: We noticed yesterday that the existing `SelectorItem` entries have the UserMini details merged at the top layer instead of under item. For existing functionality, item only contained `text` and `value`.

src/api/FileCollaborators.js
```diff
@@ -58,11 +72,13 @@ class FileCollaborators extends MarkerBasedAPI {
      */
     getFileCollaborators(
         id: string,
-        successCallback: Function,
+        successCallback: ({ entries: Array<SelectorItem<UserMini | GroupMini>>, next_marker: ?string }) => void,
```
booc0mtaco: `SelectorItems<UserMini | GroupMini>` or stop using the `SelectorItems` type altogether?

src/api/FileCollaborators.js
```diff
@@ -7,6 +7,12 @@
 import MarkerBasedAPI from './MarkerBasedAPI';
 import { DEFAULT_MAX_COLLABORATORS } from '../constants';
 import type { ElementsErrorCallback } from '../common/types/api';
+import type { SelectorItem, SelectorItems, UserMini, GroupMini } from '../common/types/core';
+
+type CollaboratorsAPIResponse = {
+    entries: Array<GroupMini | UserMini>,
```
jstoffan: Minor, but in other instances this is declared as `UserMini | GroupMini`.

src/api/Feed.js
```diff
@@ -403,8 +405,7 @@ class Feed extends Base {
         try {
             await Promise.all(
                 task.addedAssignees.map(assignee =>
-                    // $FlowFixMe
-                    assignee.type === 'group'
+                    assignee.item && assignee.item.type === 'group'
```
jstoffan: It seems odd that the the prior logic didn't cause any tests to fail.

src/api/Feed.js
```diff
@@ -663,9 +664,10 @@ class Feed extends Base {
             -use result of filter to process each group sequentially */
             const taskAssignments: Array<TaskCollabAssignee> = flatten<TaskCollabAssignee, TaskCollabAssignee>(
                 await Promise.all(
-                    assignees.map((assignee: SelectorItem): Promise<Array<TaskCollabAssignee> | TaskCollabAssignee> =>
-                        // $FlowFixMe
-                        assignee.type === 'group'
+                    assignees.map((assignee: SelectorItem<UserMini | GroupMini>): Promise<
+                        Array<TaskCollabAssignee> | TaskCollabAssignee,
```
jstoffan: The logic is getting difficult to read due to the line length and nesting. Would adding some named methods help?

src/api/FileCollaborators.js
```diff
@@ -24,26 +30,34 @@ class FileCollaborators extends MarkerBasedAPI {
...
-                name,
-                item: { id, name, email: login },
-            };
-        });
+        // Transform into "mention selector" format:
+        const collaboratorSelectorItems: SelectorItems<UserMini | GroupMini> = data.entries.map(
+            (collab: UserMini | GroupMini) => {
+                let item;
+                if (collab.type === 'group') {
+                    item = collab; // flow needs assignment to happen after type refinement
```
jstoffan: To confirm, this means that the following wouldn't work?

```
const item = { ...collab };

if (item.type !== 'group') {
    item.email = item.login;
}

return {
    id: item.id,
    name: item.name,
    item,
};

File: src/elements/content-sidebar/activity-feed/task-form/TaskForm.js [hunk 11]
```diff
@@ -331,13 +329,13 @@ class TaskForm extends React.Component<Props, State> {
...
-        const renderApprovers = convertAssigneesToSelectorItems(approvers);
+        const selectedApprovers = convertAssigneesToSelectorItems(approvers);
 
         // filter out selected approvers
         // map to datalist item format
         const approverOptions = approverSelectorContacts
-            .filter(({ id }) => !renderApprovers.find(({ value }) => value === id))
-            .map(({ id, item }) => ({ ...item, text: item.name, value: id }));
+            .filter(({ id }) => !selectedApprovers.find(({ value }) => value === id))
+            .map(({ id, item = {} }) => ({ ...item, text: item.name, value: id, item }));
```
jstoffan: Should we default `item` to an empty object at a higher level to avoid needing this in two places?

src/api/Feed.js
alexkrolick: The previous code path was not connected to the API (this PR does that), so tests for it used mock data matching the expected shape. The assumption was that the selector object would be equivalent to {...item} + other keys, but this refactoring uses the safer pattern by accessing the `item` instead. The test data has `type` for item already.

src/api/Feed.js
alexkrolick: We will be replacing this code soon with a transactional API call, instead of these nested `Promise.all`s.

src/api/FileCollaborators.js
alexkrolick: Correct.

```
Cannot assign `item.login` to `item.email` because property `email` is missing in  object literal [1].Flow(InferError)
```

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
alexkrolick: Checked and we can remove this .map, the data is formatted enough already.

PR #1818, opened 2019-12-31
feat(archive): Add json to rep hint

mxiao6:


boxcla:

src/constants.js
```diff
@@ -285,7 +285,7 @@ export const METRIC_TYPE_ELEMENTS_LOAD_METRIC: 'elements_load_metric' = 'element
 export const IS_ERROR_DISPLAYED = 'isErrorDisplayed'; // used to determine if user will see some error state or message
 
 /* ------------- Representation Hints ------------------- */
-const X_REP_HINT_BASE = '[3d][pdf][text][mp3]';
+const X_REP_HINT_BASE = '[3d][pdf][text][mp3][json]';
```
jstoffan: What is the effect of adding this to this project's `/file` calls? What need does it fulfill?

mxiao6: > What is the effect of adding this to this project's `/file` calls? What need does it fulfill?

If user preview files through BUIE, the preview cache is built inside BUIE and passed into preview SDK directly. If json is not added to rep hint, the cache will not have json rep in cache but the cache is valid because it has all the fields.

PR #1820, opened 2020-01-03
Security controls ui

feixuanli:
1. Next to any policy summary text CHANGE the orange dot to a orange shield icon (NEW):
A. Right Rail
B. Drop Down picker for Classification
C. Security controls modal (not in mock)

2. Update right rail policy information:
A. Remove Tooltip Hover
B. CHANGE to "View All" blue link
C. Clicking link goes to (NEW) read-only view of Security controls modal 

3. Security controls modal  (Tooltip in mock is removed)

siderail
security constrols modal
classify modal


boxcla:

src/features/classification/security-controls/mixins.scss
```diff
@@ -0,0 +1,20 @@
+@import '../../../styles/variables';
+
+$scrollbar-width: 7px;
+
+@mixin BackgroundlessScrollbar {
+    // Currently only supported by Firefox.
+    // For Chrome/Safari we use the styles below
+    scrollbar-color: $bdl-gray-50 transparent;
```
mickr: How does this look on IE 11?

src/icons/general/IconSecurityClassification.js
```diff
@@ -31,11 +31,11 @@ const IconSecurityClassification = ({
                 d="M17,2 L5,8 L5,15 C5,21.4214876 10.6933333,29.5421488 17,31 C23.3066667,29.5421488 29,21.4214876 29,15 L29,8 L17,2 Z"
                 stroke={color}
                 strokeWidth={strokeWidth}
-                fill="none"
+                fill={className === 'reverse' ? color : 'none'}
```
jstoffan: Hm, I'm not sure we should be driving logic based on the class name. @booc0mtaco, thoughts?

src/features/classification/security-controls/mixins.scss
```diff
@@ -0,0 +1,20 @@
+@import '../../../styles/variables';
+
+$scrollbar-width: 7px;
+
+@mixin BackgroundlessScrollbar {
```
jstoffan: This mixin should be prefixed with `bdl-` if it's needed.

src/features/classification/security-controls/mixins.scss
```diff
@@ -0,0 +1,20 @@
...
+    // Currently only supported by Firefox.
```
jstoffan: These two comments are a little confusing. What's only supported by Firefox?

src/icons/general/IconSecurityClassification.js
booc0mtaco: I don't think we want to do this. 

- `reverse` isn't a class name in this case, correct? so the prop isn't the value
- if `reverse` **is** a classname, it should be SUIT-able, correct?
- as a class name, we could specify the color to use via SASS

booc0mtaco:
NOTE for reviewers: the individual commits have the proper commit format

src/features/classification/Classification.scss
```diff
@@ -8,7 +9,12 @@
 }
 
 .bdl-Classification-definition {
+    @include BackgroundlessScrollbar;
```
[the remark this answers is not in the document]
booc0mtaco: Agreed. Ideally, we'd have this somewhere in the common styles, and using a SUIT-able name, treating the scrollbar as the component perhaps 

ex. somewhere here (possibly in a new file): https://github.com/box/box-ui-elements/tree/master/src/styles/mixins

cc @jstoffan

src/icons/general/IconSecurityClassification.js
feixuanli: Thanks! I reverted the update and have the inverted icon as a separated component.

src/features/classification/security-controls/SecurityControlsModal.js
```diff
@@ -0,0 +1,84 @@
...
+import SecurityControlsItem from './SecurityControlsItem';
+import { CLASSIFICATION_COMMUNITY_URL } from '../constants';
+import { Link } from '../../../components/link';
+import messages from './messages';
+import classificationMessages from '../messages';
+import './SecurityControlsModal.scss';
+
+type Props = {
+    classificationName: string,
+    closeModal: Function,
```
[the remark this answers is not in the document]
feixuanli: I would keep it as Function to be consistent with other usages.

src/features/classification/security-controls/mixins.scss
feixuanli: Thanks for bringing this up! Confirming with design team, I reverted the update here.

src/features/classification/Classification.scss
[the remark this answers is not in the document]
feixuanli: Thanks! Update reverted.

i18n/en-US.properties
```diff
@@ -1048,6 +1052,10 @@ boxui.securityControls.mobileDownloadExternalOwnersEditors = Download restricted
 boxui.securityControls.mobileDownloadOwners = Download restricted on mobile, except Owners/Co-Owners.
 # Bullet point that summarizes mobile download restrictions applied to classification, when restriction applies to managed users except Owners/Co-Owners/Editors
 boxui.securityControls.mobileDownloadOwnersEditors = Download restricted on mobile, except Owners/Co-Owners/Editors.
+# Description for modal to display classification and security controls details
+boxui.securityControls.modalDescription = Classification labels defined by your administrator can be used to label content and apply security policies. {learnMoreLink}
```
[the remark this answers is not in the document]
feixuanli: reverted

src/features/classification/security-controls/mixins.scss
feixuanli: reverted

src/features/classification/security-controls/mixins.scss
feixuanli: reverted

src/features/classification/security-controls/SecurityControls.js
```diff
@@ -1,62 +1,102 @@
...
 import * as React from 'react';
 import classNames from 'classnames';
+import { FormattedMessage } from 'react-intl';
 
 import type { Controls, ControlsFormat } from '../flowTypes';
 
 import SecurityControlsItem from './SecurityControlsItem';
 import { getShortSecurityControlsMessage, getFullSecurityControlsMessages } from './utils';
 import { DEFAULT_MAX_APP_COUNT, SECURITY_CONTROLS_FORMAT } from '../constants';
+import SecurityControlsModal from './SecurityControlsModal';
```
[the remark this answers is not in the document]
Sunyang730: @jstoffan, is this something that will be add to the eslint rule itself? It seems like a style nit pick that I am seeing from different PR now. Would be good to make it a standard to avoid time spend during development.

src/features/classification/security-controls/SecurityControlsModal.js
```diff
@@ -0,0 +1,84 @@
...
+            isOpen={isSecurityControlsModalOpen}
+        >
+            <p>
+                <FormattedMessage {...messages.modalDescription} values={{ learnMoreLink }} />
+            </p>
+            <ClassifiedBadge name={((classificationName: any): string)} />
+            <Label text={<FormattedMessage {...classificationMessages.definition} />}>
+                <p className="bdl-Classification-definition">{definition}</p>
+            </Label>
+            <ul className="bdl-SecurityControls">
```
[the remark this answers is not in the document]
feixuanli: Thanks @jstoffan , I left this as-is bc securitycontrolsmodal is utilizing securityControlsItem (not securityControls) so it would not create a circular dependency. As discussed with @patrickpaul, we were thinking to create two separate child components for `list of securityControlsItem` and `viewAll button + modal` if that would make it cleaner. But as implementing, I realized it would create additional layers/props checking,  yet the logic in security controls component is not made simpler. Although that `list of securityControlsItem` is also used in the modal, we have assigned different classnames on the ul element. Thus I think it makes sense to keep them separated.

src/features/classification/security-controls/SecurityControls.js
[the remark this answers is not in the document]
jstoffan: I agree that would be ideal.

src/features/classification/security-controls/__tests__/SecurityControlsModal.test.js
```diff
@@ -0,0 +1,48 @@
...
+    beforeEach(() => {
+        modalItems = [
+            { id: 'msg1', defaultMessage: 'message1' },
+            { id: 'msg2', defaultMessage: 'message2' },
+        ];
+        wrapper = getWrapper();
+    });
+
+    afterEach(() => {
+        jest.resetAllMocks();
```
jstoffan: I don't think this is needed.

src/features/classification/security-controls/__tests__/SecurityControlsModal.test.js
```diff
@@ -0,0 +1,48 @@
...
+        expect(wrapper).toMatchSnapshot();
+    });
+
+    test('should render with correct number of security controls items', () => {
+        modalItems = [
+            { id: 'msg1', defaultMessage: 'message1' },
+            { id: 'msg2', defaultMessage: 'message2' },
+            { id: 'msg3', defaultMessage: 'message3' },
+        ];
+        wrapper.setProps({ modalItems });
```
jstoffan: We usually prefer to have each test encapsulate its instance to avoid side effects. If you allow each test to call `getWrapper`, you can pass these prop values in directly rather than calling `setProps`.

src/features/classification/security-controls/__tests__/SecurityControlsModal.test.js
```diff
@@ -0,0 +1,37 @@
...
+        ];
+        wrapper = getWrapper();
+    });
+
+    afterEach(() => {
+        jest.resetAllMocks();
+    });
+
+    test('should render a SecurityControlsModal', () => {
+        expect(wrapper).toMatchSnapshot();
```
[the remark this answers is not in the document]
jstoffan: The tests still seem pretty light. What are we not testing that the snapshot is helping us catch?

src/features/classification/security-controls/SecurityControls.scss
```diff
@@ -1,14 +1,22 @@
+@import '../../../styles/variables';
+
 .bdl-SecurityControls {
     margin: 0;
     list-style: none;
 }
 
+.bdl-IconSecurityClassificationSolid {
```
jstoffan: I think this will hide the icon everywhere by default. Intended? Can we choose to simply not render the icon except in the summarized view, instead?

src/features/classification/security-controls/SecurityControlsModal.scss
```diff
@@ -0,0 +1,32 @@
+@import '../../../styles/variables';
+
+.bdl-SecurityControlsModal {
+    .link {
+        color: $bdl-box-blue !important;
+        text-decoration: none;
+
+        &:hover {
```
jstoffan: Does this need a `&:focus` rule, as well?

src/features/classification/security-controls/SecurityControls.js
```diff
@@ -1,62 +1,98 @@
...
+                    <>
+                        <PlainButton className="lnk" onClick={this.openModal} type="button">
+                            <FormattedMessage {...messages.viewAll} />
+                        </PlainButton>
+                        <SecurityControlsModal
+                            classificationName={classificationName}
+                            closeModal={this.closeModal}
+                            definition={definition}
+                            itemName={itemName}
+                            isSecurityControlsModalOpen={this.state.isSecurityControlsModalOpen}
```
jstoffan: Minor, but this value should be destructured rather than accessed directly.

src/features/classification/security-controls/SecurityControlsModal.scss
```diff
@@ -0,0 +1,32 @@
+@import '../../../styles/variables';
+
+.bdl-SecurityControlsModal {
+    .link {
+        color: $bdl-box-blue !important;
```
mickr: Do we really need to have the `!important` here?

src/features/classification/security-controls/__tests__/SecurityControlsModal.test.js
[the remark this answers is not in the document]
feixuanli: This component is just rendering child components with props provided. Test using snapshot would verify the dom structure. Here we added test for the number of securitycontrols items. I'm thinking  we don't need to test for each child component in addition to the snapshot.

src/features/classification/security-controls/SecurityControlsModal.js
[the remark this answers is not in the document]
alexkrolick: Try to follow the flow strict mode guidelines - no any, Function, Object types. You can change the comment at the top of the file to enforce this.

```diff
- // @flow
+ // @flow strict
```

src/features/classification/security-controls/SecurityControls.scss
feixuanli: Yes this is intended. I keep it this way for in this case showing/hiding icon using styles simplifies the logics and props passing/checking in components.

src/features/classification/security-controls/SecurityControlsModal.scss
feixuanli: removed

src/features/classification/security-controls/SecurityControlsModal.scss
feixuanli: removed since we have removed the link element in the modal.

src/features/classification/security-controls/__tests__/SecurityControls.test.js
```diff
@@ -69,16 +60,18 @@ describe('features/classification/security-controls/SecurityControls', () => {
 
     test('should set summarized class when not using FULL controlsFormat', () => {
         wrapper.setProps({ controlsFormat: FULL });
-        expect(wrapper.hasClass('bdl-SecurityControls--summarized')).toBe(false);
+        expect(wrapper.find('ul').hasClass('bdl-SecurityControls--summarized')).toBe(false);
```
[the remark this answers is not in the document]
feixuanli: This specific test is to verify classes applied to ul element depending on controls format. There is no data to test instead.

src/features/classification/security-controls/SecurityControlsModal.js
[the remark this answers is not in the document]
Sunyang730: Isn't the closeModal just the `() => void` ?

src/features/classification/security-controls/SecurityControls.js
```diff
@@ -1,62 +1,102 @@
...
-const SecurityControls = ({ controls, controlsFormat, maxAppCount }: Props) => {
-    let items = [];
-    let tooltipItems;
+type State = {
+    isSecurityControlsModalOpen: boolean,
+};
+
+class SecurityControls extends React.Component<Props, State> {
+    static defaultProps = {
+        classificationName: '',
```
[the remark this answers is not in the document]
Sunyang730: For the newly added props, you will need to have default. I believe @jstoffan is referring to controls, and controlsFormat didn't have a default before

src/features/classification/security-controls/SecurityControlsItem.js
```diff
@@ -1,49 +1,30 @@
 // @flow
 import * as React from 'react';
-import { FormattedMessage } from 'react-intl';
+import { FormattedMessage, type MessageDescriptor } from 'react-intl';
```
Sunyang730: nit: Is this a common pattern for imports? @jstoffan I thought we have the stand on separating imports and type should be on its own

src/features/classification/security-controls/SecurityControlsItem.js
```diff
@@ -1,49 +1,30 @@
...
 
-import type { MessageDescriptor } from 'react-intl';
-
-import Tooltip from '../../../components/tooltip';
-import IconInfo from '../../../icons/general/IconInfo';
-import { bdlBoxBlue } from '../../../styles/variables';
+import { bdlYellorange } from '../../../styles/variables';
+import IconSecurityClassificationSolid from '../../../icons/general/IconSecurityClassificationSolid';
+import { SECURITY_CONTROLS_FORMAT } from '../constants';
+import type { ControlsFormat } from '../flowTypes';
```
Sunyang730: nit: Similar here. imports should be modules, new line, full path, new line, dot dot path, new line, and then type @jstoffan confirmation please

src/features/classification/security-controls/__tests__/SecurityControlsModal.test.js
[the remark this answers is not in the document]
Sunyang730: @jstoffan I agree, however, do we have a standard on when to use snapshot matching and when not to? In this case I see a valid usage but I also understand that snapshot will get oversight in PRs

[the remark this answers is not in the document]
mickr: I think they are complementary to each other but using ONLY snapshots requires a higher level of discipline from a developer perspective as they can be affected from outside forces quite often and then it becomes second nature just to update them without verifying they are correct or not. Once a bad snapshot gets committed, your tests are useless.

PR #1823, opened 2020-01-07
chore(deps): Upgrade Cypress to 3.8.1

jstoffan:
[Cypress Changelog](https://docs.cypress.io/guides/references/changelog.html)


boxcla:

test/integration/content-explorer/ContentExplorer.int.test.js
```diff
@@ -192,14 +192,14 @@ describe('ContentExplorer', () => {
             helpers.checkRowSelections(1);
             helpers.selectRow(4);
             helpers.checkRowSelections(4);
-            helpers.previewItemFromRow(2);
```
jstoffan: This is a rather large 3d file that can occasionally time out. I just switched the test to use smaller files.

booc0mtaco: do you think this could be why we were occasionally seeing this error? https://docs.cypress.io/guides/references/error-messages.html#The-Chromium-Renderer-process-just-crashed

jstoffan: No idea. I guess we'll find out!

PR #1835, opened 2020-01-09
fix(textinput): have TextInput respect description accessibility

omgrief:


boxcla:
Hi @omgrief, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

src/components/text-input/TextInput.js
```diff
@@ -37,6 +38,7 @@ type Props = {
 
 const TextInput = ({
     className = '',
+    id,
```
jstoffan: Is there a strong need for `id` to be passed in or can we just generate a unique id every time?

src/components/label/LabelPrimitive.js
```diff
@@ -3,14 +3,15 @@ import * as React from 'react';
...
-    children: React.Node,
+    children?: React.Node,
     className?: string,
+    htmlFor?: string,
     labelContent: React.Node,
 };
 
-const LabelPrimitive = ({ children, className, labelContent, ...rest }: Props) => (
+const LabelPrimitive = ({ children, className, labelContent, htmlFor, ...rest }: Props) => (
     // eslint-disable-next-line jsx-a11y/label-has-for
```
jstoffan: Can we re-enable this rule?

src/components/text-input/TextInput.js
omgrief: probably not other than to prevent breaking existing implementations

omgrief:
addressed comments regarding the allowing of label/input by removing the change to make `children` optional.

former is unneeded, if the standard is what it is

src/components/label/LabelPrimitive.js
```diff
@@ -5,12 +5,12 @@ import classNames from 'classnames';
...
     className?: string,
+    htmlFor?: string,
     labelContent: React.Node,
 };
 
-const LabelPrimitive = ({ children, className, labelContent, ...rest }: Props) => (
-    // eslint-disable-next-line jsx-a11y/label-has-for
-    <label>
+const LabelPrimitive = ({ children, className, labelContent, htmlFor, ...rest }: Props) => (
+    <label htmlFor={htmlFor}>
```
jstoffan: I don't think `htmlFor` is necessary when the content is nested inside the `label` tag. The browser should associate them automatically.

alexkrolick: We should support using it unnested, though, right? Since the fact that this is a plain `label` without classname is not guaranteed.

omgrief: because we decided to go with making it all have to be nested anyways, I'm just gonna remove it and resolve this.

src/components/text-input/TextInput.js
omgrief: removing `htmlFor` and relying on just nesting everything removes the need for this

src/components/label/Label.js
```diff
@@ -52,7 +61,7 @@ const Label = ({ text, tooltip, infoTooltip, infoIconProps, showOptionalText, hi
     }
 
     return (
-        <StandardLabel labelContent={labelContent} tooltip={tooltip}>
+        <StandardLabel labelContent={labelContent} tooltip={tooltip} {...rest}>
```
jstoffan: It looks like all the changes to `Label` can be reverted, I think?

omgrief: done

jstoffan:
I spoke offline with @omgrief and I'm going to merge on behalf, as his team hasn't yet been added to the necessary org in Github.

PR #1836, opened 2020-01-10
feat(theme): Adding a new theme component and provider

akahn87:
- First cut of getting theming and theming generation into Box UI Elements
- Simple Styled Components implementation example for wrapping Box UI Elements Buttons
- Contains Unit test of a randomly generated set of 1k color variations to ensure accessible contrast output.

Please also review the object keys to consider use cases throughout the application. This isn't meant to capture all usage upfront, but to act as a base implementation.

**Next Steps**
- Fine tune secondary and accent generation to the Box Design Language.
- Determine how this will look for partner integrations. Consider how this will look for implementation in components and whether ThemeProvider will be a requirement for partners.
- Extend example use case to demo more of the theming functionality.
- Write Storybook story for theme component with knob for color picking.
- Implement various color modes with preset primaries, and potentially secondary / accent. Will need to determine if we want to have this outside the dynamic scope or not.
- Monitor Styled Components over next week for 5.x release


boxcla:

examples/src/ThemeExample.js
```diff
@@ -0,0 +1,43 @@
...
+        border-color: ${props => props.theme.primary.borderActive};
+    }
+`;
+
+type Props = {
+    colorKey: ?String,
+};
+
+const ThemeExample = ({ colorKey }: Props) => {
+    let dynamicTheme = {};
```
jstoffan: Looks like this can be simplified a bit:

```
const dynamicTheme = colorKey ? generateContrastColors(colorKey) : {};
const theme = merge({}, useTheme(), dynamicTheme);
```

examples/src/ThemeExample.js
```diff
@@ -0,0 +1,43 @@
...
+    if (colorKey) {
+        dynamicTheme = generateContrastColors(colorKey);
+    }
+    const theme = useTheme();
+
+    return (
+        <ThemeProvider theme={merge(theme, dynamicTheme)}>
```
jstoffan: Should this be `merge({}, theme, dynamicTheme})` to avoid mutating `theme`? Or does `useTheme` already creat a clone for you?

scripts/styleguide.config.js
```diff
@@ -98,6 +98,7 @@ const allSections = [
             '../src/components/text-area/TextArea.js',
             '../src/components/text-input/TextInput.js',
             '../src/components/text-input-with-copy-button/TextInputWithCopyButton.js',
+            '../src/components/theme/index.js',
```
jstoffan: I believe the new standard is to create Storybook stories rather than Styleguidist examples.

src/components/theme/__tests__/theme.test.js
```diff
@@ -0,0 +1,164 @@
...
+import * as vars from '../../../styles/variables';
+
+describe('Validate generated color accessibility', () => {
+    const accessibleKeys = ['backgroundHover', 'backgroundActive', 'borderHover', 'borderActive'];
+    const contrastKey = 'foreground';
+
+    // Potentially problematic colors merged with random 1000 swatches from fixture set.
+    const colorCases = ['#ffffff', '#000000', '#232300', '#c50213', ...randomSwatches];
+
+    test.each(colorCases)('given %p as argument, returns accessible color contrast', colorKey => {
```
jstoffan: This is going to add 1000+ tests to our runs. Are so many random swatches needed? Any impact on speed? Should this just be a single test with a loop inside it, instead?

src/components/theme/__tests__/theme.test.js
```diff
@@ -0,0 +1,164 @@
...
+            "foreground": "#fff",
+          },
+        }
+    `);
+});
+
+test('should generate box brand colors', () => {
+    const dynamicTheme = generateContrastColors(vars.bdlBoxBlue);
+
+    expect(dynamicTheme).toMatchInlineSnapshot(`
```
jstoffan: I think we've been avoiding inline snapshots. Can we use a Jest matcher like `toContain`, instead?

src/components/theme/__tests__/theme.test.js
```diff
@@ -0,0 +1,164 @@
...
+          },
+        }
+    `);
+});
+
+test('should generate box brand colors and merge into base theme', () => {
+    const dynamicTheme = generateContrastColors(vars.bdlBoxBlue);
+    const baseTheme = theme;
+
+    expect(merge(baseTheme, dynamicTheme)).toMatchInlineSnapshot(`
```
jstoffan: It seems like we're testing the behavior of `merge` here.

src/components/theme/__tests__/theme.test.js
```diff
@@ -0,0 +1,164 @@
+import merge from 'lodash/merge';
+import Color from 'color';
+import randomSwatches from '../../../../test/fixtures/theme/colors';
+import { theme } from '../theme';
+import { generateContrastColors, MIN_CONTRAST } from '../contrast-colors';
+import * as vars from '../../../styles/variables';
+
+describe('Validate generated color accessibility', () => {
```
jstoffan: Most of our tests have an outer `describe` with the path of the component.

src/components/theme/contrast-colors.js
```diff
@@ -0,0 +1,82 @@
...
+// When converting from rgb/hsl to hex there is potential for
+// flattening of the color space, so we add an offset factor to account for it.
+const OFFSET_FACTOR = 0.05;
+export const MIN_CONTRAST = 4.5;
+
+// The yiq coefficients in the color library are incorrect
+// http://poynton.ca/notes/colour_and_gamma/ColorFAQ.html#RTFToC9
+function getYiq(color: string) {
+    const rgb = Color(color).rgb().color;
+    const yiq = (rgb[0] * 2126 + rgb[1] * 7152 + rgb[2] * 722) / 10000;
```
jstoffan: We can just return this value immediately to avoid a one-use variable.

src/components/theme/contrast-colors.js
```diff
@@ -0,0 +1,82 @@
...
+import Color from 'color';
+import type { ColorType } from 'color';
+import * as vars from '../../styles/variables';
+
+// When converting from rgb/hsl to hex there is potential for
+// flattening of the color space, so we add an offset factor to account for it.
+const OFFSET_FACTOR = 0.05;
+export const MIN_CONTRAST = 4.5;
+
+// The yiq coefficients in the color library are incorrect
```
jstoffan: What is `yiq`?

src/components/theme/theme.js
```diff
@@ -0,0 +1,38 @@
+/* eslint-disable import/prefer-default-export */
```
jstoffan: Can we avoid disabling eslint in these files?

src/components/theme/theme.js
```diff
@@ -0,0 +1,38 @@
...
+// Or using import { withTheme } from 'styled-components';
+
+import * as vars from '../../styles/variables';
+
+export const theme = {
+    base: {
+        background: vars.white,
+        backgroundHover: vars.bdlGray30,
+        backgroundActive: vars.bdlGray62,
+        // backgroundMuted: vars.bdlGray10, // STATIC
```
jstoffan: Let's try to avoid commented code.

test/fixtures/theme/colors.js
```diff
@@ -0,0 +1,1012 @@
+// const NUM_SWATCHES = 1000;
+// const colorCases = [];
+
+// for (let i = 0; i < NUM_SWATCHES; i += 1) {
```
jstoffan: Can we move this to a function to avoid the commented code?

src/components/theme/__tests__/theme.test.js
akahn87: Yeah, this is a pretty gnarly one.. I think the solution that i'd like to do here is pull in a library like: https://www.npmjs.com/package/jest-expect-message so that I can get the proper information out of a failed expectation. The problem I encountered and why i'm using a jest each here is so that I know the failing swatch. Using jest-expect-message, I can output the swatch and also the particular accessible key check i'm doing. 

If possible, i'd like to do that one in a follow-up since it requires addition to the Jest configuration, unless you have some other thoughts around that one.

package.json
```diff
@@ -298,6 +300,7 @@
         "react-virtualized": "^9.21.0",
         "regenerator-runtime": "^0.13.2",
         "scroll-into-view-if-needed": "^2.2.20",
+        "styled-components": "5.0.0-rc.3",
```
j3tan: use `^` if possible

src/components/theme/__tests__/theme.test.js
j3tan: overriding `expect()` might be problematic if they add a param of their own later. We may just want to write our own custom expect matcher instead

src/components/theme/__tests__/theme.test.js
akahn87: I'll move these over to a separate snapshot instead of inline for convention

scripts/styleguide.config.js
akahn87: I'll be looking to create the story for this component in the next iteration where I can get a color knob in place as well.

src/components/theme/contrast-colors.js
akahn87: This is the way we're checking the luminance of a particular colorKey to determine if it's a light or dark color.

src/components/theme/__tests__/theme.test.js
akahn87: A customer matcher could potentially work, but would still need a way to pipe in the metadata about the contrast ratios we are comparing.

package.json
```diff
@@ -269,6 +270,7 @@
         "@hapi/address": "^2.0.0",
         "axios": "^0.18.1",
         "classnames": "^2.2.5",
+        "color": "^3.1.2",
```
priyajeet: should add breaking change to PR

examples/src/ThemeExample.js
```diff
@@ -0,0 +1,40 @@
+// @flow
+import React from 'react';
+import styled, { useTheme, ThemeProvider } from 'styled-components';
+import merge from 'lodash/merge';
+
+import BoxButton from '../../src/components/button';
+import { generateContrastColors } from '../../src/components/theme';
+
+const Button = styled(BoxButton)`
```
priyajeet: Should Button.js export this?

priyajeet: Basically, if theme provider exists, then spit out the styled button, otherwise spit out the regular button.

package.json
akahn87: 👍 do we do that because of the change to a peerdep? or is that because of the minor version bump from 3.0.x -> 3.1.x

examples/src/ThemeExample.js
akahn87: I'm looking to make a design doc around this to see how we might be able to provide a default theme into each of the BUIE components and allow it to be overridden by the provider context. This way, we'll have a single, consistent way of theming our components.

Doing this would mean the override path for styling would be `.btn` as the theming would be a hash of the styled component styling.

That implementation would mean, the export from the Button component would contain the styled component instead of a separate export.

e.g.
```
const StyledButton = button.styled`
  background: ${props => props.theme.primary.background}
`

...

return (
    <StyledButton
        className={styleClassName}
    >
        <span className="btn-content">{children}</span>
    </StyledButton>
)
```

priyajeet: Btw one should still discuss alternatives like the older approach now that one knows (before we were basing this approach on the premise that more components will need to be styled and after talking with deisn sounds like thats not the case) that nothing but the button and left bar will be themed. And since only those two, does it still make sense to introduce styled components or is that an overkill compared to prior style override approach.

PR #1846, opened 2020-01-13
feat(uploads): save the mb uploaded on resume

danubevictoria:
Adds a new optional field on upload item that saves the mb uploaded so far for an item when user resumes an upload

### DIFF
--- a/src/elements/content-uploader/ContentUploader.js
+++ b/src/elements/content-uploader/ContentUploader.js
@@ -999,6 +999,7 @@ class ContentUploader extends Component<Props, State> {
                 break;
             case STATUS_ERROR:
                 if (isResumable) {
+                    item.bytesUploadedOnLastResume = item.api.totalUploadedBytes;
                     this.resumeFile(item);
                 } else {
                     this.resetFile(item);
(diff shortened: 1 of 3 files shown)


boxcla:
Verified that @danubevictoria has signed the CLA. Thanks for the pull request!

src/elements/content-uploader/ContentUploader.js
```diff
@@ -999,6 +999,7 @@ class ContentUploader extends Component<Props, State> {
                 break;
             case STATUS_ERROR:
                 if (isResumable) {
+                    item.loadedLastResume = item.api.totalUploadedBytes;
```
stevegcoder: Did you consider just adding a field in item called totalUploadedBytes and update that in onProgress?  Seems like this field loadedLastResume is very specific to resuming an upload and I think it may be useful for other stuff too.

danubevictoria: It's technically already doing that with `item.api.totalUploadedBytes` but it updates it every time a part is completed, but it's not being stored in qubole. For our purposes, didn't we just want to cache it only if it's a resumable upload when they click the resume button? And only want to log to qubole on successful upload. The table in qubole too is specific to just uploads that are resumed.

PR #1848, opened 2020-01-14
feat(changelog): Add CHANGELOG.md and changelog generation package.

stevegcoder:


boxcla:

stevegcoder:
This is a Work in Progress.

priyajeet:
Why is this needed? You can see the changelog under the releases tab. Is this needed for something else?

booc0mtaco:
Agree with @priyajeet . We have the auto-generated changelog in GitHub already, and there's an API to pull that description as [`body`](https://developer.github.com/v3/repos/releases/#get-a-release-by-tag-name). If this is needed for scripting, would it be possible to use the API instead? that way the changelog info would never diverge. Also, since `tag_name` is stable, that may be enough to get the proper `body` content, which would be the changelog in this case.

PR #1850, opened 2020-01-14
feat(collapsible): display action button when collapsed and hovered

indiramuru:
UX change was requested to add hover effect to the Collapsible component, and they are ok with having this as the default behavior of the component.
Expected behavior: In collapsed state, hovering over the card header should display the Edit button and underline the card title. 
Screen Shot 2020-01-15 at 10 02 46 PM

### DIFF
--- a/src/components/collapsible/Collapsible.js
+++ b/src/components/collapsible/Collapsible.js
@@ -82,6 +82,14 @@ class Collapsible extends React.PureComponent<Props, State> {
         this.setState({ isHovered: false });
     };
 
+    handleFocus = () => {
+        this.setState({ isHovered: true });
+    };
+
+    handleBlur = () => {
+        this.setState({ isHovered: false });
+    };
+
     render() {
         const { isHovered, isOpen }: State = this.state;
         const {
@@ -119,8 +127,12 @@ class Collapsible extends React.PureComponent<Props, State> {
             <div className={sectionClassName}>
                 <div
                     className={buttonClassName}
+                    onBlur={this.handleBlur}
+                    onFocus={this.handleFocus}
                     onMouseEnter={this.handleMouseEnter}
                     onMouseLeave={this.handleMouseLeave}
+                    // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
+                    tabIndex="0"
                 >
                     <PlainButton
                         {...modifiedButtonProps}
--- a/src/components/collapsible/Collapsible.js
+++ b/src/components/collapsible/Collapsible.js
@@ -117,7 +117,7 @@ class Collapsible extends React.PureComponent<Props, State> {
                         {title}
                         <IconCaretDown className="collapsible-card-header-caret" color={bdlGray50} width={8} />
                     </PlainButton>
-                    {!!headerActionItems && <span className="bdl-collapsible-actionItems">{headerActionItems}</span>}
+                    {!!headerActionItems && <span className="bdl-Collapsible-actionItems">{headerActionItems}</span>}
                 </div>
                 <AnimateHeight duration={animationDuration} height={isOpen ? 'auto' : 0}>
                     <div className="collapsible-card-content">{children}</div>
--- a/src/components/collapsible/Collapsible.js
+++ b/src/components/collapsible/Collapsible.js
@@ -37,6 +37,7 @@ type Props = {
 };
 
 type State = {
+    isHovered: boolean,
     isOpen: boolean,
 };
 
@@ -51,6 +52,7 @@ class Collapsible extends React.PureComponent<Props, State> {
     constructor(props: Props) {
         super(props);
         this.state = {
+            isHovered: false,
             isOpen: props.isOpen,
         };
     }
@@ -72,8 +74,16 @@ class Collapsible extends React.PureComponent<Props, State> {
         );
     };
 
+    handleMouseEnter = () => {
+        this.setState({ isHovered: true });
+    };
+
+    handleMouseLeave = () => {
+        this.setState({ isHovered: false });
+    };
+
     render() {
-        const { isOpen }: State = this.state;
+        const { isHovered, isOpen }: State = this.state;
         const {
             animationDuration,
             buttonProps = {},
@@ -107,7 +117,11 @@ class Collapsible extends React.PureComponent<Props, State> {
 
         return (
             <div className={sectionClassName}>
-                <div className={buttonClassName}>
+                <div
+                    className={buttonClassName}
+                    onMouseEnter={this.handleMouseEnter}
+                    onMouseLeave={this.handleMouseLeave}
+                >
                     <PlainButton
                         {...modifiedButtonProps}
                         className="collapsible-card-title"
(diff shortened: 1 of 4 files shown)


boxcla:

src/components/collapsible/Collapsible.js
```diff
@@ -72,8 +74,16 @@ class Collapsible extends React.PureComponent<Props, State> {
         );
     };
 
+    handleMouseEnter = () => {
```
priyajeet: I think you should try to do this with just CSS, if possible. You can always render the button, with visibility:hidden when its collapsed. And when open, visible. Likewise when hovering.

priyajeet: You may also want to consider the focus use case when it is collapsed, and the user tabs into the header with keyboard navigation

priyajeet:
BTW, PR title should ideally indicate what behavior is being changed, not the how. So a better title description would be something like "display action button when collapsed and hovered"

src/components/collapsible/Collapsible.js
indiramuru: In this component, Button is passed via headerActionItems pro and isOpen prop controls the visibility of this button. I don't think with just CSS the hover effect can be controlled from where ever its consumed.

priyajeet:
```jsx
{(isOpen && headerActionItems) || (!isOpen && isHovered && headerActionItems)}
```

can be changed to

```jsx
<span className="this_class_can_be_manupilated">{headerActionItems}</span>
```

indiramuru: `{!!headerActionItems && <span className="this_class_can_be_manupilated">{headerActionItems}</span>}` this should help to accommodate the UX requirement with just the CSS. With this change the Button gets displayed all the time (collapsed or not). Shield is using this component I guess. Once this is out they may have to put some CSS to maintain the existing behavior. I am trying to test this out. will be updating the PR soon

src/components/collapsible/Collapsible.js
```diff
@@ -107,7 +113,13 @@ class Collapsible extends React.PureComponent<Props, State> {
 
         return (
             <div className={sectionClassName}>
-                <div className={buttonClassName}>
+                <div
+                    className={buttonClassName}
+                    onBlur={onBlur}
+                    onFocus={onFocus}
+                    // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
+                    tabIndex="0"
```
priyajeet: Does the div need tabindex? Shouldn't it auto-tab into the PlainButton?

src/components/collapsible/Collapsible.js
```diff
@@ -117,7 +129,9 @@ class Collapsible extends React.PureComponent<Props, State> {
                         {title}
                         <IconCaretDown className="collapsible-card-header-caret" color={bdlGray50} width={8} />
                     </PlainButton>
-                    {isOpen && headerActionItems}
+                    {!!headerActionItems && (
+                        <span className="collapsible-card-header-actionItems">{headerActionItems}</span>
```
priyajeet: I don't see CSS changes done in the PR? Wouldnt you want something like

```css
.collapsible-card {
   .collapsible-card-title + .collapsible-card-header-actionItems {
       display: none;
    }

  .collapsible-card-title:focus + .collapsible-card-header-actionItems,
   .collapsible-card-title:hover + .collapsible-card-header-actionItems {
       display: block;
    }

   &.is-open {
     .collapsible-card-title + .collapsible-card-header-actionItems {
         display: block;
      }
  }
}
```

priyajeet: Also for new classes, i would switch to SUIT style

indiramuru: I was under the assumption that these CSS changes can be done where ever its used. Can include it.

src/components/collapsible/Collapsible.js
indiramuru: you are right. its not required

src/components/collapsible/Collapsible.js
indiramuru: Are you referring to the class `collapsible-card-header-actionItems`?

src/components/collapsible/Collapsible.js
```diff
@@ -107,7 +113,7 @@ class Collapsible extends React.PureComponent<Props, State> {
 
         return (
             <div className={sectionClassName}>
-                <div className={buttonClassName}>
+                <div className={buttonClassName} onBlur={onBlur} onFocus={onFocus}>
```
priyajeet: BTW, whats the use case of onBlur and onFocus? Just need them for something in parent component? Also should they be on the button instead, since thats the focus target.

src/components/collapsible/Collapsible.js
```diff
@@ -28,8 +28,12 @@ type Props = {
     isBordered?: boolean,
     /** Initial state of the collapsible card */
     isOpen: boolean,
+    /** callback called when collapsible is unfocused */
+    onBlur?: Function,
```
priyajeet: If these are needed, for new types, I would add full typings like () => {...} instead of just Function

src/components/collapsible/Collapsible.js
indiramuru: after the CSS adjustments, I don't see a use case for these functions at least for now. have removed it

src/components/collapsible/Collapsible.js
indiramuru: have removed onFocus and onBlur

src/components/collapsible/Collapsible.js
```diff
@@ -117,7 +117,7 @@ class Collapsible extends React.PureComponent<Props, State> {
                         {title}
                         <IconCaretDown className="collapsible-card-header-caret" color={bdlGray50} width={8} />
                     </PlainButton>
-                    {isOpen && headerActionItems}
+                    {!!headerActionItems && <span className="bdl-collapsible-actionItems">{headerActionItems}</span>}
```
priyajeet: bdl-Collapsible-
uppercase, should match component name

src/components/collapsible/Collapsible.scss
```diff
@@ -38,14 +38,40 @@
...
     }
 
-    &.is-open .collapsible-card-header-caret {
-        transform: rotateZ(180deg);
+    &.is-open {
+        .collapsible-card-header-caret {
+            transform: rotateZ(180deg);
+        }
+
+        .collapsible-card-title {
```
priyajeet: should this be prefixed with .btn-plain like the above rule?

src/components/collapsible/Collapsible.scss
```diff
@@ -38,14 +38,40 @@
...
+            &:focus,
+            &:hover {
+                text-decoration: none;
+            }
```
priyajeet: btw you can make a css mixin with this, and use it both places instead of duplicating

indiramuru: The almost similar css block above has  `+ .bdl-collapsible-actionItems { display: block }` inside the focus and hover. So they are not entirely  same to make mixin.

PR #1851, opened 2020-01-15
feat(targeting): add types used for targeting

azuruce:
most of the types were copied from EUA's in-app-messenger directory. Another PR will remove EUA's same type and import those here.

### DIFF
--- a/src/features/targeting/constants.js
+++ b/src/features/targeting/constants.js
@@ -1,0 +1,9 @@
+// @flow
+const ADHOC: 'adhoc' = 'adhoc';
+const BANNER: 'banner' = 'banner';
+const MODAL: 'modal' = 'modal';
+const POPOUT: 'popout' = 'popout';
+
+const MESSAGE_TYPES = { ADHOC, BANNER, MODAL, POPOUT };
+
+export { ADHOC, BANNER, MODAL, POPOUT, MESSAGE_TYPES };
--- a/src/features/targeting/contexts.js
+++ b/src/features/targeting/contexts.js
@@ -5,6 +5,15 @@ import { type FtuxContextValue, type TargetingContextValue } from './types';
 
 const FtuxContext = createContext<FtuxContextValue>({});
 
-const TargetingContext = createContext<TargetingContextValue>({});
+const defaultTargetingContextValue = {
+    getTargetingApi: () => ({
+        shouldShow: () => false,
+        onSeen: () => {},
+        onClosed: () => {},
+    }),
+    setEligibleMessageHeaders: () => {},
+};
+
+const TargetingContext = createContext<TargetingContextValue>(defaultTargetingContextValue);
 
 export { FtuxContext, TargetingContext };
--- a/src/features/targeting/index.js
+++ b/src/features/targeting/index.js
@@ -1,4 +1,6 @@
 // @flow
-export * from './contexts';
-
+export * from './constants';
 export * from './types';
+export * from './message-api';
+export * from './targeting-api';
+export * from './contexts';
--- a/src/features/targeting/targeting-api.js
+++ b/src/features/targeting/targeting-api.js
@@ -1,0 +1,23 @@
+// @flow
+import type { MessageHeader } from './types';
+
+/**
+ * TargetingApi is provided by TargetingContext. It is provided on a per message basis.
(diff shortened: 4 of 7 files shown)


boxcla:

src/features/targeting/targeting-api.js
```diff
@@ -0,0 +1,24 @@
...
+import type { MessageHeader } from './types';
+
+/**
+ * TargetingApi is provided by TargetingContext. It is provided on a per message basis.
+ * const { getTargetingApi } = useContext(TargetingContext);
+ * const targetingApi = getTargetingApi(messageName)
+ * will return the TargetingApi for the given message. See README.
+ */
+export type TargetingApi = {|
+    // return a memorized state bit for whether the targeting message should show
```
DanDeMicco: comments look out of place

src/features/targeting/contexts.js
```diff
@@ -5,6 +5,15 @@ import { type FtuxContextValue, type TargetingContextValue } from './types';
 
 const FtuxContext = createContext<FtuxContextValue>({});
 
-const TargetingContext = createContext<TargetingContextValue>({});
+const defaultTargetingContextValue = {
```
DanDeMicco: is this temporary? Maybe I am missing something, but how do these values get passed in from parent app? Would it make sense to expose a function which creates a context given the params?

src/features/targeting/constants.js
```diff
@@ -0,0 +1,9 @@
+// @flow
+const adhoc: 'adhoc' = 'adhoc';
+const banner: 'banner' = 'banner';
+const modal: 'modal' = 'modal';
+const popout: 'popout' = 'popout';
+
```
priyajeet: empty line needed?

src/features/targeting/index.js
```diff
@@ -1,4 +1,10 @@
 // @flow
-export * from './contexts';
+export * from './constants';
 
 export * from './types';
+
```
priyajeet: Likewise, empty lines necessary?

src/features/targeting/types.js
```diff
@@ -1,4 +1,37 @@
...
+
+type MessageHeaderIdentifiers = {|
+    id: MessageID, // an integer used to load message in getMessageRepresentation
+    name: MessageName, // used to form messageIdentifier where namespace is 'contextual'
+|};
+
+type MessageHeaderParams = {|
+    canBeShownMultipleTimes: boolean, // whether it can be shown again
+    createdDate: number,
+    isCustomMessage: boolean, // if it is custom message, custom message maker must exist to show
```
priyajeet: Can the comments be moved to the top of each line? Easier to read and makes it consistent with the above file + allows multi lines.

src/features/targeting/types.js
```diff
@@ -1,4 +1,37 @@
...
+    isExemptFromSuppression: boolean, // is this message exempt from suppression
+    isMessageTypeSuppressed: boolean, // is this message type suppressed for the user
```
priyajeet: Few of these comments in here seem unnecessary, feels like they are just paraphrased from the attribute name.

src/features/targeting/types.js
```diff
@@ -1,4 +1,37 @@
...
+    priority: number,
+    shouldAutoShow: boolean,
+    type: MessageType,
+|};
+
+export type MessageHeader = {|
+    ...MessageHeaderIdentifiers,
```
priyajeet: If these two individually are just local and aren't being used on their own anywhere else, does it makes sense to just have all the attributes listed here directly instead of adding extra types?

src/features/targeting/constants.js
azuruce: in this case, I think empty line serve as a separation from individual constants to the object. It might be acceptable to put empty line there. I looked through existing code in box-ui-elements and found they use capital case for constants naming and I will change to use capital case.
Thanks for review.

src/features/targeting/index.js
azuruce: remove empty lines as there is no real need for separation

src/features/targeting/types.js
azuruce: Definitely agree. My excuse for making this more than necessary is historically, we use MessageHeaderIdentifiers in another object but it is no longer a thing.
Thanks for review.

PR #1857, opened 2020-01-16
fix(typescript): migrate to explicit exports

priyajeet:
`export * from foo` currently seems to cause a problem with jest/rewire and breaks tests in projects using rewire.


boxcla:

priyajeet:
We could also potentially not choose to export other things like Props if there isn't much need of it being consumed cross-component. Like in this PR, only SVGProps and ButtonProps are imported cross-component, while other *Props aren't so they don't necessarily need to be in the index files.

PR #1858, opened 2020-01-17
fix(renderers): Fix column header not centered vertically

mxiao6:


boxcla:

mxiao6:
> Does this work as expected in all supported browsers?

Yeah, even in IE11

PR #1863, opened 2020-01-17
fix(form): make description more customizable for components

omgrief:
DatePicker
- make description a \<div\> instead of a \<i\> and have it respect accessibility
- add error accessibility

TextArea
- make description a \<div\> instead of a \<i\>

TextInput
- make description a \<div\> instead of a \<i\>

### DIFF
--- a/src/components/date-picker/DatePicker.scss
+++ b/src/components/date-picker/DatePicker.scss
@@ -19,7 +19,6 @@
     }
 
     .date-picker-description {
-        display: block;
         color: $bdl-gray-62;
     }
 
--- a/src/components/text-area/TextArea.scss
+++ b/src/components/text-area/TextArea.scss
@@ -13,7 +13,6 @@
     }
 
     .text-area-description {
-        display: block;
         color: $bdl-gray-62;
         word-wrap: break-word;
     }
--- a/src/components/text-input/TextInput.scss
+++ b/src/components/text-input/TextInput.scss
@@ -24,7 +24,6 @@
     }
 
     .text-input-description {
-        display: block;
         color: $bdl-gray-62;
         word-wrap: break-word;
     }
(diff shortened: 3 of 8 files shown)


boxcla:
Verified that @omgrief has signed the CLA. Thanks for the pull request!

src/components/text-input/TextInput.scss
```diff
@@ -26,6 +26,7 @@
     .text-input-description {
         display: block;
```
priyajeet: block may be unnecessary since its a div now

priyajeet:
Can you also add a proper description for the PR

src/components/text-area/TextArea.scss
```diff
@@ -15,6 +15,7 @@
     .text-area-description {
         display: block;
```
priyajeet: likewise

src/components/date-picker/DatePicker.scss
```diff
@@ -21,6 +21,7 @@
     .date-picker-description {
         display: block;
```
priyajeet: likewise

src/components/date-picker/DatePicker.scss
```diff
@@ -19,8 +19,8 @@
     }
 
     .date-picker-description {
-        display: block;
         color: $bdl-gray-62;
+        font-style: italic;
```
ehoogerbeets: Does this have to be italic? Italic doesn't work so well with Asian characters and fonts....

omgrief: it was originally a \<i\> tag, so this was just to keep default behavior.  This can be modified by overriding the styling during usage

priyajeet: @ehoogerbeets checking with design

omgrief: per offline conversations, design input is to remove italics

PR #1866, opened 2020-01-17
fix(typescript): remove prop exports

priyajeet:
Since they get compiled out and don't exist in the final output. It was causing unnecessary warnings when compiling in consuming repos.

### DIFF
--- a/src/components/primary-button/PrimaryButton.tsx
+++ b/src/components/primary-button/PrimaryButton.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import Button, { ButtonProps } from '../button';
+import Button, { ButtonProps } from '../button/Button';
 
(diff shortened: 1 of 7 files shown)


boxcla:

src/components/primary-button/PrimaryButton.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import Button, { ButtonProps } from '../button';
+import Button, { ButtonProps } from '../button/Button';
```
jstoffan: I thought we were moving toward using an implicit import pattern via `index.js` rather than the component module. Is that no longer possible due to Typescript support?

priyajeet: Its possible like it was working yesterday, but requires `export * from foo` style in index.js, which we can't have due to rewire in parent repos

PR #1868, opened 2020-01-21
chore(typescript): migrate generated icons

priyajeet:


boxcla:

priyajeet:
> For each asset, there will be a .flow.js and a .tsx variant now. Will this cause issues in apps that import the existing flow-based assets?

@booc0mtaco apps always import downcompiled es5, in which case there is no typescript or flow. the flow.js file is only there temporarily to prevent VScode and flow from complaining about other JS files within BUIE trying to import TSX files.

PR #1870, opened 2020-01-22
feat(nudge): add in nudge component

laughnan:
nudge-stories-test-tsx-components-nudge-nudge-looks-visually-correct-when-using-story-components-nudge-regular-1-snap


boxcla:

src/components/nudge/Nudge.tsx
```diff
@@ -0,0 +1,54 @@
+import * as React from 'react';
+import classNames from 'classnames';
+
+import PrimaryButton from '../primary-button';
+import X16 from '../../icon/fill/X16';
+
+import './Nudge.scss';
+
+export interface NudgeProps {
```
laughnan: open thought: should we just do `children` props versus having specific ones for elements on the Nudge.

src/components/nudge/Nudge.tsx
```diff
@@ -0,0 +1,54 @@
...
+    header: React.ReactNode;
+    /** onClick callback for the nudge button */
+    onButtonClick?: () => void;
+}
+
+function Nudge({ buttonText, className, content, illustration, header, onButtonClick }: NudgeProps) {
+    const [isClosed, setIsClosed] = React.useState<boolean>(false);
+    const classes = classNames(['Nudge', className], { 'Nudge--isClosed': isClosed });
+
+    const closeButton = (
```
laughnan: open thought: we should probably pass down a function for the close action rather than just hiding 🤔

src/components/nudge/Nudge.scss
```diff
@@ -0,0 +1,45 @@
+@import '../../styles/variables';
+
+/**************************************
+ * Nudge
+ **************************************/
+
+.Nudge {
```
priyajeet: bdl-

src/components/nudge/Nudge.scss
```diff
@@ -0,0 +1,45 @@
+@import '../../styles/variables';
+
+/**************************************
```
priyajeet: Comment sort of unnecessary

src/components/nudge/Nudge.scss
```diff
@@ -0,0 +1,45 @@
...
+    right: 20px;
+    bottom: 40px;
+    z-index: 40;
+    width: 221px;
+    max-height: 450px;
+    padding: 25px;
+    background-color: $white;
+    border: solid 1px $bdl-gray-02;
+    border-radius: 30px;
+    box-shadow: 0 3px 6px 0 rgba(0, 0, 0, .1);
```
priyajeet: I think rgba can also take in $black

src/components/nudge/Nudge.scss
```diff
@@ -0,0 +1,45 @@
...
+
+    &--isClosed {
```
priyajeet: bdl-is-closed

src/components/nudge/Nudge.scss
```diff
@@ -0,0 +1,45 @@
...
+        display: none;
+    }
+
+    &-button button {
```
priyajeet: with SUIT style, you shouldnt need to do nesting. So all these can move out.

.bdl-Nudge-closeButton { ... }

src/components/nudge/Nudge.tsx
```diff
@@ -0,0 +1,54 @@
...
+    /** classname to add to the container element. */
+    className?: string;
+    /** DOM element for the content of the nudge. */
+    content: React.ReactNode;
+    /** DOM element for the illustration image */
+    illustration: React.ReactNode;
+    /** DOM element for the header of the nudge */
+    header: React.ReactNode;
+    /** onClick callback for the nudge button */
+    onButtonClick?: () => void;
```
priyajeet: (event: React.SyntheticEvent<HTMLButtonElement>) => void

src/components/nudge/Nudge.tsx
```diff
@@ -0,0 +1,54 @@
...
+            onClick={() => {
+                setIsClosed(true);
+            }}
+        >
+            <X16 height={18} width={18} />
+        </button>
+    );
+
+    return (
+        <div className={classes}>
```
priyajeet: Can/should any of these use `<section>` or `<article>`

src/components/nudge/Nudge.tsx
priyajeet: Kind of depends on how flexible or rigid you want the structure to be.

src/components/nudge/Nudge.scss
priyajeet: Search for "CSS Class naming" in confluence.

src/components/nudge/Nudge.scss
```diff
@@ -0,0 +1,45 @@
+@import '../../styles/variables';
+
+/**************************************
+ * Nudge
+ **************************************/
+
+.Nudge {
+    position: fixed;
```
priyajeet: Another thing to consider is if the positioning, widths and heights all should be responsibility of the consuming app instead of this component. Can a nudge be within a section of the page as opposed to fixed to the whole window?

src/components/nudge/Nudge.tsx
DanDeMicco: I dont think this should be optional

src/components/nudge/Nudge.tsx
DanDeMicco: agreed

src/components/nudge/Nudge.tsx
```diff
@@ -0,0 +1,54 @@
...
+            {closeButton}
+            <div className="Nudge-illustration">{illustration}</div>
+            <div className="Nudge-header">{header}</div>
```
priyajeet: Should this be an h1 (normally it is expected if you decide to change the parent div to an article)? Instead of expecting it to be passed down via the prop. Maybe prop should be simple text?

src/components/nudge/Nudge.tsx
laughnan: i chose `section` for the overall parent.

src/components/nudge/Nudge.scss
laughnan: i think based on the design language we're building here, we want it to be forced to the bottom right and fixed to the page.

src/components/nudge/Nudge.tsx
```diff
@@ -15,39 +15,44 @@ export interface NudgeProps {
     content: React.ReactNode;
     /** DOM element for the illustration image */
     illustration: React.ReactNode;
+    /** Boolean value from the parent to determine if the nudge is shown */
+    isNudgeShown: boolean;
```
DanDeMicco: [nit] can be `isShown` to be consistent with other components

src/components/nudge/Nudge.tsx
```diff
@@ -15,39 +15,44 @@ export interface NudgeProps {
...
     const closeButton = (
-        // eslint-disable-next-line react/button-has-type
-        <button
-            aria-label="close-nudge"
-            className="Nudge-closeButton"
-            onClick={() => {
-                setIsClosed(true);
-            }}
-        >
+        <button aria-label="close-nudge" type="button" className="bdl-Nudge-closeButton" onClick={onNudgeCloseClick}>
```
DanDeMicco: use `Button`

src/components/nudge/Nudge.tsx
```diff
@@ -0,0 +1,59 @@
...
+            <X16 height={18} width={18} />
+        </button>
+    );
+
+    return (
+        <section className={classes}>
+            {closeButton}
+            <div className="bdl-Nudge-illustration">{illustration}</div>
+            <h1 className="bdl-Nudge-header">{header}</h1>
```
DanDeMicco: I'm not an accessibility expert, but not sure if h1 is appropriate

laughnan: i can't see any docs that tell me otherwise (i see it a lot of examples and i don't see explicit messaging to say "NO!")

PR #1872, opened 2020-01-22
chore: typescript/storybook cleanup/upgrades

priyajeet:


boxcla:

src/styles/colors.stories.tsx
```diff
@@ -1,35 +1,35 @@
-// @flow
 import * as React from 'react';
-import * as colors from './variables';
+import * as variables from './variables';
 import mdNotes from './colors.md';
 
-const bdlColors = {};
+const bdlColors: Record<string, Array<any>> = {}; // eslint-disable-line @typescript-eslint/no-explicit-any
```
jstoffan: Should we consider disabling the `@typescript-eslint/no-explicit-any` rule for now? I've already run into some trouble with it in box-content-preview, as well.

src/styles/colors.stories.tsx
```diff
@@ -1,35 +1,35 @@
...
+const colors: Record<string, string | Array<string>> = variables;
```
jstoffan: I would think TS could infer the type based on the object literal imported from variables.js? Maybe you just need to add an index type?

PR #1875, opened 2020-01-23
chore(node): Upgrade node dependency to 12.x

j3tan:
This commit also removes a duplicate node-sass version from yarn.lock

If you run encounter an  issue with `yarn install` such as:
```
node-pre-gyp ERR! command \"/Users/jtan/.nvm/versions/node/v12.14.1/bin/node\" \"/Users/jtan/box/box-ui-elements/node_modules/sass-variable-parser/node_modules/fsevents/node_modules/node-pre-gyp/bin/node-pre-gyp\" \"install\" \"--fallback-to-build\"
```

Verify you have `node@12` installed via `nvm` and remove `node_modules/sass-variable-parser`

```
$ node -v
v12.14.1
$ rm -fr node_modules/sass-variable-parser
$ yarn install
```


boxcla:

j3tan:
Also a question on whether we need to upgrade `node` engine in `package.json` to `>=12.x` though it should work fine still with both `8.x` and `10.x`

yarn.lock
```diff
@@ -18618,31 +18564,6 @@ node-sass@^4.13.1:
     stdout-stream "^1.4.0"
     "true-case-path" "^1.0.2"
 
-node-sass@^4.9.3:
```
j3tan: this is the duplicate `yarn.lock` that causes issues if you try to `yarn install` on `node@12.x`

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
```
jstoffan: Looks like this file can be deleted?

package.json
```diff
@@ -140,7 +140,6 @@
         "@box/languages": "^1.0.0",
         "@commitlint/cli": "^8.3.5",
         "@commitlint/config-conventional": "^8.3.4",
-        "@commitlint/travis-cli": "^8.3.5",
```
priyajeet: line 120

priyajeet: should it be a breaking change?

j3tan: it's not actually required to use node 12, but it is the recommended. Node 8 should still work but I'm going to update it to use `>=10.x`

j3tan: actually let me update that to a breaking change since engine does break people installing box-ui-elements

yarn.lock
```diff
@@ -12837,21 +12791,13 @@ fs.realpath@^1.0.0:
   resolved "https://registry.yarnpkg.com/fs.realpath/-/fs.realpath-1.0.0.tgz#1504ad2523158caa40db4a2787cb01411994ea4f"
   integrity sha1-FQStJSMVjKpA20onh8sBQRmU6k8=
 
-fsevents@^1.2.2:
```
j3tan: also had Node 12 issue

PR #1876, opened 2020-01-23
feat(content-uploader): Expose contentUploaderProps in Explorer & Picker

randyjohnston:
Expose contentUploaderProps as a top-level option for the Content Explorer so that chunked uploads can be turned off for platforms such as Salesforce which do not support JavaScript workers for chunked uploads.


boxcla:
Hi @randyjohnston, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -213,6 +214,7 @@ class ContentExplorer extends Component<Props, State> {
         initialPageSize: DEFAULT_PAGE_SIZE,
         contentPreviewProps: {
             contentSidebarProps: {},
+        contentUploaderProps: {},
```
priyajeet: This should not be nested in, should be at the same level as contentPreviewProps

src/elements/common/upload-dialog/UploadDialog.js
```diff
@@ -56,9 +58,11 @@ const UploadDialog = ({
         portalClassName={`${CLASS_MODAL} be-modal-upload`}
     >
         <ContentUploader
+            {...contentUploaderProps}
             apiHost={apiHost}
             onClose={onClose}
             onComplete={onUpload}
+            contentUploaderProps={contentUploaderProps}
```
priyajeet: This line seems unnecessary. ContentUploader doesn't have a prop name contentUploaderProps

priyajeet:
https://github.com/box/box-ui-elements/blob/master/CONTRIBUTING.md#step-4-push-your-feature-branch-to-your-fork has an example of PR title/message. You will have to additionally amend your commit too and force push it again since the commit is also not commitlint style.

src/elements/content-explorer/ContentExplorer.js
randyjohnston: Corrected

src/elements/common/upload-dialog/UploadDialog.js
randyjohnston: Removed

priyajeet:
@randyjohnston you can do the same/similar change in ContentPicker, like the one you did in ContentExplorer

randyjohnston:
> @randyjohnston you can do the same/similar change in ContentPicker, like the one you did in ContentExplorer

Equivalent changes made to ContentPicker

randyjohnston:
I re-signed the new CLA. Please let me know if any additional changes are required.

S-Harris:
Hello. We are really looking forward to using the functionality in this PR. Can we get this merged, or is the CLA still pending? 

@randyjohnston mentioned he did re-sign it.

dotnetcanuck:
My team is also waiting for the functionality from this PR, and we'd rather not fork Randy's branch if we can avoid it.

@priyajeet Is the code review completed on this PR? It says above that there is still a review required.

@randyjohnston I think the CLA is still pending; the message above says you need to associate the email address from your commit(s) with your GitHub account. And thanks for adding this functionality!!

sahiga:
@randyjohnston Looks good. Can you update the tests for UploadDialog, ContentExplorer, and ContentPicker to include the new props?

boxcla:
Hi, I have verified manually that `randyjohnston` has signed the updated CLA on Feb 27, 2020 4:11:38 PM. CLAassistant seems to be lagging behind so once this commit is approved, we will merge it manually.

sahiga:
@randyjohnston Let's merge this PR. I'll take care of the tests in a follow-up PR.

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -1655,6 +1658,7 @@ class ContentExplorer extends Component<Props, State> {
                             parentElement={this.rootElement}
                             appElement={this.appElement}
                             onUpload={onUpload}
+                            contentUploaderProps={contentUploaderProps}
```
jstoffan: This seems a little odd, since a lot of the sibling props here are constituents of `contentUploaderProps`. What are the actual props we need to pass in to support enabling/disabling of chunked uploads per the customer's needs?

jstoffan:
Should there be documentation updates to account for this new top-level prop?

src/elements/content-explorer/ContentExplorer.js
randyjohnston: This seems to follow the approach with `contentPreviewProps` and the hierarchy of props following that convention. For this specific customer case, we only need to expose `chunked` option, but other ContentUploader props (e.g. `fileLimit`) make sense to be options of the Content Explorer, and I think they should be nested contentUploaderProps and not top-level Content Explorer options.

jstoffan: I see the argument to expose it as its own thing; my concern was passing them into UploadDialog rather than spreading them, but it looks like PreviewDialog works the same way.

PR #1879, opened 2020-01-23
fix(task-form): enable any/all checkbox for completion rule with groups

booc0mtaco:
- update tests to support checking each type separately, and a mix of both


boxcla:

PR #1882, opened 2020-01-23
feat(taskform): error handling for groups

booc0mtaco:
- error handling for assigning a task to a group exceeding 250 limit
- address PR/CR comments after handoff

Co-authored-by: Andrew Holloway <booc0mtaco@users.noreply.github.com>
Co-authored-by: Alex Krolick <alexkrolick@users.noreply.github.com>


boxcla:

src/api/__tests__/Groups.test.js
```diff
@@ -0,0 +1,60 @@
...
+                groups.getGroupCount({
+                    file,
+                    group,
+                    successCallback,
+                    errorCallback,
+                });
+
+                // getGroupCount() returns Promise
+                // Verify whether groups.get resolves or rejects
+                // cannot use toBeCalledWith()
```
il10: Are we still using `toBeCalledWith` here?

booc0mtaco: yah, we can use with `expect.any` so that we verify the right type is being emitted. i'll clean up these comments to reflect what the test does

src/api/Feed.js
```diff
@@ -615,18 +646,55 @@ class Feed extends Base {
...
-            successCallback: (taskData: Task) => {
-                this.addPendingItem(this.file.id, currentUser, pendingTask);
-                this.createTaskNewSuccessCallback(file, uuid, taskData, assignees, successCallback, errorCallback);
-            },
-            errorCallback: (e: ElementsXhrError, code: string) => {
-                this.feedErrorCallback(false, e, code);
-            },
-        });
+        // create request for the size of each group by ID
+        const groupInfoPromises: Array<Promise<any>> = assignees
```
prakash-kumar: this new block of code here looks almost the same as the one above added to `updateTaskNew`.. is this intentional? are there opportunities to reuse?

src/constants.js
```diff
@@ -399,6 +400,9 @@ export const TASK_COMPLETION_RULE_ANY: 'ANY_ASSIGNEE' = 'ANY_ASSIGNEE';
 export const TASK_EDIT_MODE_CREATE: 'CREATE' = 'CREATE';
 export const TASK_EDIT_MODE_EDIT: 'EDIT' = 'EDIT';
 
+/* ----------------- Task Validation ---------------- */
+export const TASK_MAX_GROUP_ASSIGNEES: 250 = 250;
```
prakash-kumar: oooohhhh.. what is this? what does `:250` do here?

src/api/Feed.js
booc0mtaco: in the other block, i added a TODO for reconciling these two: this one uses promise pattern, while the other uses asyn/await. I didn't want to confuse this review with also converting one of them to the other format, since that would likely change additional code in non-functional ways.

src/constants.js
booc0mtaco: using `: number` as type means any numeric value is valid as long as `typeof` is number. Making the type a literal value means two useful things: the constant is locked to that exact value, and flow (or typescript i guess?) will show the value in the IDE or complain more specifically if it mutates.

src/api/Feed.js
prakash-kumar: 👍

src/constants.js
prakash-kumar: 👍

src/elements/content-sidebar/activity-feed/task-form/TaskError.js
```diff
@@ -9,23 +9,23 @@ import getProp from 'lodash/get';
...
     editMode?: TaskEditMode,
-    error?: { status: number },
+    error?: { status: number }, // TODO: update to ElementXhrError once API supports it
     taskType: TaskType,
 };
-
 const TaskError = ({ editMode, error, taskType }: Props) => {
     const isEditMode = editMode === TASK_EDIT_MODE_EDIT;
     const isForbiddenErrorOnEdit = getProp(error, 'status') === 403 && isEditMode;
+    const taskGroupExceedsError = getProp(error, 'code') === 'group_exceeds_limit';
```
swfree: should you use the constant `ERROR_CODE_GROUP_EXCEEDS_LIMIT` here?

booc0mtaco: good catch. will update

PR #1893, opened 2020-01-28
chore(typescript): Update bdlClassnameManager to handle TypeScript files

sahiga:
This PR enables bdlClassnameManager to modify class names within TypeScript files (specifically, TSX files).

### DIFF
--- a/scripts/utils/bdlClassnameManager
+++ b/scripts/utils/bdlClassnameManager
@@ -218,7 +218,7 @@ async function main() {
                 .join('|'),
             'g',
         );
-    } else if (jsExtensions.has(fileType) || fileType === '') {
+    } else if (jsExtensions.has(fileType)) {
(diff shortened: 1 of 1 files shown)


boxcla:

scripts/utils/bdlClassnameManager
```diff
@@ -211,7 +218,7 @@ async function main() {
                 .join('|'),
             'g',
         );
-    } else if (fileType === 'js') {
+    } else if (jsExtensions.has(fileType) || fileType === '') {
```
manishPh: Why `|| fileType === ''` ?

sahiga: Good catch. I was originally adding `|| fileType === 'tsx'` to all the `if` statements, so I must have started to do that here and then forgotten to remove the extra clause.

PR #1896, opened 2020-01-29
fix(preview): Annotations header misalignment in IE11

ConradJChan:


boxcla:

src/elements/content-preview/preview-header/PreviewHeader.scss
```diff
@@ -45,6 +45,7 @@
     .ba-point-post-cancel-container,
     .ba-draw-post-cancel-container {
         position: absolute;
+        top: 6px;
```
mxiao6: Will this affect the other browsers? Why the misalignment only happened in IE11, not the other browsers?

ConradJChan:
@jstoffan - yeah, it looks good in the other browsers

src/elements/content-preview/preview-header/PreviewHeader.scss
ConradJChan: Not sure, seems like the other browsers just handle it more gracefully

PR #1898, opened 2020-01-30
fix(usm): Check view permissions on classification in share modal

patrickpaul:
The current item classification is shown in the title of the UnifiedShareModal. There are some circumstances where the user does not have permission to see this classification and that is governed by the `canUserSeeClassification` property on the item record. 

This permission was not currently being checked.


boxcla:

PR #1900, opened 2020-01-31
fix(datepicker): Fix wrong first day of the week

mxiao6:
A temporary solution to https://github.com/box/box-ui-elements/pull/1572

### DIFF
--- a/src/components/date-picker/DatePicker.js
+++ b/src/components/date-picker/DatePicker.js
@@ -16,6 +16,7 @@ import PlainButton from '../plain-button';
 import Tooltip from '../tooltip';
 
 import { convertDateToUnixMidnightTime } from '../../utils/datetime';
+import { get as getCookie } from '../../api/box-edit/cookies';
 
 import './DatePicker.scss';
 
@@ -46,6 +47,8 @@ const messages = defineMessages({
         id: 'boxui.datePicker.chooseDate',
     },
 });
+
+const localesWhereWeekStartsOnSunday = ['en-US', 'en-CA', 'jp-JP'];
 
 const TOGGLE_DELAY_MS = 300;
 const ISO_STRING_DATE_FORMAT = 'isoString';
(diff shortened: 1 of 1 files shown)


boxcla:

src/components/date-picker/DatePicker.js
```diff
@@ -195,6 +198,7 @@ class DatePicker extends React.Component<Props> {
             setDefaultDate: true,
             defaultDate: defaultValue,
             field: this.dateInputEl,
+            firstDay: localesWhereWeekStartsOnSunday.includes(getCookie('lang') || 'en-US') ? 0 : 1,
```
sahiga: Can you add a unit test for this change?

mxiao6: > Can you add a unit test for this change?

The DatePicker test is changed to `DatePicker.test.js.fixme`. So I want to hear @alexkrolick 's feedback first (author of #1572), then fix the tests.

src/components/date-picker/DatePicker.js
```diff
@@ -16,6 +16,7 @@ import PlainButton from '../plain-button';
 import Tooltip from '../tooltip';
 
 import { convertDateToUnixMidnightTime } from '../../utils/datetime';
+import { get as getCookie } from '../../api/box-edit/cookies';
```
jstoffan: I don't think we want to use a Box Edit API helper in a common component. We probably should find another way to get the value rather than via a cookie.

ConradJChan: Would a better option be to get the `intl.locale`?

mxiao6: `intl.locale` only contain language info like `en` or `zh`. But here we need country info like `en-US` or `en-UK`. US first day is Sunday and CA is Monday.

mxiao6: https://git.dev.box.net/Box/EndUserApp/pull/4669

alexkrolick: Whatever we do here, this needs to work with public APIs (partner platforms) and safely fall back when the locale doesn't have the country code.

priyajeet: Also see https://github.com/tc39/ecma402/issues/6

priyajeet: cc @ehoogerbeets @jeeyi

mxiao6:
Will merge #1572 instead.

PR #1901, opened 2020-01-31
fix(checkbox): fix tooltips on checkbox labels that wrap

tjgupta:
Before:
Screen Shot 2020-01-30 at 10 58 42 AM

After:
Screen Shot 2020-01-30 at 10 51 59 AM


boxcla:

src/components/checkbox/Checkbox.js
```diff
@@ -35,48 +37,59 @@ type Props = {
     value?: any,
 };
 
-const Checkbox = ({
```
priyajeet: You can use useRef instead of converting to class. eg: TextArea.js

src/components/checkbox/Checkbox.scss
```diff
@@ -62,19 +64,20 @@
                 border-color: $bdl-gray-30;
             }
         }
+    }
 
-        + span + span {
-            margin-left: 10px;
-        }
+    .checkbox-label-tooltip-wrapper {
```
priyajeet: Should use bdl-SUIT style for new classes

tjgupta:
Updated @priyajeet .

PR #1902, opened 2020-01-31
fix(sidebar): Tasks and mentions should not have Uploaders as options

ConradJChan:


boxcla:

PR #1903, opened 2020-01-31
feat(theme): Updating theming logic per design

akahn87:


boxcla:

src/components/theme/contrast-colors.js
```diff
@@ -18,14 +18,14 @@ function getYiq(color: string) {
 
 // Lighten and Darken should be absolute
 // https://github.com/Qix-/color/issues/53
```
priyajeet: Should this folder be moved under `utils` or `styles`? Since its mostly JS.
Also I think you can convert this to .ts pretty fast.

akahn87: Potentially so.. after talking through some potential solutions around theming other components like Button, I don't think we'll need a custom theme provider anymore for providing "defaults" we can just do something like so:

`
// import styled from 'styled-components'
// import defaultTheme from 'components/theme';

const StyledButton = styled.button(({ theme = defaultTheme, ...props }) => ({
    backgroundColor: theme.primary.background,
    borderColor: theme.primary.border,
}))

const Button = (props) => (
    <StyledButton>
        {text}
    </StyledButton>
)

export default Button;`

src/utils/createTheme.js
```diff
@@ -0,0 +1,160 @@
...
+
+// The yiq coefficients in the color library are incorrect
+// http://poynton.ca/notes/colour_and_gamma/ColorFAQ.html#RTFToC9
+function getYiq(color: string) {
+    const rgb = Color(color).rgb().color;
+
+    return (rgb[0] * 2126 + rgb[1] * 7152 + rgb[2] * 722) / 10000;
+}
+
+function colorModifier(color: ColorType, amount: number) {
```
j3tan: `adjustLightness` might be more apt?

src/utils/createTheme.js
```diff
@@ -0,0 +1,160 @@
...
+                hover: -4,
+            },
+        },
+    };
+
+    // Filter down the color map to the object that's in the proper YIQ range
+    const filteredColorMap = Object.keys(colorMap).reduce((obj, key) => {
+        if (
+            (colorKeyYiq >= colorMap[key].yiqRange[0] && colorKeyYiq < colorMap[key].yiqRange[1]) ||
+            (colorMap.lightnessThreshold && colorKeyLightness >= colorMap.lightnessThreshold)
```
j3tan: `lightnessThreshold` not defined

src/utils/createTheme.js
```diff
@@ -0,0 +1,160 @@
...
+            modifiers: {
+                active: -8,
+                gradient: -5,
+                hover: -4,
+            },
+        },
+    };
+
+    // Filter down the color map to the object that's in the proper YIQ range
+    const filteredColorMap = Object.keys(colorMap).reduce((obj, key) => {
```
j3tan: maybe `find()` instead?

src/utils/createTheme.js
```diff
@@ -0,0 +1,160 @@
...
+
+    const activeContrast = modifiedColors.active.contrast(foregroundObj);
+    const active =
+        activeContrast >= MIN_CONTRAST + OFFSET_FACTOR && activeContrast !== 21
+            ? modifiedColors.active
+            : modifiedColors.activeInverse || colorModifier(colorKeyObj, -filteredColorMap.modifiers.active);
+
+    return {
+        primary: {
+            background: colorKey,
```
j3tan: maybe output `colorKeyObj.hex()` here?

j3tan: and for other colorKey references, so the file stays `hex` only instead of randomly having strings like `white`

src/utils/createTheme.js
```diff
@@ -0,0 +1,159 @@
...
+    const activeContrast = modifiedColors.active.contrast(foregroundObj);
+    const active =
+        activeContrast >= MIN_CONTRAST + OFFSET_FACTOR && activeContrast !== 21
+            ? modifiedColors.active
+            : modifiedColors.activeInverse || adjustLightness(colorKeyObj, -colorRangeConfig.modifiers.active);
+
+    // TODO: Fix case inconsitencies between scss variables and color output.
+    return {
+        primary: {
+            background: colorKeyObj.hex().toLowerCase(),
```
j3tan: nit: save to var since its used multiple times

src/components/theme/contrast-colors.js
akahn87: Moved to utils and styles hybrid

PR #1904, opened 2020-01-31
feat(assets): add asset importer script

booc0mtaco:
- Used when importing new icons, illustrations, and logos from design team


boxcla:

j3tan:
Add an `npm` script to run + some documentation in `DEVELOPING.md`

booc0mtaco:
@j3tan it's more the latter, so left instructions out in this PR. Will update with a "script" declaration in package.json to make the cron usage more brief, tho

package.json
```diff
@@ -55,6 +55,7 @@
         "build:prod:storybook": "LANGUAGE=en-US REACT=true BROWSERSLIST_ENV=production BABEL_ENV=development NODE_ENV=development build-storybook -c .storybook -o styleguide/storybook",
         "build:sync": "LANGUAGE=en-US BABEL_ENV=development NODE_ENV=development RSYNC=true webpack --config scripts/webpack.config.js --mode development",
         "build:variables": "cat `find src/styles/constants -name '*.scss'` > src/styles/variables.scss | node ./scripts/build-style-vars.js src/styles/variables.scss && rm src/styles/variables.scss",
+        "build:assets": "./scripts/asset-import/cli.js",
```
priyajeet: does this need `node` in front?
`node ./scripts/asset-import/cli.js`

package.json
```diff
@@ -282,13 +291,15 @@
         "tabbable": "^1.1.3",
         "ts-loader": "^6.2.1",
         "typescript": "^3.7.5",
+        "uniqid": "^5.2.0",
```
priyajeet: Can we reuse lodash/uniqueID? or is this a guid generator?

package.json
booc0mtaco: oh this one includes a hashbang specifying node at the top `#!/usr/bin/env node`. any preference?

package.json
booc0mtaco: I was hesitant to use lodash's, since it is numeric and counts upward per run. this `uniqid` package produces strings like 'a38gckeaa3fkqm2', but is only used for uniqueness for a temp directory. I can switch to using uuid

priyajeet: its fine

PR #1908, opened 2020-02-04
fix(virtualized-table): Fix upside down sort icon

mxiao6:
The original svg icon is descending icon. So we should rotate 180 degrees when the class is ascending.

Original svg:
image


boxcla:

PR #1909, opened 2020-02-04
feat(label-pill): add LabelPill component

booc0mtaco:
label-pill

### DIFF
--- a/src/components/label-pill/LabelPill.scss
+++ b/src/components/label-pill/LabelPill.scss
@@ -1,0 +1,51 @@
+@import '../../styles/variables';
+
+.bdl-LabelPill {
+    padding: 3px;
+    color: $bdl-gray;
+    font-size: 10px;
+    line-height: 10px;
+    background-color: $bdl-gray-10;
+    border: 1px solid;
+    border-color: $bdl-gray-20;
+    border-radius: 10px;
+
+    &.bdl-LabelPill--warning {
+        background-color: $bdl-yellorange-10;
+        border-color: $bdl-yellorange;
+    }
+
+    &.bdl-LabelPill--error {
+        background-color: $bdl-watermelon-red-10;
+        border-color: $bdl-watermelon-red;
+    }
+
+    &.bdl-LabelPill--success {
+        background-color: $bdl-green-light-10;
+        border-color: $bdl-green-light;
+    }
+
+    &.bdl-LabelPill--alert {
+        background-color: $bdl-watermelon-red-10;
+        border-color: $bdl-watermelon-red;
+    }
+
+    &.bdl-LabelPill--highlight {
+        background-color: $bdl-light-blue-10;
+        border-color: $bdl-light-blue;
+    }
+
+    &.bdl-LabelPill--info {
+        background-color: $bdl-light-blue-10;
+        border-color: $bdl-light-blue;
+    }
+}
+
+.bdl-LabelPill-iconContent {
+    position: relative;
+    top: .125em;
+}
+
+.bdl-LabelPill-textContent {
+    margin: 0 3px;
+}
--- a/src/components/label-pill/LabelPill.stories.md
+++ b/src/components/label-pill/LabelPill.stories.md
@@ -1,0 +1,3 @@
+`import LabelPill from 'box-ui-elements/es/components/label-pill';`
+
+_Note_: this component is meant to be used with one of the formatted strings produced by react-intl (`FormattedMessage`, `FormattedRelativeDate`, etc.) **or** one of our icon assets.
--- a/src/components/label-pill/LabelPill.tsx
+++ b/src/components/label-pill/LabelPill.tsx
@@ -1,0 +1,43 @@
+import * as React from 'react';
+import classNames from 'classnames';
+import LabelPillIcon from './LabelPillIcon';
+import LabelPillText from './LabelPillText';
+
+import './LabelPill.scss';
+
+export enum LabelPillStatus {
+    DEFAULT = 'default',
+    INFO = 'info',
+    HIGHLIGHT = 'highlight',
+    SUCCESS = 'success',
+    WARNING = 'warning',
+    ALERT = 'alert',
+    ERROR = 'error',
+}
+
+export interface LabelPillProps {
+    children: Array<React.ReactChild> | React.ReactChild;
+    type?: LabelPillStatus;
+}
+
+const LabelPillContainer = React.forwardRef((props: LabelPillProps, ref: React.Ref<HTMLSpanElement>) => {
+    const { children, type = LabelPillStatus.DEFAULT, ...rest } = props;
+    const labelPillClasses = classNames('bdl-LabelPill', {
+        [`bdl-LabelPill--${type}`]: type !== LabelPillStatus.DEFAULT,
+    });
+    return (
+        <span ref={ref} className={labelPillClasses} {...rest}>
+            <span className="bdl-LabelPill-content">{children}</span>
+        </span>
+    );
+});
+
+LabelPillContainer.displayName = 'LabelPill';
+
+const LabelPill = {
+    Pill: LabelPillContainer,
+    Text: LabelPillText,
+    Icon: LabelPillIcon,
+};
+
+export default LabelPill;
--- a/src/components/label-pill/LabelPillText.tsx
+++ b/src/components/label-pill/LabelPillText.tsx
@@ -1,0 +1,11 @@
+import * as React from 'react';
+
+export interface LabelPillTextProps {
+    children: React.ReactChild;
+}
+
+const LabelPillText = ({ children }: LabelPillTextProps) => (
+    <span className="bdl-LabelPill-textContent">{children}</span>
+);
+
+export default LabelPillText;
--- a/src/components/label-pill/index.ts
+++ b/src/components/label-pill/index.ts
@@ -1,0 +1,1 @@
+export { default, LabelPillStatus } from './LabelPill';
--- a/src/components/index.ts
+++ b/src/components/index.ts
@@ -13,6 +13,9 @@ export * from './button-group';
 export { default as Fieldset } from './fieldset';
 export * from './fieldset';
 
+export { default as LabelPill } from './label-pill';
+export * from './label-pill';
+
 export { default as Link } from './link';
 export * from './link';
 
--- a/src/components/label-pill/LabelPill.stories.tsx
+++ b/src/components/label-pill/LabelPill.stories.tsx
@@ -1,0 +1,59 @@
+import * as React from 'react';
+import { select } from '@storybook/addon-knobs';
+import ClockBadge16 from '../../icon/fill/ClockBadge16';
+import LabelPill, { LabelPillStatus } from './LabelPill';
+import notes from './LabelPill.stories.md';
+
+const typeLabel = 'Pill Types';
+
+const options: LabelPillStatus[] = [
+    LabelPillStatus.DEFAULT,
+    LabelPillStatus.ALERT,
+    LabelPillStatus.ERROR,
+    LabelPillStatus.HIGHLIGHT,
+    LabelPillStatus.INFO,
+    LabelPillStatus.SUCCESS,
+    LabelPillStatus.WARNING,
+];
+
+export const withText = () => (
+    <LabelPill.Pill type={select(typeLabel, options, LabelPillStatus.DEFAULT, 'g01')}>
+        <LabelPill.Text>TEST TEXT</LabelPill.Text>
+    </LabelPill.Pill>
+);
+
+export const withIcon = () => (
+    <LabelPill.Pill type={select(typeLabel, options, LabelPillStatus.DEFAULT, 'g01')}>
+        <LabelPill.Icon Component={ClockBadge16} />
+    </LabelPill.Pill>
+);
+
+export const withBoth = () => (
+    <LabelPill.Pill type={select(typeLabel, options, LabelPillStatus.DEFAULT, 'g01')}>
+        <LabelPill.Icon Component={ClockBadge16} />
+        <LabelPill.Text>TEST TEXT</LabelPill.Text>
+    </LabelPill.Pill>
+);
+
+export const severalComponents = () => (
+    <div style={{ textAlign: 'center' }}>
+        <LabelPill.Pill type={select(typeLabel, options, LabelPillStatus.DEFAULT, 'g01')}>
+            <LabelPill.Text>TEXT ONLY</LabelPill.Text>
+        </LabelPill.Pill>
+        <LabelPill.Pill type={select(typeLabel, options, LabelPillStatus.DEFAULT, 'g01')}>
+            <LabelPill.Icon Component={ClockBadge16} />
+        </LabelPill.Pill>
+        <LabelPill.Pill type={select(typeLabel, options, LabelPillStatus.DEFAULT, 'g01')}>
+            <LabelPill.Icon Component={ClockBadge16} />
+            <LabelPill.Text>ICON AND TEXT</LabelPill.Text>
+        </LabelPill.Pill>
+    </div>
+);
+
+export default {
+    title: 'Components|LabelPill',
+    component: LabelPill,
+    parameters: {
+        notes,
+    },
+};
--- a/src/components/label-pill/LabelPillIcon.tsx
+++ b/src/components/label-pill/LabelPillIcon.tsx
@@ -1,0 +1,12 @@
+import * as React from 'react';
+import { SVGProps } from '../../components/accessible-svg/AccessibleSVG';
+
+export interface LabelPillIconProps {
+    Component: React.FunctionComponent<SVGProps>;
+}
+
+const LabelPillIcon = ({ Component }: LabelPillIconProps) => (
+    <Component className="bdl-LabelPill-iconContent" width={10} height={10} />
+);
+
+export default LabelPillIcon;
--- a/src/components/label-pill/__tests__/LabelPill.test.tsx
+++ b/src/components/label-pill/__tests__/LabelPill.test.tsx
@@ -1,0 +1,82 @@
+import React from 'react';
+import { mount } from 'enzyme';
+
+import LabelPill, { LabelPillStatus } from '..';
+
+describe('components/label-pill/LabelPill', () => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const FormattedMessage = (msgObject: any) => <span>{msgObject.defaultMessage}</span>;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const MockIcon = (props: any) => (
+        <svg {...props}>
+            <g />
+        </svg>
+    );
+
+    test('should render default DOM with message', () => {
+        const fullComponent = mount(
+            <LabelPill.Pill>
+                <LabelPill.Text>
+                    <FormattedMessage
+                        {...{
+                            id: 'id.test',
+                            defaultMessage: 'test message',
+                            description: 'test message description',
+                        }}
+                    />
+                </LabelPill.Text>
+            </LabelPill.Pill>,
+        );
+
+        expect(fullComponent).toMatchSnapshot();
+    });
+
+    test('should render default DOM with icon component', () => {
+        const fullComponent = mount(
+            <LabelPill.Pill>
+                <LabelPill.Icon Component={MockIcon} />
+            </LabelPill.Pill>,
+        );
+
+        expect(fullComponent).toMatchSnapshot();
+    });
+
+    test('should include DOM for tooltip if specified', () => {
+        const fullComponent = mount(
+            <LabelPill.Pill>
+                <LabelPill.Text>
+                    <FormattedMessage
+                        {...{
+                            id: 'id.test',
+                            defaultMessage: 'test message',
+                            description: 'test message description',
+                        }}
+                    />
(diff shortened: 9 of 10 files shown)


boxcla:

booc0mtaco:
@alexkrolick yah, i originally used the new strings as examples in the stories for this component, but opted not to, since the component is more generic. 

I'll split them out as a separate commit. If we had more discrete releases, it'd be useful to know when they were added.

booc0mtaco:
split strings into #1912

src/components/label-pill/LabelPill.js
```diff
@@ -0,0 +1,43 @@
+// @flow
```
priyajeet: Should be using typescript hence forth

src/components/label-pill/LabelPill.js
```diff
@@ -0,0 +1,43 @@
...
+import classNames from 'classnames';
+import Tooltip from '../tooltip';
+
+import './LabelPill.scss';
+
+// TODO: Unify notification types `InlineNoticeType` from core
+type Props = {
+    children: React.Node,
+    tooltip?: string,
+    type?: 'info' | 'warning' | 'highlight' | 'error' | 'alert' | 'success',
```
priyajeet: should be typescript enum

src/components/label-pill/LabelPill.js
```diff
@@ -0,0 +1,43 @@
...
+                    contentType.indexOf('Formatted') === 0 ? 'bdl-LabelPill-has-text' : 'bdl-LabelPill-has-component',
+                )}
+            >
+                {children}
+            </div>
+        </div>
+    );
+
+    return tooltip ? (
+        <Tooltip position="top-center" text={tooltip}>
```
priyajeet: isDisabled prop for tooltip can help with with this ternary

src/components/label-pill/LabelPill.js
```diff
@@ -0,0 +1,43 @@
...
+    const contentType = React.Children.map(children, child => child.type.name)[0];
+    const labelPill = (
+        <div
+            className={classNames(classNames('bdl-LabelPill', type !== 'default' ? `bdl-LabelPill--${type}` : ''))}
+            {...rest}
+        >
+            <div
+                className={classNames(
+                    'bdl-LabelPill-content',
+                    contentType.indexOf('Formatted') === 0 ? 'bdl-LabelPill-has-text' : 'bdl-LabelPill-has-component',
```
priyajeet: I dont think this should be done. Someone can use intl.format too.

src/components/label-pill/LabelPill.scss
```diff
@@ -0,0 +1,54 @@
...
+        border-color: $bdl-light-blue;
+    }
+}
+
+.bdl-LabelPill-content {
+    font-size: 10px;
+    line-height: 10px;
+    vertical-align: middle;
+
+    &.bdl-LabelPill-has-text {
```
priyajeet: hasText
hasComponent

src/components/label-pill/LabelPill.js
booc0mtaco: any ideas on how to tell whether there's a text node or an icon in the -content?

booc0mtaco:
@priyajeet the difference is that all stylistic details are unique here, and its usage is also different.

Differences:
- font-weight
- text-color
- border
- border-radius
- background-color
- padding

Also, this component receives text or an icon and can adapt based on its children. 

I *really* want to have just one pill component, but but these don't actually share that much

src/components/label-pill/LabelPill.scss
priyajeet: has-text / has-component should be modifiers, since they are not state.
So, &.bdl-LabelPill--text and &.bdl-LabelPill--component

src/components/label-pill/LabelPill.js
priyajeet: You don't you make two different components to choose from

booc0mtaco: actually, i'll go with a less clever approach here to see how that looks

priyajeet: You could potentially move all this under the pill folder.
Make your component PillBase, and then export out two components - LabelPill, IconPill or something like that. And when we consolidate, LabelPill can replace the existing Pill.js that used inside pill selector

src/components/label-pill/LabelPill.scss
```diff
@@ -0,0 +1,54 @@
...
+        padding: 3px 6px;
+    }
+
+    &.bdl-LabelPill-has-component {
+        padding: 3px 4px;
```
alexkrolick: I wonder if you could remove the icon/text special cases if you used flex to align the items instead of padding? ie. an icon would collapse to a circle and the content would be centered horizontally and vertically.

src/components/label-pill/LabelPill.js.flow
```diff
@@ -0,0 +1,28 @@
+// @flow
```
priyajeet: You probably don't need this file, its a new component not used by anybody else

src/components/label-pill/LabelPill.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from 'react';
+import classNames from 'classnames';
+import Tooltip, { TooltipPosition } from '../tooltip';
+
+import './LabelPill.scss';
+
+export enum StatusLevelType {
```
priyajeet: I would probably prefix this with `LabelPill` so that we don't have ambiguous exports. Also is the word `type` needed? Can it just be statusLevel (even the prop name)? Or maybe its just `type` and no status level

src/components/label-pill/LabelPill.tsx
```diff
@@ -0,0 +1,36 @@
...
+    children: React.ReactNode;
+    tooltip?: string;
+    type?: StatusLevelType;
+}
+
+const LabelPill = ({ children, type = StatusLevelType.DEFAULT, tooltip = '', ...rest }: LabelPillProps) => {
+    return (
+        <Tooltip position={TooltipPosition.TOP_CENTER} text={tooltip} isDisabled={!tooltip}>
+            <span
+                className={classNames(classNames('bdl-LabelPill', type !== 'default' ? `bdl-LabelPill--${type}` : ''))}
```
priyajeet: 2 nested classNames? Should extract it out as a variable for readability and debuggability. Also default string should be changed to enum value.

src/components/label-pill/LabelPill.scss
```diff
@@ -1,13 +1,19 @@
...
+    padding: 3px;
     color: $bdl-gray;
+    font-size: 10px;
+    line-height: 10px;
     background-color: $bdl-gray-10;
     border: 1px solid;
     border-color: $bdl-gray-20;
     border-radius: 10px;
 
+    & + .bdl-LabelPill {
```
alexkrolick: I believe we rolled back something like this for Buttons because the increased specificity of the selector made it hard to manual adjust spacing

src/components/label-pill/LabelPill.tsx
alexkrolick: It can be this:

```js
  classNames('bdl-LabelPill', { [`bdl-LabelPill--${type}`]: type !== 'default' })
```

alexkrolick: Not sure if `type` is a reserved word in Typescript, I guess not

src/components/label-pill/LabelPill.scss
booc0mtaco: i'd be a little less worried about this one, only because it's for multiple instances of this component next to one another. except for this edge case, it doesn't do any external spacing .

Unless that button use case was also its positioning relative to other buttons...

alexkrolick: It was for 2+ buttons next to each other; it created a situation similar to the other issue with button's :hover/:focus states, where every time you want to change the default, you need to recreate the selector of make one of higher specificity.

alexkrolick: May also be an issue if you have several pills in a line that wraps, where the wrapped ones have an extra left margin. I'm not sure the ideal way to handle that though, it may end up occurring anyways if spacing was implemented by consumer.

src/components/label-pill/LabelPill.tsx
```diff
@@ -0,0 +1,37 @@
...
+export interface LabelPillProps {
+    children: React.ReactNode;
+    tooltip?: string;
+    type?: LabelPillStatus;
+}
+
+const LabelPill = ({ children, type = LabelPillStatus.DEFAULT, tooltip = '', ...rest }: LabelPillProps) => {
+    const labelPillClasses = classNames(
+        'bdl-LabelPill',
+        type !== LabelPillStatus.DEFAULT ? `bdl-LabelPill--${type}` : '',
```
alexkrolick:
```suggestion
{ [`bdl-LabelPill--${type}`]: type !== LabelPillStatus.DEFAULT },
```

FYI this is the "classnames" way. It will drop the `key: value` pairs where values are false.

alexkrolick: Should this accept className as a prop, too? And forwardRef so it can work with tooltips and stuff?

src/components/label-pill/LabelPill.tsx
```diff
@@ -0,0 +1,37 @@
...
+    );
+    return (
+        <Tooltip position={TooltipPosition.TOP_CENTER} text={tooltip} isDisabled={!tooltip}>
```
alexkrolick: I feel like this will eventually expand to having all the props to customize the tooltip... maybe better left to the consumers to put their own tooltips on it?

booc0mtaco: It really depends on the strength of the designs, IMO. if all of these had carefully considered layouts that aren't as flexible as `Tooltip`, I would prefer this method. If the positioning and such is just as flexible as the options provided to `Tooltip`, then they don't need to be combined. I don't know the answer, but this is a good point since things tend to be the latter. I'll decouple.

src/components/label-pill/LabelPill.tsx
booc0mtaco: It looks gross, but is more terse and avoids generating an empty string. ++

I avoided adding a classname since the pill size, shape, and colors should be stable (should force people to either try and override the class names of the component directly (BAD), or ask questions (good).)

I don't know about `forwardRef`. I don't foresee this being more than just a visual component, and any behaviors will be added thru composition. Then again, the ref stuff confuses me so I don't know

booc0mtaco: it does have `...rest` so I suppose a person could add `className="blah"` if they really wanted to

src/components/label-pill/LabelPill.scss
booc0mtaco: I hardened the "between spacing" and will have this up in the next push. using `:last-child` to cancel margin-right

src/components/label-pill/LabelPill.tsx
booc0mtaco: For reference, it's useful to convert an enum to it's individual values, for use in storybook knobs. However, this appears to be some unspecified functionality in typescript. You can search for it and see all the confusion around it. Kinda nasty that there's no clear construct around this use case.

Just in case you're wondering why I have that manual array in the .stories for this...

src/components/label-pill/LabelPill.tsx
alexkrolick: That would override the original classNames, removing the styles. Usually we pass through classname in order to add layout classes, mostly flex or margins. I agree we don't expect colors to be customized here through CSS overrides.

alexkrolick: I think we need forwardRef for all "leaf nodes" of the design system, so that you can get a ref to a DOM node to attach listeners and work with wrappers like Popper/the new version of Tether.

booc0mtaco: We don't seem to be doing this now, as most of the existing components with any reference to fowardRef are some form of HOC. All of the other components with forwardRef are controls (select button, dropdown, nav button). I don't know the exact pattern with this, but this component doesn't seem to fit with the other things that implement forwarding refs around.

src/components/label-pill/LabelPill.js
booc0mtaco: If it's ok, will postpone this until we gather up all the pill-like things, and evaluate them. maybe there's an intermediate step that would conflict with this idea

src/components/label-pill/LabelPill.scss
```diff
@@ -0,0 +1,56 @@
+@import '../../styles/variables';
+
+.bdl-LabelPill {
+    margin-right: 4px;
```
j3tan: To keep this flexible, we should not include margins for core components and allow the parent apps to designate spacing as appropriate. Expected usage of this could be something like:

`<LabelPill className="mrs" />`

This is pretty clear with our usage of `<Button>` where we were frequently forced to override the outer margin

booc0mtaco: I agree, but if it is pretty clear, why was it added initially to these other components? has whatever existed then that made it a good idea changed? Perhaps the frequent overrides are a problem with the designs...that elements' associations vary more than they ought to. In past component libraries that I've built, the spacing between components was defined up front so everything fell into predicable units (and made overrides almost non-existent)

Also do we have a suite of mixins/classes for layout, or is "mrs" hypothetical?

src/components/label-pill/LabelPill.stories.md
```diff
@@ -0,0 +1,3 @@
+`import LabelPill from 'box-ui-elements/es/components/label-pill/LabelPill';`
+
+_Note_: this component is meant to be used with one of the formatted strings produced by react-intl (`FormattedMessage`, `FormattedRelativeDate`, etc.) **or** one of our icon assets.
```
priyajeet: Still need the note?

src/components/label-pill/LabelPill.scss
```diff
@@ -0,0 +1,51 @@
...
+        border-color: $bdl-light-blue;
+    }
+
+    &.bdl-LabelPill--info {
+        background-color: $bdl-light-blue-10;
+        border-color: $bdl-light-blue;
+    }
+}
+
+.bdl-LabelPill-iconContent {
```
priyajeet: Do we need these classes? Can they just be importable mixins? Or let the parent app deal with it.

src/components/label-pill/LabelPill.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export interface LabelPillProps {
+    children: React.ReactNode;
+    type?: LabelPillStatus;
+}
+
+const LabelPill = React.forwardRef((props: LabelPillProps, ref: React.Ref<HTMLSpanElement>) => {
+    const { children, type = LabelPillStatus.DEFAULT, ...rest } = props;
+    const labelPillClasses = classNames('bdl-LabelPill', {
+        [`bdl-LabelPill--${type}`]: type !== LabelPillStatus.DEFAULT,
```
priyajeet: You might be able to get away with just not having this condition, as in, just always add `bdl-LabelPill--${type}`. Having a default class that does nothing is a noop. And helps parent to override for whatever reason.

src/components/label-pill/LabelPill.scss
priyajeet: There was some talk a while back to use https://github.com/suitcss/suit/tree/master/packages/utils but unsure if it has the granularity of how we used to have "mrs".

If you think most parent apps will have to do 4px, maybe a SCSS variable can be provided here that they can use in their css.

alexkrolick: > Also do we have a suite of mixins/classes for layout, or is "mrs" hypothetical?

We do have utility classes: https://github.com/box/box-ui-elements/blob/master/src/styles/modifiers/_margins.scss

But we need to review the values and align with our grid size. After that I would say that is the primary unit of layout between elements, including add flex positioning helpers etc.

src/components/label-pill/LabelPill.stories.md
booc0mtaco: this is still true, tho there's nothing to enforce it. If someone tries to put different content in this component, it may not have a good layout.

src/components/label-pill/LabelPill.tsx
booc0mtaco: i actually prefer this, as it allows for a classless default look and feel. the current approach only adds a modifier when the modifier will have an effect on the layout.

src/components/label-pill/LabelPill.scss
booc0mtaco: these classes define the content types and spacing specific to this component. I don't think any other component defines these specific values, so they aren't generic. 

another idea was to borrow an approach like we use in the media card component. if we created "sub-components" of LabelPill with these styles, the pattern could look like:

```
<LabelPill>
    <LabelPill.Text>{'some text content'}</LabelPill.Text>
</LabelPill>
```

having mixins seemed a bit tricky to me, because it would require the user to create some class for the content, then use the mixin in that class, which is an additional step. Here, we document the usage and class names in examples, so the user can follow this approach without adding any additional classes

priyajeet: +2 for the sub-comp approach

alexkrolick: Throwing out another idea:

Since most of the time this is used with text content, having to add an inner component seems like a little extra work. Could we instead have default padding & alignment for text and then an `--round` modifier + prop to force it to be a circle for the icon use case?

booc0mtaco: @alexkrolick this could work. One additional use case that was proposed in a design chat was to have an icon has text which is right next to the icon, i.e., no additional padding. that would be a bit in conflict with a default padding approach.

I'm inclined to try the sub-comp approach as it gives us clear error messaging on usage, and would likely allow for us to:

- more easily specify sizing for this pill to account for future sharing in other contexts, like having a prop for pill size,
- add a "mode" prop similar to icon bars in macOS where you can specify both the icon **and** text, but use a "text only" mode in some contexts
- extend the sub-component list for cases where we have text/icon combos that follow different rules

Screen Shot 2020-02-06 at 2 26 44 PM

src/components/label-pill/LabelPill.tsx
priyajeet: You can always move the defaults into the default modifier

src/components/label-pill/LabelPill.scss
alexkrolick: The subcomponent approach sounds good for the icon + text mode, or even this avatar pill:

Screen Shot 2020-02-06 at 14 29 34

src/components/label-pill/LabelPill.tsx
```diff
@@ -0,0 +1,35 @@
...
+    ALERT = 'alert',
+    ERROR = 'error',
+}
+
+export interface LabelPillProps {
+    children: React.ReactNode;
+    type?: LabelPillStatus;
+}
+
+const LabelPill = React.forwardRef((props: LabelPillProps, ref: React.Ref<HTMLSpanElement>) => {
```
priyajeet: Should we now call it just "Pill" hoping its one Pill to rule them all

booc0mtaco: I went with "labelPill" in this case to avoid a conflict with the PillCloud component (which already uses the namespace "bdl-Pill" unfortunately).

booc0mtaco: oh yeah, and there's a Pill inside PillSelector dropdown. the styles are shared between these two..

src/components/label-pill/LabelPillText.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react';
+
+export interface LabelPillTextProps {
+    children: React.ReactNode;
```
priyajeet: React.ReactChild

src/components/label-pill/LabelPill.tsx
```diff
@@ -0,0 +1,43 @@
...
+    INFO = 'info',
+    HIGHLIGHT = 'highlight',
+    SUCCESS = 'success',
+    WARNING = 'warning',
+    ALERT = 'alert',
+    ERROR = 'error',
+}
+
+export interface LabelPillProps {
+    children: React.ReactNode;
```
priyajeet: React.ReactChild

src/components/label-pill/index.ts
```diff
@@ -0,0 +1 @@
+export { default, LabelPillStatus } from './LabelPill';
```
priyajeet: Also add to index.ts one level higher

priyajeet: there seems to be no default anymore btw

src/components/label-pill/LabelPill.stories.md
```diff
@@ -0,0 +1,3 @@
+`import LabelPill from 'box-ui-elements/es/components/label-pill/LabelPill';`
```
priyajeet: import { LabelPill }

priyajeet: ends with the directory

src/components/label-pill/index.ts
priyajeet: I mean we should not export `LabelPillCombo` as the named export

src/components/label-pill/LabelPill.tsx
booc0mtaco: This can have a single of multiple instances of its subcomponents, so using `React.ReactChild | Array<React.ReactChild>`

PR #1910, opened 2020-02-04
fix(table): Fix broken test for makeSelectable HOC

rwatari:
Fixing test so I can verify an upcoming refactor of this HOC


boxcla:
Verified that @rwatari has signed the CLA. Thanks for the pull request!

PR #1914, opened 2020-02-04
fix(test): Fix tests using rewire

rwatari:
Use jest mocks to fix tests that used __Rewire__


boxcla:
Verified that @rwatari has signed the CLA. Thanks for the pull request!

src/components/hotkeys/__tests__/HotkeyLayer.test.js
```diff
@@ -3,19 +3,22 @@ import sinon from 'sinon';
...
 const sandbox = sinon.sandbox.create();
 
+jest.mock('../HotkeyService');
+
 describe('components/hotkeys/HotkeyLayer', () => {
     beforeEach(() => {
         // This is required to prevent actually invoking HotkeyService, which causes
         // HotkeyService tests to fail
-        HotkeyLayer.__Rewire__('HotkeyService', class {});
+        HotkeyService.mockImplementation(() => {});
```
mxiao6: Is it necessary to mock every time? Can you mock once on the top then `mockReset` before or after each?

rwatari: probably not. I just did a direct translation of the existing tests. I'll see if that works

PR #1916, opened 2020-02-04
chore: Pin @hapi/address to 2.1.1 to prevent IE11 breakage

j3tan:
Note: v2.1.2+ requires a TextEncoder polyfill


boxcla:

PR #1917, opened 2020-02-04
refactor(css): Add backwards-compatible names to common components

sahiga:
BREAKING CHANGE: Many CSS and JS/TSX files have been updated to include bdl-namespaced class names.


boxcla:

mickr:
Should this be a chore if it is a breaking change?

sahiga:
> Should this be a chore if it is a breaking change?

@mickr Good point. I'll change the type to `refactor`.

scripts/utils/bdlClassnameManager
```diff
@@ -29,12 +29,12 @@ const writeFile = promisify(fs.writeFile);
...
-    // 'btn-group': 'bdl-ButtonGroup',
-    // 'btn-plain': 'bdl-Button--plain',
-    // 'btn-primary': 'bdl-Button--primary',
-    // 'is-disabled': 'bdl-is-disabled',
-    // 'label': 'bdl-Label',
+    btn: 'bdl-Button',
+    'btn-group': 'bdl-ButtonGroup',
+    'btn-plain': 'bdl-Button--plain',
+    'btn-primary': 'bdl-Button--primary',
+    'is-disabled': 'bdl-is-disabled',
```
manishPh: SUIT [naming convention doc](https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md#components) says <ComponentName> is not optional. `is-disabled` looks more like a modifier and should be used in conjunction with the component itself, e.g. `bdl-SelectButton--is-disabled`.

manishPh: @sahiga `is-disabled` is correct, you can ignore above comment.

scripts/utils/bdlClassnameManager
```diff
@@ -29,12 +29,12 @@ const writeFile = promisify(fs.writeFile);
...
 const conversionMap = {
-    // 'btn': 'bdl-Button',
-    // 'btn-group': 'bdl-ButtonGroup',
-    // 'btn-plain': 'bdl-Button--plain',
-    // 'btn-primary': 'bdl-Button--primary',
-    // 'is-disabled': 'bdl-is-disabled',
-    // 'label': 'bdl-Label',
+    btn: 'bdl-Button',
+    'btn-group': 'bdl-ButtonGroup',
+    'btn-plain': 'bdl-Button--plain',
```
priyajeet: BTW, `PlainButton` component, doesn't use `Button` component (something we should eventually fix). So this can't be a modifier atm.

sahiga: @phora Would you recommend changing this to `bdl-PlainButton` or keeping `btn-plain` for now?

alexkrolick:
Will adding the new classes for buttons lock us into something resembling the current API when we rewrite the Button components to remove their current problematic :hover styles and margins?

src/features/unified-share-modal/SharedLinkSection.js
```diff
@@ -344,7 +344,7 @@ class SharedLinkSection extends React.Component<Props> {
                 <hr />
                 {/* eslint-disable-next-line jsx-a11y/label-has-for */}
                 <label>
-                    <span className="label">
+                    <span className="label bdl-Label">
```
il10: why is `label` being kept here as well as adding `bdl-Label` ?

tjgupta: I'd guess it's for backward compatibility, but since this isn't a `<Label>` component should this be something like `SharedLinkSection-label`? Or perhaps shouldn't be part of this change at all?

src/features/unified-share-modal/UnifiedShareModal.scss
```diff
@@ -92,6 +92,7 @@
     }
 
     .bdl-PillSelectorDropdown .bdl-PillSelector, // need both selectors for specificity
+    .pill-selector-wrapper .pill-selector-input-wrapper,
```
tjgupta: Out of curiosity, why are these being added back now? Haven't we already done the work to use the new classNames in the parent app where needed?

src/features/unified-share-modal/SharedLinkSection.js
booc0mtaco: It'd be on us to figure out why we have a regular label tag instead of the label component here. I can't recall, so might be some refactor we need to do.

booc0mtaco:
@alexkrolick what's up with the button styles and :hover? 

@sahiga My review is potent; let's make sure we cover any comments made here before merging

src/features/unified-share-modal/SharedLinkSection.js
sahiga: Yes, I'm keeping the old class names just in case there are overrides in EUA and AC. This `span` element is still using the `.label` class from Label.scss, so I think it's safe to add `.bdl-Label` to it.

src/features/unified-share-modal/UnifiedShareModal.scss
sahiga: Good point. I will remove these.

sahiga:
> Will adding the new classes for buttons lock us into something resembling the current API when we rewrite the Button components to remove their current problematic :hover styles and margins?

@alexkrolick Can you provide an example of how the Button components will be rewritten?

scripts/utils/bdlClassnameManager
priyajeet: bdl-PlainButton i would say

alexkrolick:
@booc0mtaco @sahiga 

https://github.com/box/box-ui-elements/blob/8446680ade41b16e5e7ef90ebc178ce6520c0a33/src/styles/common/_buttons.scss#L85-L93

The way this is written requires everyone that adds a className with an alternate margin for `.btn-plain` (`<PlainButton />`) to also add `:hover`, `:active` states for that class. Otherwise it will jump back to margin 0 when the user hovers. This is a frequent issue.

https://github.com/box/box-ui-elements/blob/8446680ade41b16e5e7ef90ebc178ce6520c0a33/src/styles/common/_buttons.scss#L5-L9

This default margin also breaks the rule of component composition that we have established where spacing between elements is controlled by the consumer.

Additionally, the React API doesn't make a whole lot of sense - having 4 different components with different import paths for what could just be a single `Button` with a prop to change between plain, secondary, or primary styles.

priyajeet:
@alexkrolick @sahiga
Beyond the scope of this PR, but Button should eventually import its own SCSS file. Likewise PrimaryButton and PlainButton (this one should use Button also underneath). Obviously we can combine all the 3 components (or atleast in the same src/components/button folder) and props define what kind of button, but this can be the next step after removing the global _buttons.scss (or the bandaid can be ripped off at one go)

alexkrolick:
@phora, right, which is why I am concerned that if we take over the .bdl-Button namespace and people start targetting that, the overrides won't be compatible with a future, better version.

sahiga:
@alexkrolick That's a valid concern. I will remove the Button-related changes from this PR, but we should prioritize the refactoring of the Button components.

PR #1918, opened 2020-02-05
chore(textarea): add tooltip position

danielkim328:


boxcla:
Hi @danielkim328, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

src/components/text-area/TextArea.js
```diff
@@ -5,13 +5,16 @@ import uniqueId from 'lodash/uniqueId';
 
 import Label from '../label';
 import Tooltip from '../tooltip';
+import type { Position } from '../tooltip';
```
priyajeet: Just a FYI, tooltip's index.ts no longer exports this. It works right now because of a fake index.js.flow file. Ideally this component should be converted to TypeScript, and you can utilize TooltipPosition import.

danielkim328: @priyajeet Can that be a follow up PR? or is it now a requirement to convert this component to Typescript

priyajeet: Can be followup

src/components/text-area/TextArea.js
```diff
@@ -5,13 +5,16 @@ import uniqueId from 'lodash/uniqueId';
...
 
 import './TextArea.scss';
 
 type Props = {
     className?: string,
     description?: React.Node,
     error?: React.Node,
+    /** Renders error tooltip at the specified position (positions are those from Tooltip) */
+    errorPosition?: Position,
```
priyajeet: maybe `errorTooltipPosition` is a more descriptive prop

PR #1921, opened 2020-02-06
feat(css): add utility classes for spacing

alexkrolick:
- Creates utility classes like "bdl-pr-0" to implement spacing based on multipliers of our grid unit
- Pulls in a "generate-utility" scss mixin from Bootstrap that we can use to generate a variety of utility classes; modified slightly to add .bdl- prefix
- Adds deprecation notice to old "modifiers" SCSS as it does not follow the grid and doesn't use prefixes (code kept for legacy use)

Screen Shot 2020-02-05 at 23 28 37


boxcla:

src/styles/utility/_spacing.scss
```diff
@@ -0,0 +1,126 @@
+/*
+Generates utility classes for spacing (margin and padding)
+
+The convention is ".bdl-py-0"
```
priyajeet: Maybe we should stick to SUIT style, and make the property classes below full words.
https://github.com/suitcss/suit/blob/master/doc/utilities.md
https://github.com/suitcss/suit/tree/master/packages (utils packages)

.bdl-u-marginLeft
.bdl-u-marginX

.bdl-u- prefix
margin, padding
null, X, Y, Top, Right, Bottom, Left
0,1,2... multiplier for grid units

src/styles/utility/_spacing.scss
```diff
@@ -0,0 +1,126 @@
...
+        "margin": (
+            responsive: false,
+            property: margin,
+            class: m,
+            values: map-merge($spacers, (auto: auto))
+        ),
+        "margin-x": (
+            responsive: false,
+            property: margin-right margin-left,
+            class: mx,
```
priyajeet: I think maybe we can get rid of X and Y. And force people to be explicit. Unless there are use cases were we do this often.

src/styles/utility/_spacing.scss
```diff
@@ -0,0 +1,126 @@
...
+            responsive: false,
+            property: padding,
+            class: p,
+            values: $spacers
+        ),
+        "padding-x": (
+            nsive: false,
+            property: padding-right padding-left,
+            class: px,
+            values: $spacers
```
priyajeet: Likewise

src/styles/utility/_spacing.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+const Spacing = () => (
+    <>
+        <h2>Spacing Utility Classes</h2>
+
+        <SpacingDemo>
+            <Box className="bdl-px-3 bdl-py-4 bdl-m-3" />
+        </SpacingDemo>
+
+        <SpacingDemo>
+            <Box className="bdl-px-5 bdl-py-2 bdl-m-4" />
```
priyajeet: So with above changes, this will read as
```jsx
<Box className="bdl-u-paddingTop-5 bdl-u-paddingRight-5 bdl-u-paddingLeft-2 bdl-u-paddingBottom-2  bdl-u-margin-4" />
```

src/styles/utility/_spacing.scss
```diff
@@ -0,0 +1,126 @@
...
+        5: $spacer * 3,
+    ),
+    $spacers
+);
+
+$spacing-utilities: () !default;
+$spacing-utilities: map-merge(
+    (
+        "margin": (
+            responsive: false,
```
priyajeet: is "responsive" being read anywhere? Needed?

alexkrolick: I didn't want to deviate too much from the bootstrap generate-utility function; this allows creating utility classes that are aware of breakpoints.

priyajeet:
We should have a discussion if people are indeed going to use utilities. Historically we used to have them but they were just ignored.

src/styles/utility/_spacing.scss
alexkrolick: x & y (v and h) are very useful actually, every utility framework supports them

For example: https://tailwindcss.com/docs/padding/

src/styles/utility/_spacing.scss
alexkrolick: Maybe on the `u`... but I think compactness is an advantage here.

alexkrolick:
@priyajeet I'll send a doc internally. It's part of our design system push to have components that don't have any external positioning by default.

booc0mtaco:
@alexkrolick let's come back to this again once you free up

booc0mtaco:
Probably time to close this one

PR #1922, opened 2020-02-06
Implement a number abbreviation formatter

ehoogerbeets:
This will abbreviate large numbers in a locale-friendly way, using the correct groupings and localized abbreviation text for the locale. The data for this comes from the Unicode CLDR project, which we distilled down to only the parts we need and put into the @box/cldr-data package.

Usage:

```javascript
import numAbbr from '../../utils/numbAbbr';

console.log(`The shortened number is ${numAbbr(1234567)}.`);
// -> prints "1M" on the console
```
The `numAbbr` function will shorten the number by rounding to 1-4 significant digits and adding in the suffix/prefix for the size. It is not 3 digits because some cultures, such as Chinese, use groups of ten thousands (1 Man) instead of thousands like in Western cultures, so you can have up to 4 digits for the "number of man" in a number.
 
By default, `numAbbr` will format number with the "short" style, meaning that the text will be "M" for "millions". In the long style, you have it write out the full text:

```javascript
import numAbbr from '../../utils/numbAbbr';

console.log(`The shortened number is ${numAbbr(1234567, {length: "long")}.`);
// -> prints "1 million" on the console
```

Technical Stuff:

There is a new alias 'box-locale-data' which includes the subfile within @box/cldr-data that contains all the locale data we need for the current locale. That means that we will not include all of the data for all locales into the webpack bundle, which should help with the footprint.

The `numAbbr` code relies on the same `IntlMessageFormat` function that react-intl uses in order to format the string with the size and to format the number in a locale sensitive manner. That means it will use native digits for some locales. The only locale that box supports that uses native digits is Bengali for India (bn-IN). All our other locales use the regular "western" digits.


boxcla:

ehoogerbeets:
I'll be doing a "squash and merge", so it doesn't matter about the commit messages.

ehoogerbeets:
Actually, I realized this branch is based off of Jee's branch instead of master. I'm going to have to create a new branch and a new PR so I don't accidentally submit some of her code along with mine. Closing this one.

PR #1923, opened 2020-02-06
feat(i18n): implement a number abbreviation formatter

ehoogerbeets:
BREAKING CHANGE: You must add `@box/cldr-data` to your dependencies.

This will abbreviate large numbers in a locale-friendly way, using the correct groupings and localized abbreviation text for the locale. The data for this comes from the Unicode CLDR project, which we distilled down to only the parts we need and put into the `@box/cldr-data` package.

Usage:

```javascript
import numAbbr from '../../utils/numbAbbr';

console.log(`The shortened number is ${numAbbr(1234567)}.`);
// -> prints "1M" on the console
```
The `numAbbr` function will shorten the number by rounding to 1-4 significant digits and adding in the suffix/prefix for the size. It is not 3 digits because some cultures, such as Chinese, use groups of ten thousands (1 Man) instead of thousands like in Western cultures, so you can have up to 4 digits for the "number of man" in a number.
 
By default, `numAbbr` will format number with the "short" style, meaning that the text will be "M" for "millions". In the long style, you have it write out the full text:

```javascript
import numAbbr from '../../utils/numbAbbr';

console.log(`The shortened number is ${numAbbr(1234567, {length: "long")}.`);
// -> prints "1 million" on the console
```

Technical Stuff:

There is a new alias 'box-locale-data' which includes the subfile within @box/cldr-data that contains all the locale data we need for the current locale. That means that we will not include all of the data for all locales into the webpack bundle, which should help with the footprint.

The `numAbbr` code relies on the same `IntlMessageFormat` function that react-intl uses in order to format the string with the size and to format the number in a locale sensitive manner. That means it will use native digits for some locales. The only locale that box supports that uses native digits is Bengali for India (bn-IN). All our other locales use the regular "western" digits.

There is a way of doing this same thing with `react-intl`. However, the polyfill for it does not work on IE11 (which is where you need the polyfill in the first place!). This function works nicely on IE11 as a ponyfill.

If you don't care about IE11, then you can just use `react-intl` like this:

```typescript
<NumberFormat value={myValue} notation="compact"/>

or programmatically:

const myValueAsAString: string = intl.formatNumber(myValue, {notation: "compact"});
```


boxcla:

.storybook/webpack.config.js
```diff
@@ -12,6 +12,7 @@ module.exports = async ({ config }) => {
     config.resolve.alias = {
         ...config.resolve.alias,
         'box-ui-elements-locale-data': path.resolve(`i18n/${language}`),
+        'box-locale-data': path.resolve(`node_modules/@box/cldr-data/locale-data/${language}`),
```
jeeyi: We need to apply the same changes to `scripts/webpack.config.js`.

src/utils/numAbbr.js
```diff
@@ -0,0 +1,82 @@
...
+
+import IntlMessageFormat from 'intl-messageformat';
+import { languages, numbers } from 'box-locale-data';
+
+interface NumAbbrOptions {
+    length: string;
+    locale: string;
+}
+
+export function numAbbrWithLocale(localedata: object, num: number, locale: string, options?: NumAbbrOptions): string {
```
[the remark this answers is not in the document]
ehoogerbeets: In normal usage, the caller would just call the `numAbbr` function to abbreviate the number in the current locale. They do not call `numAbbrWithLocale` directly.

But, for the unit tests, we needed a way to test other locales. So, I made this as a function that the unit tests can call where the localedata is passed in. In the unit tests (q.v.), we load the locale data from @box/cldr-data explicitly and pass them in. That way, we don't have to do any dynamic imports in the `numAbbr` code. We don't have to worry about the situation where the caller passes in the wrong locale spec compared to the localedata, because it is not a public API.

.storybook/webpack.config.js
ehoogerbeets: Good catch! Missed that one. I've added it now.

src/utils/numAbbr.js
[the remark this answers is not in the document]
jeeyi: Can you rename `localedata` to something like `numbersLocaleData`? My initial question was why can't you get `locale` from `localedata`. Now I realized that `localedata` here is just numbers.

yarn.lock
```diff
@@ -3089,6 +3089,11 @@
   resolved "https://registry.yarnpkg.com/@bcoe/v8-coverage/-/v8-coverage-0.2.3.tgz#75a2e8b51cb758a7553d6804a5932d7aace75c39"
   integrity sha512-0hYQ8SB4Db5zvZB4axdMHGwEaQjkZzFjQiN9LVYvIFB2nSUHW9tYpxWriPrWDASIxiaXax83REcLxuSdnGPZtw==
 
+"@box/cldr-data@^34.0.0":
+  version "34.0.0"
+  resolved "https://box.jfrog.io/box/api/npm/boxnpm/@box/cldr-data/-/cldr-data-34.0.0.tgz#bae9b3a7200b7692afa2ad02f7fd6f0ec6bc20c7"
```
jeeyi: This is not correct @ehoogerbeets. It should be:
```
"@box/cldr-data@^34.0.0":
  version "34.0.0"
  resolved "https://registry.yarnpkg.com/@box/cldr-data/-/cldr-data-34.0.0.tgz#bae9b3a7200b7692afa2ad02f7fd6f0ec6bc20c7"
  integrity sha512-QhzBMLYx9qOZeOf4OR6KWbpP72RJ2Wh1KCF2qL8EosIRkdDSYjoNCcNXMQTaDnkRQ12R7kcASG869cMAn8KBpg==
```

src/utils/numAbbr.js
```diff
@@ -0,0 +1,83 @@
...
+    numbersData = numbersData || numbers;
+
+    switch (typeof num) {
+        case 'boolean':
+            num = num ? 1 : 0;
+            break;
+
+        case 'string':
+            num = parseInt(num, 10);
+            if (Number.isNaN(num)) return '0';
```
priyajeet: Number.isNaN doesn't work in IE11. Probably use lodash/isNaN

src/utils/numAbbr.js
```diff
@@ -0,0 +1,83 @@
...
+ * For locales that have complex plurals, such as Russian or Polish, this function
+ * returns the correctly pluralized suffix/prefix to go along with the scaled number.
+ *
+ * @param {*} num the number or numbers to abbreviate. (also accepts numbers in string form,
+ * or an array of numbers).
+ * @param {NumAbbrOptions=} options options governing how the output is generated
+ * @return {string} the number in abbreviated form as a string, or '0' if there was
+ * invalid input such as null instead of a number
+ */
+export default function(num: number, options?: NumAbbrOptions): string {
```
priyajeet: Can you move the export to the bottom to keep it consistent with other files

src/utils/numAbbr.js
```diff
@@ -0,0 +1,83 @@
+/**
```
priyajeet: This should be a TS file, not using flow anymore.
Likewise its test should also be a TS file.

src/utils/numAbbr.js
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * The abbreviation words/letters can have the length "short" or "long", specified
+ * with the "length" property in the options. If "short",
+ * then this function uses an abbreviation of the bucket such as "12M". If "long",
+ * then the name of the bucket is written out in full, such as "12 million".
+ *
+ * For locales that have complex plurals, such as Russian or Polish, this function
+ * returns the correctly pluralized suffix/prefix to go along with the scaled number.
+ *
+ * @param {*} num the number or numbers to abbreviate. (also accepts numbers in string form,
```
priyajeet: Nuke the JSDocs, TS definitions should suffice.

src/utils/numAbbr.js
priyajeet: You can see examples in src/components/button or src/components/tooltip. They have mostly tsx files, but yours should just be ts as there is no jsx in here.

src/utils/numAbbr.js
```diff
@@ -0,0 +1,83 @@
...
+import { languages, numbers } from 'box-locale-data';
+
+interface NumAbbrOptions {
+    length: string;
+    locale: string;
+    numbersData: object;
+}
+
+function numAbbrWithLocale(numbersData: object, num: number, locale: string, options?: NumAbbrOptions): string {
+    if (!num) return '0';
```
priyajeet: { } around all if/else blocks. So that we can easily put breakpoints.

src/utils/numAbbr.js
```diff
@@ -0,0 +1,83 @@
...
+    let { length } = options || {};
+    length = length || 'short';
```
priyajeet: I think 'short' and other allowed values should become an enum. See Tooltip.tsx or Button.tsx for examples.

src/utils/numAbbr.js
```diff
@@ -0,0 +1,83 @@
...
+    const digits: number = exponent >= formats.length ? exponent - formats.length + 3 : formats[exponent].digits;
+    const count: number = Math.round(num / 10 ** (exponent - digits + 1));
+    const template = new IntlMessageFormat(
+        formats[exponent > formats.length ? formats.length - 1 : exponent].msg,
+        locale,
+    );
+    return template.format({ count });
+}
+
+/**
```
priyajeet: Should we be using https://www.npmjs.com/package/filesize?
We already use it inside src/utils/getFileSize.js
(also need to nuke/replace src/utils/size.js)

priyajeet: Hmm, I guess this utility is not for file sizes, but numbers in general. nvm.

package.json
```diff
@@ -121,6 +121,9 @@
         "node": ">=10.x",
         "yarn": ">=1.10.0"
     },
+    "dependencies": {
```
priyajeet: probably move this to dev deps and peer deps.

ehoogerbeets: This is a real dependency, though. If this dependency is not satisfied, the code will not run properly and the unit tests will not work....

src/utils/numAbbr.js
```diff
@@ -0,0 +1,83 @@
...
+ * @flow
+ * @file Function to abbreviate a number as a string in a locale-sensitive manner.
+ * @author Box
+ */
+
+import IntlMessageFormat from 'intl-messageformat';
+import { languages, numbers } from 'box-locale-data'; // eslint-disable-line
```
priyajeet: How would this work for external platform customers?
Right now documentation suggests we do whats listed here https://github.com/formatjs/react-intl/blob/master/docs/Getting-Started.md#runtime-requirements

priyajeet: This is how an external customer will use https://github.com/box/box-ui-elements-demo/blob/master/src/Main.js

priyajeet: One way is that the cldr data should be published along with buie as an importable module?

package.json
priyajeet: Dev deps are also real dependencies. And with peer deps, the peer needs to install it also.

ehoogerbeets: moved to peer dependencies now

src/utils/numAbbr.js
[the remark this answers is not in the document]
ehoogerbeets: renamed

src/utils/numAbbr.js
ehoogerbeets: done

src/utils/numAbbr.js
ehoogerbeets: That was a heck of a lot of work, but it is converted to TS now

src/utils/numAbbr.js
ehoogerbeets: done

src/utils/numAbbr.js
ehoogerbeets: done

src/utils/numAbbr.js
ehoogerbeets: done

src/utils/numAbbr.js
ehoogerbeets: done

yarn.lock
ehoogerbeets: done

src/utils/numAbbr.js
ehoogerbeets: Yes, the @box/cldr-data module is importable separately. The way react-intl does it is have your app explicitly require the locale data for all the locales the app needs, so that they all make it into the webpacked result.

The way that our own locale data is included in box-ui-elements is similar to the way that the translated strings are included. That is, the 'box-locale-data' module is actually an alias that causes `node_modules/@box/cldr-data/es/locale-data/${language}` to be loaded instead. That way, only the locale data for the current locale is included in the webpacked result, instead of the locale data for all locales at once and picking the right one at run time. Helps reduce the disk/memory footprint!

ehoogerbeets:
@priyajeet since this is a completely new function/file, we don't need the *.flow files, right?

package.json
```diff
@@ -315,6 +315,7 @@
         "formik": "^2.0.0",
         "ilib-tree-node": "^1.2.0",
         "immutable": "^3.7.4",
+        "intl-messageformat": "^7.8.4",
```
jstoffan: This commit will need to be marked as a breaking change.

src/cldr-data/bn-IN.ts
```diff
@@ -0,0 +1,244 @@
+/**
+ * Auto-generated Locale Data File for bn-IN.
+ * This is generated in the @box/cldr-data project.
```
jstoffan: Can we import these files from `@box/cldr-data` directly rather than committing copies of them?

src/utils/numAbbr.ts
```diff
@@ -0,0 +1,92 @@
...
+ * returns the correctly pluralized suffix/prefix to go along with the scaled number.
+ */
+function numAbbr(num: unknown, options: NumAbbrOptions = { length: Lengths.short }): string | string[] {
+    if (!num) return '0';
+
+    // languages contains info about the current locale
+    const { locale, numbersData } = options;
+    const loc = locale || languages.bcp47Tag || 'en-US';
+    const numData: NumbersData = numbersData || numbers;
+    let value;
```
jstoffan: Can this be declared in the block where it's used? It doesn't seem to be used anywhere else.

src/utils/numAbbr.ts
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * For locales that have complex plurals, such as Russian or Polish, this function
+ * returns the correctly pluralized suffix/prefix to go along with the scaled number.
+ */
+function numAbbr(num: unknown, options: NumAbbrOptions = { length: Lengths.short }): string | string[] {
+    if (!num) return '0';
+
+    // languages contains info about the current locale
+    const { locale, numbersData } = options;
+    const loc = locale || languages.bcp47Tag || 'en-US';
```
jstoffan: It looks like these values can only be undefined, not null or otherwise falsy. Can we provide default values during assignment to avoid the intermediate variables?

```
const { locale = languages. bcp47Tag || 'en-US', numbersData = numbers } = options;
```

src/utils/numAbbr.ts
```diff
@@ -0,0 +1,92 @@
...
+    let value;
+
+    switch (typeof num) {
+        case 'boolean':
+            num = num ? 1 : 0;
+            break;
+
+        case 'string':
+            value = parseInt(num, 10);
+            if (isNaN(value)) {
```
jstoffan: Should this check instead be part of `numAbbrWithLocale` rather than a special case here?

src/utils/numAbbr.ts
```diff
@@ -0,0 +1,92 @@
...
+                return '0';
+            }
+            break;
+
+        case 'object':
+            return Array.isArray(num)
+                ? num.map(n => {
+                      return numAbbrWithLocale(n, options, numData, loc);
+                  })
+                : '0';
```
jstoffan: Should we throw an error rather than failing silently?

src/utils/numAbbr.ts
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * The abbreviation words/letters can have the length "short" or "long", specified
+ * with the "length" property in the options. If "short",
+ * then this function uses an abbreviation of the bucket such as "12M". If "long",
+ * then the name of the bucket is written out in full, such as "12 million".
+ *
+ * For locales that have complex plurals, such as Russian or Polish, this function
+ * returns the correctly pluralized suffix/prefix to go along with the scaled number.
+ */
+function numAbbr(num: unknown, options: NumAbbrOptions = { length: Lengths.short }): string | string[] {
```
jstoffan: We generally want to avoid abbreviation of variable names, where possible. Can this be `input`, maybe?

package.json
ehoogerbeets: We already depend on intl-messageformat implicitly, as it is the basis for react-intl. However, lint complained about it because we didn't explicitly mention it in the package.json file when we explictly import it in the new code. So, it's not really a dependency change.

src/cldr-data/bn-IN.ts
ehoogerbeets: @priyajeet wanted me to move them away from there and put them directly into BUIE (see the discussion in #moji on slack).

src/utils/numAbbr.ts
ehoogerbeets: If you do that, then lint complains.

src/utils/numAbbr.ts
ehoogerbeets: I'll change that

src/utils/numAbbr.ts
ehoogerbeets: The idea behind `numAbbrWithLocale` is that it only works with actual numbers and `numAbbr` is the wrapper that does whatever is necessary to convert the input to a number first. Also, in order to be able to test in multiple languages, the unit tests had to have access to the subfunction with the locale, or else we would be stuck testing US English only.

src/utils/numAbbr.ts
ehoogerbeets: I'm not sure it is a good idea for a formatter that takes anything as its input should ever throw. It's used for display, not for real business logic. If it does throw and the caller doesn't catch properly, that breaks the UI.

src/utils/numAbbr.ts
ehoogerbeets: will do

src/utils/numAbbr.ts
jstoffan: In that case, is it needed? It's only used once:

```
if (isNaN(parseInt(input, 10)) {
    return '0';
}
```

src/utils/numAbbr.ts
jstoffan: My concern was around the handling of nested values in arrays, but it looks like you've addressed that with recursion.

booc0mtaco:
@ehoogerbeets could we use [numbro](https://numbrojs.com/format.html) here? I've used it in the past, but not sure if it overlaps with our use cases

ehoogerbeets:
@booc0mtaco We could use that, but it doesn't follow the CLDR conventions. We are only adding this as a sort of a ponyfill for Intl.NumberFormat on IE11, since the polyfill that formatjs provides does not seem to work on IE11. (It crashes -- though they just added it 3 weeks ago, so maybe they are working on a fix soon?)

PR #1925, opened 2020-02-09
fix(nudge): update css height property

laughnan:
Screen Shot 2020-02-08 at 23 36 24


boxcla:

jstoffan:
Seems like this should be a `fix` rather than a `chore`.

sahiga:
@laughnan Does the `Nudge` visual regression test snapshot need to be updated?

laughnan:
@sahiga nope! the existing snapshot has content greater than 450 pixels (i checked).

PR #1926, opened 2020-02-10
fix(textinputwithcopybutton): clicks on URL within USM should select all

il10:
bug-fix: clicks on URL within USM should select all


boxcla:

PR #1928, opened 2020-02-11
fix(modal): Fix shared link setting modal input overflow

mxiao6:
Before:
image

After:
image


boxcla:

PR #1930, opened 2020-02-11
fix(explorer): Fix search dropdown too long

mxiao6:
Before:
dropdown-before

After:
dropdown-after


boxcla:

PR #1932, opened 2020-02-12
feat(unified-share-modal): Add fuzzy search

stevenzyang:
Cleaned version of #1140
Adds utility method to compute suggested collaborators given a list of contacts and cached suggested collaborators.


boxcla:

src/features/unified-share-modal/utils/computeSuggestedCollabs.js
```diff
@@ -0,0 +1,67 @@
...
+ */
+import type { SuggestedCollab, SuggestedCollabLookup } from '../flowTypes';
+import fuzzySearch from '../../../utils/fuzzySearch';
+
+/**
+ * Function to compute suggested collaborators given a list of contacts and cached suggested collaborators.
+ * Used by input components to help augment API results with cached suggestions using a fuzzy search.
+ *
+ * @param {Array<C>} contacts Contacts returned from the server with exact matches
+ * @param {SuggestedCollabLookup} suggestedCollaborators Cached suggested collab lookup to fuzzy match on
```
priyajeet: Get rid of the jsdocs, we dont add them anymore as people rarely update them or read them. Types and arg names should be self documenting.

src/features/unified-share-modal/utils/computeSuggestedCollabs.js
```diff
@@ -0,0 +1,67 @@
...
+ * @param {string} inputValue Value the user typed into the input box searching for collabs
+ * @param {number} maxSuggestions Maximum number of items to suggest/return, default 3
+ * @param {number} minCharacters Minimum number of characters before searching, default 3
+ * @returns {Array<SuggestedCollab>, Array<C>}
+ * [0] Suggested collaborators in recommended order
+ * [1] The rest of the contacts not suggested
+ */
+export default function computeSuggestedCollabs<C: { id: number | string }>(
```
priyajeet: Move the export-ing to the bottom of the file for consistency with other files.

src/features/unified-share-modal/utils/computeSuggestedCollabs.js
```diff
@@ -0,0 +1,67 @@
...
+    contacts: Array<C>,
+    suggestedCollabLookup: SuggestedCollabLookup,
+    inputValue: string,
+    maxSuggestions: number = 3,
+    minCharacters: number = 3,
+    maxGaps: number = 2,
+): [Array<SuggestedCollab>, Array<C>] {
+    const contactIdSet = new Set(contacts.map(c => c.id.toString()));
```
priyajeet: Does it make sense to use Set as the input type for this function? Can the parent component be changed to use Set or is it something thats only needed here internally?

src/features/unified-share-modal/utils/computeSuggestedCollabs.js
```diff
@@ -0,0 +1,67 @@
...
+                      return false;
+                  }
+                  const nameMatches = fuzzySearch(inputValue, option.name || '', minCharacters, maxGaps);
+                  const emailAddress = option.email || '';
+                  const emailAlias = emailAddress.substring(0, emailAddress.indexOf('@'));
+                  const emailMatches = inputValue.length >= minCharacters && emailAlias.indexOf(inputValue) > -1;
+                  return nameMatches || emailMatches;
+              });
+
+    const sortSuggestionList = list => {
```
priyajeet: Can this function move out on its own, does it need to be inlined into the parent function?

src/features/unified-share-modal/utils/computeSuggestedCollabs.js
```diff
@@ -0,0 +1,67 @@
...
+    inputValue: string,
+    maxSuggestions: number = 3,
+    minCharacters: number = 3,
+    maxGaps: number = 2,
+): [Array<SuggestedCollab>, Array<C>] {
+    const contactIdSet = new Set(contacts.map(c => c.id.toString()));
+    // $FlowFixMe
+    const suggestedCollabs: Array<SuggestedCollab> = Object.values(suggestedCollabLookup);
+    const exactMatches: Array<SuggestedCollab> = suggestedCollabs.filter(
+        sc => sc.id && contactIdSet.has(sc.id.toString()),
```
priyajeet: a lot of id.toString() is being done. Can we change such that the incoming objects always has string ids from the absolute beginning in the component hierarchy? Ideally the backend or parent app's network call should be sending it that way.

src/features/unified-share-modal/utils/computeSuggestedCollabs.js
```diff
@@ -0,0 +1,67 @@
...
+    const exactMatchIds = exactMatches.map(c => c.id.toString());
+
+    const suggestedFuzzyMatches =
+        exactMatches.length >= maxSuggestions
+            ? []
+            : suggestedCollabs.filter((option: SuggestedCollab) => {
+                  if (!inputValue || exactMatchIds.includes(option.id.toString())) {
+                      return false;
+                  }
+                  const nameMatches = fuzzySearch(inputValue, option.name || '', minCharacters, maxGaps);
```
priyajeet: Should fuzzySearch auto default name to empty string instead of passing it in the callers?

src/features/unified-share-modal/utils/computeSuggestedCollabs.js
```diff
@@ -0,0 +1,67 @@
...
+                  const emailAddress = option.email || '';
+                  const emailAlias = emailAddress.substring(0, emailAddress.indexOf('@'));
+                  const emailMatches = inputValue.length >= minCharacters && emailAlias.indexOf(inputValue) > -1;
```
priyajeet: prefer includes over indexOf

src/features/unified-share-modal/utils/computeSuggestedCollabs.js
```diff
@@ -0,0 +1,67 @@
...
+    const suggestedCollabs: Array<SuggestedCollab> = Object.values(suggestedCollabLookup);
+    const exactMatches: Array<SuggestedCollab> = suggestedCollabs.filter(
+        sc => sc.id && contactIdSet.has(sc.id.toString()),
+    );
+    const exactMatchIds = exactMatches.map(c => c.id.toString());
+
+    const suggestedFuzzyMatches =
+        exactMatches.length >= maxSuggestions
+            ? []
+            : suggestedCollabs.filter((option: SuggestedCollab) => {
```
priyajeet: Might be a bit more readable (and independently testable) if this whole block can be extracted out as a function of its own

src/features/unified-share-modal/utils/computeSuggestedCollabs.js
priyajeet: or do you even need to search at all if option.name is non existent?

src/features/unified-share-modal/utils/computeSuggestedCollabs.js
```diff
@@ -0,0 +1,67 @@
...
+        });
+    };
+
+    // combine both lists preferring exact id matches over fuzzy matches
+    const suggestedResult = [...sortSuggestionList(exactMatches), ...sortSuggestionList(suggestedFuzzyMatches)].slice(
+        0,
+        maxSuggestions,
+    );
+
+    const otherResults = contacts.filter(c => !suggestedResult.map(suggestion => suggestion.id).includes(c.id));
```
priyajeet: prefer full words over characters whenever possible. `c` --> `contact`

priyajeet: Probably break out the map part to its own const above. Easier to read/debug and put breakpoints.

src/features/unified-share-modal/utils/computeSuggestedCollabs.js
stevenzyang: It seems like the api returns number for all the places I've seen. Is there a standard for what format we store the ids in on the frontend? I could change the existing contact id type to just number

src/features/unified-share-modal/utils/computeSuggestedCollabs.js
stevenzyang: done

src/features/unified-share-modal/utils/computeSuggestedCollabs.js
stevenzyang: done

src/features/unified-share-modal/utils/computeSuggestedCollabs.js
stevenzyang: I'll add a default in fuzzySearch

src/features/unified-share-modal/utils/computeSuggestedCollabs.js
stevenzyang: There's a lot of variables referenced in the function so I don't think it's worth it

src/features/unified-share-modal/utils/computeSuggestedCollabs.js
stevenzyang: done

src/features/unified-share-modal/utils/computeSuggestedCollabs.js
stevenzyang: done

stevenzyang: done

src/features/unified-share-modal/ContactsField.js
```diff
@@ -90,22 +81,21 @@ class ContactsField extends React.Component<Props, State> {
...
-                    isExternalUser,
-                    text: name,
-                    type,
-                    value: email || id, // if email doesn't exist, contact is a group, use id
-                }));
+                );
 
             if (suggestedCollaborators) {
-                return this.addSuggestedContacts(fullContacts);
+                fullContacts = this.addSuggestedContacts(fullContacts);
```
booc0mtaco: Slightly prefer setting this new list into a separate variable, then merging it below, and mapping over the result, so it's more clear (and we can avoid having a variable that mutates).

Unless `.addSuggestedContacts` is a superset of what `fullContacts` is before this runs...

stevenzyang: The problem is that addSuggestedContacts not only adds contacts to the front but also removes contacts from the original list, so there's not really a way of avoiding mutation without duplicating code

src/features/unified-share-modal/flowTypes.js
```diff
@@ -45,6 +45,14 @@ export type contactType = {
     value?: number | string,
 };
 
+export type SuggestedCollab = contactType & {
```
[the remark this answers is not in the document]
stevenzyang: Per our discussion, they're not the same type and merging the 2 is nontrivial work beyond the scope of this PR

src/features/unified-share-modal/EmailForm.js
```diff
@@ -18,7 +18,7 @@ import IconGlobe from '../../icons/general/IconGlobe';
 
 import ContactsField from './ContactsField';
 import messages from './messages';
-import type { contactType as Contact, suggestedCollaboratorsType } from './flowTypes';
+import type { SuggestedCollabLookup, contactType as Contact } from './flowTypes';
```
booc0mtaco: 💯

PR #1933, opened 2020-02-12
refactor(style): add standard animation values with example usage

booc0mtaco:
updated `transform` animations in this PR

### DIFF
--- a/src/styles/constants/_animation.scss
+++ b/src/styles/constants/_animation.scss
@@ -1,1 +1,1 @@
-$bdl-transition-duration-base: .15s;
+$bdl-transitionDurationBase: .15s;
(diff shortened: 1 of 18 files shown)


boxcla:

src/styles/constants/_animation.scss
```diff
@@ -1 +1 @@
-$bdl-transition-duration-200ms: .2s;
+$bdl-transition-duration-base: .15s;
```
priyajeet: Since we are renaming this, might as well convert to camelcase SUIT style, eg: is in the CSS conventions doc

booc0mtaco: `$bdl-transitionDurationBase`?

src/styles/mixins/_animation.scss
```diff
@@ -0,0 +1,5 @@
+@mixin bdl-transition($time-length, $properties...) {
```
priyajeet: should time-length be defaulted to the base value? Though then you would have to change args order.

PR #1939, opened 2020-02-15
feat(icons): add monotone variant for ItemIcon

alexkrolick:
Screen Shot 2020-02-14 at 18 58 17

Screen Shot 2020-02-14 at 18 58 23

Screen Shot 2020-02-14 at 18 58 26

Note: this creates a new Monotone ItemIcon instead of adding a prop to the existing one because the default asset sizes are different and they are not usually used in the same place as the colorful icons, so adding 2x as many imports would bloat the module


boxcla:

alexkrolick:
> Curious why there is a js.flow file and a .tsx file? Is this for backwards compatibility?

Yes

src/icons/icons.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+    'spreadsheet',
+    'text',
+    'threed',
+    'vector',
+    'video',
+    'word-document',
+    'zip',
+    'folder-collab',
+    'folder-external',
+    'folder-plain',
```
krapspark: Seems weird that the folder types alone are not listed in alphabetical order

alexkrolick: I put the folders at the end and "default" icon at the beginning instead of strict alphabet order to make it find them instead of getting lost in the filetype list

PR #1940, opened 2020-02-15
docs(storybook): usability improvements

alexkrolick:
- move "knobs" panel to first position since it is used and "actions" is not
- add padding to canvas panel so components don't hit the edge
- ~~allow mdx docs (react components in markdown)~~ **removed**
- parse the "notes" option for stories as the default description block for the docs pages (these are written in standard markdown and appear in the Notes panel in the Canvas; now the same text will appear at the top of the Docs tab).


boxcla:

alexkrolick:
@jstoffan updated snapshots, things got bumped by the padding amount

jstoffan:
@sahiga, any concerns with the added padding? I think we had discussed it during the initial investigation of visual regression testing.

alexkrolick:
> @sahiga, any concerns with the added padding? I think we had discussed it during the initial investigation of visual regression testing.

Note that it currently takes a huge snapshot (full page) regardless of component dimensions, maybe we could retarget to a scaled container?

alexkrolick:
@sahiga rebasing off master

sahiga:
@jstoffan @alexkrolick The additional padding seems fine to me. I agree that we should use start using scaled containers, since most of the components are much smaller than the default snapshot viewport. Would you be able to address that in this PR, @alexkrolick?

alexkrolick:
> . I agree that we should use start using scaled containers, since most of the components are much smaller than the default snapshot viewport. Would you be able to address that in this PR, @alexkrolick?

No, I am not familiar with the BoxVisualTestUtils API

PR #1942, opened 2020-02-18
fix(nudge): update to use border radius style variable

JohnnyB83:


boxcla:
Hi @JohnnyB83, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

src/components/nudge/Nudge.scss
```diff
@@ -9,7 +9,7 @@
     padding: 25px;
     background-color: $white;
     border: solid 1px $bdl-gray-02;
-    border-radius: 30px;
+    border-radius: 4px;
```
alexkrolick: https://github.com/box/box-ui-elements/blob/master/src/styles/constants/_layout.scss#L9

JohnnyB83: thanks, resolved in latest commit

laughnan:
this probably has snapshot implications for storybook that we should update.

PR #1944, opened 2020-02-18
feat(classification): Configurable colors for Classification components

jpfranco:
This update will allow passing in `fillColor` and `strokeColor` props to Classification components in order to allow customizing the badge and icon colors. 

The current default color will be used when the props are not passed in, so this change should be backwards compatible.

Screen Shot 2020-02-18 at 11 42 07 AM

Screen Shot 2020-02-18 at 11 43 10 AM


boxcla:

src/features/classification/Classification.js
```diff
@@ -19,12 +19,14 @@ type Props = {
     controls?: Controls,
     controlsFormat?: ControlsFormat,
     definition?: string,
+    fillColor?: string,
```
priyajeet: Does it make sense to make it a color object if there is possibility of more color customization? Or if its just going to be these two, then I guess this is fine

priyajeet: Also, if there is going to be specific colors only, then maybe the prop should be like theme="purple". And internally you map purple to the colors

priyajeet: cc @feixuanli

jpfranco: We're planning to have a color map but in this case I'm trying to keep the Classification components agnostic to it since it might be moved to a more central location (in order to make it available to mobile, desktop, etc.). Eventually de API itself could be returning the colors so we need the components to be flexible.

jpfranco: But yes, we're not planning to support more colors beyond the fill and stroke

src/icons/general/IconSecurityClassificationSolid.js
```diff
@@ -3,44 +3,38 @@ import * as React from 'react';
...
-            />
-            <path
-                d="M23,11 L23,19"
-                stroke={white}
-                strokeWidth={strokeWidth}
-                strokeLinecap="round"
-                strokeLinejoin="round"
-                fill="none"
-            />
+        <AccessibleSVG className={classes} height={height} title={title} viewBox="0 0 10 13" width={width}>
```
Sunyang730: The icon size should be 32 x 32. I am not sure if we have something small but should follow the bdl standard.

Sunyang730: We have 32 x 32, 16 x 16, and 20 x 20

Sunyang730: Chat in person, JP will make another PR once got reply from design.

jpfranco: Icon updated

src/features/classification/Classification.js
feixuanli: Agreed. I ll include the color map in my PR with radio buttons.

src/features/classification/security-controls/SecurityControlsItem.js
```diff
@@ -11,20 +11,37 @@ import './SecurityControlsItem.scss';
...
         <li className="bdl-SecurityControlsItem">
             {shouldRenderIcon && (
-                <IconSecurityClassificationSolid color={bdlYellorange} height={11} width={11} strokeWidth={3} />
+                <span className="bdl-SecurityControlsItem-icon">
+                    <IconSecurityClassificationSolid
+                        fillColor={fillColor}
+                        strokeColor={strokeColor}
+                        height={ICON_SIZE}
+                        width={ICON_SIZE}
+                        strokeWidth={3}
```
feixuanli: Are we removing strokeWidth ?

PR #1945, opened 2020-02-18
fix(nudge): style modifications per brand

laughnan:
nudge-stories-test-tsx-components-nudge-nudge-looks-visually-correct-when-using-story-components-nudge-regular-1-snap

* Update styles per Brand
* Take a new Storybook snapshot


boxcla:

PR #1946, opened 2020-02-18
fix(targeting): fix tooltip position in dropdown menu

azuruce:


boxcla:

azuruce:
after fix
image

before fix
image

src/components/tooltip/Tooltip.tsx
```diff
@@ -90,6 +90,7 @@ export interface TooltipProps {
 
 type State = {
     isShown: boolean;
+    hasRendered: boolean;
```
DanDeMicco: can you add a comment as to why this is needed for others to understand?

ittaishay:
Approved, but please address @DanDeMicco's comments

src/components/tooltip/Tooltip.tsx
azuruce: ouch, just saw this. Will add a comment

PR #1949, opened 2020-02-19
chore(classification): Update default security icon stroke width

jpfranco:
The updated icon from the previous PR resulted in a lighter stroke than the original. Set a default stroke width on the icon in order to keep better contrast.

Note: This refers to the icon at the bottom of the screenshots, not the one on the badge.

Before:
Before

After:
After


boxcla:

PR #1950, opened 2020-02-19
feat(label-pill): size options and composition fixes

alexkrolick:
- add size (large, regular) prop
- better support for react and css composition:
  - custom classnames allowed for all subcomponents
  - lower specificity for style overrides


boxcla:

PR #1951, opened 2020-02-19
feat(avatar): size options and composition fixes

alexkrolick:
- add "small" size
- prevent flexbox parents from distorting shape of avatar


boxcla:

PR #1953, opened 2020-02-20
build(scss): scss build checks

alexkrolick:
We should do followup work to make sure the stylesheets for individual components compile (the ones that aren't pulled into a the /elements webpack build can fail without CI noticing). One way to do that is by running Storybook in CI instead of only in the deploy script.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
         "build:prod:npm": "BABEL_ENV=production OUTPUT=dist LANGUAGE=en-US REACT=true yarn build:prod:dist",
         "build:prod:storybook": "LANGUAGE=en-US REACT=true BROWSERSLIST_ENV=production BABEL_ENV=development NODE_ENV=development build-storybook -c .storybook -o styleguide/storybook",
         "build:sync": "LANGUAGE=en-US BABEL_ENV=development NODE_ENV=development RSYNC=true webpack --config scripts/webpack.config.js --mode development",
-        "build:variables": "cat `find src/styles/constants -name '*.scss'` > src/styles/variables.scss | node ./scripts/build-style-vars.js src/styles/variables.scss && rm src/styles/variables.scss",
+        "build:variables": "cat `find -s src/styles/constants -name '*.scss'` > src/styles/variables.scss | node ./scripts/build-style-vars.js src/styles/variables.scss && rm src/styles/variables.scss",
         "build:assets": "./scripts/asset-import/cli.js",
         "clean": "rm -rf dist es i18n/json i18n/*.js reports styleguide",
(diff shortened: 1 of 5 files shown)


boxcla:

package.json
```diff
@@ -54,7 +54,7 @@
         "build:prod:npm": "BABEL_ENV=production OUTPUT=dist LANGUAGE=en-US REACT=true yarn build:prod:dist",
         "build:prod:storybook": "LANGUAGE=en-US REACT=true BROWSERSLIST_ENV=production BABEL_ENV=development NODE_ENV=development build-storybook -c .storybook -o styleguide/storybook",
         "build:sync": "LANGUAGE=en-US BABEL_ENV=development NODE_ENV=development RSYNC=true webpack --config scripts/webpack.config.js --mode development",
-        "build:variables": "cat `find src/styles/constants -name '*.scss'` > src/styles/variables.scss | node ./scripts/build-style-vars.js src/styles/variables.scss && rm src/styles/variables.scss",
+        "build:variables": "cat `find -s src/styles/constants -name '*.scss'` > src/styles/variables.scss | node ./scripts/build-style-vars.js src/styles/variables.scss && rm src/styles/variables.scss",
```
alexkrolick: forces sort order to be lexographic, this resolves inconsistency in sort between mac and linux

scripts/build-style-vars.js
```diff
@@ -60,4 +60,7 @@ async function main() {
     }
 }
 
-main();
+main().catch(e => {
```
alexkrolick: make sure errors cause a nonzero exit code

booc0mtaco: This should apply to any of our node scripts. We should either:

- use the `sync` versions of any `fs` commands (which would block until completed or error out) and handle errors procedurally
- preserve any async calls as-is, but make sure the script itself contains the proper `.catch` or try-catch block(s)

We could use this PR to address them all. Would require a small audit.

package.json
booc0mtaco: The BSD version of [`find`](https://www.freebsd.org/cgi/man.cgi?find(1)) is **much** more sophisticated than gnu/linux [`find`](http://man7.org/linux/man-pages/man1/find.1.html)...

So `-s` isn't gonna work if circleCI uses linux unfortunately. 

An in-place solution would be to pipe into [`sort`](http://man7.org/linux/man-pages/man1/sort.1.html). Both macOS's BSD version and linux's version support `-n` for numeric sort.

`find src/styles/constants -name '*.scss' | sort -n` or similar should do the trick.

cc @j3tan

scripts/build-style-vars.js
j3tan: interesting, I assumed thrown errors would cause a non-zero error, but good to know

package.json
j3tan: good eye @booc0mtaco !

booc0mtaco:
Probably time to close this one

PR #1959, opened 2020-02-21
feat(radio-group): Add support for custom radio button components

jpfranco:
This update will add support for using `RadioGroup` with custom radio button components, so long as they accept and implement the same props as `RadioButton`. This will allow us to support use cases in which a single value needs to be selected, but the appearance and behavior of the UI itself is slightly different from `RadioButton` (e.g. color picker radio buttons).

This change is backward compatible because our existing usages of `RadioGroup` all declare `RadioButton` children, so the only difference is that these will now be cloned instead of replaced with a new `RadioButton` declaration.

### DIFF
--- a/src/components/radio/RadioGroup.tsx
+++ b/src/components/radio/RadioGroup.tsx
@@ -1,7 +1,5 @@
 import * as React from 'react';
 
-import RadioButton from './RadioButton';
-
 export interface RadioGroupProps {
     children: Array<React.ReactElement> | React.ReactElement;
     className: string;
@@ -50,9 +48,12 @@ class RadioGroup extends React.Component<RadioGroupProps, RadioGroupState> {
         return (
             <div className={`radio-group ${className}`} onChange={this.onChangeHandler}>
                 {React.Children.map(children, (radio: React.ReactElement) => {
-                    const { value, ...rest } = radio.props;
+                    const { value } = radio.props;
 
-                    return <RadioButton isSelected={value === stateValue} name={name} value={value} {...rest} />;
+                    return React.cloneElement(radio, {
+                        name,
+                        isSelected: value === stateValue,
+                    });
                 })}
             </div>
         );
--- a/src/components/radio/__tests__/RadioGroup.test.tsx
+++ b/src/components/radio/__tests__/RadioGroup.test.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import * as React from 'react';
 import { mount } from 'enzyme';
 import sinon from 'sinon';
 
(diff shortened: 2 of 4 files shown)


src/components/radio/RadioGroup.tsx
```diff
@@ -50,9 +48,12 @@ class RadioGroup extends React.Component<RadioGroupProps, RadioGroupState> {
         return (
             <div className={`radio-group ${className}`} onChange={this.onChangeHandler}>
                 {React.Children.map(children, (radio: React.ReactElement) => {
-                    const { value, ...rest } = radio.props;
+                    const { value } = radio.props;
 
-                    return <RadioButton isSelected={value === stateValue} name={name} value={value} {...rest} />;
+                    return React.cloneElement(radio, {
+                        name,
+                        isSelected: value === stateValue,
```
feixuanli: Would this miss other props ?

jpfranco: `cloneElement` will preserve the original element's props. We only need to pass the ones that are set dynamically by `RadioGroup`, which will be shallow merged with the rest.

jpfranco:
@Sunyang730 in this case I didn't add any new storybook visual regression tests since the update in itself does not cause any visual changes.

src/components/radio/__tests__/RadioGroup.test.tsx
```diff
@@ -11,20 +11,26 @@ describe('components/radio/RadioGroup', () => {
...
+                description="radio1desc"
+                label="Radio Button 1"
+                value="radio1"
+            />,
+            <RadioButton key="2" label="Radio Button 2" value="radio2" />,
+            <RadioButton key="3" description="radio3desc" label="Radio Button 3" value="radio3" />,
+            <RadioButton key="4" label="Radio Button 4" value="radio4" />,
+        ];
+
+        return mount(
```
sahiga: Can we use `shallow` here?

src/components/radio/stories/RadioGroup.stories.tsx
```diff
@@ -16,6 +16,41 @@ export const example = () => (
...
+export const withCustomRadioButtonComponent = () => {
+    const CustomRadioButton = ({ isSelected, label, name, value }: RadioButtonProps) => (
+        <span style={{ marginRight: '15px', position: 'relative' }} title={String(label)}>
+            <span
+                style={{
+                    backgroundColor: isSelected ? bdlPurpleRain : bdlGray20,
+                    borderRadius: '50%',
+                    display: 'inline-block',
+                    height: '20px',
+                    left: '0px',
```
sahiga: Zero values don't have measurements, so this can just be `0`

src/components/radio/__tests__/RadioGroup.test.tsx
jpfranco: I think it might've been deliberate to use `mount` originally since the implementation depends heavily on the radio input element. I'm not sure if there's a good way to migrate some of these existing tests to `shallow`

PR #1964, opened 2020-02-24
feat(upload): Allow file description option for upload

danielkim328:


src/api/uploads/PlainUpload.js
```diff
@@ -77,6 +77,7 @@ class PlainUpload extends BaseUpload {
         const attributes = JSON.stringify({
             name: this.fileName,
             parent: { id: this.folderId },
+            description: this.fileDescription,
```
j3tan: Verify that this is not a breaking change and not dependent on a server api change?

danielkim328: @j3tan I verified in EUA, where this param is undefined, our upload does not breaking with these changes. Is there anywhere else I should be testing?

j3tan: sounds good, just wanted to make sure this was backwards compatible

src/api/uploads/BaseUpload.js
```diff
@@ -49,6 +51,7 @@ class BaseUpload extends Base {
         const attributes = {
             name: this.fileName || name,
             parent: { id: this.folderId },
+            description: this.fileDescription,
```
jstoffan: Does this need to be part of the preflight OPTIONS request? If so, why?

jstoffan: @danielkim328, can you comment? It looks like this field is also sent in the `preflightSuccessHandler`. Why is it sent twice?

danielkim328: Ideally we want to verify the file submission is valid in preflight check, which should include whether description CAN be applied, such as valid length, or allow description etc

jstoffan: Okay. I'm not sure what "ideally" means in this case, but I'm not very familiar with this endpoint, so I'll take your word for it.

danielkim328: Well, preFlightRequest is made BEFORE the actual upload occurs to check whether the upload WILL be accepted. As I understand, preFlightRequest currently check only for file name conflicts, but 'ideally' it should also check if fileDescription can also be applied.

PR #1967, opened 2020-02-24
refactor(typescript): Migrate icons to TypeScript (part one)

sahiga:
This PR migrates all the icons from `AccessibleSVG` through `BoxTools` (alphabetically) to TypeScript.


src/icons/iconTypes.ts
```diff
@@ -0,0 +1,36 @@
+import { ReactElement } from 'react';
+
+export interface TwoTonedIcon {
```
booc0mtaco: As a note, we may be moving away from these types entirely (with the new assets, we won't have any that allow for two colors to be set.

This looks good, tho!

src/icons/box-tools/__tests__/IconBoxToolsInstall.test.tsx
```diff
@@ -1,4 +1,5 @@
 import React from 'react';
+import { shallow } from 'enzyme';
```
booc0mtaco: Hm, i am guessing TypeScript doesn't like magical imports?

src/icons/accessible-svg/AccessibleSVG.tsx
```diff
@@ -0,0 +1,41 @@
...
+    focusable?: boolean | 'false' | 'true' | 'auto' | undefined;
+    opacity?: number;
+}
+
+class AccessibleSVG extends React.Component<AccessibleSVGIconProps> {
+    id: string = uniqueId('icon');
+
+    render() {
+        const { children, title, ...rest } = this.props;
+        const titleID = `${this.id}-title`;
```
booc0mtaco: curious, do we need the `id: string`? if we were to put `uniqueId('icon')` in line 17, could we make this a functional component?

I don't think the change is needed now, since this is mostly a conversion, but wasn't sure if the class attribute was needed.

src/icons/box-tools/__tests__/IconBoxToolsInstall.test.tsx
sahiga: Yeah, TypeScript throws the "cannot find name 'shallow'" error without this line

src/icons/accessible-svg/AccessibleSVG.tsx
sahiga: Good idea! This seems like a minor change, so I'll see if I can include it in this PR

j3tan: well you don't want to generate a new unique id every render, so you'll have to keep it in state

sahiga: That makes sense. Let's keep this component as is for now. I'll add a note about this in the migration Box Note.

PR #1974, opened 2020-02-25
refactor(typescript): Migrate general icons (part two)

sahiga:
This PR migrates icons in icons/general (beginning with C or D) to TypeScript.


src/icons/general/IconCaretDown.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react';
+
+import AccessibleSVG from '../accessible-svg';
+
+import { Icon } from '../iconTypes';
+
+const IconCaretDown = ({ className = '', color = '#000', height = 6, title, width = 10 }: Icon) => (
```
booc0mtaco: I suppose we should be importing variables, and using `bdlBlack` here for the default color value, but we're moving away from color stuff, and this icon may already be replaced by one in the new set.

src/icons/general/IconChat.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from 'react';
+
+import AccessibleSVG from '../accessible-svg';
+
+import { Icon } from '../iconTypes';
+
+const IconChat = ({ className = '', color = '#979797', height = 16, title, width = 16 }: Icon) => (
```
booc0mtaco: Same comment as the other. Won't do each one but if we want to clean that up, it would be good to move in one direction, or ignore it.

src/icons/general/IconChevron.tsx
```diff
@@ -0,0 +1,51 @@
...
+    /** A string describing the size of the icon */
+    size?: string;
+    /** A string describing the thickness of the icon */
+    thickness?: string;
+}
+
+const IconChevron = ({
+    className = '',
+    color = '#000',
+    direction = DirectionType.UP,
```
booc0mtaco: For future reference, we won't have complex icon props like this in the future. any transforms and such would go thru CSS. Any exceptional cases would be good for discussion.

src/icons/general/IconCaretDown.tsx
sahiga: @booc0mtaco Yup, this one looks really similar to `Caret8`.  Would you recommend just importing `Caret8` into this icon, for backwards compatibility purposes?

Also, what's the timeline for deprecating these icons?

booc0mtaco: I would avoid that, so the old icons don't change under people.

No date on the deprecation yet. I wouldn't necessarily make any of the color variable changes here, which should keep it simple. If it's valuable, we can do that at a later step!

PR #1980, opened 2020-02-25
feat(usm): Add support for classification colors

jpfranco:
This update will enable classification colors on the USM badge. The usual orange color will be used by default, so this change should be backwards compatible.

Screen Shot 2020-02-24 at 5 13 00 PM


src/styles/constants/_colors.scss
```diff
@@ -79,7 +79,7 @@ $bdl-orange-20: #fce6d1 !default;
 // Purple Rains
 $bdl-purple-rain: #9f3fed !default;
 $bdl-purple-rain-50: #cf9ff6 !default;
-$bdl-purple-rain-20: #dad4f3 !default;
```
jpfranco: `#dad4f3` is actually Grimace 20

PR #1983, opened 2020-02-26
feat(readabletime): add support for uppercase timestamps

il10:
- adds a prop called uppercase to ReadableTime that returns a locale aware uppercase version
- refactored tests to reflect actual output


src/components/time/ReadableTime.js
```diff
@@ -51,20 +58,21 @@ const ReadableTime = ({
     } else if (showWeekday) {
         // e.g. Monday, Oct 5, 2018
         dateMessage = messages.eventTimeWeekdayLong;
-        weekday = <FormattedDate value={timestamp} weekday="long" />;
+        weekday = intl.formatDate(timestamp, { weekday: 'long' });
```
il10: using intl formatting here in order to uppercase string - not able to achieve this using the `Formatted` components itself

src/components/time/__tests__/ReadableTime.test.js
```diff
@@ -32,114 +36,53 @@ describe('components/time/ReadableTime', () => {
...
+        ruIntl.locale = 'ru';
+        const wrapperUppercase = mount(
+            <ReadableTime intl={ruIntl} timestamp={withinRelativeThresholdAhead} uppercase />,
+        );
+        const wrapperLowercase = mount(
+            <ReadableTime intl={ruIntl} timestamp={withinRelativeThresholdAhead} uppercase={false} />,
+        );
+
+        expect(wrapperUppercase.text()).toEqual(wrapperLowercase.text());
+    });
```
ehoogerbeets: Maybe add a test case for Japanese as well to make sure nothing in the date actually changes when you upper-case it?

src/components/time/ReadableTime.js
ehoogerbeets: 👍

src/components/time/__tests__/ReadableTime.test.js
```diff
@@ -32,114 +36,53 @@ describe('components/time/ReadableTime', () => {
...
         },
     );
 
     test('should use default relative threshold if not provided', () => {
         const wrapper = shallow(<ReadableTime intl={intl} timestamp={withinRelativeThresholdAhead} />);
         expect(wrapper).toMatchSnapshot();
     });
+    test('should not uppercase locales that do not have uppercase grammar', () => {
+        const ruIntl = createIntl({ locale: 'ru', messages: {} });
+        // [React Intl] Missing locale data for locale: "ru". Using default locale: "en" as fallback.
```
il10: @ehoogerbeets https://github.com/formatjs/react-intl/blob/b207d805a9e5609983ac1b2ca8f9708030e12337/docs/Upgrade-Guide.md#migrate-to-using-native-intl-apis

not fully running this bc node doesn't ship w/ the full ICU package

src/components/time/__tests__/ReadableTime.test.js
il10: @ehoogerbeets Can't test this locally: https://github.com/formatjs/react-intl/blob/b207d805a9e5609983ac1b2ca8f9708030e12337/docs/Upgrade-Guide.md#migrate-to-using-native-intl-apis

not fully running this bc node doesn't ship w/ the full ICU package

il10: @ehoogerbeets  We added the package. It should work on node 12 which is what we use in nvmrc. This should allow full internationalization in the test environment

src/components/time/__tests__/ReadableTime.test.js
```diff
@@ -85,4 +84,16 @@ describe('components/time/ReadableTime', () => {
 
         expect(wrapperUppercase.text()).toEqual(wrapperLowercase.text());
     });
+    test('CJK languages should look the same for uppercase and lowercase (e.g. japanese)', () => {
```
il10: adding test case for japanese

ehoogerbeets: 👍

package.json
```diff
@@ -222,6 +222,7 @@
         "flow-typed": "^2.6.2",
         "form-serialize": "^0.7.2",
         "formik": "^2.0.3",
+        "full-icu": "^1.3.1",
```
ehoogerbeets: Will this fix work in Japanese and other languages if the `full-icu` package is not installed? If not, then we need to add this to peer dependencies as well as the dev dependencies and add a message to the merge commit message like "WARNING: breaking change. This adds a new peer dependency on full-icu" so that BUIE customers will know. Also, you will need to submit a PR for adding `full-icu` to the dependencies of EUA and admin console so they don't break when they pick up the latest BUIE.

il10: The code will still work for Japanese and other languages if the full-icu package is not installed. The `full-icu` package is being added specifically for the test environment.

alexkrolick: Yep, it's just needed for the Intl global in Node. In the browser, the functionality is polyfilled or already available globally.

ehoogerbeets: ok cool

PR #1988, opened 2020-02-27
fix(tooltip): remove implementation in tooltip flow files

azuruce:


src/components/tooltip/Tooltip.js.flow
```diff
@@ -74,212 +18,29 @@ type Props = {
...
+    isDisabled?: boolean,
     /** Forces the tooltip to be shown or hidden (useful for errors) */
     isShown?: boolean,
     /** Whether to add tabindex=0.  Defaults to `true` */
     isTabbable?: boolean,
     /** Function called if the user manually dismisses the tooltip - only applies if showCloseButton is true */
     onDismiss?: () => void,
     /** Where to position the tooltip relative to the wrapped component */
-    position: Position,
+    position?: Position,
```
mickr: Shouldn't this be a required property of any tooltip

sahiga: `position` is already defined as part of the `defaultProps` in Tooltip.tsx, so components that use `Tooltip` do not have to provide it as a prop

sahiga:
@azhang, can you keep the `defaultProps` definition in the Flow component?
```
class Tooltip extends React.Component<Props> {
    static defaultProps = {
        constrainToScrollParent: false,
        constrainToWindow: true,
        isDisabled: false,
        position: TOP_CENTER,
        theme: DEFAULT_THEME,
    };
}
```

azuruce:
@sahiga can you approve?

PR #1989, opened 2020-02-27
fix(a11y): Fix a11y bug for error showing up as tooltip

jeeyi:
When text input validation fails, the error message is set to have role="alert".
This may interrupts the screen reader from reading the focused element.


src/components/tooltip/Tooltip.tsx
```diff
@@ -231,6 +231,9 @@ class Tooltip extends React.Component<TooltipProps, State> {
 
         if (showTooltip) {
             componentProps['aria-describedby'] = this.tooltipID;
+            if (theme === TooltipTheme.ERROR) {
+                componentProps['aria-errormessage'] = this.tooltipID;
```
tjgupta: Hmm, so wouldn't this still need to be on the input? The input is the source element that is in the error state, not the tooltip itself, and with this change now this tooltipID contains the target of the `aria-errormessage`?

jeeyi: Yes. `aria-errormessage` along with `aria-describedby` is on the input. It references the tooltip that contains the actual error message text `Field Required.`. 

Originally, it was set to the `errorMessageID` from `TextInput.js` which had the same error message text. Now that this is removed from `TextInput.js`, I changed it to set it from the `Tooltip.tsx` where I have the tooltip ID.

src/components/tooltip/Tooltip.tsx
```diff
@@ -266,7 +269,7 @@ class Tooltip extends React.Component<TooltipProps, State> {
             <TetherComponent ref={this.tetherRef} {...tetherProps}>
                 {React.cloneElement(React.Children.only(children) as React.ReactElement, componentProps)}
                 {showTooltip && (
-                    <div className={classes} id={this.tooltipID} role="tooltip">
+                    <div className={classes} id={this.tooltipID} role="tooltip" aria-live="polite">
```
tjgupta: Technically I don't think `aria-live` is supposed to be used on elements that are dynamically added/removed from the DOM. 

From MDN (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions): 
"Dynamically adding an element with an aria-live attribute or specialized role to the document also won't result in any announcement by assistive technologies (as, at that point, the browser/assistive technologies are not aware of the live region yet, so cannot monitor it for changes).

Always make sure that the live region is present in the document first, and only then dynamically add/change any content."

But if this makes it read it out, then 🤷‍♂ ...

src/components/tooltip/Tooltip.tsx
tjgupta: Ah yep I see now. Misunderstood where `componentProps` were getting used.

src/components/tooltip/Tooltip.tsx
jeeyi: I tested without `aria-live="polite"` on the tooltip and it fails ChromeVox from reading the tooltip. NVDA and JAWS work without `aria-live="polite"`.

PR #1990, opened 2020-02-27
feat(tooltip): add custom position for tooltip

azuruce:


src/components/tooltip/Tooltip.tsx
```diff
@@ -207,7 +213,7 @@ class Tooltip extends React.Component<TooltipProps, State> {
         const showTooltip = this.isShown();
 
         const withCloseButton = showCloseButton && isControlled;
-        const tetherPosition = positions[position];
+        const tetherPosition = typeof position === 'string' ? positions[position] : position;
```
laughnan: `[question-not-feedback]` is there a reason we support `string` and `array`?

src/components/tooltip/__tests__/Tooltip.test.tsx
```diff
@@ -162,6 +163,21 @@ describe('components/tooltip/Tooltip', () => {
...
+                targetAttachment: TetherPosition.BOTTOM_RIGHT,
+            };
+            const wrapper = shallow(
+                <Tooltip position={customPosition} text="hi">
+                    <button />
+                </Tooltip>,
+            );
+
+            expect(wrapper.prop('attachment')).toEqual('top left');
+            expect(wrapper.prop('targetAttachment')).toEqual('bottom right');
```
sahiga: Can we use the constants from `TetherPosition` here as well?

azuruce: hi, I look at it again, since it is in the test, we can retain the hard coded string. see
https://medium.com/@erik.sacre/on-tdd-literals-versus-constants-in-tests-8ac992c7873f

src/components/tooltip/Tooltip.tsx
azuruce: string was previously supported and I can't drop it. The custom position has 81 possibilities and I think using array of two items each has 9 possibility is better

booc0mtaco:
Most of my concerns are to do with the amount of indirection in these props, but we are inheriting that from the existing component. I don't have any strong objections, but do think there's some re-work to do with this at some point.

src/components/tooltip/Tooltip.tsx
```diff
@@ -25,38 +26,43 @@ export enum TooltipPosition {
     TOP_RIGHT = 'top-right',
 }
 
+export type TooltipCustomPostion = {
```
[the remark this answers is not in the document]
azuruce: fixed

azuruce:
Agree, I didn't create those known positions. It is pre-existing. I just added custom position to make sure we can position tooltips more flexibly. I can't remove enum b/c it is used all over EUA and we need to do a major release to remove feature.
> Why are we doing this limited set of known positions? Presumably, if we had enums for the vertical (top, middle, bottom) positions, and horizontal (left, center, right) positions, people could always use CustomPosition and not bother with the enum. Example:
> 
> ```
> const customPosition = {
>     attachment: `${TetherVerticalPosition.TOP} ${TetherHorizontalPosition.LEFT}`
>     targetAttachment: `${TetherVerticalPosition.BOTTOM} ${TetherHorizontalPosition.RIGHT}`
> };
> <Tooltip position={customPosition} text="test"><SomeAnchorComponent /></Tooltip>
> ```
> 
> If the enum is useful (example, the component only supports these eight known combinations), then I think it would be good to separate the partsof each `TetherPosition` value into its pieces. Having a variable name that maps to two independent variable values is an anti-pattern.
> 
> This might seem verbose, but would allow us to support any required layouts without further changes to `Tooltip`. And if we are worried about people mis-using the positions, we could verify the format in the component prop and throw meaningful warning messages.

azuruce:
screen shot was posted at the PR that use the tooltip https://git.dev.box.net/Box/EndUserApp/pull/4809
https://git.dev.box.net/storage/user/1053/files/eeb6b680-59f6-11ea-84da-edd973869c7b
image

azuruce:
I renamed PR to have the same title as commit message

azuruce:
@booc0mtaco Hi Andrew, I take your response as this is approvable but at some point, we need to refactor it. Can you approve it?

PR #1991, opened 2020-02-28
test(utils): change token service test to swap jest mock after run

booc0mtaco:


src/utils/__tests__/TokenService.test.js
```diff
@@ -127,9 +127,11 @@ describe('util/Tokenservice', () => {
 
     describe('getReadTokens()', () => {
         test('should call Tokenservice.getReadToken', () => {
```
booc0mtaco: It might be a good idea to remove this particular test entirely, since there might be some future implementation that does what we wish, and doesn't use `getReadToken`

mxiao6: > It might be a good idea to remove this particular test entirely, since there might be some future implementation that does what we wish, and doesn't use `getReadToken`

I don't think it's a good idea to decrease test coverage. In the future, if we don't use `getReadToken` anymore, we can remove the test together with the code, right? In this way the test coverage doesn't decrease.

mxiao6:
You may want to change the title to `test()`

src/utils/__tests__/TokenService.test.js
booc0mtaco: Because the second test requires a call to `.getReadToken` to succeed, the second test covers this particular line already (i.e., LN84 gets executed if the first param to `getReadTokens` is not "empty") . And since the above test mocks out the call, we aren't verifying anything about `.getReadToken`'s usage beyond the fact it was called. One improvement would be to mock the call to `.getReadToken` in such a way that you can confirm the number of calls *and* the return value in the second test, for instance.

the only condition that could cause `.getReadToken` to not be called is if `ids` is empty. This probably should be covered, actually (testing what happens if `getReadTokens` receives an empty string, an array with empty strings, etc.).

PR #1992, opened 2020-02-28
fix(printing): Fix print button on ContentPreview

mickr:
* Print button should not show for non-printable file-types


mickr:
Dependent upon https://github.com/box/box-content-preview/pull/1177

src/elements/content-preview/ContentPreview.js
```diff
@@ -686,6 +690,26 @@ class ContentPreview extends React.PureComponent<Props, State> {
...
+     * Returns whether a preview viewer supports printing
+     *
+     * @return {boolean}
+     */
+    handleCanPrint() {
+        /*
+        Setting this as default so older preview versions can still print documents without having the
+        default state have the print icon.
+         */
+        let canPrint = true;
```
ConradJChan: Would it make sense to default `canPrint` state to `true` then?

mickr: Talked about this off-line, setting it to true will flash the icon and then remove it for files that cannot print.

src/elements/content-preview/ContentPreview.js
```diff
@@ -686,6 +690,26 @@ class ContentPreview extends React.PureComponent<Props, State> {
...
+        const preview = this.getPreview();
+
+        if (preview && typeof preview.canPrint === 'function') {
+            canPrint = preview.canPrint();
+        }
+
+        return this.setState({ canPrint });
```
jstoffan: Can be simplified and `return` can be removed.

```
const preview = this.getPreview();

this.setState({
    canPrint: preview && (!preview.canPrint || preview.canPrint())
});
```

src/elements/content-preview/ContentPreview.js
```diff
@@ -686,6 +690,26 @@ class ContentPreview extends React.PureComponent<Props, State> {
         return !!showAnnotations && (this.canAnnotate() || hasViewAllPermissions || hasViewSelfPermissions);
     }
 
+    /**
+     * Returns whether a preview viewer supports printing
+     *
+     * @return {boolean}
```
jstoffan: Accurate? We generally leave JSDocs off of new code due to their tendency to become outdated.

src/elements/content-preview/ContentPreview.js
```diff
@@ -686,6 +690,26 @@ class ContentPreview extends React.PureComponent<Props, State> {
...
+     */
+    handleCanPrint() {
+        /*
+        Setting this as default so older preview versions can still print documents without having the
+        default state have the print icon.
+         */
+        let canPrint = true;
+        const preview = this.getPreview();
+
+        if (preview && typeof preview.canPrint === 'function') {
```
jstoffan: This method is only called in `onPreviewLoad`. Is it possible for preview to be falsy?

src/elements/content-preview/stories/ContentPreview.stories.js
```diff
@@ -0,0 +1,46 @@
+// @flow
```
jstoffan: Should these stories be introduced in a separate PR?

src/elements/content-preview/ContentPreview.js
mickr: Not in this scenario because the preview is loaded. But if this gets used outside of that scope it could because this.getPreview will return null if there isn't a preview instance or file, so I felt it better to be safe.

src/elements/content-preview/ContentPreview.js
```diff
@@ -686,6 +690,11 @@ class ContentPreview extends React.PureComponent<Props, State> {
         return !!showAnnotations && (this.canAnnotate() || hasViewAllPermissions || hasViewSelfPermissions);
     }
 
+    handleCanPrint() {
+        const preview = this.getPreview();
+        this.setState({ canPrint: !!preview && preview.canPrint && preview.canPrint() });
```
mxiao6: why the `preview.canPrint` is necessary

mxiao6: Should it be `!!preview.canPrint` then?

PR #1993, opened 2020-02-29
feat: add optional apiVersion parameter

omgrief:


ConradJChan:
And maybe the title should be `feat(api)`?

omgrief:
no longer needed

PR #2003, opened 2020-03-04
refactor(typescript): Migrate Checkbox and ErrorMask

sahiga:
This PR migrates Checkbox and ErrorMask to TypeScript and includes a new Storybook story for ErrorMask.


src/components/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,93 @@
...
+import uniqueId from 'lodash/uniqueId';
+import classNames from 'classnames';
+
+import CheckboxTooltip from './CheckboxTooltip';
+
+import './Checkbox.scss';
+
+export interface CheckboxProps {
+    className?: string;
+    /** Description to the checkbox */
```
booc0mtaco: These comments were originally for documentation styleguidist and would get rendered in the UI next to the props. Does storybook use these comments? If not, maybe we can remove them (... or move them to the related .stories.js)

sahiga: @booc0mtaco Yes, Storybook renders these comments as well. I will add equivalent comments to the `ErrorMask` component.

booc0mtaco: @sahiga good to know. Also worth mentioning that some of these files have a formatting bug where the interface type entries were alphabetized, while the comments describing them were not. Not sure if storybook supports putting the comment on the same line as the prop (or if we want to format like that)

sahiga: @booc0mtaco Storybook doesn't render inline comments. Thoughts on including the name of the prop in the comments, to ensure that they're always alphabetized in the same manner?

Also, what do the errors look like? I ran the linters, but didn't see anything.

booc0mtaco: @sahiga oh the "bug" is that we applied an alphabetizer to flow types some time ago, and it didn't sort the associated comments properly for some components. While converting we need to double check this as we go

sahiga: Ah, got it. Thanks for the info. I'll make sure the comments are properly alphabetized.

sahiga: @booc0mtaco I updated the doc comments. Can you review this PR again?

PR #2005, opened 2020-03-04
feat(tooltip): Tooltip can stop event propagation

azuruce:


src/components/dropdown-menu/DropdownMenu.js
```diff
@@ -47,25 +47,25 @@ class DropdownMenu extends React.Component<Props, State> {
     componentDidUpdate(prevProps: Props, prevState: State) {
         if (!prevState.isOpen && this.state.isOpen) {
             // When menu is being opened
-            document.addEventListener('click', this.handleDocumentClick, true);
-            document.addEventListener('contextmenu', this.handleDocumentClick, true);
+            document.addEventListener('click', this.handleDocumentClick, false);
```
mickr: `false` is the default param for addEventListener.

src/components/tooltip/Tooltip.tsx
```diff
@@ -275,7 +285,15 @@ class Tooltip extends React.Component<TooltipProps, State> {
             <TetherComponent ref={this.tetherRef} {...tetherProps}>
                 {React.cloneElement(React.Children.only(children) as React.ReactElement, componentProps)}
                 {showTooltip && (
-                    <div className={classes} id={this.tooltipID} role="tooltip" aria-live="polite">
+                    /* eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions */
```
jstoffan: We should avoid disabling these rules, as they are present for good reason. Can we find another way?

src/components/dropdown-menu/DropdownMenu.js
```diff
@@ -47,25 +47,25 @@ class DropdownMenu extends React.Component<Props, State> {
...
+            document.addEventListener('contextmenu', this.handleDocumentClick, false);
```
mickr: This needs to be tested pretty extensively in consuming apps to be sure this is playing nicely.

src/components/tooltip/__tests__/Tooltip.test.tsx
```diff
@@ -256,6 +256,29 @@ describe('components/tooltip/Tooltip', () => {
...
+            test('should stop event propagation as configured', () => {
+                const wrapper = shallow(
+                    <Tooltip isShown text="hi" shouldStopEventPropagation={shouldStopEventPropagation}>
+                        <button />
+                    </Tooltip>,
+                );
+                const instance = wrapper.instance();
+                const stop = jest.fn();
+                const nativeStop = jest.fn();
+                wrapper.find('div').simulate(event, {
```
mickr: We tend not to use the simulate method and call the function directly.

src/components/tooltip/Tooltip.tsx
azuruce: Hi, see
https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-noninteractive-element-interactions.md
For the lint rule and recommended solution. I think the recommended solution is good to go except it modifies DOM structure so practically it might not be the choice.

src/components/tooltip/__tests__/Tooltip.test.tsx
azuruce: Hi Mike, do you have an alternative? what is the other way we do to test this?

src/components/dropdown-menu/DropdownMenu.js
azuruce: fixed

src/components/dropdown-menu/DropdownMenu.js
azuruce: agree
Created PR https://jira.inside-box.net/browse/MONEY-5514

src/components/tooltip/Tooltip.tsx
jstoffan: So it seems like there a few options to move forward. Can you investigate those and choose one as appropriate?

azuruce: Hi, Jeff,
yes, the choices I considered is 
1) write code to attach event handler to DOM that prevent propagation
2) attach onClick to current div
3) attach onClick to a new div created inside current div

The drawback for 3) is it creates a new div that means a lot more testing and potentially more unexpected work. The drawback for 1) is it is more complex is less canonical. It doesn't really create much of a benefit. End of day we are attaching an event listener to a DOM element and in this case we have to find the DOM element before we attach, so a lot of complexity for a benefit that I can't see much.

src/components/tooltip/__tests__/Tooltip.test.tsx
mickr: It is Mick, you can call the function directly on the instance.

src/components/tooltip/Tooltip.tsx
azuruce: changed to render a presentation div to capture the event if shouldStopEventPropagation is true

src/components/dropdown-menu/DropdownMenu.js
azuruce: add shouldCloseOnEventBubbling prop that if set to true, will close the dropdown only if event bubbles up instead of capturing the event in document.

src/components/tooltip/__tests__/Tooltip.test.tsx
azuruce: done

src/components/dropdown-menu/__tests__/DropdownMenu.test.js
```diff
@@ -450,129 +450,125 @@ describe('components/dropdown-menu/DropdownMenu', () => {
...
-        test('should add click and contextmenu listeners when opening menu', () => {
-            const wrapper = mount(
-                <DropdownMenu>
-                    <FakeButton />
-                    <FakeMenu />
-                </DropdownMenu>,
-            );
-
-            const instance = wrapper.instance();
+    describe.each([[false], [true]])(
```
jstoffan: This is difficult to read. Let's leave the `describe` blocks as regular text and use `test.each`, instead. Or we can just write separate tests for the different prop values rather than iterating.

src/components/dropdown-menu/DropdownMenu.js
```diff
@@ -45,27 +46,29 @@ class DropdownMenu extends React.Component<Props, State> {
     };
 
     componentDidUpdate(prevProps: Props, prevState: State) {
+        const { shouldCloseOnEventBubbling } = this.props;
         if (!prevState.isOpen && this.state.isOpen) {
             // When menu is being opened
-            document.addEventListener('click', this.handleDocumentClick, true);
-            document.addEventListener('contextmenu', this.handleDocumentClick, true);
+            document.addEventListener('click', this.handleDocumentClick, !shouldCloseOnEventBubbling);
```
jstoffan: Can we invert this condition, give it a default value of `true`, and name it `useCapture` or something similar?

src/components/tooltip/Tooltip.tsx
```diff
@@ -146,6 +148,11 @@ class Tooltip extends React.Component<TooltipProps, State> {
         }
     };
 
+    handleTooltipEvent = (event: React.MouseEvent<HTMLElement> | React.KeyboardEvent<HTMLElement>) => {
+        event.stopPropagation();
+        event.nativeEvent.stopImmediatePropagation();
```
jstoffan: Is this needed? We're just trying to stop the click from bubbling up to the parents, right?

src/components/tooltip/Tooltip.tsx
```diff
@@ -274,16 +282,32 @@ class Tooltip extends React.Component<TooltipProps, State> {
...
-                )}
+                {showTooltip &&
+                    (shouldStopEventPropagation ? (
+                        <div className={classes} id={this.tooltipID} role="tooltip" aria-live="polite">
+                            <div
+                                role="presentation"
+                                onClick={this.handleTooltipEvent}
+                                onKeyPress={this.handleTooltipEvent}
+                            >
+                                {text}
```
jstoffan: It looks like the component has been duplicated. Can we move this to a separate method? Or move the wrapper to a separate component which returns a Fragment if the prop isn't enabled?

src/components/tooltip/__tests__/Tooltip.test.tsx
```diff
@@ -21,12 +21,13 @@ describe('components/tooltip/Tooltip', () => {
         sandbox.verifyAndRestore();
     });
 
-    describe('render()', () => {
+    describe.each([[true], [false]])('render() shouldStopEventPropagation=%o', shouldStopEventPropagation => {
```
jstoffan: This seems a little excessive. The test should be targeted on things that the prop affects.

src/components/tooltip/Tooltip.tsx
```diff
@@ -274,16 +282,32 @@ class Tooltip extends React.Component<TooltipProps, State> {
...
-                        {withCloseButton && (
-                            <PlainButton className="tooltip-close-button" onClick={this.closeTooltip}>
-                                <IconClose className="bdl-Tooltip-iconClose" width={14} height={14} />
-                            </PlainButton>
-                        )}
-                    </div>
-                )}
+                {showTooltip &&
+                    (shouldStopEventPropagation ? (
+                        <div className={classes} id={this.tooltipID} role="tooltip" aria-live="polite">
```
jstoffan: It looks like this node is the same in both conditions, as well.

jstoffan:
Should the dropdown menu change be a separate PR, @azuruce? It seems unrelated to the tooltip change.

src/components/tooltip/Tooltip.tsx
azuruce: I created a codeSandbox to show the difference. We need both.
https://codesandbox.io/s/stoppropagation-eybij

jstoffan: I see. Since we're using `onClick` on the node, React is delegating the event to the document, which requires `stopImmediatePropagation`. Makes sense.

src/components/tooltip/Tooltip.tsx
azuruce: will fix

src/components/tooltip/Tooltip.tsx
azuruce: will fix

src/components/tooltip/__tests__/Tooltip.test.tsx
azuruce: will fix

src/components/dropdown-menu/__tests__/DropdownMenu.test.js
azuruce: will fix

src/components/tooltip/Tooltip.js.flow
```diff
@@ -40,6 +40,8 @@ type Props = {
     onDismiss?: () => void,
     /** Where to position the tooltip relative to the wrapped component */
     position?: Position | CustomPosition,
+    /** Should click|keypress propagation be stopped */
+    shouldStopEventPropagation?: boolean,
```
azuruce: I will change this to "stopBubble" to echo with "useBubble" at DropDownMenu.

src/components/dropdown-menu/DropdownMenu.js
azuruce: rename the prop to useBubble and rename the corresponding pop in Tooltip.tsx as stopBubble

src/components/dropdown-menu/__tests__/DropdownMenu.test.js
azuruce: Hi Jared, I looked over the potential fix for this. If I keep existing describe block and use test.each, the diff will look a lot harder. But I made that change as we have 173 test.each and 4 describe.each so test.each is more prevalent. Just be aware the diff will be harder

src/components/dropdown-menu/DropdownMenu.js
azuruce: change to useBubble per discussion as we need to default to false to avoid changing snapshot and cause breaking change.

src/components/dropdown-menu/__tests__/DropdownMenu.test.js
jstoffan: @azuruce, rather than iterating over every test with different permutation of props, can we just create a few new tests for these new props? I don't think the former strategy is scalable as props are added.

src/components/dropdown-menu/DropdownMenu.js
```diff
@@ -21,6 +21,7 @@ type Props = {
     onMenuClose?: (event: SyntheticEvent<>) => void,
     /** Handler for dropdown menu open events */
     onMenuOpen?: () => void,
+    useBubble?: boolean,
```
jstoffan: Missing JSDoc.

PR #2008, opened 2020-03-04
chore(draft-js): update to draft-js 0.11.4

booc0mtaco:
update to latest version, from 0.10.5 https://github.com/facebook/draft-js/releases/tag/v0.11.4

BREAKING CHANGE: This will require any consumers to update their version of draft-js, as it is a
peer dependency and may cause issues otherwise.

### DIFF
--- a/package.json
+++ b/package.json
@@ -198,7 +198,7 @@
         "cssnano": "^4.1.10",
         "cypress": "3.8.2",
         "deepmerge": "^4.2.2",
-        "draft-js": "^0.10.1",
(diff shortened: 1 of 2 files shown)


booc0mtaco:
Before merging, here's a list of things to test (known regressions seen in 0.11.0):

- [ ] issue in IE11 where inputting japanese would add an extra line break

booc0mtaco:
Not sure if we will ever pick this up, but taking it off my plate for now

PR #1487, opened 2019-08-02
feat(tasks): add any/all checkbox

shikharmohan:
image

PR #2013, opened 2020-03-09
refactor(tasks): remove feature flag check for any/all tasks

booc0mtaco:
This functionality is now permanently enabled in the content sidebar


booc0mtaco:
See #1487

examples/src/features.js
```diff
@@ -6,6 +6,6 @@ export default {
         },
     },
     tasks: {
-        anyTask: true,
+        assignToGroup: true,
```
booc0mtaco: there is a current check that was missing from this object section. this is swapping the name to add that in the example

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
```diff
@@ -401,7 +401,7 @@ class TaskForm extends React.Component<Props, State> {
                         </PillSelectorDropdown>
 
                         {shouldShowCompletionRule && (
-                            <FeatureFlag feature="activityFeed.tasks.anyTask">
+                            <>
```
ConradJChan: Are these still needed?

booc0mtaco: The fragment is needed since there are two feature flip values directly within the `()`

ConradJChan: Gotcha, sorry missed that

PR #2015, opened 2020-03-10
fix(a11y): Fixing keyboard trap in SelectField

jeeyi:


src/components/select-field/BaseSelectField.js
```diff
@@ -208,6 +208,11 @@ class BaseSelectField extends React.Component<Props, State> {
                     this.closeDropdown();
                 }
                 break;
+            case 'Tab':
+                if (isOpen) {
```
booc0mtaco: Do we want the default event to still occur when `tab` is pressed?

jeeyi: Yes, the default case stops the default events, causing the keyboard trap. We want the default events (keyboard navigation) when `tab` is pressed. @booc0mtaco

PR #2023, opened 2020-03-13
refactor(typescript): Migrate Collapsible, ContextMenu, and Menu

sahiga:
This PR migrates Collapsible, ContextMenu, and all the Menu components to TypeScript.


src/components/collapsible/Collapsible.stories.tsx
```diff
@@ -7,13 +6,13 @@ import Collapsible from './Collapsible';
 import notes from './Collapsible.stories.md';
 
 export const withBorder = () => {
-    const onOpen = arg => {
+    const onOpen = () => {
```
manishPh: removal of `arg` intentional?

sahiga: Yes, I removed it since we were only using it as part of the console logs. I can add it back in and restrict it to the `string | number | boolean` type.

PR #2024, opened 2020-03-13
fix(tooltip): tooltip stop event bubble across whole tooltip

azuruce:


azuruce:
image

azuruce:
The problem with previous solution is the role=presentation div is inside the role=tooltip div, and stopBubble is at presentation layer. Since tooltip has padding, the presentation layer is not able to stop event issued in the padding of the tooltip.

The solution is make presentation layer the outer layer of the tooltip, and attach the proper class that cause the tooltip to render correctly. This means we don't want to share the code and some rewrite is necessary. The test basically passed and I attached the tooltip that is working by now.

src/components/tooltip/Tooltip.tsx
```diff
@@ -279,36 +279,40 @@ class Tooltip extends React.Component<TooltipProps, State> {
...
             <div
+                className={classes}
                 role="presentation"
                 onClick={this.handleTooltipEvent}
                 onContextMenu={this.handleTooltipEvent}
                 onKeyPress={this.handleTooltipEvent}
             >
-                {tooltip}
+                <div id={this.tooltipID} role="tooltip" aria-live="polite">
+                    {text}
```
ittaishay: can we pull out this part into a shared component that's rendered whether it's with or without stop-bubble? and then all that changes is the div that wraps it?

ittaishay: this part being the `{text}` and the `{withCloseButton && ...` part

azuruce: ok

src/components/tooltip/Tooltip.tsx
```diff
@@ -290,25 +290,31 @@ class Tooltip extends React.Component<TooltipProps, State> {
...
         );
 
-        const tooltipWithStopBubble = (
+        const renderTooltip = () => (
+            <div className={classes} id={this.tooltipID} role="tooltip" aria-live="polite">
+                {tooltipInner}
+            </div>
+        );
+
+        const renderTooltipWithStopBubble = () => (
```
DanDeMicco: Instead of separating the component like this and conditionally rendering, what about using the `stopBubble` prop in the `handleTooltipEvent`?

azuruce: Hi, yeah, I think that's another approach.
When this is originally started, I attached handleTooltipEvent to the role=tooltip div and in the handleTooltipEvent, I looked at stopBubble to decide whether I will stopBubble. However, lint complains that I can't attach event handler to role="tooltip", so I add a lint waiver. I got feedback from other parties that we should not waiver lint but instead create another div.

So a new div is created which changed layout, and once layout is changed, the priority becomes do not change existing behavior as much as I can, so it comes two different rendering path.

In the latest solution that I have two layout, for the legacy layout, I don't attach handleTooltipEvent, thus there is no need to check in the handleTooltipEvent that whether stopBubble is true or not.

DanDeMicco: I am just wondering if this can be simplified. The major difference is that in this case, there is a wrapper div with event listeners attached. Can it always render this wrapper div and take this render path (remove the id, as i dont think its necessary?), and then deal with the bubbling in `handleTooltipEvent`

azuruce: hi, if we always render the wrapper div, we will change existing tooltip dom tree. This is bigger risk to take and potentially be a major version bump

azuruce: remove the new id

azuruce:
fix(tooltip): tooltip stop event bubble across whole tooltip
✖   header must not be longer than 72 characters, current length is 77 [header-max-length]

PR #2026, opened 2020-03-17
fix(activity-feed): Deleting task when comments disabled

stevenzyang:


stevenzyang:
@ConradJChan It used to be, but isn't anymore. Permission to delete a task is purely decided by if you created the task

PR #2027, opened 2020-03-17
refactor(left-nav): make tether position props

laughnan:
Make the `tether` props customizable for different navigation types.

### DIFF
--- a/src/features/left-sidebar/LeftSidebarLinkCallout.js
+++ b/src/features/left-sidebar/LeftSidebarLinkCallout.js
@@ -10,8 +10,10 @@ import type { Callout } from './Callout';
 import './styles/LeftSidebarLinkCallout.scss';
 
 type Props = {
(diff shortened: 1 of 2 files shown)


src/features/left-sidebar/LeftSidebarLinkCallout.js
```diff
@@ -10,8 +10,10 @@ import type { Callout } from './Callout';
 import './styles/LeftSidebarLinkCallout.scss';
 
 type Props = {
+    attachmentPosition?: string,
```
ittaishay: would it be possible to type the available positions for both of these props?

laughnan: we actually have them typed here: `src/common/tether-positions.js.flow` i just need to see if that is compatible to import (the "real" constants have been migrated to typescript already).

PR #2028, opened 2020-03-17
feat(classification): Adding classification badge icon

jpfranco:
This update will add the Classification badge (tinycon) icon.

Screen Shot 2020-03-17 at 1 47 52 PM


jpfranco:
Will import through asset import script instead

PR #2029, opened 2020-03-17
fix(dropdown-menu): Call onMenuClose on document clicks

sahiga:
This PR modifies `handleDocumentClick()` in `DropdownMenu` to call the passed-in `onMenuClose()` function (if provided) in addition to `closeMenu()`.

### DIFF
--- a/src/components/dropdown-menu/DropdownMenu.js
+++ b/src/components/dropdown-menu/DropdownMenu.js
@@ -21,7 +21,7 @@ type Props = {
     /** Handler for dropdown menu close events */
     onMenuClose?: (event: SyntheticEvent<> | MouseEvent) => void,
     /** Handler for dropdown menu open events */
-    onMenuOpen: () => void,
+    onMenuOpen?: () => void,
     /** Set true to close dropdown menu on event bubble instead of event capture */
     useBubble?: boolean,
 };
@@ -36,7 +36,6 @@ class DropdownMenu extends React.Component<Props, State> {
         constrainToScrollParent: false,
         constrainToWindow: false,
         isRightAligned: false,
-        onMenuClose: noop,
     };
 
     menuID = uniqueId('menu');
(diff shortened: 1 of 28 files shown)


src/components/dropdown-menu/DropdownMenu.js
```diff
@@ -168,7 +168,13 @@ class DropdownMenu extends React.Component<Props, State> {
             !menuEl.contains(event.target) &&
             !menuButtonEl.contains(event.target)
         ) {
+            const { onMenuClose } = this.props;
+
             this.closeMenu();
+
+            if (onMenuClose) {
```
ConradJChan: Could we just default onMenuClose to noop?

sahiga: @ConradJChan I just updated this file with your suggestion. Could you review again? Thanks!

src/components/dropdown-menu/DropdownMenu.js
```diff
@@ -168,7 +169,9 @@ class DropdownMenu extends React.Component<Props, State> {
             !menuEl.contains(event.target) &&
             !menuButtonEl.contains(event.target)
         ) {
+            const { onMenuClose = noop } = this.props;
```
ConradJChan: Perhaps we could move this to `defaultProps` and then simplify `handleMenuClose` too?

sahiga: Great idea. Will fix

ConradJChan: Seems like defaulting to noop in defaultprops should be good enough? Is the default when destructuring needed?

sahiga: I ran into a bunch of Flow issues when I tried to add `onMenuClose` as a default prop while also keeping it optional, so I resorted to destructuring instead.

PR #2032, opened 2020-03-17
feat(table): Add option for blur detection to selectable tables

naeims:
The `makeSelectable` function takes a component (let's call it `BaseTable`) and returns another (let's call this one `SelectableTable`.)

`SelectableTable`, under the hood, renders the `BaseTable` and adds selection-related functionality to it. It provides specific functions as props such as `onRowClick` and `onRowFocus` to the `BaseTable` which in turn calls those functions upon the appropriate user actions (i.e. DOM events.)

`SelectableTable` internally manages a `focusIndex` state which determines which item in the list is considered "focused." The `BaseTable` implementation may show a different UI treatment for the focused item.

The problem this change solves is: Once `focusIndex` gets set, it never gets unset. As a result, once a `BaseTable` implementation shows an item as focused, there will always be an item that has focus in that table, regardless of where the actual user focus is. For example, if the user focuses the file list (to select a few items, or preview an item), even after clearing the selection or coming back from preview, an item will remain focused (with the gray highlight).

This change enables underlying `BaseTable` implementations to cause the clearing of the internal `focusIndex` state when the entire table loses focus. The underlying `BaseTable` would need to wire up the `onTableBlur` and `onTableFocus` props to the appropriate user actions (i.e. DOM events.)

### DIFF
--- a/examples/src/SelectableTableExamples.js
+++ b/examples/src/SelectableTableExamples.js
@@ -128,6 +128,7 @@ class SelectableTableExamples extends Component {
                 <SelectableTable
                     data={data.map(user => user[2])}
                     enableHotkeys
+                    enableBlurDetection
                     hotkeyType="Item Selection"
                     // eslint-disable-next-line react/no-unused-state
                     onSelect={(selectedItems, focusedItem) => this.setState({ selectedItems, focusedItem })}
--- a/examples/src/SelectableTableExamples.js
+++ b/examples/src/SelectableTableExamples.js
@@ -73,48 +73,50 @@ class FocusableTableRow extends Component {
     }
 }
 
-const SelectableTable = makeSelectable(({ className, onRowClick, onRowFocus, focusedItem, selectedItems }) => {
-    return (
-        <Table className={`has-hover-styles ${className}`}>
-            <TableHeader>
-                {columnHeaders.map(header => (
-                    <TableHeaderCell key={header}>{header}</TableHeaderCell>
-                ))}
-                <TableHeaderCell>Actions</TableHeaderCell>
-            </TableHeader>
-            <TableBody>
-                {data.map((row, index) => (
-                    <ContextMenu key={index}>
-                        <FocusableTableRow
-                            className={selectedItems.has(row[2]) ? 'is-selected ' : ''}
-                            isFocused={row[2] === focusedItem}
-                            onClick={event => onRowClick(event, index)}
-                            onFocus={event => onRowFocus(event, index)}
-                            tabIndex={0}
-                        >
-                            {row.map(cell => (
-                                <TableCell key={cell}>{cell}</TableCell>
-                            ))}
-                            <TableCell>
-                                <DropdownMenu className="dropdown-menu-test">
-                                    <Button type="button">•••</Button>
-                                    <Menu>
-                                        <MenuItem>View Profile</MenuItem>
-                                        <MenuItem>Help</MenuItem>
-                                    </Menu>
-                                </DropdownMenu>
-                            </TableCell>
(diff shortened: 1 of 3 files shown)


src/components/table/makeSelectable.js
```diff
@@ -348,22 +370,30 @@ function makeSelectable(BaseTable) {
...
+                    focusedIndex={focusedIndex}
+                    focusedItem={focusedItem}
+                    onRowClick={this.handleRowClick}
+                    onRowFocus={this.handleRowFocus}
+                />
             );
+
+            if (enableBlurDetection) {
+                table = (
+                    <div onBlur={this.handleTableBlur} onFocus={this.handleTableFocus}>
```
krapspark: Can we pass these to the `BaseTable` component directly? otherwise, we introduce another wrapping `div`.

naeims: Great feedback! This actually simplifies the BUIE change quite a bit (but puts the burden on the consumers). The change is now automatically backward compatible, so there is no need for a boolean prop anymore.

examples/src/SelectableTableExamples.js
```diff
@@ -73,48 +73,50 @@ class FocusableTableRow extends Component {
...
-                            <MenuItem>Help</MenuItem>
-                        </Menu>
-                    </ContextMenu>
-                ))}
-            </TableBody>
-        </Table>
-    );
-});
+const SelectableTable = makeSelectable(
+    ({ className, onRowClick, onRowFocus, focusedItem, selectedItems, onTableBlur, onTableFocus }) => {
```
naeims: git diff is exaggerating this. I only added onTableBlur and onTableFocus to the props list, and onBlur and onFocus to the root Table component.

krapspark: you adding the couple of props made the style change so the content of the function is indented more. haha

src/components/table/makeSelectable.js
```diff
@@ -280,6 +289,18 @@ function makeSelectable(BaseTable) {
             this.onSelect(selectedItems, index);
         };
 
+        handleTableBlur = () => {
+            const { focusedIndex } = this.state;
+            if (focusedIndex !== undefined) {
+                // table may get focus back right away in the same tick, in which case we shouldn't clear focus
+                this.blurTimerID = setTimeout(this.clearFocus);
+            }
+        };
```
macrotim: Excellent write-up of the problem statement!

I'm curious about `setTimeout()`. Why not invoke `clearFocus()` synchronously? 

`setState()` is an async method so "setTimeout on setState" seems redundant.

naeims: Thanks @macrotim !

The problem is this: React synthesizes and bubbles up focus and blur events. Therefore, a blur event that originates anywhere in DOM subtree of the table bubbles to the root of the tree where we attach our `onBlur` handler. For example, as you tab through subcomponents of an item row in the files list (rename button -> more options button -> share button), our root `onBlur` handler gets called for each change in active element. These synthetic blur events are immediately followed by a synthetic focus event that puts the focus back where it is supposed to be, i.e. the next element in tab order.

So in order to normalize (or maybe debounce is a better word) the blur event, we need the `setTimeout()`. I did not know about `setState()` being async, but I tried and it doesn't work correctly without the `setTimeout()`. Perhaps it has to do with the order in which React fires synthesized events.

This solution was inspired by this blog post:
https://medium.com/@jessebeach/dealing-with-focus-and-blur-in-a-composite-widget-in-react-90d3c3b49a9b

macrotim: Thanks for the background info! Super helpful!

PR #2034, opened 2020-03-18
docs(commits): add more notes about commitlint

alexkrolick:


CONTRIBUTING.md
```diff
@@ -50,11 +50,19 @@ include BREAKING CHANGE keyword for breaking changes.
 
 The message summary should be a one-sentence description of the change, and it must be 72 characters in length or shorter. For a list of tags, please [click here](https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional#type-enum). We use the angular style for tags. See the [default release rules](https://github.com/semantic-release/commit-analyzer/blob/master/lib/default-release-rules.js) based on the commit tag. Shown below is an example of the release type that will be done based on a commit message.
 
+In most cases, commits will be a `feat` or `fix`. Make sure to include the `BREAKING CHANGE` string in the summary if there are non-backwards-compatible changes in the commit.
+
+The tags for `refactor`, `chore`, `perf`, `docs` signal that there are **no external changes to _any_ APIs** (including non-breaking ones), and **release notes are not generated** for these tags.
```
booc0mtaco: `test` fits into which group. guessing this one?

we don't use `improvement` but that might be one of the available options. I'm guessing it'd go into the former category and appear in the changelog (since we don't use it, I don't think we need to add it except to be complete w.r.t. our config)

alexkrolick: I originally didn't want to pull in every value from the standard in case it gets out of sync, but we may as well.

alexkrolick: We can also consider adding this: https://commitlint.js.org/#/guides-use-prompt

booc0mtaco: I've been using commitizen locally. not sure how configurable this CLI is, but the benefit of something like commitizen is that we could configure the scope to our taste if we wished. Worth discussing.

CONTRIBUTING.md
```diff
@@ -50,11 +50,31 @@ include BREAKING CHANGE keyword for breaking changes.
...
 
-| Commit message                                                                                                                                          | Release type               |
-| ------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------- |
-| `fix(preview): fullscreen behavior in ie11`                                                                                                             | Patch Release              |
-| `feat(preview): add ability to fullscreen`                                                                                                              | ~~Minor~~ Feature Release  |
-| `feat(preview): remove fullscreen ability`<br><br>`BREAKING CHANGE: The fullscreen ability has been removed due to poor support of the fullscreen api.` | ~~Major~~ Breaking Release |
+#### Semantic Versioning
```
macrotim: This is documented elsewhere. What do you think about linking instead?
ref: https://docs.npmjs.com/about-semantic-versioning

alexkrolick: What do you mean by document elsewhere? By someone else or by us?

alexkrolick: I think we should be clear about our definitions even if we lean on an externally-maintained config at the moment (abstracted through our frontend config repo).

CONTRIBUTING.md
```diff
@@ -50,11 +50,31 @@ include BREAKING CHANGE keyword for breaking changes.
...
+
+"Semantic release" means that changes to the version number of the package (e.g. `3.42.11` to `3.43.0`) are done according to rules that indicate how the change will affect consumers.
+
+The version number is broken into 3 positions &mdash; `Major.Minor.Patch`. In semantic release terms, changes to the numbers follow `Breaking.Feature.Fix`.
+
```
macrotim: The content below is excellent and brings more color to the commit-types.

I'd like to see it moved under a new "Commit Types" section. I was thrown off for a moment because it currently lives under "Semantic Versioning."

CONTRIBUTING.md
```diff
@@ -50,11 +50,31 @@ include BREAKING CHANGE keyword for breaking changes.
...
+- Anything that changes or removes an API, option, or output format is a `BREAKING CHANGE`.
+- Anything that adds new functionality in a backwards-compatible way is a feature (`feat`). Consumers have to upgrade to the new version to use the feature, but nothing will break if they do so.
+- Bugfixes (`fix`) for existing behavior are a patch. Consumers don't have to do anything but upgrade.
+  - Performance fixes (`perf`) and reverts (`revert`) are treated as patch releases.
+  - Automated commits from Box's internalization team cause a patch because they will use the format `fix(moji): ***`.
+- Other prefixes, such as `docs`, don't get released and don't appear in the changelog. These tags signal that there are **no external changes to _any_ APIs** (including non-breaking ones).
+  Changes from these types of commits will get released along with the next releasable commit.
```
macrotim: What does this mean? -> `next releasable commit`.

CONTRIBUTING.md
alexkrolick: adding a link to this page at the end

CONTRIBUTING.md
alexkrolick: Right now it is under 

### Step 4: Push your feature branch to your fork

Just added that semantic release subheading - what do you think about removing it?

CONTRIBUTING.md
alexkrolick: next feat/fix/perf commit

alexkrolick: >  Changes from these types of commits will get released only when the release script detects other releasable commits (feat/fix) going out at the same time.

how does that sound?

CONTRIBUTING.md
alexkrolick: changed to `Commit Types`

PR #2035, opened 2020-03-18
feat(classification): add tooltip to display all applications name list

feixuanli:
Screen Shot 2020-03-18 at 10 52 04 AM

This PR adds a tooltip at the end of FULL formatted security controls for application restrictions when the number of apps exceeds maximum. 
The tooltip content displays the list of all applications' names.

### DIFF
--- a/src/features/classification/security-controls/SecurityControls.js
+++ b/src/features/classification/security-controls/SecurityControls.js
@@ -66,7 +66,7 @@ class SecurityControls extends React.Component<Props, State> {
             items = getFullSecurityControlsMessages(controls, maxAppCount);
         } else {
             const shortMessage = getShortSecurityControlsMessage(controls);
-            items = shortMessage.message ? [shortMessage] : [];
+            items = shortMessage ? [shortMessage] : [];
 
             if (items.length && controlsFormat === SHORT_WITH_BTN) {
                 modalItems = getFullSecurityControlsMessages(controls, maxAppCount);
@@ -80,16 +80,14 @@ class SecurityControls extends React.Component<Props, State> {
         const { isSecurityControlsModalOpen } = this.state;
         const shouldShowSecurityControlsModal =
             controlsFormat === SHORT_WITH_BTN && !!itemName && !!classificationName && !!definition;
-        const securityControlsItems = items.map(({ message, tooltipMessage }) => {
-            if (message) {
-                return <SecurityControlsItem key={message.id} message={message} appNames={tooltipMessage} />;
-            }
-            return null;
-        });
 
         return (
             <>
-                <ul className="bdl-SecurityControls">{securityControlsItems}</ul>
+                <ul className="bdl-SecurityControls">
+                    {items.map(({ message, tooltipMessage }) => (
+                        <SecurityControlsItem key={message.id} message={message} tooltipMessage={tooltipMessage} />
+                    ))}
+                </ul>
                 {shouldShowSecurityControlsModal && (
                     <>
                         <PlainButton className="lnk" onClick={this.openModal} type="button">
--- a/src/features/classification/security-controls/SecurityControls.js
+++ b/src/features/classification/security-controls/SecurityControls.js
@@ -3,7 +3,7 @@ import * as React from 'react';
 import { FormattedMessage } from 'react-intl';
 
 import { DEFAULT_MAX_APP_COUNT, SECURITY_CONTROLS_FORMAT } from '../constants';
-import { getAppsTooltip, getShortSecurityControlsMessage, getFullSecurityControlsMessages } from './utils';
+import { getShortSecurityControlsMessage, getFullSecurityControlsMessages } from './utils';
 import messages from './messages';
 import PlainButton from '../../../components/plain-button';
 import SecurityControlsItem from './SecurityControlsItem';
@@ -61,13 +61,12 @@ class SecurityControls extends React.Component<Props, State> {
 
         let items = [];
         let modalItems;
-        let appNames;
 
         if (controlsFormat === FULL) {
             items = getFullSecurityControlsMessages(controls, maxAppCount);
         } else {
             const shortMessage = getShortSecurityControlsMessage(controls);
-            items = shortMessage ? [shortMessage] : [];
+            items = shortMessage.message ? [shortMessage] : [];
 
             if (items.length && controlsFormat === SHORT_WITH_BTN) {
                 modalItems = getFullSecurityControlsMessages(controls, maxAppCount);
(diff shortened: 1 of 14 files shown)


src/features/classification/security-controls/SecurityControls.js
```diff
@@ -77,6 +78,11 @@ class SecurityControls extends React.Component<Props, State> {
             return null;
         }
 
+        // get applications name list for use in classify modal and security controls modal
+        if (controlsFormat !== SHORT) {
```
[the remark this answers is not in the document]
feixuanli: If it's SHORT_WITH_BTN, We would still need to render the message and pass it to the security controls modal.

src/features/classification/security-controls/SecurityControls.js
```diff
@@ -66,7 +66,7 @@ class SecurityControls extends React.Component<Props, State> {
             items = getFullSecurityControlsMessages(controls, maxAppCount);
         } else {
             const shortMessage = getShortSecurityControlsMessage(controls);
-            items = shortMessage ? [shortMessage] : [];
+            items = shortMessage.message ? [shortMessage] : [];
```
jpfranco: I think this should still return `null` when there's no message in order to simplify validating empty messages. I think the `message` property would ideally be required since there wouldn't be a `MessageItem` object without it

src/features/classification/security-controls/SecurityControls.js
```diff
@@ -80,14 +80,16 @@ class SecurityControls extends React.Component<Props, State> {
         const { isSecurityControlsModalOpen } = this.state;
         const shouldShowSecurityControlsModal =
             controlsFormat === SHORT_WITH_BTN && !!itemName && !!classificationName && !!definition;
+        const securityControlsItems = items.map(({ message, tooltipMessage }) => {
+            if (message) {
```
jpfranco: I think we could avoid this extra check and the `return` below with the approach i mentioned above

src/features/classification/security-controls/SecurityControlsItem.js
```diff
@@ -2,18 +2,34 @@
 import * as React from 'react';
 import { FormattedMessage, type MessageDescriptor } from 'react-intl';
 
+import { bdlBoxBlue } from '../../../styles/variables';
+import Tooltip from '../../../components/tooltip';
+import IconInfo from '../../../icons/general/IconInfo';
 import './SecurityControlsItem.scss';
 
 type Props = {
+    appNames: ?MessageDescriptor,
```
jpfranco: Sorry I think this might be related to some of my other suggestions, but since we switched the approach to returning a `MessageItem` with a `tooltipMessage`, I think it might make more sense for this to be `tooltipMessage` as well. I think `Props` could probably be just:

```
type Props = `MessageItem`;
```

src/features/classification/security-controls/SecurityControlsModal.js
```diff
@@ -11,6 +11,7 @@ import Label from '../../../components/label/Label';
 import messages from './messages';
 import SecurityControlsItem from './SecurityControlsItem';
 import './SecurityControlsModal.scss';
+import type { MessageItem } from './utils';
```
jpfranco: Should this type be in `flowTypes`?

src/features/classification/security-controls/SecurityControlsModal.js
```diff
@@ -38,6 +39,12 @@ const SecurityControlsModal = ({
     }
 
     const title = <FormattedMessage {...messages.modalTitle} values={{ itemName }} />;
+    const securityControlsItems = modalItems.map(({ message, tooltipMessage }) => {
+        if (message) {
+            return <SecurityControlsItem key={message.id} message={message} appNames={tooltipMessage} />;
+        }
+        return null;
```
jpfranco: Same here, this can probably be avoided by having `message` property as required and not returning an item at all when there's no message

src/features/classification/security-controls/utils.js
```diff
@@ -10,58 +10,63 @@ import downloadRestrictionsMessageMap from './downloadRestrictionsMessageMap';
 import messages from './messages';
 import { ACCESS_POLICY_RESTRICTION, DOWNLOAD_CONTROL, LIST_ACCESS_LEVEL, SHARED_LINK_ACCESS_LEVEL } from '../constants';
 
+type MessageItem = {
+    message: ?MessageDescriptor,
```
jpfranco: I think this should be in `flowTypes` and `message` should be required as mentioned above

src/features/classification/security-controls/utils.js
```diff
@@ -10,58 +10,63 @@ import downloadRestrictionsMessageMap from './downloadRestrictionsMessageMap';
...
+    tooltipMessage?: MessageDescriptor,
+};
+
 const { SHARED_LINK, DOWNLOAD, EXTERNAL_COLLAB, APP } = ACCESS_POLICY_RESTRICTION;
 const { DESKTOP, MOBILE, WEB } = DOWNLOAD_CONTROL;
 const { BLOCK, WHITELIST, BLACKLIST } = LIST_ACCESS_LEVEL;
 const { COLLAB_ONLY, COLLAB_AND_COMPANY_ONLY, PUBLIC } = SHARED_LINK_ACCESS_LEVEL;
 
-const getShortSecurityControlsMessage = (controls: Controls): ?MessageDescriptor => {
+const getShortSecurityControlsMessage = (controls: Controls): MessageItem => {
```
jpfranco: This could return a type `?MessageItem`. The structure of the object returned has changed, but I don't think it's necessary to change the logic of returning null when there's no message.

src/features/classification/security-controls/SecurityControls.js
feixuanli: nice !

src/features/classification/security-controls/__tests__/SecurityControlsItem.test.js
```diff
@@ -1,21 +1,36 @@
...
-        wrapper = getWrapper({ message });
+        tooltipMessage = {
+            id: 'id2',
+            defaultMessage: 'message2',
+        };
+        wrapper = getWrapper({ message, tooltipMessage });
     });
 
-    test('should render a SecurityControlsItem with a message', () => {
+    test('should render a SecurityControlsItem with a message and Tooltip', () => {
```
jpfranco: nit: The default should probably be the case without the tooltip since that is the most common

src/features/classification/security-controls/__tests__/SecurityControlsItem.test.js
```diff
@@ -1,21 +1,36 @@
...
         expect(wrapper).toMatchSnapshot();
+        expect(wrapper.find(Tooltip)).toHaveLength(1);
+        expect(wrapper.find(IconInfo)).toHaveLength(1);
+    });
+
+    test('should not render Tooltip if tooltipMessage is received as underfined', () => {
```
jpfranco: nit: Typo in `underfined`

src/features/classification/security-controls/__tests__/SecurityControlsItem.test.js
feixuanli: thanks, will leave this nit as-is.

src/features/classification/security-controls/__tests__/SecurityControlsItem.test.js
feixuanli: updated !

PR #2036, opened 2020-03-19
fix(content-preview): allow updating token in elements api

mickr:
* Add `updateToken` feature to the `APIFactory`, this fixes an issue when changing the token prop, requests are still made with the old token. This is needed when you update the token from the storybook knob.
* Add knobs for file and token to preview stories
* Add source-map-loader to get source maps from other modules (annotations in this case).


src/elements/content-preview/ContentPreview.js
```diff
@@ -339,6 +339,8 @@ class ContentPreview extends React.PureComponent<Props, State> {
             this.loadPreview();
         } else if (hasTokenChanged) {
             this.updatePreviewToken();
+            this.api.updateToken(token);
+            this.fetchFile(currentFileId);
```
ConradJChan: Is this required?

package.json
```diff
@@ -184,6 +184,7 @@
         "babel-plugin-react-remove-properties": "^0.3.0",
         "babel-plugin-rewire": "^1.0.0",
         "babel-preset-react-app": "^9.1.0",
+        "box-annotations": "^3.13.0",
```
jstoffan: Do we want to add this right now?

src/api/APIFactory.js
```diff
@@ -273,6 +273,14 @@ class APIFactory {
         }
     }
 
+    /**
+     *
+     * @param {Token} token - token used to make requests
+     */
+    updateToken(token: Token = ''): void {
```
jstoffan: Several questions about this approach:
- Should we be mutating `this.options`? 
- Should this method be named `setToken`? 
- Should this method have a default parameter value at all? 
- Should the factory destroy any in-flight requests if the token changes?

src/elements/content-preview/ContentPreview.js
```diff
@@ -339,6 +339,8 @@ class ContentPreview extends React.PureComponent<Props, State> {
             this.loadPreview();
         } else if (hasTokenChanged) {
             this.updatePreviewToken();
+            this.api.updateToken(token);
```
jstoffan: Should this be supported for all top-level Elements? It doesn't seem preview-specific.

package.json
mickr: I figured we could add now in the story, or provide instructions to develop on Annotations.

src/elements/content-preview/ContentPreview.js
mickr: It is in the case if you're changing the key and the filename. There is another strategy I can employ to refresh the component in the story.

src/elements/content-preview/ContentPreview.js
mickr: This came about because having the knobs in the storybook to switch tokens through props. I'm not sure other elements have this issue, but there is another strategy I can do inside storybook that will basically refresh reload the component with a button when we change the token and file id's

src/elements/content-preview/ContentPreview.js
mickr: I removed the call, and added the refresh button to the story.

src/elements/content-preview/ContentPreview.js
```diff
@@ -355,6 +354,16 @@ class ContentPreview extends React.PureComponent<Props, State> {
         }
     }
 
+    /**
+     * Updates the access token used by preview library
+     *
+     */
+    updateApiToken() {
+        if (this.api) {
+            this.api.setToken(this.props.token);
```
mxiao6: Should we always destruct props?

package.json
mickr: I can remove it if you think it isn't a good idea to include yet.

src/api/APIFactory.js
```diff
@@ -273,6 +273,17 @@ class APIFactory {
         }
     }
 
+    /**
+     *
```
ConradJChan: Missing description of method

src/api/APIFactory.js
```diff
@@ -273,6 +273,17 @@ class APIFactory {
...
+     * @param {Token} token - token used to make requests
+     */
+    setToken(token: Token): void {
+        if (token) {
```
ConradJChan: minor: Small method so it doesn't matter much, but thoughts on early return like `if (!token) { return; }`?

src/elements/content-preview/stories/ContentPreview.stories.js
```diff
@@ -1,54 +1,94 @@
...
 import ContentPreview from '../ContentPreview';
 import notes from './ContentPreview.notes.md';
 
-export const Preview = () => (
-    <IntlProvider locale="en">
-        <ContentPreview features={global.FEATURES} fileId={global.FILE_ID} hasHeader token={global.TOKEN} />
-    </IntlProvider>
-);
+// Used to re-render the story with props from the knobs
+let key = 0;
```
ConradJChan: Is there a way to do this without being declared in the global scope?

mickr: There are ways but this is already just a tweak for storybook to be able to force-reRender the story when if you change the access token and file at the same time and the scope should only affect the story.

src/elements/content-preview/ContentPreview.js
mickr: I'm not sure that is a requirement when we are just using the value once in a small function

mxiao6: yeah in small function destructuring doesn't have any benefits.

.storybook/webpack.config.js
```diff
@@ -21,30 +22,31 @@ module.exports = async ({ config }) => {
...
-                ],
-                loader: 'babel-loader',
-                options: {
-                    babelrc: false,
-                    compact: false,
-                    configFile: path.resolve('babel.config.js'),
-                },
-            }
-        );
+        config.module.rules.push({
```
jstoffan: Did this get run through prettier or something? Seems weird that the formatting would change.

package.json
jstoffan: I don't think we should include it right now. We can each just add it locally, I think.

src/elements/content-preview/ContentPreview.js
```diff
@@ -353,6 +354,16 @@ class ContentPreview extends React.PureComponent<Props, State> {
         }
     }
 
+    /**
+     * Updates the access token used by preview library
+     *
+     */
+    updateApiToken() {
+        if (this.api) {
```
jstoffan: Can this be falsy?

src/elements/content-preview/stories/ContentPreview.stories.js
jstoffan: If we want to force refresh when the token changes, it doesn't seem like we need the ContentPreview changes. Regarding the key, can we do something like this?

```
const fileId = text('File ID', global.FILE_ID);
const token = text('Access Token', global.TOKEN);

return (
        <IntlProvider locale="en" key={`${fileId}-${token}`>
            <ContentPreview
                features={global.FEATURES}
                fileId={fileId}
                hasHeader
                token={token}
            />
        </IntlProvider>
    );
```

.storybook/webpack.config.js
mickr: Yeah

src/elements/content-preview/stories/ContentPreview.stories.js
```diff
@@ -1,54 +1,91 @@
...
+                hasHeader
+                key={`${fileId}-${token}`}
+                token={token}
+            />
+        </IntlProvider>
+    );
+};
 
 export const PreviewWithAnnotations = () => {
+    const fileId = text('File ID', global.FILE_ID);
```
jstoffan: Can these be declared once at the top of the module?

mickr: No, they're supposed to be declared in the story.

.storybook/webpack.config.js
```diff
@@ -21,30 +22,31 @@ module.exports = async ({ config }) => {
         {
             test: /\.tsx?$/,
             exclude: /node_modules/,
-            use: [
-                { loader: 'babel-loader' },
-                { loader: 'react-docgen-typescript-loader' }
-            ]
+            use: [{ loader: 'babel-loader' }, { loader: 'react-docgen-typescript-loader' }],
+        },
+        {
```
alexkrolick: This is the new chunk? Doesn't this map to `.j, js, .t, .ts`? The `.tsx?` above maps to `.ts, .tsx`

mickr: I can scope it to only map files, the reason is we want to use Storybook to aid in development for annotations with ContentPreview and sourcemaps from annotations and preview js will aid in that effort.

alexkrolick: Isn't it _excluding_ annotations right now though?

alexkrolick: Not sure what the webpack parse order is but it looks like something else is handing *.js in the base config: ttps://github.com/box/box-ui-elements/blob/master/scripts/webpack.config.js#L151

mickr: No it is excluding everything in node_modules besides annotations.

mickr: I am also not looking to include source maps in the release or anywhere else besides storybook.

PR #2038, opened 2020-03-19
fix(icon): Use image icon for heic and heif files

stevenzyang:
Also treats file extensions case insensitively when determining file
type


.vscode/launch.json
```diff
@@ -7,19 +7,32 @@
         {
             "type": "node",
             "request": "launch",
-            "name": "Launch Jest",
+            "name": "Test single file",
```
booc0mtaco: Hm, i think this stuff shouldn't be here. not sure what caused the diff

stevenzyang: It was commited a long time ago. I can add this file to the gitignore but figured it might be useful for people

src/icons/file-icon/FileIcon.js.flow
```diff
@@ -189,7 +203,7 @@ type Props = {
 };
 
 const FileIcon = ({ dimension = 32, extension = '', title }: Props) => {
-    const IconComponent = Components[getFileIconComponent(extension)];
+    const IconComponent = Components[getFileIconComponent(extension.toLowerCase())];
```
il10: why are we lowercasing?

stevenzyang: In windows, file extensions are case insensitive, so it's possible for someone to upload a file with an uppercased extension. When I googled the heic type I found multiple examples of `.HEIC`, which I found out we currently don't handle without this change

booc0mtaco: suggest doing the lower-casing in `getFileIconComponent` instead of at the call site

src/icons/file-icon/FileIcon.js.flow
```diff
@@ -142,7 +142,21 @@ const EXTENSIONS = {
...
+        'jpg',
+        'png',
+        'ps',
+        'svs',
+        'svg',
+        'tif',
+        'tiff',
+        'heic',
+        'heif',
+    ]),
```
stevenzyang: Oh yeah I'm not sure what the protocol is for updating files with both .flow and .tsx versions. Do I just copy the change to both?

il10: As far as I can tell from other flow/tsx files in BUIE the change goes to both. @booc0mtaco can confirm

booc0mtaco: yeah, we want to mirror the changes in both for the time being.

- new components do not need to have a .flow file. typescript-only for new content
- existing components should keep the two in sync (esp. regarding type info... that's why they are still useful, i think.)

PR #2041, opened 2020-03-20
feat(uploads): add conflictCallback option

omgrief:
allows for an additional callback to be called on 409 where the client can specify how they want to rename the file they are trying to upload

### DIFF
--- a/src/api/uploads/BaseUpload.js
+++ b/src/api/uploads/BaseUpload.js
@@ -24,6 +24,8 @@ class BaseUpload extends Base {
 
     overwrite: boolean;
 
+    conflictCallback: ?(fileName: string) => string;
+
     preflightSuccessHandler: Function;
 
     retryCount: number = 0;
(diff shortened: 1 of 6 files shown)


src/api/uploads/BaseUpload.js
```diff
@@ -24,6 +24,8 @@ class BaseUpload extends Base {
 
     overwrite: boolean;
 
+    conflictCallback: Function;
```
jstoffan: We've been trying to avoid generic types like Function. Can you add the input/output parameters?

src/api/uploads/PlainUpload.js
```diff
@@ -126,8 +127,10 @@ class PlainUpload extends BaseUpload {
         successCallback = noop,
         errorCallback = noop,
         progressCallback = noop,
+        conflictCallback = null,
```
jstoffan: It looks like the callback has a default value in some places, but not in others. Should it be consistent?

src/api/uploads/MultiputUpload.js
```diff
@@ -1220,6 +1238,10 @@ class MultiputUpload extends BaseMultiput {
             this.fileId = data.context_info.conflicts.id;
             return;
         }
+        if (this.conflictCallback) {
```
jstoffan: Looks to be missing a test.

PR #2048, opened 2020-03-25
fix(upload): fallback sha1 calculation for IE11 using new package

omgrief:


src/utils/__tests__/webcrypto.test.js
```diff
@@ -64,13 +66,33 @@ describe('util/webcrypto', () => {
                     },
                 };
 
-                const expectedError = new Error('lol');
+                const expectedError = new Error('ERROR');
```
sahiga: Thank you for fixing this

src/utils/__tests__/webcrypto.test.js
```diff
@@ -64,13 +66,33 @@ describe('util/webcrypto', () => {
...
+                window.msCrypto = {
+                    subtle: {
+                        digest: digestMock,
+                    },
+                };
+
+                const hash = await digest('SHA-1', buffer);
+                expect(hash).toBeDefined();
+                expect(digestMock).not.toHaveBeenCalled();
+                expect(sha1.arrayBuffer).toHaveBeenCalledWith(buffer);
```
sahiga: Can we also test (1) that `sha1.arrayBuffer()` is not called when the algorithm is not SHA-1 and (2) that the promise is rejected when the call to `sha1.arrayBuffer1()` fails?

src/utils/__tests__/webcrypto.test.js
omgrief: 👍

src/utils/__tests__/webcrypto.test.js
omgrief: added tests

alexkrolick:
This commit is missing the BREAKING CHANGE keyword in the message, we''ll have to do something to trigger that in the autogenerated release

PR #2049, opened 2020-03-26
feat(content-sidebar): Support multifile tasks

stevenzyang:
Having the "View Task Details" button handler be a function instead of an href allows the consumer (e.g EUA) to route to the URL without causing a page refresh

### DIFF
--- a/src/elements/content-sidebar/ActivitySidebar.js
+++ b/src/elements/content-sidebar/ActivitySidebar.js
@@ -55,6 +55,7 @@ type ExternalProps = {
     onTaskCreate: Function,
     onTaskDelete: (id: string) => any,
     onTaskUpdate: () => any,
+    onTaskView: (id: string, isCreator: boolean) => any,
 } & ErrorContextProps;
 
 type PropsWithoutContext = {
@@ -591,6 +592,7 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
             getUserProfileUrl,
             activeFeedEntryId,
             activeFeedEntryType,
+            onTaskView,
         } = this.props;
         const {
             currentUser,
@@ -623,6 +625,7 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
                     onTaskCreate={this.createTask}
                     onTaskDelete={this.deleteTask}
                     onTaskUpdate={this.updateTask}
+                    onTaskView={onTaskView}
                     onTaskModalClose={this.onTaskModalClose}
                     onTaskAssignmentUpdate={this.updateTaskAssignment}
                     getApproverWithQuery={this.getApproverWithQuery}
--- a/src/elements/content-sidebar/activity-feed/task-new/Task.js
+++ b/src/elements/content-sidebar/activity-feed/task-new/Task.js
@@ -9,8 +9,6 @@ import TetherComponent from 'react-tether';
 import { withFeatureConsumer, getFeatureConfig } from '../../../common/feature-checking';
 import { withAPIContext } from '../../../common/api-context';
 import Avatar from '../Avatar';
-// $FlowFixMe LabelPill is in typescript
-import LabelPill from '../../../../components/label-pill';
 import Media from '../../../../components/media';
 import { MenuItem } from '../../../../components/menu';
 import ActivityError from '../common/activity-error';
@@ -21,7 +19,6 @@ import IconTaskApproval from '../../../../icons/two-toned/IconTaskApproval';
 import IconTaskGeneral from '../../../../icons/two-toned/IconTaskGeneral';
 import IconTrash from '../../../../icons/general/IconTrash';
 import IconPencil from '../../../../icons/general/IconPencil';
-import MoveCopy16 from '../../../../icon/line/MoveCopy16';
 import UserLink from '../common/user-link';
 import API from '../../../../api/APIFactory';
 import {
@@ -44,6 +41,7 @@ import TaskDueDate from './TaskDueDate';
 import TaskStatus from './TaskStatus';
 import AssigneeList from './AssigneeList';
 import TaskModal from '../../TaskModal';
+import TaskMultiFileIcon from './TaskMultiFileIcon';
 import commonMessages from '../../../common/messages';
 import messages from './messages';
 import type { GetAvatarUrlCallback, GetProfileUrlCallback } from '../../../common/flowTypes';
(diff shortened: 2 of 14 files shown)


src/elements/content-sidebar/activity-feed/task-new/Task.js
```diff
@@ -345,38 +368,53 @@ class Task extends React.Component<Props, State> {
...
+                                    ? onView && (
+                                          <PrimaryButton
+                                              onClick={() => onView(id, isCreator)}
+                                              data-testid="view-details-button"
+                                          >
+                                              <FormattedMessage {...messages.tasksFeedViewDetailsAction} />
+                                          </PrimaryButton>
+                                      )
+                                    : currentUserAssignment && (
+                                          <TaskActions
```
alexkrolick: What do you think about either:
- putting the multi-file action inside TaskActions, because it already has logic to decide what buttons to show (Approve/Reject | Complete)
- breaking apart TaskActions to keep the logic in one place

src/elements/content-sidebar/activity-feed/task-new/Task.js
```diff
@@ -319,7 +338,11 @@ class Task extends React.Component<Props, State> {
                         </div>
                         <div className="bcs-Task-status">
                             <TaskStatus status={status} />
-
+                            {isMultiFile && (
+                                <LabelPill.Pill data-testid="multifile-badge">
```
alexkrolick: In the same vein as https://github.com/box/box-ui-elements/pull/2049/files#r398273375, what about either pulling the logic out of the TaskCompletionRuleIcon & and TaskStatus helpers, or making a helper component that behaves similar to those?

This does its own visibility check: https://github.com/box/box-ui-elements/blob/0a6eeec68da9869e0a8238c90d52108fdd3a3747/src/elements/content-sidebar/activity-feed/task-new/TaskCompletionRuleIcon.js#L19

src/elements/content-sidebar/activity-feed/task-new/Task.scss
```diff
@@ -40,6 +40,10 @@ $bcs-task-left-offset: 10 * $bcs-task-grid + 2px;
 .bcs-Task-status {
     margin-top: 3 * $bcs-task-grid;
     margin-bottom: 2 * $bcs-task-grid;
+
+    & > * {
+        margin-right: 4px;
```
alexkrolick: this is equal to $bcs-task-grid (which should now be `$bdl-grid-unit` since that has been created)

src/elements/content-sidebar/activity-feed/task-new/__tests__/Task.test.js
```diff
@@ -138,38 +178,50 @@ describe('elements/content-sidebar/ActivityFeed/task-new/Task', () => {
...
-        const wrapper = shallow(
-            <Task currentUser={currentUser} {...task} isPending={false} onAssignmentUpdate={jest.fn()} />,
-        );
-
-        expect(wrapper.find('TaskActions')).toHaveLength(1);
+        [task, taskMultifile].forEach(eachTask => {
+            const wrapper = shallow(
+                <Task currentUser={currentUser} {...eachTask} isPending={false} onAssignmentUpdate={jest.fn()} />,
+            );
+            expect(wrapper.find('.bcs-Task-actionsContainer')).toHaveLength(1);
```
alexkrolick: If you don't trust this classname to be stable (seems likely), use a testid

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -55,6 +55,7 @@ type ExternalProps = {
     onTaskCreate: Function,
     onTaskDelete: (id: string) => any,
     onTaskUpdate: () => any,
+    onTaskView: (id: string, isCreator: boolean) => any,
```
alexkrolick: Do you have to propagate this even higher, to the ContentPreview or Preview Sidebar component props? Those are the top-level exports from BUIE.

src/elements/content-sidebar/activity-feed/task-new/Task.js
stevenzyang: Ah, I'll put it in its own component since the inbox uses the same icon as well

src/elements/content-sidebar/activity-feed/task-new/Task.scss
stevenzyang: Got it, will change

src/elements/content-sidebar/activity-feed/task-new/Task.js
stevenzyang: I'll move it into TaskActions

src/elements/content-sidebar/ActivitySidebar.js
stevenzyang: These props get passed in inside ActivitySidebarProps, which is inside ContentSidebarProps, which gets passed to ContentPreview

src/elements/content-sidebar/activity-feed/task-new/Task.js
```diff
@@ -319,7 +335,7 @@ class Task extends React.Component<Props, State> {
                         </div>
                         <div className="bcs-Task-status">
                             <TaskStatus status={status} />
-
+                            <TaskMultiFileIcon isMultiFile={isMultiFile} />
```
ConradJChan: Seems interesting to pass in a prop which determines whether the Icon will render or not. What are your thoughts on letting the Icon always render and move the condition to the client logic, so `{ isMultiFile && <TaskMultiFileIcon /> }`

alexkrolick: https://github.com/box/box-ui-elements/pull/2049#discussion_r398274700

ConradJChan: Gotcha, thanks for the context 👍

stevenzyang:
Just pushed a commit that adds this tooltip to the multifile icon. It's part of the design spec but I missed it initially.
Screen Shot 2020-03-30 at 1 33 49 PM

src/elements/content-sidebar/activity-feed/task-new/TaskMultiFileIcon.js
```diff
@@ -0,0 +1,26 @@
...
+import MoveCopy16 from '../../../../icon/line/MoveCopy16';
+
+type Props = {|
+    isMultiFile: boolean,
+|};
+
+const TaskMultiFileIcon = ({ isMultiFile }: Props): React.Node => {
+    return (
+        isMultiFile && (
+            <Tooltip position="top-center" text={<FormattedMessage {...messages.taskMultipleFilesAffordanceTooltip} />}>
```
alexkrolick: don't see the message definition here

PR #2052, opened 2020-03-27
fix(sharedlinksection): adding aria label to shared link input field

il10:
[508 Compliance] Accessibility Fixes: Adding an aria label "URL" to the input field for shared link


src/components/text-input-with-copy-button/TextInputWithCopyButton.js
```diff
@@ -215,7 +215,7 @@ class TextInputWithCopyButton extends React.PureComponent<Props, State> {
 
         return (
             <div className={wrapperClasses} {...copyEvent}>
-                <TextInput {...inputProps} onFocus={this.handleFocus} />
+                <TextInput {...inputProps} onFocus={this.handleFocus} aria-label="URL" />
```
tjgupta: Should probably use a intl message for this label, although I'm not sure if that's translatable or not...

booc0mtaco: Also, we might have this message or similar somewhere already. Not sure if we prefer "URL" or something like "Web Address"

alexkrolick: Isn't this a generic component? It doesn't always have to be a URL at all.

il10: updated.

il10: @booc0mtaco guidance provided that we add the label "URL" here. Not sure if that can be changed

alexkrolick: Maybe it should be a prop then? If you use this for an email address, for example, it can't be "URL".

il10: @alexkrolick now using label within `SharedLinkSection` instead of `TextInputWithCopyButton` to reflect shared link URL in USM

PR #2056, opened 2020-03-31
fix(draftjseditor): adding aria-labelledBy to comment input field

il10:
[508 compliance] accessibility fix: Adding aria-labelledby to comment input field

### DIFF
--- a/src/components/draft-js-editor/DraftJSEditor.js
+++ b/src/components/draft-js-editor/DraftJSEditor.js
@@ -4,6 +4,7 @@ import classNames from 'classnames';
 
 import { Editor } from 'draft-js';
 import type { EditorState } from 'draft-js';
+import Label from '../label';
 import 'draft-js/dist/Draft.css';
 
 import Tooltip from '../tooltip';
@@ -14,8 +15,11 @@ type Props = {
     ariaLabel?: string,
     editorState: EditorState,
     error?: ?Object,
+    hideLabel?: boolean,
     inputProps: Object,
     isDisabled?: boolean,
+    isRequired?: boolean,
+    label: React.Node,
     onBlur: Function,
     onChange: Function,
     onFocus: Function,
@@ -26,6 +30,7 @@ type Props = {
 class DraftJSEditor extends React.Component<Props> {
     static defaultProps = {
         inputProps: {},
+        isRequired: false,
         isFocused: false,
     };
 
@@ -67,7 +72,18 @@ class DraftJSEditor extends React.Component<Props> {
     };
 
     render() {
-        const { ariaLabel, editorState, error, inputProps, isDisabled, onFocus, placeholder } = this.props;
+        const {
+            ariaLabel,
+            editorState,
+            error,
+            hideLabel,
+            inputProps,
+            isDisabled,
+            isRequired,
+            label,
+            onFocus,
+            placeholder,
+        } = this.props;
 
         const { handleBlur, handleChange } = this;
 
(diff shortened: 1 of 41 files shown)


src/components/draft-js-editor/DraftJSEditor.js
```diff
@@ -110,6 +121,7 @@ class DraftJSEditor extends React.Component<Props> {
                         <div>
                             <Editor
                                 {...a11yProps}
+                                ariaLabel={formatMessage(messages.commentTextInputLabel)}
```
il10: For context the flow is as following:  CommentForm > DraftJSMentionSelector > DraftJSMentionSelectorCore > DraftJSEditor > actual Editor from draft-js package

adding ariaLabel prop here renders it in the DOM on the comment form

il10:
> * I don't think we want both a `label` tag and `aria-label`
> * There doesn't need to be a default, it should take a string or FormattedMessage for each use case (`Comment` is one of them)

What do you mean by for each use case?

src/components/draft-js-editor/DraftJSEditor.js
```diff
@@ -104,24 +90,23 @@ class DraftJSEditor extends React.Component<Props> {
 
         return (
             <div className={classes}>
-                <Label hideLabel={hideLabel} showOptionalText={!isRequired} text={label}>
```
il10: removing label as we are now using an aria-label

src/components/draft-js-editor/DraftJSEditor.js
```diff
@@ -104,24 +90,23 @@ class DraftJSEditor extends React.Component<Props> {
...
-                            />
-                        </div>
-                    </Tooltip>
-                </Label>
+                <Tooltip isShown={!!error} position="bottom-left" text={error ? error.message : ''} theme="error">
+                    {/* need div so tooltip can set aria-describedby */}
+                    <div>
+                        <Editor
+                            {...a11yProps}
+                            ariaLabel={ariaLabel === undefined ? placeholder : ariaLabel}
```
mxiao6: How about `ariaLabel || placeholder`? In addition, `placeholder` is not a required prop neither, you may consider add a default prop value for one of them.

src/elements/content-sidebar/activity-feed/comment-form/messages.js
```diff
@@ -12,6 +12,12 @@ const messages = defineMessages({
         defaultMessage: 'Cancel',
         description: 'Text for cancel button',
     },
+    commentLabel: {
+        id: 'be.contentSidebar.activityFeed.commentForm.commentLabel',
+        defaultMessage: 'Comment',
+        description: 'Accessible label for comment input field',
+    },
+
```
mxiao6: Unnecessary blank

src/components/draft-js-editor/DraftJSEditor.js
```diff
@@ -104,24 +90,23 @@ class DraftJSEditor extends React.Component<Props> {
...
-                                onChange={handleChange}
-                                onFocus={onFocus}
-                                placeholder={placeholder}
-                                readOnly={isDisabled}
-                                stripPastedStyles
-                            />
-                        </div>
-                    </Tooltip>
-                </Label>
+                <Tooltip isShown={!!error} position="bottom-left" text={error ? error.message : ''} theme="error">
```
alexkrolick: An `aria-label` is not visible, so it's not a replacement for a `label` tag except in the case where the the `hideLabel` prop for this component causes it to be visually hidden.

The problem we are trying to solve is that `label` can only label a "real" form element, such as an `input`, and not the virtual form element created by DraftJS. An alternative is to use `aria-labelledby="some-span-id-123"` for the visual case, and either leave it and hide it or switch over to `aria-label="some string"` for the hidden case.

Also, removing the `label` prop is a breaking change that we probably don't want to manage right now, so it would be preferable to fix this using the same props.

https://github.com/facebook/draft-js/blob/85764327e49f94397a83c3431f5a03cc02a18104/src/component/base/DraftEditorProps.js#L89-L97

src/components/draft-js-editor/DraftJSEditor.js
il10: Actually I do not think these two should be associated w/ each other. Now setting ariaLabel={ariaLabel} @mxiao6

src/components/draft-js-editor/DraftJSEditor.js
```diff
@@ -104,24 +94,26 @@ class DraftJSEditor extends React.Component<Props> {
...
-                                onChange={handleChange}
-                                onFocus={onFocus}
-                                placeholder={placeholder}
-                                readOnly={isDisabled}
-                                stripPastedStyles
-                            />
-                        </div>
-                    </Tooltip>
-                </Label>
+                <span className={classNames({ 'accessibility-hidden': hideLabel })} id={this.labelID}>
```
alexkrolick, line 97: I think that `showOptionalText` controls an indicator whether the form field is required or not: `My Text (optional)`

il10: I thought it was indicating whether the optional field was required?  `Whether to show the `(Optional)` text next to the label for an optional field`

Is this needed? @alexkrolick

alexkrolick: Yes, it adds text next to the label that says `(Optional)` (localized string). That was handled by the Label component but it's not here any more.

src/components/draft-js-editor/__tests__/DraftJSEditor.test.js
```diff
@@ -23,22 +23,13 @@ describe('components/draft-js-editor/DraftJSEditor', () => {
...
-            expect(label.length).toBe(1);
-            expect(label.prop('hideLabel')).toBeFalsy();
             expect(tooltip.length).toBe(1);
             expect(tooltip.children().is('div')).toBe(true);
             expect(wrapper.find(Editor).length).toBe(1);
         });
 
-        test('should hide label when specified', () => {
-            const wrapper = shallow(<DraftJSEditor {...requiredProps} hideLabel />);
-            expect(wrapper.find('Label').prop('hideLabel')).toBe(true);
```
alexkrolick, line 39: We should change this test to look for the `accessibility-hidden` class on the label span

src/components/draft-js-editor/DraftJSEditor.js
il10: 👍

src/components/draft-js-editor/__tests__/DraftJSEditor.test.js
il10: 👍

PR #2064, opened 2020-04-07
feat(annotations): Provide Annotations context

ConradJChan:


src/elements/content-preview/ContentPreview.js
```diff
@@ -741,6 +753,11 @@ class ContentPreview extends React.PureComponent<Props, State> {
         this.preview.addListener('preview_metric', this.onPreviewMetric);
         this.preview.addListener('thumbnailsOpen', () => this.setState({ isThumbnailSidebarOpen: true }));
         this.preview.addListener('thumbnailsClose', () => this.setState({ isThumbnailSidebarOpen: false }));
+
+        if (previewOptions.showAnnotationsControls) {
```
mxiao6, line 757: Can use `showAnnotationsControls` directly here without `previewOptions.`

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+        getActionSuffix(status: Status) {
+            return status === Status.SUCCESS || status === Status.ERROR ? 'end' : 'start';
+        }
+
+        handleAnnotationCreate = ({ annotation, error, meta }: AnnotationActionEvent): void => {
+            const { status } = meta;
+            const actionSuffix = this.getActionSuffix(status);
+            const action = `create_${actionSuffix}` as Action;
+            this.setState({ ...this.state, annotation, action, error });
```
mxiao6, line 31: Is `...this.state` needed in `setState`?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -0,0 +1,47 @@
+import * as React from 'react';
+import AnnotatorContext from './AnnotatorContext';
+import { Action, AnnotationActionEvent, AnnotatorState, Status } from './types';
```
mxiao6, line 3: I don't know should we define State type in another file?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -0,0 +1,47 @@
...
+type WrapperReturn<P> = React.ComponentClass<P & WrapperProps>;
+
+export default function withAnnotations<P extends object>(
+    WrappedComponent: React.ComponentType<P & WrapperProps>,
+): WrapperReturn<P> {
+    class ComponentWithAnnotations extends React.Component<P & WrapperProps, AnnotatorState> {
+        static displayName: string;
+
+        state: AnnotatorState = {
+            annotation: null,
```
jstoffan, line 18: How will this work when multiple operations are performed in sequence? Will this state get copied elsewhere? Any chance of conflicts between events?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -0,0 +1,47 @@
...
+            action: null,
+            error: null,
+        };
+
+        getActionSuffix(status: Status) {
+            return status === Status.SUCCESS || status === Status.ERROR ? 'end' : 'start';
+        }
+
+        handleAnnotationCreate = ({ annotation, error, meta }: AnnotationActionEvent): void => {
```
jstoffan, line 27: Is the suffix needed or can we infer an error if the `error` property is defined?

src/elements/common/annotator-context/withAnnotations.tsx
ConradJChan: Probably not, but it shouldn't hurt and without it, I get an odd eslint error 😛

src/elements/common/annotator-context/withAnnotations.tsx
ConradJChan: As discussed, I think a sequence of operations should be okay. Each discrete event would trigger a value change in the AnnotatorContext which would cause a re-render of any component using the `withAnnotatorContext` hoc. So even if the API to create an annotation took a while, we would see a sequence like: pending create 1, pending create 2, successful create 1, successful create 2, and each state change can be handled by the consuming component. 

Something to look out for is if somehow the events happened so fast that `setState` batches the change together which would result in a loss of data. That happening for a create or update should be fine since missing out on the pending state isn't a big deal but the chances of that happening for creating two annotations back to back seems low

src/elements/common/annotator-context/withAnnotations.tsx
ConradJChan: As discussed, the `AnnotatorState` happens to be the state of the hoc in this case, but any consumer of the context could use the same type as well

src/elements/content-preview/ContentPreview.js
```diff
@@ -741,6 +753,11 @@ class ContentPreview extends React.PureComponent<Props, State> {
...
+            this.preview.addListener('annotationCreate', onAnnotationCreate);
```
mickr, line 758: Do we have to do handle removing this listener?

ConradJChan: I don't think so because when preview is destroyed, we do `preview.removeAllListeners()` which should take care of it. But we may need to add some handling when we care about the annotator instance in the future. Thanks for bringing this up

PR #2065, opened 2020-04-07
chore(explorer): expose staticPath and previewLibraryVersion props

ConradJChan:
Expose `staticPath` and `previewLibraryVersion` props from `ContentExplorer` down to `ContentPreview`


src/elements/content-explorer/ContentExplorer.js
```diff
@@ -133,6 +134,7 @@ type Props = {
     sortBy: SortBy,
     sortDirection: SortDirection,
     staticHost: string,
+    staticPath: String,
```
mxiao6, line 137: Seems like you have to use `string` instead of `String` 😂

PR #2067, opened 2020-04-08
feat: add clear option to BaseSelectField

ChenCodes:
**SingleSelectField**
ezgif com-video-to-gif (25)

**MultiSelectField**
ezgif com-video-to-gif (26)

### DIFF
--- a/i18n/en-US.properties
+++ b/i18n/en-US.properties
@@ -1139,7 +1139,7 @@ boxui.securityControls.webDownloadOwners = Download restricted on web, except Ow
 # Bullet point that summarizes web download restrictions applied to classification, when restriction applies to managed users except Owners/Co-Owners/Editors
 boxui.securityControls.webDownloadOwnersEditors = Download restricted on web, except Owners/Co-Owners/Editors.
 # text shown on the Clear All option in the options list
-boxui.select-field.clearAll = Clear All
+boxui.selectField.clearAll = Clear All
 # Title for "Access Type" menu, in all capital letters
 boxui.share.accessType = ACCESS TYPE
 # Label for a Shared Link permission level
--- a/i18n/en-US.properties
+++ b/i18n/en-US.properties
@@ -286,6 +286,10 @@ be.executeIntegrationOpenWithErrorSubHeader = Please try again later.
 be.expand = Expand
 # Label for face skill section in the preview sidebar
 be.faceSkill = Faces
+# Call-to-action text describing what to do to navigate to specified feedback form
+be.feedbackCtaText = Click to provide feedback
+# Accessible text used to describe the form used for feedback
+be.feedbackFormDescription = Beta Feedback Form
 # File access stats error message
 be.fileAccessStatsErrorHeaderMessage = Something went wrong when fetching the access stats.
 # File classification error message
--- a/i18n/en-US.properties
+++ b/i18n/en-US.properties
@@ -286,10 +286,6 @@ be.executeIntegrationOpenWithErrorSubHeader = Please try again later.
 be.expand = Expand
 # Label for face skill section in the preview sidebar
 be.faceSkill = Faces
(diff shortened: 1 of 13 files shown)


src/components/select-field/BaseSelectField.js
```diff
@@ -195,7 +201,12 @@ class BaseSelectField extends React.Component<Props, State> {
             case ' ':
                 if (activeItemIndex !== -1 && isOpen) {
                     stopDefaultEvent(event);
-                    this.selectOption(activeItemIndex);
+
+                    if (shouldShowClearOption && activeItemIndex === 0) {
+                        this.handleClearClick();
+                    } else {
+                        this.selectOption(activeItemIndex);
+                    }
```
jeeyi, line 209: Nice!

src/components/select-field/BaseSelectField.js
```diff
@@ -350,22 +361,27 @@ class BaseSelectField extends React.Component<Props, State> {
...
             const itemProps: Object = {
-                className: 'select-option',
+                className: classNames('select-option', { 'is-clear-option': isClearOption }),
                 key: index,
                 /* preventDefault on click to prevent wrapping label from re-triggering the select button */
                 onClick: event => {
                     event.preventDefault();
-
-                    this.selectOption(index);
+                    if (index === 0 && shouldShowClearOption) {
```
jeeyi, line 380: Maybe replace with `if (isClearOption)`?

src/components/select-field/MultiSelectField.js
```diff
@@ -8,6 +8,20 @@ type Props = {
     onChange: Function,
 };
 
-const MultiSelectField = ({ ...rest }: Props) => <BaseSelectField {...rest} multiple />;
+const optionsWithClearOption = (options: Array<Object>, shouldShowClearOption: boolean) => {
+    if (shouldShowClearOption) {
+        const updatedOptions = options;
+        updatedOptions.unshift({
+            value: 'clear',
+            displayText: 'Clear All',
```
jeeyi, line 16: Don't we need to localize `Clear All`? Make sure to create `messages.js` and add `Clear All` string there so it can be translated. You should be using something like `<FormattedMessage {...messages.clearAllButton} />` instead of hard-coded string literal.

src/components/select-field/__tests__/BaseSelectField.test.js
```diff
@@ -578,6 +622,17 @@ describe('components/select-field/BaseSelectField', () => {
         });
     });
 
+    describe('handleClearClick', () => {
+        test('should call handleChange with empty array', () => {
+            const wrapper = shallowRenderSelectField();
+            const instance = wrapper.instance();
+
+            sandbox.mock(instance).expects('handleChange');
```
jeeyi, line 630: Don't you need `.withArgs` with empty array?

jeeyi:
Make sure to add `messages.js` with `Clear All` in it so that we can localize this button. Thank you!

src/components/select-field/__tests__/BaseSelectField.test.js
ChenCodes: Yep, will add

i18n/en-US.properties
```diff
@@ -1136,6 +1136,8 @@ boxui.securityControls.webDownloadExternalOwnersEditors = Download restricted on
 boxui.securityControls.webDownloadOwners = Download restricted on web, except Owners/Co-Owners.
 # Bullet point that summarizes web download restrictions applied to classification, when restriction applies to managed users except Owners/Co-Owners/Editors
 boxui.securityControls.webDownloadOwnersEditors = Download restricted on web, except Owners/Co-Owners/Editors.
+# text shown on the Clear All option in the options list
+boxui.select-field.clearAll = Clear All
```
jeeyi, line 1140: Let's use camel case instead of `-` to be consistent. Could you change it to `boxui.selectField.clearAll`?

Issue #1934, opened 2020-02-13
Fix select field component onblur behavior


rwatari:
**Environment:**

- Elements version: 12.0.0

**Desktop (please complete the following information):**

- OS: Mac
- Browser: Safari 13.0.4

**Expected behavior**
Select field component should fire onblur callback when clicking outside of the dropdown after it is opened. The component is based on a div, which shouldn't fire onblur unless it has a tabindex set.
https://github.com/box/box-ui-elements/blob/master/src/components/select-field/BaseSelectField.js#L421

Fix:
Add tabindex to the div or use semantic html like select/option. The 2nd option will improve a11y but may require more testing to ensure the component is styled and behaves correctly.

PR #2073, opened 2020-04-09
fix(select-field): enable onblur for select field

gregory-lyons:
Fixes https://github.com/box/box-ui-elements/issues/1934

Adds a tabIndex to the div component and sets outline:none to avoid
blue border on focus.

Eventually this should be made more accessible, but this fixes the
current bug.

### DIFF
--- a/src/components/select-field/SelectField.scss
+++ b/src/components/select-field/SelectField.scss
@@ -2,6 +2,8 @@
 
 .bdl-SelectField,
 .select-field {
+    outline: none;
+
     .overlay {
         min-width: 100%;
     }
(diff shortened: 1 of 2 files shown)


src/components/select-field/SelectField.scss
```diff
@@ -2,6 +2,8 @@
 
 .bdl-SelectField,
 .select-field {
+    outline: none;
```
tjgupta, line 5: I'm thinking we don't want to do this. There's been a fair bit of effort lately to fix the a11y of our various components and to undo this sort of CSS. Is this needed for something specific?

gregory-lyons:
This is no longer needed, it was fixed in https://github.com/box/box-ui-elements/pull/2075

PR #2074, opened 2020-04-10
feat(usm): add ability to create shared links on modal open

booc0mtaco:


src/features/unified-share-modal/SharedLinkSection.js
```diff
@@ -51,11 +53,43 @@ type Props = {
     triggerCopyOnLoad?: boolean,
 };
 
-class SharedLinkSection extends React.Component<Props> {
+type State = {
+    autoCreatingSharedLink: boolean,
```
alexkrolick, line 57: nit: the boolean naming convention (`is...`)

src/features/unified-share-modal/SharedLinkSection.js
```diff
@@ -51,11 +53,43 @@ type Props = {
...
 
+    constructor(props: Props) {
+        super(props);
+
+        this.state = {
+            autoCreatingSharedLink: false,
+        };
+    }
+
+    componentDidUpdate(prevProps: Props) {
```
alexkrolick, line 74: Is there no possibility that this should handle the `autoCreateSharedLink` prop on mount as well as update? It's rare that one applies but not the other

src/features/unified-share-modal/SharedLinkSection.js
alexkrolick: 🤔 maybe we can lint this in the future

src/features/unified-share-modal/SharedLinkSection.js
booc0mtaco: The lifecycle of the USM is such that the outer component will change the sharedlink and other props after a `getInitialData` has run. This component isn't responsible for creating the shared link, and all of that is delegated to the parent. in short:

* USM mounts, which will render this shared link component
* upon mount, USM will get initial data
* when it gets this data it will know whether it has a shared link or not. based on this, props get passed to the shared link component
* at that time, the component can tell if it should attempt to create one based on the update

@tjgupta I put it here, so that the shared link isn't added prior to the initial UI showing. If i put some of this in USM itself, it would wait until the shared link was generated to show the UI... this would prevent users from seeing the toggle slide, and/or the copied button from activating (eventually)

alexkrolick: Seems kind of fragile to rely on that order remaining the same. For example, the parent could just not render this component until it has data, then mount it with the props having the condition that needs this code to run.

booc0mtaco: i am not going to use hooks for this. if you have a concrete reason why this would fail given how this component currently works, i'm all ears

alexkrolick: I'm not saying rewrite it. But there are no tests or documentation for this component that would indicate why this stops working if called in a slightly different way. It seems like the principle of least surprise would be that we handle this now, the same as if we do for components that could in theory only have componentDidMount and not handle prop updates.

src/features/unified-share-modal/SharedLinkSection.js
```diff
@@ -51,11 +53,43 @@ type Props = {
...
 
+    constructor(props: Props) {
+        super(props);
+
+        this.state = {
+            isAutoCreatingSharedLink: false,
+        };
+    }
+
+    componentDidUpdate(prevProps: Props) {
```
alexkrolick, line 74:
```suggestion
    componentDidMount() {
        const { sharedLink, autoCreateSharedLink, addSharedLink, submitting } = this.props;

        if (
            autoCreateSharedLink &&
            !this.state.isAutoCreatingSharedLink &&
            !sharedLink.url &&
            !submitting &&
            !sharedLink.isNewSharedLink
        ) {
            this.setState({ isAutoCreatingSharedLink: true });
            addSharedLink();
        }
    }
    
    componentDidUpdate(prevProps: Props) { 
```

alexkrolick: If that doesn't work because of something like dummy data being passed in initially:

```suggestion
    // We handle didUpdate but not didMount because
    // the component initially renders with empty data
    // in order to start showing UI components.
    // When getInitialData completes in the parent we
    // rerender with correct sharedLink data and can
    // check whether to auto create a new one.
    componentDidUpdate(prevProps: Props) {
```

booc0mtaco: This makes sense. my initial concern was that this could only cause the the shared link to be added before the rest of the USM was set up. the current USM code will trigger the DidUpdate version. I couldn't coerce it to do it on mount given the props that are being checked, but this might future-proof some refactoring.

PR #2079, opened 2020-04-14
feat(annotations): update url on annotation selection

mickr:
Adding handler to update URL from annotation activity card, and call the callback from contentPreview to handle selecting annotation in Preview

- [x] Tests

### DIFF
--- a/src/elements/common/annotator-context/__tests__/withAnnotatorContext.test.tsx
+++ b/src/elements/common/annotator-context/__tests__/withAnnotatorContext.test.tsx
@@ -28,15 +28,22 @@ describe('elements/common/annotator-context/withAnnotatorContext', () => {
             annotation: { foo: 'bar' },
             action: Action.CREATE_START,
         };
+        const mockEmitAnnotatorChangeEvent = jest.fn();
 
-        mockContext.mockReturnValue(annotatorState);
+        mockContext.mockReturnValue({
+            state: annotatorState,
+            emitActiveChangeEvent: mockEmitAnnotatorChangeEvent,
+        });
 
         const wrapper = getWrapper();
         const wrappedComponent = wrapper.dive().find(Component);
+        const props = wrappedComponent.props() as WrappedComponentProps<HTMLDivElement>;
+
         expect(wrappedComponent.exists()).toBeTruthy();
-        expect((wrappedComponent.props() as WrappedComponentProps<HTMLDivElement>).annotatorState).toEqual({
+        expect(props.annotatorState).toEqual({
             annotation: { foo: 'bar' },
             action: Action.CREATE_START,
         });
+        expect(props.emitAnnotatorActiveChangeEvent).toEqual(mockEmitAnnotatorChangeEvent);
     });
 });
--- a/src/elements/common/annotator-context/types.ts
+++ b/src/elements/common/annotator-context/types.ts
@@ -7,6 +7,7 @@ export enum Action {
 }
 
 export interface Annotator {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
     emit: (event: string | symbol, ...args: any[]) => void;
 }
 
--- a/src/elements/common/annotator-context/types.ts
+++ b/src/elements/common/annotator-context/types.ts
@@ -14,10 +14,13 @@ export interface Annotator {
 export interface AnnotatorState {
     activeAnnotationId?: string | null;
     annotation?: object | null;
-    annotator: Annotator | null;
     action?: Action | null;
     error?: Error | null;
-    setActiveAnnotationId: (annotationId: string) => void;
+}
+
+export interface AnnotatorContext {
+    state: AnnotatorState;
+    emitActiveChangeEvent: (id: string) => void;
 }
 
 export enum Status {
--- a/src/elements/common/annotator-context/types.ts
+++ b/src/elements/common/annotator-context/types.ts
@@ -6,6 +6,10 @@ export enum Action {
     // Can extend to other actions: update_start, update_end, delete_start, delete_end
 }
 
+export interface Annotator {
+    emit: (event: string | symbol, ...args: any[]) => void;
+}
+
 export interface AnnotatorState {
     activeAnnotationId?: string | null;
     annotation?: object | null;
--- a/src/elements/common/annotator-context/types.ts
+++ b/src/elements/common/annotator-context/types.ts
@@ -11,6 +11,7 @@ export interface AnnotatorState {
     annotation?: object | null;
     action?: Action | null;
     error?: Error | null;
+    setActiveAnnotationId: (annotationId: string) => void;
 }
 
 export enum Status {
--- a/src/elements/common/annotator-context/types.ts
+++ b/src/elements/common/annotator-context/types.ts
@@ -14,6 +14,7 @@ export interface Annotator {
 export interface AnnotatorState {
     activeAnnotationId?: string | null;
     annotation?: object | null;
+    annotator: Annotator | null;
     action?: Action | null;
     error?: Error | null;
     setActiveAnnotationId: (annotationId: string) => void;
--- a/src/elements/common/annotator-context/withAnnotations.tsx
+++ b/src/elements/common/annotator-context/withAnnotations.tsx
@@ -13,19 +13,11 @@ export interface ComponentWithAnnotations {
     handleAnnotationChangeEvent: (id: string | null) => void;
     handleAnnotationCreate: (eventData: AnnotationActionEvent) => void;
     handleAnnotatorEvent: ({ event, data }: { event: string; data?: unknown }) => void;
-    handlePreviewDestroy: () => void;
     handleOnAnnotator: (annotator: Annotator) => void;
+    handlePreviewDestroy: () => void;
 }
 
 export type WithAnnotationsComponent<P> = React.ComponentClass<P & WithAnnotationsProps>;
-
-const defaultState: AnnotatorState = {
-    activeAnnotationId: null,
-    annotation: null,
-    action: null,
-    error: null,
-    setActiveAnnotationId: id => {},
-};
 
 export default function withAnnotations<P extends object>(
     WrappedComponent: React.ComponentType<P>,
@@ -35,7 +27,7 @@ export default function withAnnotations<P extends object>(
 
         annotator: Annotator | undefined;
 
-        handleAnnotationChangeEvent = (id: string | null) => {
+        emitActiveChangeEvent = (id: string | null) => {
             if (!this.annotator) {
                 return;
             }
@@ -43,7 +35,15 @@ export default function withAnnotations<P extends object>(
             this.annotator.emit('annotations_active_change', id);
         };
 
-        state = defaultState;
+        defaultState: AnnotatorState = {
+            activeAnnotationId: null,
+            annotation: null,
+            action: null,
+            error: null,
+            setActiveAnnotationId: this.emitActiveChangeEvent,
+        };
+
+        state = this.defaultState;
 
         getAction({ meta: { status }, error }: AnnotationActionEvent): Action {
             return status === Status.SUCCESS || error ? Action.CREATE_END : Action.CREATE_START;
@@ -77,7 +77,8 @@ export default function withAnnotations<P extends object>(
         };
 
         handlePreviewDestroy = (): void => {
-            this.setState(defaultState);
+            this.setState(this.defaultState);
+            this.annotator = undefined;
         };
 
         render(): JSX.Element {
--- a/src/elements/common/annotator-context/withAnnotations.tsx
+++ b/src/elements/common/annotator-context/withAnnotations.tsx
@@ -3,11 +3,13 @@ import AnnotatorContext from './AnnotatorContext';
 import { Action, Annotator, AnnotationActionEvent, AnnotatorState, Status } from './types';
 
 export interface WithAnnotationsProps {
+    onAnnotator: (annotator: Annotator) => void;
     onAnnotatorEvent: ({ event, data }: { event: string; data: unknown }) => void;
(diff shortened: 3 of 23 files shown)


ConradJChan:
Few thoughts on this:

- What are your thoughts on instead of `handleAnnotationSelect` is that we could put a function on the AnnotatorContext similar to how we do connected components in redux:
```
export interface AnnotatorState {
    activeAnnotationId?: string | null;
    annotation?: object;
    action?: Action;
    error?: Error;
    setActiveAnnotationId: (id: string) => void;
}
```
In this way, we can avoid the `onAnnotationSelect` passing down through the component tree.

 - For the deep-linking to the annotation item, we need to handle it being initiated by either box-annotations or clicking the link in the annotation item -- what do you think about only relying on the box-annotations `annotations_active_change` event to trigger a `history.push|replace` to simplify the flow?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -21,8 +21,13 @@ export default function withAnnotations<P extends object>(
     class ComponentWithAnnotations extends React.Component<P & WithAnnotationsProps, AnnotatorState> {
         static displayName: string;
 
+        handleActiveChange = (annotationId: string | null): void => {
+            this.setState({ activeAnnotationId: annotationId });
+        };
+
         state: AnnotatorState = {
             activeAnnotationId: null,
+            setActiveAnnotationId: this.handleActiveChange,
```
ConradJChan, line 30: I think instead of directly setting it back to state, we may want to emit the `annotations_active_set` event back to the annotator, and let that naturally come back as a `annotations_active_change` event which will trigger `handleActiveChange`

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -14,6 +14,7 @@ import AddTaskButton from './AddTaskButton';
 import API from '../../api';
 import messages from '../common/messages';
 import SidebarContent from './SidebarContent';
+import { AnnotatorState, withAnnotatorContext } from '../common/annotator-context';
```
ConradJChan, line 17: Are these changes still needed here?

src/elements/content-sidebar/Sidebar.js
```diff
@@ -102,13 +113,31 @@ class Sidebar extends React.Component<Props, State> {
...
     };
 
+    /**
+     * Handle annotation selection
+     *
+     * @param {string} annotationId - The id of the annotation being selected
+     * @return {void}
+     */
+
+    handleAnnotationSelect = (annotationId: string): void => {
```
ConradJChan, line 134: This may also be `null` so we should handle that case

src/elements/content-sidebar/ActivitySidebar.js
mickr: I could wrap the activity card with the context directly, but as of right now I'm passing it here to the activity feed.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -46,6 +47,7 @@ import './ActivitySidebar.scss';
 type ExternalProps = {
     activeFeedEntryId?: string,
     activeFeedEntryType?: FocusableFeedItemType,
+    annotatorState?: AnnotatorState,
```
ConradJChan, line 50: Could instead do `& WithAnnotatorContextProps` at the end

src/elements/content-sidebar/Sidebar.js
```diff
@@ -102,13 +113,35 @@ class Sidebar extends React.Component<Props, State> {
...
 
+    /**
+     * Handle annotation selection
+     *
+     * @param {string} annotationId - The id of the annotation being selected
+     * @return {void}
+     */
+
+    handleAnnotationSelect = (annotationId: string | null): void => {
+        if (!annotationId) {
```
ConradJChan, line 135: I think if we return here, we leave the url as a deep-linked annotation which wouldn't be accurate. Should it instead set the url to be `/`?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
```diff
@@ -72,6 +72,11 @@
             }
         }
 
+        .bcs-activity-feed-annotation-activity.bcs-is-focused {
+            padding-left: 21px;
```
ConradJChan, line 76: Do we need to override the animation? I think this results in the focused annotation activity background-color fading back to transparent

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -21,8 +21,13 @@ export default function withAnnotations<P extends object>(
     class ComponentWithAnnotations extends React.Component<P & WithAnnotationsProps, AnnotatorState> {
         static displayName: string;
 
+        handleActiveChange = (annotationId: string | null): void => {
```
jstoffan, line 24: Can this live next to the other handler, as before?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -594,6 +596,7 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
             activeFeedEntryId,
             activeFeedEntryType,
             onTaskView,
+            annotatorState = {},
```
jstoffan, line 599: Default values should live in `defaultProps`.

src/elements/content-sidebar/Sidebar.js
```diff
@@ -102,13 +113,35 @@ class Sidebar extends React.Component<Props, State> {
...
+            return;
+        }
+
+        const { history } = this.props;
+
+        const urlPrefix = this.getUrlPrefix(history.location.pathname);
+        history.replace(`/${urlPrefix}/annotations/${annotationId}`);
```
jstoffan, line 142: Do we want to `push` instead of `replace` here?

jstoffan: Is this URL going to suffice for our purposes? Will we be able to determine the version we need to load?

mickr: No, but I talked with the BE team and they have not worked through the versions url yet and what it will look like.

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
mickr: I assumed that we would want to keep the focus behaviors similar to the other cards on the activity feed. I can follow up with design.

src/elements/common/annotator-context/withAnnotations.tsx
mickr: I moved it here because TS was complaining that I attached it to the handler before it was declared. But some other feedback from Conrad for the event design will change this slightly.

src/elements/content-preview/ContentPreview.js
```diff
@@ -83,6 +83,7 @@ type Props = {
     logoUrl?: string,
     measureRef: Function,
     messages?: StringMap,
+    onAnnotator: Function,
```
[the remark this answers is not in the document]
mickr: good catch

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
mickr: Spoke with design and we should keep the background blue, can I follow up with another PR as this one is already getting pretty big?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -59,15 +75,20 @@ export default function withAnnotations<P extends object>(
             }
         };
 
+        handleOnAnnotator = (annotator: Annotator): void => {
+            this.setState({ annotator });
```
jstoffan, line 79: If we're passing the annotator instance around, it seems to defeat the purpose of the other portions of this component. Can we avoid storing the annotator instance?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -59,15 +75,23 @@ export default function withAnnotations<P extends object>(
             }
         };
 
+        handleOnAnnotator = (annotator: Annotator): void => {
+            this.annotator = annotator;
```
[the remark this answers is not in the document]
mickr: Yes, this PR is getting really big so we can follow up quickly

src/elements/common/annotator-context/__tests__/withAnnotatorContext.test.tsx
```diff
@@ -38,5 +42,8 @@ describe('elements/common/annotator-context/withAnnotatorContext', () => {
             annotation: { foo: 'bar' },
             action: Action.CREATE_START,
         });
+        expect(
+            (wrappedComponent.props() as WrappedComponentProps<HTMLDivElement>).emitAnnotatorActiveChangeEvent,
```
jstoffan, line 46: Can we avoid this cast? Or at least avoid repeating it?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -1,37 +1,53 @@
...
         static displayName: string;
 
+        annotator: Annotator | undefined;
+
         state = defaultState;
 
+        emitActiveChangeEvent = (id: string | null) => {
+            const { annotator } = this;
+
+            if (!annotator || !annotator.emit) {
```
jstoffan, line 44: Is this an error condition? Should we handle it somehow other than swallowing it?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -59,15 +75,23 @@ export default function withAnnotations<P extends object>(
             }
         };
 
+        handleOnAnnotator = (annotator: Annotator): void => {
+            this.annotator = annotator;
+        };
+
         handlePreviewDestroy = (): void => {
             this.setState(defaultState);
+            this.annotator = undefined;
```
jstoffan, line 84: If this is a nullable property, use `null` rather than explicitly setting `undefined`.

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -59,15 +75,23 @@ export default function withAnnotations<P extends object>(
...
         };
 
         render(): JSX.Element {
             return (
-                <AnnotatorContext.Provider value={this.state}>
+                <AnnotatorContext.Provider
+                    value={{ state: this.state, emitActiveChangeEvent: this.emitActiveChangeEvent }}
```
jstoffan, line 90: Alphabetize?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -59,15 +75,23 @@ export default function withAnnotations<P extends object>(
...
+                >
                     <WrappedComponent
                         {...this.props}
+                        onAnnotator={this.handleOnAnnotator}
```
jstoffan, line 94: `handleAnnotator` instead of `handleOnAnnotator`

src/elements/content-sidebar/__tests__/Sidebar.test.js
```diff
@@ -12,7 +12,8 @@ jest.mock('../../common/async-load', () => () => 'LoadableComponent');
 jest.mock('../../../utils/LocalStore');
 
 describe('elements/content-sidebar/Sidebar', () => {
-    const getWrapper = props => shallow(<Sidebar file={{ id: 'id' }} location={{ pathname: '/' }} {...props} />);
+    const getWrapper = props =>
+        shallow(<Sidebar file={{ id: 'id' }} annotatorState={{}} location={{ pathname: '/' }} {...props} />);
```
jstoffan, line 16: Should the Sidebar have a defaultProp value for annotatorState?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
```diff
@@ -260,8 +261,8 @@ class ActivityFeed extends React.Component<Props, State> {
                             isDisabled={isDisabled}
                             currentUser={currentUser}
                             onTaskAssignmentUpdate={onTaskAssignmentUpdate}
-                            onAnnotationDelete={onAnnotationDelete}
```
jstoffan, line 263: Is this move needed?

src/elements/common/annotator-context/withAnnotations.tsx
mickr: No, the annotator is loaded through an event so it starts null or undefined, TS throws an error if you don't have a gate for it.

src/elements/common/annotator-context/types.ts
```diff
@@ -6,13 +6,23 @@ export enum Action {
...
 export interface AnnotatorState {
     activeAnnotationId?: string | null;
     annotation?: object | null;
     action?: Action | null;
     error?: Error | null;
 }
 
+export interface AnnotatorContext {
+    state: AnnotatorState;
+    emitActiveChangeEvent: (id: string) => void;
```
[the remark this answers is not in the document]
mickr: I think it is more descriptive with this name, not sure if this detail needs to be abstracted away.

src/elements/content-sidebar/__tests__/Sidebar.test.js
mickr: Yes and I added it

src/elements/common/annotator-context/withAnnotations.tsx
mickr: I can follow up on this

src/elements/common/annotator-context/types.ts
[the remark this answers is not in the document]
ConradJChan: ok then at least make it `emitActiveSetEvent`?

PR #2080, opened 2020-04-14
feat(annotations): handle annotator events

ConradJChan:
TODO:

- [x] unit tests

### DIFF
--- a/src/elements/common/annotator-context/__tests__/withAnnotations.test.tsx
+++ b/src/elements/common/annotator-context/__tests__/withAnnotations.test.tsx
@@ -20,7 +20,7 @@ describe('elements/common/annotator-context/withAnnotations', () => {
 
     const WrappedComponent = withAnnotations(MockComponent);
 
-    const defaultProps = { className: 'foo', onAnnotationCreate: jest.fn() };
+    const defaultProps = { className: 'foo', onAnnotatorEvent: jest.fn() };
 
     const getWrapper = (props: WrappedComponentProps = defaultProps) =>
         shallow<WithAnnotationsComponent<WrappedComponentProps>>(<WrappedComponent {...props} />);
@@ -28,12 +28,12 @@ describe('elements/common/annotator-context/withAnnotations', () => {
     const getContextProvider = (wrapper: ShallowWrapper<WithAnnotationsComponent<WrappedComponentProps>>) =>
         wrapper.find(AnnotatorContext.Provider);
 
-    test('should pass onAnnotationCreate as a prop on the wrapped component', () => {
+    test('should pass onAnnotatorEvent as a prop on the wrapped component', () => {
         const wrapper = getWrapper();
 
         const wrappedComponent = wrapper.find(MockComponent);
         expect(wrappedComponent.exists()).toBeTruthy();
-        expect((wrappedComponent.props() as WrappedComponentProps).onAnnotationCreate).toBeTruthy();
+        expect((wrappedComponent.props() as WrappedComponentProps).onAnnotatorEvent).toBeTruthy();
     });
 
     test('should pass the state on to the AnnotatorContext.Provider', () => {
@@ -42,6 +42,7 @@ describe('elements/common/annotator-context/withAnnotations', () => {
         const contextProvider = getContextProvider(wrapper);
         expect(contextProvider.exists()).toBeTruthy();
         expect((contextProvider.props() as ContextProviderProps).value).toEqual({
+            activeAnnotationId: null,
             annotation: undefined,
             action: undefined,
             error: undefined,
@@ -73,6 +74,7 @@ describe('elements/common/annotator-context/withAnnotations', () => {
                 const contextProvider = getContextProvider(wrapper);
                 expect(contextProvider.exists()).toBeTruthy();
                 expect((contextProvider.props() as ContextProviderProps).value).toEqual({
+                    activeAnnotationId: null,
                     annotation: expectedAnnotation,
                     action: expectedAction,
                     error: expectedError,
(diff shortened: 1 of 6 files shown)


src/elements/common/annotator-context/withAnnotatorContext.tsx
```diff
@@ -9,7 +9,9 @@ export interface WithAnnotatorContextProps {
 export default function withAnnotatorContext<P extends {}>(WrappedComponent: React.ComponentType<P>) {
     return React.forwardRef<React.RefForwardingComponent<React.ComponentType<P>>, P>((props, ref) => (
         <AnnotatorContext.Consumer>
-            {annotatorState => <WrappedComponent ref={ref} {...props} annotatorState={annotatorState} />}
+            {({ activeAnnotationId, ...rest }) => (
```
jstoffan, line 12: Why is this particular property separated out? Should we just spread `{...annotatorState}`, instead?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -20,6 +21,7 @@ export default function withAnnotations<P extends object>(
         static displayName: string;
 
         state: AnnotatorState = {
+            activeAnnotationId: null,
             annotation: undefined,
```
jstoffan, line 25: You should be able to omit these entirely if you add `?` to these properties in the AnnotatorState type.

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -35,10 +34,27 @@ export default function withAnnotations<P extends object>(
             this.setState({ ...this.state, annotation, action, error });
         };
 
+        handleActiveChange = (annotationId: string | null): void => {
+            this.setState({ ...this.state, activeAnnotationId: annotationId });
```
jstoffan, line 38: You should be able to omit `...this.state` here. React will merge the state automatically.

src/elements/common/annotator-context/types.ts
```diff
@@ -24,7 +24,7 @@ export interface Metadata {
 }
 
 export interface AnnotationActionEvent {
-    annotation: object | undefined;
-    error: Error | undefined;
+    annotation?: object;
```
mickr, line 27: Should there be a more precise type for this?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -20,9 +21,7 @@ export default function withAnnotations<P extends object>(
         static displayName: string;
 
         state: AnnotatorState = {
-            annotation: undefined,
-            action: undefined,
-            error: undefined,
```
mickr, line 25: Should these be removed? We are still saving the annotation payload here so the properties should still exist in the default state?

src/elements/common/annotator-context/withAnnotations.tsx
ConradJChan: Interesting thing is, if I omit `...this.state`, eslint starts complaining about `no-unused-state` warning (https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unused-state.md)

src/elements/common/annotator-context/types.ts
ConradJChan: Yeah, but I figure I'm waiting for `box-annotations` to be imported so I can use those types

src/elements/common/annotator-context/withAnnotations.tsx
ConradJChan: Yeah, but they're all optional as @jstoffan pointed out, so removing them should be ok

src/elements/common/annotator-context/types.ts
```diff
@@ -7,10 +7,10 @@ export enum Action {
 }
 
 export interface AnnotatorState {
-    activeAnnotationId?: string;
-    annotation?: object | undefined;
-    action: Action | undefined;
-    error: Error | undefined;
+    activeAnnotationId?: string | null;
```
mxiao6, line 10: Will removing the null here and all the null in withAnnotations.tsx work?

ConradJChan: We want `null` though because that's a valid assignment for `activeAnnotationId` which represents that there is no active annotation id

src/elements/common/annotator-context/__tests__/withAnnotations.test.tsx
```diff
@@ -73,11 +74,51 @@ describe('elements/common/annotator-context/withAnnotations', () => {
...
+            'should call appropriate handler based on event $event',
+            ({ event, numCreateCalled, numActiveChangeCalled }) => {
+                const wrapper = getWrapper();
+                const instance = wrapper.instance() as Component & ComponentWithAnnotations;
+                instance.handleAnnotationCreate = jest.fn();
+                instance.handleActiveChange = jest.fn();
+
+                instance.handleAnnotatorEvent({ event });
+
+                expect((instance.handleAnnotationCreate as jest.Mock).mock.calls.length).toBe(numCreateCalled);
```
jstoffan, line 119: Is this casting needed? TypeScript should be able to infer it based on the assignment above.

src/elements/common/annotator-context/__tests__/withAnnotations.test.tsx
```diff
@@ -42,6 +42,7 @@ describe('elements/common/annotator-context/withAnnotations', () => {
         const contextProvider = getContextProvider(wrapper);
         expect(contextProvider.exists()).toBeTruthy();
         expect((contextProvider.props() as ContextProviderProps).value).toEqual({
+            activeAnnotationId: null,
             annotation: undefined,
```
jstoffan, line 46: Are these values still explicitly undefined or can they be removed here as well?

src/elements/common/annotator-context/__tests__/withAnnotations.test.tsx
ConradJChan: Yeah it's not inferred for some reason

PR #1854, opened 2020-01-15
feat(metadata-views): Add pagination to metadata view

manishPh:

PR #2081, opened 2020-04-15
fix(content-picker): Fix pagination and add tests

sahiga:
This PR fixes pagination in the Content Picker. The main changes are:
* Renaming the change function prop that the Content Picker passes to Pagination (`onChange` &#8594; `onOffsetChange`)
* Adding a unit test for Pagination and two pagination-related E2E tests for the Content Picker

### DIFF
--- a/src/elements/content-picker/ContentPicker.js
+++ b/src/elements/content-picker/ContentPicker.js
@@ -1235,7 +1235,7 @@ class ContentPicker extends Component<Props, State> {
                         >
                             <Pagination
                                 offset={offset}
-                                onChange={this.paginate}
+                                onOffsetChange={this.paginate}
                                 pageSize={currentPageSize}
                                 totalCount={totalCount}
                             />
--- a/src/elements/content-picker/ContentPicker.js
+++ b/src/elements/content-picker/ContentPicker.js
@@ -17,7 +17,7 @@ import UploadDialog from '../common/upload-dialog';
 import CreateFolderDialog from '../common/create-folder-dialog';
 import Internationalize from '../common/Internationalize';
 import makeResponsive from '../common/makeResponsive';
-import Pagination from '../../features/pagination';
+import OffsetBasedPagination from '../../features/pagination/OffsetBasedPagination';
 import { isFocusableElement, isInputElement, focus } from '../../utils/dom';
 import API from '../../api';
 import Content from './Content';
@@ -1233,7 +1233,7 @@ class ContentPicker extends Component<Props, State> {
                             chooseButtonLabel={chooseButtonLabel}
                             cancelButtonLabel={cancelButtonLabel}
                         >
-                            <Pagination
+                            <OffsetBasedPagination
                                 offset={offset}
                                 onOffsetChange={this.paginate}
                                 pageSize={currentPageSize}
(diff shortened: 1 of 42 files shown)


src/elements/content-picker/ContentPicker.js
```diff
@@ -1235,7 +1235,7 @@ class ContentPicker extends Component<Props, State> {
                         >
                             <Pagination
                                 offset={offset}
-                                onChange={this.paginate}
+                                onOffsetChange={this.paginate}
```
jstoffan, line 1238: Originally, the `Pagination` component only handled offset-based paging, which is largely incompatible with marker-based paging. Why were they merged into a single component? Perhaps just use the OffsetBasedPagination component directly?

manishPh: @jstoffan They are not merged. In fact, the earlier `<Pagination />` component only supported offset-based pagination. As part of https://github.com/box/box-ui-elements/pull/1854, we added ability for marker based pagination which resulted in `< OffsetBasedPagination  />` and `<MarkerBasedPagination />`. `<Pagination />` was just refactored to be a simple wrapper around these.

sahiga: @manishPh @jstoffan I updated this PR to use `OffsetBasedPagination` instead of `Pagination`. Could you review again?

test/integration/content-picker/ContentPicker.e2e.test.js
```diff
@@ -34,32 +34,142 @@ const helpers = {
 
         return cy.get('@row');
     },
+    getPaginationArrows: () => {
+        return cy
+            .getByTestId('content-picker')
+            .find('.bdl-Pagination-nav')
```
jstoffan, line 40: Consider using user-facing text or `data-testid` attributes instead of CSS class names for selectors.

test/integration/content-picker/ContentPicker.e2e.test.js
```diff
@@ -34,32 +34,142 @@ const helpers = {
...
-            cy.fixture('content-picker/get-sharedlink.json').then(getSharedLinkJson => {
-                cy.route('GET', `**/files/${fileId}?fields=allowed_shared_link_access_levels,shared_link`, {
-                    ...getSharedLinkJson,
-                    id: fileId,
+            // Confirm that the Content Picker now shows the first page
+            helpers.getRow(0).contains(FIRST_ITEM_OF_FIRST_PAGE);
+        });
+
+        it('Should be able to navigate to the next page using count button', () => {
+            const COUNT_BUTTON_ON_FIRST_PAGE = '1 of 2';
```
jstoffan, line 96: Should this use an i18n helper? I think the button strings are translated.

test/integration/content-picker/ContentPicker.e2e.test.js
```diff
@@ -34,32 +34,142 @@ const helpers = {
...
+            helpers.getRow(0).contains(FIRST_ITEM_OF_SECOND_PAGE);
+
+            // Stub call to first page
+            cy.route('GET', '**/folders/*', FIRST_PAGE);
+
+            // Confirm that the pagination count button says "2 of 2", and click it
+            helpers
+                .getPaginationCountButton()
+                .contains(COUNT_BUTTON_ON_SECOND_PAGE)
+                .then($el => {
```
jstoffan, line 132: Does `.click()` work here?

PR #2083, opened 2020-04-16
feat(annotations): reset annotator context when destroying preview

ConradJChan:
Todo:
 - [x] unit tests

### DIFF
--- a/src/elements/common/annotator-context/withAnnotations.tsx
+++ b/src/elements/common/annotator-context/withAnnotations.tsx
@@ -4,6 +4,7 @@ import { Action, AnnotationActionEvent, AnnotatorState, Status } from './types';
 
 export interface WithAnnotationsProps {
     onAnnotatorEvent: ({ event, data }: { event: string; data: unknown }) => void;
+    onPreviewDestroy: () => void;
 }
 
 export interface ComponentWithAnnotations {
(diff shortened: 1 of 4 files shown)


src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -52,10 +54,23 @@ export default function withAnnotations<P extends object>(
             }
         };
 
+        handlePreviewDestroy = (): void => {
+            this.setState({
+                activeAnnotationId: null,
+                annotation: undefined,
```
jstoffan, line 60: Should these be nullable fields, instead? It seems weird to explicitly define something as undefined.

src/elements/content-preview/ContentPreview.js
```diff
@@ -281,10 +282,12 @@ class ContentPreview extends React.PureComponent<Props, State> {
      * Cleans up the preview instance
      */
     destroyPreview() {
+        const { onPreviewDestroy } = this.props;
         if (this.preview) {
             this.preview.destroy();
             this.preview.removeAllListeners();
             this.preview = undefined;
+            onPreviewDestroy();
```
jstoffan, line 290: Should this happen regardless of whether the instance is defined?

PR #2087, opened 2020-04-18
feat(annotations): Move annotator event listener

mickr:
* Move event listener setup to withAnnotations
* Update ContentPreview to only listen for annotator event
* Update tests and types
* Add eslint rule for explicit-any on rest args only

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -32,6 +32,7 @@ module.exports = {
             files: ['*.ts', '*.tsx'],
             rules: {
                 '@typescript-eslint/explicit-function-return-type': 'off', // fixme
+                '@typescript-eslint/no-explicit-any': ["error", { "ignoreRestArgs": true }],
             }
         }
     ]
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -32,7 +32,6 @@ module.exports = {
             files: ['*.ts', '*.tsx'],
             rules: {
                 '@typescript-eslint/explicit-function-return-type': 'off', // fixme
-                '@typescript-eslint/no-explicit-any': ["error", { "ignoreRestArgs": true }],
             }
         }
     ]
--- a/src/elements/common/annotator-context/__tests__/withAnnotations.test.tsx
+++ b/src/elements/common/annotator-context/__tests__/withAnnotations.test.tsx
@@ -16,8 +16,8 @@ type ContextProviderProps = {
 };
 
 describe('elements/common/annotator-context/withAnnotations', () => {
+    let mockAnnotator = {} as Annotator;
     const MockComponent = (props: ComponentProps) => <div {...props} />;
-
     const WrappedComponent = withAnnotations(MockComponent);
 
     const defaultProps = {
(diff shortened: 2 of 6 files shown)


src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -75,8 +74,9 @@ export default function withAnnotations<P extends object>(
             }
         };
 
-        handleOnAnnotator = (annotator: Annotator): void => {
+        handleAnnotator = (annotator: Annotator): void => {
             this.annotator = annotator;
+            this.annotator.addListener('annotatorevent', this.handleAnnotatorEvent);
```
jstoffan, line 79: You can add listeners for the various events directly rather than listening for `annotatorevent`:

```
this.annotator.addListener('annotations_active_change', this. handleActiveChange)
this.annotator.addListener('annotations_create', this.handleAnnotationsCreate)
```

Also, what happens when the preview instance is destroyed? Should we call removeEventListener in the handlePreviewDestroy method?

mickr: Preview calls removeAllListeners on the annotator instance

jstoffan: That seems like kind of an indirect thing to rely upon. Should we explicitly remove our handlers here, as well, for completeness and safety's sake?

mickr: Sure, I brought it up earlier and thought it was what we were going to rely on, but I'll add it here 👍

ConradJChan: Yeah Preview does call `removeAllListeners`, but it seems like from the box-annotations sdk perspective, we should just clean up what we setup instead of relying on a client to clean up for us

.eslintrc.js
```diff
@@ -32,6 +32,7 @@ module.exports = {
             files: ['*.ts', '*.tsx'],
             rules: {
                 '@typescript-eslint/explicit-function-return-type': 'off', // fixme
+                '@typescript-eslint/no-explicit-any': ["error", { "ignoreRestArgs": true }],
```
ConradJChan, line 35: Maybe separate this out of this PR?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -75,12 +74,18 @@ export default function withAnnotations<P extends object>(
...
         };
 
         handlePreviewDestroy = (): void => {
             this.setState(defaultState);
+
+            if (this.annotator) {
+                this.annotator.removeAllListeners();
```
jstoffan, line 86: We should avoid using `removeAllListeners` in nearly all circumstances, since it can cause side effects for other components. We know what listeners we added, so we can remove them explicitly.

src/elements/common/annotator-context/__tests__/withAnnotations.test.tsx
```diff
@@ -151,15 +152,24 @@ describe('elements/common/annotator-context/withAnnotations', () => {
 
     describe('handlePreviewDestroy()', () => {
         test('should reset state and annotator', () => {
+            const mockAnnotator: Annotator = {
```
jstoffan, line 155: Can we reuse this mock rather than redeclaring it?

PR #2090, opened 2020-04-21
feat(annotations): add new annotation to sidebar

mickr:
* check for new action in annotatorState, start creation on create_start and update on create_end
* Update annotatoState type
* Update test
* Add new method to Feed.js to handle annotation creation

- [x]  Unit tests


src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -54,7 +56,9 @@ export default function withAnnotations<P extends object>(
         handleAnnotationCreate = (eventData: AnnotationActionEvent): void => {
             const { annotation = null, error = null } = eventData;
             const action = this.getAction(eventData);
-            this.setState({ ...this.state, annotation, action, error });
+            const feedId = action === Action.CREATE_START ? uniqueId('annotation_') : this.state.feedId;
```
ConradJChan, line 59: It seems redundant to create another uniqueId since the `annotation` itself has one. Is there a way we could leverage that? Maybe another reason for that is there is a corner case where the API to create is taking a while and in the meantime the user creates another annotation. In this case, the transient feedId be lost

mickr: Hmm, I think so we can update annotations to send the id on create_start & create_end then update create_end to have the real id?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -52,9 +54,12 @@ export default function withAnnotations<P extends object>(
...
 
         handleAnnotationCreate = (eventData: AnnotationActionEvent): void => {
-            const { annotation = null, error = null } = eventData;
+            const { annotation = null, error = null, meta = null } = eventData;
+
             const action = this.getAction(eventData);
-            this.setState({ ...this.state, annotation, action, error });
+            const isPending = action === Action.CREATE_START;
+
+            this.setState({ ...this.state, annotation, action, error, isPending, meta });
```
jstoffan, line 62: This is essentially storing the `action` value in state twice, once as a string and once as a boolean. Can we simplify?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -52,9 +54,12 @@ export default function withAnnotations<P extends object>(
         }
 
         handleAnnotationCreate = (eventData: AnnotationActionEvent): void => {
-            const { annotation = null, error = null } = eventData;
+            const { annotation = null, error = null, meta = null } = eventData;
```
mxiao6, line 57: The default value of `meta` can be `{}` then you don't have to do `const { requestId } = meta || {};` in the next file.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -123,6 +124,35 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         this.fetchCurrentUser(currentUser);
     }
 
+    componentDidUpdate({ annotatorState: prevAnnotatorState }: Props): void {
+        const { action: prevAnnotatorAction } = prevAnnotatorState;
+        const {
+            annotatorState: { action: annotatorAction },
+        } = this.props;
+
+        if (annotatorAction !== prevAnnotatorAction) {
```
ConradJChan, line 133: Should we also check something unique to the annotation? Like maybe the unique id?

src/api/Feed.js
```diff
@@ -113,6 +122,45 @@ class Feed extends Base {
...
+                id,
+                type: 'annotation',
+            };
+
+            this.addPendingItem(this.file.id, currentUser, newAnnotation);
+
+            return;
+        }
+        // Create action has completed, so update the existing pending item
+        this.updateFeedItem({ ...annotation, isPending: false }, id);
```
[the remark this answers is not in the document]
mickr: No, it is stylistic. I just think it reads better.

PR #2091, opened 2020-04-21
feat(preview): use uploader_display_name for sidebar uploader info

omgrief:


src/elements/content-sidebar/SidebarFileProperties.js
```diff
@@ -55,7 +55,7 @@ const SidebarFileProperties = ({
                     : {}
             }
             size={getFileSize(file.size, intl.locale)}
-            uploader={getProp(file, 'created_by.name')}
+            uploader={getProp(file, 'uploader_display_name')}
```
ConradJChan, line 58: Is the `uploader_display_name` always populated even in non forms use cases? Or should it be a combination of `created_by.name` and `uploader_display_name`?

omgrief: ah, good catch... let me reword the logic

omgrief: @ConradJChan addressed

src/elements/content-sidebar/SidebarFileProperties.js
```diff
@@ -55,7 +55,12 @@ const SidebarFileProperties = ({
                     : {}
             }
             size={getFileSize(file.size, intl.locale)}
-            uploader={getProp(file, 'created_by.name')}
+            // use uploader_display_name if uploaded anonymously
+            uploader={
+                getProp(file, 'created_by.id') === 2
```
ConradJChan, line 60: Is it string '2' or number 2? Could we reference a constant like https://github.com/box/box-ui-elements/blob/9fd8627c260146dbcb4e9441df068877f34f2d3c/src/constants.js#L382 to avoid magic numbers?

omgrief: addressed @ConradJChan

PR #2092, opened 2020-04-21
fix(folder-upload-node): fix subfolder name conflict errors

danubevictoria:
Omits incorrect setting of subfolder name conflict errors so that files
in subfolders can continue uploading after being added to the upload
queue.

If there are files/folders in error, the view would be set to
`VIEW_ERROR` and any subsequent uploads would be stopped until the upload
errors were resolved. In this case, subfolders will always hit a name
conflict so we shouldn't block uploading subfiles, and we want view to
be set to `VIEW_UPLOAD_IN_PROGRESS` to continue automatic uploads in
`ContentUploader.updateViewAndCollection`.

Original Bug:
image

Valid Subfolder Errors continue to block subfile uploads:
image

### DIFF
--- a/src/api/uploads/FolderUploadNode.js
+++ b/src/api/uploads/FolderUploadNode.js
@@ -147,7 +147,7 @@ class FolderUploadNode {
             progress: 100,
         };
 
-        if (errorEncountered) {
+        if (errorEncountered && errorCode !== ERROR_CODE_ITEM_NAME_IN_USE) {
             folderObject.status = STATUS_ERROR;
             folderObject.error = { code: errorCode };
         }
--- a/src/api/uploads/__tests__/FolderUploadNode.test.js
+++ b/src/api/uploads/__tests__/FolderUploadNode.test.js
@@ -116,6 +116,26 @@ describe('api/uploads/FolderUploadNode', () => {
             expect(folderUploadNodeInstance.folderId).toBe(folderId);
         });
 
+        test('should not set error for non-root folders if error is ITEM_NAME_IN_USE', async () => {
+            const folderId = '1';
+            const errorCallback = jest.fn();
+            const isRoot = false;
+            const error = {
+                code: ERROR_CODE_ITEM_NAME_IN_USE,
+                context_info: { conflicts: [{ id: folderId }] },
+            };
+            folderUploadNodeInstance.name = name;
+            folderUploadNodeInstance.createFolder = jest.fn(() => ({
+                id: folderId,
+            }));
+            folderUploadNodeInstance.addFolderToUploadQueue = jest.fn();
+
+            await folderUploadNodeInstance.createAndUploadFolder(errorCallback, isRoot);
+
+            expect(errorCallback).not.toHaveBeenCalledWith(error);
+            expect(folderUploadNodeInstance.error).toBeUndefined();
+        });
+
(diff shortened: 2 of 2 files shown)


src/api/uploads/__tests__/FolderUploadNode.test.js
```diff
@@ -116,6 +116,26 @@ describe('api/uploads/FolderUploadNode', () => {
...
+            folderUploadNodeInstance.name = name;
+            folderUploadNodeInstance.createFolder = jest.fn(() => ({
+                id: folderId,
+            }));
+            folderUploadNodeInstance.addFolderToUploadQueue = jest.fn();
+
+            await folderUploadNodeInstance.createAndUploadFolder(errorCallback, isRoot);
+
+            expect(errorCallback).not.toHaveBeenCalledWith(error);
+            expect(folderUploadNodeInstance.error).toBeUndefined();
```
booc0mtaco, line 136: in this case, it _should_ call `addFolderToUploadQueue`?

danubevictoria: yup we are testing that with the test below this

src/api/uploads/__tests__/FolderUploadNode.test.js
```diff
@@ -116,6 +116,26 @@ describe('api/uploads/FolderUploadNode', () => {
...
+        test('should not set error for non-root folders if error is ITEM_NAME_IN_USE', async () => {
+            const folderId = '1';
+            const errorCallback = jest.fn();
+            const isRoot = false;
+            const error = {
+                code: ERROR_CODE_ITEM_NAME_IN_USE,
+                context_info: { conflicts: [{ id: folderId }] },
+            };
+            folderUploadNodeInstance.name = name;
+            folderUploadNodeInstance.createFolder = jest.fn(() => ({
```
booc0mtaco, line 128: this prevents the test from hitting the catch block in the code. it should use a promise that either resolves or rejects. In this case it should reject with `error` to mock the code being tested.

src/api/uploads/__tests__/FolderUploadNode.test.js
booc0mtaco, line 136: the goal of this test is to make sure we call the right things (with the right values) so we would want to make sure both errors are tested for

src/api/uploads/__tests__/FolderUploadNode.test.js
```diff
@@ -116,6 +116,26 @@ describe('api/uploads/FolderUploadNode', () => {
...
+                id: folderId,
+            }));
+            folderUploadNodeInstance.addFolderToUploadQueue = jest.fn();
```
booc0mtaco, line 131: this test does not use this stub to assert anything. we should have an `expect` using this.

src/api/uploads/FolderUploadNode.js
```diff
@@ -147,7 +147,7 @@ class FolderUploadNode {
             progress: 100,
         };
 
-        if (errorEncountered) {
+        if (errorEncountered && errorCode !== ERROR_CODE_ITEM_NAME_IN_USE) {
```
booc0mtaco, line 150: since we do not want to treat "name in use" as an error, we should not set the values in that case. this change should not happen here; instead we should evaluate the try-catch ti set the scope variables appropriately.

PR #2093, opened 2020-04-21
fix(content-explorer): Align copy button in ShareDialog

sahiga:
The `ShareDialog` copy button sits below the input field in languages with longer words for "copy" (e.g., "kopieren" in German). This PR replaces the fixed width of the input field with flexbox-calculated values, which ensures that the copy button is vertically aligned with the input field, and introduces the [intl Storybook addon](https://github.com/truffls/storybook-addon-intl), which enables previews of components in different languages.

ce-copy-button


src/elements/content-explorer/ShareDialog.scss
```diff
@@ -36,5 +42,9 @@
     .be-modal-btns {
         align-items: center;
         justify-content: space-between;
+
+        .btn {
```
booc0mtaco, line 46: is there a minimum width we should consider? if the text for the button is 1-3 characters, the button can become a small click target (example: 复制)

sahiga: @booc0mtaco Good catch. I updated the PR and the screenshots. Can you review again?

PR #2097, opened 2020-04-23
Fix duplicate clear options

ChenCodes:
-use cloneDeep to make sure that we aren't updating the passed in object

### DIFF
--- a/src/components/select-field/__tests__/MultiSelectField.test.js
+++ b/src/components/select-field/__tests__/MultiSelectField.test.js
@@ -9,13 +9,9 @@ describe('components/select-field/MultiSelectField', () => {
         { displayText: 'Baz', value: 'baz' },
     ];
 
-    const intl = {
-        formatMessage: jest.fn().mockReturnValue('Clear All'),
-    };
-
     describe('MultiSelectField', () => {
         test('should render a BaseSelectField with a selectedValues prop matching passed in selected value when called', () => {
-            const wrapper = shallow(<MultiSelectFieldBase intl={intl} onChange={() => {}} options={options} />);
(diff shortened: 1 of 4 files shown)


src/components/select-field/__tests__/MultiSelectField.test.js
```diff
@@ -10,7 +10,7 @@ describe('components/select-field/MultiSelectField', () => {
     ];
 
     const intl = {
-        formatMessage: jest.fn(),
+        formatMessage: jest.fn().mockReturnValue('Clear All'),
```
booc0mtaco, line 13: This should probably be mocked closer to where we expect this value instead of for everything. also consider using [`mockImplementationOnce`](https://jestjs.io/docs/en/mock-function-api.html#mockfnmockimplementationoncefn)

src/components/select-field/__tests__/MultiSelectField.test.js
```diff
@@ -19,23 +19,25 @@ describe('components/select-field/MultiSelectField', () => {
...
             const baseSelectFieldWrapper = wrapper.find('BaseSelectField');
             expect(baseSelectFieldWrapper.length).toBe(1);
-            expect(baseSelectFieldWrapper.prop('options')).toBe(options);
+            expect(baseSelectFieldWrapper.prop('options')).toEqual(options);
             expect(baseSelectFieldWrapper.prop('multiple')).toBe(true);
         });
 
         test('should render a BaseSelectField with an options prop containing a clear option if shouldShowClearOption is true', () => {
             const wrapper = shallow(
                 <MultiSelectFieldBase intl={intl} onChange={() => {}} options={options} shouldShowClearOption />,
```
booc0mtaco, line 28: since we pass in `intl` you can create a test-specific mock here.

PR #2100, opened 2020-04-24
feat(annotations): Reflect deeplink annotations on hard load

ConradJChan:
Todo:
 -  [x] tests

### DIFF
--- a/src/elements/common/annotator-context/withAnnotations.tsx
+++ b/src/elements/common/annotator-context/withAnnotations.tsx
@@ -1,4 +1,7 @@
 import * as React from 'react';
+import getProp from 'lodash/get';
+import { matchPath } from 'react-router-dom';
+import { History } from 'history';
 import AnnotatorContext from './AnnotatorContext';
 import { Action, Annotator, AnnotationActionEvent, AnnotatorState, Status } from './types';
 
@@ -28,15 +31,41 @@ const defaultState: AnnotatorState = {
     meta: null,
 };
 
+type WithHistoryProps = {
+    history?: History;
+};
+
+type MatchParams = {
+    fileVersionId?: string;
+    annotationId?: string;
+};
+
+type ResultProps<P> = P & WithAnnotationsProps & WithHistoryProps;
+
 export default function withAnnotations<P extends object>(
     WrappedComponent: React.ComponentType<P>,
 ): WithAnnotationsComponent<P> {
-    class ComponentWithAnnotations extends React.Component<P & WithAnnotationsProps, AnnotatorState> {
+    class ComponentWithAnnotations extends React.Component<ResultProps<P>, AnnotatorState> {
         static displayName: string;
 
         annotator: Annotator | null = null;
 
-        state = defaultState;
+        constructor(props: ResultProps<P>) {
+            super(props);
+
+            // Determine by url if there is already a deeply linked annotation
+            const { history } = props;
+            const pathname = getProp(history, 'location.pathname');
+            const match = matchPath<MatchParams>(pathname, {
+                path: '/activity/annotations/:fileVersionId/:annotationId',
+                exact: true,
+                strict: false,
+            });
+            const { params: { annotationId: activeAnnotationId = null } = {} } = match || {};
+
+            // Seed the initial state with the activeAnnotationId if any from the location path
+            this.state = { ...defaultState, activeAnnotationId };
+        }
 
         emitActiveChangeEvent = (id: string | null) => {
             const { annotator } = this;
--- a/src/elements/common/annotator-context/withAnnotations.tsx
+++ b/src/elements/common/annotator-context/withAnnotations.tsx
@@ -58,7 +58,6 @@ export default function withAnnotations<P extends object>(
             const match = matchPath<MatchParams>(pathname, {
                 path: '/:sidebar/annotations/:fileVersionId/:annotationId',
                 exact: true,
-                strict: false,
             });
             const { params: { annotationId: activeAnnotationId = null } = {} } = match || {};
 
--- a/src/elements/common/annotator-context/withAnnotations.tsx
+++ b/src/elements/common/annotator-context/withAnnotations.tsx
@@ -36,8 +36,8 @@ type WithHistoryProps = {
 };
 
 type MatchParams = {
-    fileVersionId?: string;
     annotationId?: string;
+    fileVersionId?: string;
 };
 
 type ResultProps<P> = P & WithAnnotationsProps & WithHistoryProps;
--- a/src/elements/common/annotator-context/withAnnotations.tsx
+++ b/src/elements/common/annotator-context/withAnnotations.tsx
@@ -57,7 +57,7 @@ export default function withAnnotations<P extends object>(
             const { history } = props;
             const pathname = getProp(history, 'location.pathname');
             const match = matchPath<MatchParams>(pathname, {
-                path: '/activity/annotations/:fileVersionId/:annotationId',
+                path: '/:sidebar/annotations/:fileVersionId/:annotationId',
                 exact: true,
                 strict: false,
             });
--- a/src/elements/common/annotator-context/withAnnotations.tsx
+++ b/src/elements/common/annotator-context/withAnnotations.tsx
@@ -127,6 +127,7 @@ export default function withAnnotations<P extends object>(
                 >
                     <WrappedComponent
                         {...this.props}
+                        annotatorState={this.state}
                         onAnnotator={this.handleAnnotator}
                         onPreviewDestroy={this.handlePreviewDestroy}
                     />
--- a/src/elements/common/annotator-context/withAnnotations.tsx
+++ b/src/elements/common/annotator-context/withAnnotations.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react';
 import getProp from 'lodash/get';
 import { matchPath } from 'react-router-dom';
-import { History } from 'history';
 import AnnotatorContext from './AnnotatorContext';
 import { Action, Annotator, AnnotationActionEvent, AnnotatorState, Status } from './types';
 
--- a/src/elements/content-preview/ContentPreview.js
+++ b/src/elements/content-preview/ContentPreview.js
@@ -290,9 +290,9 @@ class ContentPreview extends React.PureComponent<Props, State> {
             this.preview.destroy();
             this.preview.removeAllListeners();
             this.preview = undefined;
-        }
 
-        onPreviewDestroy();
+            onPreviewDestroy();
+        }
 
         this.setState({ selectedVersion: undefined });
     }
--- a/src/elements/content-preview/ContentPreview.js
+++ b/src/elements/content-preview/ContentPreview.js
@@ -14,6 +14,7 @@ import omit from 'lodash/omit';
 import getProp from 'lodash/get';
 import flow from 'lodash/flow';
 import noop from 'lodash/noop';
+import setProp from 'lodash/set';
 import Measure from 'react-measure';
 import type { RouterHistory } from 'react-router-dom';
 import { decode } from '../../utils/keys';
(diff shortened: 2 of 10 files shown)


src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -405,14 +441,23 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+    fetchFeedItems(
+        shouldRefreshCache: boolean = false,
+        shouldDestroy: boolean = false,
+        successCallback: (feedItems: FeedItems) => void = noop,
+        errorCallback: (feedItems: FeedItems) => void = noop,
+    ) {
         const { file, api, features } = this.props;
         const shouldShowAppActivity = isFeatureEnabled(features, 'activityFeed.appActivity.enabled');
+        const successWrapper = feedItems =>
+            [successCallback, this.fetchFeedItemsSuccessCallback].forEach(callback => callback(feedItems));
```
jstoffan, line 453: This seems a little unfortunate. Should we introduce promises or something rather than relying on callbacks for everything?

src/elements/content-sidebar/Sidebar.js
```diff
@@ -44,16 +43,14 @@ type Props = {
...
     isDefaultOpen?: boolean,
     isLoading?: boolean,
-    location: Location,
     metadataEditors?: Array<MetadataEditor>,
     metadataSidebarProps: MetadataSidebarProps,
     onVersionChange?: Function,
     onVersionHistoryClick?: Function,
     versionsSidebarProps: VersionsSidebarProps,
-};
+} & RouteComponentProps;
```
jstoffan, line 53: Oh, nice, I didn't know about this.

src/elements/content-sidebar/Sidebar.js
```diff
@@ -133,9 +130,12 @@ class Sidebar extends React.Component<Props, State> {
      */
 
     handleAnnotationSelect = (annotationId: string | null): void => {
-        const { history } = this.props;
+        const { file, history } = this.props;
         const urlPrefix = this.getUrlPrefix(history.location.pathname);
-        const annotationUrl = annotationId ? `/${urlPrefix}/annotations/${annotationId}` : `/${urlPrefix}`;
+        const currentFileVersionId = file.file_version.id;
```
jstoffan, line 135: Consider using destructuring here to avoid the rename-only reassignment.

src/elements/content-sidebar/Sidebar.js
```diff
@@ -133,9 +130,12 @@ class Sidebar extends React.Component<Props, State> {
      */
 
     handleAnnotationSelect = (annotationId: string | null): void => {
```
jstoffan, line 132: Should this move into ActivitySidebar? It seems odd for it to live in this file while other aspects live further down.

src/elements/content-sidebar/SidebarPanels.js
```diff
@@ -162,22 +162,26 @@ class SidebarPanels extends React.Component<Props> {
                         exact
                         path={[
                             `/${SIDEBAR_VIEW_ACTIVITY}`,
+                            `/${SIDEBAR_VIEW_ACTIVITY}/annotations/:fileVersionId/:annotationId`,
                             `/${SIDEBAR_VIEW_ACTIVITY}/:activeFeedEntryType(annotations|comments|tasks)/:activeFeedEntryId?`,
```
jstoffan, line 166: Should annotations still be in this list? Could we do something like this to avoid the need for the annotation-specific logic in the render function below?

```
path={[
    `/${SIDEBAR_VIEW_ACTIVITY}`,
    `/${SIDEBAR_VIEW_ACTIVITY}/:activeFeedEntryType(annotations)/:fileVersionId/:activeFeedEntryId?`,                        
    `/${SIDEBAR_VIEW_ACTIVITY}/:activeFeedEntryType(comments|tasks)/:activeFeedEntryId?`,
]}

File: src/elements/content-sidebar/ActivitySidebar.js [hunk 6]
```diff
@@ -199,6 +204,37 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+        // If annotatorState does not have an activeAnnotationId but the router path has an annotationId,
+        // then we are dealing with a hard load of a deeply linked annotation
+        if (activeAnnotationId || !annotationId) {
+            return;
+        }
+
+        if (currentFileVersionId === fileVersionId) {
+            // If the pathname matched file version is the current version of the file, attempt to set
+            // the active annotation in box-annotations to the specified annotationId
+            emitAnnotatorActiveChangeEvent(annotationId);
```
jstoffan, line 223: Seems like we should emit this immediately rather than waiting on the feed items fetch to finish. We should know if the version is current or not much sooner. It doesn't really seem to fit here.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -199,6 +204,37 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+        } else {
+            // If the pathname matched file version is an older version of the file, call onVersionChange
+            // to signal that we need to change the version
+            const selectedVersion = feedItems.find(item => item.type === 'file_version' && item.id === fileVersionId);
+
+            if (selectedVersion) {
```
jstoffan, line 229: What's the negative case here? Can a version be permanently deleted? Do we want to swallow it or show the user an error or something?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -199,6 +204,37 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+                onVersionChange(selectedVersion, {
+                    currentVersionId: currentFileVersionId,
+                    activeAnnotationId: annotationId,
```
jstoffan, line 232: Alphabetize? Would also help to standardize on `active` versus `current` for both of these items.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -199,6 +204,37 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         this.fetchFeedItems();
     };
 
+    handleActiveAnnotation = (feedItems: FeedItems): void => {
+        const { emitAnnotatorActiveChangeEvent, file, match, onVersionChange } = this.props;
+        const { annotatorState: { activeAnnotationId } = {} } = this.state;
```
jstoffan, line 209: Should `annotatorState` default to an empty object to avoid needing this default parameter?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -199,6 +204,37 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+        const { annotationId } = match.params;
+        const { fileVersionId } = match.params;
+        const currentFileVersionId = file.file_version.id;
```
jstoffan, line 212: Rename-only reassignment.

src/elements/content-sidebar/Sidebar.js
jstoffan: Thoughts on this?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -28,15 +31,41 @@ const defaultState: AnnotatorState = {
...
 
-        state = defaultState;
+        constructor(props: ResultProps<P>) {
+            super(props);
+
+            // Determine by url if there is already a deeply linked annotation
+            const { history } = props;
+            const pathname = getProp(history, 'location.pathname');
+            const match = matchPath<MatchParams>(pathname, {
+                path: '/activity/annotations/:fileVersionId/:annotationId',
```
jstoffan, line 60: Could this be `path: /:sidebar/annotations/:fileVerisonId/:annotationId` to make it less activity sidebar-sspecific?

src/elements/content-sidebar/Sidebar.js
ConradJChan: I think we can leave it for now, because moving it up would probably require moving `handleVersionHistoryClick` up as well. We could address this in a separate PR

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -1,4 +1,7 @@
 import * as React from 'react';
+import getProp from 'lodash/get';
+import { matchPath } from 'react-router-dom';
+import { History } from 'history';
```
jstoffan, line 4: I don't think we currently import from `history` anywhere else, so this could complicate the parent app's dependencies. Can we import `RouterHistory` from `react-router-dom`, instead?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -28,15 +31,41 @@ const defaultState: AnnotatorState = {
     meta: null,
 };
 
+type WithHistoryProps = {
+    history?: History;
+};
+
+type MatchParams = {
+    fileVersionId?: string;
+    annotationId?: string;
```
jstoffan, line 40: Alphabetize?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -28,15 +31,41 @@ const defaultState: AnnotatorState = {
...
 
         annotator: Annotator | null = null;
 
-        state = defaultState;
+        constructor(props: ResultProps<P>) {
+            super(props);
+
+            // Determine by url if there is already a deeply linked annotation
+            const { history } = props;
+            const pathname = getProp(history, 'location.pathname');
```
jstoffan, line 58: Can these be undefined/null? Does `matchPath` handle those values adequately?

src/elements/common/annotator-context/withAnnotations.tsx
ConradJChan: It doesn't appear that `react-router-dom` has any exported History types -- seems like that worked for flow types but not typescript?

src/elements/common/annotator-context/withAnnotations.tsx
ConradJChan: yeah, if there's no `history` (i.e. outside of the webapp) this could be undefined but `matchPath` handles it gracefully and returns `null`

src/elements/common/annotator-context/withAnnotations.tsx
jstoffan: Gotcha. Can we bring in https://www.npmjs.com/package/@types/history as a dev dependency to avoid the explicit `history` dependency?

src/elements/content-preview/ContentPreview.js
```diff
@@ -742,6 +743,11 @@ class ContentPreview extends React.PureComponent<Props, State> {
             fileOpts[fileId].fileVersionId = selectedVersion.id;
         }
 
+        if (activeAnnotationId) {
+            fileOpts[fileId] = fileOpts[fileId] || {};
+            fileOpts[fileId].annotations = { activeId: activeAnnotationId };
```
jstoffan, line 748: What happens if this property is defined and passed in outside of the element? Should we try to merge into it or is overwriting what we want.

src/elements/content-preview/ContentPreview.js
```diff
@@ -742,6 +743,11 @@ class ContentPreview extends React.PureComponent<Props, State> {
             fileOpts[fileId].fileVersionId = selectedVersion.id;
         }
 
+        if (activeAnnotationId) {
```
jstoffan, line 746: Did we decide on if/how to standardize between active/current/selected?

src/elements/content-preview/__tests__/ContentPreview.test.js
```diff
@@ -296,6 +296,31 @@ describe('elements/content-preview/ContentPreview', () => {
...
+            const instance = wrapper.instance();
+            await instance.loadPreview();
+            expect(instance.preview.show).toHaveBeenCalledWith(
+                file.id,
+                expect.any(Function),
+                expect.objectContaining({
+                    container: expect.stringContaining('.bcpr-content'),
+                    header: 'none',
+                    showDownload: false,
+                    skipServerUpdate: true,
```
jstoffan, line 311: Probably omit properties that aren't under test.

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -28,15 +30,41 @@ const defaultState: AnnotatorState = {
...
+            const match = matchPath<MatchParams>(pathname, {
+                path: '/:sidebar/annotations/:fileVersionId/:annotationId',
+                exact: true,
+                strict: false,
```
jstoffan, line 61: I think this is the default value.

src/elements/content-preview/ContentPreview.js
ConradJChan: I guess I would lean toward overwriting it since it should be the more "correct" `activeId` since it's derived from the passed in `history` location pathname

src/elements/content-preview/ContentPreview.js
ConradJChan: No, I'll try to address it in a follow up PR

PR #2106, opened 2020-05-01
feat(draftjs): Extract functions into utils for other libs to reuse

mxiao6:
I think no new tests are needed cause the original tests have already covered all of the content in utils file.

### DIFF
--- a/src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelectorCore.js
+++ b/src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelectorCore.js
@@ -2,15 +2,17 @@
 import * as React from 'react';
 import { FormattedMessage } from 'react-intl';
 import classNames from 'classnames';
-import { EditorState, Modifier } from 'draft-js';
+import { EditorState } from 'draft-js';
 
 import DatalistItem from '../../datalist-item';
 import DraftJSEditor from '../../draft-js-editor';
 import SelectorDropdown from '../../selector-dropdown';
+import { getActiveMentionForEditorState, addMention } from './utils';
 
 import messages from './messages';
 
 import type { SelectorItems } from '../../../common/types/core';
+import type { Mention } from './utils';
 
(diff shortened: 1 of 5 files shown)


jstoffan:
Do any tests need to move over?

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelectorCore.js
```diff
@@ -67,19 +69,23 @@ class DraftJSMentionSelector extends React.Component<Props, State> {
         contacts: [],
         isDisabled: false,
         isRequired: false,
-        mentionTriggers: ['@', '＠', '﹫'],
```
jstoffan, line 70: If we do `mentionTriggers: defaultMentionTriggers`, does the logic for regex below need to change?

src/components/form-elements/draft-js-mention-selector/__tests__/utils.test.js
```diff
@@ -0,0 +1,113 @@
...
+        [
+            // empty input
+            {
+                editorState: EditorState.createEmpty(),
+            },
+            // input has ne mention
+            {
+                editorState: noMentionEditorState,
+            },
+        ].forEach(({ editorState }) => {
```
jstoffan, line 50: Should we convert these to `test.each`?

src/components/form-elements/draft-js-mention-selector/__tests__/utils.test.js
```diff
@@ -0,0 +1,113 @@
...
+    end: 18,
+};
+const twoMentionCursorInsideExpectedMention = {
+    mentionString: 'ba',
+    mentionTrigger: '@',
+    start: 14,
+    end: 17,
+};
+
+describe('components/form-elements/draft-js-mention-selector/DraftJSMentionSelector', () => {
```
jstoffan, line 38: Does this string need to be updated?

PR #2107, opened 2020-05-01
fix(annotations): fix styling for annotation cards

mickr:
* Fix annotation card vertical shifting on focus
* Make pending cards non-interactive

### DIFF
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
@@ -59,6 +59,14 @@
                 background-color: $bdl-box-blue-05;
                 border-bottom: none;
                 box-shadow: 0 $container-padding 0 -#{$container-padding - 1} $bdl-gray-10;
+            }
+        }
+
+        .bcs-activity-feed-task,
+        .bcs-activity-feed-task-new,
+        .bcs-activity-feed-comment,
+        .bcs-activity-feed-app-activity {
+            &.bcs-is-focused {
                 animation: fade-out 6s ease-out 2s 1 normal forwards;
 
                 @keyframes fade-out {
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
@@ -52,6 +52,7 @@
                 $container-padding: 25px;
 
                 margin-right: -$container-padding;
+                margin-bottom: 1px;
                 margin-left: -$container-padding;
                 padding-right: $container-padding;
                 padding-left: $container-padding;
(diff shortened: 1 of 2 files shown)


src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
```diff
@@ -73,8 +73,10 @@
         }
 
         .bcs-activity-feed-annotation-activity.bcs-is-focused {
+            margin-bottom: 1px;
```
jstoffan, line 76: What is this overriding? Can we fix it at the source?

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.scss
```diff
@@ -12,6 +12,7 @@
 
     &.bcs-is-pending {
         opacity: .4;
+        pointer-events: none;
```
jstoffan, line 15: Should we set `cursor: default` as well?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
```diff
@@ -73,8 +73,10 @@
...
             padding-left: 21px;
             border-left: 4px solid $bdl-yellorange;
+            animation: none;
```
jstoffan, line 79: Rather than overriding, can we avoid applying the animation in the first place?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
mickr: The bottom border is removed when focused which causes a very slight shift. I didn't want to cause any problems with existing cards, but I can move it up.

jstoffan: Rather than removing the border, can we make it transparent? That will resolve the issue for all cards, it sounds like.

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.scss
mickr: I don't think we need to, we should not have any interactivity while the pending card is showing.

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
mickr: Using a transparent bottom border will leave a slight angle on the new focused left border, which is why I went with the margin instead.

jstoffan: I see. Can we still solve the issue for all cards in another way, then? Should we put the margin in the shared `.bcs-is-focused` rule?

mickr: Yes, I will do that

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
mickr: I don't think we should remove the animation at this point. I could create a new rule for the annotation activity cards but that seems a bit overkill just for the animation. Maybe we should just keep the animation and update the animation behavior at a later time.

PR #2109, opened 2020-05-01
feat(annotations): Show versioned annotation activity cards

ConradJChan:
Propagate the current file_version down so the `AnnotationActivityItem` is aware of whether the annotation is on the current file version or not

TODO:
 - [x] unit tests


src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -79,6 +81,7 @@ const ActiveState = ({
             {items.map((item: FeedItem) => {
                 const isFocused = item === activeEntry;
                 const refValue = isFocused ? activeFeedItemRef : undefined;
+                const fileVersionId = getProp(item, 'file_version.id');
```
jstoffan, line 84: Maybe `itemFileVersionId` to be somewhat more clear that this is activity item-specific rather than file-specific?

src/elements/content-sidebar/activity-feed/activity-feed/__tests__/ActiveState.test.js
```diff
@@ -148,4 +153,28 @@ describe('elements/content-sidebar/ActiveState/activity-feed/ActiveState', () =>
...
     });
+
+    test.each`
+        currentFileVersionId | isCurrentVersion
+        ${'123'}             | ${true}
+        ${'456'}             | ${false}
+    `(
+        'should correctly reflect annotation activity isCurrentVersion as $isCurrentVersion based on file version id as $currentFileVersionId',
+        ({ currentFileVersionId, isCurrentVersion }) => {
+            const wrapper = shallow(
```
jstoffan, line 164: I know this is an existing test file, but would a `getWrapper` helper be worthwhile to avoid redeclaring all the base props for each test?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -144,12 +145,15 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         const { requestId } = meta || {};
         const { currentUser } = this.state;
         const isPending = action === 'create_start';
+        const currentFileVersionId = getProp(file, 'file_version.id');
+        // Patch the pending annotation with the current file version id
+        const patchedAnnotation = { ...annotation, file_version: { id: currentFileVersionId } };
```
jstoffan, line 150: Can users create annotations on older versions?

ConradJChan: Good callout, patching with the current version won't work. I'll make the change upstream in box-annotations

src/elements/content-sidebar/activity-feed/annotations/__tests__/AnnotationActivity.test.js
```diff
@@ -25,8 +27,13 @@ describe('elements/content-sidebar/ActivityFeed/annotations/AnnotationActivity',
         created_by: { name: 'Jane Doe', id: 10 },
         currentUser,
         description: { message: 'test' },
+        file_version: {
+            id: '456',
+            version_number: '2',
```
jstoffan, line 32: This field won't be present on a "new" annotation. Is that okay?

ConradJChan: I think that should be okay for now. It's a corner case that might come up with real-time updates but we're not there yet

PR #2111, opened 2020-05-04
fix(versions): show version uploaded by File Request

omgrief:


src/elements/common/selectors/version.js
```diff
@@ -32,8 +34,15 @@ const getVersionAction = ({ restored_at, trashed_at, version_promoted }: $Shape<
     return action;
 };
 
-const getVersionUser = ({ modified_by, restored_by, trashed_by }: $Shape<BoxItemVersion>): User => {
-    return restored_by || trashed_by || modified_by || PLACEHOLDER_USER;
+const getVersionUser = (
+    { modified_by, restored_by, trashed_by, uploader_display_name }: $Shape<BoxItemVersion>,
+    intl: IntlShape,
+): User => {
+    if (restored_by || trashed_by) return restored_by || trashed_by;
```
jstoffan, line 41: I find this to be very difficult to read. We usually avoid inline conditional returns.

src/elements/common/selectors/version.js
```diff
@@ -32,8 +34,15 @@ const getVersionAction = ({ restored_at, trashed_at, version_promoted }: $Shape<
...
+    if (modified_by && modified_by.id === PLACEHOLDER_USER.id && uploader_display_name) {
+        return { ...modified_by, name: internationalizeDisplayName(uploader_display_name, intl) };
```
jstoffan, line 43: Selectors should ideally not be aware of user-facing text or internationalization. Can we handle this in a presentational component elsewhere?

src/elements/common/selectors/version.js
```diff
@@ -32,8 +34,15 @@ const getVersionAction = ({ restored_at, trashed_at, version_promoted }: $Shape<
...
 };
 
-const getVersionUser = ({ modified_by, restored_by, trashed_by }: $Shape<BoxItemVersion>): User => {
-    return restored_by || trashed_by || modified_by || PLACEHOLDER_USER;
+const getVersionUser = (
+    { modified_by, restored_by, trashed_by, uploader_display_name }: $Shape<BoxItemVersion>,
+    intl: IntlShape,
+): User => {
+    if (restored_by || trashed_by) return restored_by || trashed_by;
+    if (modified_by && modified_by.id === PLACEHOLDER_USER.id && uploader_display_name) {
```
jstoffan, line 42: I think `modified_by === PLACEHOLDER_USER` would work as a simplified version of this condition.

src/elements/content-sidebar/activity-feed/version/Version.js
```diff
@@ -40,7 +40,7 @@ const Version = (props: Props): React.Node => {
     const action = selectors.getVersionAction(props);
     const { id, intl, onInfo, version_number, version_promoted } = props;
     // $FlowFixMe
-    const { name } = selectors.getVersionUser(props);
+    const { name } = selectors.getVersionUser(props, intl);
```
jstoffan, line 43: I believe `name` is empty for PLACEHOLDER_USER. Can we do something like the following?

```
const { name } = selectors.getVersionUser(props);
const uploaderName = uploader_display_name === FILE_REQUEST ? intl.formatMessage(messages.displayName) : uploader_display_name;
const displayName = name || uploaderName || '';
```

src/elements/content-sidebar/versions/__tests__/VersionsItem.test.js
```diff
@@ -3,7 +3,7 @@ import { FormattedMessage } from 'react-intl';
 import { shallow } from 'enzyme/build';
 import messages from '../messages';
 import selectors from '../../../common/selectors/version';
-import VersionsItem from '../VersionsItem';
+import { VersionsItem } from '../VersionsItem';
```
jstoffan, line 6: What is the reason for this change?

src/elements/content-sidebar/versions/VersionsItem.js
```diff
@@ -82,7 +84,7 @@ const VersionsItem = ({
     const versionInteger = versionNumber ? parseInt(versionNumber, 10) : 1;
     const versionTime = restoredAt || trashedAt || createdAt;
     const versionTimestamp = versionTime && new Date(versionTime).getTime();
-    const versionUserName = selectors.getVersionUser(version).name || (
+    const versionUserName = selectors.getVersionUser(version, intl).name || (
```
jstoffan, line 87: Can we use `FormattedMessage` to avoid injecting `intl`?

src/elements/common/selectors/__tests__/version.js
```diff
@@ -7,12 +7,19 @@ import {
     VERSION_UPLOAD_ACTION,
 } from '../../../../constants';
 
+import { FILE_REQUEST } from '../../../../features/file-request/constants';
+import messages from '../../../../features/file-request/messages';
```
jstoffan, line 11: We should avoid importing feature-specific modules into a common module, if possible.

src/elements/content-sidebar/versions/__tests__/VersionsItem.test.js
jstoffan: I saw after the fact that we've introduced a HOC to the component. Disregard.

src/utils/fields.js
```diff
@@ -134,6 +134,7 @@ const FILE_VERSION_FIELDS_TO_FETCH = [
     FIELD_RESTORED_FROM,
     FIELD_SIZE,
     FIELD_VERSION_NUMBER,
+    FIELD_UPLOADER_DISPLAY_NAME,
```
jstoffan, line 137: Please alphabetize all declarations, imports, object properties, etc. when possible.

src/elements/content-sidebar/versions/VersionsItem.js
```diff
@@ -82,9 +85,18 @@ const VersionsItem = ({
     const versionInteger = versionNumber ? parseInt(versionNumber, 10) : 1;
     const versionTime = restoredAt || trashedAt || createdAt;
     const versionTimestamp = versionTime && new Date(versionTime).getTime();
-    const versionUserName = selectors.getVersionUser(version).name || (
-        <FormattedMessage {...messages.versionUserUnknown} />
-    );
+    const versionUser = selectors.getVersionUser(version);
+    let versionUserName = versionUser.name || <FormattedMessage {...messages.versionUserUnknown} />;
+    if (uploader_display_name) {
+        if (versionUser.id === PLACEHOLDER_USER.id) {
```
jstoffan, line 91: Do these conditions need to be nested or can we make judicious use of `&&`?

src/elements/content-sidebar/versions/VersionsItem.js
```diff
@@ -72,6 +74,7 @@ const VersionsItem = ({
         trashed_at: trashedAt,
         version_number: versionNumber,
         version_promoted: versionPromoted,
+        uploader_display_name,
```
jstoffan, line 77: Let's rename to use camelCase to match the conventions of the file.

src/elements/content-sidebar/activity-feed/version/Version.js
```diff
@@ -38,17 +41,25 @@ const ACTION_MAP = {
 const Version = (props: Props): React.Node => {
     // $FlowFixMe
     const action = selectors.getVersionAction(props);
-    const { id, intl, onInfo, version_number, version_promoted } = props;
+    const { id, intl, onInfo, version_number, version_promoted, uploader_display_name } = props;
     // $FlowFixMe
-    const { name } = selectors.getVersionUser(props);
-
+    const user = selectors.getVersionUser(props);
+    let display_name = user.name;
```
jstoffan, line 47: We prefer to use destructuring to rename properties rather than rename-only reassignments. We should also use camelCase for all variable names.

src/constants.js
```diff
@@ -381,6 +381,7 @@ export const VERSION_RETENTION_INDEFINITE: 'indefinite' = 'indefinite';
 
 /* ------------------ Placeholder Feed Items ------------------------- */
 export const PLACEHOLDER_USER = { type: 'user', id: '2', name: '' };
+export const FILE_REQUEST = 'File Request';
```
jstoffan, line 384: Is this meant to be equivalent a user name? Can we call it `FILE_REQUEST_NAME` to indicate that?

src/elements/content-sidebar/versions/messages.js
```diff
@@ -3,6 +3,11 @@
 import { defineMessages } from 'react-intl';
 
 const messages = defineMessages({
+    fileRequestDisplayName: {
+        defaultMessage: 'File Request',
+        description: 'File Request feature display name',
```
jstoffan, line 8: We may want to provide a bit more context so the translator can provide a meaningful translation.

src/elements/content-sidebar/versions/VersionsItem.js
```diff
@@ -82,9 +85,18 @@ const VersionsItem = ({
...
     const versionTimestamp = versionTime && new Date(versionTime).getTime();
-    const versionUserName = selectors.getVersionUser(version).name || (
-        <FormattedMessage {...messages.versionUserUnknown} />
-    );
+    const versionUser = selectors.getVersionUser(version);
+    let versionUserName = versionUser.name || <FormattedMessage {...messages.versionUserUnknown} />;
+    if (uploader_display_name) {
+        if (versionUser.id === PLACEHOLDER_USER.id) {
+            versionUserName =
+                uploader_display_name === FILE_REQUEST ? (
```
jstoffan, line 93: I'm seeing some repetition between components. It seems like we're trying to do two things: 

1. Determine if the user is anonymous and, if so, substitute the blank name for `uploader_display_name` if available
2. Translate the `uploader_display_name` if it's a generic server-generated value (I assume)

It seems like the code for goal 1 should live in `selectors.getVersionUser` with a small helper method to extend/modify PLACEHOLDER_USER as needed.

It seems like the code for goal 2 should live in each component, since they each use different implementations to internationalize content (`injectIntl` vs `FormattedMessage`).

src/elements/content-sidebar/versions/VersionsItem.js
omgrief: ooh yes, that can probably happen

src/elements/content-sidebar/activity-feed/version/__tests__/Version.test.js
```diff
@@ -2,10 +2,11 @@ import * as React from 'react';
 import { shallow } from 'enzyme';
 import selectors from '../../../../common/selectors/version';
 import { VersionBase as Version } from '../Version';
-import { PLACEHOLDER_USER, VERSION_UPLOAD_ACTION } from '../../../../../constants';
+import { FILE_REQUEST_NAME, PLACEHOLDER_USER, VERSION_UPLOAD_ACTION } from '../../../../../constants';
+import messages from '../../../../common/messages';
 
 const translationProps = {
-    intl: { formatMessage: () => {} },
+    intl: { formatMessage: anyString => anyString },
```
sahiga, line 9: Where is `anyString` defined?

omgrief: it's just a function argument that makes formatMessage return whatever is passed in

PR #2112, opened 2020-05-04
feat(classification): New classification badge design

patrickpaul:
Moves Classification labels to the new design with a single color and larger border radius.

### DIFF
--- a/src/features/security/_mixins.scss
+++ b/src/features/security/_mixins.scss
@@ -4,13 +4,12 @@
     display: inline-flex;
     align-items: center;
     margin: 0;
-    padding: 0 5px;
-    border: 1px solid;
-    border-radius: $bdl-border-radius-size;
+    padding: 2px 8px;
+    border-radius: $bdl-border-radius-large-size;
 }
 
 @mixin bdl-SecurityBadge-name {
-    margin-left: 5px;
+    margin-left: $bdl-grid-unit;
     font-weight: bold;
     font-size: 10px !important; // To inhibit overriding
     line-height: 16px;
--- a/src/icons/general/IconSecurityClassification.tsx
+++ b/src/icons/general/IconSecurityClassification.tsx
@@ -5,31 +5,16 @@ import AccessibleSVG from '../accessible-svg';
 import { bdlGray } from '../../styles/variables';
 import { Icon } from '../iconTypes';
 
-const IconSecurityClassification = ({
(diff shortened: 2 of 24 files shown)


src/components/time/__tests__/__snapshots__/ReadableTime.test.js.snap
```diff
@@ -1,10 +1,10 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`components/time/ReadableTime should use default relative threshold if not provided 1`] = `"in 1,800,000 seconds"`;
```
booc0mtaco, line 3: this should probably not be changing. double check your node version; for buie it should use >=12.x 

cc @alexkrolick

src/features/security/_mixins.scss
```diff
@@ -4,13 +4,12 @@
...
-    padding: 0 5px;
-    border: 1px solid;
-    border-radius: $bdl-border-radius-size;
+    padding: 2px 8px;
+    border-radius: $bdl-border-radius-large-size;
 }
 
 @mixin bdl-SecurityBadge-name {
-    margin-left: 5px;
+    margin-left: 4px;
```
booc0mtaco, line 12: we now have `$bdl-grid-unit` (4px) for margins/paddings. let's use where possible

src/icons/general/IconSecurityClassification.tsx
```diff
@@ -5,31 +5,15 @@ import AccessibleSVG from '../accessible-svg';
...
 
-const IconSecurityClassification = ({
-    className = '',
-    height = 32,
-    color = bdlGray,
-    title,
-    strokeWidth = 2,
-    width = 32,
-}: Icon) => {
+const IconSecurityClassification = ({ className = '', height = 12, color = bdlGray, title, width = 12 }: Icon) => {
```
booc0mtaco, line 8: can we put a comment here to eventually move to the asset repo version of this? Its a small bit of refactor work so OK to come back to

is it this icon more or less? http://opensource.box.com/box-ui-elements/storybook/?path=/story/icon-line-shield16--shield-16

src/components/time/__tests__/__snapshots__/ReadableTime.test.js.snap
patrickpaul: Resolved.

PR #2113, opened 2020-05-05
fix(content-uploader): allow existing sub-folders to be uploaded

booc0mtaco:
only set error conditions when there are handled or unhandled errors. in the case of a folder that
exists the API returns an error, but we are OK with existing names


booc0mtaco:
RE #2092

src/api/uploads/FolderUploadNode.js
```diff
@@ -117,18 +117,20 @@ class FolderUploadNode {
...
+                errorEncountered = true;
+                errorCode = error.code;
                 errorCallback(error);
             } else {
                 // If this is a child folder of the folder being uploaded, this errorCallback will set
                 // an error message on the root folder being uploaded. Set a generic messages saying that a
                 // child has caused the error. The child folder will be tagged with the error message in
                 // the call to this.addFolderToUploadQueue below
+                errorEncountered = true;
+                errorCode = error.code;
```
jstoffan, line 133: This variable doesn't seem particularly useful. Can we use `error.code` directly on line 154?

src/api/uploads/FolderUploadNode.js
```diff
@@ -117,18 +117,20 @@ class FolderUploadNode {
             const data = await this.createFolder();
             this.folderId = data.id;
         } catch (error) {
-            errorEncountered = true;
-            errorCode = error.code;
             // @TODO: Handle 429
             if (error.code === ERROR_CODE_ITEM_NAME_IN_USE) {
                 this.folderId = error.context_info.conflicts[0].id;
             } else if (isRoot) {
+                errorEncountered = true;
```
jstoffan, line 124: It looks like this variable wouldn't be used if `isRoot` is true due to line 139. Is this needed? Should we test this case, if so?

src/api/uploads/FolderUploadNode.js
booc0mtaco: probably so. i was simply moving the error var.s to each block and removing the one that should not be set, so I left as much of the logic as it was before

src/api/uploads/FolderUploadNode.js
booc0mtaco: i can remove the duplicates. by removing 120-121, this reveals the two variables were being set in cases where they weren't being used

src/api/uploads/FolderUploadNode.js
booc0mtaco: oh actually, we still need `errorCode` as `error` only exists in the catch block

PR #2117, opened 2020-05-06
feat(annotation): View annotations on previous file version

ConradJChan:
TODO:
 - [x] unit tests


src/elements/common/flowTypes.js
```diff
@@ -25,8 +25,14 @@ type GetAvatarUrlCallback = string => Promise<?string>;
...
 type GetProfileUrlCallback = string => Promise<string>;
 
+type Page = {
+    type: 'page',
+    value: number,
+};
+
 type AdditionalVersionInfo = {
     currentVersionId?: ?string,
+    location?: Page,
```
jstoffan, line 35: Hm, this doesn't really seem to fit on the version object, since it's specific to an annotation. Is this something we can derive based on the active annotation id?

src/elements/content-preview/ContentPreview.js
```diff
@@ -1113,11 +1123,18 @@ class ContentPreview extends React.PureComponent<Props, State> {
...
+        const { location } = additionalVersionInfo;
+
         this.updateVersionToCurrent = additionalVersionInfo.updateVersionToCurrent;
 
         onVersionChange(version, additionalVersionInfo);
+
         this.setState({
             selectedVersion: version,
+            startAt: {
+                unit: location.type === 'page' ? 'pages' : undefined,
```
jstoffan, line 1135: Will this easily extend to support other file/location types in the future?

src/elements/content-preview/ContentPreview.js
```diff
@@ -1113,11 +1123,18 @@ class ContentPreview extends React.PureComponent<Props, State> {
...
+                value: location.value,
```
jstoffan, line 1136: Does the Preview SDK handle these properties being undefined if the `startAt` object is defined? Maybe we should make the whole `startAt` object undefined or null?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -591,6 +600,31 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         return api.getUsersAPI(false).getAvatarUrlWithAccessToken(userId, file.id);
     };
 
+    handleAnnotationSelect = (annotationId: string, { fileVersionId, location }: AnnotationSelectMeta): void => {
```
jstoffan, line 603: Do we already have something like this elsewhere? I see a simliar handler in one of the other Sidebar components. Can/should we combine them to avoid sprawl for this code path?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -591,6 +600,31 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+        const { emitAnnotatorActiveChangeEvent, file, getAnnotationsMatchPath, history, onVersionChange } = this.props;
+        const { feedItems } = this.state;
+        const currentFileVersionId = getProp(file, 'file_version.id');
+
+        const match = getAnnotationsMatchPath(history);
+        const { params: { fileVersionId: currentPreviewFileVersionId = currentFileVersionId } = {} } = match || {};
```
jstoffan, line 609: Seems like `lodash/get` might help here.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -591,6 +600,31 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+
+        emitAnnotatorActiveChangeEvent(annotationId);
+
+        if (fileVersionId !== currentPreviewFileVersionId) {
```
jstoffan, line 613: What happens if we go from oldVersion -> oldVersion instead of currentVersion -> oldVersion?

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
```diff
@@ -54,7 +55,9 @@ const AnnotationActivity = (props: Props) => {
     };
 
     const handleOnSelect = () => {
-        onSelect(id);
+        const { location } = target;
+        const { id: fileVersionId } = file_version;
+        onSelect(id, { fileVersionId, location });
```
jstoffan, line 60: Should we just pass up the whole annotation?

src/elements/common/flowTypes.js
ConradJChan: That's a good idea, let me give that a shot

src/elements/content-preview/ContentPreview.js
ConradJChan: Not really, preview sdk only supports `pages` and `seconds` so I figured when we enhance this in the future we could address it then but I'll add a helper method to do the mapping to make it more extensible

src/elements/content-preview/ContentPreview.js
ConradJChan: It looks like it does, but I think you're right, it's safer to make the whole object undefined

src/elements/content-sidebar/ActivitySidebar.js
ConradJChan: In this case, we rely on `emitAnnotatorActiveChangeEvent` to do it's job and update the AnnotatorContext with the new `activeAnnotationId` and we bypass the `onVersionChange` logic

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
ConradJChan: 👍

src/elements/content-sidebar/ActivitySidebar.js
ConradJChan: Sure let me see how we can consolidate this

src/elements/common/flowTypes.js
ConradJChan: Actually that might not work because `ContentPreview` has no access to neither `feedItems` nor a list of annotations. Maybe `AdditionalVersionInfo` can contain the whole annotation object?

jstoffan: The annotation isn't directly related to the selected version, though. It seems like there should be two separate callbacks, one for version change and one for annotation change.

src/elements/content-sidebar/ActivitySidebar.js
jstoffan: That seems to be referring to annotations within the same version. The scenario I'm proposing is as follows: The user is viewing version 2 of 3, then selects an annotation on version 1 (the oldest version).

src/elements/content-preview/ContentPreview.js
```diff
@@ -354,6 +372,24 @@ class ContentPreview extends React.PureComponent<Props, State> {
...
+        } else if (prevSelectedVersion !== selectedVersion && activeAnnotationId) {
+            // If the selected version has changed, update the deep linked annotation's fileVersionId
+            history.replace(`/activity/annotations/${previewedFileVersionId}/${activeAnnotationId}`);
+        } else if (activeAnnotationId !== prevActiveAnnotationId) {
+            // If the active annotation has changed, update the deep linked annotationId
+            const path = activeAnnotationId
+                ? `/activity/annotations/${previewedFileVersionId}/${activeAnnotationId}`
+                : '/activity/';
+            history.replace(path);
+        }
```
mickr: Should this logic reside in a function and then called from `componentDidUpdate`

src/elements/content-preview/ContentPreview.js
```diff
@@ -354,6 +372,31 @@ class ContentPreview extends React.PureComponent<Props, State> {
...
             this.updatePreviewToken();
         }
+
+        if (!prevFile && !!file && !selectedVersion && activeAnnotationId) {
+            // Replace any deep linked annotation fileVersionId with the file's current version because the
+            // current functionality is to load the current version of a deep linked annotation on a previous version
+            // on a hard load.
+            // NOTE: When ContentPreview can support mounting with an older file version this behavior can
+            // likely go away
+            history.replace(`/activity/annotations/${currentFileVersionId}/${activeAnnotationId}`);
```
jstoffan, line 382: It feels like this logic should live inside a Route/Redirect component, similar to the one at the bottom of SidebarPanels. Maybe it could live in the ActivitySidebar, since that's what it's affecting?

src/elements/content-preview/ContentPreview.js
```diff
@@ -1123,6 +1170,54 @@ class ContentPreview extends React.PureComponent<Props, State> {
         });
     };
 
+    /**
+     * Gets the Preview SDK startAt unit based on the Annotation target.location.type property.
+     * Returns undefined if type is not recognized.
+     * @param {string} type - The type returned from an Annotation target.location property
+     * @return {string | undefined}
+     */
+    getStartAtUnitFromLocationType(type: string): string | undefined {
```
jstoffan, line 1179: Can we use a map or object literal as a dictionary?

```
const startAtTypes = {
    'page': 'pages',
}

return startAtTypes[type];
```

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -591,6 +597,17 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         return api.getUsersAPI(false).getAvatarUrlWithAccessToken(userId, file.id);
     };
 
+    handleAnnotationSelect = (annotation: Annotation): void => {
+        const {
+            file_version: { id: fileVersionId },
+        } = annotation;
+        const { onAnnotationSelect } = this.props;
+        const { feedItems } = this.state;
+        const version = feedItems.filter(item => item.type === 'file_version').find(item => item.id === fileVersionId);
```
jstoffan, line 606: Is it possible for the version to not yet be loaded? If we introduce pagination to the feed calls, will this continue to work? What happens if the version isn't present?

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
```diff
@@ -54,7 +55,20 @@ const AnnotationActivity = (props: Props) => {
     };
 
     const handleOnSelect = () => {
-        onSelect(id);
+        // Return an object of properties matching only the Annotation type
+        onSelect(
```
jstoffan, line 59: If this is needed, can we use whitelist instead of a blacklist? The annotation shape is probably less likely to change than the component props. Or should we just pass in `item` rather than spreading it?

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityLink.tsx
```diff
@@ -11,7 +11,11 @@ export interface AnnotationActivityLinkProps {
 }
 
 const AnnotationActivityLink = ({ id, message, onClick = noop }: AnnotationActivityLinkProps): JSX.Element => {
-    const handleClick = () => {
+    const handleClick = (event: React.SyntheticEvent) => {
+        event.preventDefault();
+        event.stopPropagation();
+        event.nativeEvent.stopImmediatePropagation(); // Prevents document event handlers from executing
```
jstoffan, line 17: Which document listeners are we preventing and why?

src/elements/content-sidebar/ActivitySidebar.js
ConradJChan: Currently it's not possible for the version to not yet be loaded -- except for the case where there is only 1 version of the file. Once we introduce pagination, we may need to dynamically fetch the missing version. That or if we are able to refactor the way ContentPreview loads previous versions, we may be able to get away with just sending the version id instead of the whole object. If the version isn't present, the logic upstream just treats it as a change in selected annotation id on the current version

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityLink.tsx
ConradJChan: We're preventing the native click event from propagating because that would trigger the `useEventOutside` hook logic in the box-annotations `RegionList` and deselect the annotation

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -7,8 +7,11 @@
 import * as React from 'react';
 import debounce from 'lodash/debounce';
 import flow from 'lodash/flow';
+import getProp from 'lodash/get';
 import noop from 'lodash/noop';
 import { FormattedMessage } from 'react-intl';
+import { withRouter } from 'react-router-dom';
```
jstoffan, line 13: I think these lines can be simplified to `import { withRouter, type ContextRouter } from 'react-router-dom'`;

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityLink.tsx
jstoffan: Can we add a comment with that information?

Issue #2119, opened 2020-05-07
openUrlInsideIframe fails once all subsequent calls will also fail


BLamy:
**Steps to reproduce the problem:**
Use the `openUrlInsideIframe` function but have the first request fail. The second time you call `openUrlInsideIframe` the function will crash on the following line because `iframe.contentDocument` is null.
https://github.com/box/box-ui-elements/blob/e282d5f53e3d0e3c15a728e9b113888269acfea6/src/utils/iframe.js#L27

**What is the expected behavior? (Screenshots can be helpful here)**
Since the previous iframe failed to load there is nothing to cleanup in the content document

**What went wrong? (Screenshots, console logs, or HAR files can be helpful here)**
Can never download again until refresh

PR #2121, opened 2020-05-07
fix(utils): if openUrlInsideIframe fails it will fail everytime after

BLamy:


booc0mtaco:
addresses #2119

booc0mtaco:
@BLamy to help me understand, would this happen in a case where the iframe somehow went to some different origin, ie, if a session timed out, or something similar?

https://developer.mozilla.org/en-US/docs/Web/API/HTMLIFrameElement/contentDocument

BLamy:
In my experience, it's been that my download_URL threw a 500 error when I attempted to visit it. If the user then attempts to just re-download the file because they couldn't get it every subsequent attempt will fail.

PR #2123, opened 2020-05-07
feat(select-field): add search input to BaseSelectField

ChenCodes:
Changes in this PR:
- add SearchForm as part of the  BaseSelectField component

ezgif com-video-to-gif (1)


src/components/select-field/BaseSelectField.js
```diff
@@ -188,16 +217,16 @@ class BaseSelectField extends React.Component<Props, State> {
         }
     };
 
-    handleBlur = () => {
+    handleBlur = (event: FocusEvent) => {
         const { isOpen } = this.state;
-        if (isOpen) {
+        if (isOpen && event.relatedTarget && event.relatedTarget.className !== 'search-input') {
```
ChenCodes, line 222: There is a call later on where this.searchInputRef.focus() is being called which means that handleBlur will be called here. We don't actually want to close the dropdown though because the user hasn't clicked outside of modal.

src/components/select-field/BaseSelectField.js
```diff
@@ -221,6 +250,11 @@ class BaseSelectField extends React.Component<Props, State> {
                 break;
             case 'Enter':
             case ' ':
+                // When the search input is active, we want to allow space key presses in the search string
+                if (key === ' ' && shouldShowSearchInput) {
```
ChenCodes, line 254: When the user wants to use the SearchForm, they will likely enter in a space keypress. Thus, we should allow space keypresses to go through rather than either "selecting an option" or "closing the dropdown". This is why we now have an early `break` statement.

src/components/select-field/BaseSelectField.js
```diff
@@ -247,22 +281,25 @@ class BaseSelectField extends React.Component<Props, State> {
                 }
                 break;
             default: {
-                stopDefaultEvent(event);
-                const lowerCaseKey = key.toLowerCase();
-                const optionIndex = options.findIndex(
-                    option => option.displayText.toLowerCase().indexOf(lowerCaseKey) === 0,
-                );
+                if (!shouldShowSearchInput) {
```
ChenCodes, line 284: If the user has toggled shouldShowSearchInput to be on, this means that we want to allow character keypresses to go through without performing the default setActiveItem behavior. This is because we want the search input field to be able to keep track of what the current search string is.

If the toggle is off, then we still proceed how we did before.

src/components/select-field/BaseSelectField.js
```diff
@@ -247,22 +281,25 @@ class BaseSelectField extends React.Component<Props, State> {
...
                 }
             }
         }
     };
 
     openDropdown = () => {
+        const { shouldShowSearchInput } = this.props;
         if (!this.state.isOpen) {
-            this.setState({ isOpen: true });
+            this.setState({ isOpen: true }, () => shouldShowSearchInput && this.searchInputRef.focus());
```
ChenCodes, line 302: We should be focusing the search input by default when the dropdown is opened.

src/components/select-field/__tests__/__snapshots__/BaseSelectField.test.js.snap
```diff
@@ -5,7 +5,7 @@ exports[`components/select-field/BaseSelectField renderSelectButton() should sen
   aria-activedescendant={null}
   aria-autocomplete="list"
   aria-expanded={false}
-  aria-owns="selectfield11"
+  aria-owns="selectfield12"
```
ChenCodes, line 8: Not sure if this change is needed. This happened after I used the injectIntl HOC on BaseSelectField component. Would like feedback on whether or not this is needed.

src/components/select-field/BaseSelectField.js
```diff
@@ -247,22 +281,25 @@ class BaseSelectField extends React.Component<Props, State> {
...
             default: {
-                stopDefaultEvent(event);
-                const lowerCaseKey = key.toLowerCase();
-                const optionIndex = options.findIndex(
-                    option => option.displayText.toLowerCase().indexOf(lowerCaseKey) === 0,
-                );
+                if (!shouldShowSearchInput) {
+                    stopDefaultEvent(event);
+                    const lowerCaseKey = key.toLowerCase();
+                    const optionIndex = options.findIndex(
```
booc0mtaco, line 287: this is not supported in IE11. is there a polyfill? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex

src/components/select-field/messages.js
```diff
@@ -6,6 +6,11 @@ const messages = defineMessages({
         description: 'text shown on the Clear All option in the options list',
         id: 'boxui.selectField.clearAll',
     },
+    searchPlaceholder: {
+        defaultMessage: 'Search',
```
booc0mtaco, line 10: I feel like we must have this as a common string somewhere already...

ChenCodes: export const VIEW_SEARCH: 'search' = 'search';

We do have this but it doesn't look like it's the right search constant that we should be using.

booc0mtaco: No I mean something like `boxui.contentExplorer.searchPlaceholder`. Unfortunately its not in a common file so it can't be used as-is.

src/components/select-field/__tests__/__snapshots__/BaseSelectField.test.js.snap
alexkrolick: This is an autoincremented counter for the generated IDs used to associate aria labels with their elements. It went up because a new test was added, so the uid generator was called again

src/components/select-field/messages.js
ChenCodes: Ah got it.

src/components/select-field/BaseSelectField.js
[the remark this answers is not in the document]
ChenCodes: Yeah I can use classList instead as it is safer.

ConradJChan:
Should this be a separate component build on top of `BaseSelectField`? Like a  `SearchableSelectField`?

src/components/select-field/SelectField.scss
```diff
@@ -12,6 +12,16 @@
     }
 }
 
+.search-input-container {
+    width: 100%;
+    padding-right: 5px;
```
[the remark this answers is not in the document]
ChenCodes: I can reference that

booc0mtaco:
re: our chat earlier, we probably want balanced padding around the search field, so check with design on the bottom padding for that

src/components/select-field/BaseSelectField.js
ChenCodes: I can use findIndex from Lodash, it looks like it supports IE11.

ChenCodes:
> Should this be a separate component built on top of `BaseSelectField`? Like a `SearchableSelectField`?

I mentioned offline that I think we do not need a separate component for this because this functionality that I am introducing is a core functionality of the dropdown.

src/components/select-field/BaseSelectField.js
```diff
@@ -112,8 +114,6 @@ class BaseSelectField extends React.Component<Props, State> {
 
         this.selectFieldContainerRef = React.createRef();
 
-        this.searchInputRef = React.createRef();
```
ChenCodes: Removed because we already have this.searchInputRef being set later in the code.

src/components/select-field/SelectField.scss
```diff
@@ -4,7 +4,6 @@
 .select-field {
     .overlay {
         min-width: 100%;
-        max-height: 300px;
```
ChenCodes: Originally added this for testing purposes.

src/components/select-field/BaseSelectField.js
```diff
@@ -386,9 +452,18 @@ class BaseSelectField extends React.Component<Props, State> {
 
     renderSelectOptions = () => {
         const { optionRenderer, options, selectedValues, separatorIndices, shouldShowClearOption } = this.props;
-        const { activeItemIndex } = this.state;
+        const { activeItemIndex, searchText } = this.state;
 
-        const selectOptions = options.map<React.Element<typeof DatalistItem | 'li'>>((item, index) => {
+        let filteredOptions;
+        if (searchText === '') {
```
booc0mtaco, line 458: you can somewhat simplify this if you check for `''` in `.filter`. 

`option => !!searchText ? option.displayText.toLowerCase().includes(searchText.toLowerCase()) : true`

src/components/select-field/BaseSelectField.js
```diff
@@ -434,6 +509,14 @@ class BaseSelectField extends React.Component<Props, State> {
             selectOptions.splice(separatorIndex + index, 0, <li key={`separator${separatorIndex}`} role="separator" />);
         });
 
+        if (selectOptions.length === 0) {
```
booc0mtaco, line 512: we know the list is empty before we run the `.map` call, so we can exit early in that case. You can use `filteredOptions`

also, this will appear in the same slot as a normal dropdown item, correct? are you producing a dom tree here that differs? this is just some DIV with a non-suit, non-component-specific class name. This should look like a disabled data item using the regular containers

src/components/select-field/SelectField.scss
```diff
@@ -12,6 +12,23 @@
     }
 }
 
+.search-input-container {
+    width: 100%;
+    padding-right: $bdl-grid-unit;
```
booc0mtaco, line 17: Think about why I asked for using the grid. We can use SASS to get multiples of the grid unit so all measures are relative to the base unit. `padding-bottom: $bdl-grid-unit * 2`

src/components/select-field/SelectField.scss
```diff
@@ -12,6 +12,23 @@
...
+    padding-bottom: 8px;
+    padding-left: $bdl-grid-unit;
+
+    input {
+        width: 100%;
+    }
+}
+
+.no-results {
+    padding-top: 5px;
```
booc0mtaco, line 27: these are not using the grid either. can you use our variable for padding/margin values? Also consider a more specific class name (or as suggested above, use/create a disabled datalist item so you get styling for free

src/components/select-field/BaseSelectField.js
ChenCodes: Good call, I can do that instead.

src/components/select-field/SelectField.scss
ChenCodes: Yep I can add the following:
```
.select-option.is-disabled {
    pointer-events:none;
}
```

src/components/select-field/SelectField.scss
ChenCodes: Thanks for clarifying!

src/components/select-field/BaseSelectField.js
```diff
@@ -151,6 +178,10 @@ class BaseSelectField extends React.Component<Props, State> {
 
     selectFieldContainerRef: { current: null | HTMLDivElement };
 
+    searchInputRef: React.Element<any>;
```
booc0mtaco, line 181: *two* `searchInputRet`s???

ChenCodes: Addressed offline

ChenCodes: The second line is needed due to a flow error

ChenCodes: There should only be one, the second is not needed since HTMLInputElement has a focus attribute.

src/components/select-field/BaseSelectField.js
```diff
@@ -469,18 +546,22 @@ class BaseSelectField extends React.Component<Props, State> {
...
                     {this.renderSelectButton()}
-                    <SelectFieldDropdown
-                        isScrollable={isScrollable}
-                        multiple={multiple}
-                        selectedValues={selectedValues}
-                        selectFieldID={this.selectFieldID}
-                    >
-                        {this.renderSelectOptions()}
-                    </SelectFieldDropdown>
+                    <>
```
booc0mtaco, line 549: i dont think you need this fragment

ChenCodes: Yeah this should be removed. It was originally added since I had the search field as a sibling to SelectorDropdown

src/components/select-field/__tests__/BaseSelectField.test.js
```diff
@@ -127,7 +135,47 @@ describe('components/select-field/BaseSelectField', () => {
...
     describe('renderSelectOptions()', () => {
+        test('should render FormattedMessage if searchText is not a substring of any of the options', () => {
+            const searchText = 'abc';
+            const wrapper = shallowRenderSelectField();
+            const instance = wrapper.instance();
+            instance.setState({
+                searchText,
+            });
+
+            const message = wrapper.find(FormattedMessage);
```
booc0mtaco, line 159: Normally I would expect to see this as a string, but I am guessing using the instance works?

ChenCodes: https://enzymejs.github.io/enzyme/docs/api/ShallowWrapper/find.html

In the docs it looks like we are able to find the component by passing in a component constructor.

PR #2124, opened 2020-05-08
feat(contentpreview): Make ContentPreview location aware

ConradJChan:


src/elements/common/nav-router/NavRouter.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react';
+import { MemoryRouter, MemoryRouterProps, Router } from 'react-router';
+import { History } from 'history';
+
+type Props = {
+    children: React.ReactNode;
+    history?: History;
+} & MemoryRouterProps;
```
jstoffan, line 8: This isn't present on the current component. Will it require developers to pass in a router and/or its props? It's meant to be an optional feature. Will it work with non-memory routers?

src/elements/common/nav-router/index.js
```diff
@@ -1 +1,2 @@
 export { default } from './NavRouter';
```
jstoffan, line 1: Should we remove NavRouter.js or convert it to a `.js.flow` file?

src/elements/content-preview/__tests__/PreviewNavigation.test.js
```diff
@@ -48,16 +51,18 @@ describe('elements/content-preview/PreviewNavigation', () => {
...
-                    intl={{
-                        formatMessage: jest.fn(),
-                    }}
-                    collection={['a', 'b', 'c']}
-                    currentIndex={2}
-                    onNavigateLeft={onNavigateLeftMock}
-                    onNavigateRight={jest.fn()}
-                    history={historyMock}
-                />,
+                <Router history={historyMock}>
```
jstoffan, line 54: Should this use `getWrapper`?

src/elements/common/nav-router/NavRouter.tsx
ConradJChan: Yeah I think due to flow, it’s not restricted but ContentSidebar makes use of the “hidden” props for the MemoryRouter. They’re optional anyway so users don’t have to pass anything in. I think this is just a more accurate typing

src/elements/content-preview/__tests__/PreviewNavigation.test.js
ConradJChan: Yeah I’ll try to clean this up

src/elements/common/nav-router/NavRouter.tsx
jstoffan: I see. Does this work with non-memory routers? If I pass in a BrowserRouter, will Typescript complain?

PR #2125, opened 2020-05-08
fix(classifications): Classification tinycons have their own color map

patrickpaul:
Classification's showing up in tinycons in the file list have a unique color compared to labels shown across web and mobile. Add those colors in and make them available.


src/features/classification/classificationColorsMap.js
```diff
@@ -16,34 +16,42 @@ const classificationColorsMap = {
...
     },
     [CLASSIFICATION_COLOR_ID_6]: {
         name: messages.classificationGreen,
         color: vars.bdlGreenLight50,
+        tinycon: vars.bdlGreenLight,
     },
     [CLASSIFICATION_COLOR_ID_7]: {
         name: messages.classificationGrey,
         color: vars.bdlGray20,
+        tinycon: vars.bdlGray62,
```
jpfranco, line 54: Was it deliberate to use `bdlGray62` instead of `bdlGray` for this one?

patrickpaul: Yep!

PR #2126, opened 2020-05-08
fix(colors): correct orange base hex value

alexkrolick:


alexkrolick:
@booc0mtaco design told me the current value was a typo and the other hues in the spectrum were based on this value

PR #2127, opened 2020-05-11
fix(tasks): remove assign tasks to groups FF

il10:


src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -508,7 +508,7 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
     getApproverWithQuery = debounce(
         (searchStr: string) =>
             this.getCollaborators(this.getApproverContactsSuccessCallback, this.errorCallback, searchStr, {
-                includeGroups: isFeatureEnabled(this.props.features, 'activityFeed.tasks.assignToGroup'),
+                includeGroups: true,
```
booc0mtaco, line 511: does it make sense to preserve this as an exposed option, i.e., will there be cases where we want to get collaborators without groups?

il10: Hm from what I know I don't believe there will be a case where we want to dismiss groups functionality and only get collaborators

PR #2129, opened 2020-05-12
feat(usm): Separate USM into two components

sahiga:
**Ready for review**

This PR moves the body of the Unified Share Modal into a new component (the Unified Share Form), which allows applications to use the form without the modal wrapper.

The USM is responsible for data retrieval, toggling the link removal confirmation modal, and instantiating the USF with or without the modal wrapper. The USF is responsible for all state management that the modal wrapper does not need to know about.


src/features/unified-share-modal/flowTypes.js
```diff
@@ -171,3 +172,128 @@ export type tooltipComponentIdentifierType =
...
+    /** ID for the item being shared - for use with V2 APIs only */
+    itemID?: string,
+    /** Type of the item being shared - for use with V2 APIs only */
+    itemType?: string,
+    /** Language for the USM/USF */
+    language?: string,
+    /** Handler function for when the USM is closed */
+    onRequestClose?: Function,
+    /** Access token - for use with V2 APIs only */
+    token?: string,
```
j3tan, line 284: would this need to be passed in? I would assume the wrapper component would handle all of the logic around APIs and tokens

sahiga: Yes, good catch. I removed the API-specific props. Can you review this PR again?

src/features/unified-share-modal/flowTypes.js
```diff
@@ -171,3 +172,123 @@ export type tooltipComponentIdentifierType =
     | 'shared-link-copy-button'
     | 'shared-link-settings'
     | 'shared-link-toggle';
+
+type BaseUSMProps = {
```
booc0mtaco, line 176: This may be an opportunity to add additional sanity here. We could specify base types based on the USM feature they are for? some are for the invite collab section, while others are for shared link section, etc.

- functionality types would aid in future enhancements and refactoring/componentization 

Mostly an idea to help organize future use cases (assuming we need to introduce a way to disable some functionality in certain UIs)

src/features/unified-share-modal/UnifiedShareForm.js
```diff
@@ -0,0 +1,577 @@
...
+import InviteePermissionsMenu from './InviteePermissionsMenu';
+import messages from './messages';
+import SharedLinkSection from './SharedLinkSection';
+import EmailForm from './EmailForm';
+import getDefaultPermissionLevel from './utils/defaultPermissionLevel';
+import mergeContacts from './utils/mergeContacts';
+import type { contactType as Contact, USFProps } from './flowTypes';
+import type { SelectOptionProp } from '../../components/select-field/props';
+
+import './UnifiedShareModal.scss';
```
j3tan, line 23: nit: should this be UnifiedShareForm.scss now?

src/features/unified-share-modal/__tests__/UnifiedShareModal.stories.test.js
```diff
@@ -1,17 +1,25 @@
...
-    test.each(USM_STORIES)('looks visually correct when using story %s', async id => {
+    const MODAL_STORIES = [STORIES[0], STORIES[1], STORIES[2]];
+
+    test.each(STORIES)('looks visually correct when using story %s', async id => {
         const image = await BoxVisualTestUtils.takeScreenshot(id);
         return expect(image).toMatchImageSnapshot();
     });
 
-    test.each(USM_STORIES)('looks visually correct when button is clicked in story %s', async id => {
+    test.each(MODAL_STORIES)('looks visually correct when button is clicked in story %s', async id => {
```
booc0mtaco, line 16: as a coding tip, `STORIES.slice(0,-1)` would do the same, and avoid keeping the arrays in sync. 

Here, tho, might actually be better to define each test with its own list of keys. later changes to the order of `STORIES` could lead to failing tests and confusion.

src/features/unified-share-modal/UnifiedShareForm.js
sahiga: I'll remove this import, since `UnifiedShareForm` inherits the import from `UnifiedShareModal`

PR #2131, opened 2020-05-12
fix(select-field): no longer selecting the wrong item

ChenCodes:
Changes in this PR:
- Bugfix for BaseSelectField: when a search filter is applied and the user tries to select an item, we should be updating the correct item in the original options array

ezgif com-video-to-gif (2)

### DIFF
--- a/src/components/select-field/BaseSelectField.js
+++ b/src/components/select-field/BaseSelectField.js
@@ -14,6 +14,7 @@ import SelectFieldDropdown from './SelectFieldDropdown';
 import type { SelectOptionValueProp, SelectOptionProp } from './props';
 import { PLACEMENT_BOTTOM_END, PLACEMENT_BOTTOM_START } from '../popper/constants';
 import SearchForm from '../search-form/SearchForm';
+import CLEAR from './constants';
 
 import messages from './messages';
 
@@ -336,20 +337,21 @@ class BaseSelectField extends React.Component<Props, State> {
         }
     };
 
-    getItemFromFilteredOptions = (index: number) => {
+    getFilteredOptions = () => {
         const { options } = this.props;
         const { searchText } = this.state;
 
-        const filteredOptions = options.filter(option =>
-            searchText ? option.displayText.toLowerCase().includes(searchText.toLowerCase()) : true,
-        );
+        return options.filter(option => {
+            const isSubstring = option.displayText.toLowerCase().includes(searchText.toLowerCase());
+            const isClearOption = option.value === CLEAR;
 
-        return filteredOptions[index];
+            return searchText ? isSubstring && !isClearOption : true;
+        });
     };
 
     selectSingleOption(index: number) {
         const { selectedValues } = this.props;
-        const item = this.getItemFromFilteredOptions(index);
+        const item = this.getFilteredOptions()[index];
         // If item not previously selected, fire change handler
         if (!selectedValues.includes(item.value)) {
             this.handleChange([item]);
@@ -360,7 +362,7 @@ class BaseSelectField extends React.Component<Props, State> {
     selectMultiOption = (index: number) => {
         const { defaultValue, options, selectedValues } = this.props;
         const hasDefaultValue = defaultValue != null; // Checks if not undefined or null
-        const item = this.getItemFromFilteredOptions(index);
+        const item = this.getFilteredOptions()[index];
 
         // If we are already using the default option, just return without firing onChange
         if (hasDefaultValue && defaultValue === item.value) {
@@ -460,12 +462,10 @@ class BaseSelectField extends React.Component<Props, State> {
     };
 
     renderSelectOptions = () => {
-        const { optionRenderer, options, selectedValues, separatorIndices, shouldShowClearOption } = this.props;
-        const { activeItemIndex, searchText } = this.state;
+        const { optionRenderer, selectedValues, separatorIndices, shouldShowClearOption } = this.props;
+        const { activeItemIndex } = this.state;
 
-        const filteredOptions = options.filter(option =>
-            searchText ? option.displayText.toLowerCase().includes(searchText.toLowerCase()) : true,
-        );
+        const filteredOptions = this.getFilteredOptions();
 
         if (filteredOptions.length === 0) {
             return (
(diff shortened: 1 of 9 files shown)


src/components/select-field/BaseSelectField.js
```diff
@@ -336,9 +336,20 @@ class BaseSelectField extends React.Component<Props, State> {
...
+        );
+
+        return filteredOptions[index];
+    };
+
     selectSingleOption(index: number) {
-        const { options, selectedValues } = this.props;
-        const item = options[index];
+        const { selectedValues } = this.props;
+        const item = this.getItemFromFilteredOptions(index);
```
ChenCodes, line 352: This is the bulk of the fix. Before, we were setting item = options[index] but this logic will not work because index no longer refers to the correct index of the item that we just selected. Index now refers to the filtered index.

So for example, if we have options A, B, and C and then we perform a filter with the search string of 'C', when we click on C, the index that is being called in selectSingleOption is 0 because that is the only option being shown after the filter is applied. We then want to use the filtered list of options and index search in that instead.

When we want to get the correct item, we need to look through the filtered list of options and then apply the index to that.

src/components/select-field/BaseSelectField.js
```diff
@@ -349,7 +360,7 @@ class BaseSelectField extends React.Component<Props, State> {
     selectMultiOption = (index: number) => {
         const { defaultValue, options, selectedValues } = this.props;
         const hasDefaultValue = defaultValue != null; // Checks if not undefined or null
-        const item = options[index];
+        const item = this.getItemFromFilteredOptions(index);
```
ChenCodes, line 363: Same goes for this line. Single / Multi Select can both be filtered on so this functionality holds true for this case.

src/components/select-field/BaseSelectField.js
```diff
@@ -469,7 +480,7 @@ class BaseSelectField extends React.Component<Props, State> {
 
             const isSelected = selectedValues.includes(value);
 
-            const isClearOption = shouldShowClearOption && index === 0;
+            const isClearOption = shouldShowClearOption && value === 'clear';
```
ChenCodes, line 483: This change is needed because the original implementation was brittle. There can be items that are index 0 but are not the clear button so we need something a little more robust.

alexkrolick: Is there a risk that a user-provided item could have the value "clear"?

ChenCodes: Good call. I just tested it and it looks like if the user provides an item with the value clear, this will not work. We still need the `index === 0` check since we always insert the Clear option at the front of the options array.

alexkrolick: Do we need the "value" check then?

ChenCodes: We still need the value check because when we are filtering via the search field, the index === 0 element may not be be the clear option so we don't want to apply the clear option styles on that.

src/components/select-field/__tests__/BaseSelectField.test.js
```diff
@@ -1023,6 +1067,24 @@ describe('components/select-field/BaseSelectField', () => {
...
+            const indexBeforeFilter = 1; // Index of audio option in default options array
+            const indexAfterFilter = 0; // Index of audio option after user inputs a search string of 'Audio'
+            const wrapper = shallowRenderSelectField();
+            const instance = wrapper.instance();
+
+            instance.setState({
+                searchText: 'Audio',
+            });
+
+            sandbox
```
sahiga, line 1080: We're moving away from Sinon mocks in BUIE. Can we use Jest mocks for the new tests in this file?

ChenCodes: Addressed!

src/components/select-field/BaseSelectField.js
```diff
@@ -336,20 +337,21 @@ class BaseSelectField extends React.Component<Props, State> {
         }
     };
 
-    getItemFromFilteredOptions = (index: number) => {
+    getFilteredOptions = () => {
```
ChenCodes, line 340: Refactored this because renderSelectOptions method uses the filtering mechanism so I thought I'd have this method only handle the filtering, getting an element with the index can be done at a lower level.

src/components/select-field/BaseSelectField.js
```diff
@@ -336,20 +337,21 @@ class BaseSelectField extends React.Component<Props, State> {
...
         const { options } = this.props;
         const { searchText } = this.state;
 
-        const filteredOptions = options.filter(option =>
-            searchText ? option.displayText.toLowerCase().includes(searchText.toLowerCase()) : true,
-        );
+        return options.filter(option => {
+            const isSubstring = option.displayText.toLowerCase().includes(searchText.toLowerCase());
+            const isClearOption = option.value === CLEAR;
```
ChenCodes, line 346: As requested by design, we should be filtering out the Clear Option as part of the options when we're performing searching.

src/components/select-field/constants.js
```diff
@@ -0,0 +1,4 @@
+// @flow
+const CLEAR: '__clear__' = '__clear__'; // value of clear option used in select-field components
```
ChenCodes, line 2: Using a magic constant like this will help us avoid any collisions where an enduser might have an option with a value of "clear". It is highly unlikely that they will have the same constant as us now.

src/components/select-field/__tests__/BaseSelectField.test.js
```diff
@@ -1227,17 +1245,15 @@ describe('components/select-field/BaseSelectField', () => {
...
             });
 
-            sandbox
-                .mock(instance)
-                .expects('handleOptionSelect')
-                .withArgs(options[indexBeforeFilter]); // audio + video
-
             instance.selectMultiOption(indexAfterFilter);
+
+            expect(spy).toHaveBeenCalledWith(options[indexBeforeFilter]); // audio + video
```
ChenCodes, line 1256: No longer using sandbox.

src/components/select-field/BaseSelectField.js
```diff
@@ -480,7 +480,7 @@ class BaseSelectField extends React.Component<Props, State> {
 
             const isSelected = selectedValues.includes(value);
 
-            const isClearOption = shouldShowClearOption && value === 'clear' && index === 0;
+            const isClearOption = shouldShowClearOption && value === CLEAR;
```
ChenCodes, line 483: Using the magic constant CLEAR should be robust enough for us to not have to use index === 0.

PR #2132, opened 2020-05-13
fix(collabsiblesidebar): add a label for expand/collapse button

alexkrolick:


src/features/collapsible-sidebar/__tests__/__snapshots__/CollapsibleSidebarLogo.test.js.snap
```diff
@@ -1,86 +1,296 @@
...
+</CollapsibleSidebarLogo>
 `;
 
 exports[`components/core/collapsible-sidebar/CollapsibleSidebar render 2`] = `
-<div
-  className="bdl-CollapsibleSidebar-logo"
+<CollapsibleSidebarLogo
+  canEndTrial={true}
 >
-  <CollapsibleSidebarItem
```
booc0mtaco, line 50: Should there be a `CollapsibleSidebarLogo` inside itself?

alexkrolick: I guess that is what happens when it serializes the `injectIntl(Comp)`

PR #2133, opened 2020-05-14
fix(select-field): use custom selector for SearchForm component

ChenCodes:
Changes in this PR:
- we should be using a custom selector to target the search container when it is being used inside of SelectField components so that we are not affecting other components on the page that have a SearchForm even though the SearchForm does not live inside the SelectField component

Screen Shot 2020-05-14 at 11 47 19 AM

ezgif com-video-to-gif (4)

### DIFF
--- a/src/components/select-field/SelectField.scss
+++ b/src/components/select-field/SelectField.scss
@@ -12,6 +12,7 @@
     }
 
     .select-field-search-container {
+        width: 100%;
         padding-right: $bdl-grid-unit;
         padding-bottom: $bdl-grid-unit * 2;
         padding-left: $bdl-grid-unit;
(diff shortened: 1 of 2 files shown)


src/components/select-field/SelectField.scss
```diff
@@ -10,16 +10,15 @@
         margin: 10px 0;
         border-bottom: 1px solid $bdl-gray-20;
     }
-}
 
-.search-input-container {
-    width: 100%;
```
ChenCodes, line 16: Removed because unnecessary

src/components/select-field/SelectField.scss
```diff
@@ -10,16 +10,15 @@
...
-    padding-right: $bdl-grid-unit;
-    padding-bottom: $bdl-grid-unit * 2;
-    padding-left: $bdl-grid-unit;
+    .select-field-search-container {
```
ChenCodes, line 14: Moved to be a descendant selector of .select-field

PR #2135, opened 2020-05-15
feat(annotations): Add delete flow

mickr:
* Add deleteAnnotation handler
* Add emitting function for annotation removal

- [x] unit tests

### DIFF
--- a/src/elements/common/annotator-context/types.ts
+++ b/src/elements/common/annotator-context/types.ts
@@ -29,6 +29,7 @@ export type GetMatchPath = (history: History) => match<MatchParams> | null;
 
 export interface AnnotatorContext {
     emitActiveChangeEvent: (id: string) => void;
+    emitRemoveEvent: (id: string) => void;
     getAnnotationsMatchPath: GetMatchPath;
     state: AnnotatorState;
 }
--- a/src/elements/common/annotator-context/withAnnotations.tsx
+++ b/src/elements/common/annotator-context/withAnnotations.tsx
@@ -11,6 +11,7 @@ export interface WithAnnotationsProps {
 
 export interface ComponentWithAnnotations {
     emitActiveChangeEvent: (id: string | null) => void;
+    emitRemoveEvent: (id: string) => void;
     getAction: (eventData: AnnotationActionEvent) => Action;
     getMatchPath: GetMatchPath;
     handleActiveChange: (annotationId: string | null) => void;
@@ -65,6 +66,16 @@ export default function withAnnotations<P extends object>(
             }
 
             annotator.emit('annotations_active_set', id);
+        };
+
+        emitRemoveEvent = (id: string) => {
+            const { annotator } = this;
+
+            if (!annotator) {
+                return;
+            }
+
+            annotator.emit('annotations_remove', id);
         };
 
         getAction({ meta: { status }, error }: AnnotationActionEvent): Action {
@@ -123,6 +134,7 @@ export default function withAnnotations<P extends object>(
             return (
                 <AnnotatorContext.Provider
                     value={{
+                        emitRemoveEvent: this.emitRemoveEvent,
                         emitActiveChangeEvent: this.emitActiveChangeEvent,
                         getAnnotationsMatchPath: this.getMatchPath,
                         state: this.state,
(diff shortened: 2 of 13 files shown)


src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -35,7 +29,7 @@ type Props = {
     getUserProfileUrl?: GetProfileUrlCallback,
     items: FeedItems,
     mentionSelectorContacts?: SelectorItems<>,
-    onAnnotationDelete?: ({ id: string, permissions?: BoxAnnotationPermission }) => void,
+    onAnnotationDelete?: (id: string) => void,
```
ConradJChan, line 32: Why are the permissions being removed?

src/elements/common/annotator-context/types.ts
```diff
@@ -29,6 +29,7 @@ export type GetMatchPath = (history: History) => match<MatchParams> | null;
 
 export interface AnnotatorContext {
     emitActiveChangeEvent: (id: string) => void;
+    emitDeleteChangeEvent: (id: string) => void;
```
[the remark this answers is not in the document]
mickr: Yeah, that sounds better.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
mickr: I don't think we need them in the call to delete. The permissions should apply whether you see the delete item / context menu but shouldn't be required as part of the call to delete. When I added this flow originally I modeled it after comment.

ConradJChan: That's probably true, but it might still be good to include in the api call -- there appears to be a `checkApiCallValidity` method used by the API classes which would have a trickle up(?) effect on the affected code. Developers might use the API classes directly like how we are for BoxAnnotations

mickr: Re-added the permissions and we can implement them in the api call when we do the permissions slice of work.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -227,6 +227,22 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         }
     };
 
+    handleAnnotationDelete = ({ id }: { id: string, permissions?: BoxAnnotationPermission }) => {
+        const { api, emitRemoveEvent, file } = this.props;
+
+        api.getFeedAPI(false).deleteAnnotation(
+            file,
+            id,
+            () => {
```
ConradJChan, line 236: This might make it a little harder to test with mock functions -- would it be better to create a named function?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -67,6 +68,16 @@ export default function withAnnotations<P extends object>(
...
         };
 
+        emitRemoveEvent = (id: string) => {
+            const { annotator } = this;
+
+            if (!annotator) {
+                return;
+            }
+
+            annotator.emit('annotations_remove', id);
```
jstoffan, line 78: If we get more of these, we'll probably want to invest in a factory for method -> event pairs. Not needed right now, though.

src/elements/common/annotator-context/withAnnotatorContext.tsx
```diff
@@ -5,18 +5,20 @@ import { AnnotatorState, GetMatchPath } from './types';
...
     return React.forwardRef<React.RefForwardingComponent<React.ComponentType<P>>, P>((props, ref) => (
         <AnnotatorContext.Consumer>
-            {({ emitActiveChangeEvent, getAnnotationsMatchPath, state }) => (
+            {({ emitActiveChangeEvent, emitRemoveEvent, getAnnotationsMatchPath, state }) => (
                 <WrappedComponent
                     ref={ref}
                     {...props}
                     annotatorState={state}
                     emitAnnotatorActiveChangeEvent={emitActiveChangeEvent}
+                    emitRemoveEvent={emitRemoveEvent}
```
jstoffan, line 21: Should this be `emitAnnotatorRemoveEvent` to stay consistent with the the above?

PR #2138, opened 2020-05-18
feat(permissions): Add permissions checks to annotations api endpoints

jstoffan:


src/api/Annotations.js
```diff
@@ -1,7 +1,18 @@
 // @flow
 import merge from 'lodash/merge';
+import type { BoxAnnotationPermission } from '../common/types/feed';
```
ConradJChan, line 3: Should this be located below with the other type imports?

jstoffan: 😱

PR #2139, opened 2020-05-19
Fix select field stuck open / enter key reloads page

ChenCodes:
Changes in this PR:
- no longer call closeDropdown in handleBlur if the user clicks on the select field button because handleButtonClick will be called after handleBlur (before there was an issue where the dropdown is closed via handleBlur, then incorrectly opened again by handleButtonClick)
- ignore 'Enter' press when the search field is active and no item is selected (before there was an issue where there would be a page navigation which is happening since the search form uses a form input

ezgif com-video-to-gif (6)


src/components/select-field/BaseSelectField.js
```diff
@@ -256,9 +257,17 @@ class BaseSelectField extends React.Component<Props, State> {
...
-                    break;
+                if (shouldShowSearchInput) {
+                    // Allow space key presses in the search string when search field is active
+                    if (key === ' ') {
+                        break;
+                    }
+
+                    // Enter presses should be ignored when no item is active
+                    if (key === 'Enter' && activeItemIndex === -1) {
+                        stopDefaultEvent(event);
```
ChenCodes, line 268: This change is needed because we do not want the 'Enter' keypress to submit the form if there is no active item selected. The bug that was happening was that when the search field was empty (or when no results were shown) and the user pressed the Enter key, the page would reload. We want to have Enter keypresses ignored when Enter is pressed and no item is highlighted.

src/components/select-field/BaseSelectField.js
```diff
@@ -224,7 +224,8 @@ class BaseSelectField extends React.Component<Props, State> {
             isOpen &&
             event &&
             event.relatedTarget &&
-            !(event.relatedTarget: window.HTMLInputElement).classList.contains('search-input')
+            !(event.relatedTarget: window.HTMLInputElement).classList.contains('search-input') &&
+            !(event.relatedTarget: window.HTMLInputElement).classList.contains('select-button')
```
ChenCodes, line 228: The reason for this change is as follows:
Since we now have a search input component inside of the selector dropdown, that search input element becomes focused immediately when the dropdown is open. When the user clicks on the Select Field button, this will fire a `handleBlur` call. After the `handleBlur` call is fired, `handleButtonClick` is then fired. The issue that we run into then is, we call `closeDropdown` in `handleBlur`, then `openDropdown` in `handleButtonClick` since in state `isOpen` is already set to `false`. Instead, we only want closeDropdown to be called once, and that should be done in handleButtonClick rather than handleBlur (in the scenario where the user is clicking on the .select-button.

ChenCodes: This is why the dropdown seems to be stuck open; we close the dropdown then immediately open the dropdown because `handleButtonClick` is fired immediately after `handleBlur`.

src/components/select-field/BaseSelectField.js
```diff
@@ -256,9 +257,17 @@ class BaseSelectField extends React.Component<Props, State> {
...
-                if (key === ' ' && shouldShowSearchInput) {
-                    break;
+                if (shouldShowSearchInput) {
+                    // Allow space key presses in the search string when search field is active
+                    if (key === ' ') {
+                        break;
+                    }
+
+                    // Enter presses should be ignored when no item is active
+                    if (key === 'Enter' && activeItemIndex === -1) {
```
sahiga, line 267: Minor: can we use the constants defined in common/keyboard-events.js?

PR #222, opened 2018-03-14
Chore: Minor folder structure cleanup

priyajeet:
Move skills related code under sidebar folder.
Use left/right navigation arrows from box-react-ui.

PR #2141, opened 2020-05-20
feat(virtualized-table): Adding DraggableVirtualizedTable component

jpfranco:
This update adds the new `DraggableVirtualizedTable` component, which is basically a wrapper for `VirtualizedTable` that adds drag and drop support for the rows. 

The new component uses the same interface as `VirtualizedTable`, so it can be used in the same way without any additional setup apart from passing a required `tableId` prop. The logic to re-order rows needs to be implemented in the parent component as is shown on the storybook example.

Note: Given that the new code depends on components with haven't been migrated to TypeScript yet, the new component is using flow types for now. TypeScript migration is outside the scope of this PR.

DraggableVirtualizedTable


src/features/virtualized-table-renderers/__tests__/__snapshots__/draggableRowRenderer.test.js.snap
```diff
@@ -0,0 +1,29 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`features/virtualized-table-renderers/draggableRowRenderer should wrap row in Draggable and set correct props 1`] = `
```
booc0mtaco, line 3: I think for snapshots like this, we should consider regular `expect` checks. There is no DOM structure here in the snapshot.

src/features/virtualized-table/DraggableVirtualizedTable.js
```diff
@@ -0,0 +1,75 @@
+// @flow
+import * as React from 'react';
+import noop from 'lodash/noop';
+import classNames from 'classnames';
+import { Column } from 'react-virtualized/dist/es/Table/index';
+import { DragDropContext, Droppable } from 'react-beautiful-dnd';
+
+import { black } from '../../styles/variables';
```
booc0mtaco, line 8: Most things shouldn't be pure black; here we may actually want #222 (`$bdl-gray`). Can you confirm with design on this?

src/features/virtualized-table/__tests__/__snapshots__/DraggableVirtualizedTable.test.js.snap
```diff
@@ -0,0 +1,89 @@
...
+      headerRenderer={[Function]}
+      label="Description"
+      style={Object {}}
+      width={100}
+    />
+  </VirtualizedTable>
+</div>
+`;
+
+exports[`features/virtualized-table/DraggableVirtualizedTable should render a drag icon when drag handle cell renderer is called 1`] = `
```
booc0mtaco, line 83: Another example where snapshot is probably not useful and would be better as an assertion

src/features/virtualized-table/VirtualizedTable.js
```diff
@@ -10,17 +10,17 @@ import './VirtualizedTable.scss';
 
 type SortHandler = SortParams => void;
 
-type Props = {
+export type VirtualizedTableProps = {
     children: Node | (any => Node),
     className?: string,
     height?: number,
-    intl: any,
+    intl?: any,
```
booc0mtaco, line 17: I think the type is `InjectIntlProvidedProps`, no?

src/features/virtualized-table/DraggableVirtualizedTable.stories.js
```diff
@@ -0,0 +1,60 @@
...
+import DraggableVirtualizedTable from './DraggableVirtualizedTable';
+
+export const Basic = () => {
+    const initialRowData = [
+        { id: '1', name: 'Product A', description: 'Product Description A', price: 4.99 },
+        { id: '2', name: 'Product B', description: 'Product Description B', price: 2.99 },
+        { id: '3', name: 'Product C', description: 'Product Description C', price: 3.99 },
+        { id: '4', name: 'Product D', description: 'Product Description D', price: 1.99 },
+    ];
+    const [rowData, setRowData] = React.useState(initialRowData);
```
sahiga, line 19: +1 on using Hooks here

src/features/virtualized-table-renderers/__tests__/__snapshots__/draggableRowRenderer.test.js.snap
jpfranco: Agreed. I actually have some regular checks on that test. Let me see if I can replace the snapshot with some more `expect`s.

src/features/virtualized-table/VirtualizedTable.js
jpfranco: Ah yes, missed this. Will update

src/features/virtualized-table/DraggableVirtualizedTable.js
```diff
@@ -0,0 +1,75 @@
...
+
+    const handleDragEnd = ({ destination, source }) => {
+        const destinationIndex = destination ? destination.index : source.index;
+        return onDragEnd(source.index, destinationIndex);
+    };
+
+    return (
+        <DragDropContext onDragEnd={handleDragEnd}>
+            <Droppable droppableId={tableId}>
+                {droppableProvided => (
```
sahiga, line 45: What is the type of `droppableProvided`?

src/features/virtualized-table-renderers/draggableRowRenderer.js
```diff
@@ -0,0 +1,59 @@
...
+    const { index, key, rowData } = params;
+    const defaultRow = defaultRowRenderer(params);
+    // Keys are auto-generated by the virtualized table based on
+    // the row index. We give preference to id when available since
+    // the index changes (and thus also the key) when the rows are re-ordered
+    const draggableId = rowData.id || key;
+
+    return (
+        <Draggable draggableId={draggableId} index={index} key={draggableId}>
+            {(draggableProvided, draggableSnapshot) => {
```
sahiga, line 21: Types for `draggableProvided` and `draggableSnapshot`?

src/features/virtualized-table/DraggableVirtualizedTable.scss
```diff
@@ -0,0 +1,29 @@
...
+            margin-right: 0;
+        }
+    }
+}
+
+.bdl-DraggableVirtualizedTable-dragHandleColumn {
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+    width: 24px;
```
booc0mtaco, line 28: if this is 24px for the same reason `ICON_SIZE` is 24px, then we could create a token in our variables for such a thing. there is no effort to do this right now, but something to think about

PR #2142, opened 2020-05-20
feat(classification): New solid security classification icon

patrickpaul:
Added the new version of the solid security classification icon which has a filled circle behind the non-solid version.

Additionally fixed a missing export from /index.js that I saw.

Screenshot:
Screen Shot 2020-05-20 at 2 55 09 PM


booc0mtaco:
can you include a screenshot for this? it looks like it has two colors and one is always grey

PR #2143, opened 2020-05-20
feat(task): use atomic api for create/update task, tack-link, and task-collab

sliu2013:


src/api/Feed.js
```diff
@@ -772,20 +771,31 @@ class Feed extends Base {
                     return;
                 }
 
+                this.addPendingItem(this.file.id, currentUser, pendingTask);
```
booc0mtaco, line 774: this line seems suspicious. previously, we did this in `successCallback` after calling the old `.createTask`

sliu2013: Ya, you are right. I should move this back inside the successCallBack so that if the createTaskWithDeps API call fails, we won't end up with a leftover pendingItem.

PR #2144, opened 2020-05-20
style(tooltips): update tooltip to BDL Dark Blue

laughnan:
We're changing all the purple tooltips to the dark blue.

Screen Shot 2020-05-20 at 10 02 33


src/components/tooltip/Tooltip.scss
```diff
@@ -47,8 +47,9 @@ $tooltip-offset: 4px; /* amount of space between target and tooltip arrow */
     }
 
     &.is-callout {
-        background-color: $bdl-purple-rain;
-        border-color: $bdl-purple-rain;
+        color: $bdl-dark-blue-02;
+        background-color: $bdl-dark-blue;
+        border-color: $bdl-dark-blue;
```
jellee, line 52: double checking that these variables already exist (in case these are new colors)

laughnan: they exist 😄

jellee: 👍 😂

PR #2145, opened 2020-05-20
feat(draftjs): Move more functions into utils

mxiao6:


src/components/form-elements/draft-js-mention-selector/createMentionSelectorState.js
```diff
@@ -1,45 +0,0 @@
...
-    }
-    const [fullMatch, triggerSymbol, id, name] = matchArray;
-    const start = matchArray.index;
-    const end = start + fullMatch.length;
-    const data = { id, name, content: `${triggerSymbol}${name}` };
-    return { start, end, data };
-};
-
-// creates draftjs state with mentions parsed into entities
-const createMentionSelectorState = (message: string = '') => {
```
jstoffan, line 21: Does this need to move? The utils file is still in the same `form-elements` component folder. I think we should generally try to minimize catch-all util files rather than expand them.

PR #2149, opened 2020-05-22
feat(avatar): add support for showing avatars with a badge for external

booc0mtaco:
- add two new prop.s for `Avatar`: `isExternal` which indicates if the current user avatar is an external user in the given context, and `shouldShowExternal` (default `false`) which indicates if the component should show any external indicator. 
- uses `Badgeable` which has been updated to have a TypeScript version and to include mixins for any positioning overrides when used


src/components/avatar/Avatar.scss
```diff
@@ -54,6 +68,19 @@
         .avatar-initials {
             font-size: 11px;
         }
+
+        &.avatar--isExternal {
+            .avatar-external-badge {
+                width: 16px;
```
danubevictoria, line 74: minor comment: just not sure why we have a bigger sizing for a small avatar? why not just use the above class styling?

src/components/avatar/Avatar.tsx
```diff
@@ -19,21 +21,29 @@ export interface AvatarProps {
...
+    /** Whether this avatar should be labeled as external in the current context */
+    isExternal?: boolean;
     /**
      * Users full name.
      *
      * Required if "avatarUrl" is not specified.
      */
     name?: string | null;
+    /** Show the external avatar marker if the avatar is marked as for an external user */
+    shouldShowExternal?: boolean;
```
danubevictoria, line 33: is shouldShowExternal the feature flip? shouldn't we always just show the external badge if it's an external user?

booc0mtaco: yes we should, tho some interfaces can reveal this information in unique ways, and others may already include a badge that needs to be removed

src/components/avatar/Avatar.scss
booc0mtaco: this accounts for the total size, including the border. due to how the asset sizing is calculated, and the design requirements, we want the asset to have 14px size with 2px border

src/components/avatar/Avatar.tsx
```diff
@@ -19,21 +21,29 @@ export interface AvatarProps {
...
-function Avatar({ avatarUrl, className, name, id, size = '' }: AvatarProps) {
+function Avatar({ avatarUrl, className, name, id, isExternal, shouldShowExternal = false, size = '' }: AvatarProps) {
     const [hasImageErrored, setHasImageErrored] = React.useState<boolean>(false);
     const [prevAvatarUrl, setPrevAvatarUrl] = React.useState<AvatarProps['avatarUrl']>(null);
 
-    const classes = classNames(['avatar', className, { [`avatar--${size}`]: size && SIZES[size] }]);
+    const classes = classNames([
+        'avatar',
+        className,
+        { [`avatar--${size}`]: size && SIZES[size], 'avatar--isExternal': shouldShowExternal && isExternal },
```
ConradJChan, line 45: Should these classes be namespace'd and SUIT convention'ed?

src/components/badgeable/Badgeable.tsx
```diff
@@ -0,0 +1,35 @@
...
+    topLeft?: React.ReactNode;
+    /** Component to render when badging the top right corner of the rendered container */
+    topRight?: React.ReactNode;
+};
+
+const Badgeable = (props: Props) => {
+    const { children, className = '', topLeft = null, topRight = null, bottomLeft = null, bottomRight = null } = props;
+
+    return (
+        <div className={`badgeable-container ${className}`}>
```
ConradJChan, line 23: Should this component's CSS classes be converted to SUIT convention?

Should we use `classNames` here to avoid the extra space if `className` prop is empty?

booc0mtaco: i would defer refactoring in this case, as there are likely many overrides associated with this component. That's why i moved to the mixins; we can then refactor to use those in consuming apps, and any class name changes automatically woke on future refactoring.

src/components/avatar/Avatar.tsx
booc0mtaco: So s/`avatar`/`bdl-Avatar`/ ?

PR #2150, opened 2020-05-22
feat(contact-datalist-item): add support for avatars in the dropdown

booc0mtaco:


src/features/unified-share-modal/ContactsField.js
```diff
@@ -87,7 +92,7 @@ class ContactsField extends React.Component<Props, State> {
                 fullContacts = this.addSuggestedContacts(fullContacts);
             }
 
-            return fullContacts.map<Object>(({ email, id, isExternalUser, name, type }) => ({
+            return fullContacts.map<any>(({ email, id, isExternalUser, name, type }) => ({
```
danubevictoria, line 95: was there a reason this was changed to 'any'?

src/components/contact-datalist-item/__tests__/ContactDatalistItem.test.js
```diff
@@ -9,11 +9,19 @@ describe('components/contact-datalist-item/ContactDatalistItem', () => {
         expect(wrapper.find('DatalistItem').length).toBe(1);
         expect(wrapper.find('.contact-name').text()).toEqual('name');
         expect(wrapper.find('.contact-sub-name').text()).toEqual('subtitle');
+        expect(wrapper.find('Avatar').length).toBe(0);
```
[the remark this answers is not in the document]
booc0mtaco: to me, it made sense to check for the default state in the existing tests, to makes sure it does not get enabled by default

src/features/unified-share-modal/ContactsField.js
booc0mtaco: this was a carry over from the old PR. `Object` doesn't have a strong meaning in flow. 

https://github.com/facebook/flow/issues/7291

The correct type here would be whatever object results from this `.map` call, which wasn't typed earlier. That is part of the TODO directly above.

src/features/unified-share-modal/ContactsField.js
```diff
@@ -28,6 +28,7 @@ type Props = {
     onInput?: Function,
     onPillCreate?: (pills: Array<SelectOptionProp | Contact>) => void,
     selectedContacts: Array<Contact>,
+    showInviteeAvatars?: boolean,
```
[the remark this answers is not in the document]
booc0mtaco: actually, a better name here would be `showContactAvatars` probably, since other prop names deal with contacts and both components mention contact. I will see how that feels in practice

src/features/unified-share-modal/ContactsField.js
il10, line 95: any reason the type here is being changed from `Object` to `any` ?

booc0mtaco: this was pulled from the previous big PR, so i kept it. since there are a few questions around this, lemme add a more useful type to this file (This has to do with the TODO near this line)

PR #2151, opened 2020-05-26
fix(collapsiblesidebar): Update menu item to take custom content

indiramuru:
Design: https://app.zeplin.io/project/5d813c56b87bf954d7b9074e/screen/5ec6da8e0bcf374d9a3b7cf9

Screen Shot 2020-05-26 at 2 34 35 PM


booc0mtaco:
Are there no constraints on what can be added to a particular menu item? for instance, what if someone got confused and tried to embed a link within a menu item, or inserted something which has a height or behavior inappropriate to a single menu item?

More of a design question but, what is an example of how this will be used in the future?

indiramuru:
@booc0mtaco This change is mainly to accommodate a requirement in Admin Console where beta/trial tag can be displayed next to menu item on need basis. Have added an example in *.stories.js to show the usage.
if the user does have an additional element to add to the menu then `content` prop is expected to be used. This would let them add anything but the css should be adjusted at the users side to keep the UI intact.

booc0mtaco:
@indiramuru I see some concerns with this approach that may come up later

## translations

- What should happen when a locale uses a much longer word than 6 characters? 
- What if the additional content text in the pills also translates to much longer words. What should remain on screen, and what should collapse into the ellipsis?

## Component API

- what would happen if a user uses `text`, `icon`, and `content` for a menu item? it seems that your example implies that one should not use `text` if you use `content`, and that one must know that `text` means implementing `StyledMenuItemLabel`? 
- There could be a use case to somehow alter the collapsed view in some way as well. This may require changes to the icon component. 

## positioning

- because the content sits outside of `StyledMenuItemLabel`, it cannot take advantage of any of the menu item handling built into that container. 
- ideally there would be some mechanism to enforce layout when using `content`, but right now it is placed directly into the menu. `icon` and `text` props use wrappers to constrain the content, and i would think this one should as well

## designs

- Recent designs have a used a more rounded pill (see `LabelPill` or the Classification Badge feature component). Should we be using this style of component in our examples?

---

Instead of a react node being added directly to the menu item via a prop, I would consider another option, like allowing `CollapsibleSidebarMenuItem` to have child nodes where a user can supply the structure they desire (and warn against using both simultaneously, or not including the menu item sub components). 

Maybe there is another alternative? (cc @alexkrolick @krapspark )

indiramuru:
@booc0mtaco Agree with your concerns. Came up with this initial approach based on the discussion with Sam Park. Idea was to make the content as generic as possible and not to restrict on what they can take. 
translations - idea was to let user send in the css to control the overflowing text.
component API -Yes. For the expanded view, user should be passing in either `text` or `content`. 
positioning - yeah wrapping the content in styled component would give it a structure
design - Can change it in the example. used those(beta/trial)  as we are going to use it in AC until the new badge is ready in BUIE. 

Q on your suggestion: Do you mean to either pass in `children` or `icon`, `text/content`, etc.. So letting the user to define menu item content to be anything (ex: with/without icons + anything)? How the css is expected to be handled in this case?

booc0mtaco:
@indiramuru There is no indication in this PR or in the code that one should pass either text or content, but not both. Although i do think there exists a design that can avoid this ambiguity, can we document this somewhere in the component if we take this path?

As for my suggestion, I will give an example. With the props, one currently can control the icon and/or text:

```
<CollapsibleSidebarMenuItem className="something" icon={icon} text="some text" />
```

We can add an additional prop, but that does introduce some ambiguity as we have discussed. Instead, I was making a naïve suggestion to do something like this in cases where we want to control the output:

```
<CollapsibleSidebarMenuItem className="something">
    <StyledIconWrapper className="bdl-CollapsibleSidebar-menuItemIcon"><CheckmarkBadge16 height={20} width={20} /></StyledIconWrapper>
    <StyledMenuItemLabel ref={textRef} className="bdl-CollapsibleSidebar-menuItemLabel">SomeText</StyledMenuItemLabel>
    <BetaBadge style={{marginLeft: 8}} />
</CollapsibleSidebarMenuItem>
```

then I realized the refs and class names would need to be passed every time based on how it is built. Probably not a small change after all.

----

Instead, if we know that one should use either `text` or `content` but not both at the same time, we can add a check that prevents both from being used simultaneously.

indiramuru:
@booc0mtaco have addressed your comments. Can I get a review again?

src/features/collapsible-sidebar/__tests__/CollapsibleSidebarMenuItem.test.js
```diff
@@ -32,6 +32,35 @@ describe('components/core/collapsible-sidebar/__tests__/CollapsibleSidebarMenuIt
...
+
+        const wrapper = getWrapper(
+            {
+                className: 'foo',
+                content: <div>Custom Content</div>,
+                icon: 'bold',
+            },
+            { isScrolling: false },
+        );
+        expect(wrapper).toMatchSnapshot();
```
krapspark, line 46: for the actual check, we should check for the presence of the content instead of relying on snapshots.

src/features/collapsible-sidebar/__tests__/CollapsibleSidebarMenuItem.test.js
```diff
@@ -32,6 +32,35 @@ describe('components/core/collapsible-sidebar/__tests__/CollapsibleSidebarMenuIt
...
+        const wrapper = getWrapper(
+            {
+                className: 'foo',
+                content: <div>Custom Content</div>,
+                icon: 'bold',
+                text: 'bar',
+            },
+            { isScrolling: false },
+        );
+        expect(wrapper).toMatchSnapshot();
```
krapspark, line 61: same here

PR #2153, opened 2020-05-27
feat(annotations): Rewrite deep-linked file version to current

ConradJChan:
Handles the scenario where preview is direct loaded with a partial deep link to an annotation, i.e. `?sb=/activity/annotations/:fileVersionId` instead of `?sb=/activity/annotations/:fileVersionId/:annotationId`.

This can lead to a bug where clicking on an annotation that is actually on the previous file version will not trigger a file version change because according to the location, we're already on that file version.

In this scenario we will rewrite the url to be the current file version id, so as to maintain the consistency of the `sb` location being the source of truth for which version is being previewed

### DIFF
--- a/src/elements/content-sidebar/ActivitySidebar.js
+++ b/src/elements/content-sidebar/ActivitySidebar.js
@@ -151,7 +151,7 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         const currentFileVersionId = getProp(file, 'file_version.id');
         const fileVersionId = getProp(match, 'params.fileVersionId');
 
-        if (fileVersionId && annotationId && fileVersionId !== currentFileVersionId) {
+        if (fileVersionId && fileVersionId !== currentFileVersionId) {
             history.replace(this.getAnnotationsPath(currentFileVersionId, annotationId));
         }
     };
--- a/src/elements/content-sidebar/__tests__/ActivitySidebar.test.js
+++ b/src/elements/content-sidebar/__tests__/ActivitySidebar.test.js
@@ -863,7 +863,7 @@ describe('elements/content-sidebar/ActivitySidebar', () => {
             ${'124'}      | ${undefined} | ${1}              | ${'/activity/annotations/123'}
         `(
             'should call history.replace appropriately if router location annotationId=$annotationId and fileVersionId=$fileVersionId',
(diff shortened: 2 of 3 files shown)


src/elements/content-sidebar/__tests__/ActivitySidebar.test.js
```diff
@@ -856,11 +856,11 @@ describe('elements/content-sidebar/ActivitySidebar', () => {
...
-            ${'124'}      | ${'987'}     | ${1}
-            ${'124'}      | ${undefined} | ${0}
+            fileVersionId | annotationId | expectedCallCount | expectedPath
+            ${undefined}  | ${'987'}     | ${0}              | ${undefined}
+            ${'123'}      | ${'987'}     | ${0}              | ${undefined}
+            ${'124'}      | ${'987'}     | ${1}              | ${'/activity/annotations/123/987'}
+            ${'124'}      | ${undefined} | ${1}              | ${'/activity/annotations/123'}
         `(
             'should call history.replace appropriately if router location annotationId=$annotationId and fileVersionId=$fileVersionId',
             ({ annotationId, fileVersionId, expectedCallCount }) => {
```
mickr, line 866: Does this variable need updated?

ConradJChan: Good catch, updated

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -151,7 +151,7 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         const currentFileVersionId = getProp(file, 'file_version.id');
         const fileVersionId = getProp(match, 'params.fileVersionId');
 
-        if (fileVersionId && annotationId && fileVersionId !== currentFileVersionId) {
+        if (fileVersionId && fileVersionId !== currentFileVersionId) {
```
jstoffan, line 154: This method isn't really specific to annotations anymore, since it's always called on mount and doesn't check for annotation id. Should we name it `redirectInitialPath` or something?

ConradJChan: I think I'll leave it for now, it still relies on the `getAnnotationsMatchPath` to check for and set back an annotations location path

PR #2155, opened 2020-05-27
fix(usm): Remove base styles; enable loading indicator

sahiga:
This PR enables the loading indicator in the USM by default and removes base styles from the USM CSS file, which were causing conflicts in parent apps. I caused these bugs in #2129. Additionally, I updated the visual regression tests for the USM to reflect these changes and the new tooltip colors introduced in #2144.


src/features/unified-share-modal/UnifiedShareModal.js
```diff
@@ -45,7 +45,7 @@ class UnifiedShareModal extends React.Component<USMProps, State> {
             getInitialDataCalled: false,
             isConfirmModalOpen: false,
             isEmailLinkSectionExpanded: false,
-            isFetching: false,
```
booc0mtaco, line 48: Why do we change this default?

src/features/unified-share-modal/UnifiedShareModal.scss
```diff
@@ -275,6 +274,7 @@
 .bdl-UnifiedShareForm {
     @extend .unified-share-modal;
 
+    position: relative;
```
booc0mtaco, line 277: What is this relative to?

src/features/unified-share-modal/UnifiedShareModal.js
sahiga: It was originally set to `true`. I accidentally changed it to `false` in my last PR: https://github.com/box/box-ui-elements/pull/2129/files#diff-a6be9fd7eb644ea8a0a16cc96b5a65b5L179. This ensures that the loading indicator shows up on load.

src/features/unified-share-modal/UnifiedShareModal.scss
sahiga: Good catch. I forgot to remove this.

PR #2157, opened 2020-05-28
chore(sidebar): hoc to forward ref through withRouter

ConradJChan:
This fixes the warning that appears in the console because react-router `withRouter` used with ActivitySidebar complains about forwarding refs and functional components.

Introduces a new HOC `withRouterAndRef` that forwards the ref on to the wrapped component


src/elements/common/with-router-and-ref/withRouterAndRef.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+import * as React from 'react';
+import { withRouter } from 'react-router-dom';
+
+export default function withRouterAndRef(Wrapped: React.ComponentType<any>) {
+    const WithRouter = withRouter(({ forwardRef, ...otherProps }) => <Wrapped ref={forwardRef} {...otherProps} />);
+    const WithRouterAndRef = React.forwardRef<Object, React.Ref<any>>((props, ref) => (
+        <WithRouter {...props} forwardRef={ref} />
+    ));
```
mxiao6, line 9: Can we combine these two?

ConradJChan: Hmm I'm not sure, what did you have in mind?

src/elements/common/with-router-and-ref/__tests__/withRouterAndRef.test.js
```diff
@@ -0,0 +1,26 @@
+// @flow
+import React from 'react';
+import { MemoryRouter } from 'react-router-dom';
+import { mount } from 'enzyme';
+import withRouterAndRef from '../withRouterAndRef';
+
+describe('elements/common/with-router-and-ref/withRouterAndRef', () => {
```
jstoffan, line 7: Could this be `elements/common/routing` or something to try to avoid folder sprawl?

src/elements/common/with-router-and-ref/withRouterAndRef.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+import * as React from 'react';
+import { withRouter } from 'react-router-dom';
+
+export default function withRouterAndRef(Wrapped: React.ComponentType<any>) {
+    const WithRouter = withRouter(({ forwardRef, ...otherProps }) => <Wrapped ref={forwardRef} {...otherProps} />);
```
jstoffan, line 6: Seems like you could just use `<Route>` at this point?

src/elements/common/with-router-and-ref/withRouterAndRef.js
mxiao6: Never mind, I just ended up moving `WithRouter` into `WithRouterAndRef` body, which is not elegant.

PR #2161, opened 2020-05-29
feat(thumbnail-card): add hover state to thumbnail card

ChenCodes:
Changes in this PR:
- thumbnail cards now show a border on hover

Screen Shot 2020-07-09 at 12 51 17 PM
Screen Shot 2020-07-09 at 12 51 11 PM


examples/src/ThumbnailCardExamples.js
```diff
@@ -1,63 +1,118 @@
+// @flow
 import * as React from 'react';
 
 import ThumbnailCard from '../../src/components/thumbnail-card';
+import PlainButton from '../../src/components/plain-button/PlainButton';
```
ConradJChan, line 5: minor: alphabetizing the imports helps readability

examples/src/ThumbnailCardExamples.js
```diff
@@ -1,63 +1,118 @@
...
-            <div>
-                <h1>Single Card Example With Icon and Subtitle</h1>
-                <ThumbnailCard icon={icon} subtitle={subtitle} thumbnail={thumbnail} title={title} />
+type Props = {};
+
+type State = {
+    isActionItemActive: boolean,
+};
+
+class ThumbnailCardExamples extends React.Component<Props, State> {
```
ConradJChan, line 17: Could leverage `useState` hook instead of changing it to a class component and avoid the extra Props and State type declarations

examples/src/ThumbnailCardExamples.js
```diff
@@ -1,63 +1,118 @@
...
+        );
+        const subtitle = <div>I&#39;m a subtitle!</div>;
+        const title = <div>Hello World!</div>;
+        const thumbnail = <div>Thumbnail goes here</div>;
+        const actionItem = (
+            <DropdownMenu
+                onMenuOpen={() => this.updateIsActionItemActive(true)}
+                onMenuClose={() => this.updateIsActionItemActive(false)}
+            >
+                <PlainButton className="lnk">
```
ConradJChan, line 54: Might want to specify a type="button" instead of letting it default to "submit"

src/components/thumbnail-card/ThumbnailCard.js
```diff
@@ -1,23 +1,35 @@
...
 import ThumbnailCardDetails from './ThumbnailCardDetails';
 import ThumbnailCardThumbnail from './ThumbnailCardThumbnail';
 
 import './ThumbnailCard.scss';
 
 type Props = {
+    actionItem?: React.Node,
     className?: string,
     icon?: React.Node,
+    shouldExpandOnHover?: boolean,
```
ConradJChan, line 14: Maybe just `expandOnHover`?

src/components/thumbnail-card/ThumbnailCard.scss
```diff
@@ -13,6 +13,21 @@
     box-shadow: 0 5px 10px 0 rgba(198, 198, 198, .2);
 }
 
+// We should expand the card only when it is hovered or the action item is active
+.thumbnail-card.is-expandable {
```
ConradJChan, line 17: Should this be namespaced and SUIT styled?

src/components/thumbnail-card/ThumbnailCard.scss
```diff
@@ -13,6 +13,21 @@
...
+    &.is-action-item-active,
+    &:hover {
```
ConradJChan, line 19: Should these styles apply to `focus` too for accessibility?

src/components/thumbnail-card/ThumbnailCard.js
```diff
@@ -1,23 +1,35 @@
...
+    shouldExpandOnHover = false,
+    subtitle,
+    title,
+    thumbnail,
+    ...rest
+}: Props) => (
+    <div className={classNames('thumbnail-card', className, { 'is-expandable': shouldExpandOnHover })} {...rest}>
         <ThumbnailCardThumbnail thumbnail={thumbnail} />
-        <ThumbnailCardDetails icon={icon} subtitle={subtitle} title={title} />
+        <ThumbnailCardDetails actionItem={actionItem} icon={icon} subtitle={subtitle} title={title} />
```
ConradJChan, line 32: Thoughts on following the pattern `Media` has with `MediaBody`, `MediaMenu` instead of passing React nodes down as props?

src/components/thumbnail-card/__tests__/ThumbnailCardDetails.test.js
```diff
@@ -25,4 +25,11 @@ describe('components/thumbnail-card/ThumbnailCardDetails', () => {
 
         expect(wrapper).toMatchSnapshot();
     });
+
+    test('should render actionItem if it is passed in', () => {
+        const actionItem = <div>Action Item</div>;
+        const wrapper = getWrapper({ actionItem });
+
+        expect(wrapper).toMatchSnapshot();
```
booc0mtaco, line 33: please specify more specifically what this test expects to be rendered, instead of a snapshot

src/components/thumbnail-card/ThumbnailCard.scss
```diff
@@ -13,6 +13,21 @@
...
+        box-shadow: 0 4px 24px 0 rgba(0, 0, 0, .2);
+        transform: translate3d(0, -1px, 0) scale(1.25);
+        transition-duration: .2s;
```
booc0mtaco, line 22: please use `bdl-transitionDurationBase` (either 1x or 2x)

src/components/thumbnail-card/ThumbnailCard.scss
ChenCodes: I can make it say `is-expanded` instead since I think that's how the SUIT convention works

src/components/thumbnail-card/ThumbnailCard.scss
ChenCodes: I don't think this card is focusable. It doesn't look like the original cards are focusable either

ChenCodes: Should the cards be focusable?

examples/src/ThumbnailCardExamples.js
ChenCodes: Yep! Good call

src/components/thumbnail-card/ThumbnailCard.scss
ConradJChan: I'm not sure how they're used, but I would learn towards an experience where interaction elements are accessible via keyboard and mouse.

src/components/thumbnail-card/ThumbnailCard.js
booc0mtaco: Using named sub-components allows us to compartmentalize changes, and give users some guardrails for making cleaner implementations. Also gives us an enforcement mechanism that we can `throw` against

src/components/thumbnail-card/ThumbnailCard.scss
booc0mtaco: presumably, if one can have a menu in this card, then focusing the menu should both focus the card (bring it into view), and the menu

booc0mtaco:
Should postpone until we sync with design, since there are diverging requirements around how a menu should look. 

@ChenCodes does your PR include any code that effects global styles for menus or selected states?

examples/src/ThumbnailCardExamples.js
```diff
@@ -48,7 +57,11 @@ const ThumbnailCardExamples = () => {
                     {new Array(3).fill(true).map((_, i) => (
                         <ThumbnailCard
                             key={`thumbnailcard-${i}`}
+                            className={isCardInFocus === i ? 'is-card-in-focus' : ''}
                             icon={icon}
+                            id={i}
+                            setIsCardInFocus={setIsCardInFocus}
+                            shadowOnHover
```
booc0mtaco, line 64: we should use a prop name that describes the state it is in, and not how that state is currently implemented. Think `highlightOnHover` or `showFocusState` or similar.

src/components/thumbnail-card/ThumbnailCard.js
```diff
@@ -9,13 +11,37 @@ import './ThumbnailCard.scss';
...
+    id,
+    shadowOnHover = false,
+    setIsCardInFocus = noop,
+    subtitle,
+    title,
+    thumbnail,
+    ...rest
+}: Props) => (
+    <div
+        onFocus={() => id && setIsCardInFocus(id)}
```
booc0mtaco, line 34: Can you explain why onFocus and onBlur were needed? This code seems more about making the example work.

it should be possible to implement this using `tabIndex` and `:active` on the relevant elements in the component and css respectively.

ChenCodes: Good call out! Discussed offline that we will remove this functionality - it was originally added due to the actionItem use case.

examples/src/ThumbnailCardExamples.js
ChenCodes: I think that makes sense. Let's go with `highlightOnHover`.

PR #2162, opened 2020-05-29
fix(annotations): scroll to emitted annotation

mickr:
* emit scrolltoannotation event to box-content-preview to handle scrolling. This will decouple setting the active state and scrolling.


src/elements/content-preview/ContentPreview.js
```diff
@@ -1145,15 +1145,18 @@ class ContentPreview extends React.PureComponent<Props, State> {
         });
     };
 
-    handleAnnotationSelect = ({
-        file_version: { id: annotationFileVersionId },
-        target: { location = {} },
-    }: Annotation) => {
+    handleAnnotationSelect = (annotation: Annotation) => {
+        const {
```
jstoffan, line 1149: Did this need to move out of the function signature?

src/elements/content-preview/__tests__/ContentPreview.test.js
```diff
@@ -1146,12 +1147,14 @@ describe('elements/content-preview/ContentPreview', () => {
                     },
                 };
 
+                instance.getViewer = jest.fn().mockReturnValue({ emit });
```
jstoffan, line 1150: I've been tending to prefer `jest.spyOn` lately since it leaves the underlying method in place. Should we test the case where the viewer is undefined?

src/elements/content-preview/ContentPreview.js
mickr: It seemed clearer to move it out and pull the data from `annotation` and then emit the whole annotation to `content-preview`

src/elements/content-preview/ContentPreview.js
```diff
@@ -1145,15 +1145,18 @@ class ContentPreview extends React.PureComponent<Props, State> {
...
+            file_version: { id: annotationFileVersionId },
+            target,
+        } = annotation;
+        const { location = {} } = target;
         const { selectedVersion } = this.state;
```
ConradJChan, line 1154: Hmm looks like I goofed with this destructuring from `this.state` twice, would you mind fixing that for me?

src/elements/content-preview/ContentPreview.js
```diff
@@ -1163,6 +1166,10 @@ class ContentPreview extends React.PureComponent<Props, State> {
                 },
             });
         }
+
+        if (viewer) {
+            viewer.emit('scrolltoannotation', annotation);
```
ConradJChan, line 1171: Does this fix the issue where, after creating an annotation, the Preview slightly scrolls the annotation into view?

jstoffan: We would need to remove the line in BaseAnnotator that scrolls when the active id is set, which we can do once this change is merged.

src/elements/content-preview/ContentPreview.js
mickr: No problem, surprised I didn't catch it 😱 .

src/elements/content-preview/ContentPreview.js
mickr: The overall move to decouple scrolling with the active state should fix that.

PR #2163, opened 2020-05-29
fix(annotations): Avoid path rewrite when ActivitySidebar remounts

ConradJChan:
Since Sidebar unmounts and mounts the individual Sidebar panels when collapsing and expanding, I added knowledge in the `SidebarPanels` to keep track of whether the component has loaded for the first time and pass that as a prop into the individual sidebars.

`ActivitySidebar` can use this knowledge to know whether or not to treat it as a first mount or a remount

### DIFF
--- a/src/elements/content-sidebar/DetailsSidebar.js
+++ b/src/elements/content-sidebar/DetailsSidebar.js
@@ -47,6 +47,7 @@ type ExternalProps = {
     hasNotices?: boolean,
     hasProperties?: boolean,
     hasRetentionPolicy?: boolean,
+    hasSidebarInitialized: boolean,
     hasVersions?: boolean,
     onAccessStatsClick?: Function,
     onClassificationClick?: (e: SyntheticEvent<HTMLButtonElement>) => void,
--- a/src/elements/content-sidebar/DetailsSidebar.js
+++ b/src/elements/content-sidebar/DetailsSidebar.js
@@ -47,7 +47,7 @@ type ExternalProps = {
     hasNotices?: boolean,
     hasProperties?: boolean,
     hasRetentionPolicy?: boolean,
-    hasSidebarInitialized: boolean,
+    hasSidebarInitialized?: boolean,
     hasVersions?: boolean,
     onAccessStatsClick?: Function,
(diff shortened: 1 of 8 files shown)


src/elements/content-sidebar/SidebarPanels.js
```diff
@@ -81,10 +81,16 @@ class SidebarPanels extends React.Component<Props> {
...
     detailsSidebar: ElementRefType = React.createRef();
 
+    isFirstLoad: boolean = true;
+
     metadataSidebar: ElementRefType = React.createRef();
 
     versionsSidebar: ElementRefType = React.createRef();
 
+    componentDidMount() {
+        this.isFirstLoad = false;
```
jstoffan, line 91: Seems like this should be tracked in state so changes to it cause a re-render. Also, maybe `isMounted` or `isInitialized`? That way the default prop value might not be necessary in each sidebar?

src/elements/content-sidebar/MetadataSidebar.js
```diff
@@ -74,6 +75,7 @@ class MetadataSidebar extends React.PureComponent<Props, State> {
 
     static defaultProps = {
         isFeatureEnabled: true,
+        isFirstLoad: true,
```
jstoffan, line 78: Unused?

src/elements/content-sidebar/__tests__/SidebarPanels.test.js
```diff
@@ -115,5 +115,13 @@ describe('elements/content-sidebar/SidebarPanels', () => {
                 });
             });
         });
+
+        describe('first loaded behavior', () => {
+            test('should update isFirstLoad on mount', () => {
```
jstoffan, line 120: Looks like the test description needs to be updated.

src/elements/content-sidebar/DetailsSidebar.js
```diff
@@ -47,6 +47,7 @@ type ExternalProps = {
     hasNotices?: boolean,
     hasProperties?: boolean,
     hasRetentionPolicy?: boolean,
+    hasSidebarInitialized: boolean,
```
jstoffan, line 50: Should this be optional?

PR #2164, opened 2020-05-29
fix(collaborator-avatars): use standard position for external contacts

booc0mtaco:
the contact detail view had a custom implementation of the external badge for users, which sat in the wrong corner. use the `Avatar` implementation instead, so that they all look identical.


booc0mtaco:
Before:
Screen Shot 2020-05-28 at 4 49 28 PM

After: 
Screen Shot 2020-05-29 at 4 50 27 PM

src/features/collaborator-avatars/CollaboratorAvatarItem.js
```diff
@@ -35,9 +34,15 @@ const CollaboratorAvatarItem = (props: Props) => {
 
     const avatarInstance =
         hasCustomAvatar && avatarUrl ? (
-            <Avatar avatarUrl={avatarUrl} name={name} {...rest} />
+            <Avatar
+                avatarUrl={avatarUrl}
+                name={name}
+                {...rest}
```
sahiga, line 40: Minor: I'm curious about why the `rest` props appear in the middle of the list. Are `isExternal` and `shouldShowExternal` overriding some of the `rest` props?

src/features/unified-share-modal/README.md
```diff
@@ -101,7 +101,19 @@ class USMExample extends React.Component {
                 const collaborators = this.contacts.slice();
 
                 const collaboratorsList = {
-                    collaborators,
+                    collaborators: this.contacts.map(contact => {
```
sahiga, line 104: Should the Storybook story also be updated?

src/features/collaborator-avatars/CollaboratorAvatarItem.js
booc0mtaco: oh, yeah, i suppose the convention is to put `...rest` at the end. it wasn't done intentionally; tho given the context of this component, there would be no overriding since the props used and `...rest` have the same source.

src/features/unified-share-modal/README.md
booc0mtaco: i didnt realize there was a storybook version of this. i can update

PR #2165, opened 2020-06-01
feat(visual-tooltip): Add in VisualTooltip component

laughnan:
Screen Shot 2020-05-31 at 17 09 20


src/components/visual-tooltip/VisualTooltip.tsx
```diff
@@ -0,0 +1,43 @@
+import * as React from 'react';
+
+import Tooltip, { TooltipProps, TooltipTheme } from '../tooltip';
+
+import './VisualTooltip.scss';
+
+// We manually set "text" with our specific visual tooltip content.
+type OtherTooltipProps = Omit<TooltipProps, 'text'>;
```
laughnan, line 8: seeing some issues importing `TooltipProps`:

```
WARNING in ./src/components/tooltip/index.ts 1:0-81
"export 'TooltipProps' was not found in './Tooltip'
 @ ./src/icons/icons.stories.tsx
 @ ./src sync \.stories\.(js|tsx)$
 @ ./.storybook/config.js
 @ multi ./node_modules/@storybook/core/dist/server/common/polyfills.js ./node_modules/@storybook/core/dist/server/preview/globals.js ./node_modules/@storybook/addon-docs/dist/frameworks/common/config.js ./node_modules/@storybook/addon-docs/dist/frameworks/react/config.js ./.storybook/config.js ./node_modules/webpack-hot-middleware/client.js?reload=true&quiet=true
```

I think it's an issue with webpack and the TS interfaces. Need to explore this more.

src/components/tooltip/Tooltip.tsx
```diff
@@ -73,27 +73,27 @@ export interface TooltipProps {
     /** A CSS class for the tooltip */
     className?: string;
     /** Whether to constrain the tooltip to the element's scroll parent. Defaults to `false` */
-    constrainToScrollParent: boolean;
+    constrainToScrollParent?: boolean;
```
laughnan, line 76: These props are all required, but have default values set. Thinking about making them optional (due to the defaults).

src/components/tooltip/Tooltip.tsx
```diff
@@ -66,35 +66,41 @@ const positions = {
...
     stopBubble?: boolean;
     /** Text to show in the tooltip */
     text?: React.ReactNode;
-    /** Tooltip theme */
-    theme?: TooltipTheme;
-}
+} & Partial<DefaultTooltipProps>;
+// Fixes issue where webpack thinks TooltipProps is undefined due to how Typescript types are removed
+// when compiling.
+export const TooltipProps = undefined;
```
laughnan, line 103: seeing some issues importing `TooltipProps`:

```
WARNING in ./src/components/tooltip/index.ts 1:0-81
"export 'TooltipProps' was not found in './Tooltip'
 @ ./src/icons/icons.stories.tsx
 @ ./src sync \.stories\.(js|tsx)$
 @ ./.storybook/config.js
 @ multi ./node_modules/@storybook/core/dist/server/common/polyfills.js ./node_modules/@storybook/core/dist/server/preview/globals.js ./node_modules/@storybook/addon-docs/dist/frameworks/common/config.js ./node_modules/@storybook/addon-docs/dist/frameworks/react/config.js ./.storybook/config.js ./node_modules/webpack-hot-middleware/client.js?reload=true&quiet=true
```

I think it's an issue with webpack and the TS interfaces. Need to explore this more. Adding this did resolve the issue, I think it removes `TooltipProps` as an export when webpack runs?

src/components/tooltip/Tooltip.tsx
```diff
@@ -205,7 +208,7 @@ class Tooltip extends React.Component<TooltipProps, State> {
             constrainToWindow,
             isDisabled,
             isTabbable = true,
-            position,
+            position = TooltipPosition.TOP_CENTER,
```
ConradJChan, line 211: Would this already be defaulted by the defaultProps on L109?

src/components/visual-tooltip/VisualTooltip.scss
```diff
@@ -0,0 +1,45 @@
+@import '../../styles/variables';
+
+.VisualTooltip {
```
ConradJChan, line 3: Should this be namespaced with `bdl-`?

src/components/visual-tooltip/VisualTooltip.scss
```diff
@@ -0,0 +1,45 @@
...
+    width: 320px;
+    max-width: 320px;
+    padding: 0;
+    font-size: 14px;
+    text-align: center;
+    border-radius: 6px;
+
+    & .tooltip-close-button {
```
ConradJChan, line 11: Is the `& ` here necessary?

src/components/visual-tooltip/VisualTooltip.scss
```diff
@@ -0,0 +1,45 @@
...
+        top: 16px;
+        right: 16px;
+        width: 16px;
+    }
+
+    &.with-close-button {
+        padding-right: 0;
```
ConradJChan, line 18: Is this still needed if L6 has `padding: 0` already?

src/components/visual-tooltip/VisualTooltip.scss
```diff
@@ -0,0 +1,45 @@
...
+    box-shadow: 0 3px 6px 0 rgba($black, .1);
+}
+
+.VisualTooltip--title {
+    margin-top: 0;
+    color: $bdl-gray;
+    font-weight: bold;
+}
+
+.VisualTooltip--content {
```
ConradJChan, line 43: Should these class names use single dash `-`? They seem more like descendants rather than modifiers

src/components/visual-tooltip/__tests__/VisualTooltip.test.tsx
```diff
@@ -0,0 +1,19 @@
...
+describe('components/visual-tooltip/VisualTooltip', () => {
+    test('should correctly render default component', () => {
+        const tooltipContent = 'Hey I am content';
+        const tooltipTitle = 'I am a title';
+        const wrapper: ReactWrapper = mount(
+            <VisualTooltip content={tooltipContent} imageSrc={testImageSrc} isShown title={tooltipTitle}>
+                <Button>Callout</Button>
+            </VisualTooltip>,
+        );
+        expect(wrapper).toMatchSnapshot();
```
ConradJChan, line 17: Should we be explicit about what assertions we want to make to make the test more useful?

src/components/visual-tooltip/VisualTooltip.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+export const basic = () => (
+    <div style={{ textAlign: 'center' }}>
+        <VisualTooltip
+            content="Lorem ipsum dolor sit amet, consec tetur adipiscing elit. Ut at semper nisl."
+            imageSrc={testImageSrc}
+            isShown
+            title="Lorem ipsum dolor"
+        >
+            <Button>Callout</Button>
```
ConradJChan, line 17: Maybe set the button type to be "button", I believe it defaults to "submit"

src/components/visual-tooltip/VisualTooltip.tsx
```diff
@@ -0,0 +1,48 @@
...
+    /** A string to be used in the tooltip's paragraph content */
+    content: string;
+    /** The string source of the image used in the visual tooltip */
+    imageSrc: string;
+    /** A string to be used in the tooltip's title heading */
+    title: string;
+} & OtherTooltipProps;
+
+const VisualTooltip = ({ children, className, content, imageSrc, title, ...otherTooltipProps }: VisualTooltipProps) => {
+    const VisualTooltipContent = () => (
```
ConradJChan, line 23: Would it make sense to break this into it's own file?

src/components/tooltip/Tooltip.tsx
laughnan: I think it's because we do `Partial<DefaultTooltipProps>` in the Props definition (though we do set it in the `defaultProps`. the `tetherPosition` constant complains it could be undefined. 🤔

src/components/visual-tooltip/VisualTooltip.scss
laughnan: added a comment, we do need this specific override for specificity.

src/components/tooltip/Tooltip.tsx
ConradJChan: Maybe if you change `defaultProps` type to `TooltipProps`?

src/components/visual-tooltip/VisualTooltipContent.tsx
```diff
@@ -0,0 +1,26 @@
...
+    /** The string source of the image used in the visual tooltip */
+    imageSrc: string;
+    /** A string to be used in the tooltip's title heading */
+    title: string;
+};
+
+const VisualTooltipContent = ({ content, imageSrc, title }: VisualTooltipContentProps) => (
+    <div className="bdl-VisualTooltipContent">
+        <div className="bdl-VisualTooltipContent-imageWrapper">
+            <img src={imageSrc} alt={title} className="bdl-VisualTooltipContent-image" />
```
ConradJChan, line 17: Should the `alt` for the image always necessarily be the same as `title`?

ConradJChan: Maybe instead, could you pass the image as a `React.Node` prop? Instead of the `bdl-VisualTooltipContent-image` styling for `max-width`, could you style `bdl-VisualTooltipContent-imageWrapper` as a flexbox container that stretches content to full width?

src/components/tooltip/Tooltip.tsx
laughnan: I have it set to `Partial` because none of those are "required" but maybe that isn't getting us anything different than if I put them back in `Tooltips` and made them optional?

src/components/visual-tooltip/VisualTooltipContent.tsx
laughnan: Looking into the flexbox recommendation now. I suppose I can just change it to a `React.Node` and then let the parent pass in whatever they want.

src/components/visual-tooltip/VisualTooltip.tsx
```diff
@@ -0,0 +1,37 @@
...
+    children: React.ReactChild;
+    /** A string to be used in the tooltip's paragraph content */
+    content: string;
+    /** A React component representing the image used in the visual tooltip */
+    image: React.ReactNode;
+    /** A string to be used in the tooltip's title heading */
+    title: string;
+} & OtherTooltipProps;
+
+const VisualTooltip = ({ children, className, content, image, title, ...otherTooltipProps }: VisualTooltipProps) => {
```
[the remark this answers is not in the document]
laughnan: Great point! I was thinking `Visual` as like a visual graphic (not `visual` like sight). Perhaps something like `ImageTooltip` or `AssetTooltip` would be clearer?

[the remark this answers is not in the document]
azuruce: It seems you are building TitleImageContentTooltip if you want to be specific. On the first sight, AssetTooltip and ImageTooltip are partially accurate but still much better than visualTooltip. TitleImageContentTooltip is better?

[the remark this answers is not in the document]
jellee: I like ImageTooltip since your comments mention an image used in it.  Are there non-image assets that we would want to use?  

The only asset that comes to mind that may not be categorized as an image is an icon..  but we may want a different component for icons since they tend to be smaller and would require different styling?

src/components/visual-tooltip/VisualTooltip.scss
```diff
@@ -0,0 +1,21 @@
+@import '../../styles/variables';
+
+.bdl-VisualTooltip {
+    width: 320px;
+    max-width: 320px;
```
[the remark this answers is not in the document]
laughnan: Great question! The overall `Tooltip` component (which we're using) has a `max-width` property, so we're overriding it here 😄

src/components/visual-tooltip/VisualTooltip.tsx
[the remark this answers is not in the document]
laughnan: Great feedback, you two! Thank you! I'm going to adjust to `ImageTooltip` (more succinct and still as clear IMHO).

src/components/visual-tooltip/VisualTooltip.scss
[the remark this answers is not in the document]
jellee: Ahh thanks for explaining!

src/components/image-tooltip/ImageTooltipContent.scss
```diff
@@ -0,0 +1,29 @@
+@import '../../styles/variables';
+
+.bdl-ImageTooltipContent-image {
+    display: flex;
+
+    img {
```
booc0mtaco, line 6: lets avoid using tag selectors, since itll match on every image on the page during parsing

src/components/image-tooltip/ImageTooltipContent.scss
```diff
@@ -0,0 +1,29 @@
...
+        width: 100%;
+    }
+}
+
+.bdl-ImageTooltipContent-contentWrapper {
+    padding: 24px 32px;
+    color: $bdl-gray-80;
+    font-size: 14px;
+    background-color: $white;
+    border-bottom-right-radius: 6px;
```
booc0mtaco, line 16: we have instances of this 6px in a few places. i suggest creating a variable for tooltip radii, so the stay in sync

src/components/image-tooltip/ImageTooltip.scss
```diff
@@ -0,0 +1,21 @@
+@import '../../styles/variables';
+
+.bdl-ImageTooltip {
+    width: 320px;
+    max-width: 320px;
+    padding: 0;
+    font-size: 14px;
+    text-align: center;
+    border-radius: 6px;
```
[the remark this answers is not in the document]
laughnan: let me check with design on this question, I agree it's inconsistent.

src/components/image-tooltip/ImageTooltip.tsx
```diff
@@ -0,0 +1,37 @@
...
+import ImageTooltipContent from './ImageTooltipContent';
+
+import './ImageTooltip.scss';
+
+// We manually set "text" with our specific visual tooltip content.
+type OtherTooltipProps = Omit<TooltipProps, 'text'>;
+
+export type ImageTooltipProps = {
+    /** A React element to put the tooltip on */
+    children: React.ReactChild;
```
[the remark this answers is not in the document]
laughnan: Currently, yes and it mirrors the Tooltip's supported implementation.

src/components/image-tooltip/ImageTooltipContent.scss
laughnan: Reaching out for design for confirmation first and then if he wants 6 pixels I'll create a variable. Thanks!

src/components/image-tooltip/ImageTooltipContent.tsx
```diff
@@ -1,19 +1,37 @@
...
     title: string;
 };
 
-const ImageTooltipContent = ({ content, image, title }: ImageTooltipContentProps) => (
+function appendTooltipImageChildClass(child: React.ReactElement): React.ReactElement | null {
+    if (child) {
+        const {
+            props: { className: existingClasses },
+        } = child;
+        const className = classNames(existingClasses, 'bdl-ImageTooltipContent-imageChild');
```
booc0mtaco, line 20: nit: you might inline this in your return statement to be a bit more terse

PR #2168, opened 2020-06-01
feat(icons): IconGlobeInverted

wenboyu2:
image

IconGlobeInverted is going to be used in the login page header
image


booc0mtaco:
Prefer to use https://github.com/box/box-ui-elements/blob/master/src/icon/fill/GlobeBadge16.tsx if possible (then to set the one override color,  set a backdrop color, etc. via CSS)

wenboyu2:
Going to use GlobeBadge16.tsx instead

PR #2169, opened 2020-06-01
fix(annotations): Move logic to withSidebarAnnotations hoc

ConradJChan:
Lifting the add annotation logic from `ActivitySidebar` to `Sidebar` allows for a smoother UX when adding annotations with the sidebar collapsed.

TODO
 - [x] unit tests


src/elements/content-sidebar/Sidebar.js
```diff
@@ -122,6 +128,44 @@ class Sidebar extends React.Component<Props, State> {
...
+        const isOpen = this.getIsOpen();
+        const isPending = action === 'create_start';
+        const isAnnotationsPath = getAnnotationsMatchPath(history);
+
+        if (!currentUser) {
+            throw getBadUserError();
+        }
+
+        const feedAPI = api.getFeedAPI(false);
+        const { id } = file;
```
jstoffan, line 155: Should we use the `fileId` prop, instead?

src/elements/content-sidebar/Sidebar.js
```diff
@@ -114,6 +116,10 @@ class Sidebar extends React.Component<Props, State> {
             this.setState({ isDirty: true });
         }
 
+        if (prevAnnotation !== annotation) {
```
jstoffan, line 119: Should we check that `annotation` is defined, as well?

src/elements/content-sidebar/Sidebar.js
```diff
@@ -122,6 +128,44 @@ class Sidebar extends React.Component<Props, State> {
...
+        const {
+            annotatorState: { action, annotation, meta },
+            api,
+            currentUser,
+            file,
+            getAnnotationsMatchPath,
+            history,
+        } = this.props;
+        const { requestId } = meta || {};
+        const isOpen = this.getIsOpen();
```
jstoffan, line 146: Should all of this code live in SidebarPanels, instead, where `isOpen` is a simple prop? Or maybe even as its own HOC around SidebarPanels?

src/elements/content-sidebar/Sidebar.js
```diff
@@ -122,6 +128,44 @@ class Sidebar extends React.Component<Props, State> {
         }
     }
 
+    getIsOpen = () => {
+        const { isDefaultOpen } = this.props;
+        return this.isForcedSet() ? this.isForcedOpen() : !!isDefaultOpen;
+    };
+
+    addAnnotation() {
```
jstoffan, line 136: It seems kind of unfortunate that the orchestration component is becoming more aware of the internals of its children. Do we have an end state for what this should look like?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -156,43 +156,16 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+        const { match } = this.props;
         const prevFileVersionId = getProp(prevMatch, 'params.fileVersionId');
         const fileVersionId = getProp(match, 'params.fileVersionId');
 
-        if (prevAnnotation !== annotation) {
-            this.addAnnotation();
-        }
-
         if (prevFileVersionId !== fileVersionId) {
             this.updateActiveVersion();
```
jstoffan, line 165: Will this need to move up as well or no?

ConradJChan: Yeah, all his friends have moved up... but I think this should actually stay here, at least for now. `feedItems` is where we have the full version objects and moving it up would require the fetching of versions to be done higher up. Maybe we can revisit this as part of the future refactor of having ContentPreview load a previous versions directly

src/elements/content-sidebar/Sidebar.js
ConradJChan: Good point

src/elements/content-sidebar/Sidebar.js
ConradJChan: Yeah agreed. I think ideally we would have something like redux to manage the API state and have some HOC consuming the AnnotatorContext and updating the state data. I'll work on moving this to a HOC as a step in that direction, something like `WithSidebarAnnotations`

src/elements/content-sidebar/Sidebar.js
```diff
@@ -122,6 +128,44 @@ class Sidebar extends React.Component<Props, State> {
...
+            getAnnotationsMatchPath,
+            history,
+        } = this.props;
+        const { requestId } = meta || {};
+        const isOpen = this.getIsOpen();
+        const isPending = action === 'create_start';
+        const isAnnotationsPath = getAnnotationsMatchPath(history);
+
+        if (!currentUser) {
+            throw getBadUserError();
```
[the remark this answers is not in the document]
ConradJChan: Seems like `currentUser` could be undefined. If an error throws, it would result in the sad cloud for the `ContentSidebar` component I believe

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -156,43 +156,16 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         this.fetchCurrentUser(currentUser);
```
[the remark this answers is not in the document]
ConradJChan: Yeah looks like it's mainly been a pass through prop until ActivitySidebar made use of it. Worth looking into whether we need to lift it up in a follow up change

src/elements/content-sidebar/withSidebarAnnotations.js
```diff
@@ -0,0 +1,179 @@
...
+            }
+
+            // Active annotation id changed. If location is currently an annotation path or
+            // if location is not currently an annotation path but the active annotation id
+            // transitioned from falsy to truthy, update the location accordingly
+            if (hasActiveAnnotationChanged && (isAnnotationsPath || isTransitioningToAnnotationPath)) {
+                this.updateActiveAnnotation();
+            }
+
+            if (prevFileVersionId !== fileVersionId) {
```
jstoffan, line 79: Does this only handle annotation-related actions or does it also handle VersionSidebar?

src/elements/content-sidebar/withSidebarAnnotations.js
```diff
@@ -0,0 +1,179 @@
...
+
+            const feedAPI = api.getFeedAPI(false);
+            const { items: hasItems } = feedAPI.getCachedItems(fileId) || {};
+
+            // If there are existing items in the cache for this file, then patch the cache with the new annotation
+            if (hasItems) {
+                feedAPI.addAnnotation(file, currentUser, annotation, requestId, isPending);
+            }
+
+            if (isAnnotationsPath && isOpen && this.sidebarPanels.current) {
```
jstoffan, line 112: We should destructure `current` out above.

src/elements/content-sidebar/withSidebarAnnotations.js
```diff
@@ -0,0 +1,179 @@
...
+    fileId: string,
+    isOpen: boolean,
+    onVersionChange?: Function,
+} & WithAnnotatorContextProps &
+    ContextRouter;
+
+export default function withSidebarAnnotations(
+    WrappedComponent: React.ComponentType<Props>,
+): React.ComponentType<Props> {
+    class WithSidebarAnnotations extends React.Component<Props> {
```
jstoffan, line 22: Does this need to be a class? If it was a function component, we could remove `withAnnotatorContext` from the parent and use the `useContext` hook, instead. That may be too large of a rework for this PR, though.

src/elements/content-sidebar/withSidebarAnnotations.js
```diff
@@ -0,0 +1,179 @@
+// @flow
```
jstoffan, line 1: Should this be a TS file?

src/elements/content-sidebar/withSidebarAnnotations.js
```diff
@@ -0,0 +1,179 @@
...
+                annotatorState: { action, annotation, meta },
+                api,
+                currentUser,
+                file,
+                fileId,
+                getAnnotationsMatchPath,
+                isOpen,
+                location,
+            } = this.props;
+            const { requestId } = meta || {};
```
jstoffan, line 95: If `meta` and `requestId` are undefined, should we abort? We use the latter as the object id below.

src/elements/content-sidebar/withSidebarAnnotations.js
```diff
@@ -0,0 +1,179 @@
...
+            // TODO: need to address in follow on -- currentUser may be undefined here but is never fetched for sure until ActivitySidebar
+            if (!currentUser) {
+                throw getBadUserError();
+            }
+
+            const feedAPI = api.getFeedAPI(false);
+            const { items: hasItems } = feedAPI.getCachedItems(fileId) || {};
+
+            // If there are existing items in the cache for this file, then patch the cache with the new annotation
+            if (hasItems) {
```
jstoffan, line 108: Is this check needed? It looks like we do a check within `updateFeedItem`.

src/elements/content-sidebar/withSidebarAnnotations.js
```diff
@@ -0,0 +1,179 @@
+// @flow
+import * as React from 'react';
+import getProp from 'lodash/get';
+import noop from 'lodash/noop';
+import type { ContextRouter } from 'react-router-dom';
```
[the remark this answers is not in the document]
ConradJChan: flow-typed [definition](https://github.com/flow-typed/flow-typed/blob/c4f47bdda4b38cbe1a9c1a77957a4d0f7424bb3b/definitions/npm/react-router-dom_v5.x.x/flow_v0.104.x-/react-router-dom_v5.x.x.js#L83)

src/elements/content-sidebar/withSidebarAnnotations.js
```diff
@@ -0,0 +1,179 @@
...
+import { getBadUserError } from '../../utils/error';
```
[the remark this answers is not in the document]
ConradJChan: Yeah, but I think there's a conflict with the eslint rules

src/elements/content-sidebar/withSidebarAnnotations.js
```diff
@@ -0,0 +1,179 @@
...
+            const { annotatorState, getAnnotationsMatchPath, location }: Props = this.props;
+            const { annotatorState: prevAnnotatorState, location: prevLocation }: Props = prevProps;
+            const { activeAnnotationId, annotation } = annotatorState;
+            const { activeAnnotationId: prevActiveAnnotationId, annotation: prevAnnotation } = prevAnnotatorState;
+
+            const match = getAnnotationsMatchPath(location);
+            const prevMatch = getAnnotationsMatchPath(prevLocation);
+            const fileVersionId = getProp(match, 'params.fileVersionId');
+            const hasActiveAnnotationChanged = prevActiveAnnotationId !== activeAnnotationId;
+            const isAnnotationsPath = !!match;
```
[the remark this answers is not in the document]
ConradJChan: I'll force the other usages to be a boolean as well to be consistent

src/elements/content-sidebar/withSidebarAnnotations.js
ConradJChan: It should be annotation only since it returns matching results only if it matches the annotation path

src/elements/content-sidebar/withSidebarAnnotations.js
ConradJChan: I'll add to the comment in the code, but yeah it's needed because without the check, a fresh load of the `ActivitySidebar` will check the cache and see that there is already an entry for the feed items, in which case it will render that single annotation before refreshing with all the activity items

src/elements/content-sidebar/withSidebarAnnotations.js
ConradJChan: I'll look into converting it to a functional component in a follow up PR

src/elements/content-sidebar/withSidebarAnnotations.js
ConradJChan: I gave it a shot, but it looks like there is a lot of work to get the current flow types converted over to TS, can take a look in a follow up PR as well

src/elements/content-sidebar/withSidebarAnnotations.js
```diff
@@ -0,0 +1,179 @@
...
+                getAnnotationsMatchPath,
+                getAnnotationsPath,
+                history,
+                location,
+            } = this.props;
+            const match = getAnnotationsMatchPath(location);
+            const currentFileVersionId = getProp(file, 'file_version.id');
+            const fileVersionId = getProp(match, 'params.fileVersionId', currentFileVersionId);
+            const newLocationState = activeAnnotationId
+                ? { ...location.state, open: true, silent: false }
```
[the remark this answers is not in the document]
ConradJChan: I initially didn't have `silent` included but in the move to a HOC, it appears it need to be explicitly set to properly update the location in all cases. I think including the prior state just future proofs it

src/elements/content-sidebar/withSidebarAnnotations.js
```diff
@@ -0,0 +1,190 @@
...
+                : location.state;
+
+            // Update the location pathname and open state if transitioning to an active annotation id, force the sidebar open
+            history.push({
+                ...location,
```
jstoffan, line 147: What are we pulling forward here? If it's the query parameter, we could just do it forward it explicitly instead.

src/elements/content-sidebar/withSidebarAnnotations.js
```diff
@@ -0,0 +1,190 @@
...
+            const feedAPI = api.getFeedAPI(false);
+            const match = getAnnotationsMatchPath(location);
+            const currentFileVersionId = getProp(file, 'file_version.id');
+            const fileVersionId = getProp(match, 'params.fileVersionId');
+            const { items: feedItems = [] } = feedAPI.getCachedItems(fileId) || {};
+            const version = feedItems
+                .filter(item => item.type === 'file_version')
+                .find(item => item.id === fileVersionId);
+
+            if (version) {
```
jstoffan, line 173: Should we do this check before we derive all the other stuff above?

src/elements/content-sidebar/__tests__/withSidebarAnnotations.test.js
```diff
@@ -0,0 +1,267 @@
...
+            ${'234'}           | ${'456'}      | ${{ pathname: '/' }}                        | ${'/activity/annotations/456/234'} | ${{ open: true, silent: false }}
+            ${'234'}           | ${undefined}  | ${{ pathname: '/' }}                        | ${'/activity/annotations/123/234'} | ${{ open: true, silent: false }}
+            ${null}            | ${'456'}      | ${{ pathname: '/' }}                        | ${'/activity/annotations/456'}     | ${undefined}
+            ${null}            | ${'456'}      | ${{ pathname: '/', state: { foo: 'bar' } }} | ${'/activity/annotations/456'}     | ${{ foo: 'bar' }}
+            ${'234'}           | ${'456'}      | ${{ pathname: '/', state: { foo: 'bar' } }} | ${'/activity/annotations/456/234'} | ${{ foo: 'bar', open: true, silent: false }}
+        `(
+            'should set location path based on match param fileVersionId=$fileVersionId and activeAnnotationId=$activeAnnotationId',
+            ({ activeAnnotationId, fileVersionId, location, expectedPath, expectedState }) => {
+                const annotatorState = {
```
jstoffan, line 223: Should this variable live near the others in the test? It doesn't seem to be used until later.

src/elements/content-sidebar/__tests__/withSidebarAnnotations.test.js
```diff
@@ -0,0 +1,267 @@
...
+            ${'annotation deselected on annotations path'}     | ${'123'}               | ${null}            | ${true}           | ${1}
+            ${'annotation deselected not on annotations path'} | ${'123'}               | ${null}            | ${false}          | ${0}
+            ${'annotation selected not on annotations path'}   | ${null}                | ${'123'}           | ${false}          | ${1}
+            ${'annotation selected on annotations path'}       | ${null}                | ${'123'}           | ${true}           | ${1}
+        `(
+            'should call updateActiveAnnotation $expectedCount times if $condition',
+            ({ prevActiveAnnotationId, activeAnnotationId, isAnnotationsPath, expectedCount }) => {
+                const wrapper = getWrapper({ annotatorState: { activeAnnotationId: prevActiveAnnotationId } });
+
+                wrapper.instance().updateActiveAnnotation = jest.fn();
```
jstoffan, line 82: I've been favoring `jest.spyOn` for instance methods to keep from overwriting the property.

src/elements/content-sidebar/__tests__/withSidebarAnnotations.test.js
```diff
@@ -0,0 +1,267 @@
...
+                const wrapper = getWrapper({ file, getAnnotationsMatchPath, getAnnotationsPath, history });
+                const instance = wrapper.instance();
+                getAnnotationsMatchPath.mockReturnValue({ params: { annotationId, fileVersionId } });
+                getAnnotationsPath.mockReturnValue(expectedPath);
+
+                instance.redirectDeeplinkedAnnotation();
+
+                expect(history.replace).toHaveBeenCalledTimes(expectedCallCount);
+
+                if (expectedCallCount) {
```
jstoffan, line 137: Should this be a separate test?

src/elements/content-sidebar/__tests__/withSidebarAnnotations.test.js
```diff
@@ -0,0 +1,267 @@
...
+
+                annotatorContextProps.getAnnotationsMatchPath.mockReturnValueOnce(isAnnotationsPath);
+                feedAPI.getCachedItems.mockReturnValueOnce({ items: hasItems });
+                instance.sidebarPanels = {
+                    current: {
+                        refresh: jest.fn(),
+                    },
+                };
+
+                instance.addAnnotation();
```
jstoffan, line 195: If we want to convert this to a function component later, we should consider how we'd test it if we didn't have direct access to the inner instance methods. Is there a way we can test the inputs/outputs while treating the logic like a black box?

src/elements/content-sidebar/withSidebarAnnotations.js
ConradJChan: Not sure how much we save -- maybe just `const currentFileVersionId = getProp(file, 'file_version.id');`? In order to get the version, we still need to get `match`, and then use that to determine `fileVersionId`, etc...

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -84,20 +85,20 @@ export default function withAnnotations<P extends object>(
...
-                annotationId,
+                annotationId: annotationId === null ? undefined : annotationId,
                 fileVersionId,
             });
         }
 
-        getMatchPath(history?: History): matchType<MatchParams> | null {
-            const pathname = getProp(history, 'location.pathname');
+        getMatchPath(location?: Location): matchType<MatchParams> | null {
+            const pathname = getProp(location, 'pathname', '');
```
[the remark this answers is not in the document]
ConradJChan: yeah, `matchPath` type requires a string fo the pathname

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -680,7 +609,7 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
             onAnnotationSelect,
         } = this.props;
         const currentFileVersionId = getProp(file, 'file_version.id');
-        const match = getAnnotationsMatchPath(history);
+        const match = getAnnotationsMatchPath(history.location);
```
jstoffan, line 612: Is `history.location` and `location` always the same thing? It looks like we use the former here, but the latter elsewhere. Should we standardize?

ConradJChan: Should be same, I'll change it to be consistent

PR #2170, opened 2020-06-02
fix(avatar): strip any bracketed text when forming avatar initials

booc0mtaco:


src/components/avatar/__tests__/AvatarInitials.test.tsx
```diff
@@ -9,9 +9,20 @@ describe('components/avatar/AvatarInitials', () => {
...
-        expect(wrapper.text()).toEqual('HW');
+    test.each`
+        name                               | initials
+        ${'hello'}                         | ${'HH'}
+        ${'hello world'}                   | ${'HW'}
+        ${'hello world (personal)'}        | ${'HW'}
+        ${'hello world (personal acct)'}   | ${'HW'}
+        ${'hello world  (personal acct)'}  | ${'HW'}
+        ${' hello world (personal acct)'}  | ${'HW'}
+        ${'hello  world  (personal acct)'} | ${'HW'}
```
sahiga, line 20: Should we also test names with other types of brackets?

booc0mtaco: added a few like that, and ability to handle 1+ pairs good call

PR #2171, opened 2020-06-03
feat(pill-selector-dropdown): add new avatar pill for pill selector

danubevictoria:
Storybook updates for rounded pills (non-avatar pills) and imageless avatar pills (with and without external badge). These pills will be used in USM.

Rounded Pills (Non-Avatar Pills) - if `showRoundedPills` is true and `showAvatars` is false
image

Avatar Pills - if `showRoundedPills` and `showAvatar` is true
image

Additional pill states appear in order: 

1. error - if prop `isValid` is false. error trumps warning.
2. warning - if prop `hasWarning` is true
3. disabled - if prop `isDisabled` is true. disabled trumps all.
4. selected - if prop `isSelected` is true and `isDisabled` is false. selected trumps errors and warnings.

Rounded Pills (Non-Avatar Pills) Additional States in Pill Selector Dropdown
image

Avatar Pills Additional States in Pill Selector Dropdown
image

### DIFF
--- a/src/components/pill-selector-dropdown/PillSelector.js
+++ b/src/components/pill-selector-dropdown/PillSelector.js
@@ -2,13 +2,15 @@
 import * as React from 'react';
 import classNames from 'classnames';
 import uniqueId from 'lodash/uniqueId';
+import { List } from 'immutable';
 
 import Tooltip from '../tooltip';
 import { KEYS } from '../../constants';
 
+import RoundPill from './RoundPill';
 import Pill from './Pill';
 import SuggestedPillsRow from './SuggestedPillsRow';
-import type { Option, OptionValue, SelectedOptions, SuggestedPillsFilter } from './flowTypes';
+import type { RoundOption, Option, OptionValue, SuggestedPillsFilter } from './flowTypes';
 
 function stopDefaultEvent(event) {
     event.preventDefault();
@@ -25,7 +27,11 @@ type Props = {
     onRemove: Function,
     onSuggestedPillAdd?: Function,
     placeholder: string,
-    selectedOptions: SelectedOptions,
+    selectedOptions: Array<Object> | List<Object>,
+    /** Whether to show avatars in pills (if rounded style is enabled) */
+    showAvatars?: boolean,
+    /** Whether to use rounded style for pills */
+    showRoundedPills?: boolean,
     suggestedPillsData?: Array<Object>,
     suggestedPillsFilter?: SuggestedPillsFilter,
     suggestedPillsTitle?: string,
@@ -168,6 +174,8 @@ class PillSelector extends React.Component<Props, State> {
             onSuggestedPillAdd,
             placeholder,
             selectedOptions,
+            showAvatars,
+            showRoundedPills,
             suggestedPillsData,
             suggestedPillsFilter,
             suggestedPillsTitle,
@@ -199,17 +207,39 @@ class PillSelector extends React.Component<Props, State> {
                     onFocus={this.handleFocus}
                     onKeyDown={this.handleKeyDown}
                 >
-                    {selectedOptions.map((option: Option, index: number) => (
-                        <Pill
-                            isValid={allowInvalidPills ? validator(option) : true}
-                            isDisabled={disabled}
-                            isSelected={index === selectedIndex}
-                            key={option.value}
-                            onRemove={onRemove.bind(this, option, index)}
-                            // $FlowFixMe option.text is for backwards compatibility
-                            text={option.displayText || option.text}
-                        />
-                    ))}
+                    {showRoundedPills
+                        ? selectedOptions.map((option: RoundOption, index: number) => {
+                              return (
+                                  <RoundPill
+                                      isValid={allowInvalidPills ? validator(option) : true}
+                                      isDisabled={disabled}
+                                      isSelected={index === selectedIndex}
+                                      key={option.value}
+                                      onRemove={onRemove.bind(this, option, index)}
+                                      // $FlowFixMe option.text is for backwards compatibility
+                                      text={option.displayText || option.text}
+                                      showAvatar
+                                      id={option.id}
+                                      hasWarning={option.hasWarning}
+                                      isExternal={option.isExternalUser}
+                                  />
+                              );
+                          })
+                        : selectedOptions.map((option: Option, index: number) => {
+                              // TODO: This and associated types will be removed once all views are updates with round pills.
+                              return (
+                                  <Pill
+                                      isValid={allowInvalidPills ? validator(option) : true}
+                                      isDisabled={disabled}
+                                      isSelected={index === selectedIndex}
+                                      key={option.value}
+                                      onRemove={onRemove.bind(this, option, index)}
+                                      // $FlowFixMe option.text is for backwards compatibility
+                                      text={option.displayText || option.text}
+                                  />
+                              );
+                          })}
+
                     {/* hidden element for focus/key events during pill selection */}
                     <span
                         aria-hidden="true"
@@ -224,7 +254,9 @@ class PillSelector extends React.Component<Props, State> {
                         {...rest}
                         {...inputProps}
                         autoComplete="off"
-                        className={classNames('bdl-PillSelector-input', 'pill-selector-input', className)}
+                        className={classNames('bdl-PillSelector-input', 'pill-selector-input', className, {
+                            'bdl-PillSelector-input--showAvatars': showAvatars,
+                        })}
                         disabled={disabled}
                         onInput={onInput}
                         placeholder={this.getNumSelected() === 0 ? placeholder : ''}
--- a/src/components/pill-selector-dropdown/PillSelectorDropdown.stories.js
+++ b/src/components/pill-selector-dropdown/PillSelectorDropdown.stories.js
@@ -2,13 +2,13 @@
 import * as React from 'react';
 import { State, Store } from '@sambego/storybook-state';
 
-import DatalistItem from '../datalist-item/DatalistItem';
+import ContactDatalistItem from '../contact-datalist-item';
 import PillSelectorDropdown from './PillSelectorDropdown';
 import notes from './PillSelectorDropdown.notes.md';
 
 const users = [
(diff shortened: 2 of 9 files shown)


booc0mtaco:
We should set the border of the avatar to match the pill background color

src/components/pill-selector-dropdown/AvatarPill.js
```diff
@@ -0,0 +1,75 @@
...
+import LabelPill from '../label-pill';
+import Avatar from '../avatar';
+
+import './AvatarPill.scss';
+
+type Props = {
+    avatarUrl?: string,
+    className?: string,
+    hasWarning?: boolean,
+    id?: string | number,
```
booc0mtaco, line 16: we should consolidate around `string | number | null`

src/components/pill-selector-dropdown/AvatarPill.js
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    if (!isValid) {
+        pillType = 'error';
+    }
+
+    const handleClickRemove = isDisabled ? noop : onRemove;
+
+    const avatar = showAvatar ? (
+        <Avatar id={id} isExternal={isExternal} name={text} size="small" avatarUrl={avatarUrl} shouldShowExternal />
```
booc0mtaco, line 61: What happens if you use `LabelPill.Icon` here? ideally, whatever we put inside `LabelPill.Pill` will be recognized sub-components of label pill only.

src/components/pill-selector-dropdown/AvatarPill.js
```diff
@@ -0,0 +1,75 @@
...
+    avatarUrl?: string,
+    className?: string,
+    hasWarning?: boolean,
+    id?: string | number,
+    isDisabled?: boolean,
+    isExternal?: boolean,
+    isSelected?: boolean,
+    isValid?: boolean,
+    onRemove: () => any,
+    showAvatar?: boolean,
```
booc0mtaco, line 22: We are merging two concept with this one component: the new pill design, and an avatar pill which is a variation of that new pill. We should name the pill such that non-contact forms dont require this misleading name.

- BdlPill?
- RoundPill?
- etc.

We will eventually delete Pill.js, replacing it fully with the non-avatar pill component.

src/components/pill-selector-dropdown/PillSelectorDropdown.stories.js
```diff
@@ -151,6 +151,101 @@ export const withPills = () => {
...
+                    onSelect={handleSelect}
+                    selectedOptions={state.selectedOptions}
+                    selectorOptions={state.selectorOptions}
+                    showRoundedPills
+                    showAvatars
+                    validateForError={validateForError}
+                    validator={validator}
+                >
+                    {state.selectorOptions.map(option => (
+                        <DatalistItem key={option.value}>{option.displayText}</DatalistItem>
```
booc0mtaco, line 241: `ContactDatalistItem`

src/components/pill-selector-dropdown/__tests__/AvatarPill.test.js
```diff
@@ -0,0 +1,96 @@
...
+
+        expect(wrapper.find('LabelPill').prop('type')).toBe('error');
+    });
+
+    test('should generate LabelPill with warning type when hasWarning prop is true', () => {
+        const wrapper = shallow(<AvatarPill hasWarning onRemove={onRemoveStub} text="box" />);
+
+        expect(wrapper.find('LabelPill').prop('type')).toBe('warning');
+    });
+
```
booc0mtaco, line 76: Test for when both is warning and is valid are used at the same time

src/components/select-field/props.js
```diff
@@ -3,6 +3,9 @@
 export type SelectOptionValueProp = string | number | null;
 
 export type SelectOptionProp = {
+    avatarUrl?: string,
     displayText: string,
+    hasWarning?: boolean,
+    id?: string | number,
```
booc0mtaco, line 9: `string | number | null`

src/components/select-field/props.js
```diff
@@ -3,6 +3,9 @@
 export type SelectOptionValueProp = string | number | null;
 
 export type SelectOptionProp = {
+    avatarUrl?: string,
```
booc0mtaco, line 6: this is a little too specific to be used for all select options. which component is requiring this change?

src/components/pill-selector-dropdown/AvatarPill.js
danubevictoria: I'll do Round Pill.

src/components/select-field/props.js
danubevictoria: this is being used in the avatar

booc0mtaco: I understand that, but an avatar URL is not part of the generic selector option when it comes to this field. and it doesnt easily map to any equivalent in a standard `<select>`. we should handle this in a way so that we do not have a sprawling list of one-off types for any current and future select options

src/components/pill-selector-dropdown/AvatarPill.js
danubevictoria: good callout, it does work

src/components/select-field/props.js
danubevictoria: ok i'm goign to take this out for now because we are not doing images yet anyway

src/components/pill-selector-dropdown/AvatarPill.js
danubevictoria: adding null is an incompatible type so keeping it as is

src/components/select-field/props.js
danubevictoria: i couldn't do null as a test said it was an incompatible type with rendering the PillComponent

src/components/select-field/props.js
```diff
@@ -4,5 +4,8 @@ export type SelectOptionValueProp = string | number | null;
 
 export type SelectOptionProp = {
     displayText: string,
+    hasWarning?: boolean,
```
booc0mtaco, line 7: Let's make the same change you did for the avatar URL for these other new options here. `id` is OK b/c a `<select>` and all elements can have an ID. We want to separate out `hasWarning`, `isExternalUser`. Where did `avatarUrl` move to?

src/components/pill-selector-dropdown/RoundPill.scss
```diff
@@ -0,0 +1,79 @@
+/**************************************
+ * Pills
```
booc0mtaco, line 2: Name comment needs to change, ya?

src/components/pill-selector-dropdown/PillSelector.js
```diff
@@ -189,6 +198,8 @@ class PillSelector extends React.Component<Props, State> {
             'aria-errormessage': this.errorMessageID,
         };
 
+        const PillComponent = showRoundedPills ? RoundPill : Pill;
```
ConradJChan, line 201: It seems like `RoundPill` could be a prop on a `Pill` component, do they need to be separate components?

src/components/pill-selector-dropdown/PillSelector.js
```diff
@@ -199,17 +210,39 @@ class PillSelector extends React.Component<Props, State> {
...
+                                      isValid={allowInvalidPills ? validator(option) : true}
+                                      isDisabled={disabled}
+                                      isSelected={index === selectedIndex}
+                                      key={option.value}
+                                      onRemove={onRemove.bind(this, option, index)}
+                                      // $FlowFixMe option.text is for backwards compatibility
+                                      text={option.displayText || option.text}
+                                  />
+                              );
+                          })}
```
ConradJChan, line 244: Instead, could we spread an object of properties if `showAvatar` is true? Something like:
```
const avatarProps = {
  id: option.id,
  hasWarning: option.hasWarning,
  isExternal: option.isExternalUser,
};

const pill = <PillComponent ... />;

return showAvatars ? React.cloneElement(pill, avatarProps) : pill;
```

src/components/pill-selector-dropdown/PillSelector.js
```diff
@@ -45,6 +51,7 @@ class PillSelector extends React.Component<Props, State> {
         inputProps: {},
         placeholder: '',
         selectedOptions: [],
+        showRoundedPills: false,
```
jstoffan, line 54: Is this needed or does the falsy `undefined` value suffice? That way unaffected component snapshots don't need to be updated.

src/components/pill-selector-dropdown/RoundPill.js
```diff
@@ -0,0 +1,82 @@
...
+        <LabelPill.Icon
+            Component={Avatar}
+            id={id}
+            isExternal={isExternal}
+            name={text}
+            size="small"
+            shouldShowExternal
+        />
+    ) : (
+        undefined
```
ConradJChan, line 70: Should this be `null`?

src/components/pill-selector-dropdown/PillSelector.js
ConradJChan: or conditionally construct the props before spreading them on the component?

danubevictoria: Yea we couldn't group it together because of the different flow type options for when its a RoundedPill vs a Pill. wanted to type the content used in each .map and if you put the map outside, all paths have to use a type without conflicts. suggestions?

booc0mtaco: The desired logic is such that if `showRoundPills` is false, we should always use `Pill. if `showRoundPills` is true, we would or would not use an avatar based on `showAvatars` being passed to `RoundPill`. Given that, we could remove the line for abstracting the pill component name, `.map` based on which pill we want to show, and add `showAvatar={showAvatars}` on ln223 or the equivalent, no?

booc0mtaco:
FYI force-pushing is probably why you keep losing the merged commit. You can tidy this up by `git rebase -i` and interactively squashing the previous commits down, including that merge commit. then, any subsequent pushes (which shouldnt require `--force`) wont lose that change

PR #2172, opened 2020-06-03
feat(annotations): Handle API fetch error

mxiao6:
- [x] Tests

Screen Shot 2020-06-03 at 11 08 08 PM

### DIFF
--- a/src/api/Feed.js
+++ b/src/api/Feed.js
@@ -71,7 +71,7 @@ const TASK_NEW_INITIAL_STATUS = TASK_NEW_NOT_STARTED;
 const TASK = 'task';
 
 type FeedItemsCache = {
-    hasError: boolean,
+    errors: ErrorResponseData[],
     items: FeedItems,
 };
 
@@ -119,14 +119,15 @@ class Feed extends Base {
     file: BoxItem;
 
     /**
-     * @property {boolean}
+     * @property {ElementsXhrError}
      */
-    hasError: boolean;
+    errors: ElementsXhrError[];
 
     constructor(options: APIOptions) {
         super(options);
         this.taskCollaboratorsAPI = [];
         this.taskLinksAPI = [];
+        this.errors = [];
     }
 
     /**
@@ -221,7 +222,7 @@ class Feed extends Base {
         const cache = this.getCache();
         const cacheKey = this.getCacheKey(id);
         cache.set(cacheKey, {
-            hasError: !!this.hasError,
+            errors: this.errors,
             items,
         });
     }
@@ -241,7 +242,7 @@ class Feed extends Base {
         file: BoxItem,
         shouldRefreshCache: boolean,
         successCallback: Function,
-        errorCallback: (feedItems: FeedItems) => void,
+        errorCallback: (feedItems: FeedItems, errors: ElementsXhrError[]) => void,
         onError: ErrorCallback,
         {
             shouldShowAnnotations = false,
@@ -251,9 +252,9 @@ class Feed extends Base {
         const { id, permissions = {} } = file;
         const cachedItems = this.getCachedItems(id);
         if (cachedItems) {
-            const { hasError, items } = cachedItems;
-            if (hasError) {
-                errorCallback(items);
+            const { errors, items } = cachedItems;
+            if (errors.length) {
+                errorCallback(items, errors);
             } else {
                 successCallback(items);
             }
@@ -264,7 +265,7 @@ class Feed extends Base {
         }
 
         this.file = file;
-        this.hasError = false;
+        this.errors = [];
         this.errorCallback = onError;
         const annotationsPromise = shouldShowAnnotations ? this.fetchAnnotations(permissions) : Promise.resolve();
         const versionsPromise = this.fetchVersions();
@@ -285,8 +286,8 @@ class Feed extends Base {
             const sortedFeedItems = sortFeedItems(versionsWithCurrent, ...feedItems);
             if (!this.isDestroyed()) {
                 this.setCachedItems(id, sortedFeedItems);
-                if (this.hasError) {
-                    errorCallback(sortedFeedItems);
+                if (this.errors.length) {
+                    errorCallback(sortedFeedItems, this.errors);
                 } else {
                     successCallback(sortedFeedItems);
                 }
@@ -979,7 +980,7 @@ class Feed extends Base {
      */
     feedErrorCallback = (hasError: boolean = false, e: ElementsXhrError, code: string): void => {
         if (hasError) {
-            this.hasError = true;
+            this.errors.push({ ...e, code });
         }
 
         if (!this.isDestroyed() && this.errorCallback) {
--- a/src/elements/common/annotator-context/__tests__/withAnnotations.test.tsx
+++ b/src/elements/common/annotator-context/__tests__/withAnnotations.test.tsx
@@ -182,6 +182,17 @@ describe('elements/common/annotator-context/withAnnotations', () => {
         );
     });
 
+    describe('handleAnnotationFetchError()', () => {
+        test('should call onError', () => {
+            const instance = getWrapper().instance();
+
+            const mockError = new Error();
+            instance.handleAnnotationFetchError({ error: mockError });
+
+            expect(defaults.onError).toHaveBeenCalledWith(mockError, 'fetch_annotations_error');
+        });
+    });
+
     describe('handlePreviewDestroy()', () => {
         test('should reset state and annotator', () => {
             const wrapper = getWrapper();
@@ -223,6 +234,10 @@ describe('elements/common/annotator-context/withAnnotations', () => {
                 instance.handleActiveChange,
             );
             expect(mockAnnotator.removeListener).toBeCalledWith('annotations_create', instance.handleAnnotationCreate);
+            expect(mockAnnotator.removeListener).toBeCalledWith(
+                'annotations_fetch_error',
+                instance.handleAnnotationFetchError,
+            );
         });
     });
 
--- a/src/elements/common/annotator-context/withAnnotations.tsx
+++ b/src/elements/common/annotator-context/withAnnotations.tsx
@@ -21,8 +21,8 @@ export type ComponentWithAnnotations = {
     handleActiveChange: ActiveChangeEventHandler;
     handleAnnotationChangeEvent: (id: string | null) => void;
     handleAnnotationCreate: (eventData: AnnotationActionEvent) => void;
+    handleAnnotationFetchError: ({ error }: { error: Error }) => void;
     handleAnnotator: (annotator: Annotator) => void;
-    handleAnnotatorEvent: ({ event, data }: { event: string; data?: unknown }) => void;
     handlePreviewDestroy: (shouldReset?: boolean) => void;
 };
 
@@ -130,10 +130,19 @@ export default function withAnnotations<P extends object>(
             this.setState({ activeAnnotationFileVersionId: fileVersionId, activeAnnotationId: annotationId });
         };
 
+        handleAnnotationFetchError = ({ error }: { error: Error }): void => {
+            const { onError } = this.props;
+
+            if (onError && error) {
+                onError(error, 'fetch_annotations_error');
+            }
+        };
+
         handleAnnotator = (annotator: Annotator): void => {
             this.annotator = annotator;
             this.annotator.addListener('annotations_active_change', this.handleActiveChange);
             this.annotator.addListener('annotations_create', this.handleAnnotationCreate);
+            this.annotator.addListener('annotations_fetch_error', this.handleAnnotationFetchError);
         };
 
         handlePreviewDestroy = (shouldReset = true): void => {
@@ -144,6 +153,7 @@ export default function withAnnotations<P extends object>(
             if (this.annotator) {
                 this.annotator.removeListener('annotations_active_change', this.handleActiveChange);
                 this.annotator.removeListener('annotations_create', this.handleAnnotationCreate);
+                this.annotator.removeListener('annotations_fetch_error', this.handleAnnotationFetchError);
             }
 
             this.annotator = null;
--- a/src/elements/content-preview/ContentPreview.js
+++ b/src/elements/content-preview/ContentPreview.js
@@ -792,7 +792,7 @@ class ContentPreview extends React.PureComponent<Props, State> {
         this.preview.addListener('thumbnailsClose', () => this.setState({ isThumbnailSidebarOpen: false }));
 
         if (showAnnotationsControls) {
-            this.preview.addListener('annotator', onAnnotator);
+            this.preview.addListener('annotator_create', onAnnotator);
         }
 
         this.preview.updateFileCache([file]);
--- a/src/elements/content-sidebar/ActivitySidebar.js
+++ b/src/elements/content-sidebar/ActivitySidebar.js
@@ -27,6 +27,7 @@ import { withLogger } from '../common/logger';
 import { withRouterAndRef } from '../common/routing';
 import {
     DEFAULT_COLLAB_DEBOUNCE,
+    ERROR_CODE_FETCH_ACTIVITY,
     ORIGIN_ACTIVITY_SIDEBAR,
     SIDEBAR_VIEW_ACTIVITY,
     TASK_COMPLETION_RULE_ALL,
@@ -442,11 +443,20 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
      * @param {Error} e - API error
      * @return {void}
      */
-    fetchFeedItemsErrorCallback = (feedItems: FeedItems): void => {
+    fetchFeedItemsErrorCallback = (feedItems: FeedItems, errors: ElementsXhrError[]): void => {
+        const { onError } = this.props;
+
         this.setState({
             feedItems,
             activityFeedError: activityFeedInlineError,
         });
+
+        if (Array.isArray(errors) && errors.length) {
+            onError(new Error('Fetch feed items error'), ERROR_CODE_FETCH_ACTIVITY, {
+                showNotification: true,
+                errors: errors.map(({ code }) => code),
+            });
+        }
     };
 
     /**
--- a/src/api/__tests__/Feed.test.js
+++ b/src/api/__tests__/Feed.test.js
@@ -63,6 +63,8 @@ const mockTask = {
     status: TASK_NEW_NOT_STARTED,
 };
 
+const mockErrors = [{ code: 'error_code_0' }, { code: 'error_code_1' }];
+
 const mockFirstVersion = {
     action: 'upload',
     type: 'file_version',
@@ -343,6 +345,7 @@ describe('api/Feed', () => {
     });
 
     afterEach(() => {
+        feed.errors = [];
         jest.restoreAllMocks();
     });
 
@@ -386,8 +389,8 @@ describe('api/Feed', () => {
             feed.setCachedItems(id, feedItems);
             expect(feed.getCacheKey).toHaveBeenCalledWith(id);
             expect(cache.set).toHaveBeenCalledWith(cacheKey, {
+                errors: [],
                 items: feedItems,
-                hasError: false,
             });
         });
     });
@@ -447,13 +450,13 @@ describe('api/Feed', () => {
 
         test('should get feed items, sort, save to cache, and call the error callback', done => {
             feed.fetchVersions = function fetchVersionsWithError() {
-                this.hasError = true;
+                this.errors = mockErrors;
                 return [];
             };
 
             feed.feedItems(file, false, successCb, errorCb);
             setImmediate(() => {
-                expect(errorCb).toHaveBeenCalledWith(sortedItems);
+                expect(errorCb).toHaveBeenCalledWith(sortedItems, mockErrors);
                 done();
             });
         });
@@ -502,7 +505,7 @@ describe('api/Feed', () => {
 
         test('should return the cached items', () => {
             feed.getCachedItems = jest.fn().mockReturnValue({
-                hasError: false,
+                errors: [],
                 items: feedItems,
             });
             feed.feedItems(file, false, successCb, errorCb);
@@ -512,7 +515,7 @@ describe('api/Feed', () => {
 
         test('should refresh the cache after returning the cached items', done => {
             feed.getCachedItems = jest.fn().mockReturnValue({
-                hasError: false,
+                errors: [],
                 items: feedItems,
             });
             feed.feedItems(file, true, successCb, errorCb);
@@ -613,7 +616,7 @@ describe('api/Feed', () => {
     describe('updateTaskCollaboratorSuccessCallback()', () => {
         beforeEach(() => {
             feed.getCachedItems = jest.fn().mockReturnValue({
-                hasError: false,
+                errors: [],
                 items: feedItems,
             });
             feed.updateFeedItem = jest.fn();
@@ -970,7 +973,7 @@ describe('api/Feed', () => {
 
         test('should delete the feed item and call success callback', () => {
             feed.getCachedItems = jest.fn().mockReturnValue({
-                hasError: false,
+                errors: [],
                 items: feedItems,
             });
             feed.deleteFeedItem(feedItemId, successCb);
@@ -998,11 +1001,11 @@ describe('api/Feed', () => {
             jest.restoreAllMocks();
         });
 
-        test('should log the error and set the hasError property if its not destroyed and hasError is set to true', () => {
+        test('should log the error and set the errors property if its not destroyed and hasError is set to true', () => {
             const hasError = true;
             feed.feedErrorCallback(hasError, e, code);
             expect(global.console.error).toBeCalledWith(e);
-            expect(feed.hasError).toBe(true);
+            expect(feed.errors).toEqual([{ ...e, code }]);
             expect(feed.errorCallback).toHaveBeenCalledWith(e, code, {
                 error: e,
                 [IS_ERROR_DISPLAYED]: hasError,
@@ -1018,14 +1021,12 @@ describe('api/Feed', () => {
             });
         });
 
-        test('should set hasError only if hasError is set', () => {
-            feed.hasError = null;
-            expect(feed.hasError).toBe(null);
-            const hasError = false;
-            feed.feedErrorCallback(hasError, e, code);
-            expect(feed.hasError).toBe(null);
+        test('should set errors only if hasError is true', () => {
+            feed.errors = [];
+            feed.feedErrorCallback(false, e, code);
+            expect(feed.errors).toEqual([]);
             feed.feedErrorCallback(true, e, code);
-            expect(feed.hasError).toBe(true);
+            expect(feed.errors).toEqual([{ ...e, code }]);
         });
     });
 
@@ -1052,7 +1053,7 @@ describe('api/Feed', () => {
 
         test('should create an item and add it to the feed with populated cache', () => {
             feed.getCachedItems = jest.fn().mockReturnValue({
-                hasError: false,
+                errors: [],
                 items: feedItems,
             });
 
@@ -1136,7 +1137,7 @@ describe('api/Feed', () => {
         beforeEach(() => {
             feed.setCachedItems = jest.fn();
             feed.getCachedItems = jest.fn().mockReturnValue({
-                hasError: false,
+                errors: [],
                 items: feedItems,
             });
         });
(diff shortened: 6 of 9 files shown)


src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -129,16 +131,40 @@ export default function withAnnotations<P extends object>(
             }
         };
 
+        handleAnnotationsFetchError = ({ error }: { error: Error }) => {
+            this.fetchError = error;
+        };
+
         handleAnnotator = (annotator: Annotator): void => {
```
jstoffan, line 138: Should this be combined with `handleAnnotatorCreate`? The logic within seems like it could work as the result of either event.

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -129,16 +131,40 @@ export default function withAnnotations<P extends object>(
...
+
+        handleSidebarFetchError = (error: Error) => {
+            const { annotator } = this;
+
+            if (!annotator) {
+                this.fetchError = error;
+                return;
+            }
+
+            annotator.emit('annotations_visible_set', false);
```
jstoffan, line 159: We agreed that this edge case was unlikely and that handling it would add unnecessary complexity. For example, what happens if the fetch succeeds later?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -129,16 +131,40 @@ export default function withAnnotations<P extends object>(
...
+
+        handleAnnotatorCreate = (annotator: Annotator): void => {
+            annotator.addListener('annotations_fetch_error', this.handleAnnotationsFetchError);
+        };
+
+        handleSidebarFetchError = (error: Error) => {
+            const { annotator } = this;
+
+            if (!annotator) {
+                this.fetchError = error;
```
jstoffan, line 155: We should be tracking state in appropriate ways. Changes to instance properties will not cause a re-render, even though this value is used in render.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -534,8 +564,13 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
         /* eslint-enable no-console */
 
-        // eslint-disable-next-line react/prop-types
-        this.props.onError(error, code, contextInfo);
+        const { onAnnotationsFetchError, onError } = this.props;
+
+        onError(error, code, contextInfo);
+
+        if (code === ERROR_CODE_FETCH_ANNOTATIONS) {
+            onAnnotationsFetchError(error);
```
jstoffan, line 572: Should this live in the fetchAnnotations error callback, instead of in the global handler?

src/elements/common/annotator-context/types.ts
```diff
@@ -28,10 +28,12 @@ export interface AnnotatorState {
 export type GetMatchPath = (history: History) => match<MatchParams> | null;
 
 export interface AnnotatorContext {
+    annotationsFetchError: Error | null;
```
ConradJChan, line 31: Does this need to be a separate error? Is there a way to leverage the `error` already part of `state`?

src/elements/common/annotator-context/withAnnotations.tsx
ConradJChan: Do we even need to listen for the `annotator` event anymore in `ContentPreview`? Can we just use the new `annotator_create`?

src/elements/common/annotator-context/withAnnotations.tsx
ConradJChan: We need a strategy to solve the product requirement for not showing annotations in either the content or sidebar if either one of the fetch calls fail, though, unless the requirement is able to change.

src/elements/common/annotator-context/withAnnotations.tsx
mxiao6: `handleAnnotatorCreate` needs to happen right after `annotator` construction (`init` has not finished yet), but hiding annotations has to happen after `init`. So both are needed.

src/elements/content-sidebar/ActivitySidebar.js
mxiao6: There is no fetchAnnotations error callback. Annotations are fetch as part of `feedItems` api call. We can only use this global error handler.

src/elements/content-preview/ContentPreview.js
```diff
@@ -1167,6 +1179,23 @@ class ContentPreview extends React.PureComponent<Props, State> {
         }
     };
 
+    handleAnnotationsFetchError = () => {
```
ConradJChan, line 1182: We've tried to encapsulate annotations logic to `withAnnotations` HOC (apart from setting up the callbacks to Preview SDK which we can't avoid currently -- is this something we could move there as well?

src/elements/common/annotator-context/types.ts
```diff
@@ -16,13 +16,21 @@ export interface Annotator {
     removeAllListeners: () => void;
     removeListener: (event: string | symbol, listener: (...args: any[]) => void) => void;
 }
+
+export type ElementsError = {
```
ConradJChan, line 20: Should this move to `src/types.ts`?

src/elements/common/annotator-context/types.ts
```diff
@@ -16,13 +16,21 @@ export interface Annotator {
...
+    code: string;
+    context_info: Record<string, unknown>;
```
ConradJChan, line 22: Could we use `object` instead? Would that help avoid the cast [below](https://github.com/box/box-ui-elements/pull/2172/files#diff-9f2f85578bd7824108ee496c56e3692eR111)?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -81,6 +100,24 @@ export default function withAnnotations<P extends object>(
...
+        escalateAnnotationsFetchError = (error: ElementsError) => {
+            const { onError } = this.props;
+
+            if (onError) {
+                onError(
+                    {
+                        code: ERROR_CODE_FETCH_ANNOTATIONS,
+                        // eslint-disable-next-line @typescript-eslint/camelcase
+                        context_info: error as Record<string, unknown>,
+                        message: error.message || '',
```
ConradJChan, line 112: Could we destructure `message` and `origin` and assign defaults above?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -130,9 +167,46 @@ export default function withAnnotations<P extends object>(
             }
         };
 
+        handleAnnotationsFetchError = ({ error }: { error: ElementsError }): void => {
```
ConradJChan, line 170: Would the error emitted from annotations be of type `ElementsError`?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -130,9 +167,46 @@ export default function withAnnotations<P extends object>(
...
+        handleSidebarFetchError = (error: ElementsError): void => {
+            const { annotator } = this;
+
+            if (!annotator) {
+                this.setState({ error });
+                return;
+            }
+
+            this.escalateAnnotationsFetchError(error);
+            annotator.emit('annotations_visible_set', false);
```
ConradJChan, line 209: How would this handle the scenario where the sidebar fetch failed before the annotator was emitted? Would it make it easier to reason about if we added some state for `hasAnnotator` or something similar and in `componentDidUpdate`:

- `if (hasAnnotator && error)` then 
- escalate the error
- `annotator.emit('annotations_visible_set', false)`

Though that would potentially require separating the component state from `AnnotatorState`

mxiao6: If sidebar fetch failed before annotator was emitted, `handleSidebarFetchError` would set error in state. And when annotator is emitted, `handleAnnotator` will check the error in state and call `handleSidebarFetchError` again.

mxiao6: Yeah, that'll simplify the logic. Updated.

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -3,26 +3,44 @@ import getProp from 'lodash/get';
...
     emitActiveChangeEvent: (id: string | null) => void;
     emitRemoveEvent: (id: string) => void;
+    escalateAnnotationsFetchError: (error: ElementsError) => void;
     getAction: (eventData: AnnotationActionEvent) => Action;
     getAnnotationsPath: (fileVersionId?: string, annotationId?: string | null) => string;
     getMatchPath: GetMatchPath;
     handleActiveChange: (annotationId: string | null) => void;
     handleAnnotationChangeEvent: (id: string | null) => void;
     handleAnnotationCreate: (eventData: AnnotationActionEvent) => void;
+    handleAnnotationsFetchError: ({ error }: { error: SerializedError }) => void;
```
ConradJChan, line 38: Does this need to have `SerializedError` type? Would the existing `Error` type work?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -61,6 +80,25 @@ export default function withAnnotations<P extends object>(
             this.state = { ...defaultState, activeAnnotationId };
         }
 
+        componentDidUpdate(
+            _prevProps: WithAnnotationsProps,
```
ConradJChan, line 84: Is prepending with underscore a convention for ignored arguments?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -61,6 +80,25 @@ export default function withAnnotations<P extends object>(
...
+        ) {
+            const { hasAnnotator, error } = this.state;
+
+            if (
+                (prevHasAnnotator !== hasAnnotator || prevError !== error) &&
+                hasAnnotator &&
+                this.annotator &&
+                error &&
+                'origin' in error &&
+                error.origin === ORIGIN_ACTIVITY_SIDEBAR
```
ConradJChan, line 95: Maybe this can be simplified if:
- `handleAnnotationsFetchError` doesn't worry about calling `this.escalateAnnotationsFetchError` but relies on `componentDidUpdate`
- we pull out conditions as variables

Something like:
```
const hasAnnotatorChanged = this.annotator && prevHasAnnotator !== hasAnnotator;
const hasErrorChanged = error && prevError !== error;
if (hasAnnotatorChanged && hasErrorChanged) {
  this.escalateAnnotationsFetchError(error);
  this.annotator.emit('annotations_visible_set', false);
}
```

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -130,9 +187,41 @@ export default function withAnnotations<P extends object>(
...
+
         handleAnnotator = (annotator: Annotator): void => {
             this.annotator = annotator;
             this.annotator.addListener('annotatorevent', this.handleAnnotatorEvent);
+
+            this.setState({ hasAnnotator: true });
+        };
+
+        handleAnnotatorCreate = (annotator: Annotator): void => {
+            annotator.addListener('annotations_fetch_error', this.handleAnnotationsFetchError);
```
ConradJChan, line 220: It seems a bit strange to call `addListener` on the passed in `annotator` imo. Could `handleAnnotatorCreate` do the assignment to `this.annotator` and add the listeners for `annotations_fetch_error` and `annotatorevent` and then have `handleAnnotator` just set the state for `hasAnnotator`?

And along those lines, what do you think about changing the state from `hasAnnotator` => `hasAnnotatorInit`

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -137,6 +139,31 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         this.fetchCurrentUser(currentUser);
     }
 
+    componentDidUpdate({ annotatorState: prevAnnotatorState }: Props): void {
+        const { error: prevError } = prevAnnotatorState;
+        const {
+            annotatorState: { error },
+        } = this.props;
+
+        if (!prevError && error && error.code === ERROR_CODE_FETCH_ANNOTATIONS) {
```
ConradJChan, line 148: Should this just check that `prevError !== error` instead of `!prevError && error` in case there was previously another error such as annotation create?

src/elements/common/annotator-context/withAnnotations.tsx
mxiao6: Yeah, and Eslint and TS won't complain it

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -69,6 +81,24 @@ export default function withAnnotations<P extends object>(
...
+        componentDidUpdate(
+            _prevProps: WithAnnotationsProps,
+            { error: prevError, hasAnnotatorInit: prevHasAnnotatorInit }: AnnotatorState,
+        ) {
+            const { error, hasAnnotatorInit } = this.state;
+
+            const hasAnnotatorChanged = prevHasAnnotatorInit !== hasAnnotatorInit;
+            const hasErrorChanged = prevError !== error;
+
+            if ((hasAnnotatorChanged || hasErrorChanged) && error) {
```
ConradJChan, line 93: Should this be `(hasAnnotatorChanged || hasErrorChanged) && error && hasAnnotatorInit`?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -89,6 +119,26 @@ export default function withAnnotations<P extends object>(
...
+                const origin = 'origin' in error && error.origin ? error.origin : ORIGIN_CONTENT_PREVIEW;
+                onError(
+                    {
+                        code: ERROR_CODE_FETCH_ANNOTATIONS,
+                        // eslint-disable-next-line @typescript-eslint/camelcase
+                        context_info: error,
+                        message,
+                        origin,
+                        type: 'error',
+                    },
```
ConradJChan, line 136: Could create a `createElementsError` or something like that to create the formatted error and avoid repeating it here and L198

src/elements/common/annotator-context/__tests__/withAnnotations.test.tsx
```diff
@@ -63,12 +77,31 @@ describe('elements/common/annotator-context/withAnnotations', () => {
         });
     });
 
+    describe('componentDidUpdate()', () => {
+        test('should call escalateAnnotationsFetchError and emit event', () => {
```
ConradJChan, line 81: Should we test the other branches of logic to make sure it doesn't emit if no error, or hasn't init, or if it has init but has a new error?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -137,6 +139,37 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         this.fetchCurrentUser(currentUser);
     }
 
+    componentDidUpdate({ annotatorState: prevAnnotatorState }: Props): void {
+        const { error: prevError } = prevAnnotatorState;
+        const {
+            annotatorState: { error },
+        } = this.props;
+
+        if (prevError !== error && error) {
```
ConradJChan, line 148: Maybe `error && prevError !== error` instead?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -137,6 +139,37 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+            switch (error.code) {
+                case ERROR_CODE_FETCH_ANNOTATIONS:
+                    this.removeAnnotations();
```
ConradJChan, line 151: Maybe `filterAnnotations` to be consistent?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -432,7 +465,15 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
      */
     fetchFeedItemsSuccessCallback = (feedItems: FeedItems): void => {
-        this.setState({ feedItems, activityFeedError: undefined });
+        const {
+            annotatorState: { error },
+        } = this.props;
+
+        this.setState({
+            feedItems:
+                error && error.code === ERROR_CODE_FETCH_ANNOTATIONS ? this.filterAnnotations(feedItems) : feedItems,
```
ConradJChan, line 474: Maybe split out the filtering into a separate variable to ease readability?

src/elements/common/annotator-context/types.ts
```diff
@@ -1,5 +1,6 @@
 import { Location } from 'history';
 import { match } from 'react-router-dom';
+import { ElementsError } from '../../../types';
```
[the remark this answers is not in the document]
mxiao6: I think relative import should come after absolute import, right

src/elements/content-sidebar/ActivitySidebar.js
mxiao6: There's already one `filterAnnotations` which is called by `removeAnnotations`.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -432,7 +465,20 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
     fetchFeedItemsSuccessCallback = (feedItems: FeedItems): void => {
-        this.setState({ feedItems, activityFeedError: undefined });
+        const {
+            annotatorState: { error },
+        } = this.props;
+
+        let newFeedItems = feedItems;
+
+        if (error && error.code === ERROR_CODE_FETCH_ANNOTATIONS) {
+            newFeedItems = this.filterAnnotations(feedItems);
```
jstoffan, line 475: What scenario does this serve? If annotations failed to load, would they not already be absent from the feed items collection?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -137,6 +139,37 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
     }
 
+    componentDidUpdate({ annotatorState: prevAnnotatorState }: Props): void {
+        const { error: prevError } = prevAnnotatorState;
+        const {
+            annotatorState: { error },
+        } = this.props;
+
+        if (error && prevError !== error) {
+            switch (error.code) {
```
jstoffan, line 149: Do we expect further codes to be handled here? It seems like this could be simplified.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -137,6 +139,37 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+    }
+
+    removeAnnotations = (): void => {
+        const { feedItems } = this.state;
+
+        if (!feedItems) {
+            return;
+        }
+
+        this.setState({
```
jstoffan, line 166: Rather than directly manipularting the local state, should we update the feedAPI cache, instead? We could add a new `clearFeedItems` method that accepts a type parameter, then refresh the feed. Otherwise, the local state may be restored from the "stale" cache at a later time.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -463,8 +509,13 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
         /* eslint-enable no-console */
 
-        // eslint-disable-next-line react/prop-types
-        this.props.onError(error, code, contextInfo);
+        const { onAnnotationsFetchError, onError } = this.props;
+
+        onError(error, code, contextInfo);
+
+        if (code === ERROR_CODE_FETCH_ANNOTATIONS) {
+            onAnnotationsFetchError({ ...error, origin: ORIGIN_ACTIVITY_SIDEBAR });
```
jstoffan, line 517: It looks like this would lead to ContentSidebar/ContentPreview's `onError` prop being called twice for the same error. Can the Annotations HOC be limited to only managing the `annotator` instance?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -463,8 +509,13 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         console.error(error);
         /* eslint-enable no-console */
 
-        // eslint-disable-next-line react/prop-types
-        this.props.onError(error, code, contextInfo);
+        const { onAnnotationsFetchError, onError } = this.props;
+
+        onError(error, code, contextInfo);
```
jstoffan, line 514: Can we use the `contextInfo` object to notify the parent application of how we want it to behave? We already do something similar for other errors.

src/elements/common/annotator-context/withAnnotatorContext.tsx
```diff
@@ -22,6 +31,7 @@ export default function withAnnotatorContext<P extends {}>(WrappedComponent: Rea
                     emitRemoveEvent={emitRemoveEvent}
                     getAnnotationsMatchPath={getAnnotationsMatchPath}
                     getAnnotationsPath={getAnnotationsPath}
+                    onAnnotationsFetchError={onAnnotationsFetchError}
```
jstoffan, line 34: Does each context property need to be destructured individually or can we use a spread operator here?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -138,9 +191,28 @@ export default function withAnnotations<P extends object>(
             }
         };
 
-        handleAnnotator = (annotator: Annotator): void => {
+        handleAnnotationsFetchError = ({ error }: { error: Error }): void => {
+            if (!error) {
+                return;
+            }
+
+            this.setState({
```
jstoffan, line 199: Does this need to be set in state or can we just immediately trigger `this.props.onError`?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -69,6 +82,24 @@ export default function withAnnotations<P extends object>(
...
+        ) {
+            const { error, hasAnnotatorInit } = this.state;
+
+            const hasAnnotatorChanged = prevHasAnnotatorInit !== hasAnnotatorInit;
+            const hasErrorChanged = prevError !== error;
+
+            if ((hasAnnotatorChanged || hasErrorChanged) && error && hasAnnotatorInit) {
+                this.escalateAnnotationsFetchError(error);
+
+                if ('origin' in error && error.origin === ORIGIN_ACTIVITY_SIDEBAR && this.annotator) {
```
jstoffan, line 97: Can this be simplified to `if (this.annotator && error && error.origin === ORIGIN_ACTIVITY_SIDEBAR)`

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -69,6 +82,24 @@ export default function withAnnotations<P extends object>(
...
+                    this.annotator.emit('annotations_visible_set', false);
```
jstoffan, line 98: Should we instead tell preview to destroy the annotator and/or prevent it from creating it in the first place? Do we want the region toggle to show if the fetch call fails?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -89,6 +120,28 @@ export default function withAnnotations<P extends object>(
...
         };
 
+        createElementsError = (error: ElementsError | Error): ElementsError => {
+            const { message = '' } = error;
+            const origin = 'origin' in error && error.origin ? error.origin : ORIGIN_CONTENT_PREVIEW;
+
+            return {
+                code: ERROR_CODE_FETCH_ANNOTATIONS,
+                // eslint-disable-next-line @typescript-eslint/camelcase
+                context_info: error,
```
jstoffan, line 130: I think the main purpose of `context_info` is to provide data that's not directly available in the error. Can you investigate how the feed API uses this?

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -89,6 +120,28 @@ export default function withAnnotations<P extends object>(
             annotator.emit('annotations_remove', id);
         };
 
+        createElementsError = (error: ElementsError | Error): ElementsError => {
+            const { message = '' } = error;
+            const origin = 'origin' in error && error.origin ? error.origin : ORIGIN_CONTENT_PREVIEW;
```
jstoffan, line 125: It seems like the origin would always be `ORIGIN_CONTENT_PREVIEW`, since this HOC wraps that element.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -442,11 +443,20 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
-    fetchFeedItemsErrorCallback = (feedItems: FeedItems): void => {
+    fetchFeedItemsErrorCallback = (feedItems: FeedItems, errors: ElementsXhrError[]): void => {
+        const { onError } = this.props;
+
         this.setState({
             feedItems,
             activityFeedError: activityFeedInlineError,
         });
+
+        if (errors.length) {
```
jstoffan, line 454: Is `errors` always defined as an array?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -442,11 +443,20 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+            onError(errors[0], ERROR_CODE_UNKNOWN, {
```
jstoffan, line 455: It seems like this will be difficult to differentiate the parent application, since the error will be dynamic. Should we create a new error object and code, such as `ERROR_CODE_FETCH_ACTIVITY`, instead?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -442,11 +443,20 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+                showNotification: true,
+                errors,
```
jstoffan, line 457: Should we compile the failed activity type strings here rather than passing up the complete error objects?

src/api/Feed.js
```diff
@@ -980,6 +987,7 @@ class Feed extends Base {
     feedErrorCallback = (hasError: boolean = false, e: ElementsXhrError, code: string): void => {
         if (hasError) {
             this.hasError = true;
```
jstoffan, line 989: Can this be derived from `this.errors` now?

src/api/Feed.js
```diff
@@ -286,7 +293,7 @@ class Feed extends Base {
             if (!this.isDestroyed()) {
                 this.setCachedItems(id, sortedFeedItems);
                 if (this.hasError) {
-                    errorCallback(sortedFeedItems);
+                    errorCallback(sortedFeedItems, this.errors);
```
jstoffan, line 296: Should we update each constituent promise to bind its `reject` callback to `fetchFeedItemErrorCallback` along with `resolve`? Each error callback could then pass its error object to `reject`. That would allow us to use `.catch`  on the main Promise.all promise chain above, which would remove the need for an instance level `errors` property.

mxiao6: `Promise.all` will reject immediately upon any of the input promises rejecting, which is not the current behavior.

src/api/Feed.js
```diff
@@ -221,7 +222,7 @@ class Feed extends Base {
         const cache = this.getCache();
         const cacheKey = this.getCacheKey(id);
         cache.set(cacheKey, {
-            hasError: !!this.hasError,
```
jstoffan, line 224: Does any other module use this value from the cache?

src/api/Feed.js
```diff
@@ -979,7 +980,7 @@ class Feed extends Base {
      */
     feedErrorCallback = (hasError: boolean = false, e: ElementsXhrError, code: string): void => {
         if (hasError) {
```
jstoffan, line 982: It looks like this is only true if the error is considered "user-correctable". Does that cover our use cases?

src/elements/content-preview/ContentPreview.js
```diff
@@ -792,7 +792,7 @@ class ContentPreview extends React.PureComponent<Props, State> {
         this.preview.addListener('thumbnailsClose', () => this.setState({ isThumbnailSidebarOpen: false }));
 
         if (showAnnotationsControls) {
-            this.preview.addListener('annotator', onAnnotator);
+            this.preview.addListener('annotator_create', onAnnotator);
```
jstoffan, line 795: Can you remind me of the reason for this change? Does the annotator `annotations_fetch_error` event fire before preview's `annotator` event?

mxiao6: Yeah, `annotations_fetch_error` event fires before `annotator` event.

src/api/Feed.js
mxiao6: No, only `items` value fro the cache is used in other module.

src/api/Feed.js
mxiao6: I actually don't understand the concept "user-correctable". Our use cases is fetch error, the possible reason might be timeout (504) or service unavailable (503). I looked into the code and if the error code is larger than 500, `hasError` will be true. So I think it covers our use cases.

PR #2173, opened 2020-06-05
feat(annotations): handle fileVersionId when active annotation changes

ConradJChan:


src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -11,13 +11,20 @@ export interface WithAnnotationsProps {
     onPreviewDestroy: () => void;
 }
 
+type ActiveChangeEvent = {
+    annotationId: string | null;
+    fileVersionId: string;
```
jstoffan, line 16: Is this optional/nullable or do we expect it to be defined even if `annotationId` is null? Seems a little odd.

src/elements/content-sidebar/withSidebarAnnotations.js
```diff
@@ -135,7 +135,11 @@ export default function withSidebarAnnotations(
             } = this.props;
             const match = getAnnotationsMatchPath(location);
             const currentFileVersionId = getProp(file, 'file_version.id');
-            const fileVersionId = getProp(match, 'params.fileVersionId', currentFileVersionId);
+            const fileVersionId = getProp(
+                match,
+                'params.fileVersionId',
+                activeAnnotationFileVersionId || currentFileVersionId,
```
jstoffan, line 141: Pull out into a `defaultFileVersionId` variable, maybe, for readability?

src/elements/common/annotator-context/withAnnotations.tsx
ConradJChan: I was expecting `fileVersionId` to always be defined since null is actually a valid active annotation value

PR #2174, opened 2020-06-05
fix(annotations): cleanup Preview when changing versions

ConradJChan:


src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -136,8 +136,6 @@ export default function withAnnotations<P extends object>(
         };
 
         handlePreviewDestroy = (): void => {
-            this.setState(defaultState);
```
jstoffan, line 139: No longer needed? Does it get reset elsewhere?

src/elements/content-preview/ContentPreview.js
```diff
@@ -366,8 +364,10 @@ class ContentPreview extends React.PureComponent<Props, State> {
 
         if (hasFileIdChanged) {
             this.destroyPreview();
+            this.setState({ selectedVersion: undefined });
             this.fetchFile(currentFileId);
         } else if (this.shouldLoadPreview(prevState)) {
+            this.destroyPreview();
```
jstoffan, line 370: This seems like a fairly risky change since it'll affect lots of cases. Are we confident about it? Does it cause any visual issues when navigating between files?

ConradJChan: At the very least, the majority of the use cases were file id changing and that has always called `destroyPreview`, so changing files works fine. It appears to work fine switching between versions as well

src/elements/common/annotator-context/withAnnotations.tsx
ConradJChan: A full reset overrides the location changes made by the sidebar, particularly around changing versions to an active annotation id. Maybe I’ll reset other parts of the state but leave active annotation alone

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -135,8 +135,10 @@ export default function withAnnotations<P extends object>(
             this.annotator.addListener('annotatorevent', this.handleAnnotatorEvent);
         };
 
-        handlePreviewDestroy = (): void => {
-            this.setState(defaultState);
+        handlePreviewDestroy = (totally = true): void => {
```
jstoffan, line 138: How about `shouldReset`, instead? 🏄

PR #2176, opened 2020-06-08
feat(tooltip): adjust border-radius for tooltips

laughnan:
### Changes:

- Globally change `Tooltip` border radius to be `6px`
- Include missing `Avatar` image snapshots
- Add in global border radius various for Tooltips
- Update all snapshots that include `Tooltip` components

### Screenshots:
Screen Shot 2020-06-07 at 19 02 10
Screen Shot 2020-06-07 at 19 02 18


laughnan:
Per Design, we're going to keep with 4 pixels for now. If we want to tweak tooltips in the future we can/will revisit.

PR #2178, opened 2020-06-08
fix(annotations): scroll to annotation on load

mickr:
Adding logic that will scroll to an annotation from a hard load.

* Add logic to let Activity sidebar know when the viewer is ready.
* Handle scrolling to annotation in Activity Sidebar
* Add method to find cached item

- [ ] Unit Tests

### DIFF
--- a/src/elements/content-sidebar/ActivitySidebar.js
+++ b/src/elements/content-sidebar/ActivitySidebar.js
@@ -141,15 +141,15 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
 
     componentDidUpdate({ isViewerReady: prevIsViewerReady }: Props, { feedItems: prevFeedItems }: State) {
         const { getAnnotationsMatchPath, isViewerReady, location } = this.props;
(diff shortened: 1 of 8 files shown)


src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -133,10 +136,39 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
 
     componentDidMount() {
         const { currentUser } = this.props;
-        this.fetchFeedItems(true);
         this.fetchCurrentUser(currentUser);
     }
 
+    componentDidUpdate({ isViewerReady: prevIsViewerReady }: Props, { feedItems: prevFeedItems }: State) {
+        const { getAnnotationsMatchPath, isViewerReady, location } = this.props;
+        const match = getProp(getAnnotationsMatchPath(location), 'params.annotationId', null);
```
ConradJChan, line 144: Maybe rename to `activeAnnotationId`

PR #2179, opened 2020-06-08
feat(pill-selector-dropdown): update styles to latest design

booc0mtaco:


booc0mtaco:
Before: 
Screen Shot 2020-06-08 at 3 53 29 PM

Screen Shot 2020-06-08 at 4 01 24 PM

After:

Screen Shot 2020-06-08 at 2 51 41 PM

Screen Shot 2020-06-08 at 4 00 59 PM

src/components/datalist-item/DatalistItem.scss
```diff
@@ -1,7 +1,9 @@
 @import '../../styles/variables';
 
 .datalist-item {
-    padding: 5px 35px 5px 15px;
+    margin: 0 ($bdl-grid-unit * 2);
```
danubevictoria, line 4: Does this affect other things that could be using data list items?

booc0mtaco: yeah, that is the first screenshot in the "after" section

PR #2181, opened 2020-06-10
feat(dropdown-menu): add ability to pass down tetherProps to menu

booc0mtaco:
- reorder the documentation of the props as they were not aligned to the proper prop
- pass down `tetherProps` to corresponding menu components


booc0mtaco:
Given #2183, should we instead pluck out the single tether key/value pair and add it as a prop to this and other components, to conform to how other components in buie behave?

booc0mtaco:
There is no interest in standardizing any of this in this project, so closing.

PR #2186, opened 2020-06-11
feat(unified-share-modal): support method to get image url for avatars

booc0mtaco:
allow for specifying a method for retrieving the avatar URL. flexible to support custom schemes or url patterns to tie into other systems, or refer to other assets

### DIFF
--- a/src/components/contact-datalist-item/ContactDatalistItem.js
+++ b/src/components/contact-datalist-item/ContactDatalistItem.js
@@ -7,17 +7,25 @@ import DatalistItem from '../datalist-item';
 import './ContactDatalistItem.scss';
 
 type Props = {
-    id?: string | number | null,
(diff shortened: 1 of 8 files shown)


src/components/contact-datalist-item/ContactDatalistItem.js
```diff
@@ -7,17 +7,25 @@ import DatalistItem from '../datalist-item';
 import './ContactDatalistItem.scss';
 
 type Props = {
-    id?: string | number | null,
+    getContactAvatarUrl?: (contact: { id: string | number, [key: string]: any }) => string,
```
booc0mtaco, line 10: including the only required field and allowing the rest to vary for now. if we had a top-level type matching contact data, we could support a stricter type here.

PR #2192, opened 2020-06-18
feat(content-sharing): Add ContentSharing Element

sahiga:
This PR adds a UI Element for the USM that connects to Box APIs. The connection is currently limited to GET requests to the Item/Shared Link/Users APIs. POST requests, as well as connections to the Collaborators API, will be added in a follow-up PR.

**Normal state**
content-sharing-without-error

**Error state**
content-sharing-with-error


swfree:
Can you add a screenshot for the element & the error mask state? Also, do you need to export the element from `src/elements/index.js`?

src/elements/unified-share-modal-element/UnifiedShareModalElement.js
```diff
@@ -0,0 +1,172 @@
...
+    useEffect(() => {
+        const getUserData = async () => {
+            await api.getUsersAPI(false).getUser(itemID, getUserSuccess, getError, {
+                fields: [FIELD_ENTERPRISE, FIELD_HOSTNAME],
+            });
+        };
+        if (item && sharedLink && !currentUserID) {
+            getUserData();
+        }
+    });
```
swfree, line 103: More a question for my own knowledge, since I haven't used effects much: is there any recommended sorting order for useEffect calls? Just wondering since this one doesn't have skipping logic like the other two, maybe it should come first (since I'm equating this one to a typical call made in componentDidMount)?

src/elements/unified-share-modal-element/__tests__/utils.test.js
```diff
@@ -0,0 +1,284 @@
...
+jest.mock('../../../utils/file', () => ({
+    getTypedFileId: () => 'f_190457309',
+    getTypedFolderId: () => 'd_190457309',
+}));
+
+describe('normalizeItemResponse()', () => {
+    const TYPED_FILE_ID = 'f_190457309';
+    const TYPED_FOLDER_ID = 'd_190457309';
+    const ITEM_ID = '190457309';
+    const ITEM_NAME = '<3 Box UI Elements <3';
```
swfree, line 14: 😄

src/features/unified-share-modal/UnifiedShareModal.js
```diff
@@ -41,11 +41,13 @@ class UnifiedShareModal extends React.Component<USMProps, State> {
...
 
+        const { initialDataReceived } = props;
+
         this.state = {
-            getInitialDataCalled: false,
+            getInitialDataCalled: !!initialDataReceived,
             isConfirmModalOpen: false,
             isEmailLinkSectionExpanded: false,
-            isFetching: true,
+            isFetching: !initialDataReceived,
```
macrotim, line 50: If initial-data was received, I'd expect `isFetching` to be false. Unless I'm missing something?

src/elements/unified-share-modal-element/UnifiedShareModalElement.js
```diff
@@ -0,0 +1,172 @@
...
+            );
+        }
+
+        return null;
+    };
+
+    return renderElement();
+}
+
+export default UnifiedShareModalElement;
```
macrotim, line 172: At a high level, this looks good 👍

src/constants.js
```diff
@@ -187,6 +190,7 @@ export const CLIENT_NAME_CONTENT_EXPLORER = 'ContentExplorer';
 export const CLIENT_NAME_CONTENT_PREVIEW = 'ContentPreview';
 export const CLIENT_NAME_CONTENT_SIDEBAR = 'ContentSidebar';
 export const CLIENT_NAME_OPEN_WITH = 'ContentOpenWith';
+export const CLIENT_NAME_USM_ELEMENT = 'USMElement';
```
booc0mtaco, line 193: Could we consider a different name for the exposed component? `ContentSharing` perhaps?

src/elements/unified-share-modal-element/UnifiedShareModalElement.js
```diff
@@ -0,0 +1,172 @@
...
+                await api.getFolderAPI().getFolderFields(itemID, getItemSuccess, getError, {
+                    fields: [
+                        FIELD_DESCRIPTION,
+                        FIELD_EXTENSION,
+                        FIELD_ID,
+                        FIELD_NAME,
+                        FIELD_PERMISSIONS,
+                        FIELD_SHARED_LINK,
+                        FIELD_SHARED_LINK_FEATURES,
+                        FIELD_TYPE,
```
booc0mtaco, line 132: I know what this means, but the name is a bit misleading in this context. The field name is "type" but this could be misconstrued as "a type of field".

src/elements/unified-share-modal-element/UnifiedShareModalElement.js
```diff
@@ -0,0 +1,172 @@
...
+        if (errorExists) {
+            return <ErrorMask errorHeader={<FormattedMessage {...elementMessages.loadingError} />} />;
+        }
+
+        if (item && sharedLink) {
+            return (
+                <Internationalize language={language} messages={messages}>
+                    <UnifiedShareModal
+                        canInvite={sharedLink.canInvite}
+                        collaboratorsList={{ collaborators: [] }}
```
booc0mtaco, line 153: Can this be derived from having access to the item and an API instance, or do we have to provide it via an external prop?

src/elements/unified-share-modal-element/types.js
```diff
@@ -0,0 +1,82 @@
...
+    etag: string,
+    extension: string,
+    id: string,
+    name: string,
+    permissions: {
+        can_annotate: boolean,
+        can_comment: boolean,
+        can_delete: boolean,
+        can_download: boolean,
+        can_edit?: boolean,
```
booc0mtaco, line 49: any reason why this one is optional? (cannot have this field on a folder item type?)

src/elements/unified-share-modal-element/__tests__/UnifiedShareModalElement.test.js
```diff
@@ -0,0 +1,294 @@
...
+import ErrorMask from '../../../components/error-mask/ErrorMask';
+import UnifiedShareModalElement from '../UnifiedShareModalElement';
+import UnifiedShareModal from '../../../features/unified-share-modal/UnifiedShareModal';
+import { DEFAULT_HOSTNAME_API, TYPE_FILE, TYPE_FOLDER } from '../../../constants';
+import { normalizeItemResponse, normalizeUserResponse } from '../utils';
+
+jest.mock('../../../api');
+jest.mock('../utils');
+
+const SHARED_LINK = {
```
booc0mtaco, line 14: Each of these represent the internal data format for the USM component, i think. may be helpful to label them here, so they arent confused with the expected element types. Might also help to pull them into a separate file if they solely exist as testing helper objects.

src/elements/unified-share-modal-element/UnifiedShareModalElement.js
sahiga: Good idea. I updated the order of the `useEffect` calls to reflect that the call to the Item API occurs before the call to the Users API.

src/features/unified-share-modal/UnifiedShareModal.js
sahiga: Yes, that's correct. When the ContentSharing element instantiates the USM, it passes `initialDataReceived` as a prop, so `isFetching` will be false.

src/elements/unified-share-modal-element/UnifiedShareModalElement.js
sahiga: Yes, I'll request this data from the Collaborators API in a follow-up PR. I added a to-do note next to this line.

src/elements/unified-share-modal-element/types.js
sahiga: Good catch. It turns out that `can_edit` is only available for annotations and comments, so I updated the Item API requests to include the `allowed_invitee_roles` field. The normalizer function now checks that array for the presence of the "editor" role.

src/elements/content-sharing/ContentSharing.js
```diff
@@ -0,0 +1,175 @@
...
+                        FIELD_ITEM_TYPE,
+                    ],
+                });
+            }
+        };
+
+        if (!item && !sharedLink) {
+            getItem();
+        }
+    }, [api, getError, item, itemID, itemType, sharedLink, currentUserID]);
```
booc0mtaco, line 133: This may be a silly question. item, itemID, itemType, and sharedLink are all related. I am not fully informed on how `useEffect` works, but curious if it will be triggered unnecessarily, given these somewhat derived variables. 

Example: item has `.id` and `.type`, and `.sharedLink (`BoxItem`). `useEffect` presumably responds to these values changing. could we use fewer of these without triggering the correct number of side-effects? same for the others with this subset

src/elements/content-sharing/utils.js
```diff
@@ -0,0 +1,148 @@
+// @flow
```
booc0mtaco, line 1: A small part of me thinks this one should live with the USM or as a separate feature utility, as other BUIE features may need to use/extend the normalizer. I could imagine there would be one function for each defined API object. Responses that already match could be simple pass-thru methods. ones that are not possible/implemented could return `null` or similar. 

`normalize${Box API Object}Response` is a serviceable name. In a similar case, I might also suggest convert as the prefix. Normalize might imply some DB-type operation like converting redundant data into lists and objects.

src/elements/content-sharing/ContentSharing.js
```diff
@@ -0,0 +1,175 @@
...
+    const { apiHost, itemID, itemType, language, token }: ContentSharingProps = props;
+    const [item, setItem] = useState<?itemFlowType>(null);
+    const [sharedLink, setSharedLink] = useState<?ContentSharingSharedLinkType>(null);
+    const [currentUserID, setCurrentUserID] = useState<?string>(null);
+    const [errorExists, setErrorExists] = useState<boolean>(false);
+
+    const api = new API({
+        apiHost,
+        clientName: CLIENT_NAME_CONTENT_SHARING,
+        id: `${itemType}_usm`,
```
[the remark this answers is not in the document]
sahiga: Thanks for catching this. I noticed that ContentExplorer uses the format `folder_${itemID}`, so I updated the ContentSharing ID to follow that.

src/elements/content-sharing/ContentSharing.js
sahiga: By default, React runs each `useEffect` function whenever any of the values in the component scope changes; specifying this array allows React to skip the effect if none of the values in the array has changed. So, this is actually an optimization! The array has to include all values that the effect uses, so that the references to the values remain current (also, I get a lint error if I try to remove any of them)

src/elements/content-sharing/utils.js
sahiga: Good idea. I moved this file into the USM utils folder under the name `convertData`. For now, it's still specific to the ContentSharing API use cases, but we can make it more generic later on if necessary.

src/elements/content-sharing/ContentSharing.js
```diff
@@ -0,0 +1,171 @@
...
+    itemType: ItemType,
+    language: string,
+    showFormOnly?: boolean,
+    token: string,
+};
+
+const elementMessages = defineMessages({
+    loadingError: {
+        defaultMessage: 'Could not load shared link for this item.',
+        description: 'Message that appears when the USM cannot be loaded.',
```
swfree, line 46: minor: do we want to description to say "Content Sharing" instead of USM?

src/elements/content-sharing/ContentSharing.js
```diff
@@ -0,0 +1,171 @@
...
+    useEffect(() => {
+        const getUserSuccess = userData => {
+            const { id, userEnterpriseData } = convertUserResponse(userData);
+            setSharedLink({ ...sharedLink, ...userEnterpriseData });
+            setCurrentUserID(id);
+            setErrorExists(false);
+        };
+
+        const getUserData = async () => {
+            await api.getUsersAPI(false).getUser(itemID, getUserSuccess, getError, {
```
stevegcoder, line 131: getUsersAPI.getUser takes in a callback for success and returns a promise?  Ideally it would just return a promise that you can use to get the success/failure state.  This is probably pre-existing functionality though

src/features/unified-share-modal/UnifiedShareModal.js
```diff
@@ -41,11 +41,13 @@ class UnifiedShareModal extends React.Component<USMProps, State> {
     constructor(props: USMProps) {
         super(props);
 
+        const { initialDataReceived } = props;
+
         this.state = {
-            getInitialDataCalled: false,
+            getInitialDataCalled: !!initialDataReceived,
```
stevegcoder, line 47: Is there a way to infer getInitialDataCalled based on the data available?

src/elements/content-sharing/ContentSharing.js
```diff
@@ -0,0 +1,171 @@
...
+        const getItemSuccess = (itemData: ContentSharingItemAPIResponse) => {
+            const { item: itemFromAPI, sharedLink: sharedLinkFromAPI } = convertItemResponse(itemData);
+            setItem(itemFromAPI);
+            setSharedLink(sharedLinkFromAPI);
+            setErrorExists(false);
+        };
+
+        const getItem = async () => {
+            if (itemType === TYPE_FILE) {
+                await api.getFileAPI().getFile(itemID, getItemSuccess, getError, {
```
tjgupta, line 85: Perhaps a nitpick, but it looks like `TYPE_FILE` and `TYPE_FOLDER` do very similar things. Would it make sense to normalize this somehow, either by creating some kind of `getItem()` or at least saving the `fields` array as a variable?
```
let fetchItem;
if (itemType === TYPE_FILE) {
  fetchItem = api.getFileAPI().getFile;
} else if (itemType === TYPE_FOLDER) {
  fetchItem = api.getFolderAPI().getFolderFields;
}
```

src/elements/content-sharing/ContentSharing.js
```diff
@@ -0,0 +1,171 @@
...
+        id: 'be.contentSharing.loadingError',
+    },
+});
+
+function ContentSharing(props: ContentSharingProps) {
+    const { apiHost, itemID, itemType, language, token } = props;
+    const [item, setItem] = useState<?itemFlowType>(null);
+    const [sharedLink, setSharedLink] = useState<?ContentSharingSharedLinkType>(null);
+    const [currentUserID, setCurrentUserID] = useState<?string>(null);
+    const [errorExists, setErrorExists] = useState<boolean>(false);
```
tjgupta, line 56: This might be a bit more robust if we store the error itself rather than a boolean for `errorExists`. This way in the future if we ever might display different error messages this won't need to be redone. Is that a possibility?

src/elements/content-sharing/ContentSharing.js
sahiga: Yes, good catch. I removed the async/await wrappers from the API calls.

src/features/unified-share-modal/UnifiedShareModal.js
sahiga: I originally tried to use `!!item && !!sharedLink`, but that ended up changing a bunch of USM snapshot tests. `componentDidUpdate()` in the USM also checks for `getInitialDataCalled`, `type`, and `typeID`, rather than `item`/`sharedLink`. I'm not sure whether there's a case in which we'd still want to call `getInitialData` even when `item`/`sharedLink` are defined, but I figured it would be best for backwards-compatibility to add a new prop that doesn't affect existing instantiations of the USM.

src/elements/content-sharing/ContentSharing.js
sahiga: For some reason, I ran into errors when I defined `fetchItem` like this, so I moved the `fields` array into a variable.

src/elements/content-sharing/ContentSharing.js
```diff
@@ -0,0 +1,161 @@
...
+                        changeSharedLinkPermissionLevel={() => Promise.resolve([])} // to do: replace with a POST to the Shared Link API
+                        collaboratorsList={{ collaborators: [] }} // to do: replace with Collaborators API
+                        currentUserID={currentUserID}
+                        getCollaboratorContacts={() => Promise.resolve([])} // to do: replace with Collaborators API
+                        getSharedLinkContacts={() => Promise.resolve([])} // to do: replace with Collaborators API
+                        initialDataReceived
+                        item={item}
+                        onAddLink={() => Promise.resolve([])} // to do: replace with a POST to the Shared Link API
+                        sharedLink={sharedLink}
+                        showFormOnly
```
tjgupta, line 149: Should this be passing in `props.showFormOnly`? Also, I don't know if this is a super intuitive sounding name to me... but maybe just me? Does it make sense to opt out of the modal, or opt-in? Which is the most common use case? Maybe something like `displayInModal` or `renderInModal` (it could default to true if we'd rather do opt-out)...?

sahiga: I believe the modal is the more common use case. `displayInModal` sounds good to me; I updated the code with the new prop.

src/elements/content-sharing/ContentSharing.js
```diff
@@ -0,0 +1,165 @@
...
+};
+
+function ContentSharing(props: ContentSharingProps) {
+    const { apiHost, itemID, itemType, language, token } = props;
+    const [item, setItem] = useState<?itemFlowType>(null);
+    const [sharedLink, setSharedLink] = useState<?ContentSharingSharedLinkType>(null);
+    const [currentUserID, setCurrentUserID] = useState<?string>(null);
+    const [errorMessage, setErrorMessage] = useState<?Object>(null);
+
+    const api = new API({
```
jstoffan, line 53: The API instance will be recreated every time the component re-renders. Does the prior instance need to be cleaned up? If so, should the instance be stored in a ref?

src/constants.js
```diff
@@ -187,6 +194,7 @@ export const CLIENT_NAME_CONTENT_EXPLORER = 'ContentExplorer';
 export const CLIENT_NAME_CONTENT_PREVIEW = 'ContentPreview';
 export const CLIENT_NAME_CONTENT_SIDEBAR = 'ContentSidebar';
 export const CLIENT_NAME_OPEN_WITH = 'ContentOpenWith';
+export const CLIENT_NAME_CONTENT_SHARING: 'ContentSharing' = 'ContentSharing';
```
jstoffan, line 197: Can we keep things alphabetized where it's not burdensome?

src/elements/content-sharing/ContentSharing.js
```diff
@@ -0,0 +1,165 @@
+/**
+ * @flow
+ * @file ContentSharing Element
+ * @author Box
+ */
+import React, { useCallback, useEffect, useState } from 'react';
```
jstoffan, line 6: I've found it easier to mock and spy on these helpers when they're attached to a namespace:

```

React.useState<Whatever | null>(null);
```

src/elements/content-sharing/ContentSharing.js
```diff
@@ -0,0 +1,165 @@
+/**
+ * @flow
```
jstoffan, line 2: Can this be written in TypeScript? We're going to struggle to fully convert if we continue to write new components using Flow.

src/elements/content-sharing/ContentSharing.js
```diff
@@ -0,0 +1,165 @@
...
+    useEffect(() => {
+        const getItemSuccess = (itemData: ContentSharingItemAPIResponse) => {
+            const { item: itemFromAPI, sharedLink: sharedLinkFromAPI } = convertItemResponse(itemData);
+            setItem(itemFromAPI);
+            setSharedLink(sharedLinkFromAPI);
+            setErrorMessage(null);
+        };
+
+        const getItem = () => {
+            const fields = {
```
jstoffan, line 91: Does this need to be redeclared each time its containing function is executed?

src/elements/content-sharing/ContentSharing.js
```diff
@@ -0,0 +1,165 @@
...
+    );
+
+    useEffect(() => {
+        setItem(null);
+        setSharedLink(null);
+        setCurrentUserID(null);
+    }, [token, itemID, itemType]);
+
+    useEffect(() => {
+        const getItemSuccess = (itemData: ContentSharingItemAPIResponse) => {
```
jstoffan, line 83: Can this be moved up to a component-level variable? Should it be named something like `handleGetItemSuccess` to retain its status as a verb?

src/elements/content-sharing/ContentSharing.js
```diff
@@ -0,0 +1,165 @@
...
+            api.getUsersAPI(false).getUser(itemID, getUserSuccess, getError, {
+                fields: [FIELD_ENTERPRISE, FIELD_HOSTNAME],
+            });
+        };
+        if (item && sharedLink && !currentUserID) {
+            getUserData();
+        }
+    }, [api, getError, item, itemID, itemType, sharedLink, currentUserID]);
+
+    const renderElement = () => {
```
jstoffan, line 134: Does this need to be a separate variable that's declared per-render? It seems like the logic within can be inlined at the bottom of the component.

src/elements/content-sharing/__tests__/ContentSharing.test.js
```diff
@@ -0,0 +1,339 @@
...
+    MOCK_ITEM_API_RESPONSE,
+    MOCK_CONVERTED_ITEM_DATA,
+    MOCK_CONVERTED_USER_DATA,
+    MOCK_SHARED_LINK,
+    MOCK_SHARED_LINK_DATA_AFTER_NORMALIZATION,
+    MOCK_USER_API_RESPONSE,
+} from '../../../features/unified-share-modal/utils/__mocks__/USMMocks';
+
+jest.mock('../../../api');
+jest.mock('../../../features/unified-share-modal/utils/convertData');
```
jstoffan, line 22: It seems like the entire API and USM component modules should be mocked.

src/elements/content-sharing/__tests__/ContentSharing.test.js
```diff
@@ -0,0 +1,339 @@
...
+    });
+
+    afterEach(() => {
+        jest.resetModules();
+        jest.restoreAllMocks();
+    });
+
+    describe('with successful API calls', () => {
+        test('should call getFileAPI().getFile() if itemType is "file"', async () => {
+            const getFile = jest.fn().mockImplementation((id, successFn) => {
```
jstoffan, line 40: These mock implementations are duplicated a few times for various methods. Can we mock the methods at the API module level, instead?

src/elements/content-sharing/__tests__/ContentSharing.test.js
```diff
@@ -0,0 +1,339 @@
...
+            await act(async () => {
+                wrapper = getWrapper({ itemType: TYPE_FILE });
+            });
+
+            wrapper.update();
+            const usm = wrapper.find(UnifiedShareModal);
+
+            expect(getFile).toHaveBeenCalled();
+            expect(convertItemResponse).toHaveBeenCalledWith(MOCK_ITEM_API_RESPONSE);
+            expect(usm.length).toBe(1);
```
jstoffan, line 60: Can be replaced with `wrapper.exists(UnifiedShareModal)`, if desired. Since we have other assertions that rely on it existing, though, it seems like this isn't necessary.

src/elements/content-sharing/messages.js
```diff
@@ -0,0 +1,34 @@
...
+        id: 'be.contentSharing.noAccessError',
+    },
+    notFoundError: {
+        defaultMessage: 'Could not find shared link for this item.',
+        description: 'Message that appears when the item for the ContentSharing Element cannot be found.',
+        id: 'be.contentSharing.notFoundError',
+    },
+});
+
+const CONTENT_SHARING_ERRORS = {
```
jstoffan, line 26: Should we move these to a more general location outside of the messages module?

src/elements/content-sharing/types.js
```diff
@@ -0,0 +1,82 @@
...
+    serverURL: string, // SLS
+};
+
+export type ContentSharingUserDataType = {
+    id: string,
+    userEnterpriseData: ContentSharingEnterpriseDataType,
+};
+
+export type ContentSharingSharedLinkType =
+    | { canInvite: boolean }
```
jstoffan, line 17: Should these type options be aliased to hint at the scenario in which they'd be used?

src/elements/content-sharing/types.js
```diff
@@ -0,0 +1,82 @@
...
+};
+
+export type ContentSharingItemAPIResponse = {
+    allowed_invitee_roles: Array<string>,
+    description: string,
+    etag: string,
+    extension: string,
+    id: string,
+    name: string,
+    permissions: {
```
jstoffan, line 45: Should this share permissions with BoxItem? It appears to be missing some new permissions added to annotations, for example.

src/elements/content-sharing/types.js
```diff
@@ -0,0 +1,82 @@
...
+        can_invite_collaborator: boolean,
+        can_preview: boolean,
+        can_rename: boolean,
+        can_set_share_access: boolean,
+        can_share: boolean,
+        can_upload: boolean,
+        can_view_annotations_all: boolean,
+        can_view_annotations_self: boolean,
+    },
+    shared_link?: {
```
jstoffan, line 59: Duplicated from the SharedLink type?

src/elements/content-sharing/ContentSharing.js
sahiga: Yes, good call. I stored the API in another `useEffect` call that only runs when there are updates to the API props.

src/elements/content-sharing/ContentSharing.js
sahiga: Unfortunately, no, because of the dependencies on Flow types (src/common/types/core.js, src/common/types/api.js, etc.)

src/elements/content-sharing/ContentSharing.js
sahiga: I initially wrote this as a component-level function, but because it alters the component state, I had to wrap it in a `useCallback` hook. The Hooks FAQ recommends placing functions that reference props/state variables inside of effects instead of wrapping them in `useCallback`, so I moved it.

src/elements/content-sharing/messages.js
```diff
@@ -23,12 +23,4 @@ const contentSharingMessages = defineMessages({
...
-const CONTENT_SHARING_ERRORS = {
-    400: 'badRequestError',
-    401: 'noAccessError',
-    403: 'noAccessError',
-    404: 'notFoundError',
-    500: 'loadingError',
-};
-
-export { contentSharingMessages, CONTENT_SHARING_ERRORS };
+export default contentSharingMessages;
```
jstoffan, line 26: The module itself is namespaced by its directory, so this can be `export default messages`

src/elements/content-sharing/types.js
```diff
@@ -1,34 +1,39 @@
...
+// This type is used when an item does not have a shared link.
+type SharedLinkNotCreatedType = { canInvite: boolean };
+
+// This is the full shared link type, which extends the internal USM shared link with
+// data necessary for instantiating the Shared Link Settings modal.
+type SharedLinkCreatedType = USMSharedLinkType & {
+    canChangeDownload: boolean, // SLS
+    canChangePassword: boolean, // SLS
+    canChangeVanityName: boolean, // SLS
+    canInvite: boolean,
```
jstoffan, line 25: Should this type include an intersection with SharedLinkNotCreatedType?

src/elements/content-sharing/constants.js
```diff
@@ -0,0 +1,35 @@
...
+        FIELD_ID,
+        FIELD_NAME,
+        FIELD_PERMISSIONS,
+        FIELD_SHARED_LINK,
+        FIELD_SHARED_LINK_FEATURES,
+        FIELD_ITEM_TYPE,
+    ],
+};
+
+export { CONTENT_SHARING_ERRORS, CONTENT_SHARING_ITEM_FIELDS };
```
jstoffan, line 35: You can also export these inline to avoid maintaining a separate object, `export const CONTENT_SHARING_ITEM_FIELDS = {`, etc.

src/elements/content-sharing/ContentSharing.js
```diff
@@ -3,62 +3,60 @@
...
+        }),
+    );
+    const [item, setItem] = React.useState<itemFlowType | null>(null);
+    const [sharedLink, setSharedLink] = React.useState<ContentSharingSharedLinkType | null>(null);
+    const [currentUserID, setCurrentUserID] = React.useState<string | null>(null);
+    const [errorMessage, setErrorMessage] = React.useState<Object | null>(null);
+
+    React.useEffect(() => {
+        setAPI(
+            new API({
```
jstoffan, line 49: Can we avoid the duplication by having a tiny `createAPI` helper function?

src/elements/content-sharing/ContentSharing.js
```diff
@@ -0,0 +1,151 @@
...
+    apiHost: string,
+    displayInModal?: boolean,
+    itemID: string,
+    itemType: ItemType,
+    language: string,
+    token: string,
+};
+
+function ContentSharing(props: ContentSharingProps) {
+    const { apiHost, displayInModal, itemID, itemType, language, token } = props;
```
jstoffan, line 33: These can be destructured in the function signature, if you want.

src/elements/content-sharing/ContentSharing.js
```diff
@@ -0,0 +1,151 @@
...
+                errorObject = contentSharingMessages[CONTENT_SHARING_ERRORS[error.status]];
+            } else if (error.response && error.response.status) {
+                errorObject = contentSharingMessages[CONTENT_SHARING_ERRORS[error.response.status]];
+            } else {
+                errorObject = contentSharingMessages.loadingError;
+            }
+
+            setErrorMessage(errorObject);
+        },
+        [setErrorMessage],
```
jstoffan, line 71: Should this be `[errorMessage]` so the callback memoization is keyed off the string rather than the setter?

src/elements/content-sharing/ContentSharing.js
sahiga: Yes, I believe so. I tried a couple of things here:
- Moving the logic into the `return` statement (worked, but failed linting because of `no-nested-ternary` rule)
- Inlining the logic with the rest of the component (didn't work; the component returned `null` and didn't re-render after making API calls)

src/elements/content-sharing/__tests__/ContentSharing.test.js
sahiga: Yup, I initially tried to do this as well. Unfortunately, I wasn't able to grab the `props` from the USM when I mocked it.

src/elements/content-sharing/ContentSharing.js
sahiga: That makes more sense to me, too, but the React linter requires `setErrorMessage` (since the function references `setErrorMessage` instead of `errorMessage` directly)

jstoffan:
The Element and API changes look good to me. Nice job!

PR #2194, opened 2020-06-22
feat(content-explorer): custom status to action buttons

feixuanli:
Screen Shot 2020-06-22 at 10 59 36 AM

This PR adds onSelectedClick as optional prop and converts `status` to a clickable button when onSelectedClick is passed along.


src/features/content-explorer/content-explorer-modal/ContentExplorerModal.js
```diff
@@ -14,6 +14,7 @@ type Props = {
     description?: string,
     isOpen?: boolean,
     onRequestClose?: Function,
+    onSelectedClick?: Function,
```
jpfranco, line 17: nit: I know it's not being done above, but we could use more explicit type for this new handler

feixuanli: updated

PR #2196, opened 2020-06-22
fix(pill-selector-dropdown): adds padding back for scrollable dropdowns

danubevictoria:
Fixes padding on top of overlay for pill selector dropdowns and dropdowns in USM.

Before:
image

After:
image

with Suggested Collabs:
image

with only one result:
image

styleguidist:
image

styleguidist pill selector dropdown with `overlayTitle="suggested"` prop for suggested users:
image


booc0mtaco:
@danubevictoria can you confirm the spacing when there is only one item in the result list?

src/components/selector-dropdown/SelectorDropdown.js
```diff
@@ -256,7 +256,7 @@ class SelectorDropdown extends React.Component<Props, State> {
         }
 
         const list = (
-            <ul className="overlay" id={listboxID} role="listbox">
+            <ul className={`overlay ${overlayTitle ? overlayTitle.toLowerCase() : ''}`} id={listboxID} role="listbox">
```
booc0mtaco, line 259: can you use `classNames` library here? also, where does overlayTitle come from?

danubevictoria: will use classNames. overlayTitle comes from PillSelectorDropdown as a prop, which is passed in from ContactSelector in EUA.

src/components/selector-dropdown/SelectorDropdown.js
```diff
@@ -256,7 +256,11 @@ class SelectorDropdown extends React.Component<Props, State> {
         }
 
         const list = (
-            <ul className="overlay" id={listboxID} role="listbox">
+            <ul
+                className={classNames('overlay', overlayTitle ? overlayTitle.toLowerCase() : '')}
```
booc0mtaco, line 260: See [docs](https://www.npmjs.com/package/classnames). you dont wanna include ternary operations when using this function

danubevictoria: since overlayTitle is optional we discussed that the ternary is ok.

PR #2198, opened 2020-06-23
feat(annotations): Handle file versions unavailable

ConradJChan:
When a file version is unavailable, disable the activity link button and show a grayed out message `Version Unavailable`

Screen Shot 2020-06-23 at 3 16 29 PM

### DIFF
--- a/src/elements/content-sidebar/ActivitySidebar.js
+++ b/src/elements/content-sidebar/ActivitySidebar.js
@@ -613,7 +613,7 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
 
         emitAnnotatorActiveChangeEvent(nextActiveAnnotationId);
 
-        if (annotationFileVersionId !== selectedFileVersionId) {
+        if (annotationFileVersionId && annotationFileVersionId !== selectedFileVersionId) {
             history.push(getAnnotationsPath(annotationFileVersionId, nextActiveAnnotationId));
         }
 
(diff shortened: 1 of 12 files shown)


jstoffan:
Should we update the flow types to reflect that `file_version` is nullable now? Any other files need to be updated with guard logic?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -609,6 +606,7 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
             location,
             onAnnotationSelect,
         } = this.props;
+        const annotationFileVersionId = getProp(file_version, 'id');
```
jstoffan, line 609: Can `null` get pushed as part of the history path? Should we do an identity check in the if on line 618?

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.scss
```diff
@@ -46,5 +46,9 @@
     .bcs-AnnotationActivity-link {
         color: $bdl-box-blue;
         font-weight: bold;
+
+        &[aria-disabled='true'] {
+            color: $bdl-gray-62;
```
jstoffan, line 51: Do we need to add any focus/hover rules in here? Buttons can be really picky.

ConradJChan: Design said disabled button that is focusable is 👌

PR #2199, opened 2020-06-24
fix(icon): Tweak region annotation icon

ConradJChan:
Made a slight tweak to the icon viewBox to make it a little more horizontally centered


jstoffan:
Side note, I think the `style` commit type usually refers to code style. This should probably be a `fix`, instead.

PR #2201, opened 2020-06-26
fix(CollapsibleSidebarMenuItem): fix ref hook

krapspark:
Fixing misleading hooks for measuring text overflow that was used in `CollapsibleSidebarMenuItem`


src/utils/dom.js
```diff
@@ -127,24 +127,27 @@ export function scrollIntoView(itemEl: ?HTMLElement, options?: Object = {}): voi
...
  * i.e., if the text is wider than the box around it.
  *
- * @param {{ current: null | typeof undefined | HTMLElement }} contentRef
- * @return {boolean}
+ * @param {void}
+ * @return {{isRefOverflowed: boolean, ref: Function}}
  */
-export function useIsContentOverflowed(contentRef: { current: null | typeof undefined | HTMLElement }): boolean {
-    const [isContentOverflowed, setIsContentOverflowed] = React.useState<boolean>(false);
-    const { current } = contentRef;
```
alexkrolick, line 138: I believe the previous implementation would run after each render (correctly) if this destructure was moved inside the hook and it was not passed as a dep to the function

alexkrolick: The function should also `setIsContentOverflowed(false);` to reset when the condition inverts, which it's not currently doing

alexkrolick:
We'll probably want to move this to use ResiveObserver later so that it can handle the user resizing or otherwise reflowing elements without triggering React state updates.

PR #2204, opened 2020-06-30
feat(radaranimation): add offset prop

DanDeMicco:
Screen Shot 2020-06-30 at 10 13 36 AM

without offset
Screen Shot 2020-06-30 at 10 49 58 AM


DanDeMicco:
@laughnan added screenshot

Issue #557, opened 2018-09-14
Document Uploader:  Drag and Drop does not work in IE


mremigio:
Please fill out the following template so we can reproduce and fix your issue as quickly as possible!

**Environment:**
- Elements version: 
https://cdn01.boxcdn.net/platform/elements/6.0.0/en-US/explorer.js
https://cdn01.boxcdn.net/platform/elements/6.0.0/en-US/explorer.css

**Desktop (please complete the following information):**
 - OS: [e.g. Windows 10]
Windows 10
Internet Explorer (IE)
11.547.16299.0

**Steps to reproduce the problem:**
1. From Content Explorer click on Upload
2. Once in Uploader, drag and drop files using IE (chrome, edge and firefox work)
3. Documents are not added

**What is the expected behavior? (Screenshots can be helpful here)**
Drag and drop via IE should work.  It worked in version 4.3

**What went wrong? (Screenshots, console logs, or HAR files can be helpful here)**
Documents not added when dragged and dropped via IE.  Chrome, Edge and Firefox work with drag and drog.

cmuriyadan:
We are using Box Content explorer plugin in our JS based application and we are also facing the same issue. Request box to take a look at it and let us know if this can fixed

mremigio:
Hi:
Any luck with this ?
Thank you

stevebingsoo:
This problem still exists in UI Elements v11.0.2 and the latest beta v12.  I believe the cause of the problem is this [dropValidator] (https://github.com/box/box-ui-elements/blob/0b233b5aa4e07cfcc3e95f0812dadea8b459ca5f/src/elements/content-uploader/DroppableContent.js#L34)

Internet explorer's dataTransfer.types property does not return a string array like other browsers.  Instead, it returns a DOMStringList.  As a result, the dropValidator function should not use the Array.indexOf function for IE and should instead use the DOMStringList.contains function.

PR #2209, opened 2020-07-01
fix(uploader): fix drag and drop of files in content uploader for IE11

stevebingsoo:
fixes [#557](https://github.com/box/box-ui-elements/issues/557)


jstoffan:
@stevebingsoo, this pull request contains many changes that are unrelated to the uploader element. Can you rebase and force push with only the relevant changes?

src/common/types/core.js
```diff
@@ -406,6 +406,12 @@ type Reply = {
     type: 'reply',
 };
 
+type DOMStringList = {
```
ConradJChan, line 409: Maybe add a comment as to why we need to define this type ourselves

src/elements/content-uploader/ContentUploader.js
```diff
@@ -346,6 +346,58 @@ class ContentUploader extends Component<Props, State> {
...
+    /**
+     * Add DataTransfer Files to the upload queue
+     *
+     * @private
+     * @param {FileList} dataTransferFiles
+     * @param {Function} itemUpdateCallback
+     * @returns {Promise<any>}
+     */
+    addDataTransferFilesToUploadQueue = (dataTransferFiles: FileList, itemUpdateCallback: Function): void => {
+        if (!dataTransferFiles || dataTransferFiles.length === 0) {
```
ConradJChan, line 374: Is it possible that any of the items are folders?

stevebingsoo: based on my research IE11 can't support drag and drop of folders.  Here's one of the resources I found explaining that the DataTransfer.files object is empty if you drop a folder in IE11.  That can be used to detect a folder was dropped but there's no way to access the actual files in the folder.
[cross browser drag and drop notes](https://github.com/leonadler/drag-and-drop-across-browsers)

src/common/types/core.js
stevebingsoo: I've added a comment

src/elements/content-uploader/ContentUploader.js
```diff
@@ -346,6 +346,58 @@ class ContentUploader extends Component<Props, State> {
...
+            return;
+        }
+
+        const fileItems = [];
+
+        for (let idx = 0; idx < dataTransferFiles.length; idx += 1) {
+            fileItems.push(dataTransferFiles.item(idx));
+        }
+
+        this.addFileDataTransferFilesToUploadQueue(fileItems, itemUpdateCallback);
```
ConradJChan, line 384: Can this call `addFilesToUploadQueue` directly with the array of files?

src/elements/content-uploader/ContentUploader.js
ConradJChan: It seems unfortunate that we need to basically duplicate this logic to account for the fact that the `dataTransferFiles` object is represented in a different form. Can we instead normalize the array to some common file shape and then call to a single `addDataTransferItemsToUploadQueue` method?

src/elements/content-uploader/DroppableContent.js
```diff
@@ -31,18 +31,24 @@ const dropDefinition = {
...
-        Array.from(types).some(type => allowedTypes.indexOf(type) > -1),
+    dropValidator: (
+        { allowedTypes }: { allowedTypes: Array<string> },
+        { types }: { types: Array<string> | DOMStringList },
+    ) => {
+        if (types instanceof Array) {
+            return Array.from(types).some(type => allowedTypes.indexOf(type) > -1);
+        }
+
+        const allowedList = allowedTypes.filter(allowed => types.contains(allowed));
```
ConradJChan, line 42: Is this still needed?

stevebingsoo: Yes because the types object is not an array for non-IE browsers and has no indexOf method. For IE11, we have to use DOMStringList and its contains method

src/elements/content-uploader/ContentUploader.js
```diff
@@ -346,6 +346,24 @@ class ContentUploader extends Component<Props, State> {
         );
     };
 
+    /**
+     * Add dropped items to the upload queue
+     *
+     * @private
+     * @param {DataTransfer} droppedItems
+     * @param {Function} itemUpdateCallback
+     * @returns {Promise<any>}
```
swfree, line 355: should this be `@returns {void}`?

PR #2211, opened 2020-07-01
feat(content-sharing): Enable creation of shared links

sahiga:
This PR updates the new ContentSharing Element to enable shared link creation.

image

Additionally, a notification banner appears when a PUT request fails:
image

PR #2214, opened 2020-07-06
feat(content-sharing): Enable shared link deletion

sahiga:
This PR enables the deletion of shared links within the ContentSharing Element. (Follow-up to #2211)


src/elements/content-sharing/__tests__/ContentSharing.test.js
```diff
@@ -373,6 +373,33 @@ describe('elements/content-sharing/ContentSharing', () => {
...
+            });
+            wrapper.update();
+            expect(share).toHaveBeenCalledWith(
+                { id: MOCK_ITEM_ID, permissions: {} },
+                ACCESS_NONE,
+                expect.anything(),
+                expect.anything(),
+                CONTENT_SHARING_SHARED_LINK_UPDATE_PARAMS,
+            );
+            expect(wrapper.find(UnifiedShareModal).prop('sharedLink')).toEqual(MOCK_NULL_SHARED_LINK);
```
mxiao6, line 401: Can this be simplified to `usm.prop('sharedLink')`?

sahiga: That was my initial thought as well. However, `usm` refers to the `UnifiedShareModal` in the wrapper _before_ the update, so we have to retrieve a new reference here.

src/elements/content-sharing/SharingNotification.js
```diff
@@ -6,7 +6,7 @@ import API from '../../api';
 import Notification from '../../components/notification/Notification';
 import NotificationsWrapper from '../../components/notification/NotificationsWrapper';
 import { convertItemResponse } from '../../features/unified-share-modal/utils/convertData';
-import { ACCESS_COLLAB, STATUS_ERROR, TYPE_FILE, TYPE_FOLDER } from '../../constants';
+import { ACCESS_COLLAB, ACCESS_NONE, STATUS_ERROR, TYPE_FILE, TYPE_FOLDER } from '../../constants';
```
booc0mtaco, line 9: i am not sure what ACCESS_NONE means. does it actually correspond to a value returned by the API, or is it more of a placeholder when removing?

src/elements/content-sharing/ContentSharing.js
```diff
@@ -159,7 +161,7 @@ function ContentSharing({ apiHost, displayInModal, itemID, itemType, language, t
                         initialDataReceived
                         item={item}
                         onAddLink={onAddLink}
```
booc0mtaco, line 163: @il10 See here: 
- `onAddLink`
- `onRemoveLink`

src/elements/content-sharing/SharingNotification.js
sahiga: Yup, it's a placeholder. `ShareAccessSelect` also uses `ACCESS_NONE`, so I retained that part of the Item `share()` code to ensure backwards compatibility.

PR #2215, opened 2020-07-07
feat(content-sharing): Enable shared link access level changes

sahiga:
This PR adds shared link access level change functionality to the ContentSharing Element. Available access levels are "open," "company," and "collaborators." (Follow-up to #2211 and #2214; screenshots are in #2211)


src/api/Item.js
```diff
@@ -283,20 +289,22 @@ class Item extends Base {
...
      * @param {boolean|void} [options.forceFetch] - Optionally bypasse the cache
      * @param {boolean|void} [options.refreshCache] - Optionally update the cache
+     * @param {SharedLinkUpdateType} [sharedLinkRequestBody] - Optional custom request body for updating the shared link
      * @return {Promise<void>}
      */
     async share(
         item: BoxItem,
-        access: string,
+        access: ?string,
         successCallback: Function,
```
jstoffan, line 304: It's unusual to have required parameters follow optional parameters. Can we make it so that `access` is always defined?

src/features/unified-share-modal/utils/convertData.js
```diff
@@ -74,20 +55,16 @@ const convertItemResponse = (itemAPIData: ContentSharingItemAPIResponse): Conten
...
         const canChangeVanityName = canChangeAccessLevel && isVanityNameAvailable;
 
         sharedLink = {
             accessLevel,
-            allowedAccessLevels: {
-                peopleInThisItem: accessLevel === PEOPLE_IN_ITEM,
-                peopleInYourCompany: accessLevel === ANYONE_IN_COMPANY,
-                peopleWithTheLink: accessLevel === ANYONE_WITH_LINK,
-            },
+            allowedAccessLevels: ALLOWED_ACCESS_LEVELS,
```
swfree, line 67: Just confirming based on the comment in constants.js - these default to `true` but may be updated in the future?

src/features/unified-share-modal/constants.js
```diff
@@ -1,13 +1,21 @@
 // @flow
+import {
+    ACCESS_COLLAB,
+    ACCESS_COMPANY,
+    ACCESS_OPEN,
+    PERMISSION_CAN_DOWNLOAD,
+    PERMISSION_CAN_PREVIEW,
+    ACCESS_NONE,
```
swfree, line 8: minor: alphasort?

src/common/types/core.js
```diff
@@ -269,6 +272,21 @@ type SharedLink = {
     vanity_url?: string,
 };
 
+type SharedLinkUpdateType = {
```
jstoffan, line 275: This seems like it's been duplicated from the `SharedLink` type with all optional parameters. Should we use `$Shape<SharedLink>`, instead?

src/features/unified-share-modal/constants.js
```diff
@@ -23,21 +31,61 @@ const COLLAB_GROUP_TYPE = 'group';
 const COLLAB_USER_TYPE = 'user';
 const COLLAB_PENDING_TYPE = 'pending';
 
+/**
+ * The following constants are used for converting API requests
+ * and responses into objects expected by the USM, and vice versa
+ */
+
+const API_TO_USM_ACCESS_LEVEL_MAP = {
+    [ACCESS_COLLAB]: PEOPLE_IN_ITEM,
```
booc0mtaco, line 40: nit: I know you created some data conversion functions in previous commits. Perhaps instead of mapping here, we should define the API constants only, and keep the conversions together by putting the maps with the converter functions?

src/common/types/core.js
```diff
@@ -254,12 +255,14 @@ type FileRepresentationResponse = {
     entries: Array<FileRepresentation>,
 };
 
+type SharedLinkPermission = typeof PERMISSION_CAN_DOWNLOAD | typeof PERMISSION_CAN_PREVIEW;
```
jstoffan, line 258: This can stay inline if we use the `$Shape` mapped type as mentioned below.

src/api/Item.js
```diff
@@ -283,20 +289,22 @@ class Item extends Base {
...
      *
      * @param {Object} item - Item to share
-     * @param {string} access - Shared access level
+     * @param {string|undefined} access - Shared access level; potentially undefined if there are no access changes
      * @param {Function} successCallback - Success callback
      * @param {Function|void} errorCallback - Error callback
      * @param {Array<string>|void} [options.fields] - Optionally include specific fields
      * @param {boolean|void} [options.forceFetch] - Optionally bypasse the cache
      * @param {boolean|void} [options.refreshCache] - Optionally update the cache
+     * @param {SharedLinkUpdateType} [sharedLinkRequestBody] - Optional custom request body for updating the shared link
```
jstoffan, line 298: What is the purpose of this parameter? It doesn't appear to be used currently and seems to conflict with the `access` parameter. Should `share` be broken up into multiple methods to better support CRUD?

sahiga: Good catch. I meant to add this in my next PR. Will remove from this PR.

src/features/unified-share-modal/utils/convertData.js
sahiga: Yup, exactly. This data is currently not available in the API, so we will be showing all access levels by default.

sahiga:
@manishPh @swfree @booc0mtaco @jstoffan Thanks for reviewing! Could you take another look?

PR #2219, opened 2020-07-08
feat(content-sharing): Enable permissions level changes

sahiga:
This PR enables shared link permissions level changes in the ContentSharing Element. (Follow-up to #2211, #2214, and #2215)


src/features/unified-share-modal/utils/convertData.js
```diff
@@ -164,3 +164,19 @@ export const convertUserResponse = (userAPIData: User): ContentSharingUserDataTy
         },
     };
 };
+
+/**
+ * Create a shared link permissions object for the API based on a USM permission level.
+ * @param {string} newSharedLinkPermissionLevel
+ */
+export const createSharedLinkPermissions = (newSharedLinkPermissionLevel: string): $Shape<BoxItemPermission> => {
```
booc0mtaco, line 172: `convert*` since it maps a-to-b for a given permission level or no?

src/api/Item.js
```diff
@@ -279,6 +285,27 @@ class Item extends Base {
         }
     };
 
+    /**
+     * Validate an item update request
+     *
+     * @param {string|void} itemID - ID of item to share
+     * @param {BoxItemPermission|void} itemPermissions - Permissions for item
+     * @return {Error|null}
```
booc0mtaco, line 293: I am not sure if returning `null` here makes sense when validated successfully. I think `true` may make more sense, since the specified item is indeed valid. 

Perhaps it should `throw` when invalid, using the values from those functions. Conforming to existing patters should override this suggestion.

sahiga: @booc0mtaco I updated the code to reuse the `catch` blocks for invalid requests. Could you review again? 🙏

src/api/Item.js
```diff
@@ -302,37 +329,68 @@ class Item extends Base {
...
+            this.errorCallback = errorCallback;
 
-        const { can_share, can_set_share_access }: BoxItemPermission = permissions;
-        if (!can_share || !can_set_share_access) {
-            errorCallback(getBadPermissionsError(), this.errorCode);
-            return Promise.reject();
-        }
+            this.validateRequest(id, permissions);
 
-        const cache: APICache = this.getCache();
```
sahiga, line 318: Removing the cache from `share()` and `updateSharedLink()` because we should always make an API call for creating, removing, and updating shared links, even if the item is already cached. (Also, this breaks the sharing modals in ContentExplorer and ContentPicker.)

src/api/Item.js
```diff
@@ -302,37 +327,66 @@ class Item extends Base {
...
+            const { fields } = options;
+            const requestData: RequestData = {
+                url: this.getUrl(this.id),
+                data: {
+                    shared_link: access === ACCESS_NONE ? null : { access },
+                },
+            };
+            if (fields) {
+                requestData.params = { fields: fields.toString() };
+            }
```
mxiao6, line 347: Maybe can use `Object.assign` to get rid of `if`, like `Object.assign(requestData, fields && { params: {fields: fields.toString()}})`

src/features/unified-share-modal/utils/convertData.js
```diff
@@ -164,3 +164,19 @@ export const convertUserResponse = (userAPIData: User): ContentSharingUserDataTy
...
     };
 };
+
+/**
+ * Create a shared link permissions object for the API based on a USM permission level.
+ * @param {string} newSharedLinkPermissionLevel
+ */
+export const convertSharedLinkPermissions = (newSharedLinkPermissionLevel: string): $Shape<BoxItemPermission> => {
+    const sharedLinkPermissions = {};
+    Object.keys(USM_TO_API_PERMISSION_LEVEL_MAP).forEach(level => {
```
booc0mtaco, line 174: i keep thinking there is a way to optimize this but would need to dig a bit deeper into how exactly.

PR #2223, opened 2020-07-10
fix(activityfeed): 16px margin around version activity

ConradJChan:
Before:
Screen Shot 2020-07-09 at 4 47 50 PM

After:
Screen Shot 2020-07-10 at 1 09 57 PM

### DIFF
--- a/src/elements/content-sidebar/activity-feed/version/Version.scss
+++ b/src/elements/content-sidebar/activity-feed/version/Version.scss
@@ -6,8 +6,6 @@
         align-items: center;
         justify-content: space-between;
         width: 100%;
-        margin-top: $bdl-grid-unit * 4;
-        margin-bottom: $bdl-grid-unit * 4;
         padding: 9px;
         color: $bdl-green-light;
         background-color: fade-out($bdl-green-light, .9);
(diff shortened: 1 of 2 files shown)


src/elements/content-sidebar/activity-feed/version/Version.scss
```diff
@@ -6,7 +6,8 @@
         align-items: center;
         justify-content: space-between;
         width: 100%;
-        margin-top: 10px;
+        margin-top: $bdl-grid-unit * 4;
+        margin-bottom: $bdl-grid-unit * 4;
         padding: 9px;
```
booc0mtaco, line 11: not part of your PR obvs, but this pixel value is a bit suspicious

src/elements/content-sidebar/activity-feed/version/Version.scss
```diff
@@ -6,7 +6,8 @@
         align-items: center;
         justify-content: space-between;
         width: 100%;
-        margin-top: 10px;
+        margin-top: $bdl-grid-unit * 4;
+        margin-bottom: $bdl-grid-unit * 4;
```
jstoffan, line 10: It's usually safer to apply padding/margin to only the top/left of items in a list. Does this create extra spacing if the version is the last item in the list, for example?

src/elements/content-sidebar/activity-feed/version/Version.scss
ConradJChan: Yeah that's fair, can try to address it in a follow up

PR #2225, opened 2020-07-11
feat(thumbnail-card): Add actionItem to ThumbnailCard

ChenCodes:
Changes in this PR:
- add actionItem prop to ThumbnailCard
- fixed small CSS issue when the card is selected & hovered (should only show blue border on hover, not selection via tabbing)

**BUI Example**
Screen Shot 2020-07-11 at 8 48 00 AM

**Parent App Example**
Screen Shot 2020-07-11 at 8 47 50 AM


src/components/thumbnail-card/ThumbnailCard.scss
```diff
@@ -15,8 +15,7 @@
 
 // We should apply border to the card only when it is hovered
 .thumbnail-card.is-highlight-applied {
-    &:focus,
```
ChenCodes, line 18: Should not apply blue border when the card is focused because there is a little janky UI when there is the orange focus border on to of the blue border.

src/components/thumbnail-card/ThumbnailCardDetails.js
```diff
@@ -2,17 +2,21 @@
...
 };
 
-const ThumbnailCardDetails = ({ icon, subtitle, title }: Props) => (
+const ThumbnailCardDetails = ({ actionItem, icon, subtitle, title }: Props) => (
     <div className="thumbnail-card-details">
         {icon}
         <div className="thumbnail-card-details-content">
-            <div className="thumbnail-card-title">{title}</div>
-            {subtitle && <div className="thumbnail-card-subtitle">{subtitle}</div>}
+            <div className="thumbnail-card-details-body">
```
ChenCodes, line 15: Open to suggestions for the naming of this div. I was trying to create a separation between the title + subtitle and the actionItem. Body -> title + subtitle

examples/src/ThumbnailCardExamples.js
```diff
@@ -43,6 +45,11 @@ const ThumbnailCardExamples = () => {
                 <br />
                 <ThumbnailCard highlightOnHover thumbnail={thumbnail} title={title} />
             </div>
+            <div>
+                <h1>Action Item Example</h1>
+                <br />
```
sahiga, line 50: Is this line break necessary?

src/components/thumbnail-card/ThumbnailCard.js
```diff
@@ -8,6 +8,7 @@ import ThumbnailCardThumbnail from './ThumbnailCardThumbnail';
 import './ThumbnailCard.scss';
 
 type Props = {
+    actionItem?: React.Node,
```
sahiga, line 11: Could this type be more specific, e.g., `React.Element`?

src/components/thumbnail-card/ThumbnailCard.scss
sahiga: I'd recommend creating a Storybook story and image snapshot for this component, to ensure that there are no visual regressions.

src/components/thumbnail-card/ThumbnailCardDetails.js
```diff
@@ -2,17 +2,21 @@
 import * as React from 'react';
 
 type Props = {
+    actionItem?: React.Node,
```
sahiga, line 5: Same comment about using a more specific type here

src/components/thumbnail-card/ThumbnailCardDetails.js
sahiga: New CSS classes should use the SUIT naming convention: https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md. So, this could be `ThumbnailCardDetails-bodyText`.

examples/src/ThumbnailCardExamples.js
ChenCodes: That was put there to provide a bit of spacing but I am cool with removing it here and in the other places above.

src/components/thumbnail-card/ThumbnailCard.js
ChenCodes: Good call out!

src/components/thumbnail-card/ThumbnailCardDetails.js
ChenCodes: Thanks for the callout!

examples/src/ThumbnailCardExamples.js
sahiga: Since the effect is purely visual, you could replace the line break with an inline style on the `h1` element.

PR #2226, opened 2020-07-13
feat(content-sharing): Display collaborators

sahiga:
Display collaborators in the ContentSharing Element. Avatars will be added in a follow-up PR.

content-sharing-collabs


src/elements/content-sharing/messages.js
```diff
@@ -26,6 +26,11 @@ const messages = defineMessages({
         description: 'Message that appears when the shared link in the ContentSharing Element cannot be updated.',
         id: 'be.contentSharing.sharedLinkUpdateError',
     },
+    collaboratorsLoadingError: {
+        defaultMessage: 'Could not retrieve collaborators for this item.',
+        description: 'Message that appears when the shared link in the ContentSharing Element cannot be updated.',
```
swfree, line 31: should this description be something like "...when collaborators cannot be retrieved"?

src/features/unified-share-modal/utils/convertData.js
```diff
@@ -180,3 +184,52 @@ export const convertSharedLinkPermissions = (newSharedLinkPermissionLevel: strin
     });
     return sharedLinkPermissions;
 };
+
+/**
+ * Convert a response from the Item Collaborations API into the object that the USM expects.
+ * @param {Collaborations} collabsAPIData
```
swfree, line 190: should the other params be added here, `ownerEmail` and `isCurrentUserOwner`?

src/elements/content-sharing/messages.js
sahiga: Good catch! Fixed.

src/elements/content-sharing/SharingNotification.js
```diff
@@ -167,6 +179,66 @@ function SharingNotification({
...
+        createNotification,
+        currentUserID,
+        itemID,
+        itemType,
+        notificationID,
+        notifications,
+        ownerEmail,
+        ownerID,
+        retrievedCollaborators,
+        setCollaboratorsList,
```
ConradJChan, line 239: Does this effect need to re-run based on all of these values?

src/features/unified-share-modal/flowTypes.js
```diff
@@ -155,7 +158,7 @@ export type sharedLinkType = {
 };
 
 export type collaboratorType = {
-    collabID: number,
+    collabID: number | string,
```
ConradJChan, line 161: This doesn't seem ideal -- is there a way to keep it as one type?

src/features/unified-share-modal/utils/convertData.js
```diff
@@ -180,3 +184,54 @@ export const convertSharedLinkPermissions = (newSharedLinkPermissionLevel: strin
...
+export const convertCollabsResponse = (
+    collabsAPIData: Collaborations,
+    ownerEmail: ?string,
+    isCurrentUserOwner: boolean,
+): collaboratorsListType => {
+    let collaborators = [];
+
+    const { entries } = collabsAPIData;
+
+    if (entries.length) {
```
ConradJChan, line 203: Could we reverse this to be a guard clause and reduce the indentation?

src/features/unified-share-modal/utils/convertData.js
```diff
@@ -180,3 +184,54 @@ export const convertSharedLinkPermissions = (newSharedLinkPermissionLevel: strin
...
+ * @param {boolean} isCurrentUserOwner
+ */
+export const convertCollabsResponse = (
+    collabsAPIData: Collaborations,
+    ownerEmail: ?string,
+    isCurrentUserOwner: boolean,
+): collaboratorsListType => {
+    let collaborators = [];
+
+    const { entries } = collabsAPIData;
```
ConradJChan, line 201: I've heard that convention is to have `const` before `let`

src/features/unified-share-modal/utils/convertData.js
```diff
@@ -180,3 +184,54 @@ export const convertSharedLinkPermissions = (newSharedLinkPermissionLevel: strin
...
+
+    if (entries.length) {
+        const ownerEmailDomain = ownerEmail ? ownerEmail.split('@')[1] : null;
```
ConradJChan, line 204: Is there any chance that the email is not well formed? i.e. doesn't have `@`

src/api/ItemCollaborations.js
```diff
@@ -0,0 +1,64 @@
...
+class ItemCollaborations extends MarkerBasedAPI {
+    /**
+     * API URL for collaborations
+     *
+     * @param {string} id - Item id
+     * @protected
+     * @return {string} Base url for collaborations
+     */
+    getUrl(id: string): string {
+        return `getUrl(${id}) should be overriden`;
```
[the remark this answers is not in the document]
sahiga: This follows the Item API pattern, so I think it's okay to return a string here. The Base API throws, but has to implement the `no-unused-vars` ESLint rule to bypass the provided `id` argument.

src/elements/content-sharing/SharingNotification.js
sahiga: Yup, this is actually an optimization. React will only run this effect if any of the variables in the dependency array has changed; otherwise, it will skip the effect. (Also, the React linter throws an error if I try to remove any of the dependencies)

src/features/unified-share-modal/flowTypes.js
sahiga: I agree. Unfortunately, the API uses string IDs, and the internal USM uses number IDs, so we have to support both :(

src/features/unified-share-modal/utils/convertData.js
sahiga: Should be fine, since the emails are saved directly from API responses. I added a regex check to this line, just in case.

src/features/unified-share-modal/flowTypes.js
sahiga: On second thought, I'll convert the IDs to numbers in `convertCollabsResponse()`. I can convert them back to strings in future API requests, if necessary.

sahiga:
Thanks for reviewing, @swfree and @ConradJChan! Could you take another look?

PR #2228, opened 2020-07-14
feat(targeting): update targeting feature

DanDeMicco:
BREAKING CHANGE: several files have been added/removed in the features/targeting folder and targetingApi type has been updated


src/features/targeting/useMessage.js
```diff
@@ -1,36 +1,48 @@
...
-            setMessageStateMap({ ...messageStateMap, [name]: MESSAGE_STATES.SHOWING });
-            markMessageAsSeen(eligibleMessageIDMap[name]);
+        function onShow() {
+            if (canShow && !messageStateMap[name]) {
+                setMessageStateMap({ ...messageStateMap, [name]: MESSAGE_STATES.SHOWING });
+                // FIXME markMessageAsSeen action was currently throttled to prevent we make multiple
+                // backend call if onShow is called multiple times before setMessageStateMap
+                // actually alter the messageState. But it is preferrable to prevent
+                // markMessageAsSeen from being called multiple times instead of throttling
+                // after it is called.
```
JohnnyB83, line 25: Is this something that can be fixed once we implement more of the state machine?

DanDeMicco: Yeah, actually the useShowOne hook will alleviate this so its not a big deal for now but we should figure out a better way eventually

azuruce: hi, we internally discussed this. The "shown" state should be a closure state but not react state. Once we fix that way, the bugs will be all gone @JohnnyB83 , however, it is not urgent.

src/features/targeting/alwaysTargeted.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+import type { TargetingApi } from './types';
+
+/**
+ * Return a targetingApi that always targeting
```
jasminevle, line 5:
```suggestion
 * Return a targetingApi that is always targeting
```

src/features/targeting/neverTargeted.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+import type { TargetingApi } from './types';
+
+/**
+ * Return a targetingApi that always targeting
```
jasminevle, line 5:
```suggestion
 * Return a targetingApi that is never targeting
```

PR #2233, opened 2020-07-15
refactor(annotations): Separate mouse event handlers for links

jstoffan:
Sorry for the multiple PRs, but I wanted to avoid `onMouseDown` managing the invocation of the `onClick` callback.

### DIFF
--- a/src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityLink.tsx
+++ b/src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityLink.tsx
@@ -17,28 +17,33 @@ const AnnotationActivityLink = ({
     message,
     onClick = noop,
 }: AnnotationActivityLinkProps): JSX.Element => {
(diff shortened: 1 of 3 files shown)


src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityLink.tsx
```diff
@@ -17,28 +17,33 @@ const AnnotationActivityLink = ({
     message,
     onClick = noop,
 }: AnnotationActivityLinkProps): JSX.Element => {
+    const handleClick = (event: React.SyntheticEvent<HTMLButtonElement>) => {
```
ConradJChan, line 20: Should this also have a `if (isDisabled) { return; }` guard clause?

jstoffan: The behavior for clicks when disabled is handled by the PlainButton component. This is pretty much how the logic was before.

PR #2234, opened 2020-07-15
feat(content-explorer): add optional prop to content explorer modal

feixuanli:
This PR added onSelectItem as optional prop to ContentExplorerModalContainer,  ContentExplorerModal, and pass to ContentExplorer as select handler for the purpose of tracking selected items on upstream modal components.


src/features/content-explorer/content-explorer-modal-container/ContentExplorerModalContainer.js
```diff
@@ -37,6 +37,13 @@ class ContentExplorerModalContainer extends Component {
         /** Folder is in the process of being created. */
         onSelectedClick: PropTypes.func,
         /** Called when selected button is clicked */
+        onSelectItem: PropTypes.func,
+        /**
```
jpfranco, line 41: Looks like this comment should be above the prop. Looks like some other pre-existing comments are also misplaced. I think this happens when auto-formatting re-orders the props alphabetically but it doesn't move the comments

feixuanli: Thanks for catching this! Updated.

PR #2235, opened 2020-07-15
feat(content-sharing): Display contacts in collaborator form

sahiga:
Display contacts in the "Invite people" form in the ContentSharing Element.

Filtered contacts:
content-sharing-display-users-autofill

List of roles for the new collaborator:
content-sharing-display-users-roles


src/api/Users.js
```diff
@@ -36,6 +36,19 @@ class Users extends Base {
...
+     * API URL for fetching all users in the current user's enterprise
+     *
+     * @param {string} filterTerm Filter for enterprise users
+     * @return {string} URL for fetching enterprise users
+     */
+    getUsersInEnterpriseUrl(filterTerm: string): string {
+        if (!filterTerm) {
+            throw new Error('Missing filter term');
+        }
+        return `${this.getBaseApiUrl()}/users?filter_term=${filterTerm}`;
```
booc0mtaco, line 49: Perhaps we need go get in the habit of using [`query-string`](https://www.npmjs.com/package/query-string) to handle these cases. this will automatically handled escaping of `filterTerm` where needed (and if desired).

We already have this package as a dependency.

src/features/unified-share-modal/utils/convertData.js
```diff
@@ -233,3 +233,27 @@ export const convertCollabsResponse = (
...
+ * @param {UserCollection} contactsAPIData
+ * @param {string|null} currentUserID
+ */
+export const convertContactsResponse = (
+    contactsAPIData: UserCollection,
+    currentUserID: string | null,
+): Array<contactType> => {
+    const { entries } = contactsAPIData;
+
+    if (!entries || !entries.length) return [];
```
booc0mtaco, line 249: you should be able to apply `.filter` to the return statement below so you avoid having multiple returns in this. More functional that way, and fewer branches to test.

src/api/Users.js
```diff
@@ -36,6 +36,19 @@ class Users extends Base {
...
     }
 
+    /**
+     * API URL for fetching all users in the current user's enterprise
+     *
+     * @param {string} filterTerm Filter for enterprise users
+     * @return {string} URL for fetching enterprise users
+     */
+    getUsersInEnterpriseUrl(filterTerm: string): string {
+        if (!filterTerm) {
```
booc0mtaco, line 46: Hm, after looking at this, looks like we treat `''` as a possible filter term by default on LN123? It that's the case, when using the default value, it will throw?

sahiga: @booc0mtaco Thanks for catching this. I updated the PR based on your comments -- could you take another look?

src/api/Users.js
```diff
@@ -36,6 +36,23 @@ class Users extends Base {
         return `${this.getBaseApiUrl()}/users/${id}/avatar`;
     }
 
+    /**
+     * API URL for fetching all users in the current user's enterprise
+     *
+     * @param {string} [filterTerm] Optional filter for enterprise users
+     * @return {string} URL for fetching enterprise users
```
booc0mtaco, line 43: `@returns`

src/features/unified-share-modal/utils/convertData.js
```diff
@@ -233,3 +233,25 @@ export const convertCollabsResponse = (
 
     return { collaborators };
 };
+
+/**
+ * Convert an enterprise users API response into an array of internal USM contacts.
+ *
+ * @param {UserCollection} contactsAPIData
+ * @param {string|null} currentUserID
+ */
```
booc0mtaco, line 242: `@returns {[contactType]}` or whatever the JSDoc syntax for this is

src/api/Base.js
```diff
@@ -170,11 +170,13 @@ class Base {
      * Generic success handler
      *
      * @param {Object} data - The response data
+     * @returns {Promise|null|undefined}
      */
-    successHandler = (data: any): void => {
+    successHandler = (data: any): Promise<any> | null | typeof undefined => {
```
ConradJChan, line 175: This change to have the success handler return a promise seems not ideal to me. It seems like we should adhere to a callback pattern or a Promise based pattern rather than a hybrid of the two. Is this change necessary?

src/elements/content-sharing/SharingNotification.js
```diff
@@ -239,6 +248,43 @@ function SharingNotification({
         setCollaboratorsList,
     ]);
 
+    // Set the getContacts function, which is used for inviting collaborators in the USM
+    React.useEffect(() => {
```
ConradJChan, line 252: tbh, having these large `useEffect` calls make it a bit hard to read the `SharingNotification` component. Would it make sense to create custom hooks that could be reused by other components and simplify this component? Maybe a `useContacts` and `useCollaborators` hooks?

src/features/unified-share-modal/utils/convertData.js
```diff
@@ -233,3 +233,25 @@ export const convertCollabsResponse = (
...
+export const convertContactsResponse = (
+    contactsAPIData: UserCollection,
+    currentUserID: string | null,
+): Array<contactType> => {
+    const { entries = [] } = contactsAPIData;
+
+    // Return all users except for the current user
+    return entries.reduce((filteredContacts, contact) => {
```
ConradJChan, line 250: This could also be written in a more functional style like `entries.filter(...).map(...)`

src/api/Base.js
sahiga: Yeah, I agree that it's not ideal. `getContactsPromise()` in `ContactsField` (https://github.com/box/box-ui-elements/blob/master/src/features/unified-share-modal/ContactsField.js#L114) uses Promises, but the existing flow uses callbacks. To get the response to show up properly in `ContactsField`, I returned the result of `successHandler()`. Do you know if there are other ways to handle this case?

src/elements/content-sharing/SharingNotification.js
sahiga: Good idea. Will look into refactoring this component.

src/api/Base.js
ConradJChan: Could we do something like what the [Feed.js](https://github.com/box/box-ui-elements/blob/4858ac4bd28c0f7a49b547f300f8145a9b5c01f3/src/api/Feed.js#L1261) does? If `updatedGetContactsFn` needs to return a promise, it could create a `new Promise(...)` and pass that `resolve` into the success and error callbacks.

sahiga:
@booc0mtaco @ConradJChan Thanks for reviewing :) I updated the PR; could you take another look?

src/elements/content-sharing/ContentSharing.js
```diff
@@ -92,7 +99,7 @@ function ContentSharing({ apiHost, displayInModal, itemID, itemType, language, t
         setChangeSharedLinkAccessLevel(null);
         setChangeSharedLinkPermissionLevel(null);
         setCollaboratorsList(null);
-        setCurrentUserID(null);
+        // setCurrentUserID(null);
```
swfree, line 102: just checking this is an intentional change?

src/elements/content-sharing/hooks/useCollaborators.js
```diff
@@ -0,0 +1,74 @@
...
+ * @param {API} api
+ * @param {string} itemID
+ * @param {ItemType} itemType
+ * @param {string} ownerEmail
+ * @param {boolean} isCurrentUserOwner
+ * @param {Function} [handleSuccess]
+ * @param {Function} [handleError]
+ * @returns {collaboratorsListType | null}
+ */
+function useCollaborators(
```
swfree, line 27: now that these hooks are separated out, do we need unit tests for them or are they covered by the tests in ContentSharing.test.js?

sahiga: They are covered by ContentSharing.test.js and SharingNotification.test.js, but I'll add more tests specifically for the hooks.

src/elements/content-sharing/hooks/useCollaborators.js
```diff
@@ -0,0 +1,70 @@
...
+): collaboratorsListType | null {
+    const [collaboratorsList, setCollaboratorsList] = React.useState<collaboratorsListType | null>(null);
+
+    React.useEffect(() => {
+        if (collaboratorsList) return;
+
+        const handleGetCollaborationsSuccess = (response: Collaborations) => {
+            const updatedCollaboratorsList = convertCollabsResponse(response, ownerEmail, isCurrentUserOwner);
+            setCollaboratorsList(updatedCollaboratorsList);
+            if (handleSuccess) {
```
ConradJChan, line 44: could `handle*` be defaulted to `noop` and avoid these if checks?

src/elements/content-sharing/hooks/useContacts.js
```diff
@@ -0,0 +1,54 @@
...
+
+        const updatedGetContactsFn: GetContactsFnType = () => (filterTerm: string) => {
+            return new Promise((resolve: (result: Array<contactType>) => void) => {
+                api.getUsersAPI(false).getUsersInEnterprise(
+                    itemID,
+                    (response: UserCollection) => {
+                        if (handleSuccess) {
+                            handleSuccess();
+                        }
+                        return resolve(convertContactsResponse(response, currentUserID));
```
ConradJChan, line 41: Should this massaging of data be done outside of the hook to keep this more general purpose?

src/elements/content-sharing/hooks/useContacts.js
```diff
@@ -0,0 +1,54 @@
...
+    React.useEffect(() => {
+        if (getContacts) return;
+
+        const updatedGetContactsFn: GetContactsFnType = () => (filterTerm: string) => {
+            return new Promise((resolve: (result: Array<contactType>) => void) => {
+                api.getUsersAPI(false).getUsersInEnterprise(
+                    itemID,
+                    (response: UserCollection) => {
+                        if (handleSuccess) {
+                            handleSuccess();
```
ConradJChan, line 39: Should the `handleSuccess` return the data?

src/elements/content-sharing/SharingNotification.js
```diff
@@ -179,65 +194,23 @@ function SharingNotification({
...
+        null,
+        () => handleError(contentSharingMessages.collaboratorsLoadingError),
+    );
+    setCollaboratorsList(collaboratorsList);
+
+    // Set the getContacts function
+    const getContactsFn: GetContactsFnType | null = useContacts(api, currentUserID, itemID, null, () =>
+        handleError(contentSharingMessages.getContactsError),
+    );
+    setGetContacts(() => getContactsFn);
```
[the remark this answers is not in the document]
sahiga: Yup, I think so. Without the wrapper, the state gets reset automatically and causes an infinite rendering loop.

sahiga:
@swfree @ConradJChan - could you review again? 🙏

src/elements/content-sharing/SharingNotification.js
```diff
@@ -199,16 +202,20 @@ function SharingNotification({
...
         itemType,
-        ownerEmail,
-        currentUserID === ownerID,
-        null,
+        data => convertCollabsResponse(data, ownerEmail, currentUserID === ownerID),
         () => handleError(contentSharingMessages.collaboratorsLoadingError),
     );
-    setCollaboratorsList(collaboratorsList);
+    if (collaboratorsList && !collabsExist) {
+        setCollaboratorsList(collaboratorsList);
```
ConradJChan, line 209: Typically, I haven't seen our `successCallback`s return data (L205) -- could we call `setCollaboratorsList(convertCollabsResponse(...))` here?

src/elements/content-sharing/hooks/useCollaborators.js
```diff
@@ -28,29 +26,21 @@ function useCollaborators(
...
     React.useEffect(() => {
         if (collaboratorsList) return;
 
         const handleGetCollaborationsSuccess = (response: Collaborations) => {
-            const updatedCollaboratorsList = convertCollabsResponse(response, ownerEmail, isCurrentUserOwner);
-            setCollaboratorsList(updatedCollaboratorsList);
-            if (handleSuccess) {
-                handleSuccess();
-            }
+            setCollaboratorsList(handleSuccess(response));
```
ConradJChan, line 38: Should we rely on the `handleSuccess` to return the data? What about something like:
```
setCollaboratorsList(response);
handleSuccess(response);
```

src/elements/content-sharing/SharingNotification.js
```diff
@@ -198,14 +199,15 @@ function SharingNotification({
     ]);
 
     // Set the collaborators list
-    const collaboratorsList: collaboratorsListType | null = useCollaborators(
-        api,
-        itemID,
-        itemType,
-        data => convertCollabsResponse(data, ownerEmail, currentUserID === ownerID),
-        () => handleError(contentSharingMessages.collaboratorsLoadingError),
+    const collaboratorsListFromAPI: Collaborations | null = useCollaborators(api, itemID, itemType, undefined, () =>
```
ConradJChan, line 202: Thoughts on having an options argument for the custom hook instead? That way you can avoid having to pass the `undefined` for the `successCallback`

src/elements/content-sharing/SharingNotification.js
```diff
@@ -198,14 +199,15 @@ function SharingNotification({
...
+        handleError(contentSharingMessages.collaboratorsLoadingError),
     );
-    if (collaboratorsList && !collabsExist) {
+    if (collaboratorsListFromAPI && !collabsExist) {
```
ConradJChan, line 205: Is the `collabsExist` state needed? Is the truthiness of `collaboratorsListFromAPI` insufficient

src/elements/content-sharing/SharingNotification.js
sahiga: Good call. Fixed

src/elements/content-sharing/SharingNotification.js
sahiga: Yeah, this prevents an infinite loop in which `setCollaboratorsList()` keeps getting called. I updated the component to check for the existence of state values from the parent component.

PR #2238, opened 2020-07-17
feat(i18n): Add in localized language name

ehoogerbeets:
- Add in 4 new functions:
    - getDisplayNames: return an array of language names translated to their own language
    - getDisplayNamesWithIds: return an array of language names translated to their own language plus their
      Box ids
    - getLocalizedNames: return an array of language names translated to the current (or given) language
    - getLocalizedNamesWithIds: return an array of language names translated to the current (or given) language, along with
      the Box ids of the language
- Updated cldr-data to v34.4.1 to get the latest language names already pre-sorted by the current language


package.json
```diff
@@ -319,7 +319,7 @@
         "yargs": "^15.1.0"
     },
     "peerDependencies": {
-        "@box/cldr-data": "^34.2.0",
+        "@box/cldr-data": "^34.4.1",
```
j3tan, line 322: technically another breaking change, unless it'll work on the older version

src/utils/__tests__/locales.test.ts
```diff
@@ -0,0 +1,534 @@
...
+            { id: 8, name: 'Русский' },
+            { id: 71, name: 'हिन्दी' },
+            { id: 70, name: 'বাংলা' },
+            { id: 19, name: '日本語' },
+            { id: 6, name: '简体中文' },
+            { id: 63, name: '繁體中文' },
+            { id: 55, name: '한국어' },
+        ]);
+    });
+    test('should return display names with ids in order', () => {
```
j3tan, line 69: line breaks between `test` lines

package.json
ehoogerbeets: With the old version, it will return the right language names but they will come out unsorted. (Or sorted by their English name, more specifically.)

src/utils/__tests__/locales.test.ts
ehoogerbeets: fixed

package.json
ehoogerbeets: @j3tan What do you think? Does that means I should amend a commit to put "breaking change" or not?

greg-in-a-box:
@ehoogerbeets I just wanted to check if these changes are still valid since they were requested back in 2020.

ehoogerbeets:
I think this is still valid, yes. I totally forgot about it!!! Probably needs an update to the latest CLDR though

greg-in-a-box:
i18n logic will be extracted into its own package

PR #2239, opened 2020-07-17
feat(i18n): add a zoneinfo call to get time zones

ehoogerbeets:
- Return the list of time zones that Box recognizes
    - returned as a JS object with values like this:
      ```javascript
      // the Box id, same as below
      "139": {
            // the IANA name of the zone
            name: 'America/Los_Angeles',

            // the above name, localized
            nameLocalized: 'America/Los_Angeles',
 
            // the display name of the zone translated to the
            // current or given locale
            displayName: 'GMT-08:00 America/Los_Angeles PST',

            // the abbreviation used in standard time
            abbreviationStandard: 'PST',

            // the abbreviation used in daylight time
            abbreviationDaylight: 'PDT',

            // whether or not this is an enabled standard Box zone
            enabled: true,

            // The numeric Box backend id of this entry
            id: 139
      }
      ```
- Added findTimeZone to get a particular zone for a given box id
- By default, it only returns the enabled zones now. If you want
all zones anyways, you can give the option "includeZones": "all"
in the options parameter

### DIFF
--- a/src/utils/__tests__/timezones.test.ts
+++ b/src/utils/__tests__/timezones.test.ts
@@ -1,0 +1,159 @@
+/**
+ * load locale data manually that is not the current locale so that
+ * we can use it for unit testing
+ */
+
+import germanLocaleData from '@box/cldr-data/locale-data/de-DE';
+import russianLocaleData from '@box/cldr-data/locale-data/ru-RU';
+import japaneseLocaleData from '@box/cldr-data/locale-data/ja-JP';
+import timezones, { findTimeZone, TimeZone, TimeZoneEntries } from '../timezones';
+
+const germanZoneData = germanLocaleData.timezones;
+const russianZoneData = russianLocaleData.timezones;
+const japaneseZoneData = japaneseLocaleData.timezones;
+
+describe('util/timezones', () => {
+    test('should work in English', () => {
+        const zones: TimeZoneEntries = timezones({
+            includeZones: 'all',
+        }); // default is English
+        expect(zones['1']).toStrictEqual({
+            name: 'Africa/Abidjan',
+            displayName: 'GMT+00:00 Africa/Abidjan GMT',
+            id: 1,
+        });
+        expect(zones['139']).toStrictEqual({
+            name: 'America/Los_Angeles',
+            displayName: 'GMT-08:00 America/Los_Angeles PST',
+            enabled: true,
+            id: 139,
+        });
+    });
+
+    test('should work in German', () => {
+        const zones: TimeZoneEntries = timezones({
+            includeZones: 'all',
+            tzData: germanZoneData,
+        });
+        expect(zones['295']).toStrictEqual({
+            name: 'Asia/Yekaterinburg',
+            displayName: 'GMT+05:00 Jekaterinburg +05',
+            enabled: true,
+            id: 295,
+        });
+        expect(zones['301']).toStrictEqual({
+            name: 'Atlantic/Faeroe',
+            displayName: 'GMT+00:00 Färöer UTC',
+            id: 301,
+        });
+    });
+
+    test('should work in Japanese', () => {
+        const zones: TimeZoneEntries = timezones({
+            includeZones: 'all',
+            tzData: japaneseZoneData,
+        });
+        expect(zones['321']).toStrictEqual({
+            name: 'Australia/Melbourne',
+            displayName: 'GMT+10:00 メルボルン AEST',
+            id: 321,
+        });
+        expect(zones['390']).toStrictEqual({
+            name: 'Europe/Amsterdam',
+            displayName: 'GMT+01:00 アムステルダム CET',
+            enabled: true,
+            id: 390,
+        });
+    });
+
+    test('should work in Russian', () => {
+        const zones: TimeZoneEntries = timezones({
+            includeZones: 'all',
+            tzData: russianZoneData,
+        });
+        expect(zones['295']).toStrictEqual({
+            name: 'Asia/Yekaterinburg',
+            displayName: 'GMT+05:00 Екатеринбург +05',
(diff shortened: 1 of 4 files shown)


src/utils/__tests__/timezones.test.ts
```diff
@@ -0,0 +1,159 @@
...
+            includeZones: 'all',
+        }); // default is English
+        expect(zones['1']).toStrictEqual({
+            name: 'Africa/Abidjan',
+            displayName: 'GMT+00:00 Africa/Abidjan GMT',
+            id: 1,
+        });
+        expect(zones['139']).toStrictEqual({
+            name: 'America/Los_Angeles',
+            displayName: 'GMT-08:00 America/Los_Angeles PST',
```
jeeyi, line 27: What about when we are in daylight saving time? Would it return `GMT-07:00 America/Los Angeles PDT`?

ehoogerbeets: As far as I can tell, this list come from the php `DateTimeZone` class in php from the monolith source file `./modular/i18n/i18n/legacy/lib/Box/Timezone.php`. I think what it is doing is checking today's date to find the current offset for each of these time zones as it constructs the list dynamically. We don't have that functionality available in React/JS unless we somehow successfully integrate ilib which has it.

Some possible solutions:

1. Do the work of finishing the ilib integration, including fixing the ilib webpack plugins to get all of the time zone json files into the webpack. This has footprint implications, and is a heavy-weight solution to a small problem. Plus it may take a sprint or two more, which admin-console cannot wait for.
2. Extract the parts of ilib that do the timezone work into a separate project, and use it box-ui-elements (and ilib too). That way we can avoid the ilib webpack plugin problem. These parts are relatively independent of the rest of ilib, so it is possible, especially if we abandon the idea of multi-calendar-type support. (ie. only support Gregorian). Also, put the tz data into that module as well so that this new code has the knowledge of when and if DST starts and ends for each zone, and we can make a query "what is the offset for this time zone at this datetime". This is also a heavy-weight solution, though ultimately more useful than #1.
3. Punt on the dynamic idea and just make strings like:

`GMT-08:00 PST/-07:00 PDT America/Los_Angeles`

That is, for those zones that use DST, put a joint string. For those string that do not, just put something like:

`GMT-07:00 MST America/Phoenix`

This is easy and quick to do, but is slightly less useful to the user.

ehoogerbeets: Or we can #3 now, and #2 later?

ehoogerbeets: Or, we can add:
4. Extract the parts of ilib that do the timezone work and put them directly into BUIE. This presents no integration problems and we could jettison the non-Gregorian parts of it which simplifies it a bunch, but has the problems that it increases the webpack footprint and would have to be maintained separately from ilib.

ehoogerbeets: 5. Rely on Intl.DateTimeFormat() which can do time zone calculations. Pros: fast, no code from ilib needs to be added to BUIE. Cons: does not work on IE 11. Alternate idea: use strategy #3 above for IE11, and Intl.DateTimeFormat if it is available.

ehoogerbeets: Decided to do strategy number 5 above with the fall back to static strings for IE or any other platform that doesn't support the latest Intl stuff.

yarn.lock
```diff
@@ -3104,10 +3104,10 @@
   resolved "https://registry.yarnpkg.com/@bcoe/v8-coverage/-/v8-coverage-0.2.3.tgz#75a2e8b51cb758a7553d6804a5932d7aace75c39"
   integrity sha512-0hYQ8SB4Db5zvZB4axdMHGwEaQjkZzFjQiN9LVYvIFB2nSUHW9tYpxWriPrWDASIxiaXax83REcLxuSdnGPZtw==
 
-"@box/cldr-data@^34.2.0":
-  version "34.2.0"
-  resolved "https://registry.yarnpkg.com/@box/cldr-data/-/cldr-data-34.2.0.tgz#d2a50e778d28988faaf5e56194b17b46c6b98ae5"
-  integrity sha1-0qUOd40omI+q9eVhlLF7Rsa5iuU=
+"@box/cldr-data@^34.5.0":
+  version "34.5.0"
```
laughnan, line 3108: I know we're doing this in another PR but if this merges first we should include the BREAKING CHANGE piece here.

src/utils/timezones.ts
```diff
@@ -0,0 +1,141 @@
...
+        }
+    } else {
+        return displayName;
+    }
+}
+
+function timezones(options: TimeZoneOptions = {}): TimeZoneEntries {
+    const zones: TimeZoneEntries = {};
+
+    const { includeZones = 'enabled' }: { includeZones?: string } = options;
```
laughnan, line 111: we force it to be `all` in the call and yet here we're setting it to `enabled` as a default. is there any reason for this?

ehoogerbeets: If you mean that we use `all` in the `findTimeZone` function, that's correct. The `findTimeZone` function searches all zones for the one with the given id. The `timezones` function itself (which is the default export) only returns the Box-enabled time zones (ie. the ones that are used in the Box locales) by default. You can get the full list only by passing in `includeZones: "all"`

yarn.lock
ehoogerbeets: I will change the commit msg

ehoogerbeets: Actually, already did that in commit becd6ab

greg-in-a-box:
@ehoogerbeets same with this PR, is it still valid?

greg-in-a-box:
i18n logic will be extracted into its own package

PR #2240, opened 2020-07-17
fix(searchform): update icons with latest assests

indiramuru:
Replacing old search and clear icons in SearchForm with new assets


indiramuru:
@sahiga existing test pass without any update. They wrapper of the icon(button) is being checked so the test passes

PR #2244, opened 2020-07-17
fix(collapsiblesidebar): Fix sidebar footer style and shadow

viodragon2:
### Issue Resolved
- `CollapsibleMenuItem` did not apply correct styling.
- Footer did not apply shadow correctly upon content height change. (Notice how shadow is applied when it scrolls)

sidebarfootershadow-before-1

### Resolution
- Apply `bdl-CollapsibleSidebar-menuItem` to root element of `CollapsibleSidebarMenuItem`
- Handle `onUpdate` (see [react-scrollbars-custom documentation](https://github.com/xobotyi/react-scrollbars-custom#props)) to measure content element height change.

sidebarfootershadow-1


src/features/collapsible-sidebar/CollapsibleSidebarNav.js
```diff
@@ -99,6 +96,14 @@ class CollapsibleSidebarNav extends React.Component<Props, State> {
         this.debouncedTurnOffScrollingState();
     };
 
+    onUpdateHandler = (scrollValues: ScrollState, prevScrollValues: ScrollState) => {
+        const { clientHeight, contentScrollHeight } = scrollValues;
+        const { clientHeight: prevClientHeight, contentScrollHeight: prevContentScrollHeight } = prevScrollValues;
+        if (clientHeight !== prevClientHeight || contentScrollHeight !== prevContentScrollHeight) {
+            this.setScrollShadowState();
+        }
+    };
```
shahzadaziz, line 105: Can we add test coverage for this ?

src/features/collapsible-sidebar/CollapsibleSidebarNav.js
```diff
@@ -69,10 +70,6 @@ class CollapsibleSidebarNav extends React.Component<Props, State> {
         this.setScrollShadowState();
     }
 
-    componentDidUpdate() {
-        this.setScrollShadowState();
```
viodragon2, line 73: `componentDidUpdate` won't execute If the content's height is changed.

There are alternative solutions to `onUpdate`.
- `MutationObserver`: We could observe `this.scrollRef.current.contentElement` and look for attributes changes. However, this is not extensible and requires this component to be aware of children's change. Also, even if we listen to `style` attributes, it cannot detect a height change after an animation (i.e. `transition`)
- `ontransitionend` event handler: `transitionend` event gets fired from the animation or scroll. Though it may be more performant compared to `onUpdate` since it fires less, it does not guarantee to handle height change.

PR #2247, opened 2020-07-21
feat(preview): add ftux tooltip to AdditionalTab

ctuckz:
Add a tooltip wrapper to AdditionalTab to allow optionally displaying an FTUX message for individual items

Screen Shot 2020-07-20 at 7 42 23 PM


src/elements/content-sidebar/additional-tabs/AdditionalTab.js
```diff
@@ -83,6 +86,22 @@ class AdditionalTab extends React.PureComponent<Props, State> {
...
     }
 
+    withFtuxTooltip(body: React.Node) {
+        const { ftuxTooltipData, isLoading } = this.props;
+
+        if (!ftuxTooltipData || isLoading) {
+            return body;
+        }
+
+        const { targetingApi, text } = ftuxTooltipData;
```
ConradJChan, line 96: Could spread this directly onto the `AdditionalTabFtuxTooltip` since the other props are the two destructured

src/elements/content-sidebar/additional-tabs/AdditionalTabFtuxTooltip.js
```diff
@@ -0,0 +1,33 @@
+/**
```
ConradJChan, line 1: Should this be written in typescript?

src/elements/content-sidebar/additional-tabs/AdditionalTabFtuxTooltip.js
```diff
@@ -0,0 +1,33 @@
...
+
+type Props = {
+    children: React.Node,
+    targetingApi: UseTargetingApi,
+    text: string,
+};
+
+const AdditionalTabFtuxTooltip = ({ children, targetingApi, text }: Props) => (
+    <TargetedClickThroughGuideTooltip
+        className="rec-app-ftux-tooltip"
```
ConradJChan, line 23: Can this be named with SUIT style convention?

src/elements/content-sidebar/additional-tabs/__tests__/AdditionalTab.test.js
```diff
@@ -69,4 +69,23 @@ describe('elements/content-sidebar/additional-tabs/AdditionalTab', () => {
...
+            callback: () => {},
+            ftuxTooltipData: {
+                targetingApi: () => {},
+                text: 'ftux text',
+            },
+        };
+
+        const wrapper = getWrapper(props);
+
+        expect(wrapper).toMatchSnapshot();
```
ConradJChan, line 89: Snapshots tend to be pretty brittle, can we do an explicit assertion here on what we expect?

src/elements/content-sidebar/additional-tabs/AdditionalTabs.js
```diff
@@ -19,6 +20,8 @@ type State = {
     isLoading: boolean,
 };
 
+const AdditionalTabWithFtux = withFtuxTooltip(AdditionalTab);
```
ConradJChan, line 23: Could also use the hoc directly in `AdditionalTab`, i.e. `export default withFtuxTooltip(AdditionalTab)`

src/elements/content-sidebar/additional-tabs/AdditionalTabFtuxTooltip.scss
```diff
@@ -1,3 +1,3 @@
-.rec-app-ftux-tooltip.tooltip {
+.bdl-AdditionalTabFtuxTooltip.tooltip {
```
ConradJChan, line 1: Is the `.tooltip` part of the selector needed?

src/elements/content-sidebar/additional-tabs/AdditionalTabFtuxTooltip.js
ConradJChan: Roger that

src/elements/content-sidebar/additional-tabs/__tests__/AdditionalTabFtuxTooltip.test.js
```diff
@@ -0,0 +1,19 @@
...
+        shallow(<AdditionalTabFtuxTooltip {...props}>{children}</AdditionalTabFtuxTooltip>);
+
+    test('should render the tooltip', () => {
+        const children = <div>Child content</div>;
+        const targetingApi = () => {};
+        const text = 'FTUX Text';
+
+        const wrapper = getWrapper({ targetingApi, text }, children);
+
+        expect(wrapper).toMatchSnapshot();
```
jstoffan, line 17: We should try to use explicit assertions rather than snapshots whenever possible.

src/elements/content-sidebar/additional-tabs/AdditionalTabFtuxTooltip.js
```diff
@@ -0,0 +1,33 @@
+/**
+ * @flow
+ * @file Sidebar Additional Tab FTUX tooltip
+ * @author Box
+ */
+
+import * as React from 'react';
+
+import TargetedClickThroughGuideTooltip from '../../../features/targeting/TargetedClickThroughGuideTooltip';
+
```
jstoffan, line 10: Is the whitespace between each import needed? Most of our other files group imports together.

src/elements/content-sidebar/additional-tabs/AdditionalTabs.js
jstoffan: I agree with @ConradJChan. The details of how `AdditionalTab` work should remain hidden from its parent.

src/elements/content-sidebar/additional-tabs/AdditionalTabFtuxTooltip.js
```diff
@@ -0,0 +1,33 @@
...
+import './AdditionalTabFtuxTooltip.scss';
+
+type Props = {
+    children: React.Node,
+    targetingApi: UseTargetingApi,
+    text: string,
+};
+
+const AdditionalTabFtuxTooltip = ({ children, targetingApi, text }: Props) => (
+    <TargetedClickThroughGuideTooltip
```
jstoffan, line 22: Does this component require the host Element to be wrapped in a new context? Does this affect third-party use-cases? If so, it would be a breaking change.

src/elements/content-sidebar/flowTypes.js
```diff
@@ -11,8 +12,14 @@ type NavigateOptions = {
     isToggle?: boolean,
 };
 
+type AdditionalSidebarTabFtuxData = {
```
jstoffan, line 15: Is this unique to this particular component or is it a more generic type?

src/elements/content-sidebar/additional-tabs/withFtuxTooltip.js
```diff
@@ -0,0 +1,33 @@
...
+
+import AdditionalTabFtuxTooltip from './AdditionalTabFtuxTooltip';
+import type { AdditionalSidebarTabFtuxData } from '../flowTypes';
+
+type Props = {
+    ftuxTooltipData?: AdditionalSidebarTabFtuxData,
+    isLoading: boolean,
+};
+
+const withFtuxTooltip = (WrappedComponent: React.ComponentType<any>) => {
```
jstoffan, line 17: If this is only used once (within AdditionalTab), does it need to be a HOC at all? It seems like the logic is relatively simple and could be inlined in AdditionalTab's render logic.

src/elements/content-sidebar/additional-tabs/__tests__/AdditionalTabs.test.js
```diff
@@ -122,4 +122,29 @@ describe('elements/content-sidebar/AdditionalTabs', () => {
...
+                    status: 'ADDED',
+                },
+            ],
+        };
+
+        const wrapper = getWrapper(props);
+        wrapper.setState({ isLoading: false });
+
+        const ftuxTooltipComponent = wrapper.find('AdditionalTabFtuxTooltipComponent');
+        expect(ftuxTooltipComponent.dive().find(AdditionalTabFtuxTooltip)).toHaveLength(1);
```
jstoffan, line 148: You can also use `.exists` for this type of assertion.

src/elements/content-sidebar/additional-tabs/AdditionalTab.js
```diff
@@ -14,11 +14,13 @@ import PlainButton from '../../../components/plain-button/PlainButton';
 import IconEllipsis from '../../../icons/general/IconEllipsis';
 import AdditionalTabPlaceholder from './AdditionalTabPlaceholder';
 import messages from './messages';
-import type { AdditionalSidebarTab } from '../flowTypes';
+
```
jstoffan, line 17: Revert whitespace-only change?

src/elements/content-sidebar/additional-tabs/AdditionalTabFtuxTooltip.scss
```diff
@@ -0,0 +1,3 @@
...
+    max-width: 250px;
```
jstoffan, line 2: This value seems kind of magic number-y. Is it sized relative to something else?

src/elements/content-sidebar/additional-tabs/AdditionalTabFtuxTooltip.scss
ctuckz: Without it, `.bdl-GuideTooltip.bdl-Tooltip` wins out and overrides the max-width

src/elements/content-sidebar/additional-tabs/AdditionalTabFtuxTooltip.js
ctuckz: I was trying to order the imports similarly to how we do it in EUA, but you're right - the other files don't do that. I'll get rid of the whitespace!

src/elements/content-sidebar/flowTypes.js
ctuckz: Specific to this component, but I guess there's no reason why another component doing something similar couldn't use it. The data is passed into the component from the consumer of the sidebar, so anyone else that wants to use it would need to make sure that they're using the same format to pass data in

src/elements/content-sidebar/additional-tabs/withFtuxTooltip.js
ctuckz: I originally had it in AdditionalTab, but @ConradJChan suggested moving it to a HOC. I'm not really an expert on front-end patterns, but I'm happy to put it wherever you think makes the most sense

src/elements/content-sidebar/additional-tabs/AdditionalTabFtuxTooltip.js
ctuckz: If you're asking about the HTML structure, it would, since the element would go from `<AdditionalTab />` to `<Tooltip><AdditionalTab /></Tooltip`. The data being passed into the sidebar wouldn't change though, since all of the data used for this tooltip is optional. Additional tabs aren't available to third parties, so we don't need to worry about breaking anyone

src/elements/content-sidebar/additional-tabs/AdditionalTabFtuxTooltip.scss
ctuckz: The default width was too wide to fit on the right hand side of the sidebar, and we don't have official designs to work from, so I just picked a number that looked good and fit in the sidebar. Is there a better place to get this number from?

jstoffan: There's `$sidebarContentWidth` available as a variable somewhere, but it's larger than what you have here. Maybe just leave a comment explaining the reason for the value?

src/elements/content-sidebar/additional-tabs/withFtuxTooltip.js
jstoffan: It looks like it was in HOC form in the initial commit, just as a method rather than a component. Maybe we could modify `AdditionalTabFtuxTooltip `to accept an `isVisible` prop and return `children` directly if it's false?

src/elements/content-sidebar/additional-tabs/AdditionalTabFtuxTooltip.js
```diff
@@ -0,0 +1,33 @@
...
+        className="bdl-AdditionalTabFtuxTooltip"
+        body={text}
+        position="middle-right"
```
jstoffan, line 25: Does this work if the sidebar is collapsed?

src/elements/content-sidebar/additional-tabs/withFtuxTooltip.js
ctuckz: Ah, I see what you're saying. Doing it in AdditionalTabFtuxTooltip would be pretty straightforward. Do you think it would be confusing to have the component no-op based on the isVisible prop? Or have we followed that pattern with other components?

src/elements/content-sidebar/additional-tabs/AdditionalTabFtuxTooltip.js
ctuckz: Yep! It shifts over to the left side:
Screen Shot 2020-07-21 at 12 00 38 PM

src/elements/content-sidebar/additional-tabs/withFtuxTooltip.js
jstoffan: It's basically what the HOC is doing, right? We just accomplish it with one fewer component. I think simplicity should win out here, especially since none of this is being reused elsewhere.

src/elements/content-sidebar/additional-tabs/AdditionalTabFtuxTooltip.js
```diff
@@ -0,0 +1,37 @@
...
+
+type Props = {
+    children: React.Node,
+    isVisible: boolean,
+    targetingApi?: UseTargetingApi,
+    text?: string,
+};
+
+const AdditionalTabFtuxTooltip = ({ children, isVisible, targetingApi, text }: Props) => {
+    if (isVisible && targetingApi && text) {
```
jstoffan, line 20: Consider converting this to a guard clause:

```
if (!isVisible || !targetingAPI || !text) {
    return children;
}
```

src/elements/content-sidebar/additional-tabs/AdditionalTab.js
```diff
@@ -94,19 +105,22 @@ class AdditionalTab extends React.PureComponent<Props, State> {
...
-                    className={className}
-                    data-testid="additionaltab"
-                    type="button"
-                    isDisabled={isDisabled}
-                    onClick={() => callbackFn({ id, callbackData: rest })}
-                >
-                    {this.getTabIcon()}
-                </PlainButton>
-            </Tooltip>
+            <AdditionalTabFtuxTooltip isVisible={isFtuxTooltipVisible} {...(ftuxTooltipData || {})}>
```
jstoffan, line 111: Consider passing in `ftuxTooltipData` as its own prop.

src/elements/content-sidebar/additional-tabs/AdditionalTab.js
```diff
@@ -94,19 +105,22 @@ class AdditionalTab extends React.PureComponent<Props, State> {
...
-                    className={className}
-                    data-testid="additionaltab"
-                    type="button"
-                    isDisabled={isDisabled}
-                    onClick={() => callbackFn({ id, callbackData: rest })}
-                >
-                    {this.getTabIcon()}
-                </PlainButton>
-            </Tooltip>
+            <AdditionalTabFtuxTooltip ftuxTooltipData={ftuxTooltipData} isVisible={isFtuxTooltipVisible}>
```
laughnan, line 111: Can this render multiple tooltips? The targeted one and then the "default" (`Tooltip`) one?

ctuckz: Yeah, it'll render the default on the left side and the targeted one on the right:
Screen Shot 2020-07-22 at 10 54 24 AM

I thought about suppressing the default one when there's a targeted tooltip, but I think we'd want to keep both for situations where the targeted toolip doesn't say the app name. For example, if we had a targeted tooltip for video conferencing apps that says something like "Working from home? Try out one of our video apps" and shows for either Zoom, Webex, or Hangouts, you'd still want to see the app name when you hover over the icon so you know which app you're launching

laughnan: Yeah I see what you're saying. I wonder if Design/Product have opinions on this. I think it looks like a lot with two tooltips, but I understand that's the only place you can see the application name aside from the logo.

ctuckz: Just updated the PR - we decided to suppress the app name tooltip when the FTUX tooltip is showing

src/elements/content-sidebar/additional-tabs/AdditionalTab.js
```diff
@@ -94,9 +104,14 @@ class AdditionalTab extends React.PureComponent<Props, State> {
...
 
         const tooltipText = isDisabled ? this.getDisabledReason() : title;
+        const isFtuxTooltipVisible = !isLoading && !!ftuxTooltipData;
 
         return (
-            <Tooltip position="middle-left" text={tooltipText}>
+            <AdditionalTabFtuxTooltip
+                defaultTooltipText={tooltipText}
+                ftuxTooltipData={ftuxTooltipData}
+                isFtuxVisible={isFtuxTooltipVisible}
```
jstoffan, line 113: I think we can do `isFtuxVisible={!isLoading}` and check for the existence of `ftuxTooltipData` inside AdditionalTabFtuxTooltip.

src/elements/content-sidebar/additional-tabs/AdditionalTabFtuxTooltip.js
```diff
@@ -0,0 +1,44 @@
...
+import './AdditionalTabFtuxTooltip.scss';
+
+type Props = {
+    children: React.Node,
+    defaultTooltipText: ?string | React.Node,
+    ftuxTooltipData?: AdditionalSidebarTabFtuxData,
+    isFtuxVisible: boolean,
+};
+
+const AdditionalTabFtuxTooltip = ({ children, defaultTooltipText, isFtuxVisible, ftuxTooltipData }: Props) => {
```
jstoffan, line 20: Should we rename this to `AdditionalTabTooltip`, since it now handles both?

src/elements/content-sidebar/additional-tabs/__tests__/AdditionalTabFtuxTooltip.test.js
```diff
@@ -0,0 +1,56 @@
...
+    test('should render the FTUX tooltip when isFtuxVisible is true and the FTUX can be shown', () => {
+        const children = <div>Child content</div>;
+        const targetingApi = () => ({
+            canShow: true,
+        });
+        const text = 'FTUX Text';
+
+        const wrapper = getWrapper({ isFtuxVisible: true, ftuxTooltipData: { targetingApi, text } }, children);
+
+        const ftuxTooltip = wrapper.find('.bdl-AdditionalTabFtuxTooltip');
```
jstoffan, line 20: Can we target this with a `data-testid` attribute or component name? Same for the `.find('div')` and other similar selectors below.

PR #2249, opened 2020-07-22
feat(preview): Add currentFileVersionId in preview options

mxiao6:
The prerequisite of https://github.com/box/box-annotations/pull/542

### DIFF
--- a/src/elements/content-preview/ContentPreview.js
+++ b/src/elements/content-preview/ContentPreview.js
@@ -761,6 +761,7 @@ class ContentPreview extends React.PureComponent<Props, State> {
 
         if (selectedVersion) {
             setProp(fileOpts, [fileId, 'fileVersionId'], selectedVersion.id);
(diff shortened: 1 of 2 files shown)


src/elements/content-preview/ContentPreview.js
```diff
@@ -761,6 +761,11 @@ class ContentPreview extends React.PureComponent<Props, State> {
 
         if (selectedVersion) {
             setProp(fileOpts, [fileId, 'fileVersionId'], selectedVersion.id);
```
jstoffan, line 763: It seems like the logic here is the same as what was in box-annotations, just moved up a level. Why isn't it working as-is?

mxiao6: `bpx-content-preview` fetches file version and modifies `file.file_version` to be the file version that user is viewing. So they're always the same.

jstoffan: Gotcha. Should we pass down the "current" file version id as a separate property rather than the derived boolean value?

PR #2250, opened 2020-07-22
feat(classification): Render classification modified info when provided

patrickpaul:
Adds information on when a classification was last updated when it is provided in the props. Also adds a Restrictions label to Security Controls per Product.

Screen Shot 2020-07-22 at 12 11 14 PM

### DIFF
--- a/src/features/classification/Classification.js
+++ b/src/features/classification/Classification.js
@@ -2,6 +2,7 @@
 import * as React from 'react';
 import { FormattedDate, FormattedMessage } from 'react-intl';
 
+import { isValidDate } from '../../utils/datetime';
 import Label from '../../components/label/Label';
 import LoadingIndicator from '../../components/loading-indicator/LoadingIndicator';
 import ClassifiedBadge from './ClassifiedBadge';
@@ -13,12 +14,6 @@ import type { Controls, ControlsFormat } from './flowTypes';
 
 const STYLE_INLINE: 'inline' = 'inline';
 const STYLE_TOOLTIP: 'tooltip' = 'tooltip';
-
-const formattedDateOptions = {
-    month: 'long',
-    year: 'numeric',
-    day: 'numeric',
-};
 type Props = {
     className?: string,
     color?: string,
@@ -61,10 +56,12 @@ const Classification = ({
     const isControlsIndicatorEnabled = isClassified && isLoadingControls && messageStyle === STYLE_INLINE;
     const isSecurityControlsEnabled =
         isClassified && !isLoadingControls && hasSecurityControls && messageStyle === STYLE_INLINE;
-    const isModifiedMessageVisible = isClassified && hasModifiedAt && hasModifiedBy && messageStyle === STYLE_INLINE;
+    const modifiedDate = new Date(modifiedAt || 0);
+    const isModifiedMessageVisible =
+        isClassified && hasModifiedAt && isValidDate(modifiedDate) && hasModifiedBy && messageStyle === STYLE_INLINE;
 
     const formattedModifiedAt = isModifiedMessageVisible && (
-        <FormattedDate value={new Date(modifiedAt)} {...formattedDateOptions} />
+        <FormattedDate value={modifiedDate} month="long" year="numeric" day="numeric" />
     );
 
     return (
@@ -89,7 +86,7 @@ const Classification = ({
             )}
             {isModifiedMessageVisible && (
                 <Label text={<FormattedMessage {...messages.modifiedByLabel} />}>
-                    <p className="bdl-Classification-modifiedBy">
+                    <p className="bdl-Classification-modifiedBy" data-testid="classification-modifiedby">
                         <FormattedMessage
                             {...messages.modifiedBy}
                             values={{ modifiedAt: formattedModifiedAt, modifiedBy }}
--- a/src/features/classification/Classification.js
+++ b/src/features/classification/Classification.js
@@ -1,6 +1,7 @@
 // @flow
 import * as React from 'react';
-import { FormattedMessage } from 'react-intl';
+import { injectIntl, FormattedMessage } from 'react-intl';
+import type { InjectIntlProvidedProps } from 'react-intl';
 
 import Label from '../../components/label/Label';
 import LoadingIndicator from '../../components/loading-indicator/LoadingIndicator';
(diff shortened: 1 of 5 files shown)


src/features/classification/Classification.js
```diff
@@ -1,6 +1,7 @@
 // @flow
 import * as React from 'react';
-import { FormattedMessage } from 'react-intl';
+import { injectIntl, FormattedMessage } from 'react-intl';
+import type { InjectIntlProvidedProps } from 'react-intl';
```
jpfranco, line 4: nit: Should we group import types together?

src/features/classification/Classification.js
```diff
@@ -24,32 +25,51 @@ type Props = {
...
     const hasSecurityControls = !!controls;
     const isTooltipMessageEnabled = isClassified && hasDefinition && messageStyle === STYLE_TOOLTIP;
     const isInlineMessageEnabled = isClassified && hasDefinition && messageStyle === STYLE_INLINE;
     const isNotClassifiedMessageVisible = !isClassified && messageStyle === STYLE_INLINE;
     const isControlsIndicatorEnabled = isClassified && isLoadingControls && messageStyle === STYLE_INLINE;
     const isSecurityControlsEnabled =
         isClassified && !isLoadingControls && hasSecurityControls && messageStyle === STYLE_INLINE;
+    const isModifiedMessageVisible = isClassified && hasModifiedAt && hasModifiedBy;
+
+    let formattedModifiedAt;
```
jstoffan, line 63: Can we simplify this by making it a constant?

```
const formattedModifiedAt = !!modifiedAt && intl.formatDate(new Date(modifiedAt), {
    day: 'numeric',
    month: 'long',
    year: 'numeric',
});

File: src/features/classification/Classification.js [hunk 3]
```diff
@@ -24,32 +25,51 @@ type Props = {
...
+
+    if (isModifiedMessageVisible) {
+        const d = new Date(modifiedAt);
```
jstoffan, line 66: We try to avoid abbreviation, even for simple variable names.

src/features/classification/Classification.js
```diff
@@ -24,32 +25,51 @@ type Props = {
...
+        formattedModifiedAt = intl.formatDate(d, {
```
jpfranco, line 67: Is it possible to use `<FormattedDate />` here to avoid introducing `intl`?

src/features/classification/__tests__/Classification.test.js
```diff
@@ -57,6 +60,18 @@ describe('features/classification/Classification', () => {
...
+    test('should render classification last modified information when provided and message style is inline', () => {
+        const wrapper = getWrapper({
+            name: 'Confidential',
+            definition: 'fubar',
+            intl,
+            messageStyle: 'inline',
+            modifiedAt: '2020-07-16T00:51:10.000Z',
+            modifiedBy: 'A User',
+        });
+        expect(wrapper).toMatchSnapshot();
```
jstoffan, line 72: We should use explicit assertions instead of snapshots when possible.

src/features/classification/messages.js
```diff
@@ -26,6 +26,16 @@ const messages = defineMessages({
         description: 'Default message for classification in the sidebar when there is none',
         id: 'boxui.classification.missing',
     },
+    modifiedByLabel: {
```
feixuanli, line 29: nit: can we rename this like modifiedByTitle or similar

src/features/classification/Classification.js
patrickpaul: Yikes, this got away from me. Yes but I'll combine it with the shift to FormattedDate.

src/features/classification/Classification.js
patrickpaul: We sure can.

src/features/classification/Classification.js
patrickpaul: Ack. Copied from DateField.js but will make the improvement.

src/features/classification/__tests__/Classification.test.js
patrickpaul: Sure, I already have a classname on the `<p>` might as well use it to verify existence.

src/features/classification/Classification.js
```diff
@@ -36,20 +43,29 @@ const Classification = ({
...
     const isTooltipMessageEnabled = isClassified && hasDefinition && messageStyle === STYLE_TOOLTIP;
     const isInlineMessageEnabled = isClassified && hasDefinition && messageStyle === STYLE_INLINE;
     const isNotClassifiedMessageVisible = !isClassified && messageStyle === STYLE_INLINE;
     const isControlsIndicatorEnabled = isClassified && isLoadingControls && messageStyle === STYLE_INLINE;
     const isSecurityControlsEnabled =
         isClassified && !isLoadingControls && hasSecurityControls && messageStyle === STYLE_INLINE;
+    const isModifiedMessageVisible = isClassified && hasModifiedAt && hasModifiedBy && messageStyle === STYLE_INLINE;
+
+    const formattedModifiedAt = isModifiedMessageVisible && (
+        <FormattedDate value={new Date(modifiedAt)} {...formattedDateOptions} />
```
jstoffan, line 67: Looks like the date options can be passed in as explicit props rather than spread from an object.

src/features/classification/__tests__/Classification.test.js
```diff
@@ -57,6 +57,38 @@ describe('features/classification/Classification', () => {
         expect(wrapper).toMatchSnapshot();
     });
 
+    test('should not render classification last modified information when modified props are not provided', () => {
+        const wrapper = getWrapper({
+            name: 'Confidential',
+            definition: 'fubar',
+            messageStyle: 'inline',
+        });
+        expect(wrapper.find('.bdl-Classification-modifiedBy').length).toBe(0);
```
jstoffan, line 66: We should try to use `data-testid` attributes or component names to select elements, when possible. The `wrapper.exists` helper is also nice here to avoid the length check.

src/features/classification/__tests__/Classification.test.js
jstoffan: Snapshot still needed? Are there other assertions we should add to replace it?

src/features/classification/Classification.js
```diff
@@ -36,20 +43,29 @@ const Classification = ({
...
+    );
```
jpfranco, line 68: Do we need to validate the date in this case? Maybe can use `isValidDate` from datetime utils here?

PR #2251, opened 2020-07-22
fix(sidebar): Only render restrictions label when items exist

patrickpaul:
The changes to the Classifications component caused a `Restrictions` label to render even if no restrictions existed. Moved label rendering to the SecurityControls component and use a prop to selectively render it.

### DIFF
--- a/src/features/classification/security-controls/__tests__/SecurityControls.test.js
+++ b/src/features/classification/security-controls/__tests__/SecurityControls.test.js
@@ -59,6 +59,14 @@ describe('features/classification/security-controls/SecurityControls', () => {
         expect(wrapper).toMatchSnapshot();
     });
 
+    test('should render label for security controls when shouldRenderLabel prop is set', () => {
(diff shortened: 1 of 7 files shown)


src/features/classification/security-controls/__tests__/SecurityControls.test.js
```diff
@@ -59,6 +59,14 @@ describe('features/classification/security-controls/SecurityControls', () => {
         expect(wrapper).toMatchSnapshot();
     });
 
+    test('should render label for security controls when shouldRenderLabel prop is set', () => {
+        wrapper.setProps({ controlsFormat: SHORT });
```
jpfranco, line 63: I think we might not actually want the label for some of formats, but I guess that would be up to the parent component?

patrickpaul: Yes, that's how it's implemented.

PR #2255, opened 2020-07-23
feat(log): Add iscurrent resin tag for AnnotationActivityLink

mxiao6:


src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
```diff
@@ -91,6 +91,11 @@ const AnnotationActivity = ({
                         isDisabled={isFileVersionUnavailable}
                         message={activityLinkMessage}
                         onClick={handleOnSelect}
+                        resinTags={{
+                            'data-resin-iscurrent': isCurrentVersion,
+                            'data-resin-itemid': id,
+                            'data-resin-target': isCurrentVersion ? 'annotationPageLink' : 'annotationVersionLink',
```
jstoffan, line 97: Is changing the target necessary if we are adding `data-resin-iscurrent`?

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
```diff
@@ -91,6 +91,11 @@ const AnnotationActivity = ({
                         isDisabled={isFileVersionUnavailable}
                         message={activityLinkMessage}
                         onClick={handleOnSelect}
+                        resinTags={{
```
jstoffan, line 94: Can these be individual props rather than an object?

src/elements/content-sidebar/activity-feed/annotations/__tests__/AnnotationActivityMenu.test.js
```diff
@@ -28,5 +28,6 @@ describe('elements/content-sidebar/ActivityFeed/annotations/AnnotationActivityMe
         deleteButton();
 
         expect(wrapper.find('DeleteConfirmation').length).toEqual(1);
+        expect(wrapper).toMatchSnapshot();
```
jstoffan, line 31: We should favor explicit assertions over snapshot tests whenever possible.

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
mxiao6: Yeah I also noticed that the information is redundant here. But maybe it's faster for product to get the query result by reducing one query condition?

jstoffan: Can we double-check with them? Would be great to simplify this if possible.

mxiao6: Sure

PR #2260, opened 2020-07-28
chore(content-explorer): Add test ids for e2e testing

jpfranco:
Adding test ids to key elements in `ContentExplorer` in order to allow matching them in a more robust fashion while running e2e tests.

Updates include:

- Adding test id on main `ContentExplorer` element to allow scoping element searches (e.g., using `.within()` with Cypress(.
- Adding test id to each row item. Items are currently hard to match since the actual inputs are hidden and the clickable element (parent row) is a few levels above and does not have any identifying attributes apart from CSS classes.

### DIFF
--- a/src/elements/content-explorer/__tests__/ContentExplorer.test.js
+++ b/src/elements/content-explorer/__tests__/ContentExplorer.test.js
@@ -573,5 +573,10 @@ describe('elements/content-explorer/ContentExplorer', () => {
             expect(uploadDialogElement.length).toBe(1);
             expect(uploadDialogElement.prop('contentUploaderProps')).toEqual(contentUploaderProps);
         });
+
+        test('should render test id for e2e testing', () => {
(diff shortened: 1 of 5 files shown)


src/elements/content-explorer/__tests__/ContentExplorer.test.js
```diff
@@ -573,5 +573,10 @@ describe('elements/content-explorer/ContentExplorer', () => {
             expect(uploadDialogElement.length).toBe(1);
             expect(uploadDialogElement.prop('contentUploaderProps')).toEqual(contentUploaderProps);
         });
+
+        test('should render test id for e2e testing', () => {
+            const wrapper = getWrapper();
+            expect(wrapper.find('[data-testid="content-explorer"]')).toHaveLength(1);
```
sahiga, line 579: You can use `wrapper.exists('[data-testid="content-explorer"]')` here

PR #2263, opened 2020-07-28
feat(sandbox-banner): add SandboxBanner component

viodragon2:
Add new `SandboxBanner` component which displays text at the bottom of the screen.

indicator-stories-test-tsx-components-indicator-indicator-looks-visually-correct-when-using-story-components-indicator-basic-1-snap


src/components/indicator/Indicator.scss
```diff
@@ -0,0 +1,27 @@
...
+
+$indicator-max-width: 584px;
+
+.bdl-Indicator {
+    position: fixed;
+    bottom: 0;
+    left: 50%;
+    z-index: $overlay-z-index + 1;
+    max-width: $indicator-max-width;
+    height: 40px;
```
alexkrolick, line 11: I'm not sure you want to mix height and padding when it comes to centering content. I think you'd want either padding and flex with no fixed height, or fixed height and flex without vertical padding, which runs the risk of not handling overflow

viodragon2: Gotcha. Vertical padding was calculated off of `line-height: 20px`, but I think I'll use the flex  box without vertical padding

alexkrolick: Also, the paddings aren't "on the grid" $bdl-grid-unit which is a but surprising coming from a new design

src/components/indicator/Indicator.scss
```diff
@@ -0,0 +1,27 @@
...
+    border-top-right-radius: 6px;
+    transform: translate(-50%, 0);
+}
+
+.bdl-Indicator-text {
+    display: inline-block;
+    width: 100%;
+    overflow: hidden;
+    color: $white;
+    font-size: 13px;
```
alexkrolick, line 24: there's a variable for the font size, too

src/components/indicator/Indicator.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react';
+
+import './Indicator.scss';
+
+type Props = {
+    children: React.ReactNode;
```
sahiga, line 6: Can this type be more specific? Since the children are rendered within a `span`, you could use `React.ReactText`. Alternatively, if you want to allow `FormattedMessages` as children, you could change the `span` to a `div` and use `React.ReactChild`.

src/components/indicator/Indicator.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+import Indicator from './Indicator';
+
+import notes from './Indicator.stories.md';
+
+export const basic = () => (
+    <Indicator>
+        <span>
+            Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et
+            dolore magna aliqua.
```
sahiga, line 11: This text is cut off in the screenshot. Is there a way to view all the text contained within the indicator?

viodragon2: The UX guideline is to have a tooltip on hover, which can be done via `title` attribute. But given that the `children` accepts a react node and not a string, I wasn't sure whether it should expose extra prop or let the consumer of the component wrap with a proper `title` text like below.
```
<div title="translated text">
  <Indicator>
    <FormattedMessage defaultMessage="translated text" />
  </Indicator>
</div>
```
Do you think it's worth exposing a string type prop for this case?

viodragon2:
Since this feature is specific to an application logic, I have moved from `components` to `features` and renamed it to `SandboxBanner`.

src/components/indicator/Indicator.stories.tsx
alexkrolick: I think we'd want to use the Tooltip component rather than the browser's built-in one, which is slow to appear and hard to read

src/features/sandbox-banner/SandboxBanner.js.flow
```diff
@@ -0,0 +1,22 @@
+// @flow
+import * as React from 'react';
+
+import Tooltip from '../../components/tooltip';
+
+import './SandboxBanner.scss';
+
+type Props = {
+    children: React.Node,
```
sahiga, line 9: `React.Element<any>`?

src/features/sandbox-banner/__tests__/SandboxBanner.stories.test.tsx
```diff
@@ -0,0 +1,6 @@
+describe('features/sandbox-banner/SandboxBanner', () => {
+    test.each([['features-sandboxbanner--basic']])('looks visually correct when using story %s', async id => {
```
sahiga, line 2: Can you add another visual test for the hover state (using `takeScreenshotAfterInput`)?

PR #2264, opened 2020-07-29
feat(content-sharing): Add ES6 wrapper

ctuckz:
Still some issues with styling not being applied properly. Will fix in a separate PR

Screen Shot 2020-07-28 at 5 02 25 PM


src/elements/content-sharing/ContentSharing.js
```diff
@@ -30,12 +29,17 @@ import type {
...
+import '../common/fonts.scss';
+import '../common/modal.scss';
+
 type ContentSharingProps = {
     apiHost: string,
     displayInModal?: boolean,
     itemID: string,
     itemType: ItemType,
     language: string,
+    messages?: StringMap,
```
sahiga, line 42: Will these always be `FormattedMessages`? If so, can we use `MessageDescriptor` from `react-intl` here?

test/sharing-no-react.html
```diff
@@ -0,0 +1,93 @@
...
+        </style>
+
+        <link rel="stylesheet" type="text/css" href="../dev/en-US/sharing.css" />
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/react/16.9.0/umd/react.development.js"></script>
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/react-dom/16.9.0/umd/react-dom.development.js"></script>
+        <script src="../dev/en-US/sharing.no.react.js"></script>
+    </head>
+    <body>
+        <div>
+            <div class="inputs">
```
sahiga, line 27: Since these are all form elements, can they be wrapped inside of a `form`?

test/sharing-no-react.html
```diff
@@ -0,0 +1,93 @@
...
+
+                    const itemType = localStorage.getItem('itemType');
+                    document.getElementById('itemType').value = itemType || 'file';
+
+                    const token = localStorage.getItem('token');
+                    document.getElementById('token').value = token;
+
+                    const displayInModal = localStorage.getItem('displayInModal');
+                    document.getElementById('displayInModal').checked =
+                        displayInModal !== undefined ? displayInModal === 'true' : false;
```
sahiga, line 67: Can this be simplified to `displayInModal === 'true'`?

src/elements/content-sharing/ContentSharing.js
ctuckz: The messages get passed in as a map of id => text, so I don't think MessageDescriptor will work

PR #2266, opened 2020-07-30
feat(content-sharing): Launch ContentSharing from a custom button

sahiga:
This PR adds a `ContentSharingLauncher` component that instantiates `ContentSharing` as a form within the containing page or as a modal on button click, depending on the presence of the optional `customButton` property.

content-sharing-launch-button


manishPh:
Question: What is the top level component for this element? Is it `<ContentSharingLauncher />` or `<ContentSharing.js />`. Or it depends upon use-cases?

sahiga:
Good question, @manishPh. The top level is `ContentSharingLauncher`, as that's where the API is initialized. Would you recommend renaming these components to clarify the hierarchy?

manishPh:
@sahiga yeah, I would keep `<ContentSharing />` as the top level element, consistent with other elements.

sahiga:
@manishPh Fixed. The diff looks big, but I mostly just renamed ContentSharingLauncher -> ContentSharing and ContentSharing -> SharingModal. The actual changes to SharingModal (the old ContentSharing component) can be found in the first commit: https://github.com/box/box-ui-elements/pull/2266/commits/a6664f9e8a9a78d7f628c8320a67891b25eb1ce2

PR #2276, opened 2020-08-06
feat(thumbnail-card): User sees tooltip on hover thumbnail card

ChenCodes:
Changes in this PR:
- [x] User now sees a tooltip showing the full title text on hovering over the title
- [x] Updated tests

Demo:
Screen Shot 2020-08-07 at 11 25 13 AM

### DIFF
--- a/src/components/thumbnail-card/ThumbnailCard.scss
+++ b/src/components/thumbnail-card/ThumbnailCard.scss
@@ -86,11 +86,3 @@
         }
     }
 }
-
-.thumbnail-card-title-tooltip {
-    background-color: black;
-
-    &::before {
-        display: none;
-    }
-}
--- a/src/components/thumbnail-card/ThumbnailCard.scss
+++ b/src/components/thumbnail-card/ThumbnailCard.scss
@@ -86,3 +86,11 @@
         }
     }
 }
+
+.thumbnail-card-title-tooltip {
+    background-color: black;
+
+    &::before {
+        display: none;
+    }
+}
--- a/src/components/thumbnail-card/ThumbnailCardDetails.js
+++ b/src/components/thumbnail-card/ThumbnailCardDetails.js
@@ -1,6 +1,7 @@
 // @flow
 import * as React from 'react';
 import Tooltip from '../tooltip';
+import { useIsContentOverflowed } from '../../utils/dom';
 
 type Props = {
     actionItem?: React.Element<any>,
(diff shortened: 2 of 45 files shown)


src/components/thumbnail-card/ThumbnailCard.scss
```diff
@@ -86,3 +86,11 @@
         }
     }
 }
+
+.thumbnail-card-title-tooltip {
+    background-color: black;
```
alexkrolick, line 91: cc @booc0mtaco we probably don't want one-off tooltip styling?

src/components/thumbnail-card/ThumbnailCardDetails.js
```diff
@@ -1,19 +1,42 @@
...
 type Props = {
     actionItem?: React.Element<any>,
     icon?: React.Node,
+    shouldShowTooltipOnTitleHover: boolean,
     subtitle?: React.Node,
     title: React.Node,
 };
 
-const ThumbnailCardDetails = ({ actionItem, icon, subtitle, title }: Props) => (
+const RenderTitle = (shouldShowTooltipOnTitleHover: boolean, title: React.Node) => {
```
alexkrolick, line 14: style nit: if you're not calling this is a component, don't capitalize the function name

ChenCodes: I had an error earlier where it was saying that renderTitle wasn't a functional component so it couldn't use hooks so I changed it to be capitalized.

src/components/thumbnail-card/ThumbnailCard.scss
booc0mtaco: Correct, we also want to stick with our variables for colors instead of this CSS value.

if we want to define a tooltip style that is pure black (i am guessing this is the ask from design?), we should name that case and use the same param we do for callout (that navy blue color we sometimes see)

booc0mtaco:
Can you describe how the alignment works here? it seems broken

ChenCodes:
> Can you describe how the alignment works here? it seems broken

When the tooltip is displayed, it should be positioned directly above the title text. I've sent the above screens to design and they have confirmed that is the correct behavior.

src/components/thumbnail-card/ThumbnailCard.scss
ChenCodes: So it looks like design has agreed that we should stick with the consistent tooltip. Will remove this block.

src/components/thumbnail-card/ThumbnailCardDetails.js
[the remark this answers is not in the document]
ChenCodes: Fixed

PR #2278, opened 2020-08-06
feat(content-sharing): Add config for email button

sahiga:
Add a configuration object that hides the email button in the USM.

An example of the ContentSharing UI Element without the email button:
content-sharing-email-button-hidden

The default USM:
regular-usm

### DIFF
--- a/src/features/unified-share-modal/flowTypes.js
+++ b/src/features/unified-share-modal/flowTypes.js
@@ -264,9 +264,9 @@ type EmailFormTypes = {
     sendSharedLinkError: React.Node,
 };
 
-export type SharingConfig = {
-    /** Whether the send email button in the USM should be rendered */
-    allowEmails: boolean,
+export type USMConfig = {
+    /** Whether the "Email Shared Link" button and form should be rendered in the USM/USF */
+    showEmailSharedLinkForm: boolean,
 };
 
 // Prop types shared by both the Unified Share Modal and the Unified Share Form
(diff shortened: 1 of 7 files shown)


booc0mtaco:
@sahiga will this also act on sending emails for invite collaborations?

sahiga:
@booc0mtaco Good question! Collaborators can still be invited via the "Invite People" form. This change just affects the "Email Shared Link" form, since we don't have an API for that feature yet.

booc0mtaco:
@sahiga in that case, i am not sure it he key name and associated comment documenting it are specific enough. `SharingOptions` seems like it would (eventually) affect both. if it is meant specifically for disabling the email option for shared links, we may use a name that calls that out more deliberately.

sahiga:
@booc0mtaco I updated the variable and type names. Let me know if these are clear enough, or if you have other thoughts.

src/features/unified-share-modal/flowTypes.js
```diff
@@ -264,6 +264,11 @@ type EmailFormTypes = {
     sendSharedLinkError: React.Node,
 };
 
+export type USMConfig = {
+    /** Whether the "Email Shared Link" button and form should be rendered in the USM/USF */
+    showEmailSharedLinkForm: boolean,
+};
+
```
manishPh, line 271: question: Do we expect more settings to get added in future to this `USMConfig` object?

sahiga: Good question @manishPh. It's definitely a possibility, so I wanted to make the configuration object extensible.

PR #2279, opened 2020-08-07
feat(content-sharing): Retrieve allowed access levels from the Item API

sahiga:
Display allowed access levels and disabled access level reasons in the ContentSharing UI Element; show default values if these fields are unavailable.

content-sharing-disabled-access-levels


src/features/unified-share-modal/utils/convertData.js
```diff
@@ -160,7 +181,8 @@ export const convertItemResponse = (itemAPIData: ContentSharingItemAPIResponse):
 
         sharedLink = {
             accessLevel,
-            allowedAccessLevels: ALLOWED_ACCESS_LEVELS,
+            accessLevelsDisabledReason: shared_link_access_levels_disabled_reasons || {},
```
manishPh, line 184: Minor: add default value `{}` while destructuring `shared_link_access_levels_disabled_reasons ` from `itemAPIData`

sahiga: I tried to do this initially, but unfortunately, it didn't work. The `fillMissingProperties` function (called after a successful response from the API) sets all missing properties to `null`, and destructuring with default values only works when the property is `undefined`.

PR #2284, opened 2020-08-10
feat(metadata-views): Support adding removing values

manishPh:
This PR enables metadata views to add/remove metadata values in addition to replace (PUT) them.


src/features/metadata-based-view/MetadataBasedItemList.js
```diff
@@ -204,16 +206,15 @@ class MetadataBasedItemList extends React.Component<Props, State> {
...
-
-            case FIELD_TYPE_INTEGER:
-                return parseInt(value, 10);
+        if (type === FIELD_TYPE_FLOAT && !isNil(value)) {
+            return parseFloat(value);
+        }
 
-            default:
-                return value;
+        if (type === FIELD_TYPE_INTEGER && !isNil(value)) {
```
manishPh, line 213: using `isNil()` to separate `null`/`undefined` with values like `0`

src/features/metadata-based-view/MetadataBasedItemList.js
```diff
@@ -258,17 +259,18 @@ class MetadataBasedItemList extends React.Component<Props, State> {
                     return cellData;
                 }
                 const { type, value, options = [] } = field;
+                const shouldShowEditIcon = isCellEditable && isString(type);
```
manishPh, line 262: Making sure data type of metadata field is valid, so that we can show edit icon and corresponding <MetadataField /> for that type.

src/features/metadata-based-view/MetadataBasedItemList.js
```diff
@@ -258,17 +259,18 @@ class MetadataBasedItemList extends React.Component<Props, State> {
...
                         {!isCellBeingEdited && <ReadOnlyMetadataField dataValue={value} displayName="" type={type} />}
-                        {isCellEditable && (
+                        {shouldShowEditIcon && (
                             <IconWithTooltip
                                 type={EDIT_ICON_TYPE}
                                 tooltipText={<FormattedMessage {...messages.editLabel} />}
                                 onClick={() => this.handleEditIconClick(columnIndex, rowIndex, value)}
                             />
                         )}
-                        {isCellBeingEdited && valueBeingEdited && (
```
manishPh, line 271: Show edit icon even for the case where value is not present to start with; i.e. for `JSON patch add` operation

src/features/metadata-based-view/MetadataQueryAPIHelper.js
```diff
@@ -54,18 +59,55 @@ export default class MetadataQueryAPIHelper {
...
+                fields: [
+                    'name', // base representation field for an item (name, size, etag etc.)
+                    'metadata.enterprise_12345.myAwesomeTemplateKey.field_1', // metadata instance field
+                    'metadata.enterprise_12345.myAwesomeTemplateKey.field_2', // metadata instance field
+                    'metadata.enterprise_12345.myAwesomeTemplateKey.field_3' // metadata instance field
+                ],
+                ancestor_folder_id: 0,
+            };
+
+            This function will return ['field_1', 'field_2', 'field_3']
```
manishPh, line 106: `['field_1', 'field_2', 'field_3']` will become the columns in the grid view.

src/features/metadata-based-view/MetadataQueryAPIHelper.js
```diff
@@ -75,26 +117,27 @@ export default class MetadataQueryAPIHelper {
...
-                if (includes(SELECT_TYPES, type)) {
-                    // get "options" for enums or multiselects
-                    field.options = getProp(templateField, 'options');
-                }
-
-                return field;
-            });
+        const queryFields = this.getMetadataQueryFields();
+
+        const fields = queryFields.map((queryField: string) => {
```
manishPh, line 122: Iterate over the fields mentioned in the query instead of fields returned in the response. The `metadata_queries` API will return *only* the existing fields.

src/features/metadata-based-view/MetadataBasedItemList.js
```diff
@@ -95,7 +97,7 @@ class MetadataBasedItemList extends React.Component<Props, State> {
             isUpdating: false,
             scrollLeftOffset: 0,
             scrollRightOffset: 0,
-            valueBeingEdited: null,
+            valueBeingEdited: undefined,
```
manishPh, line 100: Changing from `null` to `undefined` to make use of default `propTypes` in components deep down in the component tree.

src/common/types/metadataQueries.js
```diff
@@ -18,6 +18,7 @@ type MetadataQueryOrderByClause = {
 
 type MetadataQuery = {
     ancestor_folder_id: string,
+    fields?: string[],
```
swfree, line 21: Alternatively you could write this as `Array<string>` which I see used in BUIE often

src/features/metadata-based-view/MetadataBasedItemList.js
swfree: would you want to remove `valueBeingEdited` from here then, since its `undefined`?

src/features/metadata-based-view/MetadataQueryAPIHelper.js
```diff
@@ -54,18 +59,55 @@ export default class MetadataQueryAPIHelper {
...
-        },
-    ];
+        };
+
+        if (operation === JSON_PATCH_OP_REMOVE) {
+            delete patchOp.value;
+        }
+
+        const ops = operation === JSON_PATCH_OP_ADD ? [patchOp] : [testOp, patchOp];
+        return ops;
```
sahiga, line 89: Minor: could you return line 88 directly from this function?

src/features/metadata-based-view/__tests__/MetadataBasedItemList.test.js
```diff
@@ -169,6 +169,8 @@ describe('features/metadata-based-view/MetadataBasedItemList', () => {
             ${FIELD_TYPE_FLOAT}       | ${'123.456'}                  | ${123.456}
             ${FIELD_TYPE_MULTISELECT} | ${['Yes', 'No']}              | ${['Yes', 'No']}
             ${FIELD_TYPE_STRING}      | ${'str'}                      | ${'str'}
+            ${FIELD_TYPE_STRING}      | ${undefined}                  | ${undefined}
+            ${null}                   | ${'some value'}               | ${'some value'}
```
sahiga, line 173: Should we also test the cases in which the type is `FIELD_TYPE_INTEGER` or `FIELD_TYPE_FLOAT`, but the value is `null` or `undefined`?

src/features/metadata-based-view/MetadataBasedItemList.js
sahiga: Is there a snapshot test for this change?

src/features/metadata-based-view/MetadataQueryAPIHelper.js
```diff
@@ -54,18 +59,55 @@ export default class MetadataQueryAPIHelper {
...
+                    'metadata.enterprise_12345.myAwesomeTemplateKey.field_2', // metadata instance field
+                    'metadata.enterprise_12345.myAwesomeTemplateKey.field_3' // metadata instance field
+                ],
+                ancestor_folder_id: 0,
+            };
+
+            This function will return ['field_1', 'field_2', 'field_3']
+        */
+        const { fields = [], from } = this.metadataQuery;
+        return fields.filter(field => field.includes(from)).map(field => field.split(`${from}.`).pop());
```
sahiga, line 109: Minor: would `field.split('.').pop()` work, since the field name always appears after the last dot?

src/features/metadata-based-view/__tests__/MetadataBasedItemList.test.js
manishPh: The additional two cases cover the function behavior where it returns `value` when either `value` or `type` doesn't exist. Checking for more types for the same result may not add additional testing.

src/features/metadata-based-view/MetadataBasedItemList.js
manishPh: @sahiga there is but its difficult to test react-virtualized `<AutoSizer />` component with enzyme because of issues like [this](https://github.com/bvaughn/react-virtualized/issues/493#issuecomment-265020587). The other alternative I found out was to export one more component from the same file which is wrapped in `<VirtualScroll />` as mentioned [here](https://stackoverflow.com/questions/37801005/does-react-virtualized-work-with-airbnb-enzyme) so that we can test the functionality but I guess its too complicated for a snapshot test. I have instead relied on unit testing other functionality from the component.

PR #2287, opened 2020-08-11
feat(uploads): add upload fallback logic 

satishboggarapu:
Adds upload fallback logic to both PlainUploads and MultiputUploads

### DIFF
--- a/src/api/uploads/BaseUpload.js
+++ b/src/api/uploads/BaseUpload.js
@@ -6,7 +6,6 @@
 
 import Base from '../Base';
 import { DEFAULT_RETRY_DELAY_MS, MS_IN_S, DEFAULT_HOSTNAME_UPLOAD } from '../../constants';
-import type { APIOptions } from '../../common/types/api';
 
 const MAX_RETRY = 5;
 // Note: We may have to change this number if we add a lot more fast upload hosts.
@@ -37,12 +36,7 @@ class BaseUpload extends Base {
 
     retryTimeout: TimeoutID;
 
-    isUploadFallbackLogicEnabled: boolean;
-
-    constructor(options: APIOptions) {
-        super(options);
-        this.isUploadFallbackLogicEnabled = false;
-    }
+    isUploadFallbackLogicEnabled: boolean = false;
 
     /**
      * Sends an upload pre-flight request. If a file ID is available,
@@ -122,7 +116,7 @@ class BaseUpload extends Base {
         if (isHostReachable) {
             this.preflightSuccessHandler({ data });
         } else if (this.reachabilityRetryCount >= MAX_REACHABILITY_RETRY) {
-            this.preflightSuccessHandler();
+            this.preflightSuccessHandler({ data: {} });
         } else {
             this.reachabilityRetryCount += 1;
             this.makePreflightRequest();
--- a/src/api/uploads/BaseUpload.js
+++ b/src/api/uploads/BaseUpload.js
@@ -97,14 +97,14 @@ class BaseUpload extends Base {
             return;
         }
 
-        const uploadUrl = data.upload_url;
-        // If uploadUrl is not available, don't make reachability test
-        if (!uploadUrl) {
+        const { upload_url } = data;
+        // If upload_url is not available, don't make reachability test
+        if (!upload_url) {
             this.preflightSuccessHandler({ data });
             return;
         }
 
-        const uploadHost = this.getUploadHostFromUrl(uploadUrl);
+        const uploadHost = this.getUploadHostFromUrl(upload_url);
         // The default upload host should always be reachable
         if (uploadHost === `${DEFAULT_HOSTNAME_UPLOAD}/`) {
             this.preflightSuccessHandler({ data });
--- a/src/api/uploads/BaseUpload.js
+++ b/src/api/uploads/BaseUpload.js
@@ -5,9 +5,12 @@
  */
 
 import Base from '../Base';
-import { DEFAULT_RETRY_DELAY_MS, MS_IN_S } from '../../constants';
+import { DEFAULT_RETRY_DELAY_MS, MS_IN_S, DEFAULT_HOSTNAME_UPLOAD } from '../../constants';
+import type { APIOptions } from '../../common/types/api';
 
 const MAX_RETRY = 5;
+// Note: We may have to change this number if we add a lot more fast upload hosts.
+const MAX_REACHABILITY_RETRY = 10;
 
 class BaseUpload extends Base {
     errorCallback: Function;
@@ -30,7 +33,16 @@ class BaseUpload extends Base {
 
     retryCount: number = 0;
 
+    reachabilityRetryCount: number = 0;
+
     retryTimeout: TimeoutID;
+
+    isUploadFallbackLogicEnabled: boolean;
+
+    constructor(options: APIOptions) {
+        super(options);
+        this.isUploadFallbackLogicEnabled = false;
+    }
 
     /**
      * Sends an upload pre-flight request. If a file ID is available,
@@ -49,6 +61,14 @@ class BaseUpload extends Base {
             url = url.replace('content', `${this.fileId}/content`);
         }
 
+        if (this.isUploadFallbackLogicEnabled) {
+            // Add unreachable hosts to url
+            const unreachableHostUrls = this.uploadsReachability.getUnreachableHostsUrls();
+            if (unreachableHostUrls.length !== 0) {
+                url += `?unreachable_hosts=${unreachableHostUrls.join(',')}`;
+            }
+        }
+
         const { size, name } = this.file;
         const attributes = {
             name: this.fileName || name,
@@ -60,9 +80,53 @@ class BaseUpload extends Base {
         this.xhr.options({
             url,
             data: attributes,
-            successHandler: this.preflightSuccessHandler,
+            successHandler: response => {
+                if (this.isUploadFallbackLogicEnabled) {
+                    this.preflightSuccessReachabilityHandler(response);
+                } else {
+                    this.preflightSuccessHandler(response);
+                }
+            },
             errorHandler: this.preflightErrorHandler,
         });
+    };
+
+    /**
+     * Handles successful preflight response.
+     * Performs a upload reachability test before calling preflightSuccessHandler.
+     *
+     * @param {Object} - Request options
+     * @return {Promise} Async function promise
+     */
+    preflightSuccessReachabilityHandler = async ({ data }: { data: { upload_url?: string } }): Promise<any> => {
+        if (this.isDestroyed()) {
+            return;
+        }
+
+        const uploadUrl = data.upload_url;
+        // If uploadUrl is not available, don't make reachability test
+        if (!uploadUrl) {
+            this.preflightSuccessHandler({ data });
+            return;
+        }
+
+        const uploadHost = this.getUploadHostFromUrl(uploadUrl);
+        // The default upload host should always be reachable
+        if (uploadHost === `${DEFAULT_HOSTNAME_UPLOAD}/`) {
+            this.preflightSuccessHandler({ data });
+            return;
+        }
+
+        // If upload host reachable upload file, else make a new preflight request
+        const isHostReachable = await this.uploadsReachability.isReachable(uploadHost);
+        if (isHostReachable) {
+            this.preflightSuccessHandler({ data });
+        } else if (this.reachabilityRetryCount >= MAX_REACHABILITY_RETRY) {
+            this.preflightSuccessHandler();
+        } else {
+            this.reachabilityRetryCount += 1;
+            this.makePreflightRequest();
+        }
     };
 
     /**
@@ -158,6 +222,18 @@ class BaseUpload extends Base {
 
             reader.onerror = reject;
         });
+    }
+
+    /**
+     * Parse uploadHost from uploadUrl
+     *
+     * @param uploadUrl - uploadUrl from preflight response
+     * @return {string}
+     */
+    getUploadHostFromUrl(uploadUrl: string): string {
+        const splitUrl = uploadUrl.split('/');
+        const uploadHost = `${splitUrl[0]}//${splitUrl[2]}/`;
+        return uploadHost;
     }
 }
 
--- a/src/api/uploads/UploadsReachability.js
+++ b/src/api/uploads/UploadsReachability.js
@@ -18,8 +18,6 @@ class UploadsReachability {
 
     /**
      * [constructor]
-     *
-     * @param {Xhr} xhr
      */
     constructor() {
         this.localStore = new LocalStore();
@@ -65,15 +63,15 @@ class UploadsReachability {
     }
 
     /**
-     * Returns the cached result for the given hostUrl
+     * Returns the cached result for the given uploadHost
      *
      * @private
-     * @param {string} hostUrl - The host URL
+     * @param {string} uploadHost - The host URL
      * @return {null|StringAnyMap} The result object or null if there isn't one
      */
-    getCachedResult(hostUrl: string) {
-        if (hostUrl in this.cachedResults) {
-            const result = this.cachedResults[hostUrl];
+    getCachedResult(uploadHost: string) {
+        if (uploadHost in this.cachedResults) {
+            const result = this.cachedResults[uploadHost];
             if (this.isCachedHostValid(result)) {
                 return result;
             }
@@ -86,12 +84,12 @@ class UploadsReachability {
      * Updates a cached result. Changes both the in-memory cachedResult variable and what's stored in local store
      *
      * @private
-     * @param {string} hostUrl - The host URL that was tested
+     * @param {string} uploadHost - The host URL that was tested
      * @param {boolean} isHostReachable - Whether or not the host was reachable
      * @return {void}
      */
-    updateCachedResult(hostUrl: string, isHostReachable: boolean) {
-        this.cachedResults[hostUrl] = {
+    updateCachedResult(uploadHost: string, isHostReachable: boolean) {
+        this.cachedResults[uploadHost] = {
             isReachable: isHostReachable,
             expirationTimestampMS: Date.now() + 1000 * 86400,
         };
@@ -110,10 +108,10 @@ class UploadsReachability {
             return;
         }
 
-        Object.keys(localStoreResults).forEach(hostUrl => {
-            const result = localStoreResults[hostUrl];
+        Object.keys(localStoreResults).forEach(uploadHost => {
+            const result = localStoreResults[uploadHost];
             if (this.isCachedHostValid(result)) {
-                this.cachedResults[hostUrl] = result;
+                this.cachedResults[uploadHost] = result;
             }
         });
     }
@@ -129,10 +127,10 @@ class UploadsReachability {
             return unreachableHosts;
         }
 
-        Object.keys(this.cachedResults).forEach(hostUrl => {
-            const value = this.cachedResults[hostUrl];
+        Object.keys(this.cachedResults).forEach(uploadHost => {
+            const value = this.cachedResults[uploadHost];
             if (this.isCachedHostValid(value) && !value.isReachable) {
-                unreachableHosts.push(hostUrl);
+                unreachableHosts.push(uploadHost);
             }
         });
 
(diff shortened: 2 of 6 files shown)


src/api/uploads/UploadsReachability.js
```diff
@@ -0,0 +1,194 @@
...
+                // console.log(`${hostUrl} is not reachable`);
+                unreachableHosts.push(hostUrl);
+            }
+        });
+
+        return unreachableHosts;
+    }
+
+    /**
+     * Determines whether the given host is reachable by either making a test request to the uplaodHost
```
danubevictoria, line 145: nit: typos for uploadHost

src/api/uploads/BaseUpload.js
```diff
@@ -60,11 +69,51 @@ class BaseUpload extends Base {
...
+        if (uploadHost === `${DEFAULT_HOSTNAME_UPLOAD}/`) {
+            this.preflightSuccessHandler({ data });
+            return;
+        }
+
+        // If upload host reachable upload file, else make a new preflight request
+        const isHostReachable = await this.uploadsReachability.isReachable(uploadHost);
+        if (isHostReachable) {
+            this.preflightSuccessHandler({ data });
+        } else if (this.reachabilityRetryCount >= MAX_REACHABILITY_RETRY) {
```
danubevictoria, line 109: if we hit the max retries, should the host be considered unreachable?

also correct me if i'm wrong, but we want to implement retry logic in reaching the upload host not retry the preflight?

src/api/uploads/BaseUpload.js
```diff
@@ -60,11 +69,51 @@ class BaseUpload extends Base {
...
+            this.preflightSuccessHandler();
```
danubevictoria, line 110: should we pass data still?

src/api/uploads/BaseUpload.js
```diff
@@ -60,11 +69,51 @@ class BaseUpload extends Base {
...
+     */
+    preflightSuccessReachabilityHandler = async ({ data }: { data: { upload_url?: string } }): Promise<any> => {
+        if (this.isDestroyed()) {
+            return;
+        }
+
+        const uploadUrl = data.upload_url;
+        // If uploadUrl is not avaliable, don't make reachability test
+        if (!uploadUrl) {
+            this.preflightSuccessHandler({ data });
```
danubevictoria, line 92: why wouldn't we want to set the upload url here similar to how it's done in preflightSuccessHandler to determine if the upload host is reachable before continuing on with the upload?

src/api/uploads/BaseUpload.js
```diff
@@ -60,11 +69,51 @@ class BaseUpload extends Base {
...
+
+        // If upload host reachable upload file, else make a new preflight request
+        const isHostReachable = await this.uploadsReachability.isReachable(uploadHost);
+        if (isHostReachable) {
+            this.preflightSuccessHandler({ data });
+        } else if (this.reachabilityRetryCount >= MAX_REACHABILITY_RETRY) {
+            this.preflightSuccessHandler();
+        } else {
+            this.reachabilityRetryCount += 1;
+            this.makePreflightRequest();
```
danubevictoria, line 113: so i'm assuming we are retrying with another preflight request because we got back an unreachable host is that right? should we be expecting that the preflight request would give us back a different upload url with each call?

i would actually say that if we got an unreachable host back from the preflight response, we might consider that a preflight error and just call preflightErrorHandler which handles the retry logic and calls makePreflightRequest. that would eliminate the need to track reachabilityRetries.

what are your thoughts?

src/api/uploads/BaseUpload.js
```diff
@@ -49,6 +52,12 @@ class BaseUpload extends Base {
             url = url.replace('content', `${this.fileId}/content`);
         }
 
+        // Add unreachable hosts to url
+        const unreachableHostUrls = this.uploadsReachability.getUnreachableHostsUrls();
+        if (unreachableHostUrls.length !== 0) {
+            url += `?unreachable_hosts=${unreachableHostUrls.join(',')}`;
```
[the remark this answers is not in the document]
satishboggarapu: Yes, this param is not used anywhere else. Given that I don't know how much value we would get from replacing any pre-existing params. 

The preflight call already knows how to handle this query parameter and it would only work for customers who have the Box Accelerator feature enabled.

src/api/uploads/BaseUpload.js
```diff
@@ -60,11 +69,51 @@ class BaseUpload extends Base {
         this.xhr.options({
             url,
             data: attributes,
-            successHandler: this.preflightSuccessHandler,
+            successHandler: this.preflightSuccessReachabilityHandler,
```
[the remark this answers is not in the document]
satishboggarapu: Not sure how we can do this, since this is my first time working within this project but I'm open to suggestions.

src/api/uploads/BaseUpload.js
satishboggarapu: In this case we would only hit this `else if`  when both the reachabilityRetryCount is greaterThenOrEqual to the max retry count and when the reachability test fails. Given that it doesn't make sense to pass in `data` to the preflightSuccessHandler since we already know that host is unreachable and the upload will just fail.

If the reachability test fails for a host, it doesn't make sense to retry the reachability test again since it will just fail again. Once we know a host is not reachable, in order to get a new host we need to do a preflight check again but this time with the `unreachable_hosts` query parameter.

src/api/uploads/BaseUpload.js
satishboggarapu: See response above ^

src/api/uploads/BaseUpload.js
satishboggarapu: > so i'm assuming we are retrying with another preflight request because we got back an unreachable host is that right?

yes

> should we be expecting that the preflight request would give us back a different upload url with each call?

Yes, the preflight request would give us a different upload url with each call given that we keep updating the `unreachable_hosts` query parameter.

I do like the idea of calling the preflightErrorHandler if the preflight returns an unreachable host. But we will have to increase the `MAX_RETRY` value to a higher number. This might cause some problems since we will retry more for other scenarios. For example, if we are being rate limited 429, we might end up retrying more than originally intended. 
I also don't like the idea of completely canceling the upload if we reach max retries due to reachability tests failing. We are always adding more upload hosts, so the list of possible hosts will keep growing over time. This will cause us to increase the overall `MAX_RETRY` count which like I mentioned above might lead to some unintended consequences. If we reach the max retry count due to reachability tests failing we should fallback to the default upload hosts, which should always be reachable, instead of canceling the upload. This is why I decided to keep the `reachabilityRetryCount` separate from `retryCount`. 

Let me know what you think.

src/api/uploads/BaseUpload.js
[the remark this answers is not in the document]
danubevictoria: for persisting query params, are there other query params that could come through other than unreachable hosts? because if so, we want to persist them as well as add the new unreachable_hosts.

we are always running this reachability test. is there some way for us to identify that this is a customer who runs Box Accelerator in order to determine if we need to do reachability testing? or should we do reachability testing for all uploads?

src/api/uploads/BaseUpload.js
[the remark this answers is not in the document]
danubevictoria: here's some good starting points. You'll need access to Box/scm, Box/EndUserApp, AppConf/Gerritt.
https://confluence.inside-box.net/display/ETO/Feature+Flipping
https://cloud.box.com/s/qh34eqgyzmf3t94u4ypn1ckdwusj8zvv
example PR in BUIE (isResumableUploadsEnabled): https://github.com/box/box-ui-elements/pull/1552/files

let me know if its confusing and we can set some time up to talk through what feature flips are in webapp.

src/api/uploads/BaseUpload.js
```diff
@@ -60,11 +69,51 @@ class BaseUpload extends Base {
...
+     * @param {Object} - Request options
+     * @return {Promise} Async function promise
+     */
+    preflightSuccessReachabilityHandler = async ({ data }: { data: { upload_url?: string } }): Promise<any> => {
+        if (this.isDestroyed()) {
+            return;
+        }
+
+        const uploadUrl = data.upload_url;
+        // If uploadUrl is not avaliable, don't make reachability test
```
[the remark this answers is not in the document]
danubevictoria: ok cool, i just wanted to double check that we do typically expect to get an uploadUrl and are not just usually setting it ourselves.

src/api/uploads/BaseUpload.js
danubevictoria: what does `data` look like? is it just an `uploadUrl` that comes back?

src/api/uploads/BaseUpload.js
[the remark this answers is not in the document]
satishboggarapu: Don't quite understand your question about the query params. The base `url` to make the preflight request is created  just from the line above and it doesn't have any query params. As far as I'm aware there are no other query params.

There is no way for us to identify if a customer has the accelerated uploads option enabled. We will always try to do a reachability test, but for the majority of the customers we will never end up actually performing the test since the preflight request for customers who don't have the accelerator option enabled will just return the default upload host. We don't do a reachability test if the `upload_url` from the preflight request is the default upload host.

src/api/uploads/BaseUpload.js
satishboggarapu: Here is a sample of what `data` looks like 
`{
  "upload_token": "Pc3FIOG9vSGV4VHo4QzAyg5T1JvNnJoZ3ExaVNyQWw6WjRsanRKZG5lQk9qUE1BVQP",
  "upload_url": "https://upload-las.app.box.com/api/2.0/files/content?upload_session_id=1234"
}` https://developer.box.com/reference/options-files-content/

src/api/uploads/BaseUpload.js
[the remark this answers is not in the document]
danubevictoria: ohh gotcha, didn't see the url line above. that makes sense.

src/api/uploads/BaseUpload.js
```diff
@@ -5,9 +5,10 @@
  */
 
 import Base from '../Base';
-import { DEFAULT_RETRY_DELAY_MS, MS_IN_S } from '../../constants';
+import { DEFAULT_RETRY_DELAY_MS, MS_IN_S, DEFAULT_HOSTNAME_UPLOAD } from '../../constants';
 
 const MAX_RETRY = 5;
+const MAX_REACHABILITY_RETRY = 10;
```
danubevictoria, line 11: I don't understand how we landed on 10 as max retries. Do we know the max possible different upload hosts that could come back? I'd imagine that would determine the max reachability retries before we should default to the default upload host.

src/api/uploads/BaseUpload.js
danubevictoria: do we need to use this upload_token to make the reachability test?

src/api/uploads/UploadsReachability.js
```diff
@@ -0,0 +1,194 @@
...
+        const url = `${hostUrl}html5?reachability_test=run`;
+        const headers: StringMap = {
+            [HEADER_CONTENT_TYPE]: 'application/x-www-form-urlencoded; charset=UTF-8',
+            'X-File-Name': 'reachability_pseudofile.txt',
+            'X-File-Size': '1234',
+        };
+        const data = 'reachability_file=test_file_data';
+
+        try {
+            await axios.post(url, data, { headers });
```
danubevictoria, line 185: Would this fail faster if we first check that a GET request succeeds before we attempt a POST?

src/api/uploads/UploadsReachability.js
```diff
@@ -0,0 +1,194 @@
...
+     *
+     * @private
+     * @param {string} hostUrl - The host URL that was tested
+     * @param {boolean} isHostReachable - Whether or not the host was reachable
+     * @return {void}
+     */
+    updateCachedResult(hostUrl: string, isHostReachable: boolean) {
+        this.cachedResults[hostUrl] = {
+            isReachable: isHostReachable,
+            expirationTimestampMS: Date.now() + 1000 * 86400,
```
danubevictoria, line 96: How did we decide on how long a reachability test is considered valid?

src/api/uploads/BaseUpload.js
satishboggarapu: No, I don't think the upload_token is used any where. I'm not even sure what this token is in the first place.

src/api/uploads/BaseUpload.js
satishboggarapu: Copied the value from what the current max retries in php, which is 10. From this looks like we currently have 7 different upload hosts https://support.box.com/hc/en-us/articles/360043694474-Box-Connectivity-Tests.

src/api/uploads/UploadsReachability.js
```diff
@@ -0,0 +1,192 @@
...
+
+    /**
+     * Determines if the given hostUrl is reachable by making a test upload request to it.
+     * Does not read or modify any cached results.
+     *
+     * @param {string} hostUrl - The upload host url to make a test request against
+     * @return {Promise<boolean>}
+     */
+    async makeReachabilityRequest(hostUrl: string) {
+        const url = `${hostUrl}html5?reachability_test=run`;
```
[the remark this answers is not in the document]
satishboggarapu: The url used in the previous iteration was not meant for the reachability test. The new url that I'm using was originally created for this purpose only

src/api/uploads/UploadsReachability.js
```diff
@@ -0,0 +1,192 @@
...
+        this.saveCachedResultsToLocalStore();
+    }
+
+    /**
+     * Adds to the cachedResults object with valid entries from local storage
+     *
+     * @private
+     * @return {void}
+     */
+    populateCachedResultsWithValidLocalStoreValues() {
```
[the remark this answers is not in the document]
satishboggarapu: It's not being used by anything else. I would agree with your suggestion if this method was being used by something other than the constructor.

src/api/uploads/UploadsReachability.js
```diff
@@ -0,0 +1,192 @@
...
+     * Determines whether the given host is reachable by either making a test request to the uploadHost
+     * or returning the result of the last reachability test it did
+     *
+     * @param {string} uploadHost - The upload host URL that will be stored in the cached test result and returned in
+     * getUnreachableHostsUrls() if test fails (this is usually a prefix of the uploadUrl)
+     * @return {Promise<boolean>} Promise that resolved to true if the host is reachable, false if it is not
+     */
+    async isReachable(uploadHost: string) {
+        // The default upload host should always reachable
+        if (uploadHost === `${DEFAULT_HOSTNAME_UPLOAD}/`) {
```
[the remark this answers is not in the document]
satishboggarapu: Yes in BaseUpload, when the uploadHost is parsed from the uploadUrl, it always ends in `/` . But the `DEFAULT_HOSTNAME_UPLOAD` constant doesn't.

src/api/uploads/UploadsReachability.js
```diff
@@ -0,0 +1,192 @@
...
+import LocalStore from '../../utils/LocalStore';
+import { DEFAULT_HOSTNAME_UPLOAD, HEADER_CONTENT_TYPE } from '../../constants';
+import type { StringAnyMap, StringMap } from '../../common/types/core';
+
+const CACHED_RESULTS_LOCAL_STORE_KEY = 'bcu-uploads-reachability-cached-results';
+
+class UploadsReachability {
+    localStore: LocalStore;
+
+    cachedResults: StringAnyMap;
```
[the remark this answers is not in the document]
satishboggarapu: I'm a newbie to React, how can I make this a more specific type?

src/api/uploads/BaseUpload.js
```diff
@@ -30,8 +33,17 @@ class BaseUpload extends Base {
...
 
+    reachabilityRetryCount: number = 0;
+
     retryTimeout: TimeoutID;
 
+    isUploadFallbackLogicEnabled: boolean;
+
+    constructor(options: APIOptions) {
+        super(options);
+        this.isUploadFallbackLogicEnabled = false;
```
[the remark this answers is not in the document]
satishboggarapu: Yes, I will fix it.

src/api/uploads/BaseUpload.js
```diff
@@ -60,11 +80,55 @@ class BaseUpload extends Base {
...
+     *
+     * @param {Object} - Request options
+     * @return {Promise} Async function promise
+     */
+    preflightSuccessReachabilityHandler = async ({ data }: { data: { upload_url?: string } }): Promise<any> => {
+        if (this.isDestroyed()) {
+            return;
+        }
+
+        const uploadUrl = data.upload_url;
```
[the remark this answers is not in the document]
satishboggarapu: Whats the difference between the two?

src/api/uploads/BaseUpload.js
```diff
@@ -159,6 +223,18 @@ class BaseUpload extends Base {
...
+
+    /**
+     * Parse uploadHost from uploadUrl
+     *
+     * @param uploadUrl - uploadUrl from preflight response
+     * @return {string}
+     */
+    getUploadHostFromUrl(uploadUrl: string): string {
+        const splitUrl = uploadUrl.split('/');
+        const uploadHost = `${splitUrl[0]}//${splitUrl[2]}/`;
```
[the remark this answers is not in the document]
satishboggarapu: The `unreachable_hosts` query parameter, that is part of the preflight request in `makePreflightRequest`, it requires the unreachable hosts to be in the format `https://upload.box.com/`.

src/api/uploads/UploadsReachability.js
```diff
@@ -0,0 +1,192 @@
...
+
+    /**
+     * Determines whether the given host is reachable by either making a test request to the uploadHost
+     * or returning the result of the last reachability test it did
+     *
+     * @param {string} uploadHost - The upload host URL that will be stored in the cached test result and returned in
+     * getUnreachableHostsUrls() if test fails (this is usually a prefix of the uploadUrl)
+     * @return {Promise<boolean>} Promise that resolved to true if the host is reachable, false if it is not
+     */
+    async isReachable(uploadHost: string) {
```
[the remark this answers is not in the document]
satishboggarapu: fixed

src/api/uploads/BaseUpload.js
```diff
@@ -60,11 +80,55 @@ class BaseUpload extends Base {
...
+            this.preflightSuccessHandler({ data });
+            return;
+        }
+
+        // If upload host reachable upload file, else make a new preflight request
+        const isHostReachable = await this.uploadsReachability.isReachable(uploadHost);
+        if (isHostReachable) {
+            this.preflightSuccessHandler({ data });
+        } else if (this.reachabilityRetryCount >= MAX_REACHABILITY_RETRY) {
+            this.preflightSuccessHandler();
```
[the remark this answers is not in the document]
satishboggarapu: Yes, it will cause an error. Fixed.

src/api/uploads/UploadsReachability.js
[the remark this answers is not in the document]
ConradJChan: You can check this out: https://flow.org/en/docs/types/aliases/ -- probably something like 
```
type CachedResult {
  isReachable: boolean,
  expirationTimestampMS: number,
}

type CachedResults = Record<string, CachedResult>
```

src/api/uploads/BaseUpload.js
```diff
@@ -60,11 +74,55 @@ class BaseUpload extends Base {
...
+            this.preflightSuccessHandler({ data });
+            return;
+        }
+
+        const uploadHost = this.getUploadHostFromUrl(uploadUrl);
+        // The default upload host should always be reachable
+        if (uploadHost === `${DEFAULT_HOSTNAME_UPLOAD}/`) {
+            this.preflightSuccessHandler({ data });
+            return;
+        }
```
ConradJChan, line 112: Could this be combined with the guard clause above on L102?

src/api/uploads/__tests__/BaseUpload.test.js
```diff
@@ -67,18 +72,172 @@ describe('api/uploads/BaseUpload', () => {
...
+
+    describe('preflightSuccessReachabilityHandler()', () => {
+        beforeEach(() => {
+            upload.isUploadFallbackLogicEnabled = jest.fn().mockReturnValueOnce(true);
+        });
+
+        test('should not do anything if API is destroyed', () => {
+            upload.isDestroyed = jest.fn().mockReturnValueOnce(true);
+            upload.makePreflightRequest = jest.fn();
+            upload.preflightSuccessReachabilityHandler({ data: {} });
```
ConradJChan, line 137: It looks like a lot of this setup could be moved to `beforeEach`

src/api/uploads/__tests__/UploadsReachability.test.js
```diff
@@ -0,0 +1,139 @@
...
+            expirationTimestampMS: currentTimeMS + 30,
+        },
+    };
+
+    beforeEach(() => {
+        uploadsReachability = new UploadsReachability();
+    });
+
+    describe('isReachable()', () => {
+        withData(
```
ConradJChan, line 42: I don't see `leche` used anywhere else in the repo. Could we accomplish the same thing with `test.each`? https://jestjs.io/docs/en/api#testeachtablename-fn-timeout

src/api/uploads/BaseUpload.js
satishboggarapu: I don't it can since, L102 checks if the `uploadUrl` is defined, while here we are checking the `uploadHost` which needs to be parsed from the `uploadUrl`

src/api/uploads/__tests__/UploadsReachability.test.js
satishboggarapu: Fixed

ConradJChan (dismissed):
Looks like there is a flow error though, misled you with the `Record<string, CachedResult>` since that's a typescript utility type 😅 sorry! There should be a flow equivalent

PR #2288, opened 2020-08-11
fix(metadata-instance-editor): allow specifying blur handler class names

krapspark:
Adding support for specifying class names which will be prevent `handleBlur` from executing in `BaseSelectField`.
This was necessary in cases where the container could have a `tabindex` set which would move focus from select field if the user interacted with the container's scrollbar.
This would trigger `handleBlur` which would then close the dropdown making some options almost impossible to select.

This change allows for adding exception to blur handler so that if the focus is moved to an element with exempted class names, we can safely ignore that blur case.


src/features/metadata-instance-editor/CascadePolicy.scss
```diff
@@ -42,7 +42,7 @@ $cascade-policy-background: #f1e2fd;
     padding: 14px 15px;
 }
 
-.metadata-cascade-toggle {
+.toggle-container.metadata-cascade-toggle {
```
krapspark, line 45: a new CSS rule on `.toggle-container` was messing with the cascade display. fixing it here.

src/components/select-field/BaseSelectField.js
```diff
@@ -221,12 +223,17 @@ class BaseSelectField extends React.Component<Props, State> {
...
+        const exceptionClasses = ['search-input', 'select-button', ...blurExceptionClassNames];
+
         if (
             isOpen &&
             event &&
             event.relatedTarget &&
-            !(event.relatedTarget: window.HTMLInputElement).classList.contains('search-input') &&
-            !(event.relatedTarget: window.HTMLInputElement).classList.contains('select-button')
+            exceptionClasses.every(
+                className => event && !(event.relatedTarget: window.HTMLInputElement).classList.contains(className),
```
swfree, line 235: minor: do you need the `event &&` check, since thats already checked on line 232?

krapspark: we don't but Flow was complaining that `event` could be empty. I think it's because of the scoping?

PR #2293, opened 2020-08-13
fix(style): move generic BDL overlay styles to shared mixin directory

booc0mtaco:
Move the styles used for the new BDL-style overlays to a common area, to be used for other components with overlay behavior.

This change should be a visual no-op; the same styles should be used on the components where originally applied.


booc0mtaco:
See #2179 for relevant screenshots

booc0mtaco:
Reference component images:
84178947-1972cb00-aa3a-11ea-9614-94d364e3f08f

PR #2295, opened 2020-08-13
fix(usm): fix space in between modal title and classification badge

feixuanli:
Screen Shot 2020-08-13 at 4 16 27 PM
Screen Shot 2020-08-13 at 4 16 48 PM
Screen Shot 2020-08-13 at 4 17 02 PM

- Unless there is not enough room, classification badge should NOT be pushed to the next line.

- In the case that there is not enough room and classification badge has to be the next line, there should not be any space before the badge - it should left align with the title.


booc0mtaco:
@feixuanli could you include some screenshots for this?

src/features/unified-share-modal/UnifiedShareModal.scss
```diff
@@ -52,13 +52,12 @@
...
-        flex-flow: row wrap;
+    .bdl-UnifiedShareModalTitle-classification {
+        display: inline;
+    }
 
-        .bdl-UnifiedShareModalTitle-classification {
-            margin-left: 6px;
-        }
+    .bdl-UnifiedShareModalTitle {
+        margin-right: 6px;
```
booc0mtaco, line 60: this should probably be on the grid `bdl-grid-unit` (*2?)

feixuanli: Thanks for the comment! For classification badges we have the space of 6px on other places such as page header/ share link settings modal. `bdl-grid-unit` is 4px. I think we should keep the space consistent as others. 
Screen Shot 2020-08-13 at 5 01 37 PM
Screen Shot 2020-08-13 at 5 01 26 PM

booc0mtaco: if so, that means these ones are inconsistent with the other spacings used in the app? if all other ones match at 6px, i suppose they should be internally consistent, but I think design will have to make the final call on that since they created the grid for us to use

PR #2296, opened 2020-08-13
fix(header-flyout): update header flyout container overly to match BDL

il10:


src/features/header-flyout/styles/HeaderFlyout.scss
```diff
@@ -13,8 +14,8 @@
 
     .overlay {
         padding: 2 * $flyout-header-spacer $flyout-list-padding 0;
-        // TODO: decide which BDL color applies here
-        border: 1px solid $bdl-gray-10;
+
+        @include bdl-Overlay-container;
```
alexkrolick, line 18: Is the `padding` value overridden by this - do we still need it?

il10: `bdl-Overlay-container` does not contain padding so we do need it still

booc0mtaco: @alexkrolick from some of our earlier discussions, i opted not to include any internal padding to these mixins, to avoid needing overrides in all use cases, like our button styles

alexkrolick: Ok - might make sense to stick to the convention of `@include` blocks coming before overrides/local values though, just in case.

PR #2297, opened 2020-08-19
refactor(metadata-view): Add back the metadataFieldsToShow prop

manishPh:


src/elements/content-explorer/Content.js
```diff
@@ -22,9 +23,9 @@ import './Content.scss';
  * @param {Object} currentCollection the current collection
  * @return {boolean} empty or not
  */
-function isEmpty(view: View, currentCollection: Collection): boolean {
+function isEmpty(view: View, currentCollection: Collection, metadataFieldsToShow: MetadataFieldsToShow): boolean {
```
swfree, line 26: you can add metadataFieldsToShow to the params definition above `* @param {MetadataFieldsToShow} metadataFieldsToShow`

src/elements/content-explorer/Content.js
```diff
@@ -64,9 +66,10 @@ const Content = ({
     tableRef,
     view,
     viewMode = VIEW_MODE_LIST,
+    metadataFieldsToShow = [],
```
swfree, line 69: if the user doesn't pass in anything on the `metadataFieldsToShow` prop, it looks like the user will only see the empty state when in the metadata view. is this expected or are there any default fields we would show?

manishPh: yeah, empty state is ok because every metadata template is different.

src/elements/content-explorer/Content.js
```diff
@@ -20,11 +21,12 @@ import './Content.scss';
...
  * @param {string} view the current view
  * @param {Object} currentCollection the current collection
+ * @param {MetadataFieldsToShow} metadataFieldsToShow list of metadata template fields to show
  * @return {boolean} empty or not
  */
-function isEmpty(view: View, currentCollection: Collection): boolean {
+function isEmpty(view: View, currentCollection: Collection, metadataFieldsToShow: MetadataFieldsToShow): boolean {
     const { items = [] }: Collection = currentCollection;
-    return view === VIEW_ERROR || items.length === 0;
+    return view === VIEW_ERROR || items.length === 0 || (view === VIEW_METADATA && metadataFieldsToShow.length === 0);
```
sahiga, line 29: Minor: could you use `!items.length` and `!metadataFieldsToShow.length` here?

src/common/types/metadataQueries.js
```diff
@@ -16,6 +16,14 @@ type MetadataQueryOrderByClause = {
     field_key: string,
 };
 
+type MetadataFieldConfig = {
+    canEdit?: boolean,
+    displayName?: string,
+    key: string,
+};
+
+type MetadataFieldsToShow = Array<MetadataFieldConfig | string>;
```
sahiga, line 25: I'm guessing that the two types are from the API, but it's not immediately clear to me why a metadata field might be a string. Can you add a comment in the code with an explanation of this case?

src/features/metadata-based-view/__tests__/MetadataBasedItemList.test.js
```diff
@@ -179,16 +200,20 @@ describe('features/metadata-based-view/MetadataBasedItemList', () => {
...
+            ${2}        | ${1}     | ${'bill'}        | ${'string'}
+            ${3}        | ${1}     | ${100.12}        | ${'float'}
+            ${4}        | ${1}     | ${createdAt}     | ${'date'}
+            ${5}        | ${1}     | ${undefined}     | ${undefined}
+            ${0}        | ${2}     | ${mp4Icon}       | ${undefined}
+            ${1}        | ${2}     | ${mp4NameButton} | ${undefined}
+            ${2}        | ${2}     | ${'receipt'}     | ${'string'}
+            ${3}        | ${2}     | ${200.88}        | ${'float'}
+            ${4}        | ${2}     | ${createdAt}     | ${'date'}
+            ${5}        | ${2}     | ${undefined}     | ${undefined}
```
sahiga, line 215: Do numerical values in this view always have the `float` data type? Or could they be integers?

manishPh: These are data types of Box metadata fields: https://developer.box.com/guides/metadata/fields/

src/common/types/metadataQueries.js
manishPh: `MetadataFieldsToShow` type is not from API. This is a prop that developers can pass to the <ContentExplorer />. It could be simple as string or it could be an config object
```
const mdFieldsToShow1: MetadataFieldsToShow = ['name', 'desc'];
const mdFieldsToShow2: MetadataFieldsToShow = [
    { key: 'name', displayName: 'Item Name', canEdit: false },
    { key: 'desc', displayName: 'Description', canEdit: true },
];
```

I will add a comment.

PR #2298, opened 2020-08-19
fix(content-uploader): allow cancellation of folder uploads

rosepierce:
Added a check for if an item is a folder before attempting to get the file size to prevent errors when attempting to cancel a single folder upload or multiple folder uploads. 

Additionally, changed the ID assigned when there are not API Options for a DataTransferItem to be identical to that for a File (where only the name is used as the ID when there are not API Options available), so that behavior for folders being uploaded is identical to that of files.

### DIFF
--- a/src/elements/content-uploader/__tests__/ContentUploader.test.js
+++ b/src/elements/content-uploader/__tests__/ContentUploader.test.js
@@ -223,7 +223,7 @@ describe('elements/content-uploader/ContentUploader', () => {
 
             instance.onClick(item);
 
-            expect(instance.removeFileFromUploadQueue).toBeCalled();
+            expect(instance.removeFileFromUploadQueue).toBeCalledWith(item);
             expect(onClickCancel.mock.calls.length).toBe(1);
         });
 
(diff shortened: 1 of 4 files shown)


src/utils/__tests__/uploads.test.js
```diff
@@ -277,7 +277,7 @@ describe('util/uploads', () => {
         Date.now = jest.fn(() => now);
 
         test('should return item id correctly when item does not contain API options', () => {
-            expect(getDataTransferItemId(mockItem, rootFolderId)).toBe(`hi_0_${now}`);
+            expect(getDataTransferItemId(mockItem, rootFolderId)).toBe(`hi`);
```
sahiga, line 280: Since this is no longer an interpolated string, you can use single quotes instead of back ticks

src/elements/content-uploader/__tests__/ContentUploader.test.js
```diff
@@ -213,6 +213,20 @@ describe('elements/content-uploader/ContentUploader', () => {
...
+            const item = { api: {}, isFolder: true, status: STATUS_IN_PROGRESS };
+            const onClickCancel = jest.fn();
+            const wrapper = getWrapper({ onClickCancel });
+            const instance = wrapper.instance();
+
+            instance.removeFileFromUploadQueue = jest.fn();
+
+            instance.onClick(item);
+
+            expect(instance.removeFileFromUploadQueue).toBeCalled();
```
sahiga, line 226: Can you also check for the parameter here, i.e., `toBeCalledWith(item)`?

PR #2299, opened 2020-08-20
fix(image-tooltip): hide ImageTooltip until image has loaded

laughnan:


src/components/image-tooltip/ImageTooltipContent.tsx
```diff
@@ -3,34 +3,45 @@ import classNames from 'classnames';
...
 
     return null;
 }
 
-const ImageTooltipContent = ({ children, content, title }: ImageTooltipContentProps) => (
+const ImageTooltipContent = ({ children, content, onImageLoad, title }: ImageTooltipContentProps) => (
     <div className="bdl-ImageTooltipContent">
         <div className="bdl-ImageTooltipContent-image">
-            {React.Children.map(children, child => appendTooltipImageChildClass(child))}
+            {React.Children.map(children, child => appendTooltipImageChildClass(child, onImageLoad))}
```
DanDeMicco, line 44: since there can be multiple children, will this be an issue? should you wait until they are all loaded?

src/components/image-tooltip/ImageTooltipContent.tsx
```diff
@@ -3,34 +3,45 @@ import classNames from 'classnames';
...
     /** A string to be used in the tooltip's paragraph content */
     content: string;
+    /** A callback triggered onLoad for the image element */
+    onImageLoad?: OnImageLoad;
     /** A string to be used in the tooltip's title heading */
     title: string;
 };
 
-function appendTooltipImageChildClass(child: React.ReactElement): React.ReactElement | null {
+function appendTooltipImageChildClass(child: React.ReactElement, onImageLoad?: OnImageLoad): React.ReactElement | null {
```
DanDeMicco, line 19: function name is not really accurate now

src/components/image-tooltip/ImageTooltipContent.tsx
```diff
@@ -3,34 +3,45 @@ import classNames from 'classnames';
...
 
+type OnImageLoad = () => void;
+
 export type ImageTooltipContentProps = {
     /** A React component representing the image used in the visual tooltip */
     children: React.ReactElement;
     /** A string to be used in the tooltip's paragraph content */
     content: string;
+    /** A callback triggered onLoad for the image element */
+    onImageLoad?: OnImageLoad;
```
DanDeMicco, line 14: should this be optional? seems it doesnt have to be (parent tooltip always passes it in i think?)

src/components/image-tooltip/ImageTooltipContent.tsx
laughnan: looking into this...this might not need to be a map since we only support a single element coming in 🤔

src/components/image-tooltip/ImageTooltipContent.tsx
laughnan: thanks! updating...

src/components/image-tooltip/ImageTooltipContent.tsx
laughnan: I was thinking about `ImageTooltipContent` as a standalone component, but maybe it's always used via `ImageToolip` 🤔  I never know how abstract we want to be with open source-y stuffs.

DanDeMicco: Since it lives within the image-tooltip folder it'd be okay to make it non-optional but doesnt really matter

src/components/image-tooltip/ImageTooltipContent.tsx
```diff
@@ -3,35 +3,34 @@ import classNames from 'classnames';
...
+        className,
+        onLoad: onImageLoad,
+    });
 }
 
-const ImageTooltipContent = ({ children, content, title }: ImageTooltipContentProps) => (
+const ImageTooltipContent = ({ children, content, onImageLoad, title }: ImageTooltipContentProps) => (
     <div className="bdl-ImageTooltipContent">
-        <div className="bdl-ImageTooltipContent-image">
-            {React.Children.map(children, child => appendTooltipImageChildClass(child))}
```
laughnan, line 33: We can replace this `.map()` with just the call to our cloning function since our parent does a `React.Children.only` check.

PR #2306, opened 2020-08-25
fix(open-with): fix onError and onExecute callback options

swfree:
BREAKING CHANGE: if a user passes in onError & onExecute options, those will now replace the equivalent methods on the ES6 wrapper. This is the expected behavior according to the docs, but I'm marking this as a breaking change since its an options update.

The Developer Docs already lists these callbacks as options in https://developer.box.com/guides/embed/ui-elements/open-with/#options


swfree:
@sahiga I found a way to test manually locally - I ran `yarn start:dev` and navigated to the open with element defined here: https://github.com/box/box-ui-elements/blob/master/test/openwith.html#L104. I added the onError option to the html file, and found that it doesn't trigger before my change, but it does trigger after my change.

PR #2309, opened 2020-08-26
feat(annotations): Add Highlight Text Icon

mxiao6:


src/icons/annotations/IconHighlightText.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
+import AccessibleSVG from '../accessible-svg';
+import { Icon } from '../iconTypes';
+
+const IconHighlightText = ({
+    className = '',
+    color = 'white',
+    height = 16,
+    title = 'Highlight Text',
```
jstoffan, line 9: Do other icons have a hardcoded en-US string like this one?

src/icons/annotations/IconHighlightText.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
+import AccessibleSVG from '../accessible-svg';
+import { Icon } from '../iconTypes';
+
+const IconHighlightText = ({
+    className = '',
+    color = 'white',
```
jstoffan, line 7: Should this be a variable?

src/icons/annotations/IconHighlightText.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
+import AccessibleSVG from '../accessible-svg';
+import { Icon } from '../iconTypes';
+
+const IconHighlightText = ({
+    className = '',
```
jstoffan, line 6: Is this needed?

mxiao6: Yeah, just following icon files convention. Supporting custom classname is not harmful, right

src/icons/annotations/IconHighlightText.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react';
+import AccessibleSVG from '../accessible-svg';
+import { white } from '../../styles/variables';
+import { Icon } from '../iconTypes';
+
+const IconHighlightText = ({
```
ConradJChan, line 6: Would it be more consistent with the existing `IconRegionAnnotation` if this was `IconHighlightTextAnnotation`?

src/icons/annotations/IconHighlightText.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react';
+import AccessibleSVG from '../accessible-svg';
+import { white } from '../../styles/variables';
+import { Icon } from '../iconTypes';
```
ConradJChan, line 4: 🔡

src/icons/annotations/IconHighlightText.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+const IconHighlightText = ({
+    className = '',
+    color = white,
+    height = 16,
```
ConradJChan, line 9: Seems like the other icons for annotations are a different size by default, should this be consistent?

src/icons/annotations/IconHighlightText.tsx
```diff
@@ -0,0 +1,28 @@
...
+    title,
+    width = 16,
+}: Icon): React.ReactElement => (
+    <AccessibleSVG
+        className={`bdl-IconHighlightText ${className}`}
```
ConradJChan, line 14: Should this use `classNames` to avoid the empty space by default?

src/icons/annotations/IconHighlightText.tsx
mxiao6: The other icons' sizes in this folder are not consistent, I saw 20/21/22/24

booc0mtaco:
Curious if there's a reason we are still manually importing these icons (or if they are duplicates of ones that get imported via the import script)?

mxiao6:
> Curious if there's a reason we are still manually importing these icons (or if they are duplicates of ones that get imported via the import script)?

Oh I found a duplicate one in icon folder. Good catch, thanks!

PR #2314, opened 2020-08-28
fix(content-sharing): Use marker-based paging for users and groups

sahiga:
Update users and groups retrieval in the ContentSharing Element to use marker-based paging instead of offset-based paging.

### DIFF
--- a/src/api/Groups.js
+++ b/src/api/Groups.js
@@ -4,9 +4,7 @@
  * @author Box
  */
 import noop from 'lodash/noop';
-import queryString from 'query-string';
 import Base from './Base';
-import { ERROR_CODE_FETCH_ENTERPRISE_GROUPS } from '../constants';
 import type { ElementsErrorCallback } from '../common/types/api';
 
 class Groups extends Base {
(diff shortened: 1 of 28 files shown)


src/api/Groups.js
```diff
@@ -72,33 +53,6 @@ class Groups extends Base {
...
-    ): void {
-        this.errorCode = ERROR_CODE_FETCH_ENTERPRISE_GROUPS;
-        this.get({
-            id,
-            successCallback,
-            errorCallback,
-            url: this.getGroupsInEnterpriseUrl(filterTerm),
-            requestData,
-        });
-    }
```
manishPh, line 101: Did groups and users APIs change? Can we still have offset based pagination for these endpoints?

sahiga: I added these functions, so they're not currently being used for anything else in BUIE. I can add them back in, though. Thoughts?

manishPh: @sahiga, if you are removing the same functions that you added recently, then it's ok.

src/api/MarkerBasedAPI.js
```diff
@@ -62,6 +62,7 @@ class MarkerBasedApi extends Base {
                 ...requestData,
                 marker,
                 limit,
+                usemarker: true,
```
swfree, line 65: just confirming - this change won't impact any other APIs that use this MarkerBasedAPI? i.e. if they don't have this query param implemented, it'll just be ignored?

swfree (dismissed):

src/api/MarkerBasedAPI.js
booc0mtaco: in most cases, we use camelCase, but this one is all lowercase. Is that what we have to use here?

sahiga: Good question. The `usemarker` parameter enforces the use of marker-based paging when both offset-based paging and marker-based paging are available. It shouldn't affect the other files that use this API, but to be safe, I'll move this parameter to the `MarkerBasedGroups` and `MarkerBasedUsers` files.

sahiga: @booc0mtaco Yeah, this parameter is specified as all lowercase in the docs. I was surprised by the format, too.

PR #2331, opened 2020-09-10
refactor(metadata-view): Generalize metadataFieldsToShow for item fields

manishPh:


manishPh:
This PR removes the `metadataFieldsToShow` prop from `<ContentExplorer />` and replaces it with `fieldsToShow` prop. Users should be able to provide metadata fields + item fields using the same prop. Here is what is looks like to have item fields (in addition to `name`) and metadata fields in the same view:
image

Props used for this view:
```
const mdQuery = {
    from: 'enterprise_123.templateKey',
    fields: [
        "name",
        "size",
        "description",
        "created_by",
        "metadata.enterprise_123.templateKey.type",
        "metadata.enterprise_123.templateKey.vendor",
        "metadata.enterprise_123.templateKey.amount",
        "metadata.enterprise_123.templateKey.created",
        "metadata.enterprise_123.templateKey.category",
        "metadata.enterprise_123.templateKey.approved",
    ],
    ancestor_folder_id: '0',
};
const defaultView = 'metadata';
const fieldsToShow = [
    'size',
    {key: 'created_by.login', canEdit: true, displayName: 'Created By'},
    {key: 'metadata.enterprise_123.templateKey.type', canEdit: true, displayName: 'TYPe'},
    {key: 'metadata.enterprise_123.templateKey.vendor', canEdit: false, displayName: 'AABC'},
    {key: 'metadata.enterprise_123.templateKey.amount', canEdit: false},
    'metadata.enterprise_123.templateKey.category',
    'metadata.enterprise_123.templateKey.created',
    {key: 'metadata.enterprise_123.templateKey.approved', canEdit: true, displayName: 'APPROVED'},
];

<ContentExplorer
    ...
    metadataQuery={mdQuery}
    defaultView={defaultView}
    fieldsToShow={fieldsToShow}
/>
```

src/features/metadata-based-view/MetadataBasedItemList.js
```diff
@@ -254,12 +263,25 @@ class MetadataBasedItemList extends React.Component<Props, State> {
                 );
                 break;
             default: {
-                const key = isString(metadataColumn) ? metadataColumn : metadataColumn.key;
-                const field = find(fields, ['key', key]);
-                if (!field) {
-                    return cellData;
+                const key = isString(fieldToShow) ? fieldToShow : fieldToShow.key;
+                let field;
+                let type = FIELD_TYPE_STRING;
```
sahiga, line 268: Just noticed this after looking at the updates in `MetadataBasedItemList.test.js`. Is `string` the default type because the code replaces `undefined` values with "No Value"?

sahiga (approved):

src/features/metadata-based-view/MetadataBasedItemList.js
manishPh: @sahiga the item fields that we are planning to make editable are mostly going to be `name` and `description`. These are of type `string`. But if its a metadata field, then we overwrite the value of `type` in line 279
```({ type, value, options = [] } = field);```

PR #2332, opened 2020-09-14
fix(content-sharing): style issues when using element via ES6 wrapper

ctuckz:
Box styling was missing or being overridden when using the element on a standalone page


src/features/unified-share-modal/UnifiedShareModal.scss
```diff
@@ -276,6 +283,10 @@
     }
 }
 
+.unified-share-form-container {
```
sahiga, line 286: Can we use a SUIT CSS format for this name, maybe `.bdl-UnifiedShareForm-container`?

sahiga (approved):

PR #2335, opened 2020-09-16
feat(pill-styling): support for pill styling updates in activity sidebar

danubevictoria:
Introduces handling of promises in RoundPill and ContactDatalistItem for getting the avatar url string.

### DIFF
--- a/src/components/contact-datalist-item/ContactDatalistItem.js
+++ b/src/components/contact-datalist-item/ContactDatalistItem.js
@@ -29,7 +29,7 @@ class ContactDatalistItem extends React.PureComponent<Props, State> {
     /**
      * Success handler for getting avatar url
      *
-     * @param {string} avatarUrl the user avatar url
+     * @param {string} [avatarUrl] the user avatar url
      */
     getAvatarUrlHandler = (avatarUrl: ?string) => {
         if (this.isMounted) {
@@ -40,19 +40,17 @@ class ContactDatalistItem extends React.PureComponent<Props, State> {
     };
 
     /**
-     * Gets the avatar URL for the user from the getAvatarUrl prop
+     * Gets the avatar URL for the user from the getContactAvatarUrl prop
      *
-     * @return {string} the avatar url string
+     * @return {void}
      */
     getAvatarUrl() {
         const { getContactAvatarUrl, id } = this.props;
-        const returnVal = getContactAvatarUrl && id ? Promise.resolve(getContactAvatarUrl({ id })) : undefined;
-
-        if (returnVal) {
-            returnVal.then(this.getAvatarUrlHandler).catch(() => {
+        Promise.resolve(getContactAvatarUrl && id ? getContactAvatarUrl({ id }) : undefined)
+            .then(this.getAvatarUrlHandler)
+            .catch(() => {
                 // noop
             });
-        }
     }
 
     componentDidMount() {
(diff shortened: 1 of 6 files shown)


src/components/contact-datalist-item/ContactDatalistItem.js
```diff
@@ -7,31 +7,84 @@ import DatalistItem from '../datalist-item';
...
+    };
+
+    /**
+     * Gets the avatar URL for the user from the getAvatarUrl prop
+     *
+     * @return {string} the avatar url string
+     */
+    getAvatarUrl() {
+        const { getContactAvatarUrl, id } = this.props;
+        const returnVal = getContactAvatarUrl && id ? Promise.resolve(getContactAvatarUrl({ id })) : undefined;
```
booc0mtaco, line 49: what happens if `getContactAvatarUrl` is a promise that ultimately rejects?

danubevictoria: we changed this to catch and basically do nothing as we thought if the promise rejects, the avatarUrl should just remain undefined and the fallback should be the user initials.

src/components/contact-datalist-item/ContactDatalistItem.js
```diff
@@ -7,31 +7,86 @@ import DatalistItem from '../datalist-item';
...
+            this.setState({
+                avatarUrl,
+            });
+        }
+    };
+
+    /**
+     * Gets the avatar URL for the user from the getAvatarUrl prop
+     *
+     * @return {string} the avatar url string
```
booc0mtaco, line 45: This isn't true. It will not return anything; instead it will trigger a promise that actually returns the string.

src/components/pill-selector-dropdown/RoundPill.js
```diff
@@ -24,61 +24,123 @@ type Props = {
...
+};
+
 const RemoveButton = ({ onClick, ...rest }: { onClick: () => any }) => (
     <X {...rest} aria-hidden="true" onClick={onClick} />
 );
 
-const RoundPill = ({
-    getPillImageUrl,
-    isDisabled = false,
-    isSelected = false,
```
booc0mtaco, line 34: `isSelected` is no longer being defaulted to false

src/components/contact-datalist-item/ContactDatalistItem.js
```diff
@@ -7,31 +7,86 @@ import DatalistItem from '../datalist-item';
...
+     */
+    getAvatarUrl() {
+        const { getContactAvatarUrl, id } = this.props;
+        const returnVal = getContactAvatarUrl && id ? Promise.resolve(getContactAvatarUrl({ id })) : undefined;
```
booc0mtaco, line 49: This isn't a return value; ideally it would always be a promise so we can have consistent types thru the rest of this method.

We could do `Promise.resolve(getContactAvatarUrl && id ? getContactAvatarUrl({ id }) : undefined)` or similar, to avoid the if-statement below.

In fact, that if now doesn't help as much because it will always be true in the same way if a promise is set.

src/components/contact-datalist-item/ContactDatalistItem.js
```diff
@@ -7,31 +7,86 @@ import DatalistItem from '../datalist-item';
...
+    state = {
+        avatarUrl: undefined,
+    };
+
+    isMounted: boolean = false;
+
+    /**
+     * Success handler for getting avatar url
+     *
+     * @param {string} avatarUrl the user avatar url
```
booc0mtaco, line 32: change `{string}` to the JSDoc equivalent for maybe param

src/components/contact-datalist-item/ContactDatalistItem.js
danubevictoria: my mistake forgot to change that

booc0mtaco (approved):

PR #2336, opened 2020-09-17
feat(animation): convert to fade-in animation across flyout components

booc0mtaco:


src/styles/mixins/_animation.scss
```diff
@@ -1,5 +1,26 @@
+@import '../constants/animation';
+
 @mixin bdl-transition($time-length, $properties...) {
-    transition-timing-function: cubic-bezier(0, 0, .6, 1);
+    transition-timing-function: $bdl-timingFunction;
     transition-duration: $time-length;
     transition-property: $properties;
 }
+
+@mixin bdl-animation($time-length) {
```
jstoffan, line 9: Should this have a more specific name?

src/styles/common/_animations.scss
```diff
@@ -7,21 +9,3 @@
         opacity: 1;
     }
 }
-
-@mixin bdl-fadeIn($fade-in-time) {
```
jstoffan, line 11: Is this used anywhere else or in any other projects? If so, removing it would be a breaking change.

jstoffan (dismissed):
Are there any visual changes or new features? Or is this more of a `refactor`?

src/styles/common/_animations.scss
booc0mtaco: i searched and only saw internal references to it here, and no references in known projects. I think it should technically be a breaking change, but wanted to discuss. Two options stand out:
 
* mark this as breaking with a note on why
* `@include` the renamed and moved mixin here with a [`warn`](https://sass-lang.com/documentation/at-rules/warn) saying what to use instead.

src/styles/mixins/_animation.scss
booc0mtaco: yah, a name could better indicate what this is for. it standardizes the fade-in animation instead of the previous "skew and fade" animation when opening any overlays (menus, flyout headers, and any other overlays)

jstoffan (approved):

PR #2343, opened 2020-09-23
fix(upload): add support for detecting and handling package files

booc0mtaco:
This change allows the upload manager to properly identify it has received a [package file](https://en.wikipedia.org/wiki/Package_(macOS)), and attempt to upload it. 

In Safari, this will fail predictably, so we allow detection of the error type, and show a specific error message detailing the issue with a workaround.

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -228,6 +228,8 @@ export const ERROR_CODE_UPLOAD_CHILD_FOLDER_FAILED = 'child_folder_failed_upload
 export const ERROR_CODE_UPLOAD_STORAGE_LIMIT_EXCEEDED = 'storage_limit_exceeded';
 export const ERROR_CODE_UPLOAD_FILE_SIZE_LIMIT_EXCEEDED = 'file_size_limit_exceeded';
 export const ERROR_CODE_UPLOAD_PENDING_APP_FOLDER_SIZE_LIMIT = 'pending_app_folder_size_limit';
+export const ERROR_CODE_UPLOAD_BAD_DIGEST = 'bad_digest';
+export const ERROR_CODE_UPLOAD_FAILED_PACKAGE = 'failed_package_upload';
 export const ERROR_CODE_FETCH_ACTIVITY = 'fetch_activity_error';
 export const ERROR_CODE_FETCH_ANNOTATION = 'fetch_annotation_error';
 export const ERROR_CODE_FETCH_ANNOTATIONS = 'fetch_annotations_error';
@@ -379,10 +381,11 @@ export const HTTP_OPTIONS: 'OPTIONS' = 'OPTIONS';
 export const HTTP_HEAD: 'HEAD' = 'HEAD';
 
 /* ------------------ HTTP Codes  ---------------------- */
+export const HTTP_STATUS_CODE_BAD_REQUEST: 400 = 400;
+export const HTTP_STATUS_CODE_UNAUTHORIZED: 401 = 401;
 export const HTTP_STATUS_CODE_FORBIDDEN: 403 = 403;
 export const HTTP_STATUS_CODE_NOT_FOUND: 404 = 404;
 export const HTTP_STATUS_CODE_CONFLICT: 409 = 409;
-export const HTTP_STATUS_CODE_UNAUTHORIZED: 401 = 401;
 export const HTTP_STATUS_CODE_RATE_LIMIT: 429 = 429;
 export const HTTP_STATUS_CODE_INTERNAL_SERVER_ERROR: 500 = 500;
 export const HTTP_STATUS_CODE_NOT_IMPLEMENTED: 501 = 501;
--- a/src/elements/content-uploader/ContentUploader.js
+++ b/src/elements/content-uploader/ContentUploader.js
@@ -20,10 +20,12 @@ import {
     getDataTransferItemId,
     getFileId,
     getFileFromDataTransferItem,
+    getPackageFileFromDataTransferItem,
     getFile,
     getFileAPIOptions,
     getDataTransferItemAPIOptions,
     isDataTransferItemAFolder,
+    isDataTransferItemAPackage,
     isMultiputSupported,
 } from '../../utils/uploads';
 import DroppableContent from './DroppableContent';
@@ -215,6 +217,7 @@ class ContentUploader extends Component<Props, State> {
             return;
         }
 
+        // TODO: this gets called unnecessarily (upon each render regardless of the queue not changing)
         this.addFilesWithOptionsToUploadQueueAndStartUpload(files, dataTransferItems);
     }
 
@@ -385,10 +388,13 @@ class ContentUploader extends Component<Props, State> {
 
         const folderItems = [];
         const fileItems = [];
+        const packageItems = [];
 
         Array.from(dataTransferItems).forEach(item => {
             const isDirectory = isDataTransferItemAFolder(item);
-            if (isDirectory && isFolderUploadEnabled) {
+            if (isDataTransferItemAPackage(item)) {
+                packageItems.push(item);
+            } else if (isDirectory && isFolderUploadEnabled) {
                 folderItems.push(item);
             } else if (!isDirectory) {
                 fileItems.push(item);
@@ -396,6 +402,7 @@ class ContentUploader extends Component<Props, State> {
         });
 
         this.addFileDataTransferItemsToUploadQueue(fileItems, itemUpdateCallback);
+        if (packageItems.length) this.addPackageDataTransferItemsToUploadQueue(packageItems, itemUpdateCallback);
         this.addFolderDataTransferItemsToUploadQueue(folderItems, itemUpdateCallback);
     };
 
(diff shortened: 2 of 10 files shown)


src/elements/content-uploader/__tests__/ContentUploader.test.js
```diff
@@ -119,6 +119,40 @@ describe('elements/content-uploader/ContentUploader', () => {
             const expected = { yoyo: true, yoyo_0_10000: true };
             expect(wrapper.state().itemIds).toEqual(expected);
         });
+
+        test.only('should handle accepting package "files" separate from folders', () => {
```
tjgupta, line 123: I think you don't want `test.only()` here right?

booc0mtaco: good catch!

src/elements/content-uploader/ContentUploader.js
```diff
@@ -385,17 +390,21 @@ class ContentUploader extends Component<Props, State> {
...
+                packageItems.push(item);
+            } else if (isDirectory && isFolderUploadEnabled) {
                 folderItems.push(item);
             } else if (!isDirectory) {
                 fileItems.push(item);
             }
         });
 
         this.addFileDataTransferItemsToUploadQueue(fileItems, itemUpdateCallback);
+        if (packageItems.length) this.addPackageDataTransferItemsToUploadQueue(packageItems, itemUpdateCallback);
```
ConradJChan, line 407: We don't check for length before calling the other `add*DataTransferItemsToUploadQueue`, do we need to here?

src/elements/content-uploader/progressCellRenderer.js
```diff
@@ -65,6 +71,9 @@ export default () => ({ rowData }: Props) => {
         case STATUS_STAGED:
             return <ItemProgress {...rowData} />;
         case STATUS_ERROR:
+            if (Browser.isSafari() && code === ERROR_CODE_UPLOAD_BAD_DIGEST && file.name.indexOf('.zip') !== -1) {
```
ConradJChan, line 74: Should we destructure out the name on L62?

ConradJChan, line 74: So this is for `.zip` files and not `.pkg` files?

src/utils/Browser.js
```diff
@@ -36,6 +36,16 @@ class Browser {
         return /Trident/i.test(Browser.getUserAgent());
     }
 
+    /**
+     * Returns whether browser is Safari.
+     *
+     * @return {boolena} Whether browser is IE
```
ConradJChan, line 42: `IE` and `boolena`

src/utils/__tests__/Browser.test.js
```diff
@@ -34,6 +34,24 @@ describe('util/Browser/isIE()', () => {
...
+    test('should return false if not Safari', () => {
+        browser.getUserAgent = jest.fn().mockReturnValueOnce('Trident');
+        expect(browser.isSafari()).toBeFalsy();
+    });
+
+    test('should not incorrectly identify Blink as WebKit', () => {
+        browser.getUserAgent = jest
+            .fn()
+            .mockReturnValueOnce('AppleWebKit/7.8 (KHTML, like Gecko) Chrome/1.2.3.4 Safari/5.6');
+        expect(browser.isSafari()).toBeFalsy();
```
ConradJChan, line 51: Would it make sense to convert this to a `test.each` to avoid the code repetition?

src/utils/uploads.js
```diff
@@ -202,6 +202,22 @@ function isDataTransferItemAFolder(itemData: UploadDataTransferItemWithAPIOption
...
+ * @returns {boolean}
+ */
+function isDataTransferItemAPackage(itemData: UploadDataTransferItemWithAPIOptions | DataTransferItem): boolean {
+    const item = getDataTransferItem(itemData);
+    const isDirectory = isDataTransferItemAFolder(item);
+
+    if (isDirectory && item.kind === 'file') {
+        return true;
+    }
+    return false;
```
ConradJChan, line 218: Could this be simplified to `return isDirectory && item.kind === 'file'`?

src/utils/uploads.js
```diff
@@ -243,6 +259,35 @@ async function getFileFromDataTransferItem(
...
+    const itemFile = item.getAsFile();
+
+    if (doesDataTransferItemContainAPIOptions(itemData)) {
+        return Promise.resolve({
+            file: ((itemFile: any): UploadFile),
+            options: getDataTransferItemAPIOptions(itemData),
+        });
+    }
+
+    return Promise.resolve(itemFile);
```
ConradJChan, line 288: Are these `Promise.resolve` required?

src/elements/content-uploader/progressCellRenderer.js
booc0mtaco: There are two kinds of "package" files in macOS. The installer type, which are suffixed with `.pkg`, and a File System concept which is essentially a directory with a plist and UTI in macOS that says it should be treated as a single file. This deals with the latter.

In our case, Safari knows that Keynote files are supposed to be a single file, while other browsers do not. Safari does some magic to convert the upload into a zip file.

src/elements/content-uploader/ContentUploader.js
booc0mtaco: If I had time, I would like to revisit this section. Currently there are some bugs in how this is handled, I think. I noticed that the lists never get cleared, so it ends up with duplicates of previously added data transfer items. I added this `if` to make the debugging less noisy.

I can take this one out, but I do wonder if the better change is to clean the queues and prevent duplicates. Will save that for another day.

src/utils/uploads.js
booc0mtaco: They may not be. it has been a bit since I added this in, so let me look again.

src/elements/content-uploader/progressCellRenderer.js
booc0mtaco: I left this one as a key reference b/c we already had a destructured `name` from rowData in this file.

src/utils/uploads.js
booc0mtaco: @ConradJChan I made this one return promises like this b/c the associated functions: `getFileFromEntry`, `getFileFromDataTransferItem` are all async. This new function already has all the info it needs, but I wanted to guard against introducing a non-async function along side the others, so we wouldn't have a mismatch in return types anywhere. 

There may be a tidier way to write these return values, e.g., having a single `Promise.resolve` with the function's logic in there.

ConradJChan (approved):

booc0mtaco:
Adding some helpful links for posterity
* https://en.wikipedia.org/wiki/Package_(macOS)
* https://en.wikipedia.org/wiki/Uniform_Type_Identifier

src/constants.js
```diff
@@ -228,6 +228,8 @@ export const ERROR_CODE_UPLOAD_CHILD_FOLDER_FAILED = 'child_folder_failed_upload
 export const ERROR_CODE_UPLOAD_STORAGE_LIMIT_EXCEEDED = 'storage_limit_exceeded';
 export const ERROR_CODE_UPLOAD_FILE_SIZE_LIMIT_EXCEEDED = 'file_size_limit_exceeded';
 export const ERROR_CODE_UPLOAD_PENDING_APP_FOLDER_SIZE_LIMIT = 'pending_app_folder_size_limit';
+export const ERROR_CODE_UPLOAD_BAD_DIGEST = 'bad_digest';
+export const ERROR_CODE_UPLOAD_FAILED_PACKAGE = 'failed_package_upload';
```
sahiga, line 232: Do we need types for these constants?

src/elements/content-uploader/progressCellRenderer.js
```diff
@@ -5,18 +5,22 @@
...
 import {
     ERROR_CODE_UPLOAD_FILE_SIZE_LIMIT_EXCEEDED,
     ERROR_CODE_ITEM_NAME_IN_USE,
     ERROR_CODE_ITEM_NAME_INVALID,
     ERROR_CODE_UPLOAD_PENDING_APP_FOLDER_SIZE_LIMIT,
     ERROR_CODE_UPLOAD_STORAGE_LIMIT_EXCEEDED,
     ERROR_CODE_UPLOAD_CHILD_FOLDER_FAILED,
-    STATUS_ERROR,
+    ERROR_CODE_UPLOAD_BAD_DIGEST,
+    ERROR_CODE_UPLOAD_FAILED_PACKAGE,
```
sahiga, line 20: Minor: can we alphabetize these imports?

src/utils/__tests__/uploads.test.js
```diff
@@ -184,6 +186,82 @@ describe('util/uploads', () => {
...
+            expect(isDataTransferItemAPackage(itemData)).toBeFalsy();
+        });
+
+        test('should be false if data transfer item thinks it is a file', () => {
+            const fileEntry = {
+                isDirectory: false,
+                isFile: true,
+            };
+
+            const folderItem = {
```
sahiga, line 232: `fileItem`? Also, could this test be combined with the test above into a `test.each` loop?

src/constants.js
booc0mtaco: I would prefer a PR to update all the types at once so we dont have a piecemeal set (and avoid adding any more complexity to this particular change).

But yah, we should move in that direction for our constants IMO.

src/elements/content-uploader/progressCellRenderer.js
booc0mtaco: Same as above; you will note that some are already out of sequence, unfortunately. we can have a "best practices" PR to clean them up so future changes conform to the tidy content.

src/utils/__tests__/uploads.test.js
booc0mtaco: this one is a bit funny. for package files, it identifies as both a file AND a folder, so I used the confusing name to represent this on purpose. maybe the added comments could be clearer about that in the code.

PR #2353, opened 2020-10-06
feat(usm): Allow red pills for external collab restrictions

jpfranco:
This is a partial update that prepares the USM to be able display red error pills, which will later be used to highlight restricted external collabs once the external collab restriction flow is updated. For now the main goal is to allow the `EmailForm` on the invite section to operate with `allowInvalidPills` enabled, so the red pills will not be visible initially. This change should be backwards compatible since the existing behavior of the USM was not altered.

Note: Please let me know if there's a preferred approach for submitting partial updates such as these. 

Screen Shot 2020-10-05 at 6 58 22 PM


src/features/unified-share-modal/ContactsField.js
```diff
@@ -127,6 +127,19 @@ class ContactsField extends React.Component<Props, State> {
...
     debouncedGetContacts = debounce(this.getContactsPromise, 200);
 
+    handleParseItems = (inputValue: string): Array<string> => {
+        const { validator } = this.props;
+
+        // ContactField allows invalid pills to be displayed in
+        // in some cases (e.g., when user is external and external
+        // collab is restricted). We don't allow, however, invalid
+        // emails from the pill selector input to be turned into pills.
+        const emails = parseEmails(inputValue);
```
jpfranco, line 137: Previously, invalid emails on the pill selector input were filtered out by `PillSelectorDropdown` automatically, but that filtering step is skipped when `allowInvalidPills` is enabled:

https://github.com/box/box-ui-elements/blob/master/src/components/pill-selector-dropdown/PillSelectorDropdown.js#L117

By filtering out invalid emails on the pill parser, we can keep the existing behavior of preventing invalid emails from generating pills, but also allow invalid pills to be shown for other scenarios.

src/features/unified-share-modal/EmailForm.js
```diff
@@ -205,14 +207,27 @@ class EmailForm extends React.Component<Props, State> {
         const { intl } = this.props;
         let contactsFieldError = '';
 
-        if (text && !this.isValidEmail(text)) {
+        if (text && !emailValidator(text)) {
```
jpfranco, line 210: `isValidContactPill` has a slightly different purpose than `isValidEmail`, so I think it no longer makes sense to call it here.

src/features/unified-share-modal/EmailForm.js
```diff
@@ -205,14 +207,27 @@ class EmailForm extends React.Component<Props, State> {
...
             contactsFieldError = intl.formatMessage(commonMessages.invalidEmailError);
         }
         this.setState({ contactsFieldError });
     };
 
-    isValidEmail = (text: string): boolean => {
-        return emailValidator(text);
+    isValidContactPill = (contactPill: string | Contact): boolean => {
```
jpfranco, line 216: Another side-effect of enabling `allowInvalidPills` is that the validator also gets called with pill objects (contacts in this case) instead of just strings.

src/features/unified-share-modal/EmailForm.js
```diff
@@ -49,6 +50,7 @@ type Props = {
     recommendedSharingTooltipCalloutName?: ?string,
     selectedContacts: Array<Contact>,
     sendButtonProps?: Object,
+    shouldInvalidateExternalCollabs?: boolean,
```
jpfranco, line 53: Note that this is falsey by default so it will not have any effect yet.

src/features/unified-share-modal/ContactsField.js
```diff
@@ -127,6 +127,19 @@ class ContactsField extends React.Component<Props, State> {
...
+        const validEmails = emails.filter(email => validator(email));
```
jpfranco, line 138: I wasn't sure where to put this since the parsing seems to be the concern of `ContactsField`, but email validation is mostly handled on `EmailForm`. I thought calling the `validator` provided by `EmailForm` would be a good compromise, but let me know if you have any suggestions.

jpfranco: Also not sure if this qualifies as a breaking change because, even though this does not change the behavior of USM itself, it does slightly change the behavior of this component when used on its own (not sure if we currently have use cases for that). Happy to mark the commit as "breaking" if needed.

alexkrolick: Sub-component APIs in `features/*/**` aren't considered public

src/features/unified-share-modal/__tests__/ContactsField.test.js
```diff
@@ -294,6 +296,42 @@ describe('features/unified-share-modal/ContactsField', () => {
...
+                },
+                {
+                    inputValue: '123',
+                    expectedItems: [],
+                },
+                {
+                    inputValue: 'not_an_email',
+                    expectedItems: [],
+                },
+            ].forEach(({ inputValue, expectedItems }) => {
```
sahiga, line 329: Can we use `test.each` here instead?

src/features/unified-share-modal/__tests__/EmailForm.test.js
```diff
@@ -311,14 +311,42 @@ describe('features/unified-share-modal/EmailForm', () => {
...
+            const wrapper = getWrapper();
+
+            expectedContacts.forEach(contact => {
+                const isValidContactPill = wrapper.instance().isValidContactPill(contact);
+
+                expect(isValidContactPill).toBe(true);
+            });
+        });
+
+        test('should invalidate parsed contact pills when shouldInvalidateExternalCollabs is true and contact is an external user', () => {
```
sahiga, line 333: The description only refers to the case in which `shouldInvalidateExternalCollabs` is `true`, but this test also covers the case in which `shouldInvalidateExternalCollabs` is `false`. Should it be split into two tests in a `test.each` loop?

sahiga (approved):

PR #2356, opened 2020-10-09
feat(multi-select-field): support dropdown positioning outside of parent

danubevictoria:


src/components/popper/PopperComponent.js
```diff
@@ -25,12 +26,12 @@ const PopperComponent = (props: Props) => {
...
             <Reference>{({ ref }) => React.cloneElement(reference, { ref })}</Reference>
             {isOpen && (
-                <Popper placement={popperPlacement} modifiers={modifiers}>
+                <Popper placement={popperPlacement} modifiers={modifiers} positionFixed={positionFixed}>
                     {({ ref, style, placement, scheduleUpdate }) => {
                         const { style: contentStyles } = popperContent.props;
                         return React.cloneElement(popperContent, {
                             ref,
-                            style: { ...contentStyles, ...style },
+                            style: { ...style, ...contentStyles },
```
booc0mtaco, line 34: Was there a reason the order of these was changed?

src/components/popper/__tests__/PopperComponent.test.js
```diff
@@ -50,4 +50,18 @@ describe('components/popper/PopperComponent', () => {
         expect(popperWrapper.length).toBe(1);
         expect(popperWrapper.prop('placement')).toBe('bottom-end');
     });
+
+    test('should apply the positionFixed to the Popper', () => {
```
booc0mtaco, line 54: this test is testing react itself. you create the wrapper on line 56, then test what is on line 56. we want to test the associated logic, so it would be good for a test to cover when `isOpen` is false but `positionFixed` is true.

src/components/select-field/BaseSelectField.js
```diff
@@ -68,6 +68,8 @@ type Props = {
     options: Array<SelectOptionProp>,
     /** The select button text shown when no options are selected. */
     placeholder?: string | React.Element<any>,
+    /** Boolean to determine whether the dropdown should be in fixed position */
+    positionFixed?: boolean,
```
booc0mtaco, line 72: We have had this crop up a few times before; when a component wraps some library component, how should we handle that libraries built-in options?

Here, we have a one-off case handled. `positionFixed` is added directly to the signature of BaseSelectField, but it is actually determined by `react-popper`. Popper props are mixed with this component's props. 

IMO, it is best to provide an object prop for the library's options. that way, there are fewer future changes to the component, and it is easier to refactor and type. for instance, popper may have a type for the options, which we could use here.

[`Popper.Defaults`](https://popper.js.org/docs/v1/#popperdefaults--codeobjectcode) is defined here.

src/components/select-field/BaseSelectField.js
```diff
@@ -83,6 +85,7 @@ type Props = {
 type State = {
     activeItemID: ?string,
     activeItemIndex: number,
+    dropdownStyle: Object,
```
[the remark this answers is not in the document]
danubevictoria: The Popper component uses popper.js to do some auto styling. For my use case I need to remove some of the styling and add the width of the dropdown to be the same as the reference button. The downside is passing in a custom style but popper.js supports using 'fixed' as a positioning strategy. And it is included in the Popper interface.

https://popper.js.org/docs/v2/constructors/#strategy

src/components/popper/PopperComponent.js
danubevictoria: The popper does some auto styling including adding a `top`, `left`, and `transform` properties that affect how the dropdown menu is positioned on the screen. With the dropdown menu, I want to override these only if `contentStyles` are passed in. Otherwise the popper styling would override what I pass in.

src/components/select-field/BaseSelectField.js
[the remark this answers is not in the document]
danubevictoria: I want to assign the width of the dropdown menu to be the same as the width of the reference button, which the BaseSelectField would know upon clicking the button.

src/components/select-field/BaseSelectField.js
danubevictoria: Looked at `Popper.Defaults` and by passing in these static properties, we are passing these into the Popper.Defaults object. The Popper component accepts the following properties defined here: https://www.npmjs.com/package/react-popper/v/1.3.5

Added it to BaseSelectField as an optional prop to leave the responsibility of overriding the default to whoever is implementing it. in this case, EUA would be the one implementing the MultiSelect component and can pass down whether they want to override `react-popper`'s default positioning.

This is also following the current design of `placement` for `dropdownPlacement` and `modifiers` for `dropdownModifiers`.

src/components/select-field/BaseSelectField.js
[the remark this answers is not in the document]
danubevictoria: Ah youre right looks like v1.x only has option to use positionFixed as a static property.

src/components/popper/__tests__/PopperComponent.test.js
danubevictoria: When `isOpen` is false, we should not render the inner `<Popper>` component. Test is already implemented here: https://github.com/box/box-ui-elements/blob/master/src/components/popper/__tests__/PopperComponent.test.js#L16

So here I'm trying to test that when we pass in `positionFixed` to `<PopperComponent>` that we maintain that prop in the inner `<Popper>` (which is the dropdown menu).

src/components/select-field/BaseSelectField.js
booc0mtaco: What I mean is that all the other props for BaseSelectField are specific to BaseSelectField. `dropdownPlacement` and `dropdownModifiers` are derived from props passed in that are not Popper-specific. By adding one of Popper's expected parameters directly to this component's props, we do to things:

- We tie BaseSelectField's implementation to the external library
- We establish the pattern of directly copying all future defaults into BaseSelectField's API

An object would make sense if `fixedPosition` were grouped in with other things, but it appears to be at the same level as the other defaults, as props on `<Popper/>`.

To match what is being done in BaseSelectField, we might add an `isPositionFixed` optional prop, and do a similar check on ln592 that we do for the other popper values, to create the popper-specific prop for `<PopperComponent/>`.

The other benefit is that we can avoid some of the confusion if we upgrade to popper 2.x, since they change the parameters around a lot.

src/components/select-field/BaseSelectField.js
```diff
@@ -200,11 +205,11 @@ class BaseSelectField extends React.Component<Props, State> {
         }
     };
 
-    handleButtonClick = () => {
+    handleButtonClick = (event: SyntheticEvent<HTMLElement>) => {
         if (this.state.isOpen) {
```
booc0mtaco, line 209: as a more terse alternative, `this[this.state.isOpen ? 'closeDropdown' : 'openDropdown'](event)`

src/components/select-field/__tests__/BaseSelectField.test.js
```diff
@@ -1000,25 +1022,60 @@ describe('components/select-field/BaseSelectField', () => {
...
+                    left: 'unset',
+                    minWidth: 'unset',
+                    transform: 'none',
+                    willTransform: 'unset',
+                    width: 120,
+                },
+            ],
+            [{}, {}],
+        ])('%o should set dropdownStyle state when position is fixed', (event, expected) => {
+            Element.prototype.getBoundingClientRect = jest.fn(() => {
```
booc0mtaco, line 1048: we probably should not do this; this is modifying *all* instances of element's definition of `getBoundingClientRect`. There may be a way to use `mockReturnValueOnce` here

PR #2362, opened 2020-10-14
feat(assets): Add ribbon icon asset

sanjui9000:
Add ribbon icon asset


sanjui9000:
@booc0mtaco I did sign the CLA but its still showing that I did not. Also, my username is sanjui9000. Not sure why its showing sveliventi in CLAassistant notification

PR #2365, opened 2020-10-15
feat(label-pill): update design to match latest from design

booc0mtaco:


booc0mtaco:
Screen Shot 2020-10-14 at 5 33 15 PM

src/components/label-pill/LabelPill.scss
```diff
@@ -3,11 +3,11 @@
 .bdl-LabelPill {
     color: $bdl-gray;
     font-weight: bold;
-    border: 1px solid;
+    letter-spacing: .3px;
```
alexkrolick, line 6: Isn't this inheriting from the default? https://github.com/box/box-ui-elements/blob/939e089d9b4ac71e6081949f2f5a3a4c5faebdf4/src/elements/common/base.scss#L13

booc0mtaco: yah, it is. I can drop this line out. I was debating whether components should re-specify such rules, in case the base changes. In such a scenario, maybe it's better for there to be overrides, than unintended regressions. Thoughts?

alexkrolick: I think we should add a design token for it, and reference that in the base CSS file, but not here

booc0mtaco: I will drop this one, and let it inherit. At some point, we can take the base font properties on `body, html {}` and set those to new tokens. Save that for a separate PR since it is unrelated to this component.

alexkrolick (approved):

PR #2372, opened 2020-10-17
fix(uploads): add detection to file type when determining package type

booc0mtaco:


src/utils/__tests__/uploads.test.js
```diff
@@ -260,6 +261,25 @@ describe('util/uploads', () => {
 
             expect(isDataTransferItemAPackage(itemData)).toBeTruthy();
         });
+
+        test('should be true if data transfer item has both identifies a directory but only has kind = file', () => {
```
sahiga, line 265: Should this say `false`?

sahiga (approved):

PR #2375, opened 2020-10-19
feat(usm): Adding UI for business justifications

jpfranco:
Partial update that adds the UI components needed for the new Business Justifications flow for external collab restrictions. The components have been integrated with `EmailForm`, but the `UnifiedShareForm` updates needed to integrate with the apis and EUA are still pending. 

**Note:** There is also some pending UI polish that will be handled separately. The updates will not yet be visible and should be backwards compatible.

**After submitting with restricted external collabs:**

Restricted External Collabs

**Restricted collabs email tooltip:**

Restricted Collabs Tooltip

**Submitting without selecting justification:**
(Note: Tooltip margin needs to be fixed)

Justification Required Error

**After justification reason is selected:**

Justification Selected

### DIFF
--- a/src/components/pill-selector-dropdown/RoundPill.scss
+++ b/src/components/pill-selector-dropdown/RoundPill.scss
@@ -30,6 +30,13 @@
     }
 }
 
+.bdl-RoundPill-avatar,
+.bdl-RoundPill-closeBtn {
+    path {
+        fill: $bdl-gray-50;
+    }
+}
+
 .bdl-RoundPill-text {
     flex-grow: 1;
     flex-shrink: 1;
@@ -52,12 +59,18 @@
 }
 
 .bdl-RoundPill--warning {
+    background-color: $bdl-yellorange-10;
+    border: 1px solid $bdl-yellorange-50;
+
     .bdl-Avatar-externalBadge {
         border-color: $bdl-yellorange-10;
     }
 }
 
 .bdl-RoundPill--error {
+    background-color: $bdl-watermelon-red-10;
+    border: 1px solid $bdl-watermelon-red-50;
+
     .bdl-Avatar-externalBadge {
         border-color: $bdl-watermelon-red-10;
     }
--- a/src/components/pill-selector-dropdown/RoundPill.scss
+++ b/src/components/pill-selector-dropdown/RoundPill.scss
@@ -80,6 +80,26 @@
     .bdl-Avatar-externalBadge {
         border-color: $bdl-gray-20;
     }
+
+    &.bdl-RoundPill--warning {
+        background-color: $bdl-yellorange-20;
+        border-color: $bdl-yellorange;
+        box-shadow: none;
+
+        .bdl-Avatar-externalBadge {
+            border-color: $bdl-yellorange-20;
+        }
+    }
+
+    &.bdl-RoundPill--error {
+        background-color: $bdl-watermelon-red-20;
+        border-color: $bdl-watermelon-red;
+        box-shadow: none;
+
+        .bdl-Avatar-externalBadge {
+            border-color: $bdl-watermelon-red-20;
+        }
+    }
 }
 
 .bdl-RoundPill--disabled {
--- a/src/features/unified-share-modal/ContactRestrictionNotice.js
+++ b/src/features/unified-share-modal/ContactRestrictionNotice.js
@@ -1,0 +1,139 @@
+// @flow
+
+import * as React from 'react';
+import noop from 'lodash/noop';
+import getProp from 'lodash/get';
+import uniqueId from 'lodash/uniqueId';
+import { FormattedMessage, injectIntl } from 'react-intl';
+import type { InjectIntlProvidedProps } from 'react-intl';
+
+import FormattedCompMessage from '../../components/i18n/FormattedCompMessage';
+import Param from '../../components/i18n/Param';
+import Label from '../../components/label';
+import PlainButton from '../../components/plain-button';
+import InlineNotice from '../../components/inline-notice';
+import LoadingIndicator from '../../components/loading-indicator/LoadingIndicator';
+import SingleSelectField from '../../components/select-field/SingleSelectField';
+
+import ContactsEmailsTooltip from './ContactsEmailsTooltip';
+import messages from './messages';
+
+import type { SelectOptionProp } from '../../components/select-field/props';
+import type { contactType as Contact } from './flowTypes';
+
+import './ContactRestrictionNotice.scss';
+
+type Props = {
+    error?: React.Node,
+    isLoading?: boolean,
+    justificationReasons: Array<SelectOptionProp>,
+    onRemoveRestrictedExternalContacts: () => void,
+    onSelectJustificationReason: (justificationReasonOption: SelectOptionProp) => void,
+    restrictedExternalEmails: Array<string>,
+    selectedContacts: Array<Contact>,
+    selectedJustificationReason: ?SelectOptionProp,
+} & InjectIntlProvidedProps;
+
+const ContactRestrictionNotice = ({
+    error,
+    intl,
+    isLoading,
+    justificationReasons,
+    onRemoveRestrictedExternalContacts,
+    onSelectJustificationReason,
+    restrictedExternalEmails,
+    selectedContacts,
+    selectedJustificationReason,
+}: Props) => {
+    const compMessageId = uniqueId('compMessage');
+    const restrictedExternalContacts = selectedContacts.filter(({ value }) => restrictedExternalEmails.includes(value));
+    const restrictedExternalContactCount = restrictedExternalContacts.length;
+
+    if (!restrictedExternalContactCount) {
+        return null;
+    }
+
+    const RemoveButton = ({ children }: { children: React.Node }) => (
+        <PlainButton
+            className="bdl-ContactRestrictionNotice-removeBtn"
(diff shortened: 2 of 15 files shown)


booc0mtaco:
@jpfranco can you attach a fer relevant screenshots?

jpfranco:
@booc0mtaco sure. I submitted this as a draft PR since it's not yet ready for review, but added the screenshots.

src/features/unified-share-modal/ContactRestrictionNotice.js
```diff
@@ -0,0 +1,139 @@
...
+}: Props) => {
+    const compMessageId = uniqueId('compMessage');
+    const restrictedExternalContacts = selectedContacts.filter(({ value }) => restrictedExternalEmails.includes(value));
+    const restrictedExternalContactCount = restrictedExternalContacts.length;
+
+    if (!restrictedExternalContactCount) {
+        return null;
+    }
+
+    const RemoveButton = ({ children }: { children: React.Node }) => (
```
jpfranco, line 56: Mostly to avoid duplicating the props. I think this does not justify a separate component though.

src/features/unified-share-modal/ContactRestrictionNotice.js
```diff
@@ -0,0 +1,139 @@
...
+            {children}
+        </PlainButton>
+    );
+
+    // TODO:
+    // Switch to react-intl v3+ api once migration is complete. FormattedCompMessage is now
+    // deprecated and has some issues with components nested within <Plural/>, hence why we
+    // use two messages and use unique keys as a workaround.
+
+    const noticeDescriptionSingular = (
```
jpfranco, line 71: Unfortunately ran into some issues with this deprecated component, which we'll have to work around since there's currently no viable alternative. Note that `<Plural />` requires you to specify the complete message for both cases, so the standard component usage is almost as verbose as the workaround.

src/features/unified-share-modal/EmailForm.js
```diff
@@ -48,9 +52,10 @@ type Props = {
     onSubmit: Function,
     openInviteCollaboratorsSection?: Function,
     recommendedSharingTooltipCalloutName?: ?string,
+    restrictedExternalEmails: Array<string>,
     selectedContacts: Array<Contact>,
     sendButtonProps?: Object,
-    shouldInvalidateExternalCollabs?: boolean,
```
jpfranco, line 53: This is now implied with the existence of `restrictedExternalEmails`, so no longer needed as a separate prop.

src/features/unified-share-modal/EmailForm.js
```diff
@@ -59,28 +64,52 @@ type Props = {
...
     } = React.createRef();
 
+    componentDidUpdate(prevProps: Props, prevState: State) {
+        const { contactsFieldError, justificationReasonsFieldError } = this.state;
+        const {
+            contactsFieldError: prevContactsFieldError,
+            justificationReasonsFieldError: prevJustificationReasonsFieldError,
+        } = prevState;
+
+        // Only display one type of error at a time and give preference
```
jpfranco, line 103: I also tried another approach in which there was a single `formError` which had a type that indicated which field it belonged to. Unfortunately with that approach the general validation that happens on submit would result in cases in which all errors would end up being hidden (depending on the validator order). It seemed to me that working around that resulted in a flow that was harder reason about, so settled for this approach instead.

src/features/unified-share-modal/EmailForm.js
```diff
@@ -107,6 +136,23 @@ class EmailForm extends React.Component<Props, State> {
...
     };
 
+    handleRemoveRestrictedExternalContacts = () => {
+        const { onContactRemove, selectedContacts, updateSelectedContacts } = this.props;
+
+        const [removedContacts, remainingContacts] = partition(selectedContacts, ({ value }) =>
+            this.isRestrictedExternalEmail(value),
+        );
+
+        updateSelectedContacts(remainingContacts);
```
jpfranco, line 146: It wasn't possible to achieve this by calling `handleContactRemove` since that relies on the contact index, which might change each time a contact was removed.

src/features/unified-share-modal/EmailForm.js
```diff
@@ -164,32 +231,41 @@ class EmailForm extends React.Component<Props, State> {
...
+                    restrictedExternalEmails.push(value);
+                }
                 emails.push(value);
             }
         });
+
         onSubmit({
             emails,
             groupIDs,
+            justificationReason: selectedJustificationReason,
```
jpfranco, line 266: Alternatively we could avoid including these new value when they do not apply (currently they will be either `null` or an empty array, respectively), but I think this is a bit cleaner. Note that `UnifiedShareForm` processes only the data that it needs.

src/features/unified-share-modal/ContactRestrictionNotice.js
```diff
@@ -0,0 +1,139 @@
...
+import ContactsEmailsTooltip from './ContactsEmailsTooltip';
+import messages from './messages';
+
+import type { SelectOptionProp } from '../../components/select-field/props';
+import type { contactType as Contact } from './flowTypes';
+
+import './ContactRestrictionNotice.scss';
+
+type Props = {
+    error?: React.Node,
```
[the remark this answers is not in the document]
jpfranco: This is for accepting any node that can be rendered in React: https://flow.org/en/docs/react/types/#toc-react-node

Most of our UI elements use `React.Node` for errors I think because you can also pass a `<FormattedMessage />` instead of just a string.

src/features/unified-share-modal/ContactRestrictionNotice.scss
```diff
@@ -0,0 +1,24 @@
+@import '../../styles/variables';
+
+.bdl-ContactRestrictionNotice {
+    .bdl-SelectButton {
+        margin-top: 10px;
```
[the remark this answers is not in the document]
jpfranco: Oh ok, wasn't aware of that variable, let me update

src/features/unified-share-modal/ContactRestrictionNotice.js
```diff
@@ -0,0 +1,139 @@
...
+        >
+            This classified content requires business justification to collaborate with{' '}
+            <ContactsEmailsTooltip contacts={restrictedExternalContacts}>
+                <Param
+                    value={restrictedExternalContactCount}
+                    description="Number of external collborators currently selected"
+                />{' '}
+                people
+            </ContactsEmailsTooltip>
+            . Select a business justification below or <RemoveButton>remove them</RemoveButton> to continue.
```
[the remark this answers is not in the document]
jpfranco: Yes, so in this case we have the two separate versions of the message, so I think it shouldn't be an issue. The problem I think is that we're not currently using a different copy for "remove them", but I think "remove them" also works when referring to one person?

The `{' '}` are being automatically added by the linter and I think would probably be added when committing the code even if I forced them to be removed. I think translators would be working mostly with the properties files, but let me double-check with moji.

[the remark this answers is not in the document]
jpfranco: But yes, let me confirm the "requires a business justification" part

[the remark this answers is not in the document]
alexkrolick: I'd check with the internationalization team on this. In English, we often have 2-3 variants for plurals (_none_, _one_ or _many_), but in other languages, it could more depending on how the verbs conjugate. That's why we use the intl framework pluralization methods.

jpfranco:
Updated pill styles:

Screen Shot 2020-10-20 at 11 28 24 AM

src/features/unified-share-modal/ContactRestrictionNotice.js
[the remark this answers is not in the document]
jpfranco: Yes, confirmed that this workaround might cause issues in some languages. Currently looking into a fix for `<FormattedCompMessage />` that might get `<Plural />` working properly.

src/features/unified-share-modal/ContactRestrictionNotice.scss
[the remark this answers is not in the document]
jpfranco: Updated most margins to use `bdl-grid-unit`, but the one for the loading indicator needs to be exact in order to prevent a jump when loading completes.

jpfranco:
Selected state for red pill:

Screen Shot 2020-10-20 at 12 32 04 PM

src/components/pill-selector-dropdown/RoundPill.scss
```diff
@@ -67,6 +80,26 @@
     .bdl-Avatar-externalBadge {
         border-color: $bdl-gray-20;
     }
+
+    &.bdl-RoundPill--warning {
+        background-color: $bdl-yellorange-20;
+        border-color: $bdl-yellorange;
+        box-shadow: none;
```
booc0mtaco, line 87: were these inheriting a box shadow from somewhere?

booc0mtaco (approved):
We need to review with design whether bordered pills is a general use case (feature of label pill), or if they are specific to user pills.

For the text, it may be good to stub it out, while we iterate on the proper implementation here.

src/components/pill-selector-dropdown/RoundPill.scss
jpfranco: Yes, the default selected pill style has a white box shadow. On pills with borders it looks like a gap though

PR #2376, opened 2020-10-19
feat(annotations): update error callback for annotation modify

ChenCodes:
Changes in this PR:
- [x] when an update call fails for an annotation comment, we should show an error message
- [x] when an update call fails for a regular comment, we should show an error message

**Annotation Modify:**
Screen Shot 2020-10-19 at 12 23 55 PM

**Comment Modify:** 
Screen Shot 2020-10-20 at 9 12 27 AM


src/api/Feed.js
```diff
@@ -199,11 +199,24 @@ class Feed extends Base {
                 }
             },
             (e: ErrorResponseData, code: string) => {
-                this.feedErrorCallback(true, e, code);
+                this.updateAnnotationErrorCallback(e, code, annotationId);
```
jstoffan, line 202: I don't think this new method is necessary, since it's still wrapped in an anonymous function. Can we just add the `this.updateFeedItem` line above?

ChenCodes: Yep will do

src/api/Feed.js
```diff
@@ -199,6 +199,7 @@ class Feed extends Base {
                 }
             },
             (e: ErrorResponseData, code: string) => {
+                this.updateFeedItem(this.createFeedError(messages.annotationUpdateErrorMessage), annotationId);
```
mxiao6, line 202: Test?

ChenCodes: Added test

src/api/__tests__/Feed.test.js
```diff
@@ -1369,6 +1369,21 @@ describe('api/Feed', () => {
...
+            feed.feedErrorCallback = jest.fn();
+            feed.updateAnnotation = jest.fn().mockImplementation(() => {
+                feed.updateFeedItem();
+                feed.feedErrorCallback(true, e, errorCode);
+            });
+
+            feed.updateAnnotation(file, annotationId, permissions, text, successCallback, errorCallback);
+
+            expect(feed.updateFeedItem).toHaveBeenCalled();
+            expect(feed.feedErrorCallback).toHaveBeenCalledWith(true, e, errorCode);
```
mxiao6, line 1385: This is testing mocked `feed.updateAnnotation` not the real code. Since you're testing `feed.updateAnnotation`, you shouldn't mock it, right? How about try to mock `feed.annotationsAPI.updateAnnotation`?

i18n/en-US.properties
```diff
@@ -16,6 +16,8 @@ be.activitySidebar.activityFeed.taskMissingError = This task no longer exists
 be.add = Add
 # Text to display when app is disabled by applied access policy
 be.additionalTab.blockedByShieldAccessPolicy = Use of this app has been disabled by the applied access policy
+# Error message when an annotation comment update fails
+be.api.annotationUpdateErrorMessage = This comment could not be modified.
```
ConradJChan, line 20: I may be confused, but do we update `messages.js` and let the `build:i18n` update the en-US.properties file?

src/api/messages.js
```diff
@@ -12,6 +12,11 @@ const messages = defineMessages({
         description: 'Error message when an app activity deletion fails',
         defaultMessage: 'There was an error deleting this item.',
     },
+    annotationUpdateErrorMessage: {
```
ConradJChan, line 15: It doesn't look like comments actually has this UX -- two comments: (1) should this message be `commentUpdateErrorMessage` since we're treating these annotations as comments and (2) should we update the existing update comment flow to behave the same as our annotation flow like we do for `deleteAnnotation`?

ChenCodes: Yes to both comments

src/api/__tests__/Feed.test.js
ChenCodes: I don't think I need this test anymore because I've moved out the callback to a helper method. I think since I am using the helper in two locations, it makes sense to have a helper rather than hardcode the logic in both locations.

src/api/__tests__/Feed.test.js
```diff
@@ -1369,20 +1369,22 @@ describe('api/Feed', () => {
...
-            });
-
-            feed.updateAnnotation(file, annotationId, permissions, text, successCallback, errorCallback);
+        });
 
-            expect(feed.updateFeedItem).toHaveBeenCalled();
-            expect(feed.feedErrorCallback).toHaveBeenCalledWith(true, e, errorCode);
+        test('should update the feed item and call the error callback', () => {
+            const id = '1';
+            feed.updateCommentErrorCallback(e, errorCode, id);
```
ConradJChan, line 1385: Should we assert that `createFeedError` was called with the appropriate message as well?

src/api/messages.js
```diff
@@ -12,8 +12,8 @@ const messages = defineMessages({
         description: 'Error message when an app activity deletion fails',
         defaultMessage: 'There was an error deleting this item.',
     },
-    annotationUpdateErrorMessage: {
-        id: 'be.api.annotationUpdateErrorMessage',
+    commentUpdateErrorMessage: {
```
ConradJChan, line 15: Can this be moved to be 🔤 ?

mxiao6 (approved):

PR #2379, opened 2020-10-20
fix(annotations): use correct value for isOpen prop

ChenCodes:
Changes in this PR:
- [x] we are now using isEditing value to determine whether or not we should be showing the Cancel / Post buttons

The issue is that before when we clicked on the Modify button, the form input would show up but the Cancel / Post buttons would not show up until the user clicks to focus the input field.

The correct UX is that when the user clicks on the Modify button, we should immediately see the Cancel / Post buttons.
Screen Shot 2020-10-20 at 9 23 05 AM

### DIFF
--- a/src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
+++ b/src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
@@ -56,6 +56,7 @@ const AnnotationActivity = ({
 
     const handleEdit = () => {
         setIsEditing(true);
+        setIsInputOpen(true);
     };
 
     const handleOnSelect = () => {
@@ -118,14 +119,14 @@ const AnnotationActivity = ({
                     <div>
                         <ActivityTimestamp date={createdAtTimestamp} />
                     </div>
-                    {isEditing && currentUser ? (
+                    {isEditing && isInputOpen && currentUser ? (
                         <CommentForm
                             className="bcs-AnnotationActivity-editor"
                             entityId={id}
                             getAvatarUrl={getAvatarUrl}
                             getMentionWithQuery={getMentionWithQuery}
                             isEditing={isEditing}
-                            isOpen={isInputOpen}
+                            isOpen={isEditing}
                             mentionSelectorContacts={mentionSelectorContacts}
                             onCancel={handleFormCancel}
                             onFocus={handleFormFocus}
--- a/src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
+++ b/src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
@@ -119,14 +119,14 @@ const AnnotationActivity = ({
                     <div>
                         <ActivityTimestamp date={createdAtTimestamp} />
                     </div>
-                    {isEditing && isInputOpen && currentUser ? (
+                    {isEditing && currentUser ? (
                         <CommentForm
                             className="bcs-AnnotationActivity-editor"
                             entityId={id}
                             getAvatarUrl={getAvatarUrl}
                             getMentionWithQuery={getMentionWithQuery}
                             isEditing={isEditing}
-                            isOpen={isEditing}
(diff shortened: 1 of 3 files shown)


src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
```diff
@@ -118,14 +119,14 @@ const AnnotationActivity = ({
                     <div>
                         <ActivityTimestamp date={createdAtTimestamp} />
                     </div>
-                    {isEditing && currentUser ? (
+                    {isEditing && isInputOpen && currentUser ? (
```
ChenCodes, line 122: If `isInputOpen` is false, then we shouldn't be showing the comment form.

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
```diff
@@ -56,6 +56,7 @@ const AnnotationActivity = ({
 
     const handleEdit = () => {
         setIsEditing(true);
+        setIsInputOpen(true);
```
ChenCodes, line 59: Both `isEditing` and `isInputOpen` will be set to true if the user clicks on Modify button.

booc0mtaco:
What are valid values for `isEditing` and `isInputOpen`? Can the UI get into a state where the former is true, and the latter is false, and vice versa?

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
ConradJChan: Do we care to maintain the `isInputOpen` state? Would basing the logic off of `isEditing` alone be sufficient? Doesn't seem like the `onFocus` behavior is needed

ChenCodes:
> What are valid values for `isEditing` and `isInputOpen`? Can the UI get into a state where the former is true, and the latter is false, and vice versa?

I realized that I didn't need to do both the isEditing / isInputOpen check. Just the `isEditing` check is enough because technically the form cannot be focused if the user hasn't clicked on the Modify button. If the user clicks on the Modify button, then both `isInputOpen` and `isEditing` will be true which means we should be showing the form.

The main fix in this change is that I was missing setting `isInputOpen` to be true in the edit click handler.

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
ChenCodes: Yep I've addressed this in the latest commit - we don't need isInputOpen logic here.

ConradJChan:
Can we get rid of `isInputOpen` state entirely? Do we have unit test coverage for the `CommentForm` being shown?

ChenCodes:
> Can we get rid of `isInputOpen` state entirely? Do we have unit test coverage for the `CommentForm` being shown?

I don't think we can get rid of isInputOpen entirely because CommentForm has `onFocus` as a required property.

I'll be adding a unit test for CommentForm being shown

src/elements/content-sidebar/activity-feed/annotations/__tests__/AnnotationActivity.test.js
```diff
@@ -95,6 +95,23 @@ describe('elements/content-sidebar/ActivityFeed/annotations/AnnotationActivity',
...
+                permissions: { can_edit: true },
+            },
+        };
+
+        const wrapper = mount(<AnnotationActivity {...mockActivity} {...activity} />);
+
+        wrapper.find('AnnotationActivityMenu').simulate('click');
+        wrapper.find('MenuItem').simulate('click');
+
+        expect(wrapper.exists('CommentForm')).toBe(true);
```
ConradJChan, line 112: Could we do a cancel button click and verify that the `CommentForm` doesn't exist as well?

ConradJChan (approved):

ConradJChan (approved):

PR #2389, opened 2020-10-22
fix(annotations): use correct pencil icon

ChenCodes:
Changes in this PR:
- [x] use Pencil16 component 

Screen Shot 2020-10-22 at 11 43 57 AM

### DIFF
--- a/src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityMenu.js
+++ b/src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityMenu.js
@@ -3,10 +3,10 @@ import * as React from 'react';
 import TetherComponent from 'react-tether';
 import { FormattedMessage } from 'react-intl';
 import DeleteConfirmation from '../common/delete-confirmation';
-import IconTrash from '../../../../icons/general/IconTrash';
 import Media from '../../../../components/media';
 import messages from './messages';
 import Pencil16 from '../../../../icon/line/Pencil16';
+import Trash16 from '../../../../icon/fill/Trash16';
 import { ACTIVITY_TARGETS } from '../../../common/interactionTargets';
 import { bdlGray } from '../../../../styles/variables';
 import { MenuItem } from '../../../../components/menu';
@@ -72,7 +72,7 @@ const AnnotationActivityMenu = ({ canDelete, canEdit, id, onDeleteConfirm, onEdi
                         data-testid="delete-annotation-activity"
                         onClick={handleDeleteClick}
                     >
-                        <IconTrash color={bdlGray} />
+                        <Trash16 color={bdlGray} />
                         <FormattedMessage {...messages.annotationActivityDeleteMenuItem} />
                     </MenuItem>
                 )}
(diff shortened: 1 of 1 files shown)


src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityMenu.js
```diff
@@ -3,10 +3,10 @@ import * as React from 'react';
 import TetherComponent from 'react-tether';
 import { FormattedMessage } from 'react-intl';
 import DeleteConfirmation from '../common/delete-confirmation';
-import IconEdit from '../../../../icons/general/IconEdit';
 import IconTrash from '../../../../icons/general/IconTrash';
 import Media from '../../../../components/media';
 import messages from './messages';
+import Pencil16 from '../../../../icon/line/Pencil16';
```
mxiao6, line 9: Can you rename it to `IconEdit`?

ChenCodes: Does it need to be `IconEdit`? In the existing implementation of `Comment.js I'm seeing that it is defined as `Pencil16`.

mxiao6: If you search `Icon*` in BUIE codebase, you'll realize it's like a naming convention. And I think it's a good practice to follow naming conventions.

booc0mtaco: Being a bit nosy :) The legacy icons (in `icons/general`) are being deprecated in favor of ones coming from `icon/line`, and `icon/fill` which come from our design asset import job. we should try to move off of those general icons, and use the equivalents in the aforementioned directories instead.

The new naming convention is icon/{type}/{ImageName}{baseResolution}.tsx

Lastly, we could move the trash icon from the `/general` version to the line-equivalent here, to keep consistency. Any styling should be applied via SASS or styled components (prefer sass since most colors are unchanging at runtime).

mxiao6 (dismissed):

mxiao6 (approved):

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityMenu.js
```diff
@@ -72,7 +72,7 @@ const AnnotationActivityMenu = ({ canDelete, canEdit, id, onDeleteConfirm, onEdi
                         data-testid="delete-annotation-activity"
                         onClick={handleDeleteClick}
                     >
-                        <IconTrash color={bdlGray} />
+                        <Trash16 color={bdlGray} />
```
booc0mtaco, line 75: @ChenCodes you will notice the new icons do not support `color` in this way, but its likely the color is correct because of the defaults set by design. this should be cleaned up. If you need a custom color here (or to handle `:hover`) use SASS

ChenCodes: Thanks for the heads up, I'll put a fix in for this.

PR #2397, opened 2020-10-26
test(cypress): Add E2E tests for Annotations in content-preview

mxiao6:


mxiao6:
Considering using a third-party CDN for annotations assets instead of committing the build.

test/integration/content-preview/Annotations.e2e.test.js
```diff
@@ -0,0 +1,89 @@
...
+            },
+        });
+
+        cy.get('.ba-RegionAnnotation').should('be.visible'); // Region target is on the first page
+        cy.get('.ba-HighlightTarget-rect').should('not.be.visible'); // Highlight target is on the second page
+
+        cy.contains('Highlight Annotation on version 2')
+            .siblings()
+            .filter('button')
+            .contains('Page 2')
```
ConradJChan, line 49: Could this reference the message keys? Similar to https://github.com/box/box-ui-elements/blob/9ae2349beaa886d8387829e4a0d596d5674221c5/test/integration/content-sidebar/TaskForm.e2e.test.js#L17

test/integration/content-preview/Annotations.e2e.test.js
```diff
@@ -0,0 +1,89 @@
...
+            .click();
+
+        cy.get('.ba-RegionAnnotation').should('not.be.visible');
+        cy.get('.ba-HighlightTarget-rect').should('be.visible');
+
+        cy.contains('Region Annotation on version 2')
+            .siblings()
+            .filter('button')
+            .contains('Page 1')
+            .click();
```
ConradJChan, line 59: Could these be turned into a helper function? Something like `clickAnnotationLinkByComment` perhaps? Is validating the link's text part of the test or could we do `filter` by a `data-testid`?

ConradJChan (commented):
Should we assert for selection classes being applied to the markup as well as the sidebar activity items?

test/integration/content-preview/Annotations.e2e.test.js
```diff
@@ -0,0 +1,89 @@
+import('https://unpkg.com/box-annotations/dist/annotations.js');
```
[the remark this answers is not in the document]
mxiao6: Yeah, e.g. "https://unpkg.com/box-annotations@4.0.0-beta.23/dist/annotations.js". However, I don't think we need to support this until we do hash verification in the future, since we always want to use the latest version.

mxiao6:
> Should we assert for selection classes being applied to the markup as well as the sidebar activity items?

The selection has some problems in the E2E tests. I plan to address it in a separate PR. I don't want to include too much content in this first Annotations E2E setup commit.

test/integration/content-preview/Annotations.e2e.test.js
```diff
@@ -0,0 +1,88 @@
+// Specify Annotations Version in command line, for example:
```
ConradJChan, line 1: Maybe note that otherwise defaults to the latest version

test/support/commands.js
```diff
@@ -2,3 +2,11 @@ Cypress.Commands.add('getByTestId', testId => cy.get(`[data-testid="${testId}"]`
 Cypress.Commands.add('getByTitle', title => cy.get(`[title="${title}"]`));
 Cypress.Commands.add('getTooltip', () => cy.get('[role="tooltip"]'));
 Cypress.Commands.add('getByAriaLabel', ariaLabel => cy.get(`[aria-label="${ariaLabel}"]`));
+
+// Annotations-specific commands
+Cypress.Commands.add('clickAnnotationLinkByComment', comment => {
```
ConradJChan, line 7: These commands seem to be very general, should this be organized in a separate file like `annotations.js` in this directory and imported? Or maybe just define the helper function locally in `Annotations.e2e.test.js`

ConradJChan (approved):

test/integration/content-preview/Annotations.e2e.test.js
```diff
@@ -0,0 +1,94 @@
...
+                    const link = document.createElement('link');
+                    link.rel = 'stylesheet';
+                    link.type = 'text/css';
+                    link.href = `https://unpkg.com/box-annotations${ANNOTATIONS_VERSION}/dist/annotations.css`;
+                    head.appendChild(link);
+                },
+            });
+        },
+        clickAnnotationLinkByComment(comment) {
+            cy.contains(comment)
```
ConradJChan, line 38: Maybe we could `return` the subject here as well so we can possibly chain any selected state expectations in the future?

ConradJChan (dismissed):

ConradJChan (approved):

PR #2401, opened 2020-10-29
feat(usm): Integrate collab restriction justification flow

jpfranco:
This is a partial update that integrates the Justification components that were previously added to `EmailForm`, with the rest of the USM modal flow. The changes are backwards compatible and the new flow will not be triggered initially. 

**Note**: Some UI polish and fixes are still pending, such as updating the messages on `ContactRestrictionNotice` so that localization issues can be avoided.

Screen Shot 2020-10-29 at 10 11 29 AM


src/features/unified-share-modal/UnifiedShareForm.js
```diff
@@ -54,15 +69,70 @@ class UnifiedShareForm extends React.Component<USFProps, State> {
...
+        const didExternalCollabRestrictionsChange =
+            !isEqual(restrictedExternalCollabEmails, prevRestrictedExternalCollabEmails) ||
+            isCollabRestrictionJustificationAllowed !== prevIsCollabRestrictionJustificationAllowed;
+
+        if (didExternalCollabRestrictionsChange && this.shouldRequireExternalCollabJustification()) {
+            this.fetchJustificationReasons(item, JUSTIFICATION_CHECKPOINT_EXTERNAL_COLLAB);
+        }
+    }
+
+    fetchJustificationReasons = (item: Item, checkpoint: JustificationCheckpoint) => {
```
jpfranco, line 100: Using a pattern similar to what we have for `getContactsByEmail`, where the results are only relevant to the component and thus don't need to be saved in a store and passed down as props. Currently there is no need to cache or share justification reasons beyond this component, so I'm avoiding having them on a store on the webapp. Let me know in case there's a preference for that approach though.

src/features/unified-share-modal/UnifiedShareForm.js
```diff
@@ -54,15 +69,70 @@ class UnifiedShareForm extends React.Component<USFProps, State> {
...
+        const { getJustificationReasons } = this.props;
+        const hasJustificationReasons = !!justificationReasons.length;
+
+        if (!getJustificationReasons || hasJustificationReasons) {
+            return Promise.resolve();
+        }
+        this.setState({ isFetchingJustificationReasons: true });
+
+        return getJustificationReasons(item.typedID, checkpoint)
+            .then(({ classificationLabelId, options }: GetJustificationReasonsResponse) => {
```
jpfranco, line 111: Using promises instead of `async`/`await` for consistency with the rest of the file.

src/features/unified-share-modal/UnifiedShareForm.js
```diff
@@ -125,6 +196,22 @@ class UnifiedShareForm extends React.Component<USFProps, State> {
...
 
+        const hasJustificationReason = !!justificationReason;
+        const hasRestrictedExternalInvitees = !isEmpty(restrictedExternalEmails);
+        const shouldSubmitJustificationReason =
+            hasJustificationReason && hasRestrictedExternalInvitees && isCollabRestrictionJustificationAllowed;
+
+        if (shouldSubmitJustificationReason) {
+            params = {
+                ...params,
+                classificationLabelId,
```
jpfranco, line 207: Noting that these will not be passed directly to the send invites api since the justification will be submitted as a separate request. I've been considering whether or not it'd be better to pass this as a second params object, but thought it might be cleaner this way since `ContentSharing` is already converting the params into the format it needs, for example.

Also note that `ContentSharing` will not reach this code path since it's not wired to trigger the justification flow. This is expected since this flow will not be initially supported for platform.

sahiga (approved):

PR #2403, opened 2020-11-06
feat(usm): Adding new external collab restriction flow

jpfranco:
Partial update that adds and integrates the UI for the new external collab restriction flow. Previously it was possible to submit non-restricted invitees along with invitees restricted by an Access Policy (unrestricted invites would succeed, but a red error banner would be shown for restricted ones). Once the new flow is enabled, it will no longer be possible to submit the form with restricted invitees, and the user will be asked to remove them.

Changes should be backwards compatible and the new flow will not yet be enabled.

Screen Shot 2020-11-05 at 7 09 06 PM

Additionally, this PR also switches the messages from `FormattedCompMessage` to `FormattedMessage`, given that currently there are some unresolved issues with `FormattedCompMessage` that prevent us from using it. Note that the design had to be altered a bit in order to have the interactive elements be their own sentences, since `react-intl` v3+ is not yet available in EUA.

Screen Shot 2020-11-06 at 10 58 14 AM

### DIFF
--- a/src/features/unified-share-modal/__tests__/ContactRestrictionNotice.test.js
+++ b/src/features/unified-share-modal/__tests__/ContactRestrictionNotice.test.js
@@ -2,6 +2,8 @@
 import React from 'react';
 import { shallow } from 'enzyme';
 
+import messages from '../messages';
+
 import { ContactRestrictionNoticeComponent as ContactRestrictionNotice } from '../ContactRestrictionNotice';
 
 describe('features/unified-share-modal/ContactRestrictionNotice', () => {
@@ -9,15 +11,13 @@ describe('features/unified-share-modal/ContactRestrictionNotice', () => {
     let selectedContacts;
     let restrictedExternalEmails;
 
-    const singularMessageId = 'boxui.unifiedShare.businessJustificationRequiredSingular';
-    const pluralMessageId = 'boxui.unifiedShare.businessJustificationRequiredPlural';
-
     const getWrapper = (props = {}) => {
         return shallow(
             <ContactRestrictionNotice
                 error=""
-                isLoading={false}
                 intl={{ formatMessage: jest.fn() }}
+                isFetchingJustificationReasons={false}
+                isRestrictionJustificationEnabled={false}
                 justificationReasons={[]}
                 onRemoveRestrictedExternalContacts={jest.fn()}
                 onSelectJustificationReason={jest.fn()}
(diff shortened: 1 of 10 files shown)


src/features/unified-share-modal/__tests__/ContactRestrictionNotice.test.js
```diff
@@ -60,9 +60,9 @@ describe('features/unified-share-modal/ContactRestrictionNotice', () => {
 
     describe('render', () => {
         test('should render default ContactRestrictionNotice component', () => {
-            expect(wrapper.is('InlineNotice')).toBe(true);
-            expect(wrapper.find(`FormattedCompMessage[id="${pluralMessageId}"]`)).toHaveLength(1);
-            expect(wrapper.find('[data-resin-target="justificationReasonsSelect"]')).toHaveLength(1);
+            expect(wrapper.is('Tooltip')).toBe(true);
+            expect(wrapper.find('InlineNotice')).toHaveLength(1);
+            expect(wrapper.find('[data-resin-target="removeBtn"]')).toHaveLength(1);
```
sahiga, line 65: Minor: you could also use `wrapper.exists()` here

src/features/unified-share-modal/__tests__/EmailForm.test.js
```diff
@@ -302,6 +306,44 @@ describe('features/unified-share-modal/EmailForm', () => {
...
+            });
+
+            wrapper.setState({ message });
+            wrapper.instance().handleSubmit(event);
+
+            expect(wrapper.state('contactsRestrictionError')).toEqual('boxui.unifiedShare.justificationRequiredError');
+            expect(onSubmit).toHaveBeenCalledTimes(0);
+        });
+
+        test('should trigger an error and abort submit action when restricted contacts are present and justification is not allowed', () => {
```
sahiga, line 328: Could this test and the previous test be combined into a `test.each` loop?

sahiga (dismissed):

src/features/unified-share-modal/__tests__/EmailForm.test.js
jpfranco: Yes, looks like most of the test's code is the same. Let me update

src/features/unified-share-modal/__tests__/ContactRestrictionNotice.test.js
jpfranco: Probably not an issue for this particular component, but I usually prefer `toHaveLength` in order to cover cases where there could be more than one component of the same kind and the wrong one could be matched by accident.

sahiga (approved):

PR #2405, opened 2020-11-10
feat(content-sharing): Display error for existing collabs

sahiga:
Show a specific error message when the user attempts to invite an existing collaborator to a shared link in Content Sharing.

content-sharing-existing-collaborator

### DIFF
--- a/src/elements/content-sharing/messages.js
+++ b/src/elements/content-sharing/messages.js
@@ -53,6 +53,12 @@ const messages = defineMessages({
             'Message that appears when collaborators cannot be added to the shared link in the ContentSharing Element.',
         id: 'be.contentSharing.sendInvitesError',
     },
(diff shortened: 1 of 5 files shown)


swfree (approved):

src/elements/content-sharing/messages.js
```diff
@@ -53,6 +53,12 @@ const messages = defineMessages({
             'Message that appears when collaborators cannot be added to the shared link in the ContentSharing Element.',
         id: 'be.contentSharing.sendInvitesError',
     },
+    sendInvitesExistingCollabError: {
+        defaultMessage: 'User is already a collaborator.',
```
ctuckz, line 57: Just wondering, does the API response let us know which user is already a collaborator?

PR #2407, opened 2020-11-10
feat(pill-selector-dropdown): Adding RoundPill variants

jpfranco:
**BREAKING CHANGE:** `isWarning` prop removed from `RoundPill` in favor of `variant`

This update introduces the `variant` prop for `RoundPill`, which will be used to set mutually exclusive display states (currently only `warning` and `waived`). Similar to how the `validator` function in `PillSelector` currently works, a new `getPillVariant` function will be called on each pill render in order to determine which variant to use for each particular pill. 

The switch to a callback instead of a boolean like `isWarning` is because it's assumed that these pill variants will depend on the parent component's state, and will thus change depending on the values selected. Without a callback, the parent component would need to iterate through all the pills when there's a state change in order to update the variant for each one.

A couple of things to note:

- Invalid state takes precedence over variant. Ideally "error" would've been a variant itself, but the pill's `isValid` prop currently depends on `validator`, which is also used to validate user input when adding new pills.
-  Other pill states which aren't mutually exclusive are still separate boolean props (e.g. `isExternal`, `isSelected`, etc.)

**Updated storybook (note the new "waived" pill style):** 
selected state

**End Goal: Display "waived" pills that were accepted after being invalid**

Screen Shot 2020-11-10 at 10 14 53 AM

### DIFF
--- a/src/components/pill-selector-dropdown/PillSelectorDropdown.stories.js
+++ b/src/components/pill-selector-dropdown/PillSelectorDropdown.stories.js
@@ -5,6 +5,7 @@ import { State, Store } from '@sambego/storybook-state';
 import ContactDatalistItem from '../contact-datalist-item';
 import PillSelectorDropdown from './PillSelectorDropdown';
 import notes from './PillSelectorDropdown.notes.md';
+import { PILL_VARIANT_DEFAULT, PILL_VARIANT_WARNING, PILL_VARIANT_WAIVED } from './constants';
 
 const users = [
     { id: 0, name: 'bob@foo.bar' },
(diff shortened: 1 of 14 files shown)


src/components/pill-selector-dropdown/RoundPill.js
```diff
@@ -48,31 +51,19 @@ class RoundPill extends React.PureComponent<Props, State> {
...
             'bdl-RoundPill--selected': isSelected && !isDisabled,
             'bdl-RoundPill--disabled': isDisabled,
-            'bdl-RoundPill--warning': hasWarning,
+            'bdl-RoundPill--warning': activeVariant === PILL_VARIANT_WARNING,
+            'bdl-RoundPill--waived': activeVariant === PILL_VARIANT_WAIVED,
             'bdl-RoundPill--error': !isValid,
         });
     };
 
-    getPillType = (): ?string => {
```
jpfranco, line 61: `RoundPill` and `LabelPill` styles have diverged a little, so it's no longer necessary to set the specific type on the label pill (i.e., `RoundPill` styles are overrides of the default `LabelPill` style.

src/features/unified-share-modal/EmailForm.js
```diff
@@ -323,6 +327,17 @@ class EmailForm extends React.Component<Props, State> {
         return isValid;
     };
 
+    getContactPillVariant = (contactPill: SelectOptionProp): RoundPillVariant => {
```
jpfranco, line 330: Happy to move this to a separate PR if that's preferred. This is the end goal of the change, but it's not directly related to the Pill Selector updates.

danubevictoria: i'd agree with moving this out to a separate pr.

jpfranco: Sounds good, removed the USM updates

src/components/pill-selector-dropdown/PillSelectorDropdown.stories.js
```diff
@@ -202,6 +203,71 @@ export const showRoundedPills = () => {
...
+                    allowCustomPills
+                    error={state.error}
+                    getPillVariant={getPillVariant}
+                    placeholder="Names or email addresses"
+                    onInput={handleInput}
+                    onRemove={handleRemove}
+                    onSelect={handleSelect}
+                    selectedOptions={state.selectedOptions}
+                    selectorOptions={state.selectorOptions}
+                    showRoundedPills
```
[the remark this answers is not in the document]
jpfranco: I was looking into breaking this into two stories but noticed that both of the current ones show the avatar. I think it's because `showAvatars` is not passed to the pill. Do you know if this is expected? 

https://github.com/box/box-ui-elements/blob/master/src/components/pill-selector-dropdown/PillSelector.js#L225

I can probably fix this if this was not intentional.

src/components/pill-selector-dropdown/RoundPill.js
jpfranco: I think they shouldn't conflict because `RoundPill` controls the rendering of `LabelPill` so it should always be set to default now.

src/components/pill-selector-dropdown/PillSelectorDropdown.stories.js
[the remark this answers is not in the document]
danubevictoria: i dont think it was intentional but it might be better to address it outside this ticket since it would impact a lot of modals that are using the RoundPill. feel free to leave it for now and file a bug thanks!

jpfranco:
Closing in favor of #2411

PR #2411, opened 2020-11-13
feat(pill-selector-dropdown): Adding function to get custom pill class

jpfranco:
This will add a `getPillClassName` prop to `PillSelectorDropdown`, which will allow setting a custom class for each pill on render. The base `RoundPill` styles were moved to a mixin in order to allow creating custom classes that match the overall appearance of `RoundPill`.

I had initially tried a different approach in which you would set a predefined pill variant, but it seemed too use case-specific to be worth adding. I believe this approach should be more future proof since it's super generic.

Note: Existing pill styles should remain the same after this update.

**Example:**
Setting a red border on a pill that was invalid but has been waived due to a separate action:
Screen Shot 2020-11-13 at 9 38 05 AM


src/components/pill-selector-dropdown/RoundPill.scss
```diff
@@ -59,47 +37,11 @@
 }
 
 .bdl-RoundPill--warning {
-    background-color: $bdl-yellorange-10;
-    border: 1px solid $bdl-yellorange-50;
-
-    .bdl-Avatar-externalBadge {
-        border-color: $bdl-yellorange-10;
-    }
+    @include bdl-RoundPill($background-color: $bdl-yellorange-10, $border-color: $bdl-yellorange-50, $selected-background-color: $bdl-yellorange-20, $selected-border-color: $bdl-yellorange);
```
jpfranco, line 40: I had each parameter on its own line, but it looks like the code gets automatically reformatted either when committing or when pushing the code.

jpfranco:
@booc0mtaco I added the storybook example shown below.

I don't think we have any other cases in which we use scss within a story (or at least I couldn't find any), so let me know what you think about this approach.

Screen Shot 2020-11-13 at 2 33 20 PM

src/components/pill-selector-dropdown/PillSelectorDropdown.stories.js
```diff
@@ -254,6 +256,71 @@ export const showAvatars = () => {
     );
 };
 
+export const customPillStyles = () => {
```
sahiga, line 259: Should there be a visual regression test for this story?

src/components/pill-selector-dropdown/_mixins.scss
```diff
@@ -0,0 +1,33 @@
+@import '../../styles/variables';
+
+@mixin bdl-RoundPill($background-color: $bdl-gray-10, $border-color: $bdl-gray-10, $selected-background-color: $bdl-gray-20, $selected-border-color: $white) {
```
sahiga, line 3: Nice! 👍

sahiga (approved):

src/components/pill-selector-dropdown/PillSelectorDropdown.stories.js
jpfranco: Yes, should be useful. I have a quick follow up PR so I can include it there, since this has been open for a few days now.

PR #2426, opened 2020-12-08
fix(usm): Fix line overlap on ContactRestrictionNotice

jpfranco:
Text was overlapping in cases in which the remove button fit within the last line of the paragraph and justifications were disabled.

**Before:**

Screen Shot 2020-12-07 at 1 16 24 PM

**After:**

Screen Shot 2020-12-07 at 1 18 09 PM

### DIFF
--- a/src/features/unified-share-modal/ContactRestrictionNotice.js
+++ b/src/features/unified-share-modal/ContactRestrictionNotice.js
@@ -103,6 +103,7 @@ const ContactRestrictionNotice = ({
                     {...restrictionNoticeMessage}
                     values={{ count: restrictedExternalContactCount, email: firstEmail }}
                 />
+                &nbsp;
                 {isRestrictionJustificationEnabled && justificationSelectSection}
                 <PlainButton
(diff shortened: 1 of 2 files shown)


src/features/unified-share-modal/ContactRestrictionNotice.js
```diff
@@ -103,6 +103,7 @@ const ContactRestrictionNotice = ({
                     {...restrictionNoticeMessage}
                     values={{ count: restrictedExternalContactCount, email: firstEmail }}
                 />
+                &nbsp;
```
feixuanli, line 106: Is it possible to replace `&nbsp;` with styles change ?

feixuanli (dismissed):

src/features/unified-share-modal/ContactRestrictionNotice.js
jpfranco: Yes, but I think it's preferable to use the `&nbsp;` in order to get the correct space width based on the current font size, etc.

feixuanli (approved):

Issue #2322, opened 2020-09-03
Uploading a large file from an old device halts the upload process.


toshimitsu-m-kamei-box:
Please fill out the following template so we can reproduce and fix your issue as quickly as possible!

**Environment:**

- Elements version:
  - Content Explorer 11.0.2
  - Content Uploader 11.0.2

**Desktop (please complete the following information):**

N/A

**Smartphone (please complete the following information):**

- iPhone 6(iOS 12.4.8)
- iPhone 7(iOS 13.3.1)

**Steps to reproduce the problem:**

- Construct a platform for Content Explorer and Content Uploader(eg: Like a Box API Navigator)
- Upload a large file

**What is the expected behavior? (Screenshots can be helpful here)**

On older devices, such as an iPhone 6 or 7, when using the Box UI Elements Content Uploader to upload a file of 500MB or so, the upload stops in the middle.
Also, if you upload with Chunked Upload turned off, a message will appear stating that there is a problem with the page and that it was reloaded, and then the page will stop.
I can't upload files properly.

As far as we can see, iPhone X, iPhone 8 and iPhone 11 Pro upload files without any problems.

I'm assuming that the crash is in the process of storing the file in memory.

Example movie is here.

**What went wrong? (Screenshots, console logs, or HAR files can be helpful here)**

- Incorrect case(iPhone 6 iOS 12.4.8):
  - https://cloud.box.com/s/lx9hxojzb3p8my62jtpijb7k6b3a1vkd
- Correct case (iPhone X iOS )
  - https://cloud.box.com/s/9iud9hka6psgbijg7asrx1eibvtzzbfs

**Link to application or sample code:**

N/A

**If relevant, link to file (or attach file here)**

N/A

**Expected behavior**

N/A

**Screenshots**

**Additional context**

N/A

PR #2441, opened 2020-12-21
fix(upload): Remove client side file change checks

prakash-kumar:
fixes https://github.com/box/box-ui-elements/issues/2322

According to https://w3c.github.io/FileAPI/#file-section and https://stackoverflow.com/a/14284215/12031375, the underlying File object and its props do not change when the underlying file changes on disk. 

> User agents should endeavor to have a File object’s snapshot state set to the state of the underlying storage on disk at the time the reference is taken. If the file is modified on disk following the time a reference has been taken, the File's snapshot state will differ from the state of the underlying storage.


booc0mtaco (approved):
awaiting being added to the team in the review list for this review to count :)

booc0mtaco (approved):

PR #2464, opened 2021-01-27
fix(activity): annotation activity width truncation after deletion

ChenCodes:
**Main Issue:**
There was an intermittent bug found in IE11 where when the user deletes a comment in the activity sidebar, sometimes the comments in the sidebar would get truncated.

**Solution:**
The fix is to move the `width: 100%` style on the `.bcs-activity-feed-active-state` element to the parent element.

**Demo:**
ezgif com-gif-maker (5)

- [x] cross browser testing

### DIFF
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
@@ -37,6 +37,7 @@
         display: flex;
         flex: 1 1 auto;
         flex-direction: column;
+        width: 100%;
         overflow-x: hidden;
         overflow-y: auto;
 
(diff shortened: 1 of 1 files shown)


src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
```diff
@@ -37,6 +37,7 @@
         display: flex;
         flex: 1 1 auto;
         flex-direction: column;
+        width: 100%;
```
ConradJChan, line 40: Maybe just add a comment for posterity -- something about how since it's flex-direction column, we want to explicitly set the width

ConradJChan (approved):

mxiao6 (dismissed):

ConradJChan (dismissed):

sahiga (approved):

PR #2480, opened 2021-02-23
 feat(storybook): create stories for Logo

jasminevle:


src/components/logo/Logo.stories.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react';
+
+import Logo from './Logo';
+import notes from './Logo.stories.md';
+
+export const regular = () => <Logo />;
```
sahiga, line 6: The Styleguidist example sets the `title` attribute to "Box" -- can you include that here as well?

sahiga (commented):
Can you add a visual regression test for the new story?

src/components/logo/__tests__/Logo.stories.test.tsx
```diff
@@ -0,0 +1,6 @@
+describe('components/logo/Logo', () => {
+    test.each([['logo']])('looks visually correct when using story %s', async id => {
```
sahiga, line 2: Can you include the image snapshot in this PR? You can get an image just for this test by launching the Storybook page and then running `yarn test:images src/components/logo/__tests__/Logo.stories.test.tsx`.

Also, since there is only one `Logo` story, do we need the `test.each()` call?

jasminevle: Oops! Thanks for reminding me about this very crucial step -- first time migrating if you couldn't tell 😅 

I saw that we did test.each() in components with only one story such as with `Header` and `ImageTooltip`, but we also just do test() in other components like `ErrorMask`. I've ended up changing it to just using test() though!

Thank you for your careful combing and comments @sahiga ! Appreciate it!

sahiga: You're welcome! Thanks so much for helping out with the migration! And good call on `Header` and `ImageTooltip`. I'll convert those files to use `test()` in a future PR.

Looks like the snapshot is missing a preview. Could you update the story id and run the test again? Story ids typically have the format `group-featureName--exampleName`, so this would be `components-logo--regular` (also try navigating to http://localhost:6061/?path=/story/components-logo--regular to confirm this).

sahiga (approved):

PR #2484, opened 2021-03-01
refactor(typescript): Migrate Section

marzelinger:


src/components/section/__tests__/Section.stories.test.tsx
```diff
@@ -0,0 +1,6 @@
+describe('components/section/Section', () => {
+    test.each(['components-section--basic'])('looks visually correct when using story %s', async id => {
```
sahiga, line 2: Since there is only one test case, could we replace the `test.each` with a regular `test` call?

sahiga (approved):

PR #2485, opened 2021-03-01
feat(left-sidebar): add isShown prop to LeftSidebarLinkCallout component

JohnnyB83:


src/features/left-sidebar/__tests__/LeftSidebarLinkCallout.test.js
```diff
@@ -44,5 +44,13 @@ describe('components/tooltip/LeftSidebarLinkCallout', () => {
             const btn = wrapper.find('.nav-link-callout-close-button');
             btn.simulate('click');
         });
+
+        test('should enable tethered component when isShown prop true', () => {
+            const callout = {
+                content: <div>Hi</div>,
+            };
+            const wrapper = getWrapper({ isShown: true, callout });
+            expect(wrapper.props().enabled).toBe(true);
```
wlyhdg, line 53: what is this necessarily doing? are you validating the component was rendered by saying the props were shown?

JohnnyB83: previously this callout showing was controlled all by internal state and since we are trying to use it more like a targeted component (like Tooltip) I wanted to have the callout be able to be controlled by a prop that is hooked up to a targetingAPI, so just trying to test to see if passing the prop in will cause the callout to show

wlyhdg (approved):

PR #2486, opened 2021-03-02
feat(date-picker): Allow permanent visibility and custom inputs

sahiga:
date-picker-always-visible-2

This PR adds more flexibility to the `DatePicker` component by allowing a custom input field and permanent visibility for the calendar.

New props
-----------
* `isAlwaysVisible`: enables permanent visibility by calling Pikaday's `show()` method right after instantiation and setting Pikaday's `hide()` method to a no-op
* `customInput`: allows replacement of the default input field with a custom input field

New stories
------------
* An example of a limited date range (uses existing props only)
* An always-visible example with a custom input field


src/components/date-picker/DatePicker.stories.tsx
```diff
@@ -53,6 +56,136 @@ export const manuallyEditable = () => (
...
+                        maxDate={maxDate}
+                        value={state.date}
+                    />
+                </IntlProvider>
+            )}
+        </State>
+    );
+};
+
+export const alwaysVisibleWithHiddenInput = () => {
```
alexkrolick, line 86: Is the component keyboard-accessible in this configuration? It should be OK with the separate input, but not sure about hiding the text box with no replacement.

sahiga: Good call, @alexkrolick. I replaced the input hiding logic with a new custom input field prop and added a visual regression test that demonstrates keyboard selection. Could you review again?

src/components/date-picker/DatePicker.tsx
```diff
@@ -442,24 +467,32 @@ class DatePicker extends React.Component<DatePickerProps> {
...
-                                    onKeyDown={this.handleInputKeyDown}
-                                    {...resinTargetAttr}
-                                    {...ariaAttrs}
-                                    {...inputProps}
-                                    {...valueAttr}
-                                />
+                                {customInput ? (
+                                    React.cloneElement(customInput, {
+                                        ref: (ref: HTMLInputElement) => {
+                                            this.dateInputEl = ref;
```
alexkrolick, line 473: Does the custom input field need to inherit the disabled/required states?

sahiga: Fixed!

alexkrolick (approved):

PR #2487, opened 2021-03-05
fix(uploads): update multiput support for mobile safari

danubevictoria:
Multiput uploads is not supported on Mobile Safari, and instead will use Plain Upload for all file sizes. This is due to a WebkitBlobResourceError when blob urls are attempted to be read to upload file data. The blob urls are created internally by the browser.

### DIFF
--- a/src/utils/Browser.js
+++ b/src/utils/Browser.js
@@ -27,7 +27,7 @@ class Browser {
      * @return {number} navigator maxTouchPoints
      */
     static getMaxTouchPoints(): number {
-        const maxTouchPoints = global.navigator.maxTouchPoints;
+        const { maxTouchPoints } = global.navigator;
         return !!maxTouchPoints && typeof maxTouchPoints === 'number' ? maxTouchPoints : 0;
     }
 
--- a/src/utils/Browser.js
+++ b/src/utils/Browser.js
@@ -18,13 +18,35 @@ class Browser {
     }
 
     /**
-     * Returns whether browser is mobile.
+     * Returns the max touch points for touchscreen devices
+     * Helps in mocking out.
+     *
+     * Only supported in the following browsers:
+     * @see https://developer.mozilla.org/en-US/docs/Web/API/Navigator/maxTouchPoints
+     *
+     * @return {number} navigator maxTouchPoints
+     */
+    static getMaxTouchPoints(): number {
+        const maxTouchPoints = global.navigator.maxTouchPoints;
+        return !!maxTouchPoints && typeof maxTouchPoints === 'number' ? maxTouchPoints : 0;
+    }
+
+    /**
+     * Returns whether browser is mobile, including tablets.
+     *
+     * We rely on user agent (UA) to avoid matching desktops with touchscreens.
+     * @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent#mobile_tablet_or_desktop
+     *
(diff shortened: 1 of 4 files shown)


src/utils/Browser.js
```diff
@@ -18,13 +18,35 @@ class Browser {
     }
 
     /**
-     * Returns whether browser is mobile.
+     * Returns the max touch points for touchscreen devices
+     * Helps in mocking out.
+     *
+     * Only supported in the following browsers:
+     * @see https://developer.mozilla.org/en-US/docs/Web/API/Navigator/maxTouchPoints
```
danubevictoria, line 25: See note about browser support. Particularly for the Mobile Safari use case, it is only available on iOS 13+.

src/utils/Browser.js
```diff
@@ -18,13 +18,35 @@ class Browser {
...
      *
      * @return {boolean} Whether browser is mobile
      */
     static isMobile(): boolean {
-        // Relying on the user agent to avoid desktop browsers on machines with touch screens.
-        return /iphone|ipad|ipod|android|blackberry|bb10|mini|windows\sce|palm/i.test(Browser.getUserAgent());
+        return (
+            /iphone|ipad|ipod|android|blackberry|bb10|mini|windows\sce|palm/i.test(Browser.getUserAgent()) ||
+            /Mobi/i.test(Browser.getUserAgent()) ||
+            (/Macintosh/i.test(Browser.getUserAgent()) && Browser.getMaxTouchPoints() > 0)
```
danubevictoria, line 48: Would like feedback if we should include this check for iOS devices with touchscreens. This would encompass use case where iOS users have a device with Request Desktop Website enabled, which changes the user agent to a desktop user agent, which cannot be pattern matched with 'Mobi'.

src/utils/uploads.js
```diff
@@ -335,11 +337,17 @@ function getDataTransferItemId(
...
 /**
- * Multiput uploads require the use of crypto, which is only supported in secure contexts
+ * Multiput uploads require the use of crypto, which is only supported in secure contexts.
+ * Multiput uploads is not supported on mobile iOS Safari.
  */
 function isMultiputSupported(): boolean {
     const cryptoObj = window.crypto || window.msCrypto;
-    return window.location.protocol === 'https:' && cryptoObj && cryptoObj.subtle;
+
+    if (Browser.isMobile() && Browser.isSafari()) {
```
danubevictoria, line 346: Even though Mobile Safari versions 11.0 & 11.1 work with multiput uploads, we are turning it off for all since user agent sniffing of the version number could be unreliable.

src/utils/Browser.js
danubevictoria: We will no longer be including this check.

src/utils/Browser.js
danubevictoria: Removing function.

src/utils/uploads.js
```diff
@@ -335,10 +337,16 @@ function getDataTransferItemId(
 }
 
 /**
- * Multiput uploads require the use of crypto, which is only supported in secure contexts
+ * Multiput uploads require the use of crypto, which is only supported in secure contexts.
+ * Multiput uploads is not supported on mobile iOS Safari.
```
shahzadaziz, line 341: nit: I am wondering if we should add any more details here, around why we took the decision to not support it ?

shahzadaziz (approved):
🚢

PR #2488, opened 2021-03-09
feat(content-explorer): Pass in initial select items

BLamy:


sahiga (commented):
Can you add a unit test for this change?

sahiga (approved):

PR #2491, opened 2021-03-12
feat(assets): update download16 icon to correct color

JohnnyB83:


wlyhdg (approved):
discussed the change reason with me. looks good.

PR #2492, opened 2021-03-13
feat(collapsiblesidebarmenuitem): add tooltipisdisabled prop

megansmith-box:
Add the ability to disable tooltip, through an optional prop.


atitoa93 (approved):
LGTM! Just wondering if these two files (`CollapsibleSidebar.stories.js`, `CollapsibleSidebar.stories.md`) need to be updated with the new props.

alexkrolick:
The tooltip text is used for the collapsed state - will that still be handled by the consumer of the component?

megansmith-box:
Hi @alexkrolick - The goal of this change is to be able to hide the tooltip even in the collapsed state. We were facing a bug in which there were overlapping tooltips. For example:

Screen Shot 2021-03-16 at 8 44 12 AM

The default will be false in order to not affect any current places were this is used. We will change `isTooltipDisbabled` to `true` in the consumer of the component when another tooltip is showing.

I also noticed, that if you first see the expanded nav, you will see no grey tooltip, then narrow the screen to the collapsed state, you will see the tooltip. And then you widen the screen to see the expanded nav again, you will still see the tooltip. But I don't think it's expected to see the grey tooltip in the expanded state.

Therefore I'm thinking of in the consumer component, setting `isTooltipDisbabled = canShowToolTip || navIsExpanded` Do you think this makes sense?

I don't fully understand what case `!isTextOverflowed` covers. Is it supposed to mean that the nav is in collapsed state? Or nav is in collapsed state or text doesn't fit in the expanded state.

Let me know your thoughts! Thank you!

megansmith-box:
@atitoa93 - Good Idea! I will add the prop to storybook

alexkrolick:
@megansmith-box OK sounds good

wlyhdg (approved):
thank you for your comments to clarify. it looks good to me in that case.

atitoa93 (dismissed):

megansmith-box:
Waiting to verify that there is no accessibility issue

box-internship-take-home (commented):
can you add a test for Tooltip?

I think Tooltip.js.flow needs to be modified as well

src/components/tooltip/Tooltip.tsx
```diff
@@ -100,6 +100,8 @@ export type TooltipProps = {
     stopBubble?: boolean;
     /** Text to show in the tooltip */
     text?: React.ReactNode;
+    /** True when should hide tooltip (usually when another tooltip is showing) */
+    shouldHideTooltip?: boolean;
```
DanDeMicco, line 104: do you need to add this prop? can you use an existing one instead e.g. isShown?

src/components/tooltip/Tooltip.js.flow
```diff
@@ -43,6 +43,8 @@ type Props = {
     onDismiss?: () => void,
     /** Where to position the tooltip relative to the wrapped component */
     position?: Position | CustomPosition,
+    /** True when should hide tooltip (usually when another tooltip is showing) */
+    shouldHideTooltip?: boolean,
```
DanDeMicco, line 47: can you explain why you need this and can't use `isShown`?

src/components/tooltip/Tooltip.tsx
megansmith-box: Thank you for catching this, I can use isShown instead

src/components/tooltip/Tooltip.js.flow
megansmith-box: same

src/components/tooltip/Tooltip.tsx
megansmith-box: @DanDeMicco - ready for review again

JohnnyB83 (approved):

PR #2493, opened 2021-03-20
feat(time-input): Add TimeInput component

sahiga:
This PR adds a new component that validates and parses time inputs.

time-input


alexkrolick (approved):

src/components/time-input/TimeInput.tsx
```diff
@@ -0,0 +1,95 @@
...
+    /** label - Label for the time input */
+    label?: React.ReactNode;
+    /** initialDate - Date object for initializing the time input */
+    initialDate?: Date;
+    /** isRequired - Whether the time input is required */
+    isRequired?: boolean;
+    /** onBlur - Function to call when the user blurs out of the time input */
+    onBlur?: Function;
+    /** onChange - Function to call when the user edits the time input */
+    onChange?: Function;
```
alexkrolick, line 35: Can you type the parameter the event handlers will be called with?

src/components/time-input/TimeInput.scss
```diff
@@ -0,0 +1,9 @@
+@import '../../styles/variables';
+
+.bdl-TimeInput {
+    .bdl-TimeInput-icon {
+        display: inline-block;
+        margin-bottom: -3px;
+        margin-left: -24px;
```
alexkrolick, line 7: Assuming these have some relationship to hardcoded dimensions elsewhere. If not, can the value come from the grid units?

src/components/time-input/TimeInput.tsx
sahiga: Fixed!

src/components/time-input/TimeInput.scss
sahiga: Yup, these are based on the margins for the verified icon in https://github.com/box/box-ui-elements/blob/master/src/components/text-input/TextInput.scss#L16. The clock is a little taller than the verified icon, though, so the vertical margins have a one-pixel difference.

src/components/time-input/TimeInput.tsx
```diff
@@ -0,0 +1,98 @@
...
+    errorTooltipPosition?: TooltipPosition;
+    /** hideLabel - Whether the label should be hidden */
+    hideLabel?: boolean;
+    /** label - Label for the time input */
+    label?: React.ReactNode;
+    /** initialDate - Date object for initializing the time input */
+    initialDate?: Date;
+    /** isRequired - Whether the time input is required */
+    isRequired?: boolean;
+    /**
```
alexkrolick, line 32: So, you'd always want the value from `onBlur`, right? `onChange` isn't parsed. I wonder if that creates any edge cases. For example, in a `<form>`, if I press cmd-enter will the blur event be called?

alexkrolick: I also wonder if you want to do something to enable refs to work.

sahiga: Good call. I removed the `onChange` prop and added handlers for certain keys (escape, enter, command, control).

src/components/time-input/TimeInput.tsx
```diff
@@ -27,33 +34,40 @@ export interface TimeInputProps extends WrappedComponentProps {
...
     /** isRequired - Whether the time input is required */
     isRequired?: boolean;
     /**
      * onBlur - Function to call when the user blurs out of the time input
      * The parsed display time, along with the hours and minutes in 24-hour format, will be passed to the handler.
      */
     onBlur?: ({ displayTime, hours, minutes }: { displayTime: string; hours: number; minutes: number }) => void;
-    /** onChange - Function to call when the user edits the time input */
-    onChange?: (value: string) => void;
+    /** shouldAutoBlur - Whether the field should automatically blur when form submit keys are pressed */
```
alexkrolick, line 46: Since it's not changing the focus target, I think this would be like "format or validate on <something>" rather than blur?

alexkrolick: Bringing back `onChange` but debounced + with this type of auto validation could also make sense

sahiga: Thanks for the suggestions, @alexkrolick. I updated the PR to include the debounced formatting function -- could you review again?

src/components/time-input/TimeInput.tsx
```diff
@@ -57,46 +69,64 @@ const TimeInput = ({
...
     const handleKeyDown = (event: React.KeyboardEvent) => {
         const { ctrlKey, key, metaKey } = event;
         const autoBlurKeyPressed = AUTO_BLUR_KEYS[key] || metaKey || ctrlKey;
-        if (autoBlurKeyPressed) handleBlur();
+        if (autoBlurKeyPressed) {
+            formatDisplayTime(displayTime, parsedData => {
+                if (onBlur) onBlur(parsedData);
+                if (onChange) onChange(parsedData);
+            });
+        }
```
alexkrolick, line 113: I don't think you need the keydown handler with debounced onChange and onBlur implemented as they are now

src/components/time-input/TimeInput.tsx
```diff
@@ -105,6 +135,14 @@ const TimeInput = ({
         } = event;
         setDisplayTime(updatedValue);
         if (error) setError(undefined);
+        if (shouldAutoFormat) debouncedFormatDisplayTime(updatedValue);
```
alexkrolick, line 138: I think you can remove the option and do this unconditionally

src/components/time-input/TimeInput.tsx
```diff
@@ -57,46 +69,64 @@ const TimeInput = ({
...
+        }
     };
 
+    /**
+     * Debounce formatDisplayTime() for use in handleChange().
+     * useCallback() memoizes the debounced function, so that the debounced function
+     * is not recreated on every re-render triggered by handleChange().
+     */
+    const debouncedFormatDisplayTime = React.useCallback(
+        debounce((latestValue: string) => formatDisplayTime(latestValue, onChange), 2000),
```
alexkrolick, line 122: That's a pretty high debounce time

src/components/time-input/TimeInput.tsx
sahiga: Yeah, I initially retained `shouldAutoFormat` to provide the option of formatting only on blur, but I think it's fine to run this function unconditionally

alexkrolick (approved):

swfree (approved):

PR #2499, opened 2021-03-24
feat(activityfeed): Only show time if within the last 12 months

ConradJChan:
Modify the `ActivityTimestamp` component to conditionally show the time only when the timestamp occurs in the last 12 months

Screen Shot 2021-03-24 at 1 27 46 PM


alexkrolick:
Isn't this experience a bit weird around the new year? Could we use a rolling offset like 3 months or something?

ConradJChan:
@alexkrolick Good point, I'll look into that, thanks!

jstoffan (approved):

jstoffan (approved):

PR #2502, opened 2021-03-25
fix(dropdownmenu): Ensure onMenuClose gets called consistently

ConradJChan:


src/components/dropdown-menu/DropdownMenu.js
```diff
@@ -147,10 +151,8 @@ class DropdownMenu extends React.Component<Props, State> {
     };
 
     handleMenuClose = (isKeyboardEvent: boolean, event: SyntheticEvent<> | MouseEvent) => {
-        const { onMenuClose = noop } = this.props;
-        this.closeMenu();
+        this.closeMenu(event);
         this.focusButton();
-        onMenuClose(event);
```
jstoffan, line 153: It seems like this could cause subtle bugs if the callback logic expects focus to be on the button before it's called. Have we seen anything like that with this change?

ConradJChan: Seems like that behavior would probably still be intact? Since the `onMenuClose` is called as a callback after `setState` is guaranteed to have completed the updates. But I did a search on uses and it looks like only RecentsCarousel uses the callback in the webapp

jstoffan (approved):

PR #2503, opened 2021-03-25
fix(NavButton): A11y add aria label navbar button

fsuarez-globant:
Fix: Add aria-label in Navbar button Component
I added an arial-label prop in SidebarNav and NavButton Components in order to improve accessibility.


src/elements/content-sidebar/SidebarNavButton.js
```diff
@@ -29,6 +30,7 @@ const SidebarNavButton = React.forwardRef<Props, React.Ref<any>>((props: Props,
         isOpen,
         sidebarView,
         tooltip,
+        'aria-label': dataAriaLabel,
```
alexkrolick, line 33: nit: variable name doesn't need `data` at the beginning, that's for the data-* attributes
Can we add a test or update the snapshots for this file showing that the DOM node receives the label?

fsuarez-globant: @alexkrolick I just changed the aria-label variable name and I also added an unit test in `SidebarNavButton.test.js `

alexkrolick (approved):

src/elements/content-sidebar/SidebarNav.js
```diff
@@ -61,6 +61,7 @@ const SidebarNav = ({
                         data-testid="sidebaractivity"
                         sidebarView={SIDEBAR_VIEW_ACTIVITY}
                         tooltip={<FormattedMessage {...messages.sidebarActivityTitle} />}
+                        aria-label={intl.formatMessage(messages.sidebarActivityTitle)}
```
jstoffan, line 64: We usually try to keep our attributes alphabetized, even though we have the lint rule disabled in this project.

src/elements/content-sidebar/SidebarNav.js
```diff
@@ -61,6 +61,7 @@ const SidebarNav = ({
                         data-testid="sidebaractivity"
                         sidebarView={SIDEBAR_VIEW_ACTIVITY}
                         tooltip={<FormattedMessage {...messages.sidebarActivityTitle} />}
```
jstoffan, line 63: Should we have the SidebarNavButton component apply the `aria-label` attribute, instead? It would save us from having to pass it twice.

fsuarez-globant: @jstoffan We should add `aria-label` attribute directly in HTML Elements. In this case we should pass that attribute until it is applied to an HTML element. I found an example where they do something similar https://github.com/box/box-ui-elements/blob/85c005a4de9cae201d3a879d08277ac9eafe5b89/src/components/grid-view/GridViewSlider.js#L38

src/elements/content-sidebar/SidebarNav.js
fsuarez-globant: @jstoffan I will fix this. Thanks

src/elements/content-sidebar/SidebarNav.js
jstoffan: I'm sorry, I think we're talking about different things. I'm suggesting that we reuse the localized string for tooltip by changing this line to:

```
tooltip={formatMessage(messages.sidebarActivityTitle)}
```

Then, in SidebarNavButton.js, we can add:

```
<Tooltip position="middle-left" text={tooltip} isTabbable={false}>
    <NavButton
        aria-label={tooltip}
        // ...other props ...
```

What do you think?

fsuarez-globant: @jstoffan sorry I misunderstood what are saying in the previous comment. I agree with you. I will update my PR. Thank you!

jstoffan: No problem, thanks for taking a look!

jstoffan (approved):
Nice! LGTM.

PR #2507, opened 2021-03-26
fix(preview): Made status pills larger to improve accessibility

msorianoBox:
Status pills on the preview page had accessibility issues, since the text was too small and could be troublesome to read for people with visual disabilities. The pills have a "Large" version which is used in this fix.

### DIFF
--- a/src/components/label-pill/LabelPill.tsx
+++ b/src/components/label-pill/LabelPill.tsx
@@ -33,7 +33,7 @@ export interface LabelPillProps {
 }
 
 const LabelPillContainer = React.forwardRef((props: LabelPillProps, ref: React.Ref<HTMLSpanElement>) => {
-    const { children, type = LabelPillStatus.DEFAULT, size = LabelPillSize.REGULAR, className, ...rest } = props;
+    const { children, type = LabelPillStatus.DEFAULT, size = LabelPillSize.LARGE, className, ...rest } = props;
     const labelPillClasses = classNames(
         'bdl-LabelPill',
         `bdl-LabelPill--${type}`,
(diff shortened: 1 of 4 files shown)


src/components/label-pill/LabelPill.tsx
```diff
@@ -33,7 +33,7 @@ export interface LabelPillProps {
 }
 
 const LabelPillContainer = React.forwardRef((props: LabelPillProps, ref: React.Ref<HTMLSpanElement>) => {
-    const { children, type = LabelPillStatus.DEFAULT, size = LabelPillSize.REGULAR, className, ...rest } = props;
+    const { children, type = LabelPillStatus.DEFAULT, size = LabelPillSize.LARGE, className, ...rest } = props;
```
alexkrolick, line 36: This is changing the default pill size everywhere. Is that intentional or is there one place where we should be passing the size prop as "large"?

msorianoBox: The change is needed in every place where the pill is, since the size is too small. :) This will fix all

alexkrolick: In that case it seems like we should adjust the values of the size enums to preserve relative sizing

alexkrolick: We'll likely have to review the usage in a number of places to make sure there aren't layout changes that need work to fix

alexkrolick: We need to discuss with the design team and make any changes to accommodate the new sizes where Pill elements are used

msorianoBox:
This is the current view os status pills, accessibility audit pointed out that the text was too small, so size would have to increase.

Screen Shot 2021-04-01 at 11 38 33

Among the classes on BUIE, there is this "large" size which basically tackles the problem with an existing option:

Screen Shot 2021-04-01 at 11 39 17

Now we need confirmation to know if this is acceptable or should we try another path to solve this issue?

PR #2508, opened 2021-03-29
feat(annotations): Selectable annotation activity card

ConradJChan:
Annotation Card Selectable

TODO:

- [x] unit tests


src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
```diff
@@ -84,67 +104,76 @@ const AnnotationActivity = ({
...
+                        ) : (
+                            <ActivityMessage id={id} tagged_message={message} getUserProfileUrl={getUserProfileUrl} />
+                        )}
+                    </Media.Body>
+                </Media>
+                {/* $FlowFixMe */}
+                {error ? <ActivityError {...error} /> : null}
+            </SelectableActivityCard>
+            {isMenuVisible && (
+                <AnnotationActivityMenu
```
jstoffan, line 166: It seems a little odd that this is totally outside the parent card. Can this stay within the Media body? If not, should it be moved to a parent component?

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
```diff
@@ -48,7 +47,10 @@ const AnnotationActivity = ({
     onSelect = noop,
 }: Props) => {
     const [isEditing, setIsEditing] = React.useState(false);
+    const [isMenuOpen, setIsMenuOpen] = React.useState(false);
     const { created_at, created_by, description, error, file_version, id, isPending, permissions = {}, target } = item;
+    const isFileVersionUnavailable = file_version === null;
```
jstoffan, line 52: Can we group all of the derived state together either before or after the handlers? The split is a little confusing right now.

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
```diff
@@ -71,10 +73,28 @@ const AnnotationActivity = ({
         onEdit(id, text, permissions);
     };
 
+    const handleMenuClose = (): void => {
+        setIsMenuOpen(false);
```
jstoffan, line 77: Given that these event handlers are pretty numerous, could we either inline them in the JSX or shorten the helpers?

```
const handleMenuClose = (): void => setIsMenuOpen(false);
const handleMenuOpen = (): void => setIsMenuOpen(true);

File: src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityMenu.js [hunk 4]
```diff
@@ -34,18 +46,38 @@ const AnnotationActivityMenu = ({ canDelete, canEdit, id, onDeleteConfirm, onEdi
         onDeleteConfirm();
     };
 
+    const handleMenuClose = () => {
+        setIsMenuOpen(false);
```
jstoffan, line 50: Can we track this state only in the parent and make this a controlled component?

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityMenu.js
```diff
@@ -34,18 +46,38 @@ const AnnotationActivityMenu = ({ canDelete, canEdit, id, onDeleteConfirm, onEdi
...
     const tetherProps = {
         attachment: 'top right',
         className: 'bcs-AnnotationActivity-deleteConfirmationModal',
         constraints: [{ to: 'scrollParent', attachment: 'together' }],
         targetAttachment: 'bottom right',
     };
 
+    React.useEffect(() => {
+        if (isConfirmingDelete || isMenuOpen) {
+            onMenuOpen();
```
jstoffan, line 66: Can we pass these props straight through to `dropdownProps` rather than using an effect?

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityMenu.js
```diff
@@ -34,18 +46,38 @@ const AnnotationActivityMenu = ({ canDelete, canEdit, id, onDeleteConfirm, onEdi
...
+
     const tetherProps = {
         attachment: 'top right',
         className: 'bcs-AnnotationActivity-deleteConfirmationModal',
         constraints: [{ to: 'scrollParent', attachment: 'together' }],
         targetAttachment: 'bottom right',
     };
 
+    React.useEffect(() => {
+        if (isConfirmingDelete || isMenuOpen) {
```
jstoffan, line 65: Is the `isConfirmingDelete` check needed? It looks like the menu is still open in that scenario, just disabled.

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityTimestamp.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+const AnnotationActivityTimestamp = ({
+    date,
+    id,
+    isDisabled = false,
+    message,
+    onAnnotationSelect,
+    ...rest
+}: Props): JSX.Element => (
+    <div className="bcs-AnnotationActivityTimestamp" {...rest}>
```
jstoffan, line 24: Does this need to be its own component? It doesn't seem like its children share any commonalities and the naming doesn't imply it contains a link.

src/elements/content-sidebar/activity-feed/SelectableActivityCard.scss
```diff
@@ -0,0 +1,24 @@
...
+}
+
+.bcs-SelectableActivityCard {
+    &[aria-disabled='false'] {
+        cursor: pointer;
+
+        &:hover,
+        &:focus {
+            outline: none;
+            box-shadow: 0 0 8px 0 rgba(114, 132, 142, .16);
```
jstoffan, line 20: Do we have a variable for this color?

src/elements/content-sidebar/activity-feed/SelectableActivityCard.scss
```diff
@@ -0,0 +1,24 @@
...
+            animation: bcs-SelectableActivityCard-hover 1s ease-in-out 25ms 1 normal forwards;
```
jstoffan, line 21: Is the one second duration a little long for a hover effect? Should this be a transition, since it only plays once? If it needs to be an animation, it looks like some of these properties are their default values and can be omitted.

ConradJChan: Good point, I'll adjust this

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityMenu.js
ConradJChan: The end goal is to maintain a `isMenuOpen` state that consists of the `DropdownMenu` and `DeleteConfirmation` states. Passing the props straight to `dropdownProps` wouldn't be sufficient alone

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityMenu.js
ConradJChan: Yeah I need to tweak this effect a bit

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
alexkrolick: There's a Media.Menu component which ideally would be compatible with this use case. It goes inside the Media as a sibling of Media.Body

ConradJChan: As discussed, the reason this menu has to live outside is to avoid event bubbling issues for events that occur within the dropdown menu which interferes with the `mousedown` event processing that is down to power the deselect annotation logic

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
```diff
@@ -48,103 +48,116 @@ const AnnotationActivity = ({
...
+        if (isCardDisabled) {
+            return;
+        }
+
+        // Prevents document event handlers from executing because box-annotations relies on
+        // detecting mouse events on the document outside of annotation targets to determine when to
+        // deselect annotations
+        event.stopPropagation();
+    };
+    const handleOnSelect = () => onSelect(item);
```
jstoffan, line 77: Should this be `handleCardSelect`?

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.scss
```diff
@@ -35,6 +34,24 @@
     z-index: $overlay-z-index;
 }
 
+.bcs-AnnotationActivityMenu {
```
jstoffan, line 37: Should this live in its component's stylesheet? Or should we pass it a class name like `bcs-AnnotationActivity-menu` to stay consistent with SUIT naming?

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
alexkrolick: Is there something we can fix about that in the Media element? I think in other parts of the app we added `evt.stopPropagation()` to some of the Media menu item event handlers to address some of that.

src/elements/content-sidebar/activity-feed/SelectableActivityCard.scss
alexkrolick: Probably worth asking design if they want to add this box-shadow style to the design system options

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
ConradJChan: Yeah @jstoffan it's related to the `mousedown` handler in `AnnotationActivity`. If the menu is a child of `SelectableActivityCard` then there are issues when the card is selected and the menu is opened -- the `mousedown` is prevented from propagating up to the logic that would deselect the annotation. 

@alexkrolick I'm not sure, this use case is very specific to how annotations works and I had done a quick search and there were only 1 or 2 other instances where these types of event handling occurred so maybe it's not widely an issue

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.scss
```diff
@@ -19,7 +19,6 @@
 // Need to be more specific to override btn-plain margin:0
```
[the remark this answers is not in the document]
ConradJChan: Doesn't seem like it, good catch

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityMenu.js
```diff
@@ -34,18 +47,45 @@ const AnnotationActivityMenu = ({ canDelete, canEdit, id, onDeleteConfirm, onEdi
...
+
     const tetherProps = {
         attachment: 'top right',
         className: 'bcs-AnnotationActivity-deleteConfirmationModal',
         constraints: [{ to: 'scrollParent', attachment: 'together' }],
         targetAttachment: 'bottom right',
     };
 
+    React.useEffect(() => {
+        const { current: isMounted } = isMountedRef;
```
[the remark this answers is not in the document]
ConradJChan: Even if I do that, the ref will have the value for `current` when the effect runs after first mount -- which will lead to `onMenuClose` being called because `!isConfirmingDelete && !isMenuOpen` is satisfied, so what I'm going for is essentially a "don't run the first time" effect

src/elements/content-sidebar/activity-feed/SelectableActivityCard.scss
```diff
@@ -0,0 +1,24 @@
...
+
+.bcs-SelectableActivityCard {
+    &[aria-disabled='false'] {
+        cursor: pointer;
+
+        &:hover,
+        &:focus {
+            outline: none;
+            box-shadow: 0 0 8px 0 rgba(114, 132, 142, .16);
+            animation: bcs-SelectableActivityCard-hover 300ms ease-in-out 25ms 1 normal forwards;
```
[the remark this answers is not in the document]
ConradJChan: I'll address this along with https://github.com/box/box-ui-elements/pull/2508/files#r604374585 in a separate PR since I want the `bcs-is-focused` animation to override this one for now

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
```diff
@@ -52,12 +52,14 @@
             transition: box-shadow $container-transition-duration ease $container-transition-delay;
 
             &.bcs-is-focused {
```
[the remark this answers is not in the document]
ConradJChan: Will address in a separate PR

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityMenu.js
```diff
@@ -13,78 +12,59 @@ import { MenuItem } from '../../../../components/menu';
...
+    canEdit,
+    className,
+    id,
+    isDisabled,
+    onDelete,
+    onEdit,
+    onMenuClose,
+    onMenuOpen,
+}: AnnotationActivityMenuProps) => (
+    <Media.Menu
```
jstoffan, line 35: Should this still be a `Media.Menu` since it's been moved out of its `Media` parent?

jstoffan (approved):
LGTM, thanks for working to iterate on it with me!

jstoffan (approved):

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
```diff
@@ -52,12 +52,14 @@
...
-                animation: bcs-ActivityItem-callout 1s ease-in-out $container-transition-delay 0 normal none; // Disabled by default
-
-                &:not(:focus-within) {
-                    animation-iteration-count: 1; // Only play the animation if the user is not interacting directly with the element or its descendents
-                    animation-fill-mode: forwards;
+                .bcs-ActivityCard {
+                    box-shadow: 0 0 $bdl-grid-unit * 4 $bdl-gray-30;
+                    animation: bcs-ActivityItem-callout 1s ease-in-out $container-transition-delay 0 normal none; // Disabled by default
+
+                    &:not(:focus-within) {
```
alexkrolick, line 59: :focus-within isn't supported on IE, but this looks non-critical (and is already in place)

alexkrolick (approved):

PR #2510, opened 2021-03-31
fix(typography): add minimum allowed font size

msorianoBox:
fix(all files): added new font-size const fo fix accesibility issues

Needed to add a new cont for font-size (11px). This prevents a drastic layout change in order to improve accesibility.


alexkrolick:
Is this meant to represent the minimum allowed font size?

msorianoBox:
> Is this meant to represent the minimum allowed font size?

Basically it does. the 9px font size triggers the "Text too small" warning

alexkrolick:
Can we make the variable name `fontSize-small` or `fontSize-minimum` then?

sahiga (approved):

PR #2513, opened 2021-04-01
fix(annotations): Truncate annotation link when too long

ConradJChan:
Screen Shot 2021-04-01 at 3 17 34 PM

TODO
 - Cross browser testing

### DIFF
--- a/src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.scss
+++ b/src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.scss
@@ -39,12 +39,18 @@
     z-index: 1;
 }
 
-.bcs-AnnotationActivity-timestamp {
-    white-space: nowrap;
-
-    .bcs-AnnotationActivity-link {
-        font-size: 12px;
+.bcs-AnnotationActivity-link {
+    &::before {
+        margin-right: 8px;
+        margin-left: 8px;
+        color: $bdl-gray-50;
+        content: '\2022';
     }
+}
+
+.bcs-AnnotationActivity-timestamp {
+    display: flex;
+    white-space: nowrap;
 }
 
 // override for activity feed defaults
@@ -56,19 +62,8 @@
     }
 
     .bcs-AnnotationActivity-link {
-        display: inline-block;
-        color: $bdl-box-blue;
+        display: flex;
         font-weight: bold;
-
-        &[aria-disabled='true'] {
-            color: $bdl-gray-62;
-        }
-
-        &::before {
-            margin-right: 8px;
-            margin-left: 8px;
-            color: $bdl-gray-50;
-            content: '\2022';
-        }
+        font-size: 12px;
     }
 }
--- a/src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityLink.scss
+++ b/src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityLink.scss
@@ -1,0 +1,18 @@
+@import '../../../common/variables';
+
+.bcs-AnnotationActivityLink {
+    display: inline-block;
+    overflow-x: hidden;
+    color: $bdl-box-blue;
+    font-weight: bold;
+
+    &[aria-disabled='true'] {
+        color: $bdl-gray-62;
+    }
+
+    .bcs-AnnotationActivityLink-message {
+        overflow-x: hidden;
+        white-space: nowrap;
+        text-overflow: ellipsis;
+    }
+}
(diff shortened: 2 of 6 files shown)


src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.scss
```diff
@@ -39,12 +39,18 @@
...
+    &::before {
+        margin-right: 8px;
+        margin-left: 8px;
+        color: $bdl-gray-50;
+        content: '\2022';
+    }
+}
+
 .bcs-AnnotationActivity-timestamp {
+    display: flex;
```
jstoffan, line 52: Were we able to validate this works properly in IE11? It has some weird quirks around flexbox.

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityLink.scss
```diff
@@ -0,0 +1,18 @@
+@import '../../../common/variables';
+
+.bcs-AnnotationActivityLink {
+    display: inline-block;
+    overflow-x: hidden;
```
jstoffan, line 5: Do we need this on both the parent and the message?

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityLink.tsx
```diff
@@ -1,23 +1,34 @@
...
+import classNames from 'classnames';
 import noop from 'lodash/noop';
-import { FormattedMessage, MessageDescriptor } from 'react-intl';
+import { injectIntl, MessageDescriptor, WrappedComponentProps } from 'react-intl';
 import PlainButton from '../../../../components/plain-button';
 import { ButtonType } from '../../../../components/button';
+import './AnnotationActivityLink.scss';
 
-export interface AnnotationActivityLinkProps {
+type MessageDescriptorWithValues = {
```
jstoffan, line 9: It seems like `react-intl` would have a type for this out of the box?

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityLink.tsx
```diff
@@ -1,23 +1,34 @@
...
 const AnnotationActivityLink = ({
+    className,
     id,
+    intl,
     isDisabled = false,
     message,
     onClick = noop,
     ...rest
 }: AnnotationActivityLinkProps): JSX.Element => {
+    const { values, ...messageDescriptor } = message;
```
jstoffan, line 29: Could we simplify this by passing in `isFileVersionUnavailable` and manage the message in here, instead?

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityLink.tsx
```diff
@@ -39,17 +50,20 @@ const AnnotationActivityLink = ({
 
     return (
         <PlainButton
-            className="bcs-AnnotationActivity-link"
+            className={classNames('bcs-AnnotationActivityLink', className)}
             data-testid="bcs-AnnotationActivity-link"
             isDisabled={isDisabled}
             onClick={handleClick}
             onMouseDown={handleMouseDown}
+            title={translatedMessage}
```
jstoffan, line 58: I think the timestamp uses a `Tooltip` component to display the full string. Should we do the same here to stay consistent, especially since they're siblings?

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.scss
ConradJChan: Yup just confirmed it works as expected

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityLink.scss
ConradJChan: Yeah seems like it

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityLink.tsx
ConradJChan: Design had said just to use the default browser tooltip. The `Tooltip` component also seems a bit aggressive -- as soon as hover occurs it popups up and in this case is exactly the same as the current text so the value add isn't very much

jstoffan: I had assumed that the title was used to display an untruncated version of the string, similar to the timestamp. Is that not the case? If not, do we need the title at all?

jstoffan (approved):

PR #2515, opened 2021-04-05
feat(classification): Update classification modified by copies

bxie-box:
Based on product decision, we want to update the modified by label copies and added a prop for modified action (which by default is true / always used unless in the case of MIP which we would use import action label instead). 

**Current**
Current

**With Update**
Updated


src/features/classification/Classification.js
```diff
@@ -28,22 +28,24 @@ type Props = {
     modifiedBy?: string,
     name?: string,
     onClick?: (event: SyntheticEvent<HTMLButtonElement>) => void,
+    useAppliedByLabel?: boolean,
```
bxie-box, line 31: Can flip to `useImportByLabel` boolean instead (since that by default would be false) -- would that be preferred?

src/features/classification/Classification.js
```diff
@@ -89,7 +91,15 @@ const Classification = ({
                     <p className="bdl-Classification-modifiedBy" data-testid="classification-modifiedby">
                         <FormattedMessage
                             {...messages.modifiedBy}
-                            values={{ modifiedAt: formattedModifiedAt, modifiedBy }}
+                            values={{
+                                modifiedActionLabel: useAppliedByLabel ? (
+                                    <FormattedMessage {...messages.appliedByText} />
+                                ) : (
+                                    <FormattedMessage {...messages.importedFromText} />
```
bxie-box, line 98: The action label is by default going to be `Applied by` (except in the instance it's by MIP which we will use `Imported from`) -- wondering if we'd prefer explicit individual flags like `useAppliedByLabel` `useImportedByLabel` and render it outside of this `FormattedMessage` as {`useAppliedByLabel && <FormattedMessage {...messages. appliedByText} />` or have an optional `useImportedByLabel` and fall back to `Applied By` when that optional flag is not existence.

i18n/en-US.properties
```diff
@@ -740,12 +742,14 @@ boxui.classification.classificationYellow = Yellow
 boxui.classification.definition = Definition
 # Button to edit classification on an item
 boxui.classification.edit = Edit
+# Text prepended to user or service that imported the classification.
+boxui.classification.importedFromText = Imported from
 # Default message for classification in the sidebar when there is none
 boxui.classification.missing = Not classified
 # Sentence stating which user or service modified a classification and on what date.
-boxui.classification.modifiedBy = {modifiedBy} on {modifiedAt}
+boxui.classification.modifiedBy = {modifiedActionLabel} {modifiedBy} on {modifiedAt}
```
jpfranco, line 750: It's usually not recommended to combine message fragments that were translated independently, since translators need to consider the text as a whole and the order of the fragments might be different depending on the language. We can probably have a separate `boxui.classification.importedBy` message in order to avoid these issues.

src/features/classification/Classification.js
jpfranco: Would it make sense to name it something like `isImportedClassification` so that it's not necessarily tied to the label? I think that way we can avoid the consumer of the component having to know when it is appropriate to use each label, or choosing a label incorrectly.

We could also use an enum if we expect that there could be more cases later on.

src/features/classification/Classification.js
```diff
@@ -89,7 +91,15 @@ const Classification = ({
                     <p className="bdl-Classification-modifiedBy" data-testid="classification-modifiedby">
                         <FormattedMessage
                             {...messages.modifiedBy}
-                            values={{ modifiedAt: formattedModifiedAt, modifiedBy }}
+                            values={{
```
jpfranco, line 94: style: I think for consistency might be good to follow the file's pattern of having conditions declared outside of the main jsx. So if we add the `importedBy` message I mentioned above, it could be something like:

```
const modifiedByMessage = isImportedClassification ? messages.importedBy : messages.modifiedBy;
```
And then:
```
<FormattedMessage {...modifiedByMessage} values={...} /> 
```

src/features/classification/Classification.js
jpfranco: I think this could probably be addressed with the prop name suggestion above. I'd say it's preferable to have a single prop since we will only be rendering one or the other.

src/features/classification/__tests__/Classification.test.js
```diff
@@ -77,16 +79,32 @@ describe('features/classification/Classification', () => {
...
-        });
-        expect(wrapper.exists('[data-testid="classification-modifiedby"]')).toBe(true);
-    });
+    test.each`
+        useAppliedByLabel | expectedActionLabelId
+        ${undefined}      | ${messages.appliedByText.id}
+        ${true}           | ${messages.appliedByText.id}
+        ${false}          | ${messages.importedFromText.id}
+    `(
+        'should render classification last modified information with expected action label when provided and message style is inline',
```
jpfranco, line 88: nit: Could we incorporate `useAppliedByLabel` and `expectedActionLabelId` to the message so that it's easier to figure out what is being tested from the command line?

bxie-box:
@jpfranco Updated based on comments and it's more consistent/neater. Thanks so much for the review!!

jpfranco (approved):

PR #2519, opened 2021-04-07
fix(flyout): opening with keyboard should be same as with mouse

alexkrolick:


src/components/flyout/Flyout.js
```diff
@@ -249,10 +249,18 @@ class Flyout extends React.Component<Props, State> {
             this.openOverlay();
         }
 
-        // If button was clicked, the detail field should hold number of clicks.
-        // If number is zero, the event was synthesized.
-        // https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/detail
-        const isButtonClicked = event.detail > 0;
```
alexkrolick, line 255: "event.detail" is the number of clicks for a mouse event. But pressing enter on a button also triggers the click event handler, so we shouldn't assume that a click count of 0 is a programmatic event. Instead, use .isTrusted.

frogzsj (approved):

PR #2523, opened 2021-04-08
feat(assets): update assets to latest

sethvoltz:


ehoogerbeets (approved):

src/styles/variables.scss
```diff
@@ -0,0 +1,123 @@
+// Box Design Language Colors
```
alexkrolick, line 1: this temp file shouldn't be checked in

alexkrolick (approved):

PR #2530, opened 2021-04-12
fix(typography): fixed low contrast text on preview

msorianoBox:
Fixed text on the Preview Panel with low contrast (updated value to bdl-gray-80)

Screen Shot 2021-04-12 at 10 06 33
Screen Shot 2021-04-12 at 10 06 25
Screen Shot 2021-04-12 at 10 03 07

### DIFF
--- a/src/features/item-details/ItemProperties.scss
+++ b/src/features/item-details/ItemProperties.scss
@@ -4,7 +4,7 @@
     margin: 0;
 
     dt {
-        color: $bdl-gray-62;
+        color: $bdl-gray-80;
     }
 
     dd {
--- a/src/elements/content-sidebar/activity-feed/comment-form/CommentForm.scss
+++ b/src/elements/content-sidebar/activity-feed/comment-form/CommentForm.scss
@@ -37,7 +37,7 @@
 
     .bcs-CommentForm-tip {
         margin-top: 10px;
-        color: $bdl-gray-50;
+        color: $bdl-gray-80;
     }
 
     .bcs-CommentFormControls {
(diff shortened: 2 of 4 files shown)


alexkrolick:
We do need to change the gray-50 text, but I thought gray-62 met contrast requirements against white backgrounds. Can you clarify why those need to be changed?

msorianoBox:
Gray-62 does meet the contrast requirements against white BG, but in the preview panel the BG color is #FBFBFB and that's why we had to improve the contrast.

alexkrolick (approved):

src/features/item-details/ItemProperties.scss
```diff
@@ -4,7 +4,7 @@
     margin: 0;
 
     dt {
-        color: $bdl-gray-62;
+        color: $bdl-gray-80;
```
jstoffan, line 7: Just as a heads-up, it looks like this change affects a shared component that is used both in the Details sidebar within preview and in other areas:

- Have we validated the visual appearance for the other usages?
- Can/should we use a lighter gray for those areas if they have a `#fff` background, per @alexkrolick's suggestion? 

Same questions for the Classification component.

msorianoBox: As you can see in the screenshots provided, the color difference it's barely perceivable but for the Accessibility audit it makes a whole difference. We're on the clock with these adjustments, so I think it could be an alternative to generate a small technical debt regarding the special case for white BG, since its not something of significant impact rather than a 20% darker gray on the font color.

Is it viable to do this?

jstoffan: We're going from `#909090` in Classification to `#4e4e4e`, so it's a pretty large color change in some places, to the point where the header isn't really distinguishable from the content. If the design team has signed off on the changes, though, the code itself LGTM.

jstoffan (approved):

greg-in-a-box:
closing because its stale, there seems to be a change with the exact same variable but to a different color and preview team cannot seem to find an corresponding ticket of the change

greg-in-a-box:
https://github.com/box/box-ui-elements/pull/3015 replaces this PR

PR #2532, opened 2021-04-13
feat(messagecenter): only show categories when more than 1 category

DanDeMicco:


src/features/message-center/components/message-center-modal/MessageCenterModal.js
```diff
@@ -227,7 +242,6 @@ function MessageCenterModal({
             className="bdl-MessageCenterModal"
             data-resin-component="messageCenterModal"
             data-testid="messagecentermodal"
-            focusElementSelector=".bdl-CategorySelector .is-selected"
```
megansmith-box, line 230: Will removing this affect the focused element for when there are categories?

DanDeMicco: yes it will, but it needs to be removed as the categories will now automatically show anymore and an error is thrown as it doing a querySelector on this on mount

megansmith-box (approved):

PR #2533, opened 2021-04-14
fix(time-input): Add support for el-GR, ko-KR, zh-CN, and zh-TW

sahiga:
This PR updates the `TimeInput` component to format p.m. times in Chinese, Korean, and Greek.


src/components/time-input/TimeInputUtils.ts
```diff
@@ -1,6 +1,6 @@
-const VALID_TIME_REGEX = /[0-9]{1,2}.?.?.?[0-9]{0,2}\s?[AaPp]?\.?\s?[Mm]?/;
-const AM_REGEX = /[Aa]\.?\s?[Mm]/;
-const PM_REGEX = /[Pp]\.?\s?[Mm]/;
+const VALID_TIME_REGEX = /[\u4e0a\u4e0b\uc624]?[\u5348\ud6c4\uc804]?[0-9]{1,2}.?.?.?[0-9]{0,2}\s?[AaPp\u03C0\u03BC]?\.?\s?[Mm\u03BC]?/;
```
ehoogerbeets, line 1: For this part:

```
[0-9]{1,2}.?.?.?[0-9]{0,2}
```

The three dots in the middle are all optional, so for non-conforming time formats, maybe we should make the middle one should not be optional? (Or does it not really matter?) Also, 2 digits are always required for the minutes, I think. So maybe like this:

```
[0-9]{1,2}.?..?[0-9]{2}
```

src/components/time-input/TimeInputUtils.ts
```diff
@@ -1,6 +1,6 @@
...
+const AM_REGEX = /[Aa\u4e0a\u03C0\uc624]\.?\s?[Mm\u5348\ud6c4\u03BC\uc804]/;
+const PM_REGEX = /[Pp\u4e0b\uc624\u03BC]\.?\s?[Mm\u5348\ud6c4\u03BC]/;
```
ehoogerbeets, line 3: I'm impressed you got all the characters, including the ones I didn't put in slack! You did your homework!

src/components/time-input/__tests__/TimeInputUtils.test.ts
```diff
@@ -74,6 +74,20 @@ const VALID_TIME_INPUTS = [
...
+    ['21:32', { hours: 21, minutes: 32 }],
+    ['4:32 PM', { hours: 16, minutes: 32 }],
+    ['23.32', { hours: 23, minutes: 32 }],
+    ['16 h 32', { hours: 16, minutes: 32 }],
+    ['5:32', { hours: 5, minutes: 32 }],
+    ['오전 5:32', { hours: 5, minutes: 32 }],
+    ['오후 5:32', { hours: 17, minutes: 32 }],
+    ['上午4:32', { hours: 4, minutes: 32 }],
+    ['下午4:32', { hours: 16, minutes: 32 }],
 ];
```
ehoogerbeets, line 91: What happens if someone puts in some other time format? like albanian:

```
morning:  1:32 e paradites
evening:  1:32 e pasdites
```

src/components/time-input/TimeInputUtils.ts
sahiga: `TimeInput` will accept 1-4 numbers, with or without a separator. I wanted to make the input as flexible as possible, so that users could type "300" or "3" and still get a correctly formatted time

src/components/time-input/__tests__/TimeInputUtils.test.ts
sahiga: Do we officially support Albanian? If so, I can add those to the am/pm regexes

src/components/time-input/TimeInputUtils.ts
ehoogerbeets: Oh, I see... so without a separator. Okay.

src/components/time-input/__tests__/TimeInputUtils.test.ts
ehoogerbeets: No, we don't, but if someone puts in something other than the 20-ish languages we do support, hopefully it comes up with something reasonable.

sahiga: Yup, the component will parse any input that has 1-4 numbers, regardless of what other symbols or letters are included. Both "1:32 e paradites" and "1:32 e pasdites" will get parsed as 1:32 in the morning.

ehoogerbeets (approved):

PR #2535, opened 2021-04-15
refactor(typescript): Migrate Badge

laughnan:


src/components/badge/Badge.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+import './Badge.scss';
+
+type Props = {
+    /** Child components */
+    children: React.ReactNode;
+    /** A predefined badge type */
+    className?: string;
+    /** Adds a class to the component */
+    type?: 'default' | 'info' | 'warning' | 'highlight' | 'error' | 'alert' | 'success';
```
sahiga, line 12: Could we convert this to an enum, similar to `ButtonType` in Button.tsx?

sahiga (commented):
Thanks for migrating these files!

src/components/badge/Badge.tsx
laughnan: Migrated to `enum`! https://github.com/box/box-ui-elements/pull/2535/commits/5b8736195ec22cf75f52cf8b411d04695c8f285f

I also exported that enum out via `index.ts` since I think that's necessary iirc.

src/components/badge/__tests__/Badge.test.tsx
```diff
@@ -19,21 +20,11 @@ describe('components/badge/Badge', () => {
         expect(wrapper.hasClass('some-badge-style')).toBe(true);
     });
 
-    test(`should render a badge with info styling when initialized`, () => {
-        const wrapper = shallow(<Badge type="info">test</Badge>);
+    [BadgeType.INFO, BadgeType.WARNING, BadgeType.HIGHLIGHT].forEach((type: BadgeType) => {
```
sahiga, line 23: Can we use `test.each` here instead?

laughnan: We can! Updated 🚀

sahiga (approved):

PR #2540, opened 2021-04-20
fix(content-explorer): Request fields for Preview SDK file cache

sahiga:
This PR updates the folder request fields in the Content Explorer to include additional fields required by the Preview SDK cache:
- file_version
- sha1
- shared_link
- watermark_info


jstoffan (approved):
Is there any performance impact on the associated calls after this change? It looks like this code is used by ContentPicker, as well, which doesn't support preview.

sahiga:
Good call, @jstoffan. I created a separate array for the Content Explorer folder fields.

src/utils/fields.js
```diff
@@ -80,9 +79,15 @@ const FOLDER_FIELDS_TO_FETCH = [
     FIELD_AUTHENTICATED_DOWNLOAD_URL,
     FIELD_IS_DOWNLOAD_AVAILABLE,
     FIELD_REPRESENTATIONS,
+    FIELD_URL,
+];
+
+// Fields needed for Content Explorer folder requests
+const CONTENT_EXPLORER_FOLDER_FIELDS_TO_FETCH = [
```
jstoffan, line 86: Should this constant live in a file within the `src/elements/content-explorer` directory?

sahiga: Yes! Just moved the constant into the `src/elements/content-explorer` directory.

jstoffan (approved):
Thanks for fixing this issue!

swfree (approved):

PR #2542, opened 2021-04-21
feat(preview/tooltip): pass preview experiences to preview

megansmith-box:
Pass PreviewExperiences From ContentPreview Component to Preview Options that will be used by Content Preview

### DIFF
--- a/src/elements/content-preview/ContentPreview.js
+++ b/src/elements/content-preview/ContentPreview.js
@@ -56,6 +56,7 @@ import {
     ERROR_CODE_UNKNOWN,
 } from '../../constants';
 import type { Annotation } from '../../common/types/feed';
+import type { TargetingApi } from '../../features/targeting/types';
 import type { ErrorType, AdditionalVersionInfo } from '../common/flowTypes';
 import type { WithLoggerProps } from '../../common/types/logging';
 import type { RequestOptions, ErrorContextProps, ElementsXhrError } from '../../common/types/api';
@@ -103,6 +104,9 @@ type Props = {
     onLoad: Function,
     onNavigate: Function,
     onVersionChange: VersionChangeCallback,
+    previewExperiences?: {
+        [name: string]: TargetingApi,
+    },
     previewLibraryVersion: string,
     requestInterceptor?: Function,
     responseInterceptor?: Function,
@@ -746,6 +750,7 @@ class ContentPreview extends React.PureComponent<Props, State> {
             fileOptions,
             onAnnotatorEvent,
             onAnnotator,
+            previewExperiences,
             showAnnotationsControls,
             token: tokenOrTokenFunction,
             ...rest
@@ -784,6 +789,7 @@ class ContentPreview extends React.PureComponent<Props, State> {
             fileOptions: fileOpts,
             header: 'none',
             headerElement: `#${this.id} .bcpr-PreviewHeader`,
+            experiences: previewExperiences,
             showAnnotations: this.canViewAnnotations(),
             showAnnotationsControls,
             showDownload: this.canDownload(),


src/components/tooltip/Tooltip.scss
```diff
@@ -64,6 +64,11 @@ $tooltip-offset: 4px; /* amount of space between target and tooltip arrow */
         }
     }
 
+    &.is-wide {
```
jstoffan, line 67: Is this change related to the preview changes or should it be a separate PR? We will need to very cautious about (or even avoid entirely) using BUIE components in Preview SDK, since it can lead to styling issues.

src/components/tooltip/Tooltip.scss
```diff
@@ -88,6 +93,23 @@ $tooltip-offset: 4px; /* amount of space between target and tooltip arrow */
         }
     }
 
+    .tooltip-callout {
+        display: flex;
+        flex-flow: column;
+        font-weight: normal;
+        font-size: 13px;
+        line-height: 1.54;
```
jstoffan, line 101: Can we use a more standardized value to avoid this magic number?

src/components/tooltip/Tooltip.scss
```diff
@@ -88,6 +93,23 @@ $tooltip-offset: 4px; /* amount of space between target and tooltip arrow */
...
+        letter-spacing: .3px;
```
jstoffan, line 102: Is this needed or can we use the default kerning?

src/components/tooltip/Tooltip.scss
```diff
@@ -88,6 +93,23 @@ $tooltip-offset: 4px; /* amount of space between target and tooltip arrow */
...
+        cursor: default;
```
jstoffan, line 103: Is this needed?

src/elements/content-preview/ContentPreview.js
```diff
@@ -1274,6 +1283,7 @@ class ContentPreview extends React.PureComponent<Props, State> {
                             onDownload={this.download}
                             contentOpenWithProps={contentOpenWithProps}
                             canAnnotate={this.canAnnotate()}
+                            previewExperiences={previewExperiences}
```
jstoffan, line 1286: This version of the preview header is only shown to platform customers, so this may not be needed.

megansmith-box: yes, I will remove, thanks!

src/components/tooltip/Tooltip.scss
megansmith-box: yes, it's not needed, thank you

src/components/tooltip/Tooltip.scss
megansmith-box: Yes, the cursor changes to select text cursor versus pointer cursor without this.

src/components/tooltip/Tooltip.scss
megansmith-box: I copied the styles from left-nav-callout css style. I'm thinking at some point we can replace left-nav-callout with this more generic class tooltip-callout. If I remove it it looks more condensed than the given style. And If i increase it, it looks more spread apart. I think I will keep it as close to design as possible and consistent with the left-nav-callout unless you think it's super important and should be changed

src/elements/content-preview/ContentPreview.js
```diff
@@ -784,6 +789,7 @@ class ContentPreview extends React.PureComponent<Props, State> {
             fileOptions: fileOpts,
             header: 'none',
             headerElement: `#${this.id} .bcpr-PreviewHeader`,
+            previewExperiences,
```
jstoffan, line 792: Minor, but since we're passing this value to preview, should we go with just `experiences` here?

megansmith-box: yes, I like that idea

jstoffan (approved):

src/components/tooltip/Tooltip.scss
alexkrolick: 1.54 * 13px = 20px, which was probably meant to be the height of the element

PR #2543, opened 2021-04-21
fix(content-picker): Add spacing between user and size tags

sahiga:
In the Content Picker, the modified-by user tag and the item size tag sit right next to each other, without any spacing. This PR fixes this bug by reusing styles from the Content Explorer in the Content Picker. There are no changes to the Content Explorer.

**Before:**
content-picker-size-spacing-before

**After:**
content-picker-size-spacing-after

### DIFF
--- a/src/elements/common/item/ItemSubDetails.scss
+++ b/src/elements/common/item/ItemSubDetails.scss
@@ -8,7 +8,8 @@
     }
 }
 
-.bce-item-row {
+.bce-item-row,
+.bcp-item-row {
     .bdl-ItemSubDetails-size {
         &::before {
             padding-right: 3px;


jstoffan (approved):

src/elements/common/item/ItemSubDetails.scss
```diff
@@ -8,7 +8,8 @@
     }
 }
 
-.bce-item-row {
+.bce-item-row,
+.bcp-item-row {
```
ConradJChan, line 12: Is this dependent on `.bce-item-row` at all? Could the extra specificity be dropped?

ConradJChan (approved):

src/elements/common/item/ItemSubDetails.scss
sahiga: @ConradJChan Yup, without the extra specificity, the grid view also displays a middot:
content-explorer-with-middot-spacing

ConradJChan: Ah missed the duality of grid vs list, thanks

swfree (approved):

PR #2544, opened 2021-04-22
fix(a11y): ensure tooltip-labelled content is accessible

tille:
The problem is the tooltip element is not in the DOM unless focused, so use cases like icons that are labeled visually with tooltips don’t have proper screenreader representation because they are defacto invisible in the screen reader tree

so we want to auto-apply the aria-label whenever there is a tooltip, unless the user has already applied an aria label to the target element, in which case the describedby relationship is applied - this is for the use case where you have an icon with obvious visual meaning + a manual aria label as a substitute, and a longer explanatory text in the tooltip, which only needs to show on focus in either the visual or screen reader mode

what we found is that a whole bunch of places where we placed tooltips on icons had to have manual aria-labels added that were equal to the tooltip text, and then the current implementation would also add the describedby relation on focus, which was redundant because the text was the same (this strategy takes care of this situation)

### DIFF
--- a/src/components/guide-tooltip/GuideTooltip.tsx
+++ b/src/components/guide-tooltip/GuideTooltip.tsx
@@ -2,8 +2,7 @@ import * as React from 'react';
 import { FormattedMessage } from 'react-intl';
 import classNames from 'classnames';
 
-// @ts-ignore flow import
-import Tooltip, { TooltipTheme, TooltipBase } from '../tooltip';
+import Tooltip, { TooltipTheme } from '../tooltip';
 import Button from '../button';
 // @ts-ignore flow import
 import messages from './messages';
--- a/src/components/guide-tooltip/GuideTooltip.tsx
+++ b/src/components/guide-tooltip/GuideTooltip.tsx
@@ -3,14 +3,14 @@ import { FormattedMessage } from 'react-intl';
 import classNames from 'classnames';
 
 // @ts-ignore flow import
-import Tooltip, { TooltipProps as TooltipBase, TooltipTheme } from '../tooltip';
+import Tooltip, { TooltipTheme, TooltipBase } from '../tooltip';
 import Button from '../button';
 // @ts-ignore flow import
 import messages from './messages';
 
 import './GuideTooltip.scss';
 
-type TooltipProps = Omit<TooltipBase, 'intl' | 'text' | 'theme'>;
+type TooltipProps = Omit<JSX.LibraryManagedAttributes<typeof Tooltip, TooltipBase['props']>, 'text' | 'theme'>;
 type Props = TooltipProps & {
     body: React.ReactNode;
     title?: React.ReactNode;
--- a/src/components/guide-tooltip/GuideTooltip.tsx
+++ b/src/components/guide-tooltip/GuideTooltip.tsx
@@ -2,15 +2,15 @@ import * as React from 'react';
 import { FormattedMessage } from 'react-intl';
 import classNames from 'classnames';
 
-import Tooltip, { TooltipTheme } from '../tooltip';
+// @ts-ignore flow import
+import Tooltip, { TooltipProps as TooltipBase, TooltipTheme } from '../tooltip';
 import Button from '../button';
 // @ts-ignore flow import
 import messages from './messages';
 
 import './GuideTooltip.scss';
 
-type TooltipProps = Omit<JSX.LibraryManagedAttributes<typeof Tooltip, Tooltip['props']>, 'text' | 'theme'>;
-
+type TooltipProps = Omit<TooltipBase, 'intl' | 'text' | 'theme'>;
 type Props = TooltipProps & {
     body: React.ReactNode;
     title?: React.ReactNode;
--- a/src/components/guide-tooltip/GuideTooltip.tsx
+++ b/src/components/guide-tooltip/GuideTooltip.tsx
@@ -9,7 +9,8 @@ import messages from './messages';
 
 import './GuideTooltip.scss';
 
-type TooltipProps = Omit<JSX.LibraryManagedAttributes<typeof Tooltip, TooltipBase['props']>, 'text' | 'theme'>;
+type TooltipProps = Omit<JSX.LibraryManagedAttributes<typeof Tooltip, Tooltip['props']>, 'text' | 'theme'>;
+
 type Props = TooltipProps & {
     body: React.ReactNode;
     title?: React.ReactNode;
--- a/src/components/image-tooltip/__tests__/ImageTooltip.test.tsx
+++ b/src/components/image-tooltip/__tests__/ImageTooltip.test.tsx
@@ -18,7 +18,7 @@ describe('components/image-tooltip/ImageTooltip', () => {
         const image = <img src={testImageSrc} alt="foo" />;
 
         const wrapper: ShallowWrapper = shallow(
-            <ImageTooltip intl={intlFake} content="Foo content" image={image} isShown title="Bar">
+            <ImageTooltip content="Foo content" image={image} isShown title="Bar">
                 <Button>Callout</Button>
             </ImageTooltip>,
         );
--- a/src/components/image-tooltip/__tests__/ImageTooltip.test.tsx
+++ b/src/components/image-tooltip/__tests__/ImageTooltip.test.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { shallow, ShallowWrapper } from 'enzyme';
+import { MessageDescriptor } from 'react-intl';
 
 import Button from '../../button/Button';
 import ImageTooltip from '../ImageTooltip';
@@ -7,11 +8,16 @@ import ImageTooltip from '../ImageTooltip';
 import testImageSrc from '../getTestImageSrc';
 
 describe('components/image-tooltip/ImageTooltip', () => {
+    const intlFake: any = {
+        formatMessage: (message: MessageDescriptor) => message.defaultMessage,
+        locale: 'en-US',
+    };
+
     test('should correctly render an ImageTooltip', () => {
         const image = <img src={testImageSrc} alt="foo" />;
 
         const wrapper: ShallowWrapper = shallow(
-            <ImageTooltip content="Foo content" image={image} isShown title="Bar">
+            <ImageTooltip intl={intlFake} content="Foo content" image={image} isShown title="Bar">
                 <Button>Callout</Button>
             </ImageTooltip>,
         );
--- a/src/components/image-tooltip/__tests__/ImageTooltip.test.tsx
+++ b/src/components/image-tooltip/__tests__/ImageTooltip.test.tsx
@@ -8,12 +8,6 @@ import ImageTooltip from '../ImageTooltip';
 import testImageSrc from '../getTestImageSrc';
 
 describe('components/image-tooltip/ImageTooltip', () => {
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const intlFake: any = {
-        formatMessage: (message: MessageDescriptor) => message.defaultMessage,
-        locale: 'en-US',
-    };
-
     test('should correctly render an ImageTooltip', () => {
         const image = <img src={testImageSrc} alt="foo" />;
 
--- a/src/components/image-tooltip/__tests__/ImageTooltip.test.tsx
+++ b/src/components/image-tooltip/__tests__/ImageTooltip.test.tsx
@@ -8,6 +8,7 @@ import ImageTooltip from '../ImageTooltip';
 import testImageSrc from '../getTestImageSrc';
 
 describe('components/image-tooltip/ImageTooltip', () => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const intlFake: any = {
         formatMessage: (message: MessageDescriptor) => message.defaultMessage,
         locale: 'en-US',
--- a/src/components/image-tooltip/__tests__/ImageTooltip.test.tsx
+++ b/src/components/image-tooltip/__tests__/ImageTooltip.test.tsx
@@ -1,6 +1,5 @@
 import React from 'react';
 import { shallow, ShallowWrapper } from 'enzyme';
-import { MessageDescriptor } from 'react-intl';
 
 import Button from '../../button/Button';
 import ImageTooltip from '../ImageTooltip';
--- a/src/components/text-input/__tests__/__snapshots__/TextInput.test.js.snap
+++ b/src/components/text-input/__tests__/__snapshots__/TextInput.test.js.snap
@@ -14,9 +14,6 @@ exports[`components/text-input/TextInput should render text input with descripti
       some description
     </div>
     <Tooltip
-      constrainToScrollParent={false}
-      constrainToWindow={true}
-      isDisabled={false}
       isShown={false}
       position="middle-right"
       text=""
--- a/src/components/text-input/__tests__/__snapshots__/TextInput.test.js.snap
+++ b/src/components/text-input/__tests__/__snapshots__/TextInput.test.js.snap
@@ -14,6 +14,9 @@ exports[`components/text-input/TextInput should render text input with descripti
       some description
     </div>
     <Tooltip
+      constrainToScrollParent={false}
+      constrainToWindow={true}
+      isDisabled={false}
       isShown={false}
       position="middle-right"
       text=""
--- a/src/components/tooltip/Tooltip.tsx
+++ b/src/components/tooltip/Tooltip.tsx
@@ -204,6 +204,17 @@ class Tooltip extends React.Component<TooltipProps, State> {
         return showTooltip;
     };
 
+    tooltipText = this.props.text
+        ? typeof this.props.text === 'object'
+            ? this.props.text.props.defaultMessage
+            : String(this.props.text)
+        : '';
+
+    labelInName = () => {
+        const ariaLabel = React.Children.only(this.props.children).props['aria-label'];
+        return ariaLabel && this.tooltipText !== ariaLabel;
+    };
+
     render() {
         const {
             bodyElement,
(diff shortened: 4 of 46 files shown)


src/components/tooltip/Tooltip.tsx
```diff
@@ -219,6 +219,12 @@ class Tooltip extends React.Component<TooltipProps, State> {
             theme,
         } = this.props;
 
+        const tooltipText =
+            text && typeof text === 'object' ? (text as React.ReactElement).props.defaultMessage : String(text);
```
laughnan, line 223: Instead of accessing `props.defaultMessage` (assuming this test is a `react-intl` MessageDescriptor) you can use `intl.formatMessage` to render the message as a string.

src/components/tooltip/Tooltip.tsx
```diff
@@ -219,6 +219,12 @@ class Tooltip extends React.Component<TooltipProps, State> {
...
+
+        const ariaLabel = (React.Children.only(children) as React.ReactElement).props['aria-label'];
+        const labelInName = ariaLabel && tooltipText !== ariaLabel;
```
laughnan, line 226: `[nit]` maybe making the variable `isLabelMatchingTooltipText`

src/components/tooltip/Tooltip.tsx
```diff
@@ -219,6 +219,12 @@ class Tooltip extends React.Component<TooltipProps, State> {
             theme,
         } = this.props;
 
+        const tooltipText =
+            text && typeof text === 'object' ? (text as React.ReactElement).props.defaultMessage : String(text);
+
+        const ariaLabel = (React.Children.only(children) as React.ReactElement).props['aria-label'];
```
alexkrolick, line 225: Rather than casting this, can you make sure `.props` is defined, so there isn't the possibility runtime error if the tooltip is wrapped around a string or other non-Element node?

src/components/tooltip/Tooltip.tsx
laughnan: +1

src/components/tooltip/Tooltip.tsx
```diff
@@ -217,10 +218,17 @@ class Tooltip extends React.Component<TooltipProps, State> {
             stopBubble,
             text,
             theme,
+            intl,
         } = this.props;
 
-        const tooltipText =
-            text && typeof text === 'object' ? (text as React.ReactElement).props.defaultMessage : String(text);
+        const textElement = text as React.ReactElement;
```
alexkrolick, line 224: ReactElement is not the same as ReactNode, so this cast is masking the real possible types for the `text` prop

src/components/tooltip/Tooltip.tsx
```diff
@@ -217,8 +219,24 @@ class Tooltip extends React.Component<TooltipProps, State> {
...
+            intl,
         } = this.props;
 
+        let ariaLabel;
+        let tooltipText;
+
+        if (text) {
+            if (typeof text === 'string') {
+                tooltipText = text;
+            } else if (React.isValidElement(text) && getProp(text, 'props.id')) {
```
alexkrolick, line 231: Can you check for the element type being FormattedMessage (because anything can have an `id`), and pull out the exact react-intl props here and when you pass them into formatMessage

```js
const { id, defaultMessage, description, values } = props;
intl.formatMessage({ id, defaultMessage, description }, values)
```

alexkrolick: cc @megansmith-box @DanDeMicco 👀 can you verify if this approach conflicts with any of the ways product adoption tooltips work? Do we put rich text formatting in there?

tille: @alexkrolick do you mean something like this?
```
      if (text) {
          if (typeof text === 'string') {
              tooltipText = text;
          } else if (React.isValidElement(text) && getProp(text, 'props')) {
              const { id, defaultMessage, description, values } = getProp(text, 'props');
              if (id && defaultMessage && description) {
                  tooltipText = intl.formatMessage({ id, defaultMessage, description }, values)
              }
          }
      }
```

alexkrolick: Yes, like that

src/components/tooltip/Tooltip.tsx
```diff
@@ -217,8 +219,27 @@ class Tooltip extends React.Component<TooltipProps, State> {
...
+                tooltipText = text;
+            } else if (React.isValidElement(text) && getProp(text, 'props')) {
+                const { id, defaultMessage, description, values } = getProp(text, 'props');
+                if (id && defaultMessage) {
+                    tooltipText = intl.formatMessage({ id, defaultMessage, description }, values);
+                }
+            }
+        }
+        if (getProp(children, 'props')) {
+            ariaLabel = getProp(children, 'props.aria-label');
```
megansmith-box, line 239: will aria-label always be a prop on children? should we have a default in case it's not or does getProp return a default?

src/components/tooltip/Tooltip.tsx
```diff
@@ -217,8 +219,27 @@ class Tooltip extends React.Component<TooltipProps, State> {
...
+            intl,
         } = this.props;
 
+        let ariaLabel;
+        let tooltipText;
+
+        if (text) {
+            if (typeof text === 'string') {
+                tooltipText = text;
+            } else if (React.isValidElement(text) && getProp(text, 'props')) {
```
DanDeMicco, line 231: What is expected behavior for tooltips with not just a single element passed in? e.g. ones with a title and a body?

`GuideTooltip` is an example of this, but there are multiple variations of this in the codebase.

src/components/tooltip/Tooltip.tsx
```diff
@@ -217,8 +219,27 @@ class Tooltip extends React.Component<TooltipProps, State> {
             stopBubble,
             text,
             theme,
+            intl,
         } = this.props;
 
+        let ariaLabel;
```
megansmith-box, line 225: is undefined a valid ariaLabel?

tille: yes it is a valid one, because if is undefined then it will be captured by the line
https://github.com/box/box-ui-elements/pull/2544/files#diff-ecdd760eb0b8901c98cc10e0b0266af590216f4f4619845564b2755d1ef90c55R241

src/components/tooltip/Tooltip.tsx
tille: yes exactly

src/components/tooltip/Tooltip.tsx
tille: Pretty God question

src/components/tooltip/Tooltip.tsx
laughnan: Is there a benefit in renaming `ariaLabel` to `childComponentAriaLabel` to keep it clear that it's not the tooltip's `aria-label`?

src/components/tooltip/Tooltip.tsx
laughnan: This logic should work fine for `CheckboxTooltip`  and `ImageTooltip`. The `GuideTooltip` is tricky because it builds a pretty complex child component in the `text` prop. It might make sense to address this in a follow-up. Theoretically, the logic above shouldn't impact (good or bad) it.

src/components/tooltip/Tooltip.tsx
```diff
@@ -217,8 +219,27 @@ class Tooltip extends React.Component<TooltipProps, State> {
...
+        let tooltipText;
+
+        if (text) {
+            if (typeof text === 'string') {
+                tooltipText = text;
+            } else if (React.isValidElement(text) && getProp(text, 'props')) {
+                const { id, defaultMessage, description, values } = getProp(text, 'props');
```
laughnan, line 232: I think this should be fine for the first pass (still wrapping my head around updating the child's `aria-label` or adding in a `describedby` label. Can we add comments in this conditional to explain that we're specifically looking for these props since they indicate a `FormattedMessage` component is passed into `text` so that it's clear?

src/components/tooltip/Tooltip.tsx
tille: Where can I find a real example of those kind of Tooltips? (ex. GuideTooltip)

```
For those Tooltips I guess it should have a `aria-describedBy` relation and `aria-hidden=false`
this strategy takes care of that (but Im needing a real example to test it)
```

src/components/tooltip/Tooltip.tsx
tille: yes!

src/components/tooltip/Tooltip.tsx
laughnan: Shared via Slack.

src/components/image-tooltip/__tests__/ImageTooltip.test.tsx
```diff
@@ -1,17 +1,23 @@
...
 import { shallow, ShallowWrapper } from 'enzyme';
+import { MessageDescriptor } from 'react-intl';
 
 import Button from '../../button/Button';
 import ImageTooltip from '../ImageTooltip';
 // @ts-ignore flow import
 import testImageSrc from '../getTestImageSrc';
 
 describe('components/image-tooltip/ImageTooltip', () => {
+    const intlFake: any = {
```
laughnan, line 11: I'm seeing these linting errors:

```
/Users/alaughnan/Code/box-ui-elements/src/components/image-tooltip/ImageTooltip.stories.tsx
  11:17  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any

/Users/alaughnan/Code/box-ui-elements/src/components/image-tooltip/__tests__/ImageTooltip.test.tsx
  11:21  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any

/Users/alaughnan/Code/box-ui-elements/src/components/tooltip/__tests__/Tooltip.test.tsx
  13:21  warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
```

laughnan (commented):
I think to fix the issue in `GuideTooltip` you can change it's type definition slightly:

```
...
import Tooltip, { TooltipProps as BaseTooltipProps, TooltipTheme } from '../tooltip';
...
type TooltipProps = Omit<BaseTooltipProps, 'intl' | 'text' | 'theme'>;
```

Basically replacing `JSX.LibraryManagedAttributes<typeof Tooltip, Tooltip['props']>` with `BaseTooltipProps` (added to import statement). And I think you actually want type `TooltipProps = Omit<BaseTooltipProps, 'intl' | 'text' | 'theme'>;` (I added intl to the Omit)

tille:
> I think to fix the issue in `GuideTooltip` you can change it's type definition slightly:
> 
> ```
> ...
> import Tooltip, { TooltipProps as BaseTooltipProps, TooltipTheme } from '../tooltip';
> ...
> type TooltipProps = Omit<BaseTooltipProps, 'intl' | 'text' | 'theme'>;
> ```
> 
> Basically replacing `JSX.LibraryManagedAttributes<typeof Tooltip, Tooltip['props']>` with `BaseTooltipProps` (added to import statement). And I think you actually want type `TooltipProps = Omit<BaseTooltipProps, 'intl' | 'text' | 'theme'>;` (I added intl to the Omit)

done!

src/components/image-tooltip/__tests__/ImageTooltip.test.tsx
tille: those are warnings ^^

src/components/guide-tooltip/GuideTooltip.tsx
```diff
@@ -2,15 +2,15 @@ import * as React from 'react';
 import { FormattedMessage } from 'react-intl';
 import classNames from 'classnames';
 
-import Tooltip, { TooltipTheme } from '../tooltip';
+// @ts-ignore flow import
```
laughnan, line 5: Do we need this?

tille: yes it needed for the cause `TooltipProps` is not passing ts linting
it has to be ignored in flow to pass all linting errors

src/features/collapsible-sidebar/__tests__/__snapshots__/CollapsibleSidebarLogo.test.js.snap
```diff
@@ -246,67 +246,80 @@ exports[`components/core/collapsible-sidebar/CollapsibleSidebar render 2`] = `
...
           isDisabled={true}
           isTabbable={false}
           position="middle-right"
-          theme="default"
         >
-          <CollapsibleSidebarLogo__StyledToggleButton
-            aria-label="Expand"
-            className="bdl-CollapsibleSidebar-toggleButton"
-            type="button"
+          <Tooltip
```
laughnan, line 253: Why does this render two `Tooltips`?

swfree: This snapshot is generated with `mount` instead of `shallow`. Same thing happened with the [SidebarToggleButton snapshot test](https://github.com/box/box-ui-elements/blob/master/src/components/sidebar-toggle-button/__tests__/SidebarToggleButton.test.js#L7)

swfree (approved):

laughnan (changes requested):
After internal discussions, here are the changes to get this merged:

- Revert larger pull request changes and just check `text` prop as a string (no longer checking FormattedMessage, etc)
- Don't add `aria-describedby` when the `aria-label` is the same

src/components/guide-tooltip/GuideTooltip.tsx
alexkrolick: This file is in TS, and the import is in TS, so it shouldn't need an ignore

src/components/text-input/__tests__/__snapshots__/TextInput.test.js.snap
```diff
@@ -14,9 +14,6 @@ exports[`components/text-input/TextInput should render text input with descripti
       some description
     </div>
     <Tooltip
-      constrainToScrollParent={false}
```
alexkrolick, line 17: What's causing these default props to be dropped from the snapshots?

tille: the InjectIntl wrapped, just remove it and updated snapshots

src/components/tooltip/__tests__/__snapshots__/Tooltip.test.tsx.snap
```diff
@@ -3,7 +3,7 @@
 exports[`components/tooltip/Tooltip render() should match snapshot when stopBubble is set 1`] = `
 <div
   className="tooltip bdl-Tooltip"
-  id="tooltip19"
+  id="tooltip22"
```
jstoffan, line 6: Is this change needed or can we revert it?

src/components/tooltip/__tests__/Tooltip.test.tsx
```diff
@@ -109,7 +109,7 @@ describe('components/tooltip/Tooltip', () => {
         test('should show tooltip when isShown state is true', () => {
             const wrapper = shallow(
                 <Tooltip text="hi">
-                    <button />
+                    <button aria-label="tales" />
```
jstoffan, line 112: Do we have a condition that tests a button with no aria label?

src/components/tooltip/Tooltip.tsx
```diff
@@ -325,12 +340,18 @@ class Tooltip extends React.Component<TooltipProps, State> {
...
                 </div>
             </div>
         ) : (
-            <div className={classes} id={this.tooltipID} role="tooltip" aria-live="polite">
+            <div
+                className={classes}
+                id={this.tooltipID}
+                role="tooltip"
+                aria-live="polite"
+                aria-hidden={isLabelMatchingTooltipText}
```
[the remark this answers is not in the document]
tille: this is to avoid repetition of texts being announced so if BOTH are the same means that `isLabelMatchingTooltipText` will be true and so aria-label and tooltipText doesn't have to be read with ONLY one of them being read is enough

src/components/tooltip/__tests__/__snapshots__/Tooltip.test.tsx.snap
tille: don't know why its being updated but anyway it doesn't make any side effect 
the ID can be 19 or 22 and doesn't generate any problem

src/components/tooltip/Tooltip.tsx
```diff
@@ -222,6 +223,9 @@ class Tooltip extends React.Component<TooltipProps, State> {
             theme,
         } = this.props;
 
+        const childAriaLabel = getProp(children, 'props.aria-label');
+        const isLabelMatchingTooltipText = Boolean(childAriaLabel && childAriaLabel === text);
```
jstoffan, line 227: Can we go with `const isLabelMatchingTooltipText = !!childAriaLabel && childAriaLabel === text;`?

src/components/tooltip/Tooltip.tsx
```diff
@@ -254,7 +258,10 @@ class Tooltip extends React.Component<TooltipProps, State> {
         }
 
         if (showTooltip) {
-            componentProps['aria-describedby'] = this.tooltipID;
+            if (!isLabelMatchingTooltipText || childAriaLabel === undefined) {
```
jstoffan, line 261: Do we have any unit tests for this change?

tille: yes `when aria-label does not exists describedBy relation should exists` 
https://github.com/box/box-ui-elements/blob/30f5e4d5d533be5a7de20804cc705d4ae818c5b8/src/components/tooltip/__tests__/Tooltip.test.tsx#L122

src/components/tooltip/Tooltip.tsx
tille: hmm I think... yes

src/components/tooltip/Tooltip.tsx
jstoffan: It looks like that's testing the positive case. Do we have any assertions for the negative case?

tille: https://github.com/box/box-ui-elements/pull/2544/commits/31c7a92b5882470b4fd730fd35ffd6f8f5a28b36

jstoffan (dismissed):
Looks good to me!

laughnan (approved):
🎉

jstoffan (approved):

PR #2545, opened 2021-04-22
fix(content-sharing): allow download settings on gsuite files

sophiayangyxy:


ctuckz (commented):
Change looks good to me. This needs to be merged after the backend changes are live, right @sophiayangyxy?

sahiga (approved):

PR #2546, opened 2021-04-22
feat(activity-sidebar): allow callback for task assignment update

yetesh:


src/api/Feed.js
```diff
@@ -504,6 +505,19 @@ class Feed extends Base {
             },
             errorCallback: handleError,
         });
+
+        const params = queryString.parse(location.search);
```
jstoffan, line 509: Can we retrieve this information in a component and pass it to this call? Doing so in the data access layer isn't ideal.

src/api/tasks/TaskCollaborators.js
```diff
@@ -15,6 +15,8 @@ import {
 } from '../../constants';
 import type { ElementsErrorCallback } from '../../common/types/api';
 import type { BoxItem } from '../../common/types/core';
+import Xhr from '../utils/Xhr';
```
jstoffan, line 18: Can we use the underlying `post` method on the API classes to avoid interacting with the XHR class directly?

src/api/tasks/TaskCollaborators.js
```diff
@@ -149,6 +151,50 @@ class TaskCollaborators extends TasksBase {
...
+        emailTraceId,
+        successCallback,
+        errorCallback
+    }): {
+
+        const url = getUrlForNotificationAnalytics();
+        const requestData = {
+            data: {           
+              notification_type: "workflow-task-" + taskStatus.toLowerCase(),
+              source: 'WEBAPP',
```
jstoffan, line 172: It seems like this logic should probably live in the webapp, since this component can be used in third-party contexts, as well.

src/api/tasks/TaskCollaborators.js
```diff
@@ -149,6 +151,50 @@ class TaskCollaborators extends TasksBase {
         });
     }
 
+    getUrlForNotificationAnalytics(): string {
```
jstoffan, line 154: Should we move portions of this logic to a separate API class? it seems to be a new resource not directly related to tasks or collaborators.

jstoffan (commented):
Hi @yetesh. Let's sync offline about some of these changes.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -317,6 +317,10 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
 
         // need to load the pending item
         this.fetchFeedItems();
+
+        if (onTaskAssignmentUpdate) {
```
jstoffan, line 321: Should we assign a `defaultProps` value of `noop` for this prop, similar to `onTaskUpdate`?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -317,6 +317,10 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+            onTaskAssignmentUpdate(taskId, taskAssignmentId, status);
```
jstoffan, line 322: It looks like we all onTaskUpdate in the success callback of the updateTask method. Can we stay consistent with that here, as well?

jstoffan (commented):
Looks good to me. Should we add a unit test?

src/elements/content-sidebar/ActivitySidebar.js
yetesh: Seems, there's already a default prop of `noop` :-) 
https://github.com/box/box-ui-elements/blob/master/src/elements/content-sidebar/ActivitySidebar.js#L116

src/elements/content-sidebar/ActivitySidebar.js
yetesh: Updated, thanks for the feedback!

yetesh:
@jstoffan Since, its part of successCallback which needs API call and I think unit testing might be hard

jstoffan:
@yetesh, we typically ask for unit tests for all changes to help ensure code coverage. Were you able to find any examples of existing tests that run for similar functionality?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -304,14 +304,23 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
     };
 
     updateTaskAssignment = (taskId: string, taskAssignmentId: string, status: TaskCollabStatus): void => {
-        const { file, api } = this.props;
+        const { file, api, onTaskAssignmentUpdate } = this.props;
+        // $FlowFixMe
```
jstoffan, line 308: Can we add more information about why we're suppressing flow here? It seems risky to override the typing system when it comes to state, since if `currentUser` is undefined, this could throw an error.

yetesh: Can you take a look again ?

jstoffan (approved):

yetesh:
Hi @jstoffan Seems I don't have write access, would you be able to merge this ?

yetesh:
oops, I can add `ready-to-merge` label

PR #2548, opened 2021-04-26
fix(AccountSettings): A11Y - update label color in 2-step verification modal

fsuarez-globant:
fix(AccountSettings): A11Y - update label color in 2-step verification modal

I did this change in order to improve color contrast in the 2-Step modal (Accounts Settings) , I changed the label font color from `$bdl-gray-50` to `$bdl-gray-62` and I also added an aria-label attribute to the radio button component.

Screenshot before change 
image

Screenshot after change
image


src/components/radio/RadioButton.tsx
```diff
@@ -31,6 +31,7 @@ const RadioButton = ({
         {/* eslint-disable-next-line jsx-a11y/label-has-for */}
         <label className="radio-label">
             <input
+                aria-label={name}
```
laughnan, line 34: Does the `label` element here not sufficiently handle the child `input`? Do we just need to add a `for` or something to connect them?

fsuarez-globant: @laughnan I just added `htmlfor` to connect the input and label as you recommended.

laughnan (approved):

swfree (approved):

PR #2559, opened 2021-05-10
fix(flyouts): made flyout button keyboard accessible

msorianoBox:
Hey guys! The Flyouts button can't be used while navigating on keyboard, so we had to make some adjustments in order to be  focusable and activated with the Enter key.

Screen Shot 2021-05-06 at 09 18 22

### DIFF
--- a/src/components/flyout/Flyout.js
+++ b/src/components/flyout/Flyout.js
@@ -290,8 +290,8 @@ class Flyout extends React.Component<Props, State> {
         }
     };
 
-    handleKeyPress = event => {
-        if (event.key === 'Enter') {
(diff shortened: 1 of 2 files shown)


src/components/flyout/Flyout.js
```diff
@@ -291,6 +290,13 @@ class Flyout extends React.Component<Props, State> {
         }
     };
 
+    handleKeyPress = (e: KeyboardEvent) => {
+        if (e.key === 'Enter') {
```
laughnan, line 294: We have a constant for `Enter` in our `KEYS` object: `src/constants.js`

laughnan (approved):
Thanks for the edits, this makes sense.

PR #2560, opened 2021-05-10
feat(tooltip): add classname to tether element component

megansmith-box:
- Add the ability to add a specific class to the tether element. We don't necessarily want to target all .tooltip-element classes. My use case changing z-index:
before:
Screen Shot 2021-05-10 at 11 46 30 AM
after (with class):
Screen Shot 2021-05-10 at 11 44 41 AM

### DIFF
--- a/src/components/tooltip/Tooltip.js.flow
+++ b/src/components/tooltip/Tooltip.js.flow
@@ -47,6 +47,8 @@ type Props = {
     showCloseButton?: boolean,
     /** stop click|keypress event bubbling */
     stopBubble?: boolean,
+    /** A CSS class for the tether element component */
+    tetherElementClassName?: string,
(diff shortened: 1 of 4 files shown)


DanDeMicco (approved):

megansmith-box:
Hi @alexkrolick - The z-index is currently placed on the class `.tooltip-element` and the className prop to the tooltip container is placed on the child component of `.tooltip-element.` There is no supported way that I could find to select a parent element using css. I wanted to figure out a way to target the `.tooltip-element` element using css without changing all instances of `.tooltip-element`. I think this use case will be useful in the future, so I added this new `tetherElementClassName` prop

Screen Shot 2021-05-12 at 8 52 45 AM

alexkrolick (approved):

PR #2563, opened 2021-05-13
fix(button-group): fix styles when wrapped with tooltip

megansmith-box:
We were seeing a gap when button wrapped in tooltip for button group
Screen Shot 2021-05-12 at 6 14 46 PM
- now we'd like to target not just the direct .btn child but also the child .btn of a tooltip (.bdl-targeted-click-through)
Screen Shot 2021-05-12 at 4 08 23 PM

### DIFF
--- a/src/components/button-group/ButtonGroup.scss
+++ b/src/components/button-group/ButtonGroup.scss
@@ -4,8 +4,6 @@
  * Button Group
  **************************************/
 
-// In the case where there is an element in between .btn and .bdl-ButtonGroup, you can add class .bdl-ButtonWrapper to parent of .btn
-
 .bdl-ButtonGroup {
     position: relative;
 
@@ -13,7 +11,7 @@
         display: inline;
 
         > .btn,
-        > .bdl-ButtonWrapper > .btn {
+        > .bdl-targeted-click-through > .btn {
             margin-left: 0;
             padding-right: 11px;
             padding-left: 11px;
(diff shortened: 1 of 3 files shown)


src/components/button-group/ButtonGroup.scss
```diff
@@ -9,7 +9,8 @@
     .toggle-overlay {
         display: inline;
 
-        > .btn {
+        > .btn,
```
alexkrolick, line 12: What if we remove the direct descendent combinator (`>`)?

megansmith-box: That works too, I wasn't sure which is better to be more specific or to allow for any child .btn. What do you think @alexkrolick ?

alexkrolick: On the one hand, it's more conservative to be specific with the selector, especially if we add support for nested button groups and things like that.

But I don't think we want to bake in specific external classnames into the implementation. What if we add a generic `.bdl-ButtonWrapper` class target here and you can add that to your targeted click through implementation?

megansmith-box: I like that idea! Thank you @alexkrolick

alexkrolick (dismissed):

src/components/button-group/ButtonGroup.scss
alexkrolick: Should probably add a comment in the SCSS file or Storybook about this

src/features/targeting/hocs/withTargetedClickThrough.js
```diff
@@ -41,7 +41,7 @@ function withTargetedClickThrough<Config>(
         return (
             <WrappedComponent showCloseButton stopBubble {...rest} isShown={shouldShow} onDismiss={onClose}>
                 <span
-                    className="bdl-targeted-click-through"
+                    className="bdl-targeted-click-through bdl-ButtonWrapper"
```
megansmith-box, line 44: Hi @alexkrolick - I realized the element I want to add the class to is in BUIE. How do you feel about adding the class directly here? Or do you think it's better to pass it as a prop? I think we would want all cases of this element to have  .bdl-ButtonWrapper so I think adding it directly makes sense. What do you think?

alexkrolick: Makes sense to add it here I think

alexkrolick: * Assuming that it always or nearly always wraps a button

jstoffan: It seems a little unsafe to make such a big assumption, since the tooltip could wrap almost any kind of element. Given that, should we revert back to the original solution, with `.bdl-targeted-click-through > .btn` in the ButtonGroup styles?

src/components/button-group/ButtonGroup.scss
jstoffan: It seems like including `.bdl-ButtonWrapper` here is largely the same as including `.bdl-targeted-click-through` here, since both are external to the ButtonGroup component.

src/features/targeting/hocs/withTargetedClickThrough.js
megansmith-box: Yes I agree with both you, it's not safe to assume it's always wraps a button, I will change it back

src/components/button-group/ButtonGroup.scss
```diff
@@ -60,32 +65,37 @@
...
+    > .bdl-targeted-click-through > .btn:focus {
         z-index: 3; /* place on top of all other buttons for accessibility */
     }
 
     &.bdl-is-disabled,
     &.is-disabled {
         > .btn,
-        > .bdl-Button {
+        > .bdl-Button,
+        > .bdl-targeted-click-through > .btn:focus {
```
jstoffan, line 98: It looks like this rule only applies to `.btn`, not `.bdl-Button`. Also, should this rule have `:focus` attached? It's not present in the rule it modifies.

src/components/button-group/ButtonGroup.scss
```diff
@@ -3,13 +3,17 @@
 /**************************************
  * Button Group
  **************************************/
+
+// In the case where there is an element in between .btn and .bdl-ButtonGroup, you can add class .bdl-targeted-click-through to parent of .btn
```
jstoffan, line 7: Do we still need this comment?

megansmith-box: No, I don't think so

src/components/button-group/ButtonGroup.scss
megansmith-box: oops, good catch, I will remove the focus/

megansmith-box: I'm not sure what you mean by `It looks like this rule only applies to .btn, not .bdl-Button`?

jstoffan (approved):

PR #2568, opened 2021-05-24
fix(CreateNewFolder): [A11Y]- Add aria described by pilled selector

fsuarez-globant:
fix(CreateNewFolder): [A11Y]- Add aria described by pilled selector

Error Messages are provided but not in an accesible way in Pill selector component (Create New Folder Form). Screen reader users get no indication of the error.  I added a missing `aria-describedby` attribute in the `textarea` to be read when a form validation occurs. `Aria-invalid` (exists in Tooltips component) and `aria-describedby` are used together to indicate an error when any field are not in the expected format.

Screen Shot 2021-05-24 at 3 51 14 PM

### DIFF
--- a/src/components/pill-selector-dropdown/PillSelector.js
+++ b/src/components/pill-selector-dropdown/PillSelector.js
@@ -201,6 +201,7 @@ class PillSelector extends React.Component<Props, State> {
         const ariaAttrs = {
             'aria-invalid': hasError,
             'aria-errormessage': this.errorMessageID,
+            'aria-describedBy': this.errorMessageID,
         };
 
         return (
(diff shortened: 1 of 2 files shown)


src/components/pill-selector-dropdown/PillSelector.js
```diff
@@ -201,6 +201,7 @@ class PillSelector extends React.Component<Props, State> {
         const ariaAttrs = {
             'aria-invalid': hasError,
             'aria-errormessage': this.errorMessageID,
+            'aria-describedBy': this.errorMessageID,
```
swfree, line 204: minor: lowercase `aria-describedby`

Also - what happens when there is no error? is this just empty or undefined?

fsuarez-globant: @swfree I just changed the` aria-describedby` attribute. This `errorMessageID` is an `uniqueId`, it means , this variable is always  created when the component is created

swfree (approved):

PR #2569, opened 2021-05-25
fix(AccountSettings): A11Y- incorrect use of nav tag in TabviewPrimitive

fsuarez-globant:
fix(AccountSettings): A11Y- incorrect use of nav tag in TabviewPrimitive

Page regions such as `<header>, <nav>, <main>, <aside>, and <footer>` programmatically define the essential semantic structure of a page. The `<nav>` region should usually be applied only to the primary navigation, for this reason I changed the nav tag for a div tag.

image

### DIFF
--- a/src/components/tab-view/TabViewPrimitive.js
+++ b/src/components/tab-view/TabViewPrimitive.js
@@ -197,7 +197,7 @@ class TabViewPrimitive extends React.Component<Props, State> {
     calculatePrevIndex = (currentIndex: number, childrenCount: number) =>
         (currentIndex - 1 + childrenCount) % childrenCount;
 
-    /* eslint-disable jsx-a11y/no-noninteractive-element-to-interactive-role */
+    /* eslint-disable jsx-a11y/interactive-supports-focus */
     renderTabs() {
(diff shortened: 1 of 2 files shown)


alexkrolick:
According to WCAG, multiple `role="navigation"` elements are allowed as long as they have a label. Can we look into that approach, or at least allow props to be spread onto the outer element so that `{ role: "navigation", aria-label: "Task Tabs" }` would work?

fsuarez-globant:
> According to WCAG, multiple `role="navigation"` elements are allowed as long as they have a label. Can we look into that approach, or at least allow props to be spread onto the outer element so that `{ role: "navigation", aria-label: "Task Tabs" }` would work?
> 
> Ref: https://www.w3.org/WAI/GL/wiki/Using_ARIA_landmarks_to_identify_regions_of_a_page#Example_2

@alexkrolick The navigation landmark role is used to identify major groups of links used for navigating through a website or page content, in this case TabViewPrimitive is a tablist component (tablist components are typically placed near usually preceding, a series of tabpanel elements).

alexkrolick (approved):

PR #2573, opened 2021-05-26
fix(Tooltip): A11Y - The errormessage should not have role=tooltip 

fsuarez-globant:
fix(Tooltip): A11Y - The error message should not have role=tooltip

While it's not a bad idea to add  (role="tooltip") it to the tooltip element, it doesn't seem to do much good either. The tooltip role does not appear to affect screen reader announcements in any meaningful way -- aria-describedby and aria-labelledby do all the heavy lifting. 

Screen Shot 2021-05-26 at 3 26 35 PM


src/components/tooltip/Tooltip.tsx
```diff
@@ -325,12 +325,10 @@ class Tooltip extends React.Component<TooltipProps, State> {
                 onContextMenu={this.handleTooltipEvent}
                 onKeyPress={this.handleTooltipEvent}
             >
-                <div role="tooltip" aria-live="polite">
-                    {tooltipInner}
-                </div>
+                <div aria-live="polite">{tooltipInner}</div>
             </div>
         ) : (
-            <div className={classes} id={this.tooltipID} role="tooltip" aria-live="polite">
```
alexkrolick, line 333: Should we be removing from every tooltip or just error types?

fsuarez-globant: It should be removed for every tooltip, it won't affect screen reader announcements in any meaningful way `aria-describedby` and `aria-labelledby` do all the heavy lifting.

laughnan: Is this true @fsuarez-globant? I'm a bit leery about removing this `role` from all Tooltips unless we have firm direction that it's not needed. Some brief research gave me these links:

- Having `role` on the tooltip - https://inclusive-components.design/tooltips-toggletips/#tooltipasprimarylabel 
- Inconsistencies with `aria-label`, `aria-labelledby`, and `aria-describedby` - https://www.tpgi.com/short-note-on-aria-label-aria-labelledby-and-aria-describedby/

src/components/tooltip/Tooltip.tsx
```diff
@@ -332,17 +332,17 @@ class Tooltip extends React.Component<TooltipProps, State> {
...
                 </div>
             </div>
         ) : (
             <div
                 className={classes}
                 id={this.tooltipID}
-                role="tooltip"
                 aria-live="polite"
                 aria-hidden={isLabelMatchingTooltipText}
+                {...(theme === TooltipTheme.DEFAULT ? { role: 'tooltip' } : {})}
```
laughnan, line 345: What about `TooltipTheme.CALLOUT`? Should we just hide this role if the theme is `TooltipTheme.ERROR`?

fsuarez-globant: @laughnan You're right, I updated this PR.

jasminevle (approved):

swfree (dismissed):

src/components/tooltip/Tooltip.tsx
```diff
@@ -332,17 +332,17 @@ class Tooltip extends React.Component<TooltipProps, State> {
...
                 </div>
             </div>
         ) : (
             <div
                 className={classes}
                 id={this.tooltipID}
-                role="tooltip"
                 aria-live="polite"
                 aria-hidden={isLabelMatchingTooltipText}
+                {...(theme !== TooltipTheme.ERROR ? { role: 'tooltip' } : {})}
```
jstoffan, line 345: Can we pull this into a variable so it can be reused and consistent on line 335? Separately, does `role={theme === TooltipTheme.ERROR ? undefined : 'tooltip'}` work? We generally prefer positive rather than negative conditions unless it's a guard clause.

test/integration/content-sidebar/ContentSidebar.e2e.test.js
```diff
@@ -128,7 +128,7 @@ describe('ContentSidebar', () => {
 
     describe('activity feed comments', () => {
         const getDraftJSEditor = () => cy.getByTestId('bcs-CommentForm-body').find('[contenteditable]');
-        const getTooltip = () => cy.get('[role="tooltip"]');
```
jstoffan, line 131: We generally want to avoid selecting on CSS classes for Cypress tests, except as a last resort. Can we either add a `data-testid` property to the tooltip or use some user-visible selector, such as the actual content of the tooltip?

src/components/tooltip/Tooltip.tsx
fsuarez-globant: @jstoffan I just updated this PR with your comments

test/integration/content-sidebar/ContentSidebar.e2e.test.js
fsuarez-globant: @jstoffan I just updated this PR with your comments

src/components/tooltip/Tooltip.tsx
```diff
@@ -312,6 +313,12 @@ class Tooltip extends React.Component<TooltipProps, State> {
             tetherProps.offset = offset;
         }
 
+        if (theme === TooltipTheme.ERROR) {
```
jstoffan, line 316: It seems like this could be simplified quite a bit with a ternary.

src/components/tooltip/Tooltip.tsx
```diff
@@ -332,17 +339,23 @@ class Tooltip extends React.Component<TooltipProps, State> {
                 onContextMenu={this.handleTooltipEvent}
                 onKeyPress={this.handleTooltipEvent}
             >
-                <div role="tooltip" aria-live="polite" aria-hidden={isLabelMatchingTooltipText}>
+                <div
+                    role={tooltipRole}
+                    aria-live="polite"
+                    aria-hidden={isLabelMatchingTooltipText}
+                    data-testid="tooltip"
```
jstoffan, line 346: Let's use `bdl-Tooltip`?

src/components/tooltip/Tooltip.tsx
fsuarez-globant: @jstoffan I just updated this PR with your comments

src/components/tooltip/Tooltip.tsx
fsuarez-globant: @jstoffan I just updated this PR with your comments

i18n/en-US.properties
```diff
@@ -62,28 +62,6 @@ be.choose = Choose
 be.close = Close
 # Message to the user to collapse the Transcript entries
 be.collapse = Collapse
-# Message that appears when the request for the ContentSharing Element is malformed.
```
jstoffan, line 65: Can we revert these text changes?

fsuarez-globant: Sorry , I reverted this change

jstoffan (approved):

alexkrolick (approved):

PR #2576, opened 2021-05-27
fix(pillselectordropdown): remove label when there is no text inside

hormaza:
PillSelectorDropdown will not add a label when "label prop" is not present for accessibility purposes.


src/components/pill-selector-dropdown/PillSelectorDropdown.js
```diff
@@ -254,48 +254,48 @@ class PillSelectorDropdown extends React.Component<Props, State> {
...
+                        validator={validator}
+                        value={this.state.inputValue}
+                    />
+                }
+            >
+                {children}
+            </SelectorDropdown>
         );
+
+        return label ? <Label text={label}>{selectorDropdownElement}</Label> : selectorDropdownElement;
```
swfree, line 298: Could you add a unit test for this? Will be helpful for documenting this logic & to prevent regressions in the future. You could use test.each to test for both conditions (with & without `label` prop), and then check whether the `Label` component exists

swfree:
@hormaza can you follow the link provided in the CLAassistant comment to sign the CLA agreement?

src/components/pill-selector-dropdown/PillSelectorDropdown.js
hormaza: Hey Sky. I have commited the test for this. Let me know if it works.

src/components/pill-selector-dropdown/__tests__/PillSelectorDropdown.test.js
```diff
@@ -73,6 +73,15 @@ describe('components/pill-selector-dropdown/PillSelectorDropdown', () => {
             instance.handleBlur();
             expect(addPillsFromInputMock).toHaveBeenCalledTimes(1);
         });
+
+        test.each([
+            ['test', true],
+            ['', false],
+        ])('label', (value, expected) => {
```
swfree, line 80: can you make the test case description more specific? e.g. "should render Label component when label exists"

hormaza: Of course. Thanks for feedback

swfree (approved):

hormaza:
Thanks

PR #2575, opened 2021-05-27
feat(in-app-messenger): add new template for popout

MateuszMamczarz:
Add a new template for in-app messenger leveraging preview component
Screenshot 2021-05-27 at 15 06 36

PR #2580, opened 2021-05-28
feat(in-app-messenger): move MessagePreviewContent to common place

MateuszMamczarz:
Move MessagePreviewContent to a common place to reuse it for in-app-messenger's template

This PR address comment under https://github.com/box/box-ui-elements/pull/2575


src/components/preview-ghost/PreviewGhost.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 import * as React from 'react';
-import Ghost from '../../../../../components/ghost/Ghost';
+import Ghost from '../ghost/Ghost';
```
DanDeMicco, line 3: Thoughts on moving to `message-preview-content` folder with the other component in this file? Unless you think they will be used standalone

src/common/types/contentPreview.js
```diff
@@ -0,0 +1,19 @@
+// @flow
+import * as React from 'react';
+
+import ContentPreview from '../../elements/content-preview';
+
+export type ContentPreviewProps = $Diff<
```
jstoffan, line 6: Can we clarify why this is needed? Can we import the props directly wherever ContentPreview is being used, since that location will already have an import?

src/common/types/contentPreview.js
```diff
@@ -0,0 +1,19 @@
+// @flow
+import * as React from 'react';
+
+import ContentPreview from '../../elements/content-preview';
```
jstoffan, line 4: We typically want to avoid importing element-specific code into common/component modules.

src/components/preview-ghost/PreviewGhost.js
```diff
@@ -1,6 +1,6 @@
...
 
 function PreviewGhost() {
```
jstoffan, line 5: Can we rename this to `MessagePreviewGhost` and move it to a non-common folder to make it clear that it's not used in preview itself?

src/features/message-center/components/templates/preview-title-body-tags-button/PreviewTitleBodyTagsButton.js
```diff
@@ -1,12 +1,13 @@
...
 import type { Token } from '../../../../../common/types/core';
-import MessagePreviewContent from '../common/MessagePreviewContent';
+import MessagePreviewContent from '../../../../../components/message-preview-content/MessagePreviewContent';
 import MessageTextContent from '../common/MessageTextContent';
 import MessageTags from '../common/MessageTags';
 import MessageFooter from '../common/MessageFooter';
 import BottomContentWrapper from '../common/BottomContentWrapper';
-import type { ContentPreviewProps, PreviewTitleBodyTagsButtonTemplateParams } from '../../../types';
+import type { PreviewTitleBodyTagsButtonTemplateParams } from '../../../types';
+import type { ContentPreviewProps } from '../../../../../common/types/contentPreview';
```
jstoffan, line 10: Rather than needing to import this everywhere, can we import it in `MessagePreviewContent` and re-export it as `MessagePreviewContentProps`?

src/components/message-preview-content/messages.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+import { defineMessages } from 'react-intl';
+
+const messages = defineMessages({
+    previewError: {
```
jstoffan, line 5: Can we update this to be `messagePreviewError`?

jstoffan (commented):
We should be cautious about labelling things "Preview", as it is already a heavily used prefix that we try to reserve for preview itself rather than its usages. Further, the `components` folder is meant for very low-level controls, like button, tooltip, etc. Should we create a new `features` folder to contain these message-center modules, instead?

src/components/preview-ghost/PreviewGhost.js
MateuszMamczarz: PreviewGhost is used not only in MessagePreviewContent, but also in message-center-modal/MessageCenterModal.js, that's why I created a new folder for it

src/components/preview-ghost/PreviewGhost.js
MateuszMamczarz: This change will require to update some integration tests, but other than that I'm fine with it. Any thoughts @DanDeMicco?

src/common/types/contentPreview.js
MateuszMamczarz: It does make sense for me to move it to `MessagePreviewContent`.  I added it to the file I'm not sure if we want to keep the type there or I should rather create a dedicated file for it. What is you opinion?

src/common/types/contentPreview.js
MateuszMamczarz: Moved the type definition to features/message-preview-content, so that's not an issue anymore, thank you!

src/features/message-center/components/templates/preview-title-body-tags-button/PreviewTitleBodyTagsButton.js
MateuszMamczarz: Done, but I'm not sure how it improves the import, could you elaborate on that a little bit more?

src/components/message-preview-content/messages.js
MateuszMamczarz: Done

MateuszMamczarz:
@DanDeMicco @jstoffan I believe I addressed your comments, could you please take a look on this PR once again?

DanDeMicco (approved):

src/features/message-preview-content/MessagePreviewContent.js
```diff
@@ -1,14 +1,28 @@
 // // @flow
 import * as React from 'react';
 import classNames from 'classnames';
-import ContentPreview from '../../../../../elements/content-preview';
-import type { Token } from '../../../../../common/types/core';
-import type { ContentPreviewProps } from '../../../types';
-import Cache from '../../../../../utils/Cache';
-import PreviewGhost from './PreviewGhost';
+import ContentPreview from '../../elements/content-preview';
```
jstoffan, line 4: My thought regarding simplifying the imports for props would look something like this, which would eliminate the direct reference to ContentPreview in all the consuming components:

```
// MessagePreviewContent.js
import ContentPreview, { type ContentPreviewProps } from '../../elements/content-preview';
export type ContentPreviewProps = ContentPreviewProps;

// PreviewTitleBodyTag.js (and others)
import MessagePreviewContent, { type ContentPreviewProps } from '../../../../message-preview-content/MessagePreviewContent';

jstoffan (approved):

PR #2581, opened 2021-05-28
fix(Toggle): A11Y - When shared link does not exist maintain focus on the switch

fsuarez-globant:
fix(Toggle): A11Y - When shared link does not exist maintain focus on the switch

When the sharing option has been deactivated and `Create shared link` is triggered, the form changes and focus is set to the read-only textbox that contains the URL. This skips over the `Link Settings` option, meaning a screen reader user will not know that this option is available. Instead of setting focus to the URL, maintain focus on the switch.

Screen Shot 2021-05-28 at 5 56 10 PM

### DIFF
--- a/src/features/unified-share-modal/SharedLinkSection.js
+++ b/src/features/unified-share-modal/SharedLinkSection.js
@@ -129,10 +129,12 @@ class SharedLinkSection extends React.Component<Props, State> {
             this.setState({ isAutoCreatingSharedLink: true });
             addSharedLink();
         }
+
         if (!prevProps.sharedLink.url && sharedLink.url) {
             this.setState({ isAutoCreatingSharedLink: false });
-        } else if (!prevProps.sharedLink.url && !sharedLink.url && this.toggleRef) {
-            this.toggleRef.focus();
+            if (this.toggleRef) {
+                this.toggleRef.focus();
+            }
         }
 
         if (
--- a/src/features/unified-share-modal/SharedLinkSection.js
+++ b/src/features/unified-share-modal/SharedLinkSection.js
@@ -156,12 +156,6 @@ class SharedLinkSection extends React.Component<Props, State> {
                     onCopyError();
                 });
         }
-
-        if (sharedLink.url === null) {
-            if (this.toggleRef) {
-                this.toggleRef.focus();
-            }
-        }
     }
 
     canAddSharedLink = (isSharedLinkEnabled: boolean, canAddLink: boolean) => {
--- a/src/features/unified-share-modal/SharedLinkSection.js
+++ b/src/features/unified-share-modal/SharedLinkSection.js
@@ -70,7 +70,7 @@ class SharedLinkSection extends React.Component<Props, State> {
         autoCreateSharedLink: false,
     };
 
-    toggleRef: HTMLInputElement | null | undefined;
+    toggleRef: HTMLInputElement | null;
 
     constructor(props: Props) {
         super(props);
--- a/src/features/unified-share-modal/SharedLinkSection.js
+++ b/src/features/unified-share-modal/SharedLinkSection.js
@@ -132,6 +132,8 @@ class SharedLinkSection extends React.Component<Props, State> {
 
         if (!prevProps.sharedLink.url && sharedLink.url) {
             this.setState({ isAutoCreatingSharedLink: false });
+        } else if (!sharedLink.url && this.toggleRef) {
+            this.toggleRef.focus();
         }
 
         if (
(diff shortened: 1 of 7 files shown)


src/features/unified-share-modal/SharedLinkSection.js
```diff
@@ -151,6 +153,12 @@ class SharedLinkSection extends React.Component<Props, State> {
                     onCopyError();
                 });
         }
+
+        if (sharedLink.url === null) {
```
swfree, line 157: Based on the flow type definition for the sharedLink prop, sharedLink.url won't be null but rather an empty string "" when there is no shared link. Also you'll want to check for a change in sharedLink url rather than anytime its empty, otherwise you'll also focus on the toggle when the modal first launches. I recommend moving your toggleRef focus logic up into this block: https://github.com/box/box-ui-elements/blob/master/src/features/unified-share-modal/SharedLinkSection.js#L131-L133

src/features/unified-share-modal/SharedLinkSection.js
```diff
@@ -151,6 +153,12 @@ class SharedLinkSection extends React.Component<Props, State> {
...
+            if (this.toggleRef) {
+                this.toggleRef.focus();
```
swfree, line 159: This will also only focus on the toggle component when the `autofocusSharedLink` prop is false https://github.com/box/box-ui-elements/blob/master/src/features/unified-share-modal/SharedLinkSection.js#L242. Is that at odds with the accessibility requirement?

fsuarez-globant: @swfree when `autofocusSharedLink = false` the requirement says that it should focus in the invite people input field.

src/features/unified-share-modal/SharedLinkSection.js
```diff
@@ -130,6 +132,8 @@ class SharedLinkSection extends React.Component<Props, State> {
 
         if (!prevProps.sharedLink.url && sharedLink.url) {
             this.setState({ isAutoCreatingSharedLink: false });
+        } else if (!sharedLink.url && this.toggleRef) {
```
swfree, line 135: This logic will still focus on the toggle when the modal first loads. Since you want to focus on the toggle only when going from deactivated (!prevProps.sharedLink.url) to activated (sharedLink.url), you want to nest it under the condition above:
```
if (!prevProps.sharedLink.url && sharedLink.url) {
    this.setState({ isAutoCreatingSharedLink: false });
    if (this.toggleRef) {
        this.toggleRef.focus();
    }
}
```

fsuarez-globant: @swfree I consider the logic is OK because I want to focus on the toggle only when `sharedLink.url` is null or empty it means when I click in shared button and it does not auto generate a link. You're right when you say that the SharedLinkSection first load it will focus to the toggle, however it will focus to `TextInputWithCopyButton` when  `sharedLink.url` changes from null or empty to any other value and it will work as expected. 

I did this change for recommendation of WebAIM in order to focus the most important feature in the modal which is Share Link toggle only when `sharedLink.url` does not exist instead to focus on invite people input.

src/features/unified-share-modal/__tests__/UnifiedShareForm.test.js
```diff
@@ -555,6 +555,7 @@ describe('features/unified-share-modal/UnifiedShareForm', () => {
 
         test('should return true if forced focus and link is loaded', () => {
             const wrapper = getWrapper({
+                createSharedLinkOnLoad: true,
```
swfree, line 558: Can you remove this line from this test? Since this test will still pass without this line, it'd be preferable to preserve this test case (which is specifically for "forced focus and link is loaded"), and instead add a new test case to document the expected behavior for createSharedLinkOnLoad since that's a new logic change

src/features/unified-share-modal/__tests__/UnifiedShareForm.test.js
```diff
@@ -567,6 +568,7 @@ describe('features/unified-share-modal/UnifiedShareForm', () => {
 
         test('should return true if new shared link and link is loaded', () => {
```
swfree, line 569: can you update this test case description to include that createSharedLinkOnLoad is true here?

swfree: can you also add a test case that is similar to this one, but without `createSharedLinkOnLoad: true`? this will help document the expected behavior for just `focusSharedLinkOnLoad: false`

fsuarez-globant:
@swfree I just updated the unit tests

swfree (approved):

swfree (approved):

alexkrolick (approved):

PR #2582, opened 2021-05-29
fix(sidebar): Change Aside element by Header

tille:
image


laughnan (commented):
This will introduce two `<header>` tags into the application, is this acceptable?

> Most pages also have only one <header> and <footer>. <header>, <main>, and <footer> must be direct children of <body> to be exposed to screen readers. They cannot be nested within other container elements. All page content must be contained in a region.
(https://webaim.org/techniques/semanticstructure/)

### Existing <header> UI
Screen Shot 2021-06-01 at 11 56 00

### Existing <header> HTML
Screen Shot 2021-06-01 at 11 55 54

laughnan (commented):
Approving since we have guidance that this change (from `<aside>` to `<div>` is acceptable for accessibility).

laughnan (approved):

PR #2584, opened 2021-06-01
refactor(typescript): Migrate Feature Checking

jstoffan:
I'm not sure this is exactly worth doing, since the system is pretty open type-wise, but maybe it's worth a try.


src/elements/common/feature-checking/FeatureConsumer.tsx
```diff
@@ -0,0 +1,6 @@
+// @flow
```
ConradJChan, line 1: Is this and the other `// @flow` still needed?

jstoffan: Drat. I copied these by hand, so this declaration snuck in. Will remove.

ConradJChan (dismissed):

alexkrolick:
Would be nice to have a hook for the FFs too

PR #2589, opened 2021-06-03
fix(sharelinksection): add isTabbable property setted as false to tooltip component

matiaslionel:
To reproduce this: you should try to share link of a file. Share a document and a Modal will be opened. If you navigate through this form in order to go to share link toggle element you will notice that you need a coulple more of tabs to get there.

This is happening because the toggle component is wrapped in a Tooltip component.

Luckly Tooltip component has a property to do not set TabIndex. We can change this behaivior changing the property tablable to false (which is comming true as default)


src/features/unified-share-modal/SharedLinkSection.js
```diff
@@ -373,6 +373,7 @@ class SharedLinkSection extends React.Component<Props, State> {
                     <span>
                         <FormattedMessage {...messages.linkShareOn} />
                         <Tooltip
+                            isTabbable={false}
```
swfree, line 376: Should this component also be set to isTabbable=false? Your bugfix is for the toggle component, whereas this is for the toggle's label. How is the behavior changing here?

swfree (approved):

PR #2590, opened 2021-06-04
feat(content-picker): pass custom buttons to ContentPicker

ChenCodes:
**Purpose behind the change:**
We have a use case in the parent app of ContentPicker where we want to have custom buttons in the Footer component that ContentPicker is currently using.

ContentPicker is implemented in a way such that we only show a Cancel / Select button in the footer. In the parent app, we want to have a few custom buttons such as Move & Copy. This change involves extending ContentPicker to allow the passing down of multiple custom buttons.


src/elements/content-picker/ContentPicker.js
```diff
@@ -290,6 +292,15 @@ class ContentPicker extends Component<Props, State> {
         }
     }
 
+    getSelectedItems = (): BoxItem[] => {
```
swfree, line 295: Can you add comments to document this method, similar to other methods in this file?

src/elements/content-picker/Footer.js
```diff
@@ -58,21 +64,25 @@ const Footer = ({
...
-                    isDisabled={!selectedCount}
-                    text={chooseButtonLabel || <FormattedMessage {...messages.choose} />}
-                >
-                    <PrimaryButton isDisabled={!selectedCount} onClick={onChoose} type="button">
-                        <IconCheck color="#fff" height={16} width={16} />
-                    </PrimaryButton>
-                </Tooltip>
-            </ButtonGroup>
+            {renderCustomActionButtons ? (
+                renderCustomActionButtons({ selectedCount, selectedItems })
```
swfree, line 68: What do you think about passing down more props, in case the use case grows in the future? E.g. I'm wondering if onChoose or onCancel might still be useful props to pass down

src/elements/content-picker/ContentPicker.js
```diff
@@ -95,6 +96,7 @@ type Props = {
     messages?: StringMap,
     onCancel: Function,
     onChoose: Function,
+    renderCustomActionButtons?: ({ selectedCount: number, selectedItems: BoxItem[] }) => Node,
```
swfree, line 99: Could you also add this new prop to the readme for ContentPicker https://github.com/box/box-ui-elements/blob/master/src/elements/content-picker/README.md?

src/elements/content-picker/ContentPicker.js
ChenCodes: Yep!

src/elements/content-picker/README.md
```diff
@@ -40,6 +40,7 @@ var ContentPicker = require('./ContentPicker').default;
 | onCancel | function |  | Callback function for when the cancel button is pressed. |
 | onChoose | function |  | Callback function for when the choose button is pressed. |
 | requestInterceptor | function | | *See the [developer docs](https://developer.box.com/guides/embed/ui-elements/picker/#Options).* |
+| renderCustomActionButtons | function | React.Node | Render prop that renders custom action buttons in the footer. |
```
swfree, line 43: you can remove `React.Node` since that column is "Default" - you can leave it empty since there is no default value for renderCustomActionButtons

swfree (approved):

PR #2592, opened 2021-06-04
feat(sign): Add ftux tooltip for Box Sign entry point

jstoffan:
box-sign-ftux

### DIFF
--- a/src/elements/content-sidebar/SidebarNavSign.tsx
+++ b/src/elements/content-sidebar/SidebarNavSign.tsx
@@ -2,6 +2,8 @@ import React from 'react';
 import { injectIntl, IntlShape } from 'react-intl';
 import BoxSign28 from '../../icon/logo/BoxSign28';
 import PlainButton, { PlainButtonProps } from '../../components/plain-button';
+// @ts-ignore Module is written in Flow
+import TargetedClickThroughGuideTooltip from '../../features/targeting/TargetedClickThroughGuideTooltip';
 import Tooltip, { TooltipPosition } from '../../components/tooltip';
 // @ts-ignore Module is written in Flow
 import messages from './messages';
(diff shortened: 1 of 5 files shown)


src/elements/content-sidebar/SidebarNavSign.tsx
```diff
@@ -10,17 +12,35 @@ import './SidebarNavSign.scss';
 export type Props = PlainButtonProps & {
     intl: IntlShape;
     status?: string;
+    targetingApi?: {
```
ConradJChan, line 15: Is there no existing type for this?

src/elements/content-sidebar/SidebarNavSign.tsx
```diff
@@ -10,17 +12,35 @@ import './SidebarNavSign.scss';
...
+        onShow: () => void;
+    };
 };
 
-export function SidebarNavSign({ intl, status, ...rest }: Props) {
+export const PlaceholderTooltip = ({ children }: { children: React.ReactNode }) => children;
+
+export function SidebarNavSign({ intl, status, targetingApi, ...rest }: Props) {
+    const isTargeted = targetingApi && targetingApi.canShow;
+    const FtuxTooltip = isTargeted ? TargetedClickThroughGuideTooltip : PlaceholderTooltip;
```
ConradJChan, line 27: Is the `PlaceholderTooltip` needed? Seems like if `targetingApi.canShow` is false, then the tooltip should not show already, do we need to not render it also?

src/elements/content-sidebar/SidebarNavSign.tsx
jstoffan: Yes, but only for Flow.

src/elements/content-sidebar/SidebarNavSign.tsx
jstoffan: The `TargetedClickThroughGuideTooltip` doesn't check to see if `targetingApi` is defined before trying to use it. I figured that pre-empting its rendering was the safest way to handle it.

ConradJChan (approved):

PR #2596, opened 2021-06-07
fix(tooltip): Modify tooltip to allow rerender after closed by user

JohnnyB83:
add optional prop `shouldTooltipShowAferUserClose` to `Tooltip.tsx`
add logic to allow for the tooltip to be opened again if it was closed by the user for restarting onboarding experiences multiple times

### DIFF
--- a/src/components/tooltip/Tooltip.tsx
+++ b/src/components/tooltip/Tooltip.tsx
@@ -87,8 +87,6 @@ export type TooltipProps = {
     children: React.ReactChild;
     /** A CSS class for the tooltip */
     className?: string;
-    /** Should the tooltip show again after user closes it once it is eligible again. Defaults to `false` */
-    shouldTooltipShowAfterUserClose?: boolean;
     /** Forces the tooltip to be shown or hidden (useful for errors) */
     isShown?: boolean;
     /** Whether to add tabindex=0.  Defaults to `true` */
@@ -128,8 +126,24 @@ class Tooltip extends React.Component<TooltipProps, State> {
         this.state = { isShown: !!props.isShown, hasRendered: false, wasClosedByUser: false };
     }
 
+    isControlled = () => {
+        const { isShown: isShownProp } = this.props;
+        return typeof isShownProp !== 'undefined';
+    };
+
     componentDidMount() {
         this.setState({ hasRendered: true });
+    }
+
+    componentDidUpdate(prevProps: TooltipProps, prevState: State) {
+        // Reset wasClosedByUser state when isShown transitions from false to true
+        if (this.isControlled()) {
+            if (!prevProps.isShown && this.props.isShown) {
+                this.setState({ wasClosedByUser: false });
+            }
+        } else if (!prevState.isShown && this.state.isShown) {
+            this.setState({ wasClosedByUser: false });
+        }
     }
 
     tooltipID = uniqueId('tooltip');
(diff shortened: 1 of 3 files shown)


src/components/tooltip/Tooltip.tsx
```diff
@@ -87,6 +87,8 @@ export type TooltipProps = {
     children: React.ReactChild;
     /** A CSS class for the tooltip */
     className?: string;
+    /** Should the tooltip show again after user closes it once it is eligible again. Defaults to `false` */
+    shouldTooltipShowAferUserClose?: boolean;
```
megansmith-box, line 91: it looks like this was copied with typo: should be shouldTooltipShowAfterUserClose

src/components/tooltip/Tooltip.tsx
```diff
@@ -195,12 +197,13 @@ class Tooltip extends React.Component<TooltipProps, State> {
     };
 
     isShown = () => {
-        const { isShown: isShownProp } = this.props;
+        const { isShown: isShownProp, shouldTooltipShowAferUserClose = false } = this.props;
```
DanDeMicco, line 200: Am wondering if you can implement a `componentDidUpdate` and watch for `isShown` changes. If it changes from false to true, set `wasClosedByUser` to false. Or, if you can somehow render the tooltip with a key which would force a re-mount. Your solution also works, just trying to avoid adding a new optional prop

src/components/tooltip/Tooltip.tsx
JohnnyB83: good catch, updated

src/components/tooltip/Tooltip.tsx
JohnnyB83: will play around with this to see if i can get it to work without the prop

src/components/tooltip/Tooltip.tsx
```diff
@@ -126,10 +126,26 @@ class Tooltip extends React.Component<TooltipProps, State> {
...
         this.setState({ hasRendered: true });
     }
 
+    componentDidUpdate(prevProps: TooltipProps, prevState: State) {
+        // Reset wasClosedByUser state when isShown transitions from false to true
+        if (this.isControlled()) {
+            if (!prevProps.isShown && this.props.isShown) {
+                this.setState({ wasClosedByUser: false });
+            }
+        } else if (!prevState.isShown && this.state.isShown) {
```
DanDeMicco, line 144: In what case would this block be executed / be needed? When would isShown be flipped from false to true, after `wasClosedByUser === true`?

I think youd only need to handle the controlled, but maybe am missing something you are seeing

JohnnyB83: looked at it again and you are right, doesn't make sense to check for the state flip, updated to just do this for the controlled case

src/components/tooltip/Tooltip.tsx
```diff
@@ -126,10 +126,24 @@ class Tooltip extends React.Component<TooltipProps, State> {
         this.state = { isShown: !!props.isShown, hasRendered: false, wasClosedByUser: false };
     }
 
+    isControlled = () => {
```
DanDeMicco, line 129: [nit] can this be in the original spot? order shouldnt matter in class component

DanDeMicco (approved):

src/components/tooltip/Tooltip.tsx
JohnnyB83: good point, updated

DanDeMicco (approved):

PR #2604, opened 2021-06-15
fix(SidebarSection): A11Y - add aria-expanded in sidebar section component

fsuarez-globant:
In the tab panel, the"Access States" and "File Properties" buttons should have aria-expanded. I just added this aria attribute in `SidebarSection`  component.

Screen Shot 2021-06-15 at 11 19 10 AM

### DIFF
--- a/src/elements/content-sidebar/SidebarSection.js
+++ b/src/elements/content-sidebar/SidebarSection.js
@@ -88,6 +88,7 @@ class SidebarSection extends React.PureComponent<Props, State> {
                         data-resin-target={interactionTarget}
                         onClick={this.toggleVisibility}
                         type="button"
+                        aria-expanded={!!isOpen}
                     >
                         {title}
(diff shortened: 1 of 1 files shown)


jasminevle (approved):

src/elements/content-sidebar/SidebarSection.js
```diff
@@ -88,6 +88,7 @@ class SidebarSection extends React.PureComponent<Props, State> {
                         data-resin-target={interactionTarget}
                         onClick={this.toggleVisibility}
                         type="button"
+                        aria-expanded={!!isOpen}
```
jstoffan, line 91: Is the `!!` needed here, since this value should always be defined? Can we keep these attributes sorted alphabetically, as well?

fsuarez-globant: @jstoffan I just updated this PR.

jstoffan (approved):

PR #2605, opened 2021-06-16
fix(contentsidebar): change background to white

hormaza:
Change background of right sidebar because contrast issue.


jstoffan (approved):
LGTM, so long as this change has been validated for all the sidebar panels by design.

PR #2606, opened 2021-06-16
feat(in-app-messenger): add new template for modal

MateuszMamczarz:
Add new template for in-app messenger leveraging preview component
Screenshot 2021-06-16 at 20 58 24

### DIFF
--- a/src/features/in-app-messenger/contextual/templates/PreviewTitleBodyTwoButtonsModalTemplate.js
+++ b/src/features/in-app-messenger/contextual/templates/PreviewTitleBodyTwoButtonsModalTemplate.js
@@ -13,7 +13,7 @@ import Button from '../../../../components/button';
 import MessagePreviewContent from '../../../message-preview-content/MessagePreviewContent';
 import type { Token } from '../../../../common/types/core';
 import { messageActions } from '../../types/message-actions';
-import { type MessageActions, type PreviewTitleBodyTwoButtonsModalParams } from '../../types';
+import { type MessageActions, type PreviewTitleBodyTwoButtonsParams } from '../../types';
 
 import './styles/PreviewTitleBodyTwoButtonsModalTemplate.scss';
 
@@ -22,7 +22,7 @@ type Props = {
     contentPreviewProps?: ContentPreviewProps,
     getToken: (folderID: string | number) => Promise<Token>,
     onAction: (MessageActions, ...Array<any>) => any,
-    params: PreviewTitleBodyTwoButtonsModalParams,
+    params: PreviewTitleBodyTwoButtonsParams,
 };
 
 const handleClose = onAction => {
--- a/src/features/in-app-messenger/contextual/templates/PreviewTitleBodyTwoButtonsModalTemplate.js
+++ b/src/features/in-app-messenger/contextual/templates/PreviewTitleBodyTwoButtonsModalTemplate.js
@@ -1,10 +1,5 @@
 // @flow
 
-/**
- * This component is intended to be passed to TemplateContainer as templateComponent prop, like this:
- *   <TemplateContainer templateID='my-template' templateComponent=MyTemplate/>
- */
-
 import * as React from 'react';
 
 import PrimaryButton from '../../../../components/primary-button';
--- a/src/features/in-app-messenger/contextual/templates/PreviewTitleBodyTwoButtonsModalTemplate.js
+++ b/src/features/in-app-messenger/contextual/templates/PreviewTitleBodyTwoButtonsModalTemplate.js
@@ -1,0 +1,98 @@
+// @flow
+
+/**
+ * This component is intended to be passed to TemplateContainer as templateComponent prop, like this:
+ *   <TemplateContainer templateID='my-template' templateComponent=MyTemplate/>
+ */
+
+import * as React from 'react';
+
+import PrimaryButton from '../../../../components/primary-button';
+import { Modal, ModalActions } from '../../../../components/modal';
+import Button from '../../../../components/button';
+import MessagePreviewContent from '../../../message-preview-content/MessagePreviewContent';
+import { messageActions } from '../../types/message-actions';
+import { MessageActions, type PreviewTitleBodyTwoButtonsModalParams } from '../../types';
+
+import './styles/PreviewTitleBodyTwoButtonsModalTemplate.scss';
+
+type Props = {
+    apiHost: string,
+    contentPreviewProps?: ContentPreviewProps,
+    getToken: (folderID: string | number) => Promise<Token>,
+    onAction: (MessageActions, ...Array<any>) => any,
+    params: PreviewTitleBodyTwoButtonsModalParams,
+};
+
+const handleClose = onAction => {
+    onAction([messageActions.close]);
+};
+
+const handleButton1Click = (onAction, button1) => {
+    if (button1) {
+        onAction(button1.actions);
+    }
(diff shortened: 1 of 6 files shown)


src/features/in-app-messenger/contextual/templates/PreviewTitleBodyTwoButtonsModalTemplate.js
```diff
@@ -0,0 +1,98 @@
...
+    getToken: (folderID: string | number) => Promise<Token>,
+    onAction: (MessageActions, ...Array<any>) => any,
+    params: PreviewTitleBodyTwoButtonsModalParams,
+};
+
+const handleClose = onAction => {
+    onAction([messageActions.close]);
+};
+
+const handleButton1Click = (onAction, button1) => {
```
megansmith-box, line 31: Why do we need 2 separate function for handleButton1Click & handleButton2Click?

src/features/in-app-messenger/contextual/templates/PreviewTitleBodyTwoButtonsModalTemplate.js
```diff
@@ -0,0 +1,98 @@
...
+
+const handleButton2Click = (onAction, button2) => {
+    if (button2) {
+        onAction(button2.actions);
+    }
+};
+
+const PreviewTitleBodyTwoButtonsModalTemplate = ({
+    apiHost,
+    contentPreviewProps,
```
megansmith-box, line 45: do we need a default prop since it's optional?

src/features/in-app-messenger/contextual/templates/__tests__/PreviewTitleBodyTwoButtonsModalTemplate.test.js
```diff
@@ -0,0 +1,84 @@
...
+describe('features/in-app-messenger/contextual/templates/PreviewTitleBodyTwoButtonsModalTemplate', () => {
+    const onAction = jest.fn();
+
+    const paramsConfigs = {
+        all: {
+            params: {
+                body: 'body',
+                button1: {
+                    label: 'button1',
+                    actions: 'actions1',
```
megansmith-box, line 15: should actions be an array?

src/features/in-app-messenger/contextual/templates/styles/PreviewTitleBodyTwoButtonsModalTemplate.scss
```diff
@@ -0,0 +1,78 @@
...
+            max-width: 400px;
+            max-height: 200px;
+
+            .bp {
+                background-color: $white;
+            }
+        }
+    }
+
+    .modal-actions {
```
megansmith-box, line 39: should these styles be more specific so that they don't apply to all `.modal-actions`? such as `.bdl-PreviewTitleBodyTwoButtonsModalTemplate-title .modal-actions`

src/features/in-app-messenger/contextual/templates/PreviewTitleBodyTwoButtonsModalTemplate.js
```diff
@@ -0,0 +1,98 @@
...
+    params: { body, button1, button2, fileUpload: { fileId, sharedLinkUrl } = {}, title },
+}: Props) => {
+    return (
+        <Modal
+            className="bdl-PreviewTitleBodyTwoButtonsModalTemplate"
+            closeButtonProps={{ 'data-resin-target': 'dismiss' }}
+            isOpen
+            onRequestClose={() => handleClose(onAction)}
+            shouldNotUsePortal
+            title=""
```
megansmith-box, line 57: can we exclude title since it's an optional prop?

src/features/in-app-messenger/contextual/templates/styles/PreviewTitleBodyTwoButtonsModalTemplate.scss
MateuszMamczarz: They are inside `.bdl-PreviewTitleBodyTwoButtonsModalTemplate` block, so it's like `.bdl-PreviewTitleBodyTwoButtonsModalTemplate .modal-actions`. The style should be applied only for modal with this specific template.

src/features/in-app-messenger/contextual/templates/__tests__/PreviewTitleBodyTwoButtonsModalTemplate.test.js
MateuszMamczarz: Yes, for testing purposes I used string to only check if correct action is being used, but I can also put the string to an array to be more precise

src/features/in-app-messenger/contextual/templates/styles/PreviewTitleBodyTwoButtonsModalTemplate.scss
megansmith-box: got it, I didn't see it was wrapped, thanks!

src/features/in-app-messenger/contextual/templates/PreviewTitleBodyTwoButtonsModalTemplate.js
MateuszMamczarz: I don't think so, we are passing these props to `ContentPreview`, in that way we can provide any props needed for it, but I believe nothing is really required

src/features/in-app-messenger/contextual/templates/__tests__/PreviewTitleBodyTwoButtonsModalTemplate.test.js
```diff
@@ -0,0 +1,84 @@
...
+                templateID: 'preview-title-body-two-buttons-modal-template',
+                title: 'title',
+            },
+        },
+        missingButton2: {
+            params: {
+                body: 'body',
+                button1: {
+                    label: 'button1',
+                    actions: 'actions1',
```
megansmith-box, line 34: One more! I think this should be an array?

src/features/in-app-messenger/types/template-params.js
```diff
@@ -18,3 +18,12 @@ export type PreviewTitleBodyTwoButtonsPopoutParams = {|
     title: string,
     ...PreviewParams,
 |};
+
+export type PreviewTitleBodyTwoButtonsModalParams = {|
```
megansmith-box, line 22: This looks the same as the props above, can we use a generic type for both? maybe named `PreviewTitleBodyTwoButtonsParams`

MateuszMamczarz: Good point, it had different props when I was creating it on the beginning, but after a refactor it became the same. Good catch!

src/features/in-app-messenger/contextual/templates/__tests__/PreviewTitleBodyTwoButtonsModalTemplate.test.js
```diff
@@ -0,0 +1,84 @@
...
+
+    beforeEach(() => {});
+
+    afterEach(() => {
+        jest.resetAllMocks();
+    });
+
+    describe.each([paramsConfigs.all, paramsConfigs.missingButton2])('%o', ({ params }) => {
+        test('renders correctly', () => {
+            getWrapper(params);
```
DanDeMicco, line 58: this has no assertions

DanDeMicco (approved):

PR #2610, opened 2021-06-18
fix(ContentPreview): A11Y - remove index zero to avoid tab stops

fsuarez-globant:
fix(ContentPreview): A11Y - remove index zero to avoid tab stops

The file preview itself is contained in one` <div>` elements with `tabindex="0"`. This `<div>` in the keyboard navigation flow, creating invisible Tab stops with no functionality. This are likely to create confusion for screen reader and keyboard users. For this reason I removed `tabindex=0 `in the `ContentPreview` component.

Before change 
Screen Shot 2021-06-18 at 3 57 07 PM
After change
Screen Shot 2021-06-18 at 3 58 52 PM

### DIFF
--- a/src/elements/content-preview/ContentPreview.js
+++ b/src/elements/content-preview/ContentPreview.js
@@ -1269,10 +1269,9 @@ class ContentPreview extends React.PureComponent<Props, State> {
         const onHeaderClose = currentVersionId === selectedVersionId ? onClose : this.updateVersionToCurrent;
 
         /* eslint-disable jsx-a11y/no-static-element-interactions */
-        /* eslint-disable jsx-a11y/no-noninteractive-tabindex */
         return (
             <Internationalize language={language} messages={messages}>
-                <div id={this.id} className={styleClassName} ref={measureRef} onKeyDown={this.onKeyDown} tabIndex={0}>
+                <div id={this.id} className={styleClassName} ref={measureRef} onKeyDown={this.onKeyDown}>
                     {hasHeader && (
                         <PreviewHeader
                             file={file}


src/elements/content-preview/ContentPreview.js
```diff
@@ -1269,10 +1269,9 @@ class ContentPreview extends React.PureComponent<Props, State> {
         const onHeaderClose = currentVersionId === selectedVersionId ? onClose : this.updateVersionToCurrent;
 
         /* eslint-disable jsx-a11y/no-static-element-interactions */
-        /* eslint-disable jsx-a11y/no-noninteractive-tabindex */
         return (
             <Internationalize language={language} messages={messages}>
-                <div id={this.id} className={styleClassName} ref={measureRef} onKeyDown={this.onKeyDown} tabIndex={0}>
+                <div id={this.id} className={styleClassName} ref={measureRef} onKeyDown={this.onKeyDown}>
```
jstoffan, line 1274: @fsuarez-globant, does the `keydown` handling on this element still work without the tabindex? The left and right arrow keys should allow the user to move to the previous and next file in a collection, respectively.

fsuarez-globant:
@jstoffan I checked that the functionality of moving files will work when users presses the right and left arrow. This feature works because inside the div where the tabIndex was removed, there are another two tabindexes ( document and thumbnails side bar) and it can trigger the onKeydown events.

jstoffan:
@fsuarez-globant, thanks for checking. However, since this attribute was added quite a while ago via 1a0ff4cc72ddfe7031b8ab43c88cdcf04a4b2bcb, I'm concerned we're missing something and risk causing a regression. Does the autofocus feature still work properly? Does left/right work with all file types (images, videos, audio, text, etc.)?

PR #2617, opened 2021-06-24
feat(content-picker): add flag for hiding pagination buttons

ChenCodes:
Changes in this PR:
- [x] add flag in ContentPicker to decide whether or not the Pagination buttons in the Footer should be shown

### DIFF
--- a/src/elements/content-picker/ContentPicker.js
+++ b/src/elements/content-picker/ContentPicker.js
@@ -89,6 +89,7 @@ type Props = {
     initialPageSize: number,
     isHeaderLogoVisible?: boolean,
     isLarge: boolean,
+    isPaginationVisible?: boolean,
     isSmall: boolean,
     isTouch: boolean,
     language?: string,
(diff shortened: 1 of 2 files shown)


src/elements/content-picker/ContentPicker.js
```diff
@@ -1282,12 +1285,14 @@ class ContentPicker extends Component<Props, State> {
                             cancelButtonLabel={cancelButtonLabel}
                             renderCustomActionButtons={renderCustomActionButtons}
                         >
-                            <OffsetBasedPagination
-                                offset={offset}
-                                onOffsetChange={this.paginate}
-                                pageSize={currentPageSize}
-                                totalCount={totalCount}
-                            />
+                            {isPaginationVisible && (
```
swfree, line 1288: Slight preference for `{isPaginationVisible ? (...) : null}`. isPaginationVisible is documented as a boolean so the user shouldn't pass in any other values, but it would be nice to cover the possible off-chance that a user passes in a different falsy value. (E.g. if a user decided to pass in `isPaginationVisible=0`, the && condition would display a `0`)

ChenCodes: Good catch! Will update

swfree (approved):

PR #2618, opened 2021-06-25
fix(commentform): A11Y - color contrast correction

hormaza:
The grey (#09090) "@mention users to notify them" text presents an insufficient contrast ratio of 3.2:1 with the white background

image

The fix should change the color to #767676:

image


ConradJChan (approved):

ConradJChan:
Should this have `A11y` in the name of the commit like the others?

hormaza:
@ConradJChan I've updated the title

PR #2620, opened 2021-06-29
fix(countbadge): tweak contrast color

tille:
since this color will only be used in this single instance, I didn't add a variable for it in the colors-palete 

Screen Shot 2021-06-29 at 3 44 05 PM

now contrast is 4.5 which pass a11y AA standards

Screen Shot 2021-06-29 at 3 46 43 PM

### DIFF
--- a/src/components/count-badge/CountBadge.scss
+++ b/src/components/count-badge/CountBadge.scss
@@ -36,7 +36,7 @@ $bdl-CountBadge-background: $bdl-watermelon-red;
     font-size: 11px;
     line-height: 16px;
     text-align: center;
-    background-color: $bdl-CountBadge-background;
+    background-color: #e9163c;
     border-radius: 16px;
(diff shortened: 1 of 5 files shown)


src/components/count-badge/CountBadge.scss
```diff
@@ -36,7 +36,7 @@ $bdl-CountBadge-background: $bdl-watermelon-red;
     font-size: 11px;
     line-height: 16px;
     text-align: center;
-    background-color: $bdl-CountBadge-background;
+    background-color: #e9163c;
```
laughnan, line 39: Can we us a Box color? `$bdl-watermelon-red`?

https://opensource.box.com/box-ui-elements/storybook/?path=/story/theming-colors--all-colors

tille: unfortunaly not this time... we talk this with David and there is no red-color with enough contrast

jasminevle (approved):

PR #999, opened 2019-03-25
chore(travis): introduce flow error

JustinHoldstock:

PR #2621, opened 2021-07-01
fix(metadata): improved color contrast on edit icon button

msorianoBox:
The edit icon on the Metadata section used to have a #999 fill color, which does not meet the minimum contrast requirement over white background. 

Screen Shot 2021-07-01 at 10 18 36

Color had to be changed to #767676 in order to meet minimum contrast required.

Screen Shot 2021-07-01 at 10 19 07

### DIFF
--- a/src/icons/general/IconEdit.tsx
+++ b/src/icons/general/IconEdit.tsx
@@ -4,7 +4,7 @@ import AccessibleSVG from '../accessible-svg';
 
 import { Icon } from '../iconTypes';
 
-const IconEdit = ({ className = '', color = '#999', height = 14, title, width = 14 }: Icon) => (
+const IconEdit = ({ className = '', color = '#767676', height = 14, title, width = 14 }: Icon) => (
     <AccessibleSVG className={`icon-edit ${className}`} height={height} title={title} viewBox="0 0 14 14" width={width}>
         <path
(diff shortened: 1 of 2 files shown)


src/icons/general/IconEdit.tsx
```diff
@@ -4,7 +4,7 @@ import AccessibleSVG from '../accessible-svg';
 
 import { Icon } from '../iconTypes';
 
-const IconEdit = ({ className = '', color = '#999', height = 14, title, width = 14 }: Icon) => (
+const IconEdit = ({ className = '', color = '#767676', height = 14, title, width = 14 }: Icon) => (
```
laughnan, line 7: I assume we had to change this everywhere? Since it's a prop, we could have just increased it for that one specific section, right?

laughnan: Chatted in person about this. It is something we want to update across the board. I'll be approving.

laughnan (approved):

src/icons/general/IconEdit.tsx
msorianoBox: In fact any lighter gray  in the page over white background should change for $bld-gray-62, in this case I had to put it right on the svg, because its not taking styles from anywhere else :(

PR #2622, opened 2021-07-02
fix(TaskForm): A11Y - Give users the option to enter a date by hand

fsuarez-globant:
When the field receives keyboard focus, a date picker dialog appears. This dialog is moderately accessible to screen reader users, but it would be better to give users the option to enter a date by hand instead. I added `isTextInputAllowed = true` in order to enable this feature. 

Screen Shot 2021-07-02 at 3 40 50 PM

### DIFF
--- a/src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
+++ b/src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
@@ -449,6 +449,7 @@ class TaskForm extends React.Component<Props, State> {
                             }}
                             isDisabled={isForbiddenErrorOnEdit}
                             isRequired={false}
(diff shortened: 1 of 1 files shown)


alexkrolick:
This was disabled because it has issues with internationalized date formats (like 12.19.21)

tille:
@alexkrolick we can change this to 3 different fields in order differentiate day/month/year but I guess it is a big change and maybe out of scope, what do you think?

alexkrolick:
It's a pretty big change; we'd have to detect the locale and put the fields in the right natural order, and it would change the look and feel. Since it's not an immediate issue, can we flag this for followup UX work?

PR #2627, opened 2021-07-09
fix(metadata): Added visual focus to search btns

msorianoBox:
Hi guys!

According to the recommendations made by the team, we introduced some changes in order to highlight the buttons on the search component in the metadata section, as requested by WebAIM

Screen Shot 2021-07-09 at 10 51 00

Screen Shot 2021-07-09 at 10 51 50

### DIFF
--- a/i18n/bn-IN.properties
+++ b/i18n/bn-IN.properties
@@ -123,7 +123,7 @@ be.contentSidebar.activityFeed.commentForm.atMentionTip = @তাদের ব�
 # Text for cancel button
 be.contentSidebar.activityFeed.commentForm.commentCancel = বাতিল করুন
 # Accessible label for comment input field
-be.contentSidebar.activityFeed.commentForm.commentLabel = মন্তব্য
+be.contentSidebar.activityFeed.commentForm.commentLabel = একটি মন্তব্য লিখুন
 # Text for post button
 be.contentSidebar.activityFeed.commentForm.commentPost = পোস্ট করুন
 # Placeholder for comment input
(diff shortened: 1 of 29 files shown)


laughnan (dismissed):

swfree:
@msorianoBox will you be removing "value" in a separate PR, since you removed that commit here (totally fine to do so, so we can test that change separately)? If so, can you remove the line "In order to fix the component buttons behavior, value had to be removed." from your PR message

swfree:
Also @msorianoBox & @laughnan what do you think about making this change after the separate fix for the search icon behavior (to change it from a button to a div if onSubmit is not defined)? Right now, clicking on the search icon doesn't do anything, so outlining it might make that more confusing to users. Just realizing this now, that the order of operations might matter here 🤔

msorianoBox:
I did the adjustment at first, and while removing the Value prop from te component made everything else work, the lint test was breaking up. I went and did a clean up, but then, the filtering from the list stopped working, so I reverted that change.

Basically the focus issue was solved, but then, without the right functionality might look pointless, since there is no action.

Might need a different solution for the toggle buttons.

swfree:
Sounds good @msorianoBox! Yeah we can come up with another approach

For now, can you add another selector for `.clear-button` so the outline just targets the cancel button here? Since there is no action for the search-button, seems like we shouldn't add an outline to it

msorianoBox:
> .clear-button

Done! :D

swfree (approved):

PR #2632, opened 2021-07-13
fix(metadata): Fixed lost focus on search

msorianoBox:
Hi guys!

We had to do some changes on the search component, since focus got out of sight after tabbing  from the input text field. The Search button has some instances where the OnSubmit action is not defined and receives focus nevertheless. the approach to fix this was to make sure that the onSubmit prop was defined in order to render it as a <button>, otherwise it gets rendered as a <div>.

Screen Shot 2021-07-13 at 17 51 25

Renders as a div when is undefined
Screen Shot 2021-07-13 at 17 53 40

This prevents lost focus on the component.

### DIFF
--- a/src/components/search-form/SearchForm.js
+++ b/src/components/search-form/SearchForm.js
@@ -129,7 +129,18 @@ class SearchForm extends React.Component<Props, State> {
     searchInput: ?HTMLInputElement;
 
     render() {
-        const { action, className, intl, isLoading, method, name, queryParams, useClearButton, ...rest } = this.props;
+        const {
+            action,
+            className,
+            intl,
+            isLoading,
+            method,
+            name,
+            queryParams,
+            onSubmit,
+            useClearButton,
+            ...rest
+        } = this.props;
         const { isEmpty } = this.state;
 
         const inputProps = omit(rest, [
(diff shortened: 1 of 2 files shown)


src/components/search-form/SearchForm.js
```diff
@@ -152,13 +163,24 @@ class SearchForm extends React.Component<Props, State> {
...
+                    <button
+                        type="button"
+                        className="action-button search-button"
+                        title={formatMessage(messages.searchButtonTitle)}
+                    >
+                        <Search16 />
+                    </button>
+                ) : (
+                    <div
+                        type="button"
```
jstoffan, line 176: Can we omit `type` and `title`, since this variation is a purely visual treatment?

src/components/search-form/__tests__/SearchForm.test.js
```diff
@@ -21,9 +21,11 @@ describe('components/search-form/SearchForm', () => {
     });
 
     test('should correctly render default component', () => {
-        const wrapper = mount(<SearchForm placeholder="search" />);
+        const onSubmitSpy = sinon.spy();
+        const wrapper = mount(<SearchForm onSubmit={onSubmitSpy} placeholder="search" value="cheese" />);
         expect(wrapper.find('form').length === 1).toBeTruthy();
         expect(wrapper.find('input').length === 1).toBeTruthy();
+
```
jstoffan, line 28: Can we revert this whitespace-only change?

src/components/search-form/__tests__/SearchForm.test.js
```diff
@@ -21,9 +21,11 @@ describe('components/search-form/SearchForm', () => {
     });
 
     test('should correctly render default component', () => {
-        const wrapper = mount(<SearchForm placeholder="search" />);
+        const onSubmitSpy = sinon.spy();
+        const wrapper = mount(<SearchForm onSubmit={onSubmitSpy} placeholder="search" value="cheese" />);
```
jstoffan, line 25: Can we add a test for when `onSubmit` is not defined?

tille:
this PR is deprecated, replaced by https://github.com/box/box-ui-elements/pull/2637

greathmaster:
Closing the PR since it is deprecated and replaced by https://github.com/box/box-ui-elements/pull/2637

PR #2633, opened 2021-07-16
A11y announce results comment field

tille:
Hi guys!

We had to do some changes on the search component, since focus got out of sight after tabbing from the input text field. The Search button has some instances where the OnSubmit action is not defined and receives focus nevertheless. the approach to fix this was to make sure that the onSubmit prop was defined in order to render it as a , otherwise it gets rendered as a

.

image
image

### DIFF
--- a/src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelectorCore.js
+++ b/src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelectorCore.js
@@ -75,6 +75,8 @@ class DraftJSMentionSelector extends React.Component<Props, State> {
         startMentionMessage: <DefaultStartMentionMessage />,
     };
 
+    mentorSelectorRef: { current: null | HTMLDivElement } = React.createRef();
+
     constructor(props: Props) {
         super(props);
         const mentionTriggers = props.mentionTriggers.reduce((prev, current) => `${prev}\\${current}`, '');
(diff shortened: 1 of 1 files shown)


swfree:
Hi @tille this looks like the same change as https://github.com/box/box-ui-elements/pull/2619? Also your PR comment looks like its related to a different PR, not this one

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelectorCore.js
```diff
@@ -229,6 +234,27 @@ class DraftJSMentionSelector extends React.Component<Props, State> {
...
+        const previousElement = document.querySelector('[data-testid="accessibility-alert"]');
+        if (previousElement) {
+            previousElement.remove();
+        }
+        if (addAlert) {
+            alertElement = document.createElement('span');
+            alertElement.setAttribute('class', 'accessibility-hidden');
+            alertElement.setAttribute('data-testid', 'accessibility-alert');
+            alertElement.setAttribute('role', 'alert');
+            alertElement.innerText = contacts.length > 0 ? `${contacts.length} users found` : 'No users found';
```
swfree, line 251: DraftJsEditor is a generic editor. These alerts look more specific to [CommentForm](https://github.com/box/box-ui-elements/blob/master/src/elements/content-sidebar/activity-feed/comment-form/CommentForm.js).

swfree: On a different note - can you internationalize this text?

tille:
yes it's a clone of #2619 since Ferney is out to maintain its branch

PR #2637, opened 2021-07-19
fix(metadata): Fixed lost focus on search

tille:
Hi guys!

We had to do some changes on the search component, since focus got out of sight after tabbing from the input text field. The Search button has some instances where the OnSubmit action is not defined and receives focus nevertheless. the approach to fix this was to make sure that the onSubmit prop was defined in order to render it as a , otherwise it gets rendered as a

.

image

Renders as a div when is undefined
image
This prevents lost focus on the component.


jasminevle:
Question: are we ever passing onSubmit in from a parent component?

tille:
@jasminevle yes there are other places in the code where such prop is passed in

jasminevle (approved):

PR #2639, opened 2021-07-21
fix(comments): Added Aria attributes to enhance accessibility

msorianoBox:
According to the communication held with WebAim, the role combobox had to be switched to textbox and the aria-multiline attribute had to be added.

Screen Shot 2021-07-21 at 10 35 37

### DIFF
--- a/src/components/draft-js-editor/DraftJSEditor.js
+++ b/src/components/draft-js-editor/DraftJSEditor.js
@@ -111,6 +111,7 @@ class DraftJSEditor extends React.Component<Props> {
                 ariaAutoComplete: inputProps['aria-autocomplete'],
                 ariaExpanded: inputProps['aria-expanded'],
                 ariaOwneeID: inputProps['aria-owns'],
+                ariaMultiline: true,
                 role: 'textbox',
             };
         }
@@ -130,7 +131,6 @@ class DraftJSEditor extends React.Component<Props> {
                     <div>
                         <Editor
                             {...a11yProps}
-                            ariaMultiline
                             ariaLabelledBy={this.labelID}
                             ariaDescribedBy={this.descriptionID}
                             editorState={editorState}
--- a/src/components/draft-js-editor/DraftJSEditor.js
+++ b/src/components/draft-js-editor/DraftJSEditor.js
@@ -111,7 +111,7 @@ class DraftJSEditor extends React.Component<Props> {
                 ariaAutoComplete: inputProps['aria-autocomplete'],
                 ariaExpanded: inputProps['aria-expanded'],
                 ariaOwneeID: inputProps['aria-owns'],
-                role: inputProps.role,
+                role: inputProps.textbox,
             };
         }
 
@@ -130,6 +130,7 @@ class DraftJSEditor extends React.Component<Props> {
                     <div>
                         <Editor
                             {...a11yProps}
+                            ariaMultiline
                             ariaLabelledBy={this.labelID}
                             ariaDescribedBy={this.descriptionID}
                             editorState={editorState}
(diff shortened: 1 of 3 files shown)


src/components/draft-js-editor/DraftJSEditor.js
```diff
@@ -111,7 +111,7 @@ class DraftJSEditor extends React.Component<Props> {
                 ariaAutoComplete: inputProps['aria-autocomplete'],
                 ariaExpanded: inputProps['aria-expanded'],
                 ariaOwneeID: inputProps['aria-owns'],
-                role: inputProps.role,
+                role: inputProps.textbox,
```
jstoffan, line 114: This change will set the `role` to the value of the `inputProps.textbox` prop, which appears to be undefined for most usages. Can we instead update the calling components to pass in the proper `role` prop?

src/components/draft-js-editor/__tests__/DraftJSEditor.test.js
```diff
@@ -75,7 +75,7 @@ describe('components/draft-js-editor/DraftJSEditor', () => {
             expect(editor.prop('ariaAutoComplete')).toEqual(inputProps['aria-autocomplete']);
             expect(editor.prop('ariaExpanded')).toEqual(inputProps['aria-expanded']);
             expect(editor.prop('ariaOwneeID')).toEqual(inputProps['aria-owns']);
-            expect(editor.prop('role')).toEqual(inputProps.role);
+            expect(editor.prop('role')).toEqual(inputProps.textbox);
```
jstoffan, line 78: It appears that `inputProps.textbox` is undefined here. Can we hardcode the expected result and see if the tests passes?

src/components/draft-js-editor/DraftJSEditor.js
msorianoBox: Done!

laughnan: @msorianoBox I think what is confusing here to me is that we have `<DraftJSEditor>` (component) which has a prop called `inputProps.role` but we're now ignoring that prop and using `inputProps.textbox`? So the calling components will go from
```
<DraftJSEditor inputProps={{role: 'textbox'}} />
```
to
```
<DraftJSEditor inputProps={{textbox: 'textbox'}} />
```

Is this expected? cc: @jstoffan

laughnan: If you want to ignore `inputProps.role` and **always** force it to `textbox` then wouldn't this work:

```
...
role: 'textbox',
```

src/components/draft-js-editor/__tests__/DraftJSEditor.test.js
```diff
@@ -66,7 +66,7 @@ describe('components/draft-js-editor/DraftJSEditor', () => {
                 'aria-autocomplete': 'list',
                 'aria-expanded': true,
                 'aria-owns': 'id',
-                role: 'combobox',
+                role: 'textbox',
```
laughnan, line 69: So we're still using `inputProps.textbox` which I don't think is what we're referencing above: https://github.com/box/box-ui-elements/pull/2639/files#diff-e198f1f72d6edc0d491d8f2b2c7050db6bf462561b0e533630439673f064717aR114

laughnan: Also, the `a11yProps` are wrapped in an `if` condition if the `inputProps.role` is set: https://github.com/box/box-ui-elements/blob/master/src/components/draft-js-editor/DraftJSEditor.js#L108-L116

Since we no longer care/require that prop, should we remove the conditional?

laughnan (approved):

PR #2640, opened 2021-07-21
fix(metadata): A11Y - Autofocus on dialog when delete metadata

hormaza:
image

When the "Remove" button is clicked, the confirmation dialog ask to the user if sure to delete the metadata. 
For voiceOver this text is not accesible and focus is set to body.
This change will focus to confirmation dialog in order to be read for people who is using voiceover.

It should be accesible via tab button now:
image


src/features/metadata-instance-editor/MetadataInstanceConfirmDialog.js
```diff
@@ -15,20 +15,33 @@ type Props = {
...
+        if (confirmationMessageElement.current) {
+            confirmationMessageElement.current.focus();
+        }
+    });
+
+    return (
+        <div className="metadata-instance-confirm-cover">
+            <div className="metadata-instance-confim-container">
+                {/* eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex */}
+                <div tabIndex="0" ref={confirmationMessageElement}>
```
ConradJChan, line 31: This seems like an unideal solution. Would using something like the `role="alert"` work instead to get the screen reader to read the text without explicitly providing focus?

hormaza: Hi. We tired role='alert' but linter still fail. Do you have any other suggestion?

(Linter says non-interactive elements can't have tab-index property.)

hormaza:
snapshot updated

src/features/metadata-instance-editor/MetadataInstanceConfirmDialog.js
ConradJChan: I was suggesting that instead of the `tabIndex` approach, whether providing the `role="alert"` on the `MetadataInstanceConfirmDialog` would solve the voiceover issue when it mounts?

alexkrolick: +1 to @ConradJChan, adding aria-role="alert" would create a live region. The autofocus, if any, should probably go to the Cancel button.

hormaza: We need to have ​autofocus on the message of `metadata-instance-confim-container` in order to VoiceOver read this section.
role="alert" does not solve the problem of VoicerOver. TabIndex does.

hormaza: @alexkrolick we need to focus the message before focus the cancel button. Otherwise VoiceOver will read Cancel but no message

ConradJChan: Maybe try `role="alert"` instead? I put together a codesandbox that seems to work as expected: https://codesandbox.io/s/confirm-dialog-a11y-29ze3

hormaza: Thanks clarify this. It works!

src/features/metadata-instance-editor/MetadataInstanceConfirmDialog.js
```diff
@@ -15,20 +15,37 @@ type Props = {
...
+    React.useEffect(() => {
+        if (cancelButtonRef.current) {
+            cancelButtonRef.current.btnElement.focus();
+        }
+    }, []);
+
+    return (
+        <div className="metadata-instance-confirm-cover">
+            <div className="metadata-instance-confim-container">
+                <div role="alert">
```
ConradJChan, line 30: Does the `role` need to be added here or can it be added to the topmost `div`? Seeing if we can do away with this extra `div` that was added here

hormaza: Yes. Done!

ConradJChan (approved):

swfree (approved):

PR #2645, opened 2021-07-27
fix(DraftJSMentionSelectorCore): Dynamically removed and reapplied alert Mention feature

tille:
In order to make multiple announcements using the same alert element, the role attribute must be dynamically removed and reapplied every time an alert is made. I used of ARIA alerts to inform screen reader users of how many results have been found (e.g., "3 users found", "No users found", etc.)

No user found

image

image


src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelectorCore.js
```diff
@@ -229,6 +234,27 @@ class DraftJSMentionSelector extends React.Component<Props, State> {
         return !!(activeMention && activeMention.mentionString && contacts.length);
     };
 
+    buildAccessibleAlert = (activeMention: Object) => {
```
ConradJChan, line 237: Should we consider a React approach instead of manually manipulating the DOM directly?

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelectorCore.js
```diff
@@ -229,6 +234,27 @@ class DraftJSMentionSelector extends React.Component<Props, State> {
...
+        const previousElement = document.querySelector('[data-testid="accessibility-alert"]');
+        if (previousElement) {
+            previousElement.remove();
+        }
+        if (addAlert) {
+            alertElement = document.createElement('span');
+            alertElement.setAttribute('class', 'accessibility-hidden');
+            alertElement.setAttribute('data-testid', 'accessibility-alert');
+            alertElement.setAttribute('role', 'alert');
+            alertElement.innerText = contacts.length > 0 ? `${contacts.length} users found` : 'No users found';
```
ConradJChan, line 251: We should also internationalize this message

ConradJChan:
Should we add tests for this functionality?

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelectorCore.js
tille: Can you recommend me a react way to attack this?

tille: I guess it is the only way to do such a particular change

ConradJChan: Seems like what you're building on L247 could be an `AccessibilityAlert` component that takes some i18n message. 

To make it re-usable, maybe you could hide and show via CSS, or maybe play with the `key` React prop to make it unique each time it renders

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelectorCore.js
```diff
@@ -75,12 +82,15 @@ class DraftJSMentionSelector extends React.Component<Props, State> {
         startMentionMessage: <DefaultStartMentionMessage />,
     };
 
+    mentorSelectorRef: { current: null | HTMLDivElement } = React.createRef();
```
ConradJChan, line 85: Is this still needed?

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelectorCore.js
```diff
@@ -31,11 +31,18 @@ const DefaultSelectorRow = ({ item = {}, ...rest }: DefaultSelectorRowProps) =>
...
 const DefaultStartMentionMessage = () => <FormattedMessage {...messages.startMention} />;
 
-type MentionStartStateProps = {
+type MentionStateProps = {
+    contacts?: Number,
     message?: React.Node,
 };
 
-const MentionStartState = ({ message }: MentionStartStateProps) => <div className="mention-start-state">{message}</div>;
+const MentionStartState = ({ message }: MentionStateProps) => <div className="mention-start-state">{message}</div>;
```
ConradJChan, line 39: We typically like to keep a single component per file. That being said, do we even need `MentionStartState` as a separate component? Seems like that could be inlined on L303. Similarly for the `MentionUsersFound`

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelectorCore.js
```diff
@@ -31,11 +31,18 @@ const DefaultSelectorRow = ({ item = {}, ...rest }: DefaultSelectorRowProps) =>
...
+
+const MentionUsersFound = ({ contacts }: MentionStateProps) => (
+    <span className="accessibility-hidden" data-testid="accessibility-alert" role="alert">
+        {contacts > 0 ? `${contacts} users found` : 'No users found'}
```
ConradJChan, line 43: Do we need to internationalize this message?

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelectorCore.js
```diff
@@ -285,6 +301,7 @@ class DraftJSMentionSelector extends React.Component<Props, State> {
                         : []}
                 </SelectorDropdown>
                 {showMentionStartState ? <MentionStartState message={startMentionMessage} /> : null}
+                {announceActiveMention ? <MentionUsersFound /> : null}
```
ConradJChan, line 304: Seems like these could be simplified to `announceActiveMention && ...JSX...` rather than the ternary operator

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelectorCore.js
```diff
@@ -180,9 +190,15 @@ class DraftJSMentionSelector extends React.Component<Props, State> {
      */
     handleChange = (nextEditorState: EditorState) => {
         const { onChange } = this.props;
-
+        const { isFocused } = this.state;
         const activeMention = this.getActiveMentionForEditorState(nextEditorState);
 
+        if (activeMention && activeMention.mentionString?.length > 0 && isFocused) {
+            this.setState({
+                announceActiveMention: true,
```
ConradJChan, line 198: It seems like `announceActiveMention` is never set to `false`. 

Also, does it trigger the screen reader to update the message when the number of contacts changes?

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelectorCore.js
ConradJChan: Should `contacts` be passed here?

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelectorCore.js
```diff
@@ -197,6 +202,11 @@ class DraftJSMentionSelector extends React.Component<Props, State> {
                 }
             },
         );
+
+        if (activeMention && activeMention.mentionString?.length > 0 && isFocused) {
```
ConradJChan, line 206: Instead of using a `setTimeout` to hide the a11y alert after showing it, would utilizing a unique for the alert span work? That way we force React to render a new DOM element instead of reusing it which should help solve the screen reader issue

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelectorCore.js
```diff
@@ -285,6 +295,15 @@ class DraftJSMentionSelector extends React.Component<Props, State> {
                         : []}
                 </SelectorDropdown>
                 {showMentionStartState ? <MentionStartState message={startMentionMessage} /> : null}
+                {announceActiveMention && (
+                    <span className="accessibility-hidden" data-testid="accessibility-alert" role="alert">
+                        {this.shouldDisplayMentionLookup() ? (
+                            <UsersFoundMessage contactsCount={contacts.length} />
+                        ) : (
+                            <NoUsersFoundMessage />
```
ConradJChan, line 303: Are these thin wrappers around `FormattedMessage` needed? Could we simplify by using it directly? Maybe something like
```
const message = this.shouldDisplayMentionLookup() ? {...messages.usersFound, values: count: contacts.length} : messages.noUsersFound;
<FormattedMessage {...message} />
```

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelectorCore.js
```diff
@@ -285,6 +295,15 @@ class DraftJSMentionSelector extends React.Component<Props, State> {
                         : []}
                 </SelectorDropdown>
                 {showMentionStartState ? <MentionStartState message={startMentionMessage} /> : null}
+                {announceActiveMention && (
```
ConradJChan, line 298: Should this be `announceActiveMention && isFocused`?

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelectorCore.js
tille: sorry I don't get it

tille: how would be the unique alert span syntax
`<span id=unique('someText')>` ?

tille: I guess it will not work since we have to wait for some miliseconds before set the value in false
I mean if I set 

`setTimeout(callback, 0); // this fail` 
`setTimeout(callback, 100); // this works`

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelectorCore.js
tille: not required

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelectorCore.js
tille: done

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelectorCore.js
ConradJChan: Sorry, I mis-typed, utilizing a unique `key` on the element, something like this:
https://codesandbox.io/s/a11y-alert-react-key-u04u4

i18n/en-US.properties
```diff
@@ -876,8 +876,12 @@ boxui.datePicker.chooseDate = Choose Date
 boxui.datePicker.dateClearButton = Clear Date
 # Date entered is invalid
 boxui.datePicker.iconAlertText = Invalid Date
+# Message for screenReader users when trying to mention a user but there are no matches
+boxui.draftjs.mentionSelector.noUsersFound = No users found
 # Message to display when a user triggers a mention
 boxui.draftjs.mentionSelector.startMention = Mention someone to notify them
+# Message for screenReader users when a certain number of users is being mentioned
+boxui.draftjs.mentionSelector.usersFound = {usersFound} users found
```
ConradJChan, line 884: Should we account for single vs. plural in this message?

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelectorCore.js
```diff
@@ -35,11 +35,16 @@ type MentionStartStateProps = {
     message?: React.Node,
 };
 
-const MentionStartState = ({ message }: MentionStartStateProps) => <div className="mention-start-state">{message}</div>;
+const MentionStartState = ({ message }: MentionStartStateProps) => (
+    <div role="alert" className="mention-start-state">
+        {message}
+    </div>
```
ConradJChan, line 41: Would it make sense to have a single alert component that just takes a message that could be used for mention start as well as when there are active mentions listed?

ConradJChan (commented):
I think with the unique key approach we wouldn't need to introduce the new `contactsLoaded` state

i18n/en-US.properties
tille: I don't think it is required "1 users found", is ok for screenReader users

src/components/form-elements/draft-js-mention-selector/DraftJSMentionSelectorCore.js
tille: because we are running out of time... I guess a refactor for the other messages and components can be worked in a different ticket

i18n/en-US.properties
ConradJChan: Shouldn't be a hard fix, see https://github.com/box/box-ui-elements/blob/830bf71bef5720a5e6ba128afa35152112674c17/src/elements/content-sidebar/versions/messages.js#L115 for an example

jasminevle (dismissed):

ConradJChan (approved):

PR #2646, opened 2021-07-28
fix: tooltip text for shield access restriction

dnsosebee:
This is part of a series of changes to unify tooltip wording for shield access restrictions.

The old tooltip:
image

The new tooltip:
Screen Shot 2021-07-27 at 12 45 54 PM


mickr:
Looks good to me

mickr (approved):

PR #2651, opened 2021-07-29
fix(tooltip): remove button role for targeted click through

wlyhdg:


JohnnyB83 (approved):
looks good, lets just double check this doesn't have any weird side effects for the tooltips

PR #2660, opened 2021-08-13
fix(forms): highlight properly radio and checkboxes

msorianoBox:


msorianoBox:
Screen Shot 2021-08-13 at 10 29 31
Focus highlight used to be kind of dim on the checkboxes and radio inputs and was reported as an accessibility issue.

Screen Shot 2021-08-13 at 10 28 35
Added a little bit more of border thickness in order to highlight better.

mickr (approved):

PR #2661, opened 2021-08-17
fix: error screen when preview is blocked by access policy

dnsosebee:
Fixes an issue where the errorType of preview errors wasn't being passed through. Now, the error screen looks like this when a user is blocked by an access policy:
Screen Shot 2021-08-09 at 1 09 01 PM

### DIFF
--- a/scripts/styleguide.config.js
+++ b/scripts/styleguide.config.js
@@ -16,6 +16,7 @@ const allSections = [
             '../src/elements/content-open-with/ContentOpenWith.js',
             '../src/elements/content-picker/ContentPicker.js',
             '../src/elements/content-preview/ContentPreview.js',
+            '../src/elements/content-sharing/ContentSharing.js',
             '../src/elements/content-sidebar/ContentSidebar.js',
             '../src/elements/content-uploader/ContentUploader.js',
         ],
@@ -85,6 +86,7 @@ const allSections = [
 
             '../src/components/avatar/Avatar.tsx',
             '../src/components/badgeable/Badgeable.tsx',
+            '../src/components/category-selector/CategorySelector.tsx',
             '../src/components/label/Label.tsx',
             '../src/components/button/Button.tsx',
             '../src/components/button-group/ButtonGroup.tsx',
@@ -101,6 +103,7 @@ const allSections = [
             '../src/components/radio/RadioButton.tsx',
             '../src/components/radio/RadioGroup.tsx',
             '../src/components/menu/SelectMenuLinkItem.tsx',
+            '../src/components/time-input/TimeInput.tsx',
             '../src/components/tooltip/Tooltip.tsx',
         ],
         description: 'Box UI Elements components implement the reusable building blocks of the Box Design Language',
@@ -223,10 +226,13 @@ const allSections = [
     {
         name: 'Features',
         components: () => [
+            '../src/features/beta-feedback/BetaFeedbackBadge.js',
             '../src/features/classification/Classification.js',
+            '../src/features/collapsible-sidebar/CollapsibleSidebarLogo.js',
             '../src/features/invite-collaborators-modal/InviteCollaboratorsModal.js',
             '../src/features/left-sidebar/LeftSidebar.js',
             '../src/features/header-flyout/HeaderFlyout.js',
+            '../src/features/message-center/components/MessageCenter.js',
             '../src/features/presence/Presence.js',
             '../src/features/presence/PresenceLink.js',
             '../src/features/security-cloud-game/SecurityCloudGame.js',
@@ -235,6 +241,7 @@ const allSections = [
             '../src/features/shared-link-modal/SharedLinkModal.js',
             '../src/features/shared-link-settings-modal/SharedLinkSettingsModal.js',
             '../src/features/unified-share-modal/UnifiedShareModal.js',
+            '../src/features/virtualized-table-renderers/FormattedUser.js',
         ],
         sections: [
             {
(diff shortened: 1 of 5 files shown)


scripts/styleguide.config.js
```diff
@@ -16,6 +16,7 @@ const allSections = [
             '../src/elements/content-open-with/ContentOpenWith.js',
             '../src/elements/content-picker/ContentPicker.js',
             '../src/elements/content-preview/ContentPreview.js',
+            '../src/elements/content-sharing/ContentSharing.js',
```
jstoffan, line 19: These changes seem unrelated to the bug fix. Can we move them to a separate PR?

src/elements/content-preview/PreviewMask.tsx
```diff
@@ -17,7 +17,7 @@ export default function PreviewMask({ errorCode, extension, isLoading }: Props):
 
     return (
         <div className="bcpr-PreviewMask">
-            {errorCode ? <PreviewError /> : isLoading && <PreviewLoading extension={extension} />}
+            {errorCode ? <PreviewError errorCode={errorCode} /> : isLoading && <PreviewLoading extension={extension} />}
```
jstoffan, line 20: Whoops, this is my bad! Thanks for correcting this behavior.

scripts/styleguide.config.js
feixuanli: Is this fix still needed after upgrading node? @dnsosebee

src/elements/content-preview/PreviewMask.tsx
```diff
@@ -17,7 +17,7 @@ export default function PreviewMask({ errorCode, extension, isLoading }: Props):
...
         </div>
```
feixuanli, line 21: Can we update test for this ?

dnsosebee: I believe updating the snapshot is sufficient, since there's already two tests for preview error, one of which deals with the 'forbidden_by_policy' error:
```
test('should render correctly', () => {
            const wrapper = getWrapper();
        });

        test('should render correctly when blocked by access policy', () => {
            const wrapper = getWrapper({ errorCode: ERROR_CODE_FETCH_FILE_DUE_TO_POLICY });
        });
```

scripts/styleguide.config.js
dnsosebee: I can run the tests successfully without changing the styleguide, however, if I run `yarn start`, or `yarn build`, it removes all of the i18n related to the classes I added here. And I don't want to push my changes with a bunch of unneeded deletes from i18n.
Does this make sense @jstoffan?
Anyways my understanding is I could move this into another PR, but I would need it to be approved first and to pull those changes in before pushing the errorScreen changes.
Let me know if any of my assumptions here are wrong, and let me know if you still think I should separate this into another PR.

dnsosebee: Here's what happens when I run `yarn start` on the upstream head, using node 12.13.0:
```
diff --git a/i18n/en-US.properties b/i18n/en-US.properties
index 3fe5652b..c897fa60 100644
--- a/i18n/en-US.properties
+++ b/i18n/en-US.properties
@@ -62,28 +62,6 @@ be.choose = Choose
 be.close = Close
 # Message to the user to collapse the Transcript entries
 be.collapse = Collapse
-# Message that appears when the request for the ContentSharing Element is malformed.
-be.contentSharing.badRequestError = The request for this item was malformed.
-# Message that appears when collaborators cannot be retrieved in the ContentSharing Element.
-be.contentSharing.collaboratorsLoadingError = Could not retrieve collaborators for this item.
-# Message that appears when users cannot be retrieved in the ContentSharing Element.
-be.contentSharing.getContactsError = Could not retrieve contacts.
-# Message that appears when the ContentSharing Element cannot be loaded.
-be.contentSharing.loadingError = Could not load shared link for this item.
-# Message that appears when the user cannot access the item for the ContentSharing Element.
-be.contentSharing.noAccessError = You do not have access to this item.
-# Message that appears when the item for the ContentSharing Element cannot be found.
-be.contentSharing.notFoundError = Could not find shared link for this item.
-# Message that appears when collaborators cannot be added to the shared link in the ContentSharing Element.
-be.contentSharing.sendInvitesError = Could not send invites.
-# Message that appears when collaborators were added to the shared link in the ContentSharing Element.
-be.contentSharing.sendInvitesSuccess = Successfully invited collaborators.
-# Message that appears when the shared link in the ContentSharing Element was removed.
-be.contentSharing.sharedLinkRemovalSuccess = The shared link for this item was removed.
-# Message that appears when the shared link settings in the ContentSharing Element were successfully updated.
-be.contentSharing.sharedLinkSettingsUpdateSuccess = The shared link for this item was successfully updated.
-# Message that appears when the shared link in the ContentSharing Element cannot be updated.
-be.contentSharing.sharedLinkUpdateError = Could not update the shared link for this item.
 # Text to show on menu item to delete the annotation comment
 be.contentSidebar.activityFeed.annotationActivity.annotationActivityDeleteMenuItem = Delete
 # Confirmation prompt text to delete the annotation comment
@@ -326,10 +304,6 @@ be.executeIntegrationOpenWithErrorSubHeader = Please try again later.
 be.expand = Expand
 # Label for face skill section in the preview sidebar
 be.faceSkill = Faces
-# Call-to-action text describing what to do to navigate to specified feedback form
-be.feedbackCtaText = Click to provide feedback
-# Accessible text used to describe the form used for feedback
-be.feedbackFormDescription = Beta Feedback Form
 # File access stats error message
 be.fileAccessStatsErrorHeaderMessage = Something went wrong when fetching the access stats.
 # File classification error message
@@ -380,22 +354,6 @@ be.loadingState = Please wait while the items load...

```

jstoffan: Hi @dnsosebee. I'm not sure why the styleguide config would affect i18n, but I still think it would be best to have a separate PR for these changes. That way, the fix code stays atomic. I'll be happy to approve such a PR once it's open.

PR #2666, opened 2021-08-18
feat(shared-link-section): add warning copy for elevated edit permission

ChenCodes:
**Changes in this PR:**
- [x] when the user changes the permissionLevel from something other than `CAN_EDIT` to `CAN_EDIT`, then we should display a special message just for that case

**Demo:**


src/features/unified-share-modal/SharedLinkSection.js
```diff
@@ -138,6 +147,14 @@ class SharedLinkSection extends React.Component<Props, State> {
             }
         }
 
+        if (
+            prevProps.sharedLink.permissionLevel !== '' &&
```
ChenCodes, line 151: Added an empty string check here because by default permissionLevel is always empty string when a sharedLink is first instantiated in the parent app. We don't want to prematurely set `isPermissionElevatedToEdit`.

src/features/unified-share-modal/SharedLinkSection.js
```diff
@@ -335,6 +352,17 @@ class SharedLinkSection extends React.Component<Props, State> {
                         )}
                     </div>
                 )}
+                {accessLevel === ANYONE_IN_COMPANY && isPermissionElevatedToEdit && (
```
ChenCodes, line 355: This warning copy is only for the special case where accessLevel is `ANYONE_IN_COMPANY` and when the permissionLevel has been elevated to Edit from another permission level.

src/features/unified-share-modal/SharedLinkSection.js
```diff
@@ -138,6 +147,21 @@ class SharedLinkSection extends React.Component<Props, State> {
...
+            prevProps.sharedLink.permissionLevel !== '' &&
+            prevProps.sharedLink.permissionLevel !== CAN_EDIT &&
+            sharedLink.permissionLevel === CAN_EDIT
+        ) {
+            this.setState({ isPermissionElevatedToEdit: true });
+        }
+
+        if (
+            isPermissionElevatedToEdit &&
+            (sharedLink.permissionLevel !== CAN_EDIT || sharedLink.accessLevel !== ANYONE_IN_COMPANY)
```
ChenCodes, line 160: We need this check as well to account for the following scenarios:

Given that the current shared link has the following:
accessLevel -> ANYONE_IN_COMPANY
permissionLevel -> CAN_EDIT

If the user then decides to update permissionLevel to be something other than `CAN_EDIT`,
then we should set `isPermissionElevatedToEdit: false`.


If the user then decides to update accessLevel to be something other than `ANYONE_IN_COMPANY`
then we should set `isPermissionElevatedToEdit: false`.

The goal behind this is that it acts as a sort of "cleanup" - we don't want the component to be in a stale state where `isPermissionElevatedToEdit` is still true even though the user has updated either the accessLevel or permissionLevel to other values.

src/features/unified-share-modal/__tests__/SharedLinkSection.test.js
```diff
@@ -297,6 +297,78 @@ describe('features/unified-share-modal/SharedLinkSection', () => {
...
+            const sharedLink = {
+                accessLevel: ANYONE_IN_COMPANY,
+                url: 'http://example.com/',
+                isNewSharedLink: false,
+                permissionLevel: CAN_VIEW_DOWNLOAD,
+            };
+
+            const wrapper = getWrapper({ sharedLink });
+
+            expect(wrapper.state().isPermissionElevatedToEdit).toBe(false);
```
mickr, line 310: Should there be introspection into the state here or can we just test the outcome?

mickr (approved):

PR #2668, opened 2021-08-24
fix(buttons): highlighting button properly on focus

msorianoBox:
Buttons didn't receive proper focus behaviour, and looked
image

In order to highlight correctly I used the hover behavior so the focus outline is noticeable
image

### DIFF
--- a/src/styles/common/_buttons.scss
+++ b/src/styles/common/_buttons.scss
@@ -121,7 +121,8 @@ button svg {
     &:not(.bdl-is-disabled),
     &:not(.is-disabled) {
         &:focus {
-            border-color: $primary-color;
+            background-color: lighten($primary-color, 8%);
+            border: 2px solid $primary-color;
             box-shadow: inset 0 0 0 1px fade-out($white, .2), 0 1px 2px fade-out($black, .9);
         }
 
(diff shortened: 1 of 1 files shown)


src/styles/common/_buttons.scss
```diff
@@ -121,7 +121,8 @@ button svg {
     &:not(.bdl-is-disabled),
     &:not(.is-disabled) {
         &:focus {
-            border-color: $primary-color;
+            background-color: lighten($primary-color, 8%);
+            border: 2px solid $primary-color;
```
alexkrolick, line 125: If the border width changes on focus, won't it cause layout shifts?

msorianoBox: I'll switch it back to 1px :D

alexkrolick: Even with 1px, the default is 0, so it will still shift.

msorianoBox:
Buttons didn't receive proper focus behaviour, and looked 
Screen Shot 2021-08-24 at 13 06 21

In order to highlight correctly I used the hover behavior so the focus outline is noticeable
Screen Shot 2021-08-24 at 13 06 36

jasminevle (approved):

PR #2669, opened 2021-08-26
feat(shared-link-permission-menu): add ftux for shared link permission menu

ChenCodes:
**Changes in this PR:**
- [x] pass `sharedLinkEditTagTargetingApi` down from parent app that helps indicate when the `New` LabelPillText should be visible within the SharedLinkPermissionMenu component

**Demo:**
Screen Shot 2021-09-01 at 1 53 20 PM

### DIFF
--- a/src/features/unified-share-modal/SharedLinkPermissionMenu.scss
+++ b/src/features/unified-share-modal/SharedLinkPermissionMenu.scss
@@ -1,0 +1,10 @@
+.ums-share-permissions-menu {
+    .ums-share-permissions-menu-item {
+        display: flex;
+        align-items: center;
+
+        .ftux-editable-shared-link {
+            margin-left: 10px;
+        }
+    }
+}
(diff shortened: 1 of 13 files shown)


src/features/unified-share-modal/SharedLinkPermissionMenu.scss
```diff
@@ -0,0 +1,10 @@
+.ums-share-permissions-menu {
+    .ums-share-permissions-menu-item {
+        display: flex;
+        align-items: center;
+
+        .ftux-editable-shared-link {
+            margin-left: 10px;
```
mickr, line 7: Should this only be for the `ftux-editable-shared-link` or less specific for other ftux enhancements?

mickr (approved):

src/features/unified-share-modal/SharedLinkPermissionMenu.scss
ChenCodes: I think this is only for the `ftux-editable-shared-link` use case - for other ftux enhancements there will probably be a specific selector to apply that styling.

PR #2674, opened 2021-09-08
feat(shared-link-section): add tooltip ftux

ChenCodes:
**Changes in this PR:**
- [x] add GuideTooltip in SharedLinkSection that will show up if the targetingApi indicates that `canShow` is set to true
- [x] the tooltip will be hidden when the tooltip is dismissed

**Demo:**
Screen Shot 2021-09-08 at 10 34 36 AM

### DIFF
--- a/src/features/unified-share-modal/SharedLinkSection.js
+++ b/src/features/unified-share-modal/SharedLinkSection.js
@@ -5,6 +5,7 @@ import { FormattedMessage } from 'react-intl';
 
 import PlainButton from '../../components/plain-button';
 import Button from '../../components/button';
+import GuideTooltip from '../../components/guide-tooltip';
 import TextInputWithCopyButton from '../../components/text-input-with-copy-button';
 import Toggle from '../../components/toggle/Toggle';
 import Tooltip from '../../components/tooltip';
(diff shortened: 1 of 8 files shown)


shahzadaziz (approved):
:shipit:

src/features/unified-share-modal/SharedLinkSection.js
```diff
@@ -107,6 +115,12 @@ class SharedLinkSection extends React.Component<Props, State> {
             this.setState({ isAutoCreatingSharedLink: true });
             addSharedLink();
         }
+
+        // if ESL ftux tooltip is showing on initial mount, we call onShow
+        if (sharedLinkEditTooltipTargetingApi && sharedLinkEditTooltipTargetingApi.canShow) {
```
megansmith-box, line 120: I think you need to check also `allowedPermissionLevels.includes(CAN_EDIT)` here too. Otherwise, it may lead to onShow being called, when it actually cannot show.

ChenCodes: Ah good catch! Will put in a fix PR for this, thanks for being thorough and finding this

PR #2677, opened 2021-09-09
fix(metadata): add ariaLabel to button

hormaza:
The edit button should have an attribute aria-label instead an attribute area-describedby.
And the value of aria-label should be the same as the Tooltip which is wrapping the button.

image

### DIFF
--- a/src/features/metadata-instance-editor/Instance.js
+++ b/src/features/metadata-instance-editor/Instance.js
@@ -1,6 +1,6 @@
 // @flow
 import * as React from 'react';
-import { FormattedMessage } from 'react-intl';
+import { injectIntl, FormattedMessage } from 'react-intl';
 import classNames from 'classnames';
 import isEqual from 'lodash/isEqual';
 import cloneDeep from 'lodash/cloneDeep';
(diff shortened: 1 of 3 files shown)


src/features/metadata-instance-editor/Instance.js
```diff
@@ -549,7 +550,7 @@ class Instance extends React.PureComponent<Props, State> {
     }
 
     renderEditButton = () => {
-        const { isDirty }: Props = this.props;
+        const { isDirty, intl }: Props = this.props;
```
ConradJChan, line 553: minor: alphabetize props

src/features/metadata-instance-editor/Instance.js
```diff
@@ -558,13 +559,15 @@ class Instance extends React.PureComponent<Props, State> {
...
+            const metadataLabelEditText = intl.formatMessage(messages.metadataEditTooltip);
             return (
-                <Tooltip position="top-left" text={<FormattedMessage {...messages.metadataEditTooltip} />}>
+                <Tooltip position="top-left" text={metadataLabelEditText}>
                     <PlainButton
                         className={editClassName}
                         data-resin-target="metadata-instanceedit"
                         onClick={this.toggleIsEditing}
                         type="button"
+                        aria-label={metadataLabelEditText}
```
ConradJChan, line 570: minor: alphabetize props -- I'm surprised the linter doesn't flag this

ConradJChan (approved):

swfree (approved):

PR #2680, opened 2021-09-10
fix(itemicon): Hide preview thumbnail for screen readers

tille:
The image links in each entry lack any alternative text or text content, making it very difficult for screen reader users to determine the purpose of the link.

for that reason we are making them `presentation` icons and not focusable (since do not have any content to read out)

Screen Shot 2021-09-10 at 5 33 02 PM

### DIFF
--- a/src/icons/item-icon/ItemIcon.tsx
+++ b/src/icons/item-icon/ItemIcon.tsx
@@ -81,7 +81,16 @@ export interface ItemIconProps {
 
 const ItemIcon = ({ className, dimension = 32, iconType, title }: ItemIconProps) => {
     const IconComponent = (itemIconTable as { [key: string]: React.FC<SVGProps> })[iconType] || IconFileDefault;
-    return <IconComponent className={className} height={dimension} title={title} width={dimension} />;
+    return (
+        <IconComponent
+            className={className}
+            height={dimension}
+            title={title}
+            width={dimension}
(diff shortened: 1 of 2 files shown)


src/icons/item-icon/ItemIcon.tsx
```diff
@@ -81,7 +81,16 @@ export interface ItemIconProps {
...
     const IconComponent = (itemIconTable as { [key: string]: React.FC<SVGProps> })[iconType] || IconFileDefault;
-    return <IconComponent className={className} height={dimension} title={title} width={dimension} />;
+    return (
+        <IconComponent
+            className={className}
+            height={dimension}
+            title={title}
+            width={dimension}
+            aria-hidden="true"
+            role="presentation"
```
ConradJChan, line 91: Does the `ItemIcon` component always need to have these props or should they be passed in based on use?

tille: yes `ItemIcon` always need to have those properties

tille: alternative is to identify the type of file and then set an `alt` property based on that
but I guess is better to just hide it for screen reader users

tille: https://github.com/box/box-ui-elements/pull/2686

PR #2681, opened 2021-09-13
feat(shared-link-permission-menu): add check for esl ftux

ChenCodes:
**Changes in this PR:** 
- [x] call onShow for `sharedLinkEditTagTargetingApi` when the SharedLinkPermissionMenu is opened
- [x] refactor `getAllowedPermissionLevels` into its own helper method for use in two different locations
- [x] call onShow for `sharedLinkEditTagTargetingApi` only if the user can edit, the ESL FF is on, and `canShow` is true

### DIFF
--- a/src/features/unified-share-modal/SharedLinkPermissionMenu.js
+++ b/src/features/unified-share-modal/SharedLinkPermissionMenu.js
@@ -22,6 +22,7 @@ type Props = {
     changePermissionLevel: (
         newPermissionLevel: permissionLevelType,
     ) => Promise<{ permissionLevel: permissionLevelType }>,
+    isSharedLinkEditTooltipShown: boolean,
     permissionLevel?: permissionLevelType,
     sharedLinkEditTagTargetingApi?: TargetingApi,
     sharedLinkEditTooltipTargetingApi?: TargetingApi,
@@ -53,6 +54,7 @@ class SharedLinkPermissionMenu extends Component<Props> {
     render() {
         const {
             allowedPermissionLevels,
+            isSharedLinkEditTooltipShown,
             permissionLevel,
             sharedLinkEditTagTargetingApi,
             sharedLinkEditTooltipTargetingApi,
@@ -61,7 +63,6 @@ class SharedLinkPermissionMenu extends Component<Props> {
         } = this.props;
         const { sharedLinkPermissionsMenuButtonProps } = trackingProps;
         const canShowTag = sharedLinkEditTagTargetingApi ? sharedLinkEditTagTargetingApi.canShow : false;
-        const canShowTooltip = sharedLinkEditTooltipTargetingApi ? sharedLinkEditTooltipTargetingApi.canShow : false;
 
         if (!permissionLevel) {
             return null;
@@ -93,11 +94,7 @@ class SharedLinkPermissionMenu extends Component<Props> {
                     }
 
                     // complete tooltip FTUX on opening of dropdown menu
-                    if (
-                        allowedPermissionLevels.includes(CAN_EDIT) &&
-                        canShowTooltip &&
-                        sharedLinkEditTooltipTargetingApi
-                    ) {
+                    if (isSharedLinkEditTooltipShown) {
                         sharedLinkEditTooltipTargetingApi.onComplete();
                     }
                 }}
--- a/src/features/unified-share-modal/SharedLinkPermissionMenu.js
+++ b/src/features/unified-share-modal/SharedLinkPermissionMenu.js
@@ -77,7 +77,14 @@ class SharedLinkPermissionMenu extends Component<Props> {
         };
 
         return (
-            <DropdownMenu constrainToWindow>
+            <DropdownMenu
+                constrainToWindow
+                onMenuOpen={() => {
+                    if (canShow) {
+                        sharedLinkEditTagTargetingApi.onShow();
+                    }
+                }}
+            >
                 <PlainButton
                     className={classNames('lnk', {
                         'is-disabled': submitting,
--- a/src/features/unified-share-modal/SharedLinkPermissionMenu.js
+++ b/src/features/unified-share-modal/SharedLinkPermissionMenu.js
@@ -63,6 +63,7 @@ class SharedLinkPermissionMenu extends Component<Props> {
         } = this.props;
         const { sharedLinkPermissionsMenuButtonProps } = trackingProps;
         const canShowTag = sharedLinkEditTagTargetingApi ? sharedLinkEditTagTargetingApi.canShow : false;
+        const canShowTooltip = sharedLinkEditTooltipTargetingApi ? sharedLinkEditTooltipTargetingApi.canShow : false;
 
         if (!permissionLevel) {
             return null;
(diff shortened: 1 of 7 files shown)


src/features/unified-share-modal/SharedLinkPermissionMenu.js
```diff
@@ -77,7 +77,14 @@ class SharedLinkPermissionMenu extends Component<Props> {
         };
 
         return (
-            <DropdownMenu constrainToWindow>
+            <DropdownMenu
+                constrainToWindow
+                onMenuOpen={() => {
+                    if (canShow && sharedLinkEditTagTargetingApi) {
```
megansmith-box, line 83: sharedLinkEditTagTargetingApi is guaranteed to exist if canShow is true
```suggestion
                    if (canShow) {
```

src/features/unified-share-modal/SharedLinkSection.js
```diff
@@ -197,6 +194,24 @@ class SharedLinkSection extends React.Component<Props, State> {
                     onCopyError();
                 });
         }
+
+        // if ESL ftux tooltip is showing on initial mount, we call onShow
+        const allowedPermissionLevels = this.getAllowedPermissionLevels(
```
megansmith-box, line 199: why do we need to pass all these as parameters?
can we directly access these props in the method `getAllowedPermissionLevels`?

src/features/unified-share-modal/SharedLinkSection.js
```diff
@@ -197,6 +194,24 @@ class SharedLinkSection extends React.Component<Props, State> {
...
+            canChangeAccessLevel,
+            permissionLevel,
+            isDownloadSettingAvailable,
+            isEditSettingAvailable,
+            isAllowEditSharedLinkForFileEnabled,
+        );
+
+        if (
```
megansmith-box, line 207: Since this is in `componentDidUpdate`, what is causing this to be shown only once?

src/features/unified-share-modal/SharedLinkPermissionMenu.js
ChenCodes: I added this second part because there was a flow issue indicating that sharedLinkEditTagTargetingApi could have been undefined.

ChenCodes: I do agree with the above statement though - it seems like flow isn't able to infer that properly.

src/features/unified-share-modal/SharedLinkSection.js
ChenCodes: Yes good catch - will fix.

src/features/unified-share-modal/SharedLinkSection.js
ChenCodes: I originally had it in `componentDidMount` but then I noticed that the `allowedPermissionlevels` array that was coming back was empty. I then looked at `componentDidUpdate` and saw this comment that had been left:
```
    // We handle didUpdate but not didMount because
    // the component initially renders with empty data
    // in order to start showing UI components.
    // When getInitialData completes in the parent we
    // rerender with correct sharedLink data and can
    // check whether to auto create a new one.
    // Note: we are assuming the 2nd render is safe
    // to start doing this check.
```

So it seems that the first mount only renders with empty data so we aren't able to perform our check there.

Regarding your question about showing it only once, I'm wondering if there is some way to check to see if `onShow` has already been called for a specific FTUX. If there is, then we could apply that flag in here to make sure that we only call `onShow` once.

megansmith-box: Ok, I believe we can add a piece of state to determine if onShow has been called similar to the if statement above with isCopySuccessful

src/features/unified-share-modal/SharedLinkPermissionMenu.js
ChenCodes: Yeah I can use that value

src/features/unified-share-modal/SharedLinkSection.js
```diff
@@ -197,6 +193,20 @@ class SharedLinkSection extends React.Component<Props, State> {
...
         }
+
+        // if ESL ftux tooltip is showing on initial mount, we call onShow
+        const allowedPermissionLevels = this.getAllowedPermissionLevels();
+
+        if (
+            allowedPermissionLevels.includes(CAN_EDIT) &&
+            sharedLinkEditTooltipTargetingApi &&
+            sharedLinkEditTooltipTargetingApi.canShow &&
+            !isFtuxTooltipShown
```
ChenCodes, line 204: We want to make sure that `onShow` is only triggered once when the tooltip ftux is shown for the first time.

src/features/unified-share-modal/SharedLinkPermissionMenu.js
```diff
@@ -77,7 +80,24 @@ class SharedLinkPermissionMenu extends Component<Props> {
...
-            <DropdownMenu constrainToWindow>
+            <DropdownMenu
+                constrainToWindow
+                onMenuClose={() => {
+                    if (canShowTag && sharedLinkEditTagTargetingApi) {
+                        sharedLinkEditTagTargetingApi.onComplete();
+                    }
+                }}
+                onMenuOpen={() => {
+                    if (canShowTag && sharedLinkEditTagTargetingApi) {
```
megansmith-box, line 91: can you include allowedPermissionLevels includes CAN_EDIT level?

src/features/unified-share-modal/SharedLinkPermissionMenu.js
```diff
@@ -77,7 +80,24 @@ class SharedLinkPermissionMenu extends Component<Props> {
         };
 
         return (
-            <DropdownMenu constrainToWindow>
+            <DropdownMenu
+                constrainToWindow
+                onMenuClose={() => {
+                    if (canShowTag && sharedLinkEditTagTargetingApi) {
```
megansmith-box, line 86: can you include allowedPermissionLevels includes CAN_EDIT level?

src/features/unified-share-modal/SharedLinkPermissionMenu.js
```diff
@@ -77,7 +80,24 @@ class SharedLinkPermissionMenu extends Component<Props> {
...
+                        sharedLinkEditTagTargetingApi.onComplete();
+                    }
+                }}
+                onMenuOpen={() => {
+                    if (canShowTag && sharedLinkEditTagTargetingApi) {
+                        sharedLinkEditTagTargetingApi.onShow();
+                    }
+
+                    // complete tooltip FTUX on opening of dropdown menu
+                    if (canShowTooltip && sharedLinkEditTooltipTargetingApi) {
```
megansmith-box, line 96: Can you pass the value of `isFtuxTooltipShown` from `SharedLinkSection` and also check for that here

src/features/unified-share-modal/SharedLinkSection.js
megansmith-box, line 204: Can we rename to isSharedLinkEditTooltipShown

src/features/unified-share-modal/SharedLinkPermissionMenu.js
```diff
@@ -22,8 +22,10 @@ type Props = {
     changePermissionLevel: (
         newPermissionLevel: permissionLevelType,
     ) => Promise<{ permissionLevel: permissionLevelType }>,
+    isSharedLinkEditTooltipShown: boolean,
```
ChenCodes, line 25: Now using the flag from state in SharedLinkSection.

megansmith-box (approved):

PR #2682, opened 2021-09-14
feat(datepicker): Option to enable pikaday's default keyboard input

swfree:
- Optional prop `isKeyboardInputAllowed` allows keyboard inputs to fall back to the default pikaday behavior, which means users can use keyboard arrow keys to navigate through the calendar popup.
- Adds isKeyboardInputAllowed prop to TaskForm for accessibility purposes
- Tested all DatePicker storybooks with & without isKeyboardInputAllowed to check for regressions
- Tested TaskForm change in:
- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] IE11
- [x] Edge

Screen Shot 2021-09-15 at 1 40 20 PM

### DIFF
--- a/src/components/date-picker/DatePicker.stories.tsx
+++ b/src/components/date-picker/DatePicker.stories.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react';
 import { IntlProvider } from 'react-intl';
 import { State, Store } from '@sambego/storybook-state';
-import { boolean } from '@storybook/addon-knobs';
 
 import { TooltipPosition } from '../tooltip';
 import DatePicker from './DatePicker';
@@ -29,7 +28,42 @@ export const basic = () => {
                             month: 'long',
                             year: 'numeric',
                         }}
-                        isKeyboardInputAllowed={boolean('isKeyboardInputAllowed', false)}
+                        label="Date"
+                        name="datepicker"
+                        onChange={(date: Date) => {
+                            componentStore.set({ date });
+                        }}
+                        placeholder="Date"
+                        value={state.date}
+                        yearRange={yearRange}
+                    />
+                </IntlProvider>
+            )}
+        </State>
+    );
+};
+
+export const basicWithKeyboardInput = () => {
+    const MIN_TIME = new Date(0);
+    const TODAY = new Date('July 18, 2018');
+    const yearRange = [MIN_TIME.getFullYear(), TODAY.getFullYear()];
+    const componentStore = new Store({
+        date: new Date('July 9, 2018'),
+        fromDate: null,
+        toDate: null,
+    });
+    return (
+        <State store={componentStore}>
+            {state => (
+                <IntlProvider locale="en-US">
+                    <DatePicker
+                        className="date-picker-example"
+                        displayFormat={{
+                            day: 'numeric',
+                            month: 'long',
+                            year: 'numeric',
+                        }}
+                        isKeyboardInputAllowed
                         label="Date"
                         name="datepicker"
                         onChange={(date: Date) => {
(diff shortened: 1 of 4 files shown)


src/components/date-picker/DatePicker.stories.tsx
```diff
@@ -259,6 +277,64 @@ export const withRange = () => {
     );
 };
 
+export const withRangeAndKeyboardInput = () => {
```
jstoffan, line 280: Is a new story needed or can we update the `withRange` story above to support the knob with a default value of `true`? It seems like we should try to avoid having too many permutations of the base story.

src/components/date-picker/DatePicker.tsx
```diff
@@ -128,6 +128,8 @@ export interface DatePickerProps extends WrappedComponentProps {
     isDisabled?: boolean;
     /** Is input required */
     isRequired?: boolean;
+    /** Enables pikaday's default keyboard input support */
+    isKeyboardInputAllowed?: boolean;
```
jstoffan, line 132: Do we want to add a default prop value for this to stay consistent with `isTextInputAllowed`?

src/components/date-picker/DatePicker.stories.tsx
swfree: I was having difficulty with adding two default knobs to `withRange`, but maybe it was just a naming conflict, will try again!

src/components/date-picker/DatePicker.tsx
swfree: Good catch, will do 👍

laughnan (commented):
No additional edits that Jared didn't mention. Looks great, super excited for this!

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
```diff
@@ -448,6 +448,7 @@ class TaskForm extends React.Component<Props, State> {
                                 'data-testid': 'task-form-date-input',
                             }}
                             isDisabled={isForbiddenErrorOnEdit}
+                            isKeyboardInputAllowed
```
tille, line 451: should this attribute have a type?
since it is being received as a prop then maybe in https://github.com/box/box-ui-elements/blob/fd9790216e6812d009a2d6bc8120afd526403464/src/elements/content-sidebar/activity-feed/task-form/TaskForm.js#L45

src/components/date-picker/__tests__/DatePicker.test.tsx
```diff
@@ -304,6 +304,22 @@ describe('components/date-picker/DatePicker', () => {
             expect(stopPropagationSpy).not.toHaveBeenCalled();
         });
 
+        test('should not stop propagation when isKeyboardInputAllowed is enabled', () => {
+            const wrapper = renderDatePicker({ isKeyboardInputAllowed: true });
+            const instance = wrapper.instance();
+            const inputEl = wrapper.find('input').at(0);
+            const stopPropagationSpy = jest.fn();
+            if (instance.datePicker) {
```
tille, line 312: since this is an static wrapper this conditional will always evaluate to true... am I right?

src/components/date-picker/DatePicker.stories.tsx
swfree: So it turns out knobs & state addons aren't compatible, and it also turns out knobs have been deprecated in favor of controls which is something to look into at a later date. Changing a knob value calls forceReRender which messes with state renders. Didn't notice when testing since I initially tested by just adding isKeyboardInputAllowed to all the stories before I remembered that knobs are a thing.

Instead I've removed the knobs and added two stories with isKeyboardInputAllowed defaulted to true.

src/elements/content-sidebar/activity-feed/task-form/TaskForm.js
swfree: yup! the type is defined here: https://github.com/box/box-ui-elements/pull/2682/files#diff-8c22974687f38296f0e4c9b61ba406c1a06dd3be54e61892b6a9ad5f546d0478R132

src/components/date-picker/__tests__/DatePicker.test.tsx
swfree: This is needed due to a typescript error that pops up when its not there: `Object is possibly 'null'`

jstoffan (approved):

laughnan (approved):

il10 (approved):

PR #2685, opened 2021-09-14
fix(versions): improved contrast on highlighted version button

msorianoBox:
Currently, the version history uses #767676 for the modified date text, which has no contrast issues on white background, but the selected state, has a light blue BG (#F5FAFD) and the contrast ratio is not enough.

Screen Shot 2021-09-14 at 12 09 52

So, for the selected class, the text is now #222 and the contrast ratio is now fixed.

Before:
Screen Shot 2021-09-14 at 11 40 46

After:
Screen Shot 2021-09-14 at 12 08 05

### DIFF
--- a/src/elements/content-sidebar/versions/VersionsItemButton.scss
+++ b/src/elements/content-sidebar/versions/VersionsItemButton.scss
@@ -44,7 +44,7 @@
         border-color: $bdl-box-blue;
 
         .bcs-VersionsItem-info {
-            color: $bdl-gray;
+            color: inherit;
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/elements/content-sidebar/versions/VersionsItemButton.scss
```diff
@@ -42,5 +42,9 @@
     &.bcs-is-selected {
         background-color: $hover-blue-background;
         border-color: $bdl-box-blue;
+
+        .bcs-VersionsItem-info {
+            color: $bdl-gray;
```
jstoffan, line 47: Does `color: inherit` work here to avoid duplicating the style declared on line 7 above?

msorianoBox: Yes it does, will send an updated commit in a bit

jstoffan (approved):

PR #2686, opened 2021-09-14
fix(itemicon): Hide preview thumbnail for screen readers

tille:
The image links in each entry lack any alternative text or text content, making it very difficult for screen reader users to determine the purpose of the link.

for that reason we are making them presentation icons and not focusable (since does not have any content to read out)

image

### DIFF
--- a/src/icons/accessible-svg/AccessibleSVG.tsx
+++ b/src/icons/accessible-svg/AccessibleSVG.tsx
@@ -4,8 +4,8 @@ import uniqueId from 'lodash/uniqueId';
 import { SVGProps } from '../../components/accessible-svg/AccessibleSVG';
 
 export interface AccessibleSVGIconProps extends SVGProps {
-    'aria-labelledby'?: string;
     'aria-hidden'?: boolean | 'false' | 'true' | undefined;
+    'aria-labelledby'?: string;
     focusable?: boolean | 'false' | 'true' | 'auto' | undefined;
     opacity?: number;
 }
--- a/src/icons/accessible-svg/AccessibleSVG.tsx
+++ b/src/icons/accessible-svg/AccessibleSVG.tsx
@@ -5,6 +5,7 @@ import { SVGProps } from '../../components/accessible-svg/AccessibleSVG';
 
 export interface AccessibleSVGIconProps extends SVGProps {
     'aria-labelledby'?: string;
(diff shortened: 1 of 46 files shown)


src/icons/item-icon/ItemIcon.tsx
```diff
@@ -81,7 +81,16 @@ export interface ItemIconProps {
...
     const IconComponent = (itemIconTable as { [key: string]: React.FC<SVGProps> })[iconType] || IconFileDefault;
-    return <IconComponent className={className} height={dimension} title={title} width={dimension} />;
+    return (
+        <IconComponent
+            className={className}
+            height={dimension}
+            title={title}
+            width={dimension}
+            aria-hidden="true"
+            role="presentation"
```
jstoffan, line 91: It looks like this should already be handled by the AccessibleSVG component if a title is not passed in: https://github.com/box/box-ui-elements/blob/master/src/icons/accessible-svg/AccessibleSVG.tsx#L29

src/icons/item-icon/ItemIcon.tsx
```diff
@@ -81,7 +81,16 @@ export interface ItemIconProps {
...
 const ItemIcon = ({ className, dimension = 32, iconType, title }: ItemIconProps) => {
     const IconComponent = (itemIconTable as { [key: string]: React.FC<SVGProps> })[iconType] || IconFileDefault;
-    return <IconComponent className={className} height={dimension} title={title} width={dimension} />;
+    return (
+        <IconComponent
+            className={className}
+            height={dimension}
+            title={title}
+            width={dimension}
+            aria-hidden="true"
```
jstoffan, line 90: Is this needed if `role="presentation"` is also present?

src/icons/accessible-svg/AccessibleSVG.tsx
```diff
@@ -5,6 +5,7 @@ import { SVGProps } from '../../components/accessible-svg/AccessibleSVG';
 
 export interface AccessibleSVGIconProps extends SVGProps {
     'aria-labelledby'?: string;
+    'aria-hidden'?: boolean | 'false' | 'true' | undefined;
```
jstoffan, line 8: Can we alphabetize these attributes?

jstoffan (approved):

jstoffan (approved):

swfree (approved):

PR #2688, opened 2021-09-16
feat(contentsidebar): Add instrumentation for data load time

ConradJChan:
Introduces instrumentation to measure the time taken to load the data needed for a sidebar in the `ContentSidebar`. In this case, the PR is applying this instrumentation to the `ActivitySidebar` to measure the time taken to load the initial feed items from component mount time to `Feed` api returning successfully.

Subsequent data fetch times are not logged as a metric.

### DIFF
--- a/src/elements/common/logger/__tests__/Logger.test.js
+++ b/src/elements/common/logger/__tests__/Logger.test.js
@@ -27,6 +27,12 @@ describe('elements/common/logger/Logger', () => {
             const { source } = instance.props;
             expect(eventName).toBe(`${source}::${name}`);
         });
+        test('should create an event name with optional uniqueId', () => {
+            const name = 'bar';
+            const eventName = instance.createEventName(name, '123');
+            const { source } = instance.props;
+            expect(eventName).toBe(`${source}::${name}::123`);
+        });
     });
 
     describe('logMetric()', () => {
@@ -139,6 +145,17 @@ describe('elements/common/logger/Logger', () => {
                 METRIC_TYPE_ELEMENTS_LOAD_METRIC,
                 EVENT_DATA_READY,
                 data,
+                undefined,
+            );
+        });
+
+        test('should log a unique metric with uniqueId', () => {
+            instance.handleDataReadyMetric(data, '123');
+            expect(instance.logUniqueMetric).toHaveBeenCalledWith(
+                METRIC_TYPE_ELEMENTS_LOAD_METRIC,
+                EVENT_DATA_READY,
+                data,
+                '123',
             );
         });
     });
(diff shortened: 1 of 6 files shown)


src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -99,6 +99,8 @@ export const activityFeedInlineError: Errors = {
     },
 };
 
+const MARK_NAME_DATA_LOADING = `${ORIGIN_ACTIVITY_SIDEBAR}_data_loading`;
```
swfree, line 102: should `data_loading` also be a constant, like `EVENT_DATA_LOADING`?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -135,6 +137,8 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
 
     componentDidMount() {
         const { currentUser } = this.props;
+
+        mark(MARK_NAME_DATA_LOADING);
```
jstoffan, line 141: Should we start this in the constructor?

src/elements/content-sidebar/ActivitySidebar.js
ConradJChan: I didn't because the constant is an actual event name being submitted to the analytics store whereas `data_loading` is just a suffix for a marker to measure performance

swfree (approved):

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -455,6 +460,17 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
      * @return {void}
      */
     fetchFeedItemsSuccessCallback = (feedItems: FeedItems): void => {
+        const { logger } = this.props;
+
+        // Only emit metric if has >1 activity feed items (there should always at least be the current version)
+        if (feedItems.length > 1) {
+            mark(MARK_NAME_DATA_READY);
```
jstoffan, line 467: Should we always set the mark, but only log if the feedItems array is greater than one?

src/elements/common/logger/__tests__/Logger.test.js
```diff
@@ -118,6 +118,31 @@ describe('elements/common/logger/Logger', () => {
...
 
+    describe('handleDataReadyMetric()', () => {
+        const END = 'end';
+        const START = 'start';
+        const data = {
+            foo: 'bar',
+            endMarkName: END,
+            startMarkName: START,
+        };
+        let instance;
```
jstoffan, line 129: Is this needed if we only have the one test? Would something like `jest.spyOn(wrapper.instance(), 'logUniqueMetric');` work?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -455,6 +460,24 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
      * @return {void}
      */
     fetchFeedItemsSuccessCallback = (feedItems: FeedItems): void => {
+        const {
```
jstoffan, line 463: This probably warrants its own method.

jstoffan (approved):

PR #2694, opened 2021-09-21
fix(metadata): added aria-expanded to collapsible buttons

msorianoBox:
Hi guys! 

One of the current issues we have is that some collapsible buttons doesn't give enough information to the users while they are being expanded or not.
Screen Shot 2021-09-21 at 9 44 26

Added the aria-expanded attribute so screen readers can inform the users while the buttons are open or not
Screen Shot 2021-09-21 at 9 45 01

### DIFF
--- a/src/components/collapsible/Collapsible.tsx
+++ b/src/components/collapsible/Collapsible.tsx
@@ -115,9 +115,9 @@ class Collapsible extends React.PureComponent<CollapsibleProps, CollapsibleState
                 <div className={buttonClassName}>
                     <PlainButton
                         {...modifiedButtonProps}
+                        aria-expanded={isOpen}
                         className="collapsible-card-title"
                         onClick={this.toggleVisibility}
-                        aria-expanded={isOpen ? 'true' : 'false'}
                         type={ButtonType.BUTTON}
                     >
                         {title}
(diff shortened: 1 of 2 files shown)


src/components/collapsible/Collapsible.tsx
```diff
@@ -102,6 +102,7 @@ class Collapsible extends React.PureComponent<CollapsibleProps, CollapsibleState
         const resinTagTarget: string = RESIN_TAG_TARGET;
         const modifiedButtonProps: { [index: string]: string } = omit(buttonProps, [resinTagTarget]);
         const interactionTarget = buttonProps[resinTagTarget];
+        const ariaExpanded = isOpen ? 'true' : 'false';
```
jstoffan, line 105: Can we in-line this code with its prop since it isn't reused?

msorianoBox: Will do! :D

msorianoBox: Done!

src/components/collapsible/Collapsible.tsx
```diff
@@ -117,6 +117,7 @@ class Collapsible extends React.PureComponent<CollapsibleProps, CollapsibleState
                         {...modifiedButtonProps}
                         className="collapsible-card-title"
                         onClick={this.toggleVisibility}
+                        aria-expanded={isOpen ? 'true' : 'false'}
```
jstoffan, line 120: It looks like `aria-expanded` expects a boolean. Can use `aria-expanded={isOpen}` here? Also, can we keep the props alphabetized?

jstoffan (approved):

PR #2695, opened 2021-09-21
fix(metadata): added aria-pressed to edit metadata button

msorianoBox:
In order to improve accessibility, the Edit metadata button required some tweaking to inform the SR users that is in use
Screen Shot 2021-09-21 at 10 53 57

The aria-pressed attribute helps telling users the button they are interacting with
Screen Shot 2021-09-21 at 10 57 38

### DIFF
--- a/src/features/metadata-instance-editor/Instance.js
+++ b/src/features/metadata-instance-editor/Instance.js
@@ -563,6 +563,7 @@ class Instance extends React.PureComponent<Props, State> {
             return (
                 <Tooltip position="top-left" text={metadataLabelEditText}>
                     <PlainButton
+                        aria-pressed={isEditing ? 'true' : 'false'}
                         aria-label={metadataLabelEditText}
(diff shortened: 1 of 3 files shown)


src/features/metadata-instance-editor/Instance.js
```diff
@@ -563,6 +563,7 @@ class Instance extends React.PureComponent<Props, State> {
             return (
                 <Tooltip position="top-left" text={metadataLabelEditText}>
                     <PlainButton
+                        aria-pressed={isEditing ? 'true' : 'false'}
```
ConradJChan, line 566: minor: alphabetize the props

Also, does just providing `aria-pressed={isEditing}` work as expected?

msorianoBox: Done

ConradJChan (approved):

swfree (approved):

PR #2697, opened 2021-09-22
feat(contentexplorermodal): Added new copy mode

ConradJChan:


src/features/content-explorer/content-explorer/ContentExplorerNewFolderButton.js
```diff
@@ -16,9 +16,10 @@ const ContentExplorerNewFolderButton = ({
     isCreateNewFolderAllowed = true,
 }) => {
     const doesModeAllowCreateNewFolder =
-        contentExplorerMode === ContentExplorerModes.SELECT_FOLDER ||
+        contentExplorerMode === ContentExplorerModes.COPY ||
```
jstoffan, line 19: Are we sure this is something we want to support for copy?

jstoffan (approved):

src/features/content-explorer/content-explorer/ContentExplorerNewFolderButton.js
ConradJChan: I feel like it should be ok because there are other props which can control the showing and hiding of the create folder button as well, this just makes it an option for `copy` mode, if so desired

PR #2700, opened 2021-09-24
fix(content-explorer): disable item name click if item disabled

jsenar:
Adds guard clause to handleItemNameClick if item is disabled. Similar to the other click handlers above (handleItemDoubleClick and handleItemClick).

It seems to me the the existence of this guard clause in handleItemDoubleClick implies that we should not be able to enter a folder if it is disabled. https://github.com/box/box-ui-elements/blob/master/src/features/content-explorer/content-explorer/ContentExplorer.js#L296-L304

Also open to separating this into a separate prop to disable navigation on click.

### DIFF
--- a/src/features/content-explorer/content-explorer/ContentExplorer.js
+++ b/src/features/content-explorer/content-explorer/ContentExplorer.js
@@ -309,6 +309,10 @@ class ContentExplorer extends Component {
         const { items } = this.props;
         const item = items[index];
 
+        if (item.isDisabled || item.isLoading) {
+            return;
+        }
+
         if (item.type !== ItemTypes.FOLDER) {
             return;
         }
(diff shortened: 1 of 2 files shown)


src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -309,6 +309,10 @@ class ContentExplorer extends Component {
         const { items } = this.props;
         const item = items[index];
 
+        if (item.isDisabled || item.isLoading) {
```
swfree, line 312: Can you add a test for this?
Here's some similar tests you can base it on: https://github.com/box/box-ui-elements/blob/master/src/features/content-explorer/content-explorer/__tests__/ContentExplorer.test.js#L272 & https://github.com/box/box-ui-elements/blob/master/src/features/content-explorer/content-explorer/__tests__/ContentExplorer.test.js#L388

jsenar: Sure thing, thanks!

jsenar: Added

swfree (approved):

PR #2702, opened 2021-09-24
fix(unified-share-modal): filter out collaborated users and groups

Machoper:
Filter out users and groups who has already collaborated on the file when sending requests.

### DIFF
--- a/src/elements/content-sharing/SharingNotification.js
+++ b/src/elements/content-sharing/SharingNotification.js
@@ -259,7 +259,7 @@ function SharingNotification({
             closeComponent();
         },
         setIsLoading,
-        transformRequest: data => convertCollabsRequest(data),
+        transformRequest: data => convertCollabsRequest(data, collaboratorsList),
     });
     if (sendInvitesFn && !sendInvites) {
         setSendInvites(() => sendInvitesFn);
--- a/src/features/unified-share-modal/flowTypes.js
+++ b/src/features/unified-share-modal/flowTypes.js
@@ -171,7 +171,7 @@ export type sharedLinkType = {
 
 export type collaboratorType = {
     collabID: number,
-    email: string,
+    email?: string,
     expiration?: {
         executeAt: string,
     },
--- a/src/features/unified-share-modal/flowTypes.js
+++ b/src/features/unified-share-modal/flowTypes.js
@@ -171,6 +171,7 @@ export type sharedLinkType = {
 
 export type collaboratorType = {
     collabID: number,
+    email: string,
     expiration?: {
         executeAt: string,
     },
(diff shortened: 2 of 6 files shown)


src/features/unified-share-modal/utils/convertData.js
```diff
@@ -458,30 +458,45 @@ export const convertCollabsResponse = (
...
     collabRequest: InviteCollaboratorsRequest,
+    collaboratorsList: collaboratorsListType | null,
 ): ContentSharingCollaborationsRequest => {
     const { emails, groupIDs, permission } = collabRequest;
     const emailArray = emails ? emails.split(',') : [];
     const groupIDArray = groupIDs ? groupIDs.split(',') : [];
+    const collabSet = new Set();
+    if (collaboratorsList) {
+        collaboratorsList.collaborators.forEach(collab => {
+            if (collab.type === COLLAB_USER_TYPE) {
```
ctuckz, line 469: Does this need to also check for falsy emails, like the check we're doing for groups?

src/features/unified-share-modal/flowTypes.js
```diff
@@ -171,6 +171,7 @@ export type sharedLinkType = {
 
 export type collaboratorType = {
     collabID: number,
+    email: string,
```
ctuckz, line 174: Is email always present, or should the type allow it to be undefined? I see in some of the tests that groups don't have an email field

src/features/unified-share-modal/utils/convertData.js
```diff
@@ -458,30 +458,45 @@ export const convertCollabsResponse = (
  */
 export const convertCollabsRequest = (
     collabRequest: InviteCollaboratorsRequest,
+    collaboratorsList: collaboratorsListType | null,
```
ctuckz, line 461: IMO this and `collabSet` below could benefit from being renamed to something like "existingCollaboratorsList" to help distinguish it from the collabs being added in the request

ctuckz (approved):

swfree (approved):

src/elements/content-sharing/SharingNotification.js
```diff
@@ -259,7 +259,7 @@ function SharingNotification({
             closeComponent();
         },
         setIsLoading,
-        transformRequest: data => convertCollabsRequest(data),
+        transformRequest: data => convertCollabsRequest(data, collaboratorsList),
```
jstoffan, line 262: It seems like this filtering logic should live in the `handleSuccess` call above or even during render rather than a util since it's accessing component state.

jstoffan: I just saw it's accessing a prop, not state. I still think it'd best be served for the logic to live in the component rather than a util, but not going to hold up the merge for it.

jstoffan (approved):
Passing around state references is a little unusual.

PR #2708, opened 2021-10-05
fix(select): changing select tooltip to be triggered by a button instead of on hover

mrscobbler:


src/components/select/Select.js
```diff
@@ -61,10 +62,14 @@ const Select = ({
...
                         {infoTooltip && (
-                            <Tooltip position="middle-right" text={infoTooltip}>
-                                <span className="tooltip-icon-container">
+                            <Tooltip isShown={infoTooltipIsOpen} position="middle-right" text={infoTooltip}>
+                                <PlainButton
+                                    type="button"
+                                    className="tooltip-icon-container"
+                                    onClick={() => setInfoTooltipIsOpen(!infoTooltipIsOpen)}
+                                >
                                     <IconInfo height={16} width={16} {...infoIconProps} />
```
alexkrolick, line 71: Can we give a default title or aria-label to the icon so that it defaults to being accessible for screen readers?

src/components/select/__tests__/__snapshots__/Select.test.js.snap
```diff
@@ -102,19 +102,22 @@ exports[`components/select/Select should render infoTooltip when specified 1`] =
           constrainToScrollParent={false}
           constrainToWindow={true}
           isDisabled={false}
+          isShown={false}
           position="middle-right"
           text="hello!!!"
           theme="default"
         >
-          <span
+          <PlainButton
```
alexkrolick, line 110: Let's also add a test for the visibility interaction

src/components/select/Select.js
```diff
@@ -61,10 +62,14 @@ const Select = ({
                             <span className="select-overlay" />
                         </span>
                         {infoTooltip && (
-                            <Tooltip position="middle-right" text={infoTooltip}>
-                                <span className="tooltip-icon-container">
+                            <Tooltip isShown={infoTooltipIsOpen} position="middle-right" text={infoTooltip}>
+                                <PlainButton
```
swfree, line 66: Are the styles for the button the same before & after? Could you add a screenshot to show that?

src/components/select/Select.js
tille: what about the other tooltip-icons that are not `Select` components
like this one https://git.dev.box.net/Box/EndUserApp/blob/master/src/components/pages/account-sharing-page/components/AppNotificationsSection.js#L27

src/components/select/Select.js
```diff
@@ -61,10 +62,14 @@ const Select = ({
                             <span className="select-overlay" />
                         </span>
                         {infoTooltip && (
-                            <Tooltip position="middle-right" text={infoTooltip}>
-                                <span className="tooltip-icon-container">
```
swfree, line 65: is there any chance that components using this could be using `span` to add override styles? preferably not, but i'm wondering if this counts as a breaking change

PR #2721, opened 2021-10-12
fix(mediafigure): set "as" property to div as default

hormatias:
Change a figure element to a div element for accessibility purposes

image

image


hormaza:
> If you want to change the semantics of this, change the Media element, not this individual implementation. You can also change the `as` prop if this particular implementation needs to be a div but figure is OK in other places: https://opensource.box.com/box-ui-elements/storybook/?path=/docs/components-media-mediafigure--example
> 
> (Would be nice to have an explanation of why that's the case).

Understood. Updated!

jstoffan:
@hormaza, thanks for the response. I think the main question is: Should we update the default value for `as` from `figure` to `div` in the [Media.Figure component](https://github.com/box/box-ui-elements/blob/master/src/components/media/MediaFigure.tsx#L20) rather than setting it here as a one-off. If not, why?

PR #2723, opened 2021-10-14
docs(badgeable): fix import of CountBadge in Badgeable readme

booc0mtaco:
The previous import was implicit due to how [styleguidist](https://react-styleguidist.js.org) collects components. After the conversion of `CountBadge` to TypeScript, this broke the link. 

This change imports the compiled JS file directly in the example.


booc0mtaco:
Before & After
  
Screen Shot 2021-10-13 at 11 12 21 PM

Screen Shot 2021-10-13 at 11 12 39 PM

PR #2739, opened 2021-11-03
feat: add freemium upsell experiment changes

jackiejou:
With Experiment Enabled. Upgrade Button
image

Without Experiment Enabled.
image


src/elements/content-uploader/ItemList.js
```diff
@@ -56,7 +65,6 @@ const ItemList = ({ isResumableUploadsEnabled = false, items, onClick, onRemoveC
                         cellRenderer={actionCell}
                         dataKey="status"
                         flexShrink={0}
-                        width={25}
```
jackiejou, line 59: removed to allow the upgrade button to show full width. See added screenshots for behavior changes

src/elements/content-uploader/UploadsManager.js
```diff
@@ -27,16 +29,18 @@ type Props = {
 };
 
 const UploadsManager = ({
+    freemiumUpsellCTACb,
```
jasminevle, line 32: love this re-alphabetization

src/elements/content-uploader/ItemList.js
```diff
@@ -16,18 +16,27 @@ import '@box/react-virtualized/styles.css';
 import './ItemList.scss';
 
 type Props = {
+    freemiumUpsellCTACb?: ?() => void,
```
jasminevle, line 19: Does the Cb stand for callback? is there convention to use that or can we rename it to `onFreemiumUpsellCTAClick`?

src/elements/content-uploader/progressCellRenderer.js
```diff
@@ -35,11 +35,19 @@ type Props = {
  * @param {string} [itemName]
  * @returns {FormattedMessage}
  */
-const getErrorMessage = (errorCode: ?string, itemName: ?string) => {
+const getErrorMessage = (errorCode: ?string, itemName: ?string, maxFileSize?: ?string) => {
     switch (errorCode) {
         case ERROR_CODE_UPLOAD_CHILD_FOLDER_FAILED:
             return <FormattedMessage {...messages.uploadsOneOrMoreChildFoldersFailedToUploadMessage} />;
         case ERROR_CODE_UPLOAD_FILE_SIZE_LIMIT_EXCEEDED:
+            if (maxFileSize) {
```
jstoffan, line 43: I'm not sure we want to have two sources of truth (client, server) for the maximum file size. Can we add a new error code on the backend that we can key off of to ask the user to upgrade, instead?

src/elements/content-uploader/ContentUploader.js
```diff
@@ -1203,17 +1205,19 @@ class ContentUploader extends Component<Props, State> {
...
             language,
+            maxFileSize,
+            measureRef,
             messages,
             onClose,
-            className,
-            measureRef,
-            isTouch,
-            fileLimit,
+            onFreemiumUpsellCTAClick,
```
jstoffan, line 1219: ContentUploader can used by platform developers in their own applications, outside of Box's webapp. We should try to keep the component API simple and general, when possible. Do we plan to add more experiences like this one? If so, should we wrap them in an object to avoid further prop bloat?

src/elements/content-uploader/ItemAction.js
```diff
@@ -66,6 +85,18 @@ const ItemAction = ({ status, onClick, intl, isResumableUploadsEnabled, isFolder
             break;
     }
 
+    if (STATUS_ERROR && code === ERROR_CODE_UPLOAD_FILE_SIZE_LIMIT_EXCEEDED && !!onFreemiumUpsellCTAClick) {
+        return (
```
jstoffan, line 89: We generally try to avoid multiple returns, as it can lead to confusing behavior. Can we tie into the status switch above and the existing `onClick` handler, instead?

src/elements/content-uploader/ItemList.js
jstoffan: The `react-virtualized` types indicate that this prop is required. Is that correct? If not, can we validate the action cell renders properly in all browsers, including IE11 and Safari? Some browsers don't handle missing flex basis very well.

src/elements/content-uploader/ContentUploader.js
jackiejou: These are the final current planned ContentUploader changes for this project. No other prop needed. I suspect that once the experiment concludes, I will make a PR to clean up these changes (2 props)

src/elements/content-uploader/ItemAction.js
jackiejou: The PrimaryButton design (a new resin tag, no tooltip) is quite different from the current Plain Button icon with tooltip. Classname `bcu-item-action` also limits the width and height of this component which would cause the PrimaryButton to be cut off or misaligned. In addition, code clean up may be easier in the future.

src/elements/content-uploader/progressCellRenderer.js
jackiejou: The plan is to get the maxFileSize from the server in the EUA (exist as CurrentUserRecord) rather than passing in a hardcoded data in the FE repo. This avoids adding new endpoint/changing current APIs for this experimental change. (reference maxFileSizeAsString in EUA)

src/elements/content-uploader/progressCellRenderer.js
```diff
@@ -58,7 +61,7 @@ const getErrorMessage = (errorCode: ?string, itemName: ?string) => {
     }
 };
 
-export default () => ({ rowData }: Props) => {
+export default (shouldShowUpgradeCTAMessage?: ?boolean) => ({ rowData }: Props) => {
```
laughnan, line 64: Is this saying that the prop is optional and it also that it's not always a boolean?

src/elements/content-uploader/ItemList.js
jackiejou: Updated to the conditional below

jstoffan (approved):

PR #2740, opened 2021-11-05
feat(buttons): new styles, large size, and icon support

alexkrolick:
New styles:
* The button components have been restyled to take on the "dejablue" theme by default. There is now no difference between dejablue and the regular styles. Overrides in other apps can be removed.

New props:
* `size` `undefined | large`
   * The standard size, used in existing buttons, is still 32px
   * The large size is 40px, which is commonly used in new designs
* `icon` `undefined | React.Element` (`<SomeIcon />`)
   * Icon elements are automatically positioned for optical alignment and height/width are controlled by the button.
   * The `icon` prop can be used without `children` to create a square icon-only button. Note that `title` should be set on AccessibleSVG-based icons to ensure an accessible name is set for the button.
   
 Documentation:
* Docs added for the `bdl-Button-margins(size)` SCSS mixin for PlainButton
* Docs added for working with built-in margins on the Button and PrimaryButton

Variables:
* Border radius size enum
     "bdl-border-radius-size": "4px",
    "bdl-border-radius-med-size": "6px",
    "bdl-border-radius-large-size": "8px",
    "bdl-border-radius-xlarge-size": "12px",


src/styles/variables.json
```diff
@@ -96,7 +96,9 @@
     "overlay-z-index": "190",
     "tooltip-z-index": "190",
     "bdl-border-radius-size": "4px",
-    "bdl-border-radius-large-size": "10px",
```
jstoffan, line 99: Is this value used anywhere today and, if so, do we need to update those usages?

src/styles/common/_buttons.scss
```diff
@@ -62,7 +114,8 @@
     }
 
     &.is-loading {
```
jstoffan, line 116: Do all of the new variants work with the loading and radar treatments?

src/styles/common/_buttons.scss
```diff
@@ -1,35 +1,87 @@
+@import '../constants/typography';
+@import '../constants/layout';
+@import '../mixins/buttons';
+
 /**************************************
  * Buttons
  **************************************/
+$bdl-btn-border-width: 1px;
+$bdl-btn-height: 32px;
+$bdl-btn-height-large: 40px;
```
jstoffan, line 10: Should we use SUIT for our variable names here, something like `$bdl-Button-height--large`?

src/styles/common/_buttons.scss
```diff
@@ -1,35 +1,87 @@
...
+$bdl-btn-icon-size: 20px;
+$bdl-btn-text-icon-size: 16px;
+$bdl-btn-padding-horizontal: $bdl-grid-unit * 4;
+$bdl-btn-padding-vertical: 10px; // Do not reuse, use flexbox or line-height instead
```
jstoffan, line 14: If we don't want some of these to be reused, should we move them inside of `.btn` to keep them from being global variables?

src/features/security/_mixins.scss
```diff
@@ -5,7 +5,7 @@
     align-items: center;
     margin: 0;
     padding: 2px 8px;
-    border-radius: $bdl-border-radius-large-size;
+    border-radius: 2px + 16px / 2; // rounded sides: vert padding + half the line height of content
```
jstoffan, line 8: Is the math needed here or should we just set the actual value?

src/components/primary-button/PrimaryButton.stories.tsx
```diff
@@ -20,7 +20,7 @@ export const loading = () => <PrimaryButton isLoading>Click Here</PrimaryButton>
 export const disabled = () => <PrimaryButton isDisabled>Click Here</PrimaryButton>;
 
 export default {
-    title: 'Components|PrimaryButton',
+    title: 'Components|Buttons/PrimaryButton',
```
jstoffan, line 23: Is the purpose of the `Buttons/` here for visual organization? These components are all top-level and aren't really grouped together at the moment.

src/components/button/__tests__/Button.test.tsx
```diff
@@ -73,6 +73,37 @@ describe('components/button/Button', () => {
...
+        const wrapper = shallow(<Button icon={<FakeIcon />}>Click Here</Button>);
+        const iconContainer = wrapper.find('.bdl-btn-icon');
+        const textContainer = wrapper.find('.btn-content');
+        const { width, height } = iconContainer.find('FakeIcon').props();
+        expect(iconContainer.length).toBe(1);
+        expect(textContainer.length).toBe(1);
+        expect(width).toEqual(height);
+        expect(width).toEqual(16);
+
+        expect(wrapper).toMatchSnapshot();
```
jstoffan, line 104: Is the snapshot needed with all the assertions?

src/components/button/__tests__/Button.test.tsx
```diff
@@ -73,6 +73,37 @@ describe('components/button/Button', () => {
...
+        expect(width).toEqual(20);
+    });
+
+    test('should render icon next to text if icon and children props are both set', () => {
+        const FakeIcon = (props: Record<string, unknown>) => <svg {...props} />;
+        const wrapper = shallow(<Button icon={<FakeIcon />}>Click Here</Button>);
+        const iconContainer = wrapper.find('.bdl-btn-icon');
+        const textContainer = wrapper.find('.btn-content');
+        const { width, height } = iconContainer.find('FakeIcon').props();
+        expect(iconContainer.length).toBe(1);
```
jstoffan, line 99: Minor, but we can also do `expect(wrapper.exists('.bdl-btn-icon')).toBe(true);`

src/components/button/Button.tsx
```diff
@@ -53,22 +57,51 @@ class Button extends React.Component<ButtonProps> {
...
+        const hasIcon = !!icon;
+        const hasText = !!children;
+
         const styleClassName = classNames(
             'btn',
             {
                 'is-disabled': isDisabled,
                 'is-loading': isLoading,
                 'is-selected': isSelected,
+                'bdl-is-large': size === 'large',
```
jstoffan, line 87: What do you think of using a modifier here, like `.bdl-Button--large`, rather than state?

src/components/button/Button.tsx
```diff
@@ -53,22 +57,51 @@ class Button extends React.Component<ButtonProps> {
...
+                'bdl-has-icon': hasIcon,
             },
             className,
         );
 
+        const textContent = children ? <span className="btn-content">{children}</span> : null;
+        let iconContent = null;
+        if (icon) {
+            // Size of text+icon is 16px, just icon is 20px
+            const iconSize = hasIcon && hasText ? 16 : 20;
```
jstoffan, line 97: Does the icon need to grow for the `large` variant?

src/components/button/Button.tsx
```diff
@@ -83,7 +116,8 @@ class Button extends React.Component<ButtonProps> {
                 type={type}
                 {...buttonProps}
             >
-                <span className="btn-content">{children}</span>
+                {textContent}
```
jstoffan, line 119: Should we continue to move toward inlining our renders and/or moving more complex inner render logic to separate components?

src/components/button/Button.tsx
```diff
@@ -53,22 +57,51 @@ class Button extends React.Component<ButtonProps> {
...
             },
             className,
         );
 
+        const textContent = children ? <span className="btn-content">{children}</span> : null;
+        let iconContent = null;
+        if (icon) {
+            // Size of text+icon is 16px, just icon is 20px
+            const iconSize = hasIcon && hasText ? 16 : 20;
+            const fixedSizeIcon = React.cloneElement<{ width: number; height: number }>(icon, {
```
jstoffan, line 98: Should the type here come from some Icon base class?

src/components/button/Button.tsx
```diff
@@ -83,7 +116,8 @@ class Button extends React.Component<ButtonProps> {
                 type={type}
                 {...buttonProps}
             >
-                <span className="btn-content">{children}</span>
+                {textContent}
+                {iconContent}
```
jstoffan, line 120: Do we need to solve for left/right positioning of the icon?

src/styles/common/_buttons.scss
```diff
@@ -1,35 +1,87 @@
...
+        // Instead of using line-height + padding + border = height like the base size,
+        // just set the line-height + border; this already handles vertical alignment.
+        padding: 0 $bdl-btn-padding-horizontal;
+        font-size: $bdl-fontSize--dejaBlue;
+        line-height: $bdl-btn-height-large - 2 * $bdl-btn-border-width;
+    }
+
+    &.bdl-has-icon {
+        display: inline-flex;
+        // Buttons with icons use flexbox to align content instead of line-height/padding
```
jstoffan, line 54: Is this comment needed?

src/components/button/Button.tsx
```diff
@@ -53,22 +57,51 @@ class Button extends React.Component<ButtonProps> {
...
+            showRadar,
+            ...rest
+        } = this.props;
         const buttonProps: Record<string, boolean> = omit(rest, ['onClick']);
         if (isDisabled) {
             buttonProps['aria-disabled'] = true;
         }
 
+        const hasIcon = !!icon;
+        const hasText = !!children;
```
jstoffan, line 79: Should we be making the assumption that `children` will only contain text? If a consumer passes in an icon on the left, for example?

src/components/button/Button.tsx
```diff
@@ -24,6 +26,8 @@ export interface ButtonProps {
     onClick?: Function;
     /** to set buttons inner ref */
     setRef?: Function;
+    /** size of the button */
+    size?: 'large' | null;
```
jstoffan, line 30: Does this need to be nullable? Should we list the "normal" size?

src/styles/common/_buttons.scss
```diff
@@ -1,35 +1,87 @@
...
+            margin-left: $bdl-btn-padding-horizontal;
+        }
+
+        .bdl-btn-icon {
+            // Icon without text:
+            margin: ($bdl-btn-height - $bdl-btn-icon-size) / 2 - $bdl-btn-border-width;
+
+            // Icon with text:
+            & + .btn-content {
+                margin-right: 3 * $bdl-grid-unit;
```
jstoffan, line 71: Can we change this to `$bdl-grid-unit * 3` to jive with the line below?

src/styles/common/_buttons.scss
```diff
@@ -1,35 +1,87 @@
...
+                margin-left: $bdl-grid-unit * 2;
```
jstoffan, line 72: Should we have a comment about why these two margin values are different?

src/styles/common/_buttons.scss
```diff
@@ -1,35 +1,87 @@
...
+            }
+        }
+
+        &.bdl-is-large {
+            height: $bdl-btn-height-large;
```
jstoffan, line 77: Should this be in the base `&.bdl-is-large` style rules above? It seems odd to only have it set for icons.

src/styles/common/_buttons.scss
```diff
@@ -1,35 +1,87 @@
+@import '../constants/typography';
```
jstoffan, line 1: Do any of these imports contain new style rules or anything else that could impact consumers?

src/components/primary-button/PrimaryButton.stories.tsx
alexkrolick: image

It creates this grouping in the docs so you don't have to jump around to get to the PrimaryButton and PlainButton  docs, which inherit the API from the base Button

src/components/button/__tests__/Button.test.tsx
alexkrolick: Wanted to check the ordering of the content/icon and make sure no additional elements popped up (negative assertion, so to speak)

src/styles/variables.json
alexkrolick: https://github.com/box/box-ui-elements/pull/2740/files#diff-fca7649e0a5992886ed74ba82a2ff242ad77703cbf9b3fd25a060f231e84837cR8

It was created for the use case in that file and used only there

src/features/security/_mixins.scss
alexkrolick: @patrickpaul preferred showing why the value was that number

src/styles/common/_buttons.scss
alexkrolick: yes they do

src/components/button/Button.tsx
alexkrolick: It actually looks OK inlined, I'll do that

src/components/button/Button.tsx
alexkrolick: No, non-standard ones will be migrated to right-side icons, except the special case dropdowns that have other styling for the padding and alignment anyways.

src/styles/common/_buttons.scss
alexkrolick: I'd like to call out that there is an important internal difference in the layout

src/styles/common/_buttons.scss
alexkrolick: It's needed here for specificity, and not needed above because the non-icon buttons define their heights using lineheight/padding so the height prop is redundant and conflicting. Could possibly refactor this to make it more intuitive.

src/styles/common/_buttons.scss
alexkrolick: Screen Shot 2021-11-05 at 3 16 16 PM

with all of the above

src/components/button/Button.tsx
alexkrolick: The main use case for icon buttons is the large size. Regular size is supported with the same size per design, but not commonly used.

src/features/security/_mixins.scss
alexkrolick: Actually, decided to change it to just a hardcoded number and explanation

src/styles/common/_buttons.scss
alexkrolick: No, just variables

src/styles/common/_buttons.scss
```diff
@@ -1,35 +1,90 @@
...
+@import '../constants/layout';
+@import '../mixins/buttons';
```
alexkrolick: Removed this, it's not used

jstoffan (approved):

PR #2743, opened 2021-11-08
feat(borders): restyle elements new border radius

alexkrolick:
PR 2 onto restyle integration branch

* Migrate to new standard border radius sizes.
   * Form elements that match the button get `$bdl-border-radius-med-size`
   * Other elements get closest matching size
   * Very small elements with 2px radius unchanged (for now)
* Fixes issues with dropdown elements where the new design should have a space between the dropdown content and the button or input field, as well as resetting list item hover state radiuses for some of these back to 0 because they don't yet have content padding and should not have been migrated to have rounded corners


src/components/selector-dropdown/SelectorDropdown.scss
```diff
@@ -36,6 +37,10 @@
         }
     }
 
+    .datalist-item {
```
ConradJChan, line 40: Doesn't seem like `SelectorDropdown` enforces its children to be of type `DataListItem`, is adding `.datalist-item` styles enough?

alexkrolick: The problem is DataListItem got restyled with curved corners, which is meant to work with a padded list style,  so resetting that unexpected change gets the edge-to-edge style of the standard dropdown back to the intended design. Other use cases shouldn't have been affected I don't think.

src/components/button-group/ButtonGroup.scss
```diff
@@ -15,7 +15,7 @@
             margin-left: 0;
             padding-right: 11px;
             padding-left: 11px;
-            border-radius: 0 4px 4px 0;
+            border-radius: 0 $bdl-border-radius-med-size $bdl-border-radius-med-size 0;
```
jstoffan, line 18: I'd suggest putting the modifier at the end to keep all these variables alphabetically together, which is similar to what we do for colors. Maybe like `bdl-border-radius-size-med`

src/components/form-elements/draft-js-mention-selector/MentionSelector.scss
```diff
@@ -16,9 +16,18 @@
...
         background: $white;
         border: 1px solid $bdl-gray-30;
+        border-radius: $bdl-border-radius-med-size;
         box-shadow: 0 2px 6px 0 fade-out($black, .9);
     }
+
+    // Normally .bdl-SelectorDropdown > .text-input-container has a bottom margin
+    // but the DraftJSEditor doesn't have one.
+    // Add spacing above dropdown wrappers here instead:
+    .overlay {
```
jstoffan, line 30: Is this a bug fix? If so, should it be a separate PR?

src/components/select-field/SelectField.scss
```diff
@@ -54,7 +55,9 @@
 .bdl-SelectFieldDropdown {
     z-index: 4;
     min-width: 100%;
+    margin-top: $bdl-grid-unit;
```
jstoffan, line 58: Do we have any use cases where a flush menu is still the desired experience?

src/components/selector-dropdown/SelectorDropdown.scss
alexkrolick: Screen Shot 2021-11-08 at 5 33 33 PM

src/components/form-elements/draft-js-mention-selector/MentionSelector.scss
alexkrolick: It aligns the spacing of this element, which customizes the SelectorDropdown with an alternative text input box, with the styling of the other selectors.

Screen Shot 2021-11-08 at 5 33 33 PM


Screen Shot 2021-11-09 at 1 20 42 PM

src/components/select-field/SelectField.scss
alexkrolick: No, we don't

jstoffan (approved):

PR #2751, opened 2021-11-15
feat(datepicker): Add option to use date type for text input

swfree:
- Introduces optional `isAccessible` prop which uses the input type "date", to support manual entry that is accessible and internationalized

Date Picker without `isAccessible`:

Screen Shot 2021-11-15 at 10 45 48 AM

Screen Shot 2021-11-15 at 10 45 35 AM

Date Picker with `isAccessible`:

datepicker-gif
Screen Shot 2021-11-15 at 11 07 06 AM
Screen Shot 2021-11-15 at 10 45 22 AM
Screen Shot 2021-11-15 at 10 44 48 AM
Screen Shot 2021-11-15 at 10 45 08 AM
Screen Shot 2021-11-15 at 10 44 41 AM
Screen Shot 2021-11-15 at 11 07 15 AM


src/components/date-picker/DatePicker.tsx
```diff
@@ -314,18 +365,23 @@ class DatePicker extends React.Component<DatePickerProps> {
...
+            event.preventDefault();
+        }
+
+        if ((isTextInputAllowed || (isAccessible && !this.canUseDateInputType)) && event.key === ENTER_KEY) {
             event.preventDefault();
         }
 
-        if (isTextInputAllowed && event.key === ENTER_KEY) {
+        // In Firefox, enter & spacebar open up the browser's default date picker. Suppress it.
+        if (isAccessible && (event.key === ENTER_KEY || event.keyCode === 32) && Browser.isFirefox()) {
```
ConradJChan, line 384: Should `Browser.isFirefox()` be moved to the beginning of the condition?

src/components/date-picker/DatePicker.tsx
```diff
@@ -337,9 +393,29 @@ class DatePicker extends React.Component<DatePickerProps> {
         }
     };
 
-    handleInputBlur = (event: React.FocusEvent<HTMLInputElement>) => {
-        const { onBlur, isTextInputAllowed } = this.props;
+    handleOnChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+        const { isAccessible, onChange } = this.props;
+        if (isAccessible && this.canUseDateInputType) {
```
ConradJChan, line 398: Should we invert this to a guard clause to reduce the indentation?

src/components/date-picker/DatePicker.tsx
```diff
@@ -381,11 +461,56 @@ class DatePicker extends React.Component<DatePickerProps> {
...
 
+    handleDivClick = (event: React.SyntheticEvent<HTMLDivElement>) => {
+        const { isDisabled } = this.props;
+
+        if (isDisabled) {
+            return;
+        }
+
+        event.preventDefault();
+        event.stopPropagation();
```
ConradJChan, line 472: Should these be called whether disabled or not?

src/components/date-picker/DatePicker.tsx
```diff
@@ -381,11 +461,56 @@ class DatePicker extends React.Component<DatePickerProps> {
...
+
+        if (this.datePicker) {
```
ConradJChan, line 474: Could this be included with the guard clause above with `isDisabled`?

src/components/date-picker/DatePicker.tsx
```diff
@@ -399,6 +524,50 @@ class DatePicker extends React.Component<DatePickerProps> {
...
+        const { intl, isAccessible, isAlwaysVisible, isDisabled } = this.props;
+        const { formatMessage } = intl;
+
+        if (isAlwaysVisible) {
+            return null;
+        }
+
+        // De-emphasizing the Pikaday date picker because it does not meet accessibility standards
+        // Screenreaders & navigating via keyboard will no longer pick up on this element
+        if (isAccessible) {
```
ConradJChan, line 537: Could we make use of `aria-hidden="true"` or something instead to remove it from being read by screenreaders?

src/components/date-picker/DatePicker.tsx
```diff
@@ -440,17 +610,33 @@ class DatePicker extends React.Component<DatePickerProps> {
...
+        }
+        let patternAttr = {};
+        if (isAccessible && !this.canUseDateInputType) {
+            patternAttr = { pattern: '\\d{4}-\\d{2}-\\d{2}' };
+        }
+        // "name" prop is required for pattern validation to be surfaced on form submit. See components/form-elements/form/Form.js
+        let nameAttr = {};
+        if (isAccessible && !this.canUseDateInputType) {
+            nameAttr = { name };
+        }
```
ConradJChan, line 637: Seems like these two can be combined, might be worth seeing if this block of code can be simplified further.

src/components/date-picker/AccessiblePikaday.ts
```diff
@@ -0,0 +1,69 @@
...
+    accessibleField: HTMLElement | null | undefined;
+
+    datePickerButtonEl: HTMLElement | null | undefined;
+
+    constructor(options: AccessiblePikadayOptions) {
+        super(options);
+        this.accessibleField = options.accessibleField;
+        this.datePickerButtonEl = options.datePickerButtonEl;
+
+        // Override behavior as if `options.field` and `options.bound` were set.
```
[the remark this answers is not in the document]
swfree: Hm not sure about this one, @viodragon2 what do you think?

src/components/date-picker/AccessiblePikaday.ts
```diff
@@ -0,0 +1,69 @@
...
+        }
+    }
+
+    handleBlur = () => {
+        // TODO: Test in IE11
+        this.hide();
+    };
+
+    handleClickOutside = (e: MouseEvent) => {
+        if (this.isVisible() && this.datePickerButtonEl && this.datePickerButtonEl.contains(e.target as HTMLElement)) {
```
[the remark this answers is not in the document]
swfree: Closing the calendar on click was already happening elsewhere, which was causing a problem where clicking the button would always keep the calendar open. This addition enables closing the calendar on clicking `datePickerButtonEl`

src/components/date-picker/AccessiblePikaday.ts
```diff
@@ -0,0 +1,69 @@
...
+
+        if (this.isVisible() && !this.el.contains(e.target as HTMLElement)) {
+            this.hide();
+        }
+    };
+
+    show() {
+        super.show();
+        if (this.accessibleField) {
+            document.addEventListener('click', this.handleClickOutside, true);
```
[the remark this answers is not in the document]
swfree: @viodragon2 what do you think?

src/components/date-picker/DatePicker.scss
```diff
@@ -38,45 +43,41 @@
...
         position: absolute;
         right: 7px;
-        bottom: 10px;
-        display: inline-block;
-        height: 17px;
+        bottom: 8px;
+        display: inline-flex;
+    }
+
+    div.date-picker-open-btn:hover {
```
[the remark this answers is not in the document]
swfree: Nope, removed

src/components/date-picker/DatePicker.scss
```diff
@@ -38,45 +43,41 @@
...
-        height: 14px;
-    }
-
     .date-picker-clear-btn {
         path {
             transition: fill linear .1s;
             fill: $bdl-gray-50;
         }
+
+        &.hide-for-firefox {
```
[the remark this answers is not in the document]
swfree: Good point, I've updated the classname to "is-hidden". There's isn't a plan to update to SUIT in the immediate future, but there's a backlog ticket to overhaul DatePicker (replace Pikaday with an accessible calendar library) so I've added to the ticket a request to update to SUIT style too

src/components/date-picker/DatePicker.tsx
```diff
@@ -21,6 +22,7 @@ import Tooltip, { TooltipPosition, TooltipTheme } from '../tooltip';
 // @ts-ignore flow import
 import { convertDateToUnixMidnightTime } from '../../utils/datetime';
 
+import AccessiblePikaday, { AccessiblePikadayOptions } from './AccessiblePikaday';
```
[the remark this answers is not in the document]
swfree: I moved it around but may not have it right - do you mean grouped with the above imports and alphabetized?

src/components/date-picker/DatePicker.tsx
swfree: I'll look into this

src/components/date-picker/DatePicker.tsx
swfree: I'll work on cleaning this part up

src/components/date-picker/AccessiblePikaday.ts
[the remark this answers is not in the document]
viodragon2: This is something `Pikaday` did and wanted to keep the behavior as similar as possible. I don't think it matters too much TBH.

src/components/date-picker/AccessiblePikaday.ts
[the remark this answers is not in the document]
viodragon2: > do we need to do any cleanup if `options.field` and `options.bound` were actually set?

Probably not, but it would be safe to do it in the component. When `AccessiblePikaday` is used in `DatePicker` https://github.com/box/box-ui-elements/pull/2751/files#diff-8c22974687f38296f0e4c9b61ba406c1a06dd3be54e61892b6a9ad5f546d0478R240 `field` property is deleted before initialized.

For `bound`, I'm not sure why it changed from my original change (https://github.com/box/box-ui-elements/pull/2733/files#diff-9b4042368d8ad5fdcc96cb31f69332780a5693527fb78ec9fbb8737e5af563bfR21-R23), but it seems like we're not overriding the behavior anymore for this case.

> Is it possible that this.el is appended to the document.body twice?

I don't think so? `this.el` is technically a private member of the `Pikaday` class and gets instantiated in its constructor.

src/components/date-picker/DatePicker.tsx
swfree: I've updated this - I also set tabindex=-1 because the button should not be navigable by keyboard: https://github.com/box/box-ui-elements/pull/2751/files#diff-8c22974687f38296f0e4c9b61ba406c1a06dd3be54e61892b6a9ad5f546d0478R540

src/components/date-picker/AccessiblePikaday.ts
```diff
@@ -0,0 +1,74 @@
...
+            document.body.appendChild(this.el);
+
+            this.accessibleFieldEl.addEventListener('blur', this.handleBlur);
+
+            this.hide();
+        }
+    }
+
+    handleBlur = () => {
+        // TODO: Test in IE11
```
alexkrolick, line 32: Did it work in IE11?

src/components/date-picker/DatePicker.tsx
```diff
@@ -150,6 +152,8 @@ export interface DatePickerProps extends WrappedComponentProps {
     placeholder?: string;
     /** Resin tag */
     resinTarget?: string;
+    /** Does the date input meet accessibility standards */
```
alexkrolick, line 155: Can we add a `@deprecated` to the displayFormat field mentioning it won't be supported with the accessible mode turned on?

src/components/date-picker/DatePicker.tsx
```diff
@@ -317,18 +356,23 @@ class DatePicker extends React.Component<DatePickerProps> {
...
             event.preventDefault();
         }
 
-        if (isTextInputAllowed && event.key === ENTER_KEY) {
+        if ((isTextInputAllowed || (isAccessible && !this.canUseDateInputType)) && event.key === ENTER_KEY) {
+            event.preventDefault();
+        }
+
+        // In Firefox, enter & spacebar open up the browser's default date picker. Suppress it.
+        if (Browser.isFirefox() && isAccessible && (event.key === ENTER_KEY || event.keyCode === 32)) {
```
alexkrolick, line 375: Would it be bad to always preventDefault, regardless of browser?

src/components/date-picker/DatePicker.tsx
```diff
@@ -367,28 +434,72 @@ class DatePicker extends React.Component<DatePickerProps> {
...
     };
 
+    formatDisplayDateType = (date?: Date | null): string => {
+        // Input type "date" only accepts the format YYYY-MM-DD
+        return date ? date.toISOString().slice(0, 10) : '';
+    };
+
+    parseDisplayDateType = (dateString?: string | null): Date | null => {
+        if (dateString) {
+            // Calling new Date('YYYY-MM-DD') without 'T00:00:00' yields undesired results:
```
alexkrolick, line 495: Do we need the timezone part of the ISO string too?

src/components/date-picker/DatePicker.tsx
```diff
@@ -443,17 +598,27 @@ class DatePicker extends React.Component<DatePickerProps> {
...
+        if (isAccessible && this.canUseDateInputType) {
+            onChangeAttr = { onChange: this.handleOnChange };
+        } else if (isTextInputAllowed || (isAccessible && !this.canUseDateInputType)) {
+            onChangeAttr = {};
+        } else {
+            onChangeAttr = { onChange: noop };
+        }
+        // "name" prop is required for pattern validation to be surfaced on form submit. See components/form-elements/form/Form.js
+        const additionalAttrs =
+            isAccessible && !this.canUseDateInputType ? { name, pattern: '\\d{4}-\\d{2}-\\d{2}' } : {};
```
alexkrolick, line 619: Is there anything we should do when we are in text fallback mode to help make sure the date is not in ddmm order instead of mmdd? Such as validate the 1-31 and 1-12 ranges?

src/elements/content-sidebar/__tests__/AddTaskButton.test.js
```diff
@@ -2,6 +2,21 @@ import React from 'react';
 import { mount } from 'enzyme';
 import { AddTaskButtonComponent as AddTaskButton } from '../AddTaskButton';
 
+jest.mock('../../../components/date-picker/DatePicker', () => props => {
```
alexkrolick, line 5: Will every test that renders a datepicker need a mock like this?

src/components/date-picker/AccessiblePikaday.ts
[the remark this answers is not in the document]
swfree: @viodragon2 I removed the click & focus listeners from your original change because the calendar popup should only open by clicking on the calendar icon. Focusing & clicking on the text box should not open the calendar.

src/elements/content-sidebar/__tests__/AddTaskButton.test.js
swfree: I believe this was only necessary because `mount` was used for rendering in the test. So for consumers of date picker that use `mount` in their tests, yes we'll likely need to add a mock like this

src/components/date-picker/AccessiblePikaday.ts
tjuanitas: It does work in IE11 but will never be called. This method is only used for a blur event listener to `this.accessibleFieldEl` which will not be set for IE11.

src/components/date-picker/DatePicker.tsx
tjuanitas: The timezone part is not required. It's only if the string is "date-only" that the issue occurs.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#parameters (scroll to dateString)

src/components/date-picker/DatePicker.scss
```diff
@@ -91,6 +92,13 @@
         }
     }
 
+    .date-picker-clear-btn:focus,
+    .date-picker-open-btn:focus {
+        margin: -1px;
+        border: 1px solid $bdl-box-blue;
+        border-radius: 4px;
```
alexkrolick, line 99:
```suggestion
        border-radius: $bdl-border-radius-size;
```

src/utils/Browser.js
```diff
@@ -41,6 +41,16 @@ class Browser {
         return /Trident/i.test(Browser.getUserAgent());
     }
 
+    /**
+     * Returns whether browser is Firefox.
+     *
+     * @return {boolean} Whether browser is Firefox
+     */
+    static isFirefox() {
```
alexkrolick, line 49: Is this still used anywhere?

alexkrolick (approved):

ConradJChan (approved):

PR #2754, opened 2021-11-17
test(screenshots): update screenshot tests

alexkrolick:
1. Stabilize the `test:images` command:
  a. Don't restart the webpack storybook watcher on __tests__ changes
  b. Don't manually launch puppeteer in each test, it's already launched by jest-puppeteer
  c. Fix flaky test assertions
  d. Make sure network connections are idle before executing tests. This makes sure font assets are loaded.
2. Update images  

There is still some instability due to screenshotting focused fields and having a flashing cursor.


jstoffan (approved):
I spot-checked some of the screenshots. It looks like some components did indeed change visually, but I'm not seeing any change in some others. That said, I also don't see any that are obviously wrong, so LGTM.

macrotim (approved):

swfree (approved):

PR #2760, opened 2021-11-22
feat(theme): Add breaking change commit for dejablue redesign

swfree:
BREAKING CHANGE: rename $bdl-border-radius-large-size to $bdl-border-radius-size-large

See https://github.com/box/box-ui-elements/pull/2749 to learn more


alexkrolick (dismissed):
Approving  - noop commit to fix incorrect breaking change syntax in previous PR.

alexkrolick (approved):

PR #2764, opened 2021-11-24
fix: remove Adobe sign integration

sujaygarlanka:
Removed the ability to show the Adobe Sign option in the Open with UI Element. Its icons and tests have been removed.


swfree:
@sujaygarlanka can you update the PR description with some more details about this change?

src/elements/content-open-with/__tests__/ContentOpenWith.test.js
```diff
@@ -299,7 +299,7 @@ describe('elements/content-open-with/ContentOpenWith', () => {
 
             displayIntegration = {
                 appIntegrationId: '1',
-                displayName: 'Adobe Sign',
+                displayName: 'Google Docs',
```
swfree, line 302: It looks like this file has a few more mentions of "adobe" that could be removed

test/integration/open-with/OpenWith.int.test.js
```diff
@@ -11,7 +11,7 @@ describe('OpenWith', () => {
     });
 
     it('A single integration', () => {
-        cy.route('GET', '**/files/*/open_with_integrations', 'fixture:open-with/integration-adobe');
+        cy.route('GET', '**/files/*/open_with_integrations', 'fixture:open-with/integration-box-edit');
```
swfree, line 14: Looks like you can remove this file then, since its no longer used: https://github.com/box/box-ui-elements/blob/master/test/fixtures/open-with/integration-adobe.json

src/elements/content-open-with/__tests__/ContentOpenWith.test.js
sujaygarlanka: Adobe is still supported because I see the ability to integrate with adobe acrobat, so that's why I left it in. Should I remove it?

swfree: Ah gotcha, I'd been assuming that only Adobe Sign is associated with Content Open With, and that ADOBE_INTEGRATION_ID was referring to Adobe Sign. It would be good to confirm with @box/partners whether "Adobe" here is referring to adobe acrobat or adobe sign.

swfree (dismissed):
lgtm! Just one open question re: adobe acrobat vs adobe sign

macrotim (dismissed):
Approving the change under scripts/

jstoffan (dismissed):

src/elements/content-open-with/__tests__/ContentOpenWith.test.js
ctuckz: I don't think we ever added adobe acrobat to Content Open With, but it is available in the webapp. It looks like all the usages of Adobe in this file are referencing mock data, and using fake IDs. You could remove it, but then we'd lose test cases for when there's more than 2 integrations. It might be better to rename it to a fake app name to retain the test coverage for when there's > 1 integration

ctuckz (commented):
Has there already been a backend change to stop Adobe Sign from being returned from the API that populates the integration list?

ctuckz (approved):

swfree (approved):

jstoffan (approved):

alexkrolick (approved):

PR #2766, opened 2021-12-03
fix(presence): Make more accessible by adding appropriate roles

ConradJChan:
In the `Flyout` component, added the `role="button"` to the button child element. Added `aria-hidden="true"` to the `PresenceAvatar` used in the `PresenceAvatarList` so that it is ignored by screen readers.

TODO:
- [x] update unit tests

### DIFF
--- a/i18n/en-US.properties
+++ b/i18n/en-US.properties
@@ -1060,8 +1060,6 @@ boxui.notification.clearNotification = Clear Notification
 boxui.presence.accessedInTheLastMinute = Viewed less than a minute ago
 # Description for someone who is currently viewing or editing a document
 boxui.presence.activeNow = Active now
-# Description for the label of the avatar for a given user
-boxui.presence.avatarLabel = Avatar for {name}
 # Description for when someone last commented on a document less than a minute ago
 boxui.presence.commentedIntheLastMinute = Commented less than a minute ago
 # Text on button to get shared link for the item
@@ -1088,8 +1086,8 @@ boxui.presence.timeSinceLastCommented = Commented {timeAgo}
 boxui.presence.timeSinceLastModified = Edited {timeAgo}
 # Description for when someone last previewed a document, {timeAgo} is a relative time like 2 months ago
 boxui.presence.timeSinceLastPreviewed = Previewed {timeAgo}
-# Text on button to toggle the presence overlay
-boxui.presence.toggleOverlayLabel = Toggle Presence Overlay
+# Description of the button to toggle the presence overlay
+boxui.presence.toggleButtonLabel = Presence Activity
 # Text on the add filter button, on click generates another filter row
 boxui.queryBar.addFilterButtonText = + Add Filter
 # Text on the apply filter button, on click applies the filters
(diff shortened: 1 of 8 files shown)


i18n/en-US.properties
```diff
@@ -1086,6 +1088,8 @@ boxui.presence.timeSinceLastCommented = Commented {timeAgo}
 boxui.presence.timeSinceLastModified = Edited {timeAgo}
 # Description for when someone last previewed a document, {timeAgo} is a relative time like 2 months ago
 boxui.presence.timeSinceLastPreviewed = Previewed {timeAgo}
+# Text on button to toggle the presence overlay
+boxui.presence.toggleOverlayLabel = Toggle Presence Overlay
```
jstoffan, line 1092: I think the guidance is to just use the name of the element and not the action/instructions. So, just `Presence Overlay` in this case.

src/features/presence/PresenceAvatar.js
```diff
@@ -26,6 +42,7 @@ const PresenceAvatar = ({
         onFocus={onFocus}
         onMouseEnter={onMouseEnter}
         onMouseLeave={onMouseLeave}
+        role="figure"
```
jstoffan, line 45: Is this correct for an interactive element? I think we've been removing the use of `figure` in most places, as well. It also looks like the line below has a condition of `isDropDownAvatar`.

src/features/presence/PresenceAvatar.js
```diff
@@ -34,20 +51,4 @@ const PresenceAvatar = ({
...
-    onBlur: PropTypes.func,
-    isDropDownAvatar: PropTypes.bool,
-};
-
-PresenceAvatar.defaultProps = {
-    isActive: false,
-};
-
-export default PresenceAvatar;
+export default injectIntl(PresenceAvatar);
```
jstoffan, line 54: Do we have any cases where we're using the ref of this component? If so, do they still work properly?

src/features/presence/PresenceAvatarList.tsx
```diff
@@ -5,6 +5,7 @@ import noop from 'lodash/noop';
 import PresenceAvatar from './PresenceAvatar';
 import PresenceAvatarTooltipContent from './PresenceAvatarTooltipContent';
 import Tooltip, { TooltipPosition } from '../../components/tooltip';
+
```
jstoffan, line 8: Can we revert this whitespace-only change?

src/features/presence/messages.js
```diff
@@ -6,6 +6,11 @@ const messages = defineMessages({
         description: 'Description for someone who is currently viewing or editing a document',
         id: 'boxui.presence.activeNow',
     },
+    avatarLabel: {
+        defaultMessage: 'Avatar for {name}',
+        description: 'Description for the label of the avatar for a given user',
```
jstoffan, line 11: Is this a description of the label or the text for the label?

src/features/presence/PresenceAvatar.js
ConradJChan: Good callout, I poked around a few other repos and didn't find any references so we should be good here

i18n/en-US.properties
```diff
@@ -1086,6 +1086,8 @@ boxui.presence.timeSinceLastCommented = Commented {timeAgo}
 boxui.presence.timeSinceLastModified = Edited {timeAgo}
 # Description for when someone last previewed a document, {timeAgo} is a relative time like 2 months ago
 boxui.presence.timeSinceLastPreviewed = Previewed {timeAgo}
+# Description of the button to toggle the presence overlay
+boxui.presence.toggleButtonLabel = Presence Activity
```
alexkrolick, line 1090: Would this string make sense in context? Something like Recent Activity (the title of the list that opens up) might be more clear.

jstoffan (approved):

PR #2772, opened 2021-12-09
fix(activity-sidebar): delete and cancel button misalignment

karelee7:
Bug fix for incorrect vertical stacking of the delete and cancel buttons on the confirmation modal that pops up when the user tries to delete a comment. Now the buttons are aligned horizontally.

### Before
COXP-9313

### After
**Example of Fix on Chrome:**
Screen Shot 2021-12-09 at 2 33 50 PM

**Example of Fix on IE11:**
Screen Shot 2021-12-09 at 11 23 43 AM
Fix on IE11

### DIFF
--- a/src/elements/content-sidebar/activity-feed/common/delete-confirmation/DeleteConfirmation.scss
+++ b/src/elements/content-sidebar/activity-feed/common/delete-confirmation/DeleteConfirmation.scss
@@ -16,11 +16,11 @@
 }
 
 // overrides for confirm/cancel button spacing
-.bcs-DeleteConfirmation-delete,
-.bcs-DeleteConfirmation-cancel {
+.btn.bcs-DeleteConfirmation-delete,
+.btn.bcs-DeleteConfirmation-cancel {
     margin-left: 0;
 }
 
-.bcs-DeleteConfirmation-cancel {
+.btn.bcs-DeleteConfirmation-cancel {
     margin-right: 10px;
 }


src/elements/content-sidebar/activity-feed/common/delete-confirmation/DeleteConfirmation.js
```diff
@@ -56,7 +56,7 @@ class DeleteConfirmation extends React.Component<Props> {
                 <div className="bcs-DeleteConfirmation-promptMessage">
                     <FormattedMessage {...message} />
                 </div>
-                <div>
+                <div className="bcs-DeleteConfirmation-btn-container">
```
ConradJChan, line 59: Instead of `*-btn-container` can we name it `bcs-DeleteConfirmation-actions`?

ConradJChan:
For the commit message we may also want to do `fix(activity-sidebar)` instead of preview

src/elements/content-sidebar/activity-feed/common/delete-confirmation/DeleteConfirmation.scss
```diff
@@ -15,6 +15,10 @@
     margin-bottom: 20px;
 }
 
+.bcs-DeleteConfirmation-btn-container {
+    display: flex;
```
alexkrolick, line 19: Can you check this in a language with long strings like French or German to make sure the buttons wrap as a block and align correctly when that happens?

src/elements/content-sidebar/activity-feed/common/delete-confirmation/DeleteConfirmation.js
karelee7: Updated! :)

src/elements/content-sidebar/activity-feed/common/delete-confirmation/DeleteConfirmation.scss
karelee7: Sure! Just to clarify, the correct behavior for long text is to have the button become taller and the text to wrap around, is that right? (while still maintaining the side by side alignment)

alexkrolick: I think it's supposed to wrap to two lines, keeping the normal button height

karelee7: Ahh I see, so the entire button wraps around, I wonder if that's why the bug originally appeared...

alexkrolick:
```suggestion
    display: flex;
    flex-wrap: wrap;
```

Screen Shot 2021-12-10 at 1 19 45 PM
Screen Shot 2021-12-10 at 1 18 31 PM

karelee7: Thank you for the image and suggestion! I actually found the root cause of the bug, turned out there was already some styling fixes in the repo, but they were being overwritten so I fixed that! :)

karelee7: **Example of German buttons:**
Screen Shot 2021-12-10 at 12 39 21 PM

**Example of Albanian buttons:**
Screen Shot 2021-12-10 at 1 55 23 PM

alexkrolick (dismissed):

ConradJChan (approved):

alexkrolick (approved):

PR #2779, opened 2021-12-20
feat: add conditional and logic to show new upgrade copy

jackiejou:
adding new prop showNewUpgradeText

OLD COPY
image

NEW COPY
image


src/features/unified-share-modal/UnifiedShareForm.js
```diff
@@ -540,28 +540,53 @@ class UnifiedShareForm extends React.Component<USFProps, State> {
...
-                                type="button"
-                                {...upgradeLinkProps}
-                            >
-                                <FormattedMessage {...messages.upgradeGetMoreAccessControlsLink} />
-                            </PlainButton>
-                        ),
-                    }}
-                    {...messages.upgradeGetMoreAccessControlsDescription}
-                />
+                <FormattedMessage values={messageValuesObj} {...description} />
```
jasminevle, line 589: is this convention to grab the FormattedMessage props like this ? Or would it be better to save the FormattedMessage as a parameter once we've determined which message we want and use that parameter here instead?

jasminevle (approved):
LGTM, but wouldn't mind someone who's more familiar with the USM to take a look!

src/features/unified-share-modal/flowTypes.js
```diff
@@ -393,6 +393,8 @@ export type USFProps = BaseUnifiedShareProps & {
     sharedLinkLoaded: boolean,
     /** Whether the FTUX tooltip should be rendered */
     shouldRenderFTUXTooltip: boolean,
+    /** Whether the new upgrade text should be rendered */
+    showNewUpgradeText: boolean,
```
ctuckz, line 397: nit - it looks like this should be an optional type since the destructure in UnifiedShareForm is setting a default value. Also depending on how you're planning to populate this value, it might not be possible for it to be set when the USM is running as a standalone UI element and needs to make v2 API calls to do stuff

src/features/unified-share-modal/UnifiedShareForm.js
```diff
@@ -540,28 +540,55 @@ class UnifiedShareForm extends React.Component<USFProps, State> {
...
-                                type="button"
-                                {...upgradeLinkProps}
-                            >
-                                <FormattedMessage {...messages.upgradeGetMoreAccessControlsLink} />
-                            </PlainButton>
-                        ),
-                    }}
-                    {...messages.upgradeGetMoreAccessControlsDescription}
-                />
+                {showNewUpgradeText ? (
```
ctuckz, line 550: Just curious, what's the plan for this prop in the future? It looks like it's not being explicitly set right now so it's always false

jackiejou: it is part of an experiment we are running. Certain buckets will have this boolean set to true. After the experiment finishes, we will roll out the winner and remove the extra props that were created.

ctuckz: Ah ok, thanks for explaining!

ctuckz (approved):

PR #2783, opened 2021-12-21
feat: changes for an updated error message for version history errors

jlarsen-box:
Adding the code to show an updated version history error for free users that includes an upsell, linking users to the upgrade modal.
Screen Shot 2021-12-21 at 1 38 47 PM
This new error and upsell will be shown if showVersionUpsell property is true. this will be true for users that are in the treatment_H treatment bucket, and will be set via the EUA when creating this component.


src/elements/content-sidebar/versions/VersionsSidebar.js
```diff
@@ -12,24 +12,36 @@ import messages from './messages';
...
 import './VersionsSidebar.scss';
 
 const MAX_VERSIONS = DEFAULT_FETCH_END;
 
 type Props = {
     error?: MessageDescriptor,
+    errorUpsell: boolean,
     fileId: string,
     isLoading: boolean,
+    onUpgradeCTAClick: openUpgradePlanModal,
```
jstoffan, line 29: Can we make this more generic, like `onUpgrade: () => void` so the element isn't aware of how this callback will be used?

src/elements/content-sidebar/versions/VersionsSidebar.js
```diff
@@ -49,8 +61,26 @@ const VersionsSidebar = ({ error, isLoading, parentName, versions, ...rest }: Pr
         >
             <LoadingIndicatorWrapper className="bcs-Versions-content" crawlerPosition="top" isLoading={isLoading}>
                 {showError && (
-                    <InlineError title={<FormattedMessage {...messages.versionServerError} />}>
+                    <InlineError
+                        title={
+                            errorUpsell ? (
```
jstoffan, line 66: I think this inline error can be used for other API call failures. It also doesn't seem like `error` and `errorUpsell` have much in common. Can set `error` to undefined in the container and pass in a `showUpgrade` prop, instead?

src/elements/content-sidebar/versions/VersionsSidebar.js
```diff
@@ -49,8 +61,26 @@ const VersionsSidebar = ({ error, isLoading, parentName, versions, ...rest }: Pr
...
+                                <FormattedMessage {...messages.versionServerErrorUpsell} />
+                            ) : (
+                                <FormattedMessage {...messages.versionServerError} />
+                            )
+                        }
+                    >
                         <FormattedMessage {...error} />
+                        {errorUpsell && (
```
jstoffan, line 74: Can we standardize on either "upgrade" (preferred) or "upsell"?

src/elements/content-sidebar/versions/VersionsSidebar.js
```diff
@@ -49,8 +61,26 @@ const VersionsSidebar = ({ error, isLoading, parentName, versions, ...rest }: Pr
...
+                            <PlainButton
+                                onClick={onUpgradeCTAClick}
+                                data-resin-target="jessica-to-update"
```
jstoffan, line 77: Can we omit this line until we have a finalized target?

src/elements/content-sidebar/versions/VersionsSidebar.js
```diff
@@ -49,8 +61,26 @@ const VersionsSidebar = ({ error, isLoading, parentName, versions, ...rest }: Pr
...
+                                type="button"
+                                className="upgradeNowButton"
```
jstoffan, line 79: Please use SUIT style naming conventions to stay consistent with the other style rules. Maybe `bcs-Versions-upgrade`?

src/elements/content-sidebar/versions/__tests__/VersionsSidebar.test.js
```diff
@@ -28,6 +28,13 @@ describe('elements/content-sidebar/versions/VersionsSidebar', () => {
             expect(wrapper).toMatchSnapshot();
         });
 
+        test('should show an upsell inline error if the errorUpsell prop is provided', () => {
+            const wrapper = getWrapper({ error: messages.versionFetchErrorUpsell, errorUpsell: true, versions: [] });
+
+            expect(wrapper.exists(InlineError)).toBe(true);
+            expect(wrapper).toMatchSnapshot();
```
jstoffan, line 35: We've been trying to reduce our dependence on snapshot testing. Can we make explicit assertions, instead?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -286,6 +291,8 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
                 onPromote={this.handleActionPromote}
                 onRestore={this.handleActionRestore}
                 parentName={parentName}
+                errorUpsell={this.props.showVersionUpsell}
```
jstoffan, line 294: Can we destructure these props, similar to the existing props in this file?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -165,9 +169,10 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
 
     handleFetchError = (): void => {
         this.setState({
-            error: messages.versionFetchError,
+            error: this.props.showVersionUpsell ? messages.versionFetchErrorUpsell : messages.versionFetchError,
             isLoading: false,
             isWatermarked: false,
+            onUpgradeCTAClick: this.props.onUpgradeCTAClick,
```
jstoffan, line 175: We should almost always try to avoid copying props to state. Is this unused?

src/elements/content-sidebar/versions/VersionsSidebar.scss
```diff
@@ -37,3 +37,8 @@
         padding-bottom: 20px;
     }
 }
+
+.upgradeNowButton {
+    color: #0061d5;
+    font-weight: bold !important;
```
jstoffan, line 43: We should almost always avoid using `!important`.

src/elements/content-sidebar/versions/VersionsSidebar.scss
```diff
@@ -37,3 +37,8 @@
         padding-bottom: 20px;
     }
 }
+
+.upgradeNowButton {
+    color: #0061d5;
```
jstoffan, line 42: Can we use an existing color variable for this value?

jstoffan (changes requested):

src/elements/content-sidebar/versions/VersionsSidebar.js
jstoffan: We could also consider passing in `errorTitle` from the container, which would simplify this quite a bit.

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
jlarsen-box: it is, will update to remove from state.

src/elements/content-sidebar/versions/VersionsSidebar.js
jlarsen-box: I am renaming my variables to make them clearer. We have the upgrade button, which should really be the only place I use upgrade. and then the rest is an upsell, which is asking people to update.

src/elements/content-sidebar/versions/VersionsSidebar.scss
jlarsen-box: So I haven't been able to get the font-weight to be bold without the !important. I suspect this is because I am using the <PlainButton> component which creates a button with the class `btn-plain` which sets `font: inherit;`. I will check with my team on how to accomplish overriding that inherit without an `!important`

src/elements/content-sidebar/versions/VersionsSidebar.js
jlarsen-box: I added an errorTitle, and cleaned up variable names, I will be using the showUpgrade approach with the other requirements, as those will no longer be errors from the API.

src/elements/content-sidebar/versions/VersionsSidebar.js
```diff
@@ -21,19 +22,31 @@ const MAX_VERSIONS = DEFAULT_FETCH_END;
 
 type Props = {
     error?: MessageDescriptor,
+    errorTitle?: MessageDescriptor,
     fileId: string,
     isLoading: boolean,
+    onUpgradeCTAClick: () => void,
```
laughnan, line 28: This is an optional prop, right?

```suggestion
    onUpgradeCTAClick?: () => void,
```

src/elements/content-sidebar/versions/VersionsSidebar.js
```diff
@@ -21,19 +22,31 @@ const MAX_VERSIONS = DEFAULT_FETCH_END;
...
     parentName: string,
+    showVersionHistoryErrorUpsell: boolean,
```
laughnan, line 30: This is an optional prop, right?

```suggestion
    showVersionHistoryErrorUpsell?: boolean,
```

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -27,18 +27,21 @@ type Props = {
     hasSidebarInitialized?: boolean,
     history: RouterHistory,
     match: Match,
+    onUpgradeCTAClick: () => void,
```
laughnan, line 30: Optional prop? 

```suggestion
    onUpgradeCTAClick?: () => void,
```

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -163,9 +166,11 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
         this.mergeResponse(data);
     };
 
-    handleFetchError = (): void => {
+    handleFetchError = (error: Object): void => {
+        const showVersionErrorWithUpsell = this.props.showVersionErrorWithUpsell && error.status === 403;
```
laughnan, line 170: Can you de-structure the prop first (it's optional, right?). Untested code approach: 
```suggestion
        const { showVersionErrorWithUpsell }  = this.props;
        const shouldShowVersionErrorWithUpsell = !!showVersionErrorWithUpsell && error.status === 403;
```

src/elements/content-sidebar/versions/VersionsSidebar.js
laughnan: Looking back through, I guess this probably isn't. Maybe at the container layer, but not here.

src/elements/content-sidebar/versions/VersionsSidebar.scss
laughnan: +1

src/elements/content-sidebar/versions/VersionsSidebar.scss
laughnan: Yeah let's dig more into what CSS classes we can use to target it @jlarsen-box and we'll figure it out.

src/elements/content-sidebar/versions/VersionsSidebar.scss
```diff
@@ -37,3 +37,8 @@
         padding-bottom: 20px;
     }
 }
+
+button {
```
jstoffan, line 41: This will affect all buttons globally. Can we use `.bcs-Versions-upgrade`?

src/elements/content-sidebar/versions/VersionsSidebar.js
```diff
@@ -49,8 +62,18 @@ const VersionsSidebar = ({ error, isLoading, parentName, versions, ...rest }: Pr
         >
             <LoadingIndicatorWrapper className="bcs-Versions-content" crawlerPosition="top" isLoading={isLoading}>
                 {showError && (
-                    <InlineError title={<FormattedMessage {...messages.versionServerError} />}>
+                    <InlineError title={<FormattedMessage {...errorTitle} />}>
                         <FormattedMessage {...error} />
+                        {showVersionHistoryErrorUpsell && (
```
jstoffan, line 67: Can we make `onUpgradeCTAClick` nullable and use its presence/absence to control the visibility of the button? We would be able to eliminate the `showVersionHistoryErrorUpsell` prop entirely, if so.

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
jstoffan, line 30: Please change to `onUpgradeClick` to avoid reference to visual treatment (CTA).

src/elements/content-sidebar/versions/__tests__/VersionsSidebar.test.js
```diff
@@ -22,12 +23,29 @@ describe('elements/content-sidebar/versions/VersionsSidebar', () => {
...
+            const wrapper = getWrapper({
+                error: messages.versionNotAvailable,
+                errorTitle: messages.versionAccessError,
+                showVersionHistoryErrorUpsell: true,
+                versions: [],
+            });
+
+            expect(wrapper.exists(InlineError)).toBe(true);
+            const errorButton = wrapper.find(PlainButton);
+            expect(errorButton.length).toBe(1);
```
jstoffan, line 46: Can be simplified to `expect(wrapper.exists(PlainButton)).toBe(true)`

src/elements/content-sidebar/versions/__tests__/VersionsSidebarContainer.test.js
```diff
@@ -69,7 +69,7 @@ describe('elements/content-sidebar/versions/VersionsSidebarContainer', () => {
         });
     });
 
-    describe('handleActionDelete', () => {
+    describe(' Delete', () => {
```
jstoffan, line 72: Is this change needed? If not, can we revert it?

src/elements/content-sidebar/versions/__tests__/VersionsSidebarContainer.test.js
```diff
@@ -191,10 +191,26 @@ describe('elements/content-sidebar/versions/VersionsSidebarContainer', () => {
         test('should set state to default values with error message', () => {
             const wrapper = getWrapper();
 
-            wrapper.instance().handleFetchError();
+            wrapper.instance().handleFetchError({ status: 500 });
 
             expect(wrapper.state()).toEqual({
```
jstoffan, line 196: Optional, but we can change `toEqual` to `toMatchObject` and only check the props we care about to reduce the surface area of these tests.

src/elements/content-sidebar/versions/__tests__/__snapshots__/VersionsSidebar.test.js.snap
```diff
@@ -37,14 +37,6 @@ exports[`elements/content-sidebar/versions/VersionsSidebar render should show an
         id="be.sidebarVersions.fetchError"
       />
     </InlineError>
-    <div
```
jstoffan, line 40: This should still be a supported experience for files without versions. Can we investigate why this snapshot was removed?

jlarsen-box: So this is why this was removed: 
This message shows up when there are no versions, which is what it was intended to, I think. It was also displaying when there was an error from the api. I fixed it so that it only appears when the file has no previous versions, which means that this test, which is testing what is displayed with an error, no longer has the message. When a file has no versions, it doesn't result in an error, just an empty versions list.

src/elements/content-sidebar/versions/__tests__/VersionsSidebarContainer.test.js
jlarsen-box: 👍  That was an accidental change I think. I have reverted it

src/elements/content-sidebar/versions/VersionsSidebar.js
jlarsen-box: Ah, yes that makes sense! Making that change now!

src/elements/content-sidebar/versions/VersionsSidebar.scss
jlarsen-box: did it!

src/elements/content-sidebar/versions/VersionsSidebar.js
jlarsen-box: yup, it is optional at the container, but always sets it here. Is that ok? Or should I make it optional everywhere?

jlarsen-box: Actually I removed this here based on @jstoffan suggestion below.

src/elements/content-sidebar/versions/__tests__/VersionsSidebarContainer.test.js
jlarsen-box: I sure can!

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -286,6 +292,7 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
                 onPromote={this.handleActionPromote}
                 onRestore={this.handleActionRestore}
                 parentName={parentName}
+                onUpgradeClick={onUpgradeClick}
```
jstoffan, line 295: Minor, but can we alphabetize this prop to stay consistent with the others?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -163,9 +166,12 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
         this.mergeResponse(data);
     };
 
-    handleFetchError = (): void => {
+    handleFetchError = (error: Object): void => {
+        const { showVersionErrorWithUpsell } = this.props;
+        const shouldShowVersionErrorWithUpsell = !!showVersionErrorWithUpsell && error.status === 403;
         this.setState({
-            error: messages.versionFetchError,
+            error: shouldShowVersionErrorWithUpsell ? messages.versionNotAvailable : messages.versionFetchError,
```
jstoffan, line 173: Do we need this boolean or can we key off the existence of `onUpgradeClick`, similar to the underlying components?

src/elements/content-sidebar/versions/VersionsSidebar.scss
```diff
@@ -37,3 +37,8 @@
         padding-bottom: 20px;
     }
 }
+
+.btn-plain.bcs-Versions-upgrade {
```
jstoffan, line 41: Is the `.btn-plain` portion of this needed? We typically see the something like the following to override the default plain button styling:

```
.bcs-Versions-upgrade {
    &,
    &:hover,
    &:active {
        color: $bdl-box-blue;
        font-weight: bold;
    }
}

File: src/elements/content-sidebar/versions/VersionsSidebarContainer.js (hunk 22 above)
Comment (jlarsen-box) [reply src/elements/content-sidebar/versions/VersionsSidebarContainer.js:173]: I thought it was but I thought through the three different scenarios where the onUpgradeClick will not be null and with checking for the error code, and the onUpgradeClick existing we can in fact infer the correct bucket and omit that prop!

File: src/elements/content-sidebar/versions/VersionsSidebarContainer.js (hunk 21 above)
Comment (jlarsen-box) [reply src/elements/content-sidebar/versions/VersionsSidebarContainer.js:295]: yup yup!

File: src/elements/content-sidebar/versions/VersionsSidebar.scss (hunk 23 above)
Comment (jlarsen-box) [reply src/elements/content-sidebar/versions/VersionsSidebar.scss:41]: So the default if I change nothing is no bold ever. If I have what you suggest there I have blue text, and it is bold only when I hover? 😭  Looking into it hahaha

**Review by laughnan — COMMENTED** [review]
Great work on incorporating feedback here, @jlarsen-box ! I don't think I have anything beyond what Jared commented (I didn't look further into that CSS feedback).

File: src/elements/content-sidebar/versions/VersionsSidebar.js [hunk 24]
```diff
@@ -21,19 +22,21 @@ const MAX_VERSIONS = DEFAULT_FETCH_END;
...
     versions: Array<BoxItemVersion>,
 };
 
-const VersionsSidebar = ({ error, isLoading, parentName, versions, ...rest }: Props) => {
+const VersionsSidebar = ({ error, errorTitle, onUpgradeClick, isLoading, parentName, versions, ...rest }: Props) => {
     const showLimit = versions.length >= MAX_VERSIONS;
     const showVersions = !!versions.length;
-    const showEmpty = !isLoading && !showVersions;
     const showError = !!error;
+    const showEmpty = !isLoading && !showVersions && !onUpgradeClick;
```
jstoffan, line 39: Minor, but can we move this back to its prior location?

jstoffan (commented):
Change looks good to me, but it also looks like end-to-end tests are failing.

jstoffan (approved):
Looks good! Thanks very much for working through the feedback.

src/elements/content-sidebar/versions/VersionsSidebar.scss
alexkrolick: I'm surprised if it works without that. `font-weight: normal` is specified for the hover state.

https://github.com/box/box-ui-elements/blob/master/src/styles/common/_buttons.scss#L137-L153

laughnan (approved):

tjuanitas:
@Mergifyio refresh

PR #2785, opened 2021-12-22
feat: Add new upgrade inline notice

jackiejou:
WITHOUT INLINE NOTICE
image

WITH INLINE NOTICE
image


alexkrolick:
Is there a need to keep both options?

jackiejou:
> Is there a need to keep both options?

@alexkrolick  yes we are going to run an experiment with both options. After PM and data team return with the results, we will implement the winner.

rdacanay (approved):

ctuckz (approved):

PR #2786, opened 2021-12-22
fix: Fix className property passed to ImageTooltip component

wpiesiak:
I noticed ImageTooltip component did not pass className property.
Instead of provided value component render 'className' string. 

Change has been tested and providing expected result.

### DIFF
--- a/src/components/image-tooltip/ImageTooltip.tsx
+++ b/src/components/image-tooltip/ImageTooltip.tsx
@@ -31,8 +31,7 @@ const ImageTooltip = ({ children, className, content, image, title, ...otherTool
         </ImageTooltipContent>
     );
 
-    const imageTooltipClasses = classNames('bdl-ImageTooltip', {
-        className,
+    const imageTooltipClasses = classNames('bdl-ImageTooltip', className, {
         'bdl-is-image-loaded': isImageLoaded,
     });
 


src/components/image-tooltip/ImageTooltip.tsx
```diff
@@ -31,8 +31,7 @@ const ImageTooltip = ({ children, className, content, image, title, ...otherTool
         </ImageTooltipContent>
     );
 
-    const imageTooltipClasses = classNames('bdl-ImageTooltip', {
-        className,
+    const imageTooltipClasses = classNames('bdl-ImageTooltip', className, {
```
megansmith-box, line 34: I think we should either make className default to an empty string or modify the object passed to classNames to include something like this:
```
{
    [className]: !!className,
}

laughnan: I think `classnames()` ignores `undefined` values (so we shouldn't need that check IMO). Wiola's change makes sense to me though we might want to see how other BUIE components are handling it.

wpiesiak: I've checked few components in BUIE and they mostly share my approach. 
Furthermore when I tried to modify object passed to classNames, I've got an error from TS:
Screenshot 2021-12-23 at 10 38 39

laughnan (approved):

src/components/image-tooltip/ImageTooltip.tsx
megansmith-box: Makes sense!! Thank you for checking

megansmith-box (approved):

PR #2794, opened 2022-01-06
fix: fix collaborator access message copy for international translations

jackiejou:
copies remain the same
image

### DIFF
--- a/src/features/unified-share-modal/UnifiedShareForm.js
+++ b/src/features/unified-share-modal/UnifiedShareForm.js
@@ -542,7 +542,7 @@ class UnifiedShareForm extends React.Component<USFProps, State> {
     }
 
     renderUpgradeLinkDescription() {
-        const { openUpgradePlanModal = () => {}, showNewUpgradeText = false, trackingProps = {} } = this.props;
+        const { intl, openUpgradePlanModal = () => {}, showNewUpgradeText = false, trackingProps = {} } = this.props;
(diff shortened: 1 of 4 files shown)


src/features/unified-share-modal/UnifiedShareForm.js
```diff
@@ -599,30 +600,28 @@ class UnifiedShareForm extends React.Component<USFProps, State> {
...
-                                target="_blank"
-                            >
-                                <FormattedMessage {...messages.collaboratorAccessLink} />
-                            </Link>
-                        ),
-                    }}
-                    {...messages.upgradeCollaboratorAccessDescription}
-                />
+                <FormattedCompMessage
+                    id="boxui.unifiedShare.upgradeCollaboratorAccessDescription"
```
laughnan, line 604: Shouldn't this generate a line in `i18n/en-US.properties` to ensure it gets exported for translation?

ehoogerbeets: No because the Box webpack plugin that extracts it is not open-source, so we cannot put it in the build scripts for BUIE. Instead, we have a special build in the globalization jenkins job which augments the BUIE build and installed this private webpack plugin. That extracts the string and gets it translated through the normal route and finally checks it all in to the translated properties files.

ehoogerbeets (approved):

Issue #2797, opened 2022-01-13
Choose button remains active when user unselect a folder


fpan225:
Please fill out the following template so we can reproduce and fix your issue as quickly as possible!

**Note:** if your issue includes a potential security vulnerability, please do not file it here. Instead, email the issue to security@box.com for support.

**Note:** if your issue involves previewing files, please report the issue [here](https://github.com/box/box-content-preview/issues/new/choose).

**Environment:**

- Elements version: 15.0.0-beta.4

**Desktop (please complete the following information):**

- OS: [e.g. Windows 10]

**Smartphone (please complete the following information):**

**Steps to reproduce the problem:**

In Box Platform sign Tab
Go to Template creation page.
Select a file.
Now click on default selected folder and 'Select destination folder' popup will display.
Now Select any folder and then unselect it.
AR: Save button remains in enabled state and no action is getting performed on clicking on it.
![Screen Shot 2022-01-13 at 4 18 29 PM](https://user-images.githubusercontent.com/97700841/149418087-b0b33844-7716-4d19-8d8a-0b1f0b938d23.png)

**What is the expected behavior? (Screenshots can be helpful here)**\

Save button should be disabled when folder is unselected

**What went wrong? (Screenshots, console logs, or HAR files can be helpful here)**
![Screen Shot 2022-01-13 at 4 18 43 PM](https://user-images.githubusercontent.com/97700841/149418093-83b7a36b-7ad0-48fb-bef6-65237af1ee19.png)
**Link to application or sample code:**

**If relevant, link to file (or attach file here)**

**Expected behavior**

**Screenshots**

**Additional context**

PR #2804, opened 2022-01-14
fix(content-explorer): choose button disabled state

fpan225:
issue #2797

### DIFF
--- a/src/features/content-explorer/content-explorer/ContentExplorer.js
+++ b/src/features/content-explorer/content-explorer/ContentExplorer.js
@@ -5,7 +5,7 @@ import omit from 'lodash/omit';
 
 import ContentExplorerHeaderActions from './ContentExplorerHeaderActions';
 import ContentExplorerEmptyState from './ContentExplorerEmptyState';
-import ContentExplorerActionButtons from './ContentExplorerActionButtons';
+import ContentExplorerActionButtons, { getChosenItemsFromSelectedItems } from './ContentExplorerActionButtons';
 import ContentExplorerSelectAll from './ContentExplorerSelectAll';
 
 import ItemList from '../item-list';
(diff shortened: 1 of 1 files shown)


src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -380,6 +380,19 @@ class ContentExplorer extends Component {
...
     };
 
+    getUpdatedChooseButtonProps = chooseButtonProps => {
+        const { selectedItems } = this.state;
+        const { contentExplorerMode } = this.props;
+        const chosenItems = getChosenItemsFromSelectedItems(selectedItems);
+        if (contentExplorerMode === ContentExplorerModes.SELECT_FOLDER && chosenItems.length === 0) {
+            return {
+                ...chooseButtonProps,
+                isDisabled: chooseButtonProps.isDisabled && this.getCurrentFolder().id === '0',
```
swfree, line 390: There could be other use cases where the consumer of this ContentExplorer feature would want to be able to select the root folder. Is there a way to make use of the `areButtonsDisabled` field instead, from within the repo where you are using this component? If the root folder has `isActionDisabled` set to true, then it should disable the choose button when items are deselected:
https://github.com/box/box-ui-elements/blob/master/src/features/content-explorer/content-explorer/ContentExplorer.js#L446-L450
https://github.com/box/box-ui-elements/blob/master/src/features/content-explorer/content-explorer/ContentExplorerActionButtons.js#L108

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -380,6 +380,19 @@ class ContentExplorer extends Component {
         return <ContentExplorerEmptyState isSearch={isViewingSearchResults} />;
     };
 
+    getUpdatedChooseButtonProps = chooseButtonProps => {
```
tjuanitas, line 383: I think this method is better suited in `src/features/content-explorer/content-explorer/ContentExplorerActionButtons.js`.

src/features/content-explorer/content-explorer/ContentExplorer.js
fpan225: However in this case when `this.getCurrentFolder().id === '0'`  File name will be **"All files"**  it means there is no root folder can be selected for user?  Try to set root folder has `isActionDisabled` to true however it still get overwritten by the chooseButtonProps.

Issue #2805, opened 2022-01-14
Add Media Query Hooks for Responsive UI Support


ivanthai:
**Describe the solution you'd like**
Add media query hooks for support for responsive UI. 

**Describe alternatives you've considered**
n/a

**Additional context**
Feature requested by client

alexkrolick:
No need to create an issue for work tracked internally

PR #2806, opened 2022-01-15
feat: add responsive media query hoc and hooks (#2805)

ivanthai:
Add media query hooks for support for responsive UI.


package.json
```diff
@@ -355,6 +355,7 @@
         "react-modal": "^3.8.1",
         "react-popper": "^1.3.4",
         "react-process-string": "^1.2.0",
+        "react-responsive": "^9.0.0-beta.5",
```
alexkrolick, line 358: Is it possible to use the non-beta version of the library?

src/elements/common/media-query/withMedia.js
```diff
@@ -0,0 +1,79 @@
...
+type PointerType = POINTER_TYPE_NONE | POINTER_TYPE_FINE | POINTER_TYPE_COARSE;
+
+type MediaShape = {
+    anyHover: 'none' | 'hover',
+    anyPointer: 'none' | 'fine' | 'coarse',
+    hover: 'none' | 'hover',
+    isExtraLarge: boolean,
+    isLarge: boolean,
+    isMedium: boolean,
+    isSmall: boolean,
```
alexkrolick, line 36: Can a page be more than one of these at a time? It seems like an enum for `size` would be better than a series of booleans.

src/elements/common/media-query/useMedia.js
```diff
@@ -0,0 +1,37 @@
...
+function useQuery(query: string, onQueryChange?: (_: boolean) => void): boolean {
+    return useMediaQuery({ query }, null, onQueryChange);
+}
+
+/**
+ * @param query
+ * @param onQueryChange
+ * @returns {boolean}
+ */
+function useMedia(query: string, onQueryChange?: (_: boolean) => void): boolean {
```
alexkrolick, line 32: I think we should consider the hook the primary API, rather than the HOC. So any utility logic that we would put into the HOC to simplify Box-specific responsive usage should go here, too.

package.json
ivanthai: The beta version seems to be just a js to typescript migration. However there seems to be a memory leak fix included in one of the beta versions. 

https://github.com/yocontra/react-responsive/commit/95b5384b19ef22d375edc5eb63347d5a452e1829

I bumping it down to 8.2.0 and verifying if this is reproducible

src/elements/common/media-query/withMedia.js
ivanthai: Switched to using enum strings for size

src/elements/common/media-query/useMedia.js
ivanthai: useMedia now returns a media props object

src/elements/common/media-query/types.js
```diff
@@ -0,0 +1,27 @@
...
+    minWidth?: number,
+    pointer?: MediaPointerType,
+};
+
+export type MediaShape = {
+    anyHover: 'none' | 'hover',
+    anyPointer: 'none' | 'fine' | 'coarse',
+    hover: 'none' | 'hover',
+    pointer: 'none' | 'fine' | 'coarse',
+    size: boolean,
```
alexkrolick, line 24: should be an enum of strings 

do we also want to provide the exact screen dimensions here, or anything else?

ivanthai: i've added view height and width

src/elements/common/media-query/index.js
```diff
@@ -0,0 +1,2 @@
+export { default } from './useMedia';
```
alexkrolick, line 1: Let's skip default exports for this and use named ons

`import { useMedia, withMedia } from 'box-ui-elements/es/components/media-query'`

src/elements/common/media-query/useMedia.js
```diff
@@ -0,0 +1,101 @@
+// @flow
```
alexkrolick, line 1: It's not actually a component per-se, but this should live in the `components` directory rather than `elements`. Elements are top-level exports like ContentExplorer. The `common` folder within that is scoped to those use cases.

ivanthai:
Please refer to the story book entry

Screen Shot 2022-01-21 at 3 50 48 PM

swfree (approved):

PR #2807, opened 2022-01-20
feat(classification): add watermarking to security controls)

feixuanli:
Screen Shot 2022-01-19 at 10 47 57 PM
Screen Shot 2022-01-19 at 10 47 34 PM
Screen Shot 2022-01-19 at 10 47 34 PM

### DIFF
--- a/src/features/classification/flowTypes.js
+++ b/src/features/classification/flowTypes.js
@@ -44,11 +44,16 @@ type SharedLinkRestrictions = {
     accessLevel: SharedLinkAccessLevel,
 };
 
+type watermarkApplied = {
+    enabled?: boolean,
+};
+
 type Controls = {
     app?: ApplicationRestriction,
     download?: DownloadRestrictions,
     externalCollab?: ExternalCollabRestriction,
     sharedLink?: SharedLinkRestrictions,
+    watermark?: watermarkApplied,
 };
 
 type ControlsFormat = $Values<typeof SECURITY_CONTROLS_FORMAT>;
--- a/src/features/classification/security-controls/utils.js
+++ b/src/features/classification/security-controls/utils.js
@@ -15,47 +15,41 @@ import {
     SHARED_LINK_ACCESS_LEVEL,
 } from '../constants';
 
-const { SHARED_LINK, DOWNLOAD, EXTERNAL_COLLAB, APP } = ACCESS_POLICY_RESTRICTION;
+const { SHARED_LINK, DOWNLOAD, EXTERNAL_COLLAB, APP, WATERMARK } = ACCESS_POLICY_RESTRICTION;
 const { DEFAULT, WITH_APP_LIST, WITH_OVERFLOWN_APP_LIST } = APP_RESTRICTION_MESSAGE_TYPE;
 const { DESKTOP, MOBILE, WEB } = DOWNLOAD_CONTROL;
 const { BLOCK, WHITELIST, BLACKLIST } = LIST_ACCESS_LEVEL;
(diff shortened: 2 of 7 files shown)


src/features/classification/flowTypes.js
```diff
@@ -44,11 +44,16 @@ type SharedLinkRestrictions = {
     accessLevel: SharedLinkAccessLevel,
 };
 
+type watermarkApplied = {
```
jpfranco, line 47: Would it be preferable to name this consistently like `WatermarkRestriction`, or is there a reason why a different convention was used in this case?

src/features/classification/security-controls/utils.js
```diff
@@ -76,6 +81,16 @@ const getSharedLinkMessages = (controls: Controls): Array<MessageItem> => {
     return items;
 };
 
+const getWatermarkingMessages = (controls: Controls): Array<MessageItem> => {
+    const items = [];
+    const isWatermarkEnabled = getProp(controls, `${WATERMARK}.enabled`);
```
jpfranco, line 86: nit: Might be good to add a default `false` value, in case this object is missing

src/features/classification/security-controls/utils.js
```diff
@@ -15,47 +15,52 @@ import {
...
-        return { message: messages.shortDownloadApp };
-    }
-
-    if (sharing) {
-        return { message: messages.shortSharing };
-    }
-
-    if (download) {
-        return { message: messages.shortDownload };
+    switch (true) {
```
jpfranco, line 32: I feel like this was a bit simpler before. I think ideally the `switch` would be used for enums and each `case` would match a unique value, but here all cases resolve to `true`. I think the same logic should be possible with the existing `if`s.

src/features/classification/security-controls/utils.js
```diff
@@ -15,47 +15,52 @@ import {
...
+            items.push({ message: messages.shortApp });
+            break;
+        default:
+            // no-op
+            break;
     }
 
-    if (app) {
-        return { message: messages.shortApp };
+    if (watermark) {
```
jpfranco, line 59: So far we've only displayed a single bullet for the short version of the summary. Just to confirm: does product/design want to make an exception for watermarking?

src/features/classification/flowTypes.js
feixuanli: Was discussing with Yuvnesh on this, watermarking is not considered as a restriction like the other controls, though we have it as part of access policy, which is also why we are printing it with an extra bullet point in the short version.

feixuanli: Copy in admin console: 
Screen Shot 2022-01-20 at 3 08 36 PM

src/features/classification/security-controls/utils.js
feixuanli: I was thinking about this too, using existing `if`s we would now gotta concat each with `else if`s since we don't return immediately, looks a bit bulky. Let me know how you think ?

src/features/classification/security-controls/utils.js
feixuanli: Yes confirm this is desired with Product.

src/features/classification/security-controls/utils.js
feixuanli: updated

src/features/classification/security-controls/utils.js
jpfranco: I think the `else if` would be easier to understand and would result in less lines of code. A `switch` of boolean values where all the cases resolve to `true` is not a common usage and feels like it could be error prone.

jpfranco (approved):

PR #2808, opened 2022-01-21
fix(datepicker): Add max and min attributes to date input

tjuanitas:
This adds the `max` and `min` attributes to the `<input type="date">` element which uses the browser's constraint validation and prevents the form from submitting if the value is not within the range.

**Before - No min attribute set**

Before Attributes

**After - Min attribute set to 2022**

After Attributes


src/components/date-picker/DatePicker.tsx
```diff
@@ -702,13 +705,20 @@ class DatePicker extends React.Component<DatePickerProps, DatePickerState> {
...
-        // "title" prop is shown during constraint validation as a description of the pattern
-        // See https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern#usability
-        const additionalAttrs =
-            isAccessible && !this.canUseDateInputType
-                ? { name, pattern: ISO_DATE_FORMAT_PATTERN.source, title: 'YYYY-MM-DD' }
-                : {};
+        let additionalAttrs;
+        if (isAccessible && this.canUseDateInputType) {
+            additionalAttrs = {
+                max: this.formatDisplayDateType(maxDate) || '9999-12-31',
```
swfree, line 711: Would it make sense to add any unit tests for testing the max & min is set properly?

tjuanitas: Yup, let me look into this!

swfree (approved):

PR #2817, opened 2022-01-28
chore(deps): upgrade @hapi/address 2.1.2 to @sideway/address 4.1.3

jcamden:
Tested on dev vm. Apparently does not need TextEncoder polyfill at this stage.
@hapi/address 2.1.2 is still in node_modules via @hapi/joi, a fact which probably insulates against unforseeable risks.

No longer seeing "invalid email" error for .cpa top-level domain:
Screen Shot 2022-01-28 at 1 12 45 PM
Screen Shot 2022-01-28 at 1 12 50 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -146,8 +146,8 @@
         "@formatjs/intl-pluralrules": "^1.5.2",
         "@formatjs/intl-relativetimeformat": "^4.5.9",
         "@formatjs/intl-unified-numberformat": "^3.2.0",
-        "@hapi/address": "2.1.1",
         "@sambego/storybook-state": "^2.0.1",
+        "@sideway/address": "^4.1.3",
         "@storybook/addon-actions": "^5.3.9",
         "@storybook/addon-docs": "^5.3.9",
         "@storybook/addon-info": "^5.3.9",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -369,7 +368,6 @@
     },
     "comments": {
         "dependencies": {
-            "@hapi/address": "Version 2.1.2+ requires a polyfill for TextEncoder. Pinning to version 2.1.1 prevents IE11 from breaking.",
```
jstoffan, line 372: Did we test IE11 to ensure this dependency still works as expected?

package.json
```diff
@@ -327,7 +327,6 @@
     "peerDependencies": {
         "@box/cldr-data": "^34.2.0",
         "@box/react-virtualized": "9.22.3-rc-box.2",
-        "@hapi/address": ">=2.0.0 <2.1.2",
```
jstoffan, line 330: The new dependency should appear in `peerDependencies`, since it's used at runtime. This is a fix and a breaking change, so let's update the commit message and PR title as described in our [contribution](https://github.com/box/box-ui-elements/blob/master/CONTRIBUTING.md#commit-types) docs.

jcamden: Ah ha, can I ask about that? I noticed that we don't use straight up dependencies in this repo. Why is that? Is the idea that, because people may only use parts of BUIE, they might not need all of its dependencies?

package.json
jcamden: ... definitely should have done that. On it now, just waiting for Parallels access and IE11 Win 81 Parallels image to download over country-bumpkin DSL.

jcamden:
We have decided not to upgrade at this time. Rather than providing a TextEncoder polyfill for IE11, we will utilize the @hapi/address tld option. Possibly, we will look back at upgrading after IE11 EOL in June.

PR #2818, opened 2022-02-01
fix(tooltip): dismiss tooltip on ESC

viodragon2:
### Issue Resolved
To meet [WCAG 1.4.13](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html) requirement, Tooltip needs to be dismissed by pressing "Escape" key when hovered or focused.

### Resolution
- Handle "Escape" key at `document` level to handle hovering over either the wrapped component or the tooltip
- Capture event and stop propagation of the Escape key event to prevent dismissing other floating components (e.g. Modal, Menu, etc.) from closing  (Note that the requirement describes it should dismiss **additional content**)
> A [mechanism](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html#dfn-mechanism) is available to dismiss the **additional content** without moving pointer hover or keyboard focus, ...

#### Escape key on focused element
tooltip-focus

#### Escape key on hovered element
tooltip-hover

#### Escape key on both focused and hovered elements
tooltip-focus hover

#### Escape key only closing tooltip
Preview

tooltip-previewStopPropagation

Modal
tooltip-modal

### Additional Consideration
- We have use cases where the tooltip is forced opened without hover or focus. In these cases, it doesn't fall under WCAG 1.4.13, and it's okay not to close by "Escape" key
-  Some controlled tooltips won't be closed with "Escape" key by this fix. (e.g. ViewSelector in EUA)
    - With the proposed fix, each controlled tooltips needs to handle key event at `document` since we do not expose proper "close" event for tooltip.
    - One idea to resolve this is by adding new event props (e.g. `onOpen`, `onClose`) so that controlled components can handle these cases without knowing the tooltip's open/close mechanism 
    - (extra consideration) We should also rethink of how `showCloseButton` and `onDismiss` props behave. Clicking on close button closes the tooltip even when tooltip is controlled.

### DIFF
--- a/src/components/tooltip/Tooltip.tsx
+++ b/src/components/tooltip/Tooltip.tsx
@@ -140,12 +140,18 @@ class Tooltip extends React.Component<TooltipProps, State> {
             }
         } else {
             if (!prevState.isShown && this.state.isShown) {
-                document.addEventListener('keydown', this.handleKeyDown);
+                // capture event so that tooltip closes before any other floating components that can be closed by
+                // "Escape" key(e.g. Modal, Menu, etc.)
+                document.addEventListener('keydown', this.handleKeyDown, true);
             }
             if (prevState.isShown && !this.state.isShown) {
-                document.removeEventListener('keydown', this.handleKeyDown);
+                document.removeEventListener('keydown', this.handleKeyDown, true);
             }
         }
+    }
+
+    componentWillUnmount() {
+        document.removeEventListener('keydown', this.handleKeyDown, true);
     }
 
     tooltipID = uniqueId('tooltip');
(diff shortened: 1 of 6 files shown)


src/components/tooltip/Tooltip.tsx
```diff
@@ -130,12 +130,21 @@ class Tooltip extends React.Component<TooltipProps, State> {
...
+        if (isControlled) {
             if (!prevProps.isShown && this.props.isShown) {
                 this.setState({ wasClosedByUser: false });
             }
+        } else {
+            if (!prevState.isShown && this.state.isShown) {
+                document.addEventListener('keydown', this.handleKeyDown);
+            }
+            if (prevState.isShown && !this.state.isShown) {
+                document.removeEventListener('keydown', this.handleKeyDown);
```
alexkrolick, line 146: May need to remove the listener in componentWillUnmount as well, so it doesn't create dangling listeners

swfree (approved):

laughnan (approved):

PR #2822, opened 2022-02-09
feat: add responsive dropdown menu

ivanthai:
image
x


ivanthai:
dropdown-menu-mobile

ivanthai:
UnifiedNewDropdownMenu EUA Demo

dropdown-unified-new-eua

ivanthai:
Storybook Documentation

image

src/components/dropdown-menu/DropdownMenu.js
```diff
@@ -5,7 +5,11 @@ import uniqueId from 'lodash/uniqueId';
 import noop from 'lodash/noop';
 
 import { KEYS } from '../../constants';
+import ResponsiveMenuContainer from './ResponsiveMenuContainer';
```
jstoffan, line 8: Containers should wrap the components they contain rather than be included as children. Do we need a new DropdownMenu wrapper component?

src/components/dropdown-menu/DropdownMenu.js
```diff
@@ -5,7 +5,11 @@ import uniqueId from 'lodash/uniqueId';
...
+import { withMediaQuery } from '../media-query';
```
jstoffan, line 9: This change introduces a new dependency on `react-responsive` to all existing imports of this component. This is a major breaking change and generally undesirable. We should consider moving the responsive-specific code into a new wrapping component so applications can opt-in to its use.

src/components/dropdown-menu/DropdownMenu.js
```diff
@@ -242,6 +253,23 @@ class DropdownMenu extends React.Component<Props, State> {
 
         const bodyEl = bodyElement instanceof HTMLElement ? bodyElement : document.body;
 
+        const isMobile = size === VIEW_SIZE_TYPE.medium || size === VIEW_SIZE_TYPE.small;
```
jstoffan, line 256: We should avoid conflating small screen size and mobile devices, as they're not mutually inclusive. Desktop users could simply have a small window, which would receive the same treatment.

src/components/dropdown-menu/DropdownMenu.scss
```diff
@@ -19,4 +19,17 @@
     margin-top: 5px;
     perspective: 500px;
     perspective-origin: center top;
+
+    &.be-is-small,
+    &.be-is-medium {
+        width: 100%;
+        height: 100%;
```
jstoffan, line 26: Use of relative height requires a parent element to have a fixed height or to have relative height all the way to the root of the document. Should we use absolute positioning or some other solution, instead?

src/components/dropdown-menu/ResponsiveDropdownMenu.stories.js
```diff
@@ -0,0 +1,109 @@
...
+import MenuToggle from './MenuToggle';
+import PlainButton from '../plain-button/PlainButton';
+import SubmenuItem from '../menu/SubmenuItem';
+import { ResponsiveDropdownMenu } from './DropdownMenu';
+import notes from './ResponsiveDropdownMenu.stories.md';
+
+function generateClickHandler(message) {
+    return event => {
+        event.preventDefault();
+        /* eslint-disable-next-line no-console */
```
jstoffan, line 21: Can we avoid disabling this eslint rule by removing the console statement?

src/components/dropdown-menu/ResponsiveMenuContainer.scss
```diff
@@ -0,0 +1,60 @@
+@import '../../styles/variables';
+
+.menu-container {
+    position: relative;
+    top: 0;
+    left: 0;
+    z-index: 9999;
```
jstoffan, line 7: Is this a safe assumption? We've seen issues with competing z-index values in the past.

src/components/dropdown-menu/ResponsiveMenuContainer.scss
```diff
@@ -0,0 +1,60 @@
+@import '../../styles/variables';
+
+.menu-container {
+    position: relative;
+    top: 0;
```
jstoffan, line 5: Are these style rules needed?

src/components/dropdown-menu/ResponsiveMenuContainer.scss
```diff
@@ -0,0 +1,60 @@
...
+    left: 0;
+    z-index: 9999;
+    display: flex;
+    flex-direction: column;
+    width: 100%;
+    height: 100%;
+    background-color: $bdl-gray-02;
+
+    .menu-header {
+        position: relative;
```
jstoffan, line 15: Is this needed?

src/components/dropdown-menu/ResponsiveMenuContainer.scss
```diff
@@ -0,0 +1,60 @@
...
+    top: 0;
+    left: 0;
+    z-index: 9999;
+    display: flex;
+    flex-direction: column;
+    width: 100%;
+    height: 100%;
+    background-color: $bdl-gray-02;
+
+    .menu-header {
```
jstoffan, line 14: Can we flatten these selectors rather than nesting them?

src/components/dropdown-menu/ResponsiveMenuContainer.js
```diff
@@ -0,0 +1,57 @@
...
+    const btnProps = {
+        onClick: onButtonClick,
+    };
+
+    const closeBtnProps = {
+        onClick: onCloseClick,
+    };
+
+    return (
+        <div className="menu-container" {...rest}>
```
jstoffan, line 38: Do these class names conflict with any other components? Can we use SUIT naming conventions?

src/components/dropdown-menu/ResponsiveMenuContainer.js
```diff
@@ -0,0 +1,57 @@
...
+    const secondaryTitle = subtitle && <div className="menu-header-subtitle">{subtitle}</div>;
+
+    const titleComponents = [secondaryTitle, primaryTitle];
+
+    const btnProps = {
+        onClick: onButtonClick,
+    };
+
+    const closeBtnProps = {
+        onClick: onCloseClick,
```
jstoffan, line 34: Can we use declarative props in JSX rather than spreading the objects?

src/components/dropdown-menu/ResponsiveMenuContainer.scss
```diff
@@ -0,0 +1,60 @@
...
+        position: relative;
+        display: flex;
+        align-items: center;
+        justify-content: space-between;
+        padding: 13px 16px;
```
jstoffan, line 19: Can we investigate if there are any existing variables we can use for padding, box shadow, font-size, height, width, etc.?

src/components/dropdown-menu/ResponsiveMenuContainer.scss
```diff
@@ -0,0 +1,60 @@
...
+                width: 32px;
+                height: 32px;
+                padding: unset;
+                background-color: $bdl-gray-10;
+                border: unset;
+            }
+        }
+    }
+
+    & > .aria-menu {
```
jstoffan, line 47: Can we omit the `&` here?

src/components/dropdown-menu/__tests__/ResponsiveMenuContainer.test.js
```diff
@@ -0,0 +1,26 @@
...
+        const menu = (
+            <Menu>
+                <MenuItem>View Profile</MenuItem>
+                <MenuItem>Help</MenuItem>
+            </Menu>
+        );
+
+        const wrapper = shallow(<ResponsiveMenuContainer {...props}>{menu}</ResponsiveMenuContainer>);
+
+        expect(wrapper).toMatchSnapshot();
```
jstoffan, line 24: Can we replace the snapshot tests with explicit assertions?

src/components/dropdown-menu/index.js
```diff
@@ -1,3 +1,3 @@
 // @flow
-export { default } from './DropdownMenu';
+export { default, ResponsiveDropdownMenu } from './DropdownMenu';
```
jstoffan, line 2: Can we move this to a separate component barrel file to avoid introducing a new dependency?

src/components/media-query/withMediaQuery.js
```diff
@@ -1,23 +1,40 @@
...
     children: React.Node,
+    className: string,
 };
 
 function withMediaQuery<Props: PropsShape>(WrappedComponent: React.ComponentType<any>): React.ComponentType<any> {
-    return ({ children, ...rest }: Props) => {
+    const MediaQueryComponent = ({ children, className, ...rest }: Props) => {
         const mediaProps: MediaShape = useMediaQuery();
 
+        const isSmall = mediaProps.size === SIZE_SMALL;
```
jstoffan, line 18: We generally want to avoid casting enums to booleans, as it creates a situation where we can have multiple conflicting states versus a single known state.

src/components/menu/Menu.tsx
```diff
@@ -199,7 +199,7 @@ class Menu extends React.Component<MenuProps> {
         const { menuItemEl }: { menuItemEl?: HTMLElement | null } =
             event.target instanceof Node ? this.getMenuItemElFromEventTarget(event.target) : {};
 
-        if (!menuItemEl) {
+        if (!menuItemEl || (event.target as Element).className.includes('submenu-target')) {
```
jstoffan, line 202: Why is this change needed?

src/components/menu/SubmenuItem.scss
```diff
@@ -10,15 +10,16 @@
         top: $bdl-submenu-offset;
         right: auto;
         left: calc(100% - #{$bdl-submenu-offset});
+        z-index: 9999;
```
jstoffan, line 13: It seems like this conflicts with the other z-index changes we've made.

src/components/menu/SubmenuItem.scss
```diff
@@ -10,15 +10,16 @@
...
         &.is-left-aligned {
             right: 100%;
             left: auto;
         }
 
         &.is-hidden {
-            right: auto;
-            left: -9999px;
+            right: auto !important;
+            left: -9999px !important;
```
jstoffan, line 22: We try to avoid the use of `!important` at all times. Can we determine what's causing the specificity issue and resolve it as the source, instead?

jstoffan (changes requested):

src/components/menu/Menu.tsx
ivanthai: SubmenuItem clicks are not distinguished from a MenuItem click and thus closes the menu when clicked on. This is a problem on mobile where users are not able to perform "hover" actions

Dropdown-submenu-click-bug

src/components/dropdown-menu/DropdownMenu.js
ivanthai: I guess isMobile is a bad name then. We wanted to have the responsive behavior for small and medium sized viewports

src/components/dropdown-menu/DropdownMenu.js
ivanthai: This actually wraps the Menu component that is passed in as a child to the DropdownMenu. Is that acceptable?

src/components/menu/SubmenuItem.scss
ivanthai: Set to 171 which is one above menu z-index (170)

PR #2823, opened 2022-02-09
fix(message-center): doesnt load list until scroll

megansmith-box:


src/features/message-center/components/message-center-modal/MessageCenterModal.js
```diff
@@ -118,7 +118,6 @@ function MessageCenterModal({
                             currentCategory={category}
                             categories={categories}
                             onSelect={value => {
-                                cache.clearAll();
```
laughnan, line 121: What's the impact to its removal? Or maybe a better phrasing: what does this cache clearing/management do with respect to the message center modal?

megansmith-box: Cache clearAll will reset cached measurements for all cells https://github.com/bvaughn/react-virtualized/blob/e360d958b99fa965fd3f1caed4506403ad652cd3/docs/CellMeasurer.md

The impact of removing this line is that when switching tabs, the height of the list won't be re-measured. Since the first tab is all and will contain the largest height, the impact of moving to a new tab with fewer messages is that there will be extra space on the bottom to scroll.

laughnan: So what's happening now is that when you switch tabs, it clears the cache, sets the height to `0` then fails to display anything until scroll?

Have we been able to replicate this with other tabs besides the `Products` one?

megansmith-box: Yes, that's correct. It's been happening consistently for me for the event tab.

laughnan: Mostly musing on an alternative, did you explore causing the `<List>` component to re-render by adding `key` to the props (an example). What happens if we force those cells to rebuild if the list filtering changes?

I think removing the cache change is fine -- assuming it doesn't have implications on virtualizing if we request more in `Product` and then go back to `All` which I assume keeps the messages -- I'm just musing.

megansmith-box: I'll try that out!

laughnan (approved):

PR #2828, opened 2022-02-16
fix(tab-view): focus outline

viodragon2:
Before
Screen Shot 2022-02-15 at 4 56 49 PM

After
Screen Shot 2022-02-15 at 4 55 01 PM

### DIFF
--- a/src/components/tab-view/Tabs.scss
+++ b/src/components/tab-view/Tabs.scss
@@ -107,10 +107,6 @@
         }
     }
 
-    .tab:focus {
-        outline: 1px dotted $primary-color;
-    }
-
     .tab:hover,
     .tab:focus {
         .tab-title {


alexkrolick:
It would be awesome if we had some way of making the outline not bump against the content so much 🤔

viodragon2:
Good call. Just as an FYI, WebAIM's first recommendation is to use default browser outline. Borderline is also okay, but we have to careful on the contrast ratio.

Here are screenshots taken from different browsers.
Safari
Screen Shot 2022-02-16 at 4 17 17 PM
Firefox
Screen Shot 2022-02-16 at 4 18 07 PM

src/components/tab-view/Tabs.scss
```diff
@@ -107,10 +107,6 @@
         }
     }
 
-    .tab:focus {
-        outline: 1px dotted $primary-color;
-    }
-
     .tab:hover,
     .tab:focus {
         .tab-title {
```
alexkrolick, line 112: What if we put the outline on the text content 🤔?

alexkrolick (approved):
Approving for now but maybe we can ask design to revisit the focus concept for the tabs later

PR #2830, opened 2022-02-16
feat: add version history upsell with version preview

jlarsen-box:
Treatment F: 
Screen Shot 2022-03-01 at 4 58 26 PM

Treatment G:
Screen Shot 2022-03-01 at 4 58 08 PM

The only difference between the two is the addition of the image in treatment G.


src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -166,16 +170,85 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
...
         const shouldShowVersionErrorWithUpsell = !!onUpgradeClick && error.status === 403;
-        this.setState({
-            error: shouldShowVersionErrorWithUpsell ? messages.versionNotAvailable : messages.versionFetchError,
-            errorTitle: shouldShowVersionErrorWithUpsell ? messages.versionAccessError : messages.versionServerError,
-            isLoading: false,
-            isWatermarked: false,
-            versionCount: 0,
-            versions: [],
-        });
+        const shouldShowVersions = showUpsell || showUpsellWithPicture;
```
ittaishay, line 175: posted a similar comment in GHE, but we're passing in multiple different upsell props and i don't think it's obvious to the caller which take precedent (e.g. what happens if all are true?)

is there a way to consolidate all potential upsell variations into a single prop?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -166,16 +170,85 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
...
+                        can_download: false,
+                        can_preview: false,
+                        can_upload: false,
+                    },
+                    created_at: '2021-11-29T15:01:19-08:00',
+                    modified_at: '2021-11-29T15:01:19-08:00',
+                    modified_by: currentUser,
+                    trashed_at: null,
+                    size: 1875887,
+                    uploader_display_name: 'Jessica Larsen',
```
ittaishay, line 207: should we use an obviously fake name instead of yours for the dummy data?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -166,16 +170,85 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
...
+                isLoading: false,
+                isWatermarked: false,
+                versionCount: 0,
+                versions: [],
+            });
+        } else {
+            const isWatermarked = false;
+            const versionLimit = Infinity;
+            const versionCount = 3;
+            const versions = [
```
ittaishay, line 192: can we pull out the dummy data to live somewhere other than this function so that we keep `handleFetchError` to a reasonable size?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -166,16 +170,85 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
...
+                    type: 'file_version',
+                    id: '28338883546',
+                    permissions: {
+                        can_delete: false,
+                        can_download: false,
+                        can_preview: false,
+                        can_upload: false,
+                    },
+                    created_at: '2021-11-29T15:01:19-08:00',
+                    modified_at: '2021-11-29T15:01:19-08:00',
```
ittaishay, line 203: should we always have this be this date, or should we make it more programmatic based on the time the user is viewing the upsell?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -166,16 +170,85 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
...
+                versions: [],
+            });
+        } else {
+            const isWatermarked = false;
+            const versionLimit = Infinity;
+            const versionCount = 3;
+            const versions = [
+                {
+                    type: 'file_version',
+                    id: '28338883546',
```
ittaishay, line 195: what does this ID represent? is there any danger in hardcoding it?

ittaishay: it looks like the other two id's are 1 and 2 above this, we might want to make that relationship more obvious if it's necessary to maintain it that way

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -166,16 +170,85 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
...
+                    permissions: {
+                        can_delete: false,
+                        can_download: false,
+                        can_preview: false,
+                        can_upload: false,
+                    },
+                    created_at: '2021-11-29T15:01:19-08:00',
+                    currentUser,
+                    size: 1875887,
+                    trashed_at: null,
```
ittaishay, line 238: a lot of these key-values are identical between the 3 file versions, i'd recommend looking into pulling out the common parts and leveraging the ES6 spread operator to copy them in for each one

src/elements/content-sidebar/versions/messages.js
```diff
@@ -187,6 +187,17 @@ const messages = defineMessages({
         defaultMessage: 'Retention policy expires on {time}.',
         description: 'Message describing when the version retention policy will expire.',
     },
+    versionUpsell: {
+        defaultMessage:
+            'Sorry, version history is not available with your current account plan. To access versioning, select from one of our paid plans.',
+        description: 'upgrade now description',
```
ittaishay, line 193: might be helpful for our translators to add a little more information to the message description

src/elements/content-sidebar/versions/VersionsSidebar.js
```diff
@@ -27,22 +29,59 @@ type Props = {
...
+}: Props) => {
     const showLimit = versions.length >= MAX_VERSIONS;
     const showVersions = !!versions.length;
     const showEmpty = !isLoading && !showVersions && !onUpgradeClick;
     const showError = !!error;
 
+    const upsellOverlay = () => {
+        if (showUpsell || showUpsellWithPicture) {
+            return (
+                <div className="bcs-upsell-wrapper">
```
ittaishay, line 58: please follow SUIT naming conventions https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md

src/elements/content-sidebar/versions/VersionsSidebar.js
```diff
@@ -27,22 +29,59 @@ type Props = {
...
+                    <div className="bcs-version-upsell">
+                        {showUpsellWithPicture && <BoxDrive140 className="bcs-version-upsell-icon" />}
+                        <p className="bcs-upgrade-now">Upgrade Now</p>
```
ittaishay, line 61: `Upgrade Now` should be translated

src/elements/content-sidebar/SidebarContent.js
```diff
@@ -36,6 +37,7 @@ const SidebarContent = ({ actions, children, className, title, elementId, sideba
             <div className="bcs-scroll-content-wrapper">
                 <div className="bcs-scroll-content">{children}</div>
             </div>
+            {overlay}
```
jstoffan, line 40: Can we encapsulate the new UI elements so as to not require a top-level overlay? I don't think this is going to be a reused experience.

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -292,6 +365,8 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
                 onRestore={this.handleActionRestore}
                 onUpgradeClick={onUpgradeClick}
                 parentName={parentName}
+                showUpsell={showUpsell}
+                showUpsellWithPicture={showUpsellWithPicture}
```
jstoffan, line 369: Can we eliminate the need for these props? Can we use the existence of `onUpgradeClick`, instead?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -166,16 +170,85 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
...
+                    },
+                    created_at: '2021-11-29T15:01:19-08:00',
+                    currentUser,
+                    size: 1875887,
+                    trashed_at: null,
+                    uploader_display_name: 'Jessica Larsen',
+                    version_number: '3',
+                },
+            ];
+            this.setState({
```
jstoffan, line 243: Can we use a static background image to display the "mock" versions rather than actually rendering them?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -166,16 +170,85 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
...
-            isWatermarked: false,
-            versionCount: 0,
-            versions: [],
-        });
+        const shouldShowVersions = showUpsell || showUpsellWithPicture;
+
+        if (!shouldShowVersions) {
+            this.setState({
+                error: shouldShowVersionErrorWithUpsell ? messages.versionNotAvailable : messages.versionFetchError,
+                errorTitle: shouldShowVersionErrorWithUpsell
```
jstoffan, line 180: Do we now have three separate upsell experiences? Can we reduce this to one?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -166,16 +170,85 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
...
+                    id: '28338883546',
+                    permissions: {
+                        can_delete: false,
+                        can_download: false,
+                        can_preview: false,
+                        can_upload: false,
+                    },
+                    created_at: '2021-11-29T15:01:19-08:00',
+                    modified_at: '2021-11-29T15:01:19-08:00',
+                    modified_by: currentUser,
```
jstoffan, line 204: I think this experience could be very confusing to the point of causing concern, as it implies a level for access to the file that the current user may not have. It seems like a static background image with clearly mock data would be preferable for simplicity and clarity.

src/elements/content-sidebar/versions/VersionsSidebar.scss
```diff
@@ -42,3 +42,47 @@
...
+    position: relative;
+    height: 100%;
+    text-align: center;
+    background: linear-gradient(to bottom, rgba(255, 255, 255, 0) 0%, rgba(255, 255, 255, 1) 30%);
+}
+
+.bcs-version-upsell {
+    position: absolute;
+    top: 30%;
+    padding: 48px;
```
jstoffan, line 56: Can we add a comment on how we arrived at this number?

src/elements/content-sidebar/versions/VersionsSidebar.scss
```diff
@@ -42,3 +42,47 @@
...
+.bcs-upsell-wrapper {
+    position: relative;
+    height: 100%;
+    text-align: center;
+    background: linear-gradient(to bottom, rgba(255, 255, 255, 0) 0%, rgba(255, 255, 255, 1) 30%);
+}
+
+.bcs-version-upsell {
+    position: absolute;
+    top: 30%;
```
jstoffan, line 55: Can we dock the upsell message to the bottom?

src/elements/content-sidebar/versions/VersionsSidebar.scss
```diff
@@ -42,3 +42,47 @@
...
+    padding: 48px;
+}
+
+.bcs-upgrade-now {
+    margin: 0 0 10px;
+    font-weight: bold;
+}
+
+.bcs-upgrade-button {
+    padding: 10px 15px;
```
jstoffan, line 65: Can we reuse any variables for the values in thi sfile?

src/elements/content-sidebar/versions/VersionsSidebar.scss
```diff
@@ -42,3 +42,47 @@
...
+        border-color: $bdl-box-blue-50;
+    }
+}
+
+.bcs-version-upsell-icon {
+    width: 80px;
+    height: 80px;
+}
+
+.bcs-upgrade-now-message {
```
jstoffan, line 86: Can we move these new styles to a separate stylesheet to encapsulate them?

src/elements/content-sidebar/versions/VersionsSidebar.js
```diff
@@ -27,22 +29,59 @@ type Props = {
...
+    showUpsellWithPicture,
+    versions,
+    ...rest
+}: Props) => {
     const showLimit = versions.length >= MAX_VERSIONS;
     const showVersions = !!versions.length;
     const showEmpty = !isLoading && !showVersions && !onUpgradeClick;
     const showError = !!error;
 
+    const upsellOverlay = () => {
```
jstoffan, line 55: Can we move all of this new code to a separate component that is fully encapsulated?

src/elements/content-sidebar/versions/VersionsSidebar.js
```diff
@@ -27,22 +29,59 @@ type Props = {
...
+        if (showUpsell || showUpsellWithPicture) {
```
jstoffan, line 56: Can we move this to a guard clause and `return null` early if either is not true?

src/elements/content-sidebar/versions/VersionsItemButton.js
```diff
@@ -53,9 +54,9 @@ class VersionsItemButton extends React.Component<Props> {
     };
 
     render() {
-        const { children, fileId, isCurrent, isDisabled, isSelected, onClick } = this.props;
+        const { children, fileId, isCurrent, isDisabled, isSelected, onClick, showVersionPreview } = this.props;
         const buttonClassName = classNames('bcs-VersionsItemButton', {
-            'bcs-is-disabled': isDisabled,
+            'bcs-is-disabled': showVersionPreview ? !showVersionPreview : isDisabled,
```
jstoffan, line 59: We're drilling this prop multiple levels to add a visual treatment. Is this something we can accomplish by adding a CSS class much higher up in the DOM?

src/elements/content-sidebar/versions/VersionsItemButton.js
```diff
@@ -17,6 +17,7 @@ type Props = {
     isDisabled: boolean,
     isSelected: boolean,
     onClick: (event: SyntheticMouseEvent<HTMLButtonElement>) => void,
+    showVersionPreview: boolean,
```
jstoffan, line 20: Can we reuse the `isDisabled` prop?

jstoffan (changes requested):
We may need to rethink our approach with regard to rendering version items that don't actually exist. I also agree with @ittaishay that the new code should be more encapsulated and following our other existing conventions.

src/elements/content-sidebar/versions/VersionsItemButton.js
jlarsen-box: great question. I will look into it! Would be great if we didn't need to  have this here. I was unhappy with it lol

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
jlarsen-box: @jstoffan There are three different ones for the experiment. I am going to try to eliminate how many like flags we're sending over, but the point of the experiment is to try out three separate upsell experiences, yes

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
laughnan: I'm curious the benefit here? Screenreaders aren't going to parse an image (not that they're doing it here for the most part).

jlarsen-box: Yeah, wondering how accessible an image would be. Also thinking maybe encapsulating all this into like a dummy, static version of the version history component to render instead of the actual one? I think that would separate
 the fake data better?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
laughnan: Added a question here: https://github.com/box/box-ui-elements/pull/2830#discussion_r809399386

I'm curious what `very confusing to the point of causing concern` could mean here?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
laughnan: ^ that's totally valid. I'm mostly just curious about the trade-off here.

jstoffan: @laughnan, good question. I think the main benefit is simplicity, since a background image can be applied with a couple of lines of CSS. We have no need for non-functional versions of these components outside of this upsell experiment, so this is a lot of code for just that.

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
jstoffan: My main concern with this particular line is that we're indicating the current user, who may or may not be a collaborator on the current file, uploaded a new version of the file. Version history is in some ways a "system of record," so displaying inaccurate data could lead to alarm/concern.

src/elements/content-sidebar/versions/StaticVersionsItem.js
```diff
@@ -0,0 +1,74 @@
...
+
+    const versionDisplayName = <FormattedMessage {...messages.versionUserUnknown} />;
+
+    const buttonClassName = classNames('bcs-VersionsItemButton', {
+        'bcs-is-disabled': false,
+        'bcs-is-selected': false,
+    });
+
+    return (
+        <div className="bcs-VersionsItem">
```
laughnan, line 37: `[nit]` can we make this wrapping element's `className` to be `bcs-StaticVersionsItem` to match the component or are we depending on this versions item class?

src/elements/content-sidebar/versions/StaticVersionSidebar.js
```diff
@@ -0,0 +1,118 @@
...
+                                </li>
+                            </ul>{' '}
+                        </div>
+                    </LoadingIndicatorWrapper>
+                </div>
+            </div>
+            <div className="bcs-upsell-wrapper">
+                <div className="bcs-version-upsell">
+                    {showUpsellWithPicture && <BoxDrive140 className="bcs-version-upsell-icon" />}
+                    <p className="bcs-upgrade-now">Upgrade Now</p>
```
jackiejou, line 101: Should we put this copy into messages as well?

src/elements/content-sidebar/versions/StaticVersionSidebar.js
```diff
@@ -0,0 +1,118 @@
...
+                                <li className="bcs-VersionsMenu-item" key="versionsToday">
+                                    <section className="bcs-VersionsGroup">
+                                        <h4 className="bcs-VersionsGroup-heading">
+                                            <FormattedMessage {...messages.versionsToday} />
+                                        </h4>
+                                        <ul className="bcs-VersionsList">
+                                            <li className="bcs-VersionsList-item" key={1}>
+                                                <StaticVersionsItem
+                                                    versionNumber="1"
+                                                    versionTime="2021-11-29T15:01:19-08:00"
```
jackiejou, line 75: seems like the versionTimes are all the same timestamp, should we save this as a constant above and re-use for all three?

Also thoughts on using the new Date() aka today as the fake time string?

src/elements/content-sidebar/versions/StaticVersionSidebar.js
```diff
@@ -0,0 +1,118 @@
...
+import '../SidebarContent.scss';
+
+type Props = {
+    isLoading: boolean,
+    onUpgradeClick?: () => void,
+    parentName: string,
+    showUpsellWithPicture: boolean,
+};
+
+const StaticVersionsSidebar = ({ onUpgradeClick, isLoading, parentName, showUpsellWithPicture }: Props) => {
```
jackiejou, line 33: nit: return type

src/elements/content-sidebar/versions/StaticVersionsItem.js
```diff
@@ -0,0 +1,74 @@
...
+import './VersionsItem.scss';
+
+type Props = {
+    versionNumber: string,
+    versionTime: string,
+};
+
+const FIVE_MINUTES_MS = 5 * 60 * 1000;
+
+const StaticVersionsItem = ({ versionNumber, versionTime }: Props) => {
```
jackiejou, line 25: nit: return type

src/elements/content-sidebar/versions/messages.js
```diff
@@ -187,6 +187,17 @@ const messages = defineMessages({
...
+        id: 'sidebar.version.upsell',
+    },
+    upgradeButton: {
+        defaultMessage: 'Upgrade',
+        description: 'upgrade button text',
+        id: 'sidebar.version.upgrade',
```
rdacanay, line 199: should the ids for these messages be:
`be.sidebarVersions.versionUpsell` and `be.sidebarVersions.upgradeButton` 
to match the id formats of the other messages?

src/elements/content-sidebar/versions/StaticVersionsItem.js
```diff
@@ -0,0 +1,74 @@
+/**
+ * @flow
+ * @file Static Versions Item component
+ * @author Box
+ */
+
+import * as React from 'react';
+import { FormattedMessage } from 'react-intl';
+import classNames from 'classnames';
```
rdacanay, line 9: [nit] new line after this

src/elements/content-sidebar/versions/StaticVersionsItem.js
jlarsen-box: Currently we are depending on this versions item class since the css is the same across the static vs actual. I didnt love that I was doing this, but at the same time it does eliminate copying of code. So I will probably work around to see what the minimum styling is and move it

src/elements/content-sidebar/versions/StaticVersionSidebar.js
jlarsen-box: yup! Moving it now

src/elements/content-sidebar/versions/StaticVersionSidebar.js
jlarsen-box: I like the idea of using Date! That will keep the date more relevant.

src/elements/content-sidebar/versions/messages.js
jlarsen-box: yup, nice catch, I'll update the ids

src/elements/content-sidebar/versions/StaticVersionSidebar.js
```diff
@@ -0,0 +1,90 @@
...
+                    className="bcs-static-Versions-content"
+                    crawlerPosition="top"
+                    isLoading={isLoading}
+                >
+                    <section>
+                        <h4 className="bcs-static-VersionsGroup-heading">
+                            <FormattedMessage {...messages.versionsToday} />
+                        </h4>
+                        <ul>
+                            <li>
```
tclancy, line 54: Matter of taste at this size (3 items), but I'd loop over `[1, 2, 3]` and assemble it dynamically to avoid having to maintain 3 entries.

src/elements/content-sidebar/versions/StaticVersionsItem.js
```diff
@@ -0,0 +1,71 @@
...
+import sizeUtil from '../../../utils/size';
+import VersionsItemBadge from './VersionsItemBadge';
+import { ReadableTime } from '../../../components/time';
+import PlainButton from '../../../components/plain-button';
+
+type Props = {
+    versionNumber: string,
+};
+
+const StaticVersionsItem = ({ versionNumber }: Props) => {
```
rdacanay, line 20: [nit] add return type

src/elements/content-sidebar/versions/StaticVersionsItem.js
```diff
@@ -0,0 +1,71 @@
...
+    const versionDisplayName = <FormattedMessage {...messages.versionStaticUser} />;
+
+    const yesterday = new Date();
+    yesterday.setDate(yesterday.getDate() - 1);
+    const versionTimestamp = new Date(yesterday).getTime();
```
rdacanay, line 25: could we just do:
```suggestion
    const versionTimestamp = yesterday.getTime();
```

src/elements/content-sidebar/versions/StaticVersionSidebar.js
```diff
@@ -0,0 +1,90 @@
...
+                    {showUpsellWithPicture && <BoxDrive140 className="bcs-version-upsell-icon" />}
+                    <p className="bcs-upgrade-now">
+                        <FormattedMessage {...messages.versionUpgradeButton} />
+                    </p>
+                    <p>
+                        <FormattedMessage {...messages.versionUpsell} />
+                    </p>
+                    <Button
+                        className="bcs-upgrade-button"
+                        data-resin-target="versioning_error_error_message_upgrade_cta"
```
sanjui9000, line 79: `error_error`

src/elements/content-sidebar/versions/StaticVersionsItem.js
```diff
@@ -0,0 +1,71 @@
+/**
+ * @flow
+ * @file Static Versions Item component
+ * @author Box
+ */
+
+import * as React from 'react';
```
sanjui9000, line 7: Nit. Can we fix order of imports here?

src/elements/content-sidebar/versions/StaticVersionSidebar.js
jlarsen-box: omg, tbh I am really proud that I changed to doing this before reading your comment lol

src/elements/content-sidebar/versions/StaticVersionSidebar.js
```diff
@@ -0,0 +1,97 @@
...
+import './StaticVersionsSidebar.scss';
+
+type Props = {
+    isLoading: boolean,
+    onUpgradeClick?: () => void,
+    parentName: string,
+    showUpsellWithPicture: boolean,
+};
+
+const StaticVersionsSidebar = ({ onUpgradeClick, isLoading, parentName, showUpsellWithPicture }: Props): Object => {
```
jackiejou, line 27: The return type should be a react element/node?

src/elements/content-sidebar/versions/StaticVersionSidebar.js
```diff
@@ -0,0 +1,97 @@
...
+    const yesterday = new Date();
+    yesterday.setDate(yesterday.getDate() - 1);
+    const versionTimestamp = new Date(yesterday);
```
jackiejou, line 30: const yesterday was set to the date for yesterday, this line might not be needed?

src/elements/content-sidebar/versions/StaticVersionSidebar.js
jackiejou: Also nit: alphabetize props

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -36,6 +37,14 @@ type Props = {
     onVersionRestore: VersionActionCallback,
     parentName: string,
     versionId?: string,
+    versionUpsellExperience: string,
```
jackiejou, line 40: should this be optional for other places that uses this component?

src/elements/content-sidebar/versions/messages.js
```diff
@@ -153,9 +153,9 @@ const messages = defineMessages({
         description: 'Heading for inline error messages with upsell.',
     },
     versionUpgradeButton: {
```
rdacanay, line 155: rename this to match new id `upgradeLink`

src/elements/content-sidebar/versions/StaticVersionSidebar.js
```diff
@@ -0,0 +1,97 @@
...
+import BoxDrive140 from '../../../illustration/BoxDrive140';
+
+import { LoadingIndicatorWrapper } from '../../../components/loading-indicator';
+
+import VersionsMenu from './VersionsMenu';
+import './StaticVersionsSidebar.scss';
+
+type Props = {
+    isLoading: boolean,
+    onUpgradeClick?: () => void,
```
tclancy, line 23: Why is this optional? There doesn't seem to be handling for it being null below.

jlarsen-box: great question, a relic from when I was doing this in the original component. It should never be optional or null here. I will update that!

src/elements/content-sidebar/versions/StaticVersionSidebar.js
```diff
@@ -0,0 +1,97 @@
+/**
+ * @flow
+ * @file Versions Sidebar component
+ * @author Box
+ */
+
+import * as React from 'react';
```
sanjui9000, line 7: Nit. Can we fix order of imports like we decided in the document?
https://confluence.inside-box.net/display/ETO/PR+Conventions

jlarsen-box: @sanjui9000 I think I got it right this time if you want to take a look!

src/elements/content-sidebar/versions/StaticVersionSidebar.js
```diff
@@ -0,0 +1,96 @@
...
+import { FormattedMessage } from 'react-intl';
+
+import BoxDrive140 from '../../../illustration/BoxDrive140';
+
+import { BackButton } from '../../common/nav-button';
+import Button from '../../../components/button';
+import { LoadingIndicatorWrapper } from '../../../components/loading-indicator';
+import VersionsMenu from './VersionsMenu';
+
+import messages from './messages';
```
rdacanay, line 17: [nit] new line after this line?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -36,6 +37,14 @@ type Props = {
     onVersionRestore: VersionActionCallback,
     parentName: string,
     versionId?: string,
+    versionUpsellExperience?: string,
```
rdacanay, line 40: Do we want the type to be `?string` instead since, looking at the EUA changes, this could also be null

src/elements/content-sidebar/versions/StaticVersionSidebar.js
```diff
@@ -0,0 +1,96 @@
+/**
+ * @flow
+ * @file Versions Sidebar component
```
jstoffan, line 3: Can we update the flow description to differentiate it from the main Versions Sidebar?

src/elements/content-sidebar/versions/StaticVersionSidebar.js
```diff
@@ -0,0 +1,96 @@
...
+            modified_by: null,
+            size: 1875887,
+            trashed_at: null,
+            uploader_display_name: messages.versionStaticUser.defaultMessage,
+        };
+    });
+
+    return (
+        <div
+            className="bcs-static-Versions"
```
jstoffan, line 49: Can we update this and the other classes to use `bcs-StaticVersionSidebar`?

src/elements/content-sidebar/versions/StaticVersionSidebar.js
```diff
@@ -0,0 +1,96 @@
...
+                    crawlerPosition="top"
+                    isLoading={isLoading}
+                >
+                    <div className="bcs-Versions-menu">
+                        <VersionsMenu versions={versions} fileId="1" versionCount={3} versionLimit={3} />
+                    </div>
+                </LoadingIndicatorWrapper>
+            </div>
+
+            <div className="bcs-upsell-wrapper">
```
jstoffan, line 74: Can we use SUIT style naming for the remaining components?

src/elements/content-sidebar/versions/StaticVersionSidebar.js
```diff
@@ -0,0 +1,96 @@
...
+                    <p className="bcs-upgrade-now">
+                        <FormattedMessage {...messages.versionUpgradeLink} />
+                    </p>
+                    <p>
+                        <FormattedMessage {...messages.versionUpsell} />
+                    </p>
+                    <Button
+                        className="bcs-upgrade-button"
+                        data-resin-target="versioning_error_upgrade_cta"
+                        onClick={onUpgradeClick}
```
jstoffan, line 86: Button components are `type="submit"` by default. Can we use `type="button"` here?

src/elements/content-sidebar/versions/StaticVersionsSidebar.scss
```diff
@@ -0,0 +1,82 @@
+@import '../../../styles/variables';
+@import '../../common/variables';
+
+.bcs-static-Versions {
+    position: relative;
+    width: 341px; // 1px extra to account for the margin left
```
jstoffan, line 6: Can we use the `SidebarContent` component since we no longer need the overlay?

src/elements/content-sidebar/versions/StaticVersionsSidebar.scss
```diff
@@ -0,0 +1,82 @@
...
+        position: absolute;
+        top: 30%;
+        padding: 48px;
+
+        .bcs-version-upsell-icon {
+            width: 80px;
+            height: 80px;
+        }
+
+        .bcs-upgrade-now {
```
jstoffan, line 59: Generally, we don't want to nest too much since it increases specificity. Classes with unique SUIT naming can be flattened out.

src/elements/content-sidebar/versions/StaticVersionsSidebar.scss
```diff
@@ -0,0 +1,82 @@
...
+            margin: 0 0 10px;
+            font-weight: bold;
+        }
+
+        .bcs-upgrade-button {
+            margin-top: -8px;
+            padding: 10px 15px;
```
jstoffan, line 66: Do we have any variables we can use for the numeric values in this file?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -20,6 +20,7 @@ import VersionsSidebarAPI from './VersionsSidebarAPI';
 import { withAPIContext } from '../../common/api-context';
 import type { VersionActionCallback, VersionChangeCallback } from './flowTypes';
 import type { BoxItemVersion, BoxItem, FileVersions } from '../../../common/types/core';
+import StaticVersionsSidebar from './StaticVersionSidebar';
```
jstoffan, line 23: Can we alphabetize this so it matches the convention above?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -36,6 +37,14 @@ type Props = {
     onVersionRestore: VersionActionCallback,
     parentName: string,
     versionId?: string,
+    versionUpsellExperience?: string,
+};
+
+// Epoxy Experiment Bucks
```
jstoffan, line 43: Do we need to reference `Epoxy` or `Bucks`?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -280,8 +290,21 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
     };
 
     render() {
-        const { fileId, parentName, onUpgradeClick } = this.props;
-
+        const { fileId, parentName, onUpgradeClick, versionUpsellExperience } = this.props;
+        if (
```
jstoffan, line 294: Can we add whitespace above the `if`?

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -280,8 +290,21 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
...
+            return (
+                <StaticVersionsSidebar
+                    onUpgradeClick={onUpgradeClick}
+                    parentName={parentName}
+                    showUpsellWithPicture={versionUpsellExperience === UPSELL_TYPE.STATIC_VERSION_HISTORY_WITH_PICTURE}
+                    {...this.state}
+                />
+            );
+        }
         return (
```
jstoffan, line 308: Whitespace above the return?

src/elements/content-sidebar/versions/StaticVersionSidebar.js
```diff
@@ -0,0 +1,96 @@
...
+            version_number: versionNumber,
+            type: 'file_version',
+            permissions: {
+                can_preview: true,
+            },
+            created_at: versionTimestamp.toUTCString(),
+            modified_by: null,
+            size: 1875887,
+            trashed_at: null,
+            uploader_display_name: messages.versionStaticUser.defaultMessage,
```
jstoffan, line 43: Do we need to internationalize the name of the uploader? If not, let's hardcode it here. If so, we'll need to bring in `intl.formatMessage` rather than referencing the default message, which is always in English.

src/elements/content-sidebar/versions/StaticVersionSidebar.js
jlarsen-box: Ack, will do

src/elements/content-sidebar/versions/StaticVersionSidebar.js
jlarsen-box: Yup!

src/elements/content-sidebar/versions/StaticVersionSidebar.js
jlarsen-box: ah, good point, we dont need to translate it, so I will switch to hardcoding it.

src/elements/content-sidebar/versions/StaticVersionSidebar.js
jlarsen-box: Yup!

src/elements/content-sidebar/versions/StaticVersionsSidebar.scss
jlarsen-box: ah, got it! Good to know

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
jlarsen-box: nope! Forgot to get rid of this comment!

src/elements/content-sidebar/versions/StaticVersionsSidebar.scss
jlarsen-box: not really, they were values from the css for the non static version.. I could put them in variables, but not really sure how helpful that would be.

jstoffan (approved):

PR #2831, opened 2022-02-16
Fix usm invite modal

mickr:


src/components/selector-dropdown/SelectorDropdown.js
```diff
@@ -296,7 +297,7 @@ class SelectorDropdown extends React.Component<Props, State> {
             </ul>
         );
 
-        // change onKeyPress/onPaste back to onInput when React fixes this IE11 bug: https://github.com/facebook/react/issues/7280
+        // change onPaste back to onInput when React fixes this IE11 bug: https://github.com/facebook/react/issues/7280
```
jstoffan, line 300: Did we test IE11 to validate this still works as before?

mickr: I am in progress with this, but I assume it will work as we are using a supported API. But will let you know...

mickr: I tested that this functionality now works with IME in IE11 under Windows 10. The `onKeyDown` event is fired when using IME.

jstoffan (approved):

PR #2837, opened 2022-02-28
fix(avatar): change unknown avatar svg

jtodd819:
- Updates Unknown User avatar to use svg from [Avatar Badge](https://opensource.box.com/box-ui-elements/storybook/?path=/story/icon-fill-avatarbadge16--avatar-badge-16) icon
- Updated icon from local storybook:
323BB3A5-E9A0-42D7-A5ED-23DB95F3DE12_4_5005_c


src/icons/avatars/UnknownUserAvatar.tsx
```diff
@@ -14,8 +14,6 @@ class UnknownUserAvatar extends React.PureComponent<TwoTonedIcon> {
         width: 28,
     };
 
-    idPrefix = `${uniqueId(ICON_CLASS)}-`;
```
alexkrolick, line 17: Should we be removing this unique ID? Do we know what depends on it?

src/components/avatar/UnknownUserAvatar.tsx
```diff
@@ -9,10 +10,10 @@ const UnknownUserAvatar = ({ className = '', height = 28, title, width = 28 }: S
         viewBox="0 0 28 28"
         width={width}
     >
-        <circle cx="14" cy="14" fill="#ededed" r="14" />
         <path
-            d="M5.5 25.1C6.7 21 10 18 14 18s7.3 3 8.5 7.1a14 14 0 0 1-17 0zM14 16a5 5 0 1 1 0-10 5 5 0 0 1 0 10z"
-            fill="#c3c3c3"
+            fill={vars.bdlGray50}
```
alexkrolick, line 14: Looks like this has WCAG contrast of 3.19, which should be ok for secondary elements.

src/components/avatar/UnknownUserAvatar.tsx
```diff
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import AccessibleSVG, { SVGProps } from '../accessible-svg/AccessibleSVG';
+import * as vars from '../../styles/variables';
 
 const UnknownUserAvatar = ({ className = '', height = 28, title, width = 28 }: SVGProps) => (
```
alexkrolick, line 5: Should this component wrap the icon comp instead of duplicating it?

src/icons/avatars/UnknownUserAvatar.tsx
jtodd819: This is only used in the definition and reference of the circle in the old avatar svg, so it is safe to remove.

src/components/avatar/UnknownUserAvatar.tsx
jtodd819: These are pretty much identical, but they each use different versions of `AccessibleSVG` components with incompatible props. The one here references the `AccessibleSVG` in `src/components/accessible-svg/AccessibleSVG.tsx` which is marginally different from the accessible svg in `src/icons/accessible-svg/AccessibleSVG.tsx` in the icon comp. We should probably standardize the props between these two components as they functionally serve the same purpose, but I wouldn't think that would be in scope for this update.

src/icons/avatars/UnknownUserAvatar.tsx
alexkrolick: Are you sure that the ID of the icon isn't referenced by other code indirectly, though? I don't mean the variable reference. What was the original purpose of having an ID on the icon?

src/components/avatar/UnknownUserAvatar.tsx
alexkrolick: Can you create a ticket to track that clean up work?

jtodd819: https://github.com/box/box-ui-elements/issues/2840 created

alexkrolick (approved):

src/components/avatar/UnknownUserAvatar.tsx
alexkrolick: Thanks - but move that to internal Jira. We don't track tasks in this repo.

jtodd819:
@alexkrolick not sure how to merge this. Based on the other repo changes it looks like I need to add the ready to merge label but I can't edit the labels on this request.

PR #2838, opened 2022-03-01
chore(unit-tests): set up @testing-library/jest-dom

jcamden:


package.json
```diff
@@ -85,7 +85,7 @@
         "test": "NODE_ICU_DATA=node_modules/full-icu BABEL_ENV=test NODE_ENV=test yarn jest",
```
alexkrolick, line 85: Can you move the root jest config next to the other one and add the -c flag here?

jcamden: Hmm, as I mentioned, moving jest.config.js out of the root dir creates some issues related to Jest's perception of what the root dir is: if I set rootDir to "../../" in jest.config.js, then jest can resolve our paths which use rootDir (e.g. '<rootDir>/scripts/jest/jest-setup.js' ), but it does not properly resolve the packages from the root of the project:
```
Cannot find module '@testing-library/jest-dom' from 'jest-setup.js'
    > 1 | import '@testing-library/jest-dom';
```
That's why I recommended that we leave the jest.config.js in the root of the project. However, I'm sure there must be a way to make to work. I'll keep looking into it, but stop me if you think it might not be worth the additional complexity.

jcamden: done :)

scripts/jest/jest.config.js
```diff
@@ -1,21 +1,25 @@
-const path = require('path');
-
 module.exports = {
```
alexkrolick, line 1: For future reference, `git mv` would have preserved the history of the 2 files

alexkrolick (approved):

PR #2841, opened 2022-03-02
feat(classification): update security controls watermarking copies

bxie-box:
**Short Summary**:
Short Summary

**Modal**:
Modal


src/features/classification/flowTypes.js
```diff
@@ -59,7 +59,7 @@ type Controls = {
 type ControlsFormat = $Values<typeof SECURITY_CONTROLS_FORMAT>;
 
 type MessageItem = {
-    message: MessageDescriptor,
+    message: React.Node,
```
jpfranco, line 62: I think this should be `MessageDesciptor | React.Node`

src/features/classification/security-controls/SecurityControlsItem.js
```diff
@@ -15,7 +15,7 @@ const ICON_SIZE = 13;
 
 const SecurityControlsItem = ({ message, tooltipMessage }: Props) => (
     <li className="bdl-SecurityControlsItem">
-        <FormattedMessage {...message} />
+        {React.isValidElement(message) ? message : <FormattedMessage {...message} />}
```
jpfranco, line 18: Can we add a unit test for when `isValidElement` returns true?

jpfranco (commented):
Just some minor comments, but I think they might be worth fixing.

bxie-box:
Adding multiple restrictions for ref
Multiple Restrictions

bxie-box:
Have been trying to resolve the flow error, but going to look into it more tomorrow.

bxie-box:
Deferring to the use of `$FlowFixMe` per offline discussion (had spent a bit on addressing this /  few approaches with type casting etc)

patrickpaul (approved):

PR #2845, opened 2022-03-04
fix(content-explorer): Pass id into custom name renderer

BLamy:
Without the id of the item being passed into the callback function you have no idea which row this is suppose to be rendered for.


tjuanitas (approved):

tjuanitas (commented):
If there's no `linkRenderer` provided, the `id` is added to a plain button. This could possibly conflict with other elements using the same `id`? Should we guard for that?

tjuanitas (dismissed):

tjuanitas (approved):

PR #2854, opened 2022-03-15
feat(media-query): Add CSS media queries to BUIE

greathmaster:


src/styles/_sizes.media.scss
```diff
@@ -52,3 +88,12 @@ $preview-medium-width: 1000px;
         @content;
     }
 }
+
+/* stylelint-disable */
+:export {
```
alexkrolick, line 93: I'm not sure I've seen this SCSS feature before. Can you explain more about it? Is it supported in all the browsers and build systems we use?

greathmaster: Actually, `:export` is not an SCSS feature. It is part of the `css-loader` module. CSS Loader implements the Interoperable CSS standard that defines `:export` and `:import` behaviors: 

https://webpack.js.org/loaders/css-loader/#separating-interoperable-css-only-and-css-module-features

jstoffan: I like this idea, but I'm concerned it could cause issues for platform developers who use the BUIE NPM package. Can we use the `variables.ts` generator script for these new variables, instead?

src/styles/_sizes.media.scss
```diff
@@ -5,6 +5,42 @@ $medium-width: 1220px;
...
+$size-large-min: 768px;
+$size-large-max: 1023px;
+$size-xlarge-min: 1024px;
+
+$SIZE_SMALL: '(max-width: #{$size-small-max})';
+$SIZE_MEDIUM: '(min-width: #{$size-medium-min}) and (max-width: #{$size-medium-max})';
+$SIZE_LARGE: '(min-width: #{$size-large-min}) and (max-width: #{$size-large-max})';
+$SIZE_XLARGE: '(min-width: #{$size-xlarge-min})';
+
+@mixin small-screen-size {
```
jstoffan, line 20: Minor, but consider using a `breakpoint` mixin with the size as as string argument. Both Bootstrap and Foundation follow use a similar convention.

```
@mixin breakpoint($size)
```

src/styles/_sizes.media.scss
```diff
@@ -5,6 +5,42 @@ $medium-width: 1220px;
...
 $preview-medium-width: 1000px;
 
+$size-small-max: 374px;
+$size-medium-min: 375px;
+$size-medium-max: 767px;
+$size-large-min: 768px;
+$size-large-max: 1023px;
+$size-xlarge-min: 1024px;
+
+$SIZE_SMALL: '(max-width: #{$size-small-max})';
```
jstoffan, line 15: We typically use lower-hyphenated-case for all variables. Can we switch and then clarify the naming?

jstoffan, line 15: We typically use lower-hyphenated-case for all variables. Can we switch and then clarify the naming?

src/styles/_sizes.media.scss
```diff
@@ -5,6 +5,42 @@ $medium-width: 1220px;
 $preview-small-width: 849px;
 $preview-medium-width: 1000px;
 
+$size-small-max: 374px;
+$size-medium-min: 375px;
```
jstoffan, line 9: Minor, but we could use SCSS math operations to halve our variable count.

src/components/media-query/__tests__/useMediaQuery.test.js
```diff
@@ -11,6 +11,15 @@ type Props = {
     children?: React.node,
 };
 
+jest.mock('../../../styles/_sizes.media.scss', () => {
+    return {
```
jstoffan, line 15: I don't think these will be needed if we use the `variables.ts` approach.

jstoffan, line 15: I don't think these will be needed if we use the `variables.ts` approach.

jstoffan:
I'm not sure why all my comments were duplicated. Looks like the new Github UI changes may have some bugs. 😄

src/styles/_sizes.media.scss
```diff
@@ -5,6 +5,42 @@ $medium-width: 1220px;
...
+$size-medium-max: 767px;
+$size-large-min: 768px;
+$size-large-max: 1023px;
+$size-xlarge-min: 1024px;
+
+$SIZE_SMALL: '(max-width: #{$size-small-max})';
+$SIZE_MEDIUM: '(min-width: #{$size-medium-min}) and (max-width: #{$size-medium-max})';
```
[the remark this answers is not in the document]
alexkrolick: To clarify, you mean using `min-width` only to create a mobile-first approach, right? I'm on board with that as well.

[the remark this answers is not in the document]
jstoffan: Yep, exactly. Thanks for clarifying, that was a much better way of saying it.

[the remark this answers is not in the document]
alexkrolick: Would it be min-width or max-width?

src/styles/_sizes.media.scss
greathmaster: Understood. Will make use of the `variables.ts` generator script.

src/styles/_sizes.media.scss
greathmaster: Added `breakpoint` mixin to _media-queries.scss file

src/styles/_sizes.media.scss
greathmaster: Switched to use lower-hyphenated-case for variables. Also changed variable names (`$small-screen`, `$medium-screen`, and `$large-screen`). Also moved variables to _media-queries.scss file.

src/styles/_sizes.media.scss
greathmaster: This is no longer needed due to changing the media queries from explicit to overlapping. The number of variables has been consolidated to `$size-small-max`, `$size-medium-max`, and `$size-large-max`.

src/styles/_sizes.media.scss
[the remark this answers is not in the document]
greathmaster: Updated the media queries to be overlapping using `max-width`.

src/components/media-query/__tests__/useMediaQuery.test.js
greathmaster: Went with `variables.ts` approach so removed these mocks.

jstoffan (approved):
LGTM 👍. FYI, @alexkrolick

PR #2862, opened 2022-03-24
feat(icons): Add color attribute to AvatarPlusBadge16

plaskowski:
Hi,
Piotr Laskowski from Box Poland here.
I need this icon in white color to use it in a GuidedTooltip.


tjuanitas (commented):
I don't think we usually update these components since they're auto-generated.

You should be able to update the path color in your stylesheet.

plaskowski:
Good to know, I will go for CSS then!

PR #2864, opened 2022-03-25
fix(editable-shared-link): if FTUX is shown also show new tag

ahorowitz123:


src/features/unified-share-modal/SharedLinkPermissionMenu.js
```diff
@@ -33,11 +33,19 @@ type Props = {
...
+    hasSeenEditTag: boolean,
+};
+
+class SharedLinkPermissionMenu extends Component<Props, State> {
     static defaultProps = {
         trackingProps: {},
     };
 
+    state = {
+        hasSeenEditTag: false,
```
jstoffan, line 46: I was under the impression that the FTUX targeting API stored this state internally and exposed helpers to help mange it. It may be worth reaching out to Product Adoption for some guidance. Maybe @megansmith-box could help?

src/features/unified-share-modal/SharedLinkPermissionMenu.js
```diff
@@ -62,8 +70,11 @@ class SharedLinkPermissionMenu extends Component<Props> {
             trackingProps,
         } = this.props;
         const { sharedLinkPermissionsMenuButtonProps } = trackingProps;
-        const canShowTag = sharedLinkEditTagTargetingApi ? sharedLinkEditTagTargetingApi.canShow : false;
         const canShowTooltip = sharedLinkEditTooltipTargetingApi ? sharedLinkEditTooltipTargetingApi.canShow : false;
+        // we show the "new" tag if the tooltip is shown, otherwise we follow the normal targeting api flow
+        const canShowTag = sharedLinkEditTagTargetingApi
+            ? (isSharedLinkEditTooltipShown && !this.state.hasSeenEditTag) || sharedLinkEditTagTargetingApi.canShow
```
jstoffan, line 76: It's usually prudent to keep ternaries to a single line and break out these compound conditions into their own variables for readability's sake.

src/features/unified-share-modal/__tests__/SharedLinkPermissionMenu.test.js
```diff
@@ -84,6 +84,45 @@ describe('features/unified-share-modal/SharedLinkPermissionMenu', () => {
 
             expect(wrapper.find('LabelPillText')).toHaveLength(length);
         });
+
+        test.each`
+            isSharedLinkEditTooltipShown | length | should
+            ${true}                      | ${1}   | ${'should render LabelPillText if tooltip was shown'}
```
jstoffan, line 90: We generally avoid parameterizing the human-readable test description. Should these be two separate tests?

src/features/unified-share-modal/SharedLinkPermissionMenu.js
ahorowitz123: You are correct about the FTUX targeting API internal state. I talked with @shahzadaziz and we decided that the desired behavior was to show the "New" tag when the tooltip is shown. The main principle of the targeting API is that it only shows one onboarding behavior at a time. However, in this case we want to show two so this was the workaround for that. Definitely interested if there is a better way.

src/features/unified-share-modal/__tests__/SharedLinkPermissionMenu.test.js
ahorowitz123: Ah ok. Yeah I'll split these up. Will definitely be more readable that way

src/features/unified-share-modal/SharedLinkPermissionMenu.js
megansmith-box: I agree that the main issue here is that the way the generic experience machine is set up, only one experience can be shown at a time per page load. Therefore, either the Tooltip or the Tag will show up independently of each other. I do see this implementation as a work around. 

I believe a better way would be to remove `shared-link-with-edit-tooltip` & `shared-link-with-edit-tag` from the generic experience machine and create a custom 2 step state machine for this experience. That way they are grouped together since they are dependent of each other. The flow for the new machine would be
idle -> show tooltip -> show tag -> done

There is some documentation about this here: https://confluence.inside-box.net/display/ETO/4+-+Creating+Tooltip+Flows and several examples of custom state machines in EUA. 

Let me know what you think!

PR #2868, opened 2022-04-04
feat(full-screen-popover): Add full screen popover for mobile overlays

ivanthai:
I decided to build FullScreenPopover for BUIE because DropdownMenu didn’t seem like the appropriate implementation of the UI element in the attached screenshot. More specifically, adding a TabView into DropdownMenu didn't seem flexible enough and felt hacky in order to work. FullScreenPopover is a more generalized component and I can just swap it in if I want to use a mobile/responsive layout

One thing I am particularly unclear about is my decision to modify Flyout to support: 
1) disableTether prop - disables tethering so I can apply css to the overlay component 
2) clickableOverrideSelector prop - an override to identify alternative elements that should be considered “clickable”. by default, Flyout considers only Button and link (a) elements as clickable

Gif demos are below

image


ivanthai:
The FullscreenFlyout allows users to create UI elements with full-screen overlays for mobile/responsive layouts. 
image

ivanthai:
Basic usage.
fullscreen-flyout-basic

ivanthai:
More complex usage with header titles and button along and custom items in the overlay triggering closeOverlay(
fullscreen-flyout-complex
)

src/components/fullscreen-flyout/OverlayHeader.scss
```diff
@@ -0,0 +1,48 @@
...
+        .close-btn {
+            width: 32px;
+            height: 32px;
+            padding: unset;
+            background-color: $bdl-gray-10;
+            border: unset;
+        }
+    }
+
+    //> .aria-menu {
```
greathmaster, line 35: Is this intended to be removed since they are commented out?

src/components/fullscreen-flyout/FullscreenFlyout.stories.js
```diff
@@ -0,0 +1,59 @@
...
+                actionButton={<PrimaryButton>Action</PrimaryButton>}
+            />
+        }
+    >
+        <PrimaryButton>Click (custom close on click elements)</PrimaryButton>
+        <Menu>
+            <MenuItem>Menu Item 1</MenuItem>
+            <MenuItem>Menu Item 2</MenuItem>
+            <MenuSeparator />
+            <MenuItem>Menu Item 3</MenuItem>
```
greathmaster, line 51: Confirming this will scroll if content exceeds the viewport height?

src/components/fullscreen-flyout/FullscreenFlyout.js
```diff
@@ -0,0 +1,53 @@
...
+            disableTether
+            onCloseClick
+            clickableOverrideSelector={clickableSelector}
+            className={classNames('fullscreen-flyout', className)}
+            {...rest}
+        >
+            {button}
+            <Overlay>
+                {header || <OverlayHeader />}
+                {content}
```
greathmaster, line 47: The search functionality on small screens includes a text input  in the header that updates the content. Can the FullscreenFlyout be used in this case?

src/components/fullscreen-flyout/OverlayHeader.js
```diff
@@ -0,0 +1,63 @@
...
+};
+
+type CloseButtonProps = {
+    onClick?: Function,
+};
+
+const CloseButton = ({ onClick }: CloseButtonProps) => {
+    return (
+        <Button onClick={onClick} className="close-btn" type="button">
+            <IconClose color={bdlGray62} height={18} width={18} />
```
greathmaster, line 32: Do we have button size constants we could use for height and width?

src/components/fullscreen-flyout/FullscreenFlyout.js
ivanthai: No, not yet. I left out customizability of the header initially. I was considering adding a way to customize it to have something other than just titles.

src/components/fullscreen-flyout/OverlayHeader.scss
ivanthai: ah. this one fell through the cracks. It was intended to be removed. Thanks for catching that

src/components/fullscreen-flyout/FullscreenFlyout.js
ivanthai: I added customizability to the header and we can move custom implementation to EUA

src/components/fullscreen-flyout/OverlayHeader.js
ivanthai: Not sure if it's appropriate here because this height/width is specific to this button

alexkrolick (commented):
Nit/Convention: Can you use `bdl-` prefix for the SCSS code to prevent CSS conflicts with the class names?

src/components/full-screen-popover/FullScreenPopover.scss
```diff
@@ -0,0 +1,29 @@
...
+
+        .aria-menu {
+            flex-grow: 1;
+            background: unset;
+            border: unset;
+            border-radius: unset;
+            box-shadow: unset;
+
+            .menu-item {
+                color: $bdl-gray-80;
```
alexkrolick, line 25: a11y: If this is for text, the lightest it should be on white is gray-62

src/components/full-screen-popover/FullScreenPopover.js
```diff
@@ -0,0 +1,135 @@
...
+    /** Components to render in the overlay */
+    children: React.Node,
+    /** Set className to the overlay wrapper */
+    className?: string,
+    /** Content nested inside of the popover header */
+    header?: React.Element<any>,
+    /** Fires this callback when overlay is closed */
+    onClose?: Function,
+    /** CSS selector that identifies clickable elements that closes the overlay */
+    onCloseClassSelector?: string,
```
alexkrolick, line 21: Technically not a "class selector" since it can be any CSS selector

src/components/full-screen-popover/FullScreenPopover.js
```diff
@@ -0,0 +1,135 @@
...
+    const focusButton = () => {
+        const buttonEl = document.getElementById(popoverButtonID);
+        if (buttonEl) {
+            buttonEl.focus();
+        }
+    };
+
+    const handleOverlayClick = (event: SyntheticEvent<>) => {
+        // Searches overlay for classes that match event.target and closes overlay if match
+        const overlayNode: HTMLElement | null = document.getElementById(overlayID);
```
alexkrolick, line 67: In theory you should be able to use a direct ref to the element instead of this query

src/components/full-screen-popover/FullScreenPopover.js
```diff
@@ -0,0 +1,135 @@
...
+    const [isVisible, setVisibility] = React.useState(false);
+
+    const overlayID = uniqueId('overlay');
+    const popoverButtonID = uniqueId('popoverbutton');
+
+    const { className, header, children, onClose, onCloseClassSelector, onOpen } = props;
+
+    const elements = React.Children.toArray(children);
+
+    if (elements.length !== 2) {
```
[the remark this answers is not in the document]
alexkrolick: I also think this API is a little bit clunky, even though DropdownMenu does it. We could just have a required prop for the button.

src/components/fullscreen-flyout/FullscreenFlyout.stories.js
ivanthai: Yes now it does

ivanthai:
Some updates...this component has now been updated to be a simple standalone component that does not rely on Flyout and supports some customizability of the header

src/components/full-screen-popover/FullScreenPopover.js
```diff
@@ -0,0 +1,143 @@
...
+    onCloseCssSelector?: string,
+    /** Fires this callback when overlay is opened */
+    onOpen?: Function,
+    /** Custom button that toggles the overlay */
+    popoverButton: React.Element<any>,
+};
+
+type Props = FullScreenPopoverProps;
+
+const FullScreenPopover = (props: Props) => {
```
jstoffan, line 31: The concept here seems similar to the existing Flyout and/or Modal components. Can we reuse that by adding an `isFullscreen` option to it?

src/components/full-screen-popover/FullScreenPopover.js
```diff
@@ -0,0 +1,143 @@
...
+        <div className={classNames('bdl-full-screen-popover', className)}>
+            {React.cloneElement(popoverButton, popoverButtonProps)}
+            {isVisible ? (
+                <Overlay {...overlayProps}>
+                    <OverlayHeader className="bdl-fs-header" onCloseClick={closeOverlay}>
+                        {header}
+                    </OverlayHeader>
+                    {children != null && (
+                        <ScrollWrapper
+                            className="bdl-fs-content"
```
jstoffan, line 129: Can we update the class naming to be internally consistent and also comply with SUIT naming [rules](https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md)?

src/components/full-screen-popover/FullScreenPopover.js
```diff
@@ -0,0 +1,143 @@
...
+
+    return (
+        <div className={classNames('bdl-full-screen-popover', className)}>
+            {React.cloneElement(popoverButton, popoverButtonProps)}
+            {isVisible ? (
+                <Overlay {...overlayProps}>
+                    <OverlayHeader className="bdl-fs-header" onCloseClick={closeOverlay}>
+                        {header}
+                    </OverlayHeader>
+                    {children != null && (
```
jstoffan, line 127: We should always use strict equality. Should we make `children` a required prop?

jstoffan (changes requested):
@ivanthai, let's sync up to refine the approach before iterating further.

PR #2869, opened 2022-04-05
feat(icon-search-juicy): create IconSearchJuicy

jcamden:
Juicier version of IconSearch [found in Zeplin mockups](https://app.zeplin.io/project/614cb647d7f89ab694666261/screen/6228da594da800bd75b6f3f1) for header on mobile screen sizes. 

Unit tests are happy:
Screen Shot 2022-04-05 at 12 12 17 PM

In action:
Screen Shot 2022-04-05 at 12 17 57 PM

### DIFF
--- a/src/icons/general/IconSearch.js.flow
+++ b/src/icons/general/IconSearch.js.flow
@@ -6,7 +6,7 @@ import { bdlGray40 } from '../../styles/variables';
 
 import type { Icon } from '../flowTypes';
 
-const IconSearch = ({ className = 'icon-search', color = bdlGray40, height = 14, title, width = 14 }: Icon) => (
+const IconSearch = ({ className, color = bdlGray40, height = 14, title, width = 14 }: Icon) => (
     <AccessibleSVG
         className={`icon-search ${className}`}
         height={height}
--- a/src/icons/general/IconSearch.js.flow
+++ b/src/icons/general/IconSearch.js.flow
@@ -6,7 +6,7 @@ import { bdlGray40 } from '../../styles/variables';
 
 import type { Icon } from '../flowTypes';
 
-const IconSearch = ({ className, color = bdlGray40, height = 14, title, width = 14 }: Icon) => (
+const IconSearch = ({ className = '', color = bdlGray40, height = 14, title, width = 14 }: Icon) => (
     <AccessibleSVG
         className={`icon-search ${className}`}
         height={height}
--- a/src/icons/general/IconSearchJuicy.js.flow
+++ b/src/icons/general/IconSearchJuicy.js.flow
@@ -1,0 +1,26 @@
+// @flow
+import * as React from 'react';
+
+import AccessibleSVG from '../accessible-svg';
+import { bdlGray40 } from '../../styles/variables';
+
+import type { Icon } from '../flowTypes';
+
+const IconSearchJuicy = ({ className = 'icon-search', color = bdlGray40, height = 14, title, width = 14 }: Icon) => (
+    <AccessibleSVG
+        className={`icon-search ${className}`}
+        height={height}
+        title={title}
+        viewBox="0 0 24 24"
+        width={width}
+    >
+        <path
+            className="fill-color"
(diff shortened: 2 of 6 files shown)


src/icons/general/IconSearchJuicy.js.flow
```diff
@@ -0,0 +1,26 @@
+// @flow
+import * as React from 'react';
+
+import AccessibleSVG from '../accessible-svg';
+import { bdlGray40 } from '../../styles/variables';
+
+import type { Icon } from '../flowTypes';
+
+const IconSearchJuicy = ({ className = 'icon-search', color = bdlGray40, height = 14, title, width = 14 }: Icon) => (
```
tjuanitas, line 9: I think the default `className` value should be an empty string since a class is already specified on line 11. That seems to match how the other icon files do it.

src/icons/general/IconSearchJuicy.js.flow
```diff
@@ -0,0 +1,26 @@
...
+    <AccessibleSVG
+        className={`icon-search ${className}`}
```
tjuanitas, line 11: I think `icon-search` should be `icon-search-juicy`?

src/icons/general/IconSearchJuicy.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react';
+
+import AccessibleSVG from '../accessible-svg';
+import { bdlGray40 } from '../../styles/variables';
+
+import { Icon } from '../iconTypes';
+
+const IconSearchJuicy = ({ className = 'icon-search', color = bdlGray40, height = 14, title, width = 14 }: Icon) => (
```
tjuanitas, line 8: Same comment as the flow file.

src/icons/general/IconSearchJuicy.tsx
```diff
@@ -0,0 +1,25 @@
...
+    <AccessibleSVG
+        className={`icon-search ${className}`}
```
tjuanitas, line 10: Same comment as the flow file.

src/icons/general/__tests__/IconSearchJuicy.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { shallow } from 'enzyme';
+import IconSearchJuicy from '../IconSearchJuicy';
+import { bdlGray40 } from '../../../styles/variables';
+
+describe('icons/general/IconSearchJuicy', () => {
+    test('should correctly render default icon with default color', () => {
+        const dimension = 14;
+        const wrapper = shallow(<IconSearchJuicy />);
+
+        expect(wrapper.hasClass('icon-search')).toEqual(true);
```
tjuanitas, line 11: Same comment about `icon-search-juicy`

src/icons/general/IconSearchJuicy.js.flow
jcamden: Hey @tjuanitas ! Thanks for the review. I just copied and pasted from `IconSearch`. Should we change, both, one, or neither?

jcamden: Probably both.

tjuanitas: Yeah probably both since it's seems more like an oversight rather than intentional.

src/icons/general/IconSearchJuicy.js.flow
tjuanitas: I'm honestly not sure if people use these classes but likely best to stay with the current pattern. Although I think these are older icons and they were intended to eventually be migrated to our internal Box/design-assets repo which auto-generates icons for BUIE.

src/icons/general/IconSearch.js.flow
```diff
@@ -6,7 +6,7 @@ import { bdlGray40 } from '../../styles/variables';
 
 import type { Icon } from '../flowTypes';
 
-const IconSearch = ({ className = 'icon-search', color = bdlGray40, height = 14, title, width = 14 }: Icon) => (
+const IconSearch = ({ className, color = bdlGray40, height = 14, title, width = 14 }: Icon) => (
     <AccessibleSVG
         className={`icon-search ${className}`}
```
tjuanitas, line 11: Does removing the default value all together potentially add `undefined` to the className prop on line 11? I don't remember exactly what happens but I think using `className = ''` on line 9 might be safer?

tjuanitas: Yeah it looks like the `undefined` is causing the lint check to fail.

tjuanitas (approved):
Looks good! Thanks

PR #2876, opened 2022-04-16
feat(modal): add responsive behavior to Modal component

greathmaster:
# Added Responsive Behavior to Modal

## Resize
Responsive-Modal-Resize

## Zoom
Responsive-Modal-Zoom

## Behavior
Modal width shrinks when viewport is resized until a threshold of `460px` is reached, and then displays in fullscreen.

### Files
- Modal.scss - Added additional CSS to support responsiveness
- _media-queries.scss - Added media query parameter for `460px` modal fullscreen threshold
- visual-adapter.js - Updated `takeModalScreenshot` to support different viewport sizes
- Modal.stories.test.js - Added snapshot test for small screen
- variables.js, variables.json, variables.ts - autogenerated

### Snapshots

- modal-stories-test-js-components-modal-modal-looks-visually-correct-for-small-screens-when-button-is-clicked-in-story-components-modal-basic-1-snap.png
- modal-stories-test-js-components-modal-modal-looks-visually-correct-for-small-screens-when-button-is-clicked-in-story-components-modal-with-custom-backdrop-click-handler-1-snap.png


src/components/modal/Modal.scss
```diff
@@ -133,3 +135,33 @@
...
+            padding: 0;
+        }
+    }
+
+    .modal-dialog-container {
+        width: 100%;
+        margin: auto;
+
+        @include breakpoint($modal-fullscreen) {
+            height: 100%;
```
jstoffan, line 152: Percentage heights are fragile, since they're relative to their direct parent. In this case, `.modal` is a flex container, so this should be equivalent to `flex-basis: 100%`.  I'd suggest testing these changes in IE11 and Safari to ensure they work properly, since those browsers can be a little touchy with flexbox.

src/components/modal/Modal.scss
```diff
@@ -133,3 +135,33 @@
...
+
+            .modal-dialog {
+                min-height: 100%;
+                padding: 22px 16px 12px 16px;
```
jstoffan, line 156: Why did we choose these specific values? It seems like they'd create inconsistent spacing. Can we keep the 30px padding from the default experience, instead?

src/components/modal/Modal.scss
```diff
@@ -133,3 +135,33 @@
...
+.is-responsive-web {
+    .modal {
+        @include breakpoint($modal-fullscreen) {
+            padding: 0;
+        }
+    }
+
+    .modal-dialog-container {
+        width: 100%;
+        margin: auto;
```
jstoffan, line 149: Is this needed if it's already declared in the base styles?

src/components/modal/Modal.scss
```diff
@@ -133,3 +135,33 @@
...
+
+        @include breakpoint($modal-fullscreen) {
+            height: 100%;
+
+            .modal-dialog {
```
jstoffan, line 154: Does this need to be nested inside of the already-nested `.modal-dialog-container` rule? We should aim for as little specificity as is needed to achieve the desired effect.

src/components/modal/Modal.scss
```diff
@@ -133,3 +135,33 @@
...
+                min-height: 100%;
+                padding: 22px 16px 12px 16px;
+                border-radius: 0;
+            }
+        }
+    }
+
+    .modal-close-button {
+        @include breakpoint($modal-fullscreen) {
+            right: 10px;
```
jstoffan, line 164: Can we retain the existing 20px value, instead?

src/styles/constants/_media-queries.scss
```diff
@@ -1,10 +1,12 @@
 $size-small-max: 374px;
 $size-medium-max: 767px;
 $size-large-max: 1023px;
+$modal-fullscreen-width: 460px;
```
jstoffan, line 4: Should the naming here refer to a screen size value like the other variables above? We're measuring the screen rather than the modal, right?

src/components/modal/Modal.scss
greathmaster: Updated this to use `flex-basis: 100%` instead. Seems to be okay on other browsers.

src/components/modal/Modal.scss
greathmaster: These were pulled from the Zeplin designs.

src/components/modal/Modal.scss
greathmaster: No we don't. Removed.

src/components/modal/Modal.scss
greathmaster: Refactored to reduce specificity.

src/components/modal/Modal.scss
greathmaster: This was also pulled from the Zeplin designs.

src/styles/constants/_media-queries.scss
greathmaster: Renamed to clarify its screen width and not modal width.

ivanthai (dismissed):
Looks great!

src/styles/constants/_media-queries.scss
```diff
@@ -1,10 +1,12 @@
 $size-small-max: 374px;
+$size-small-medium-max: 460px;
 $size-medium-max: 767px;
 $size-large-max: 1023px;
 
 $small-screen: unquote('(max-width: #{$size-small-max})');
 $medium-screen: unquote('(max-width: #{$size-medium-max})');
 $large-screen: unquote('(max-width: #{$size-large-max})');
+$small-medium-screen: unquote('(max-width: #{$size-small-medium-max})');
```
tjuanitas, line 9: Looks like these should be sorted by the `px` amount?

```suggestion
$small-screen: unquote('(max-width: #{$size-small-max})');
$small-medium-screen: unquote('(max-width: #{$size-small-medium-max})');
$medium-screen: unquote('(max-width: #{$size-medium-max})');
$large-screen: unquote('(max-width: #{$size-large-max})');
```

tjuanitas: Same suggestion applies for the new additions in `src/styles/variables.js`, `src/styles/variables.json`, and `src/styles/variables.ts`.

src/styles/constants/_media-queries.scss
```diff
@@ -1,10 +1,12 @@
 $size-small-max: 374px;
+$size-small-medium-max: 460px;
```
tjuanitas, line 2: Is this how we usually handle variable names for in-between values?

src/styles/constants/_media-queries.scss
greathmaster: Good idea. Done.

src/styles/constants/_media-queries.scss
greathmaster: I searched the code for in-between values for media queries and actually didn't find any. I'm open to suggestions, but "small-medium" follows the same logic as a "medium-large" size (ie between medium and large).

tjuanitas: Yeah that makes sense. I would defer to @jstoffan if we want to follow any specific naming convention but otherwise looks good to me!

ivanthai (approved):

jaknas (approved):
LGTM

src/components/modal/Modal.scss
```diff
@@ -133,3 +135,28 @@
...
+.is-responsive-web {
+    @include breakpoint($small-medium-screen) {
+        .modal {
+            flex-direction: column;
+            padding: 0;
+        }
+
+        .modal-dialog {
+            min-height: 100%;
+            padding: 22px 16px 12px 16px;
```
paroxyzm, line 149: Would it make sense to extract those values as variables?

ivanthai: Personally, I feel like we'd only need to do this if it's being used somewhere else and/or stated specifically in a styling guide (correct me if there is one). Otherwise it'd seem like a lot of extra work for little benefit

PR #2878, opened 2022-04-23
feat(flyout): Add responsive behaviors

ivanthai:
### **Updated 2022-04-25 8PM PST**
Added/updated tests and documentation

Important files are in `src/components/close-button` and `src/components/flyout`

**2022-04-25 4PM PST**
No longer using ResponsiveFlyout, but instead build responsiveness directly into Flyout. We now utilize Context Providers to pass in Flyout's closeOverlay() function to OverlayHeader. This allows consumers to create a composition of the required elements instead.

```
<Flyout isResponsive...>
    <Overlay...>
        <OverlayHeader>{header content}</OverlayHeader>
...
```

CloseButton is also now a standalone component. This button is going to be reused by other responsive components coming down the pipeline.

You might figured that Flyout and Overlay could have been migrated to TS...especially now that there are tsx files in the same module. That, my friend, will be targeted later.

**2022-04-25**
Created a **ResponsiveFlyout** component that implements Flyout. Responsive behaviors can be toggled using props. 

I didn't want to modify **Flyout** and **Overlay** directly as it felt like I was overloading them. Creating **ResponsiveFlyout** was inspired by **HeaderFlyout** and is a composition of **Flyout**, **Overlay**, and other responsive behaviors. It felt cleaner and I was able to keep the responsive concerns separate from their non-responsive counterparts. The one workaround I opted for is to "disable" _react-tether_ styling by using purely CSS overrides (`transform: none ! important`)

Refactored **HeaderFlyout** to implement **ResponsiveFlyout**.

responsive-flyout-resize-demo


ivanthai:
For a11y, tabbing through elements work

responsive-flyout-tabbing-demo

src/components/responsive-flyout/ResponsiveFlyout.js
```diff
@@ -0,0 +1,70 @@
...
+    flyoutRefFn?: any => any,
+    /** Header for responsive layout */
+    header?: React.Element<any>,
+    /** Overlay classname */
+    overlayClassName?: string,
+    /** Optional function to get the scrollRef in parent components */
+    scrollRefFn?: any => any,
+};
+
+class ResponsiveFlyout extends React.Component<Props> {
```
jcamden, line 30: Oh no, a class, haha! If it is because you wanted to use a ref, there is, of course, `useRef`. (You probably know that though.) About refs in general, do we want to reserve refs as last resort (c.f. https://blog.logrocket.com/why-you-should-use-refs-sparingly-in-production/). If so, would it be possible to do this without a ref---and, either way, with a function instead of a <shudder> class? Young guns like me (xp-wise) ain't got no class.

src/components/responsive-flyout/OverlayHeader.js
```diff
@@ -0,0 +1,33 @@
...
+    /** Set click handler when close button is clicked */
+    onCloseClick?: Function,
+};
+
+const OverlayHeader = (props: Props) => {
+    const { children, onCloseClick, className } = props;
+
+    const handleClick = event => {
+        event.preventDefault();
+        event.stopPropagation();
```
jcamden, line 22: This is cool. `stopPropagation` is new to me. Can help me understand, is this here to catch and essentially ignore click events? What does it do for us in the context of this fullscreen modal, as it were?

src/components/responsive-flyout/styles/OverlayHeader.scss
```diff
@@ -0,0 +1,18 @@
...
+    z-index: 1;
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    height: $overlay-header-height;
+    padding: 13px 16px;
+    background-color: white;
+    box-shadow: 0 2px 4px 0 rgba(0, 0, 0, .08);
+
+    .oh-content {
```
jcamden, line 15: I didn't know what this was when I read it in `OverlayHeader`. Some other Boxers could be as dull as I am on their worst days 😄  On our behalf, would it make sense to go with a more explicit name, such as `overlay-header-content`?

src/components/responsive-flyout/ResponsiveFlyout.js
```diff
@@ -0,0 +1,70 @@
...
+            overlayClassName,
+            flyoutRefFn,
+            ...rest
+        } = this.props;
+
+        return (
+            <Flyout ref={this.flyoutRef} className={classNames({ 'responsive-flyout': enabled }, className)} {...rest}>
+                {flyoutButton}
+                <Overlay className={overlayClassName}>
+                    <OverlayHeader onCloseClick={() => this.flyoutRef.current.closeOverlay()}>{header}</OverlayHeader>
```
ivanthai, line 58: I opted to use refs because I felt that it would be cleaner to implement without needing to make changes to Flyout or Overlay

I looked at passing it down via Context Providers, but I didn't like that solution either and seemed like it was an anti-pattern according to React. Using refs in this way also seemed like an anti-pattern as well, but was the lesser of two evils.

The "more correct" solution might be to create a ResponsiveOverlay and implement OverlayHeader in there and replace Overlay here with it. ResponsiveOverlay would take in a header prop. This way we can handle it declaratively and avoid using refs

src/components/responsive-flyout/OverlayHeader.js
```diff
@@ -0,0 +1,33 @@
...
+    };
+
+    return (
+        <div role="presentation" onClick={handleClick} className={classNames('overlay-header', className)}>
+            <div className="oh-content">{children}</div>
+            <CloseButton className="oh-close-btn" onClick={onCloseClick} />
```
jcamden, line 28: Is this class implemented?

src/components/responsive-flyout/ResponsiveFlyout.js
```diff
@@ -0,0 +1,70 @@
+// @flow
+import * as React from 'react';
+import classNames from 'classnames';
+import { Flyout, Overlay } from '../flyout';
+import ScrollWrapper from '../scroll-wrapper';
+import type { FlyoutProps } from '../flyout/Flyout';
+import OverlayHeader from './OverlayHeader';
```
jcamden, line 7: Let's make sure to sort the imports. (I get this comment on all PRs. I hate manually sorting our imports. We should really make a push to move EUA to eslint import sorting. That way, we can rely on eslint to do our busy work for us.)

src/components/responsive-flyout/styles/ResponsiveFlyout.scss
```diff
@@ -0,0 +1,37 @@
...
+    .overlay {
+        .overlay-header {
+            display: none;
+        }
+    }
+}
+
+.is-responsive-web {
+    @include breakpoint($medium-screen) {
+        .flyout-overlay.responsive-flyout {
```
jcamden, line 14: Does .responsive-flyout duplicate the functionality of .is-responsive-web?

src/features/header-flyout/HeaderFlyout.js
```diff
@@ -27,43 +29,44 @@ class HeaderFlyout extends React.Component<Props> {
...
         return (
-            <Flyout
+            <ResponsiveFlyout
                 closeOnClick={false}
                 offset={HeaderFlyout.panelOffset}
                 className={classNames('header-flyout', className)}
                 constrainToWindow
+                header={header}
+                overlayClassName="header-flyout-overlay"
+                enabled={isResponsive}
```
jcamden, line 46: If this is only used to set a class that depends on the FF, would we be able to remove this whole prop in favor using .is-responsive-web?

jcamden (commented):
Good stuff here. I had fun reading this, and learned some too.

src/components/responsive-flyout/ResponsiveFlyout.js
jcamden: Oh, I just saw this. Great thinking, for sure!

src/components/responsive-flyout/CloseButton.js
```diff
@@ -0,0 +1,25 @@
...
+import classNames from 'classnames';
+
+import Button from '../button';
+import IconClose from '../../icons/general/IconClose';
+import { bdlGray62 } from '../../styles/variables';
+import './styles/CloseButton.scss';
+
+export type CloseButtonProps = {
+    className?: string,
+    onClick?: Function,
```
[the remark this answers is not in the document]
ivanthai: Yes, it is not required to use the button. Didn't see a strong reason to require it. A button can exist but doesn't have to be clicked on

src/components/responsive-flyout/CloseButton.js
```diff
@@ -0,0 +1,25 @@
...
+};
+
+type Props = CloseButtonProps;
```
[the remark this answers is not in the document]
ivanthai: Honestly, copy and paste :)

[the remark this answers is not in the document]
ivanthai: Fixed it, for the sake of simplicity

src/components/responsive-flyout/OverlayHeader.js
```diff
@@ -0,0 +1,33 @@
+// @flow
+import * as React from 'react';
+
+import classNames from 'classnames';
+import CloseButton from './CloseButton';
+import './styles/OverlayHeader.scss';
+
+export type Props = {
```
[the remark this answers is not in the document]
ivanthai: I think that can be handled using aliases imo

src/components/responsive-flyout/OverlayHeader.js
```diff
@@ -0,0 +1,33 @@
...
+    /** Components to render in the header */
+    children?: React.Node,
+    /** Set className to the overlay header */
+    className?: string,
+    /** Set click handler when close button is clicked */
+    onCloseClick?: Function,
```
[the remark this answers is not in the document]
ivanthai: Yes

src/components/responsive-flyout/ResponsiveFlyout.js
ivanthai: That's a good concern to have. I agree its a last resort. I was kind of leaning toward using ResponsiveOverlay at the time of this PR. Might be a good idea to do so

src/components/responsive-flyout/OverlayHeader.js
ivanthai: I want to make sure that it does not conflict with Flyout/Overlay onclick handlers. It contains functionality that could break the OverlayHeader

src/components/responsive-flyout/ResponsiveFlyout.js
```diff
@@ -0,0 +1,70 @@
...
+    /** Custom button to trigger for opening/closing the flyout */
+    flyoutButton: React.Element<any>,
+    /** Optional function to get the flyoutRef in parent components */
+    flyoutRefFn?: any => any,
+    /** Header for responsive layout */
+    header?: React.Element<any>,
+    /** Overlay classname */
+    overlayClassName?: string,
+    /** Optional function to get the scrollRef in parent components */
+    scrollRefFn?: any => any,
```
tjuanitas, line 27: Could we use the prop type from [`ScrollWrapper`](https://github.com/box/box-ui-elements/blob/master/src/components/scroll-wrapper/ScrollWrapper.js#L15)?

`scrollRefFn: (?HTMLElement) => void,`

src/components/responsive-flyout/ResponsiveFlyout.js
```diff
@@ -0,0 +1,70 @@
...
+    /** Components to render in the overlay */
+    children: ?React.Node,
+    /** Set className to the overlay wrapper */
+    className?: string,
+    /** Disable responsiveness behavior for higher-level components that want control */
+    enabled?: boolean,
+    /** Custom button to trigger for opening/closing the flyout */
+    flyoutButton: React.Element<any>,
+    /** Optional function to get the flyoutRef in parent components */
+    flyoutRefFn?: any => any,
```
tjuanitas, line 21: Is this being used?

src/components/responsive-flyout/styles/CloseButton.scss
```diff
@@ -0,0 +1,10 @@
+@import '../../../styles/variables';
+@import './variables';
+
+.btn.bdl-close-btn {
+    width: $close-btn-size;
+    height: $close-btn-size;
+    padding: unset;
```
tjuanitas, line 7: `unset` is not supported in IE11. Would `padding: 0;` work?

src/components/responsive-flyout/styles/CloseButton.scss
```diff
@@ -0,0 +1,10 @@
...
+    background-color: $bdl-gray-10;
+    border: unset;
```
tjuanitas, line 9: Would `border: none;` work instead of `unset`?

src/components/responsive-flyout/styles/OverlayHeader.scss
```diff
@@ -0,0 +1,18 @@
+@import '../../../styles/variables';
+@import './variables';
+
+.overlay-header {
+    position: relative;
+    z-index: 1;
```
tjuanitas, line 6: Are these necessary?

src/components/responsive-flyout/styles/ResponsiveFlyout.scss
```diff
@@ -0,0 +1,37 @@
...
+                top: 0;
+                right: 0;
+                bottom: 0;
+                left: 0;
+                margin: 0;
+                padding: 0;
+                background-color: $white;
+                border: unset;
+                border-radius: unset;
+                box-shadow: unset;
```
tjuanitas, line 29: Are there alternatives for `unset` here?

src/components/responsive-flyout/styles/_variables.scss
```diff
@@ -0,0 +1,3 @@
+$overlay-header-height: 64px;
+
+$close-btn-size: 32px;
```
tjuanitas, line 3: Would `$bdl-btn-height` work instead of creating this?
https://github.com/box/box-ui-elements/blob/master/src/styles/constants/_buttons.scss#L2

src/components/responsive-flyout/CloseButton.js
```diff
@@ -0,0 +1,25 @@
...
+
+const CloseButton = ({ className, onClick }: Props) => {
+    return (
+        <Button onClick={onClick} className={classNames('bdl-close-btn', className)} type="button">
```
jstoffan, line 19: Ideally, this would be named in a Flyout-specific way to ensure it's scoped properly. We also want to use [SUIT naming conventions](https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md), so maybe `bdl-ResponsiveFlyoutClose`

src/components/responsive-flyout/OverlayHeader.js
```diff
@@ -0,0 +1,33 @@
...
+const OverlayHeader = (props: Props) => {
+    const { children, onCloseClick, className } = props;
+
+    const handleClick = event => {
+        event.preventDefault();
+        event.stopPropagation();
+    };
+
+    return (
+        <div role="presentation" onClick={handleClick} className={classNames('overlay-header', className)}>
```
jstoffan, line 26: Let's use SUIT naming convention in all our new code: `bdl-ResponsiveFlyout` should work here

src/components/responsive-flyout/OverlayHeader.js
jstoffan: Is this actually needed? Since `OverlayHeader` is used within `Overlay`, `Flyout` should ignore any clicks just like it would for other content within the overlay.

src/components/responsive-flyout/OverlayHeader.js
```diff
@@ -0,0 +1,33 @@
...
+            <div className="oh-content">{children}</div>
```
jstoffan, line 27: `bdl-ResponsiveFlyout-content` and so on for all new elements

src/components/responsive-flyout/ResponsiveFlyout.js
```diff
@@ -0,0 +1,70 @@
...
+        position: 'bottom-left',
+        enabled: true,
+    };
+
+    constructor(props) {
+        super(props);
+        this.flyoutRef = React.createRef();
+    }
+
+    render() {
```
jstoffan, line 41: We should try to build all new components as functions rather than classes.

src/components/responsive-flyout/ResponsiveFlyout.js
```diff
@@ -0,0 +1,70 @@
...
+            flyoutRefFn,
+            ...rest
+        } = this.props;
+
+        return (
+            <Flyout ref={this.flyoutRef} className={classNames({ 'responsive-flyout': enabled }, className)} {...rest}>
+                {flyoutButton}
+                <Overlay className={overlayClassName}>
+                    <OverlayHeader onCloseClick={() => this.flyoutRef.current.closeOverlay()}>{header}</OverlayHeader>
+                    {children != null && (
```
jstoffan, line 59: We should always use strict equality. That said, is this check needed?

src/components/responsive-flyout/styles/_variables.scss
```diff
@@ -0,0 +1,3 @@
+$overlay-header-height: 64px;
```
tjuanitas, line 1: Does this need to be a variable? It seems to be used only once?

src/components/responsive-flyout/ResponsiveFlyout.js
```diff
@@ -0,0 +1,70 @@
...
+                        <ScrollWrapper scrollRefFn={scrollRefFn} shadowSize="contain">
```
jstoffan, line 60: I don't think we want to assume the scroll behavior here. We currently allow consuming components to decide what goes in their Overlay.

src/components/responsive-flyout/styles/OverlayHeader.scss
jstoffan, line 6: We have to be careful with z-index values, since they can often conflict. Is this needed?

src/components/responsive-flyout/styles/OverlayHeader.scss
```diff
@@ -0,0 +1,18 @@
...
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    height: $overlay-header-height;
+    padding: 13px 16px;
+    background-color: white;
```
jstoffan, line 12: Let's use the bdl variables where possible?

src/components/responsive-flyout/ResponsiveFlyout.js
```diff
@@ -0,0 +1,70 @@
...
+            className,
+            overlayClassName,
+            flyoutRefFn,
+            ...rest
+        } = this.props;
+
+        return (
+            <Flyout ref={this.flyoutRef} className={classNames({ 'responsive-flyout': enabled }, className)} {...rest}>
+                {flyoutButton}
+                <Overlay className={overlayClassName}>
```
jstoffan, line 57: We currently allow the consuming components to declare their own Overlay. Do we need to have an "all-in-one" type solution here?

src/components/responsive-flyout/styles/CloseButton.scss
jstoffan, line 7: Should we be using `PlainButton`, instead?

src/components/responsive-flyout/OverlayHeader.js
ivanthai: Yes, CloseButton.js. I'm planning on moving this out to it's own module to be reused across other components

src/components/responsive-flyout/styles/OverlayHeader.scss
```diff
@@ -0,0 +1,18 @@
...
+    box-shadow: 0 2px 4px 0 rgba(0, 0, 0, .08);
```
jstoffan, line 13: Is there a standard shadow that we can reuse?

src/features/header-flyout/HeaderFlyout.js
tjuanitas, line 46: Can we sort these?

src/components/responsive-flyout/ResponsiveFlyout.js
ivanthai: I can configure Intellij to do it. I didn't get any errors on that so this was an oversight. Must not be configured, even though linting is already enabled.

src/features/header-flyout/HeaderFlyout.js
```diff
@@ -27,43 +29,44 @@ class HeaderFlyout extends React.Component<Props> {
...
+        isResponsive: false,
     };
 
     render() {
-        const { header, footer, flyoutButton, children, scrollRefFn, className, ...rest } = this.props;
+        const { header, footer, isResponsive, children, scrollRefFn, className, ...rest } = this.props;
 
         return (
-            <Flyout
+            <ResponsiveFlyout
```
jstoffan, line 39: It may be safer to move this refactor to a separate PR to ease potential reverts.

src/features/header-flyout/HeaderFlyout.js
```diff
@@ -27,43 +29,44 @@ class HeaderFlyout extends React.Component<Props> {
...
                 closeOnClick={false}
                 offset={HeaderFlyout.panelOffset}
                 className={classNames('header-flyout', className)}
                 constrainToWindow
+                header={header}
+                overlayClassName="header-flyout-overlay"
+                enabled={isResponsive}
                 {...rest}
             >
-                {flyoutButton}
```
tjuanitas, line 43: Should `flyoutButton` be removed from the prop types since it's no longer used in this file?

src/features/header-flyout/styles/HeaderFlyout.scss
```diff
@@ -19,6 +20,14 @@
     }
 }
 
+@include breakpoint($medium-screen) {
+    .responsive-flyout {
```
jstoffan, line 24: Do we need this specificity or can we target `.flyout-list-container-title` directly?

jstoffan (changes requested):

src/components/responsive-flyout/styles/ResponsiveFlyout.scss
ivanthai: No it does not. This is to preserve the non-responsive functionality of ResponsiveFlyout

src/features/header-flyout/HeaderFlyout.js
tjuanitas, line 39: Would this swapping cause any issues for existing users?

src/components/responsive-flyout/CloseButton.js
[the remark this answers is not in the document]
tjuanitas: Nvm, see https://github.com/box/box-ui-elements/pull/2878#discussion_r857949205

src/components/responsive-flyout/OverlayHeader.js
```diff
@@ -0,0 +1,33 @@
...
+    /** Components to render in the header */
+    children?: React.Node,
+    /** Set className to the overlay header */
+    className?: string,
+    /** Set click handler when close button is clicked */
+    onCloseClick?: Function,
+};
+
+const OverlayHeader = (props: Props) => {
+    const { children, onCloseClick, className } = props;
```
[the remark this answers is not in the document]
ivanthai: Don't really understand. I didn't need to add pass-through props, but I needed only a few values from it.

src/components/responsive-flyout/OverlayHeader.js
ivanthai: The problem is that Flyout does add a onClick handler to Overlay and there are behaviors associated with it. More specifically with closeOnClick

https://github.com/box/box-ui-elements/blob/a2df0ed670ec2205f777ed37833577046b073da5/src/components/flyout/Flyout.js#L405

src/components/responsive-flyout/OverlayHeader.js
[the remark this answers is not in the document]
greathmaster: I think what @jcamden is referring to hear is destructure the necessary props from the argument

ie: `({a, b, c}) => {etc}`

so we don't have pass in the `props` parameter and then destructure.

src/components/responsive-flyout/ResponsiveFlyout.js
ivanthai: This was removed already actually

src/components/responsive-flyout/OverlayHeader.js
ivanthai: "bdl-OverlayHeader"

also moved to components/flyout

src/components/responsive-flyout/ResponsiveFlyout.js
ivanthai: Oops typo. And good point

src/components/responsive-flyout/styles/OverlayHeader.scss
ivanthai: I wanted to make sure OverlayHeader is above the content since it contains a drop shadow

src/components/responsive-flyout/styles/CloseButton.scss
ivanthai: Button currently has the styling I need. However, styling might change in the next week or so according to design. Perhaps I can make the switch the: use PlainButton and provide explicit CSS styling

src/components/responsive-flyout/styles/OverlayHeader.scss
ivanthai: Not that I've seen that matches design

src/components/responsive-flyout/ResponsiveFlyout.js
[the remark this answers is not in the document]
ivanthai: Typo here, but this check has been removed

src/components/responsive-flyout/styles/OverlayHeader.scss
ivanthai: https://git.dev.box.net/Box/EndUserApp/blob/243f9a476460c77a59531506231f0aec426e6898/src/styles/_dejablue.scss#L10-L10

src/components/responsive-flyout/styles/OverlayHeader.scss
ivanthai: Changed it to bdl-OverlayHeader

src/components/responsive-flyout/OverlayHeader.js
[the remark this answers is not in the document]
jcamden: That's not a good pattern. Descriptive name ftw.

src/components/flyout/CloseButton.tsx
```diff
@@ -0,0 +1,23 @@
...
+import './CloseButton.scss';
+
+export interface CloseButtonProps {
+    className?: string;
+    onClick?: (event: React.SyntheticEvent<HTMLButtonElement>) => void;
+}
+
+const CloseButton = ({ className, onClick }: CloseButtonProps) => {
+    return (
+        <Button onClick={onClick} className={classNames('bdl-OverlayHeaderClose', className)}>
```
jstoffan, line 17: Let's make sure to include `type="button"`, since the default is `submit`, which can cause a full page reload.

src/components/flyout/OverlayHeader.scss
```diff
@@ -0,0 +1,17 @@
+@import '../../styles/variables';
+
+.bdl-OverlayHeader {
+    position: relative;
+    z-index: 1;
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    height: 64px;
```
jstoffan, line 9: Ideally, we want to avoid setting dimensions and allow content to do it for us. If this is absolutely necessary, let's use `min-height` so we don't see content overflow issues.

src/components/flyout/OverlayHeader.scss
```diff
@@ -0,0 +1,17 @@
...
+    padding: 13px 16px;
```
jstoffan, line 10: Can we use the `$bdl-grid-unit` variables for these and go with 12px and 16px to standardize?

src/components/responsive-flyout/ResponsiveFlyout.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+import './styles/ResponsiveFlyout.scss';
+
+export interface ResponsiveFlyoutProps {
+    /** Components to render in the overlay */
+    children?: React.ReactNode;
+    /** Set className to the overlay wrapper */
+    className?: string;
+    /** Disable responsiveness behavior for higher-level components that want control */
+    enabled?: boolean;
```
jstoffan, line 17: Can we go with `isEnabled` here to stick with our convention?

src/components/responsive-flyout/ResponsiveFlyout.tsx
```diff
@@ -0,0 +1,48 @@
...
+    /** Custom button to trigger for opening/closing the flyout */
+    flyoutButton: React.ReactElement;
```
jstoffan, line 19: Should this be `React.ReactNode`?

src/components/responsive-flyout/ResponsiveFlyout.tsx
```diff
@@ -0,0 +1,48 @@
...
+    const { header, flyoutButton, children, enabled = true, className, overlayClassName, ...rest } = props;
+
+    const handleCloseClick = () => {
+        if (!flyoutRef.current) return;
+
+        flyoutRef.current.closeOverlay();
+    };
+
+    return (
+        <Flyout ref={flyoutRef} className={classNames({ 'responsive-flyout': enabled }, className)} {...rest}>
```
jstoffan, line 38: Let's use SUIT naming in all our new components.

src/components/responsive-flyout/styles/ResponsiveFlyout.scss
```diff
@@ -0,0 +1,37 @@
+@import '../../../styles/variables';
+@import '../../../styles/constants/media-queries';
+
+.flyout-overlay {
+    .overlay {
```
jstoffan, line 5: Is this specificity needed?

src/components/responsive-flyout/styles/ResponsiveFlyout.scss
```diff
@@ -0,0 +1,37 @@
...
+                bottom: 0;
+                left: 0;
+                margin: 0;
+                padding: 0;
+                background-color: $white;
+                border: none;
+                border-radius: 0;
+                box-shadow: none;
+
+                .bdl-OverlayHeader {
```
jstoffan, line 31: It looks like this can be moved out of the `.overlay` wrapper to reduce its specificity.

src/components/responsive-flyout/styles/ResponsiveFlyout.scss
```diff
@@ -0,0 +1,37 @@
...
+            display: none;
+        }
+    }
+}
+
+.is-responsive-web {
+    @include breakpoint($medium-screen) {
+        .flyout-overlay.responsive-flyout {
+            // cancels out the tether styling without having to set enabled=false
+            transform: none !important;
```
jstoffan, line 16: We almost always want to avoid using `!important`, since it hampers extensibility and customization. Is there no other option here?

src/components/responsive-flyout/ResponsiveFlyout.tsx
ivanthai: I understood ReactNode could be more than just an object with type and props. ReactNode can be a fragment, array of nodes, string, etc. I don't see flyoutButton or header being more than just an element.

src/components/responsive-flyout/ResponsiveFlyout.tsx
```diff
@@ -0,0 +1,48 @@
...
+    const { header, flyoutButton, children, isEnabled = true, className, overlayClassName, ...rest } = props;
+
+    const handleCloseClick = () => {
+        if (!flyoutRef.current) return;
+
+        flyoutRef.current.closeOverlay();
+    };
+
+    return (
+        <Flyout ref={flyoutRef} className={classNames({ 'bdl-ResponsiveFlyout': isEnabled }, className)} {...rest}>
```
jstoffan, line 38: It's a little odd that `isEnabled` controls this class directly, since we'd typically expect the base class to always be applied. Should we be passing `isEnabled` down to Flyout as a prop, instead?

src/components/responsive-flyout/ResponsiveFlyout.scss
```diff
@@ -0,0 +1,32 @@
+@import '../../styles/variables';
+
+.bdl-OverlayHeader {
```
jstoffan, line 3: Should these styles should live in `OverlayHeader.scss`?

jstoffan: Or should it be a hybrid of this and what we had before, just with slightly less specificity?

```
.bdl-ResponsiveFlyout {
    .bdl-OverlayHeader {
        display: none;
    }
}

File: src/components/responsive-flyout/ResponsiveFlyout.tsx [hunk 45]
```diff
@@ -0,0 +1,48 @@
...
+    /** Components to render in the overlay */
+    children?: React.ReactNode;
+    /** Set className to the overlay wrapper */
+    className?: string;
+    /** Disable responsiveness behavior for higher-level components that want control */
+    isEnabled?: boolean;
+    /** Custom button to trigger for opening/closing the flyout */
+    flyoutButton: React.ReactNode;
+    /** Header for responsive layout */
+    header?: React.ReactNode;
```
tjuanitas, line 21: nit: can we sort this?

src/components/responsive-flyout/ResponsiveFlyout.tsx
ivanthai: I wanted to enable/disable the class and have the css styles apply correspondingly. I guess I could have isEnabled pass to Flyout via isResponsive and then apply the responsive-flyout classname to it

src/components/responsive-flyout/ResponsiveFlyout.tsx
```diff
@@ -0,0 +1,48 @@
...
+    /** Overlay classname */
+    overlayClassName?: string;
+}
+
+const ResponsiveFlyout = (props: ResponsiveFlyoutProps & FlyoutProps) => {
+    const flyoutRef = React.useRef<Flyout>(null);
+
+    const { header, flyoutButton, children, isEnabled = true, className, overlayClassName, ...rest } = props;
```
tjuanitas, line 29: nit: can we sort these props?

```
const { children, className, flyoutButton, header, isEnabled = true, overlayClassName, ...rest } = props;
```

src/components/responsive-flyout/ResponsiveFlyout.tsx
ivanthai: The other option is to just skip ResponsiveFlyout altogether

Consumers can deal with Flyout directly by passing in isResponsive and we can utilize context to pass functions to OverlayHeader like we discussed. I had a working prototype earlier

Basic structure
```
<Flyout isResponsive...>
    <Overlay ...>
        <OverlayHeader ...>
```
...
We either pass Flyout's closeOverlay or Overlay's closeOverlay handler via Context. Example below is Flyout's

In Flyout render()
```
return (
    <TetherComponent ...>
        {React.cloneElement(overlayButton, overlayButtonProps)}
        <FlyoutContext.Provider value={{closeOverlay: this.closeOverlay}}>
            {isVisible ? React.cloneElement(overlayContent, overlayProps) : null}
        </FlyoutContext.Provider>
...

```   

In OverlayHeader

```
const { closeOverlay } = React.useContext(FlyoutContext);

// closeButton onClick={closeOverlay}
```

src/styles/constants/_layout.scss
```diff
@@ -11,6 +11,9 @@ $bdl-border-radius-size-med: $bdl-border-radius-size * 1.5;
 $bdl-border-radius-size-large: $bdl-border-radius-size * 2;
 $bdl-border-radius-size-xlarge: $bdl-border-radius-size * 3;
 
+// Drop shadows
+$bdl-box-shadow-header: 0 0 10px 0 rgba(0, 0, 0, .05);
```
tjuanitas, line 15: Should this name be `$bdl-header-box-shadow` and `bdlHeaderBoxShadow`?

It seems to be the pattern used for buttons:
https://github.com/box/box-ui-elements/blob/master/src/styles/variables.js#L97-L102

src/components/responsive-flyout/ResponsiveFlyout.tsx
```diff
@@ -0,0 +1,48 @@
+// @flow
+import * as React from 'react';
+import classNames from 'classnames';
+// @ts-ignore flow
+import { Flyout, Overlay, OverlayHeader } from '../flyout';
+// @ts-ignore flow
```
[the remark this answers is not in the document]
ivanthai: Not sure, but removed ResponsiveFlyout

src/components/flyout/OverlayHeader.scss
```diff
@@ -0,0 +1,17 @@
+@import '../../styles/variables';
+
+.bdl-OverlayHeader {
+    position: relative;
```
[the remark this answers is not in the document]
ivanthai: When I remove this, the z-index isn't respected

src/components/flyout/OverlayHeader.tsx
```diff
@@ -0,0 +1,33 @@
...
+import './OverlayHeader.scss';
+
+export interface OverlayHeaderProps {
+    /** Components to render in the header */
+    children?: React.ReactNode;
+    /** Set className to the overlay header */
+    className?: string;
+}
+
+const OverlayHeader = ({ children, className }: OverlayHeaderProps) => {
```
tjuanitas, line 17: Do we need tests for this component?

src/components/close-button/CloseButton.tsx
```diff
@@ -0,0 +1,22 @@
...
+import IconClose from '../../icons/general/IconClose';
+import { bdlGray62 } from '../../styles/variables';
+import './CloseButton.scss';
+
+export interface CloseButtonProps {
+    className?: string;
+    onClick?: Function;
+}
+
+const CloseButton = ({ className, onClick }: CloseButtonProps) => {
```
tjuanitas, line 14: Since this is a new component, can we add the `index`, `stories`, `flow` files similar to: https://github.com/box/box-ui-elements/tree/master/src/components/plain-button

tjuanitas (commented):
Can the PR title be updated as well since `responsive-flyout` is no longer a component?

src/components/flyout/Flyout.scss
```diff
@@ -22,4 +22,37 @@
...
+
+    .bdl-OverlayHeader {
+        display: none;
+    }
+}
+
+.is-responsive-web {
+    @include breakpoint($medium-screen) {
+        .bdl-Flyout-responsive {
+            .flyout-overlay-enabled {
```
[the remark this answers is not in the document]
ivanthai: It's actually dynamically generated by Tether. I wanted to make sure the css override only applies to the tether component of Flyout

src/components/flyout/OverlayHeader.tsx
ivanthai: Yes, I'm going to commit that separately. I had to change a lot of things, so I wanted to focus on tests after

src/components/flyout/Flyout.js
```diff
@@ -1,11 +1,19 @@
...
 import * as React from 'react';
+import classNames from 'classnames';
+import noop from 'lodash/noop';
 import TetherComponent from 'react-tether';
 import uniqueId from 'lodash/uniqueId';
 import { KEYS } from '../../constants';
 
 import './Flyout.scss';
 
+interface FlyoutContextValues {
```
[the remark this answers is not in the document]
ivanthai: This is changing. I'm adding this to a separate file. No need to export unless it will be used elsewhere I believe

ivanthai:
Just updated the PR description

- Added/updated tests and documentation

src/components/close-button/CloseButton.tsx
```diff
@@ -0,0 +1,25 @@
...
+import IconClose from '../../icons/general/IconClose';
+import { bdlGray62 } from '../../styles/variables';
+
+import './CloseButton.scss';
+
+export interface CloseButtonProps {
+    /** Custom class for the close button */
+    className?: string;
+    /** onClick handler for the close button */
+    onClick?: Function;
```
jcamden, line 14: I love that you are using Typescript! Typescript brother! 😄  Anyway, in my experience, it is good to make types as specific as possible (like `any` is the worst type, `object` is pretty bad too, etc. Those are really escape hatches.) In this case, that means that you could upgrade `Function` to `()=>void` if you like :)

src/components/close-button/CloseButton.tsx
```diff
@@ -0,0 +1,25 @@
...
+}
+
+const CloseButton = ({ className, onClick }: CloseButtonProps) => {
```
jcamden, line 17: Another Typescript pattern that I really like is to use `React.FC` and provide props via a generic:
`const CloseButton: React.FC<CloseButtonProps> = happiness`
I don't know if any official decisions have been made, actually, about how to type React components. Might be a good topic for a very mini RFC.

Speaking of RFC, we may able to replace default exports with named exports in the context of Typescript sometime soon: https://cloud.box.com/s/50b9hr6z5t7213cn79052nfqxn5vhzun . However, that RFC is still unresolved afaik, so default for now; but I thought you'd be interested :)

src/components/close-button/CloseButton.tsx
jcamden, line 14: I love that you are using Typescript! Typescript brother! 😄 Anyway, in my experience, it is good to make types as specific as possible (like `any` is the worst type, `object` is pretty bad too, etc. Those are really escape hatches.) In this case, that means that you could upgrade `Function` to `()=>void` if you like :)

src/components/close-button/CloseButton.tsx
jcamden, line 17: Another Typescript pattern that I really like is to use `React.FC` and provide props via a generic:
`const CloseButton: React.FC<CloseButtonProps> = ({prop1, prop2}) => <>happiness</>`
I don't know if any official decisions have been made, actually, about how to type React components. Might be a good topic for a very mini RFC.

Speaking of RFC, we may able to replace default exports with named exports in the context of Typescript sometime soon: https://cloud.box.com/s/50b9hr6z5t7213cn79052nfqxn5vhzun . However, that RFC is still unresolved afaik, so default for now; but I thought you'd be interested :)

src/components/close-button/__tests__/CloseButton.test.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { shallow } from 'enzyme';
```
jcamden, line 2: I believe our policy is to compose new component-level unit tests using @testing-library/react (RTL) instead of enzyme. (That is because enzyme usually targets implementation details, whereas RTL forces a user-perspective approach to tests.) I've gotten pretty decent with RTL over the last few weeks, so if you want any guidance on this, I would be happy to help out :) 

For reference:
My most recent RTL tests are in my latest PR, e.g.: https://git.dev.box.net/Box/EndUserApp/pull/8121/files#diff-0de84049a19f96b8eca357bf3a4b566ae22a135dd1711513db91f83425dc1c87

In EUA, you could check out /src/components/core/left-nav/components/__tests__/LeftNav.test.js
In BUIE, there is also my first RTL work: src/features/collapsible-sidebar/__tests__/CollapsibleSidebar.test.js

src/components/flyout/OverlayHeader.tsx
jcamden: When you compose that test, let's make sure it's RTL instead of enzyme (c.f. comment above). I can help with that :)

src/components/close-button/__tests__/CloseButton.test.tsx
```diff
@@ -0,0 +1,14 @@
...
+import CloseButton from '../CloseButton';
+import Button from '../../button';
+
+describe('components/close-button/CloseButton', () => {
+    test('should correctly render', () => {
+        const wrapper = shallow(<CloseButton />);
+
+        expect(wrapper.find(Button).length).toEqual(1);
+        expect(wrapper.hasClass('bdl-CloseButton')).toBe(true);
+        expect(wrapper).toMatchSnapshot();
```
jstoffan, line 12: We generally want to avoid snapshot testing. Do we need to add some more explicit assertions, instead?

src/components/flyout/__tests__/OverlayHeader.test.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import sinon from 'sinon';
+
+import FlyoutContext from '../FlyoutContext';
+import OverlayHeader from '../OverlayHeader';
+
+const sandbox = sinon.sandbox.create();
```
jstoffan, line 8: We've been moving away from sinon. Can we use Jest for any mocks or spies we need?

src/components/flyout/index.js
```diff
@@ -1,3 +1,5 @@
 // @flow
 export { default as Flyout } from './Flyout';
 export { default as Overlay } from './Overlay';
+export { default as OverlayHeader } from './OverlayHeader';
+export { default as FlyoutContext } from './FlyoutContext';
```
jstoffan, line 5: Minor, but can we alphabetize our exports?

src/components/close-button/__tests__/CloseButton.stories.test.tsx
```diff
@@ -0,0 +1,9 @@
+describe('components/close-button/CloseButton', () => {
+    test.each(['components-buttons-closebutton--regular', 'components-buttons-closebutton--disabled'])(
```
tjuanitas, line 2: I think `components-buttons-closebutton--disabled` can be removed since there's no disabled in the `stories.tsx`.

tjuanitas: Then this `src/components/close-button/__tests__/__image_snapshots__/close-button-stories-test-tsx-components-close-button-close-button-looks-visually-correct-when-using-story-components-buttons-closebutton-disabled-1-snap.png` can be removed.

src/components/flyout/Flyout.scss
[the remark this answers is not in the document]
tjuanitas: Can we add a comment about this class being generated by Tether?

tjuanitas (commented):
It looks like there are some snapshots for the `collapsible` component being updated:
`src/components/collapsible/__tests__/__image_snapshots__/collapsible-stories-test-tsx-components-collapsible-collapsible-collapses-content-for-story-components-collapsible-with-border-1-snap.png`
Are those related to this update? If not, can we remove them for now?

src/components/close-button/CloseButton.scss
```diff
@@ -0,0 +1,10 @@
+@import '../../styles/variables';
+
+.btn.bdl-CloseButton {
```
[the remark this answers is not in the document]
ivanthai: In some cases, it will not

src/components/close-button/__tests__/CloseButton.stories.test.tsx
```diff
@@ -0,0 +1,6 @@
+describe('components/close-button/CloseButton', () => {
+    test.each(['components-buttons-closebutton--regular'])('looks visually correct when using story %s', async id => {
```
jstoffan, line 2: It looks like we can drop back to a regular `test` here now.

src/components/close-button/CloseButton.tsx
```diff
@@ -0,0 +1,30 @@
...
+    return (
+        <Button
+            onClick={onClick}
+            className={classNames('bdl-CloseButton', className)}
+            type={ButtonType.BUTTON}
+            data-testid="bdl-CloseButton"
```
jstoffan, line 23: Minor, but we typically keep props alphabetized where possible.

jstoffan (dismissed):
LGTM, nice job. 👍

tjuanitas (approved):
Looks great! Thank you for working through the feedback!

tjuanitas:
Force merging this since Semantic Pull Request is down and passes other checks.

PR #2880, opened 2022-04-26
feat(persistent-onboarding): Add renderAvatar property to ContentPreview

robertjk:
In Persistent Onboarding project I need to add a tooltip over `Avatar` component inside Preview. For that to work I defined `renderAvatar` property, which can be passed down in EUA and wrap this component in a tooltip. If `renderAvatar` is not defined `Avatar` will render normally, so the old code will still work as it used to.

### DIFF
--- a/src/elements/content-sidebar/activity-feed/comment-form/CommentForm.js
+++ b/src/elements/content-sidebar/activity-feed/comment-form/CommentForm.js
@@ -36,6 +36,7 @@ type Props = {
     onCancel: Function,
     onFocus?: Function,
     onSubmit?: Function,
+    renderCommentFormAvatar?: Function,
     showTip?: boolean,
     tagged_message?: string,
     updateComment?: Function,
@@ -49,6 +50,7 @@ type State = {
 class CommentForm extends React.Component<Props, State> {
     static defaultProps = {
         isOpen: false,
+        renderCommentFormAvatar: avatar => avatar,
     };
 
     state = {
@@ -119,6 +121,7 @@ class CommentForm extends React.Component<Props, State> {
             tagged_message,
             getAvatarUrl,
             showTip = true,
+            renderCommentFormAvatar,
         } = this.props;
         const { commentEditorState } = this.state;
         const inputContainerClassNames = classNames('bcs-CommentForm', className, {
@@ -129,7 +132,7 @@ class CommentForm extends React.Component<Props, State> {
             <Media className={inputContainerClassNames}>
                 {!isEditing && (
                     <Media.Figure className="bcs-CommentForm-avatar">
-                        <Avatar getAvatarUrl={getAvatarUrl} user={user} />
+                        {renderCommentFormAvatar(<Avatar getAvatarUrl={getAvatarUrl} user={user} />)}
                     </Media.Figure>
                 )}
 
--- a/src/elements/content-sidebar/ActivitySidebar.js
+++ b/src/elements/content-sidebar/ActivitySidebar.js
@@ -61,6 +61,7 @@ type ExternalProps = {
     onTaskDelete: (id: string) => any,
     onTaskUpdate: () => any,
     onTaskView: (id: string, isCreator: boolean) => any,
+    renderCommentFormAvatar?: Function,
 } & ErrorContextProps &
     WithAnnotatorContextProps;
 
@@ -719,6 +720,7 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
             activeFeedEntryId,
             activeFeedEntryType,
             onTaskView,
+            renderCommentFormAvatar,
         } = this.props;
         const {
             currentUser,
@@ -768,6 +770,7 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
                     onTaskUpdate={this.updateTask}
                     onTaskView={onTaskView}
                     onVersionHistoryClick={onVersionHistoryClick}
+                    renderCommentFormAvatar={renderCommentFormAvatar}
                 />
             </SidebarContent>
         );
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
@@ -52,6 +52,7 @@ type Props = {
     onTaskUpdate?: Function,
     onTaskView?: Function,
     onVersionHistoryClick?: Function,
+    renderCommentFormAvatar?: Function,
     translations?: Translations,
 };
 
@@ -222,6 +223,7 @@ class ActivityFeed extends React.Component<Props, State> {
             onTaskUpdate,
             onTaskView,
             onVersionHistoryClick,
+            renderCommentFormAvatar,
             translations,
         } = this.props;
         const { isInputOpen } = this.state;
@@ -327,6 +329,7 @@ class ActivityFeed extends React.Component<Props, State> {
                         user={currentUser}
                         onCancel={this.commentFormCancelHandler}
                         onFocus={this.commentFormFocusHandler}
+                        renderCommentFormAvatar={renderCommentFormAvatar}
                         getAvatarUrl={getAvatarUrl}
                     />
                 ) : null}


src/elements/content-sidebar/activity-feed/comment-form/CommentForm.js
```diff
@@ -129,7 +134,7 @@ class CommentForm extends React.Component<Props, State> {
             <Media className={inputContainerClassNames}>
                 {!isEditing && (
                     <Media.Figure className="bcs-CommentForm-avatar">
-                        <Avatar getAvatarUrl={getAvatarUrl} user={user} />
+                        {renderAvatar(<Avatar getAvatarUrl={getAvatarUrl} user={user} />)}
```
megansmith-box, line 137: Should this be renamed to something like renderAvatarWrapper? or renderAvatarTooltip?

megansmith-box, line 137: Do you have a screenshot of what this looks like?

src/elements/content-preview/ContentPreview.js
```diff
@@ -108,6 +108,7 @@ type Props = {
         [name: string]: TargetingApi,
     },
     previewLibraryVersion: string,
+    renderAvatar?: Function,
```
jstoffan, line 111: We generally want to avoid exposing more props on our top-level Elements. Can we pass this new value in via the existing `activitySidebarProps` object, instead?

src/elements/content-sidebar/activity-feed/comment-form/CommentForm.js
```diff
@@ -119,6 +123,7 @@ class CommentForm extends React.Component<Props, State> {
             tagged_message,
             getAvatarUrl,
             showTip = true,
+            renderAvatar,
```
jstoffan, line 126: It would be helpful to understand how this is intended to be used. Feel free to reach out in the Preview team channel if it's more convenient.

jstoffan (changes requested):

src/elements/content-sidebar/activity-feed/comment-form/CommentForm.js
robertjk: > renderAvatarTooltip?

I consciously avoided very specific name that would connect this property with rendering a tooltip. That is because potentially a property like this could be reused in other contexts; those contexts could add something totally different than the tooltip.

> renderAvatarWrapper

I see value in `renderAvatarWrapper` vs `renderAvatar` - it clearly expresses we are wrapping the avatar with something. One problem I see with that particular name is that Wrapper components refer only to the outside component but not the inside component, so `renderAvatarWrapper` might suggest rendering the Wrapper but not the Avatar itself. `renderAvatarWithWrapper` might be better. Or `wrapAvatar`, but that one drops `render` prefix, which connects to [Render Props](https://reactjs.org/docs/render-props.html) pattern which this exactly is.

robertjk: I have and it's in the PR for Comments tooltip, which actually pass this property to `PreviewContent`. I'll gladly share it with you, but I was told to not post links to our internal Box systems in this public repo.

And then I don't feel that's connected to this PR. This PR adds a property on general purpose component `PreviewContent`, which can be possibly reused in other contexts than just the tooltip I'm now implementing. So it should be examined in this general context as well and not in the context of the tooltip I'm working on. At least that's how I see it. Please tell me if you feel otherwise.

src/elements/content-preview/ContentPreview.js
robertjk: That sounds like a great idea, thanks. I wasn't sure what those properties are doing. Also that's my first PR to box-ui-elements and I'm still not familiar with how things are being done here. Let me refactor it...

src/elements/content-sidebar/activity-feed/comment-form/CommentForm.js
megansmith-box: I think `renderAvatarWithWrapper` makes sense.

ConradJChan: I hate to make it more verbose, but would it make sense to have it be `renderCommentFormAvatar` to be more descriptive of what it does? Initially upon reading the prop I thought it was for all avatars in the activity feed.

src/elements/content-sidebar/activity-feed/comment-form/CommentForm.js
```diff
@@ -129,7 +132,7 @@ class CommentForm extends React.Component<Props, State> {
             <Media className={inputContainerClassNames}>
                 {!isEditing && (
                     <Media.Figure className="bcs-CommentForm-avatar">
-                        <Avatar getAvatarUrl={getAvatarUrl} user={user} />
+                        {renderAvatar(<Avatar getAvatarUrl={getAvatarUrl} user={user} />)}
```
ConradJChan, line 135: Another thought is, what if instead of a decorating render prop we gave the render prop full control? The `renderCommentFormAvatar` could by default render `<Avatar ... />` but if the render prop is provided we let the parent component render whatever they want

robertjk: > I hate to make it more verbose, but would it make sense to have it be renderCommentFormAvatar to be more descriptive of what it does? Initially upon reading the prop I thought it was for all avatars in the activity feed.

That's actually a great suggestion. Indeed I didn't think that there might be more avatar occurrences than just in `CommentForm`.

This only collides a bit with suggestions in [this thread](https://github.com/box/box-ui-elements/pull/2880#discussion_r859107218). Merging them both would result in `renderCommentFormAvatarWithWrapper`, which for me feels like a little too much. Can I ask you to please voice your thoughts in that other thread?

robertjk: > Another thought is, what if instead of a decorating render prop we gave the render prop full control? The renderCommentFormAvatar could by default render <Avatar ... /> but if the render prop is provided we let the parent component render whatever they want

Isn't that what happens right now? The parent component can render whatever they want. We pass `<Avatar>` to it (as `children` property), but the parent can totally ignore it and render whatever they want. Unless I'm missing something.

src/elements/content-sidebar/activity-feed/comment-form/CommentForm.js
jstoffan: It would be helpful to have a screenshot of the intended experience, if one is available. Why are we diverging from the other FTUX tooltips, like the one in `SidebarNavSign`?

src/elements/content-sidebar/activity-feed/comment-form/CommentForm.js
robertjk: >

> I hate to make it more verbose, but would it make sense to have it be `renderCommentFormAvatar` to be more descriptive of what it does? Initially upon reading the prop I thought it was for all avatars in the activity feed.

That definitely makes sense. I didn't think there might be more avatar occurrences than just in `CommentForm`.

For now I refactored the name to `renderCommentFormAvatar`. I'm discussing with @ConradJChan how this function should work, which will affect the other changes to the name. In the current form I guess `WithWrapper` isn't the best, as `renderAvatar`/`renderCommentFormAvatar` might wrap the avatar in something or it might now - it's up to the parent component to decide what they do with the avatar; they might even replace it with some totally different implementation of avatar.

ConradJChan:
Changes look good to me, but there is an outstanding question from @jstoffan. Also, are there any unit tests that need to be updated as well?

src/elements/content-sidebar/activity-feed/comment-form/CommentForm.js
robertjk: > It would be helpful to have a screenshot of the intended experience, if one is available.

@megansmith-box [asked](https://github.com/box/box-ui-elements/pull/2880#discussion_r859109757) for the same thing. Isn't that a bit out of context of this change? I mean: This is a general purpose repository. Even if I'm extending a component here with a property I'll use in EndUserApp, the change in this repo should be examined in a more general context than that of my change in EUA. I have the screenshot of how the tooltip displayed on `CommentForm`'s `Avatar` looks like in EndUserApp's PR, which is PR that implements this tooltip. I could possibly link it here, but I was told to not post any links to our internal tools (Jira, GHE etc.) in this public repo. Is that true?

> Why are we diverging from the other FTUX tooltips, like the one in SidebarNavSign?

That's a good question! I consciously diverge from this pattern, as I think the old pattern does not make a good separation of concerns. `box-ui-elements` is a general purpose library exposing reusable UI components. They should be possible to be reused in any repository, or even in a context outside of Box organization (Am I wrong on that point?). Having concrete implementations of Box tooltips inside this repository does not follow that. First those tooltips are not in any way reusable - there are tied to some unique experiences for users in `EndUserApp` (or `AdminConsole` or some other parent repo). Second they are totally irrelevant for anyone outside of Box organisation.

So instead of defining my tooltip in `box-ui-elements` I decided it's better to define it in `EndUserApp` and pass it as a [Render Prop](https://reactjs.org/docs/render-props.html) to the component we need to display them on (wrap that component in my tooltip). That way `box-ui-elements` doesn't need to be aware of what exactly is done to the `Avatar`, how the tooltip is constructed nor what our user experiences are. All this logic will stay in `EndUserApp`, which is where I think it belongs much better.

Another reason is that this particular Persistent Onboarding tooltip I'm putting on `Avatar` is just one of the few tooltips in a bigger experience. 3 of those tooltips are applied on elements declared in `EndUserApp`, one on an element in `box-ui-elements` and one on an element in `box-content-preview`. I think's it's great to have the code for the whole experience defined in a single place, which also makes code reuse easy/possible. Following the pattern I'm divering from - declaring all tooltip logic in `box-ui-elements` (which would also mean declaring one tooltip in `box-content-preview) - would split a tightly related code in 3 repositories and would make code reuse very hard or impossible.

I'd say that's the pattern we should use in the future to declare tooltips on elements hidden somewhere deep in `box-ui-elements` - to pass render props and leave the tooltip declarations in parent repo (`EndUserApp`, `AdminConsole` etc.).

@jstoffan @ConradJChan @megansmith-box What do you think?

PR #2881, opened 2022-04-27
refactor(typescript): Migrate InfiniteScroll to typescript

apieceofbart:
My attempt at migrating InfiniteScroll to typescript.


src/components/infinite-scroll/InfiniteScroll.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+    function getContainer(): EventTarget | undefined {
+        return useWindow ? window : scrollContainerNode;
+    }
+
+    React.useEffect(() => {
+        function onContainerScroll() {
+            if (isLoading || !hasMore) return;
+
+            if (sentinelRef.current == null) return;
```
tjuanitas, line 50: Can strict equality checking be used here?

src/components/infinite-scroll/InfiniteScroll.tsx
```diff
@@ -0,0 +1,108 @@
...
+    scrollContainerNode,
+    threshold = 100,
+    throttle = 64,
+    useWindow = false,
+}: Props) {
+    const sentinelRef = React.createRef<HTMLDivElement>();
+
+    function getContainer(): EventTarget | undefined {
+        return useWindow ? window : scrollContainerNode;
+    }
```
tjuanitas, line 44: Would it work to do `const container = useWindow ? window : scrollContainerNode;` here so that `getContainer()` doesn't need to be called in `addEventListeners` and `removeEventListeners`?

src/components/infinite-scroll/InfiniteScroll.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+        function addEventListeners() {
+            const container = getContainer();
+            if (container) {
+                container.addEventListener('scroll', scrollHandler);
+                container.addEventListener('resize', resizeHandler);
+            }
+        }
+
+        addEventListeners();
```
tjuanitas, line 77: Is the function necessary/preferred since we're immediately invoking it? Would having just the contents here make sense?

src/components/infinite-scroll/InfiniteScroll.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+        return function removeEventListeners() {
```
tjuanitas, line 79: This might be preference but is the named function necessary?

src/components/infinite-scroll/InfiniteScroll.tsx
```diff
@@ -0,0 +1,108 @@
...
+    useWindow = false,
+}: Props) {
+    const sentinelRef = React.createRef<HTMLDivElement>();
+
+    function getContainer(): EventTarget | undefined {
+        return useWindow ? window : scrollContainerNode;
+    }
+
+    React.useEffect(() => {
+        function onContainerScroll() {
```
tjuanitas, line 47: Would it better to declare this outside of the `React.useEffect`?

src/components/infinite-scroll/__tests__/InfiniteScroll.test.tsx
```diff
@@ -145,31 +150,14 @@ describe('components/infinite-scroll/InfiniteScroll', () => {
...
-        expect(mockAddEventListeners).toHaveBeenCalledTimes(1);
-        expect(mockRemoveEventListeners).toHaveBeenCalledTimes(1);
-
-        component.setProps({
-            scrollContainerNode: document.createElement('div'),
-        });
-        expect(mockAddEventListeners).toHaveBeenCalledTimes(2);
-        expect(mockRemoveEventListeners).toHaveBeenCalledTimes(2);
-    });
-
```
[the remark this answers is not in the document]
tjuanitas: I think identifying if we're removing the appropriate eventListeners would be beneficial. Is there another way we could test/verify this?

src/components/infinite-scroll/__tests__/InfiniteScroll.test.tsx
```diff
@@ -145,31 +150,14 @@ describe('components/infinite-scroll/InfiniteScroll', () => {
...
     describe('with sentinel in range', () => {
-        let component;
+        let component: ReturnType<typeof mount>;
```
tjuanitas, line 160: It might be better to import `ReactWrapper` from enzyme although there doesn't seem to be a direct equivalent to `ShallowWrapper` for mount: https://enzymejs.github.io/enzyme/docs/api/ReactWrapper/mount.html

```suggestion
        let component: ReactWrapper;
```

tjuanitas (commented):
Thanks for taking initiative on this migration!

src/components/infinite-scroll/InfiniteScroll.tsx
```diff
@@ -0,0 +1,108 @@
...
+    scrollContainerNode?: HTMLElement;
+    /** Scroll threshold */
+    threshold?: number;
+    /** Throttle rate */
+    throttle?: number;
+    /** Set useWindow to true to use the window as scroll container. If set to true, will ignore scrollContainerRef. */
+    useWindow?: boolean;
+};
+
+function InfiniteScroll({
```
jcamden, line 30: This is definitely a preference thing, but I like to use React.FC and pass props as a generic (and I also like arrow functions):
`const InfiniteScroll: React.FC<InfiniteScrollProps> = ({kitty, turtle}) => <div>kittle is {kitty + turtle} </div>`
From my perspective, would be cool if we ended up standardized using React.FC<Props> in the migration docs 🙏

src/components/infinite-scroll/InfiniteScroll.tsx
```diff
@@ -0,0 +1,108 @@
...
+    isLoading,
+    onLoadMore,
+    scrollContainerNode,
+    threshold = 100,
+    throttle = 64,
+    useWindow = false,
+}: Props) {
+    const sentinelRef = React.createRef<HTMLDivElement>();
+
+    function getContainer(): EventTarget | undefined {
```
jcamden, line 42: Is it possible to move `getContainer` outside of the component? Either way, do we want getContainer() to take window as a prop instead of relying on scope?

src/components/infinite-scroll/InfiniteScroll.tsx
```diff
@@ -0,0 +1,108 @@
...
+    children,
+    hasMore,
+    isLoading,
+    onLoadMore,
+    scrollContainerNode,
+    threshold = 100,
+    throttle = 64,
+    useWindow = false,
+}: Props) {
+    const sentinelRef = React.createRef<HTMLDivElement>();
```
jcamden, line 40: I imagine that you are not the inventor of this, since this is just a Typescript migration; but refs: \<shudder\>

src/components/infinite-scroll/InfiniteScroll.tsx
```diff
@@ -0,0 +1,108 @@
...
+        isLoading,
+        onLoadMore,
+        scrollContainerNode,
+        sentinelRef,
+        threshold,
+        throttle,
+        useWindow,
+    ]);
+
+    const sentinel = <div ref={sentinelRef} data-testid="sentinel" />;
```
jcamden, line 98: What is the reason to assign this to var instead of just declaring it in line 103?

src/components/index.ts
```diff
@@ -31,6 +31,9 @@ export * from './ghost';
 export { default as Header } from './header';
 export * from './header';
 
+export { default as InfiniteScroll } from './infinite-scroll';
```
jcamden, line 34: FYI, we may soon be able to use prefer named exports in Typescript files. There is an RFC, [Coupled Migrations of Typescript and Named Exports in BUIE](https://box.slack.com/archives/CG8KSE9DJ/p1645049343055679), which I believe might be taken up by web-arch-group on Monday, so that'd be cool. For now, still default exports.

src/components/infinite-scroll/__tests__/InfiniteScroll.test.tsx
```diff
@@ -37,21 +39,20 @@ describe('components/infinite-scroll/InfiniteScroll', () => {
     it('should render with default props', () => {
         const component = mount(<InfiniteScroll {...propsList} />);
         expect(component).toMatchInlineSnapshot(`
-<InfiniteScroll
-  hasMore={false}
-  isLoading={false}
-  onLoadMore={[MockFunction]}
-  threshold={100}
-  throttle={64}
```
mickr, line 45: Should this still get passed in for the test?

src/components/infinite-scroll/InfiniteScroll.tsx
mickr: @jcamden what are the benefits of React.FC? I know there is some controversy surrounding it...

src/components/infinite-scroll/__tests__/InfiniteScroll.test.tsx
```diff
@@ -1,19 +1,21 @@
...
 
 const threshold = 100;
-const propsList = {
+const propsList: Props = {
     isLoading: false,
     hasMore: false,
     useWindow: true,
     onLoadMore: mockOnLoadMore,
     threshold,
+    children: null,
```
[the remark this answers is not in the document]
apieceofbart: what do you mean? children were a required prop and those tests were wrong - they did not provide the children (because they were not typed) 
`null` is a valid value for children and these props are only used in tests that don't require children so I think it's ok

src/components/infinite-scroll/InfiniteScroll.tsx
```diff
@@ -0,0 +1,108 @@
+/**
+ * From https://github.com/jaredpalmer/react-simple-infinite-scroll
+ * Updated to accept a scroll container React ref as the "window"
+ */
+import * as React from 'react';
+import throttleFn from 'lodash/throttle';
+
+export type Props = {
```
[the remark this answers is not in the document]
apieceofbart: I prefer type for consistency, as soon as you start writing unions you can only uses types so we'll end up with both interface and type in our code anyway. But I don't feel that strongly so I can change here.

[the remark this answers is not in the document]
apieceofbart: Sure

src/components/infinite-scroll/InfiniteScroll.tsx
apieceofbart: you mean if the name is necessary? it's for readability, I can make it anonymous if you prefer

src/components/infinite-scroll/InfiniteScroll.tsx
apieceofbart: again, readability, we can get rid of it

src/components/infinite-scroll/InfiniteScroll.tsx
apieceofbart: yeah, I would even go as far as banning `==` completely

src/components/infinite-scroll/__tests__/InfiniteScroll.test.tsx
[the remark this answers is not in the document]
apieceofbart: oh you mean, sort alphabetically? why isn't their a prettier rule in place? feels like it should be

src/components/infinite-scroll/InfiniteScroll.tsx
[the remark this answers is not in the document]
apieceofbart: Plus, I looked at some of the `tsx` files and it looks like we're mostly using `type` 🤷

src/components/infinite-scroll/InfiniteScroll.tsx
apieceofbart: I don't really think it would beneficial but I don't mind moving outside the component

src/components/infinite-scroll/__tests__/InfiniteScroll.test.tsx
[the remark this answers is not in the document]
apieceofbart: I can write a test for that but bear in mind that those removed tests were not checking that we're removing any listeners, merely that we're calling some internal methods.

src/components/infinite-scroll/InfiniteScroll.tsx
apieceofbart: Sorry, I disagree completely here and I would ban using React.FC altogether.  I saw milion times people using it by default even when their components were not using `children`  (FC adds children by default). 
So I think it's always better to be explicit and type children per component basis. 
As for arrow function - I am not a fan unless it's lambda or you need to keep `this` value. Not sure why people prefer it over `function` - for my eyes they're less readable.

So unless you feel strongly about it I would prefer to keep `function` keyword

apieceofbart: BTW: all those React.FC that you have in the code will be a problem when moving to @types/react 18 since they removed the `children` :)

src/components/infinite-scroll/InfiniteScroll.tsx
apieceofbart: That was the original code, I am happy to change that

src/components/index.ts
apieceofbart: Yes, defaults are a bad idea imo (and I think there's a general consensus about it now in the community) so I totally agree with that

src/components/infinite-scroll/__tests__/InfiniteScroll.test.tsx
apieceofbart: I don't care really, just needed something to make TS happy

src/components/infinite-scroll/InfiniteScroll.tsx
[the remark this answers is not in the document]
jcamden: I can't come up with a good reason. I've always used interfaces for objects, but there is almost nothing that interfaces can do that types can't do as well (e.g. in place of extends, types have unions). One difference is that unions and intersections of interfaces can be assigned to types, but unions and intersections of types cannot be assigned to interfaces of course. But that has basically no weight against a types-only position. I think my preference may have been based on a faulty assumption that utility types only applied to interfaces. Anyway, now it's just habit---not a valid argument.

src/components/infinite-scroll/InfiniteScroll.tsx
apieceofbart: yeah, I'll change that

src/components/infinite-scroll/InfiniteScroll.tsx
apieceofbart: I just inlined it as variable per other comment

src/components/infinite-scroll/InfiniteScroll.tsx
jcamden: Yep, you're totally right. I just learned another thing.

src/components/infinite-scroll/InfiniteScroll.tsx
jcamden: I think persuasion and consensus are really import. Let me try again, and if I'm not persuasive, no worries :) `onContainerScroll` has closure from `InfiniteScroll`. This increases the chances of accidentally making use of sideEffects from `InfiniteScroll`. Moving it out of the closure and requiring is dependencies to be passed as parameters protects against this hazard. It also has the added benefits of making that function testable and reusable. What are the benefits of leaving as is?

src/components/infinite-scroll/InfiniteScroll.tsx
jcamden: For my two cents, I have never done this, but when I saw it, I thought, hey, that's not a bad idea. In this case it's pretty clear that the function removes the event listeners, but this might be a good, explicit pattern to establish.

src/components/infinite-scroll/InfiniteScroll.tsx
jcamden: As a follow up, if we're not going to use React.FC, would it be a good idea for us to type the return value (i.e. React.ReactNode) instead of relying on inference? (Btw, thanks for helping us discuss Typescript with this PR :) )

src/components/infinite-scroll/__tests__/InfiniteScroll.test.tsx
mickr: Sounds good to me

src/components/infinite-scroll/InfiniteScroll.tsx
```diff
@@ -0,0 +1,126 @@
...
+    /** Set useWindow to true to use the window as scroll container. If set to true, will ignore scrollContainerRef. */
+    useWindow?: boolean;
+}
+
+type OnContainerScrollParams = Pick<Required<InfiniteScrollProps>, 'threshold' | 'useWindow'> &
+    Pick<InfiniteScrollProps, 'isLoading' | 'hasMore' | 'scrollContainerNode' | 'onLoadMore'> & {
+        sentinelRef: React.RefObject<HTMLDivElement>;
+    };
+
+function onContainerScroll({
```
tjuanitas, line 35: Would it make sense to still keep this inside the `InfiniteScroll` component but before the `React.useEffect` block? I.e. declare this under `const sentinelRef` on line 74.

This way we don't need to pass in any params since we should have access to the `InfiniteScroll` props.

jcamden: @tjuanitas , I suggested extracting it outside of the component: 
https://github.com/box/box-ui-elements/pull/2881#discussion_r861718607 
What d'you think?

tjuanitas: Gotcha. Yeah, I thought it might look cleaner since we're always passing in the same parameters from the parent props when we call `onContainerScroll`. The calls when we declare `scrollHandler` and `resizeHandler` and any new ones could then be condensed.

Also, this is preference, but the declaration for `OnContainerScrollParams` looked odd to me lol.

I'm okay with leaving it outside though if there's a benefit to it.

jcamden: It would be nice not to have such chunky `scrollHandler` and `resizeHandler` calls. Could we condense those props (except for sentinelRef) using a rest parameter? Here is the case that I made for breaking the function out:
> `onContainerScroll` has closure from `InfiniteScroll`. This makes it possible to accidentally make use of sideEffects from InfiniteScroll. Moving it out of the closure and requiring its dependencies to be passed as parameters protects against this hazard. It also has the added benefits of making that function testable and reusable.

We may not end up unit testing or reusing `onContainerScroll`; so I think, for the most part, the question here is, how valuable is it to isolate functions from the side effects they would have access to if composed in the scope of the component?

jcamden: I personally like the picks and intersections that @apieceofbart used in that declaration. I also can't think of an alternative except for duplicating the types 🤔

tjuanitas: Hey @apieceofbart, I talked with @jcamden and let's keep this as-is but can we sort the parameters? Here and in the declarations for `scrollHandler` and `resizeHandler`.

tjuanitas (approved):

PR #2844, opened 2022-03-03
feat(collapsible-sidebar): make CollapsibleSidebar hide-able

jcamden:
Manual test on dev vm:
width 475px:
Screen Shot 2022-03-03 at 4 48 46 PM

width 474px:
Screen Shot 2022-03-03 at 4 48 51 PM

unit tests are green:
Screen Shot 2022-03-03 at 4 59 43 PM

PR #2882, opened 2022-04-28
feat(collapsible-sidebar): make CollapsibleSidebar fullscreen

joshmarnold:


src/features/collapsible-sidebar/CollapsibleSidebar.scss
```diff
@@ -111,6 +111,16 @@ $left-sidebar-z-index: 40 !default;
     }
 }
 
+@include breakpoint($small-screen) {
+    .is-responsive-web {
+        .bdl-CollapsibleSidebar-wrapper,
+        .bdl-CollapsibleSidebar.is-expanded,
+        .bdl-CollapsibleSidebar.is-expanded .bdl-CollapsibleSidebar-nav ul {
```
greathmaster, line 118: Is it possible to achieve the same result with less specificity?

src/features/collapsible-sidebar/CollapsibleSidebar.scss
```diff
@@ -10,7 +10,7 @@ $collaspsible-side-bar-logo-v-margin: $collapsible-side-bar-v-padding + $bdl-gri
 $left-sidebar-z-index: 40 !default;
 
 .bdl-CollapsibleSidebar-wrapper {
-    position: relative;
+    // position: relative;
```
tjuanitas, line 13: We should remove this if it's no longer needed.

mickr: Are we sure this should be removed? `.bdl-CollapsibleSidebar` is still positioned `absolute` so it may not be contained in the wrapper anymore.

tjuanitas: In that case, this should be uncommented, right? @joshmarnold can you add screenshots to this PR?

mickr: That may be the expected behavior but good to check 😄

joshmarnold: Hey guys, let me get back to you. I'm sort of back to the drawing board after I found out there's a scrollbar to manage.

joshmarnold:
Moving this back over to original ticket [WEBAPP-12529](https://jira.inside-box.net/browse/WEBAPP-12529)

src/features/collapsible-sidebar/CollapsibleSidebar.scss
```diff
@@ -111,6 +111,25 @@ $left-sidebar-z-index: 40 !default;
...
+    }
+
+    .is-responsive-web {
+        .bdl-CollapsibleSidebar.is-expanded {
+            width: 100%;
+
+            .bdl-CollapsibleSidebar-nav ul {
+                // On expand, set ul to fullscreen width with no transition for performance.
+                // Use vw because relative parent (custom-scrollbar) is not full screen width
+                width: calc(100vw - (#{$collapsible-side-bar-h-padding} * 2));
```
jcamden, line 122: Would it work to disable padding of the parents, and then use `width:100%`? This `width: calc(100vw` is clever, but seems fragile versus the relative `width:100%` approach.

src/features/collapsible-sidebar/CollapsibleSidebar.scss
jcamden: What did you find here?

src/features/collapsible-sidebar/CollapsibleSidebar.scss
```diff
@@ -111,6 +111,25 @@ $left-sidebar-z-index: 40 !default;
     }
 }
 
+@include breakpoint($small-screen) {
+    // JUST FOR TESTING ******** REMOVE BEFORE PUSHING ----
+    .bdl-CollapsibleSidebar-wrapper.hidden {
```
jcamden: Can you tell us what this is for and what happens when you remove it?

src/features/collapsible-sidebar/CollapsibleSidebar.scss
joshmarnold: The comment above the code explains why I went with this approach. I don't want it to be "clever" and figured that's how it would seem. 

To explain a little further, the relative parent of the ul element is the scrollbar element which has its own dimensions that can't be relied on for getting screen width. Therefore, without using javascript, the only way without modifying the scrollbar (which is probably not possible) is to use vw.

jcamden: I entered these changes manually in the browser:

.bdl-CollapsibleSidebar.is-expanded {
    width: 100%;
}

@media (max-width: 767px)
.bdl-CollapsibleSidebar-wrapper {
    width: 100%;
}

@media (max-width: 767px)
.bdl-CollapsibleSidebar.is-expanded .bdl-CollapsibleSidebar-nav ul {
    width: 100%;
}

And I got the following result:
Screen Shot 2022-05-04 at 5 04 01 PM
Screen Shot 2022-05-04 at 5 04 09 PM

Does that seem right?

jcamden: Oh it's actually not right, because you want small screen-width rather than medium. But anyway, if you just change the breakpoint that I put there from medium to small, will that do it for ya?

joshmarnold: Yeah, so there's a gotcha with this approach; The animation choppiness due to the ul contents. Setting the ul width to 100% will cause the ul and all of it's content (which is a lot) to "follow" the animation / transition which overloads the GPU I assume and causes choppiness.

src/features/collapsible-sidebar/CollapsibleSidebar.scss
joshmarnold: I put this in there to show the wrapper. Is there a ff or cookie on or something? Anyway, it was a quick fix to see the sidebar

src/features/collapsible-sidebar/CollapsibleSidebar.scss
joshmarnold: It does because `.bdl-CollapsibleSidebar-nav ul` has a lot of elements which needs to be resized / recalculated during the layout step in the browser render which is costly. This could possibly be fixed by redesigning the li elements into flex items and reducing the number of els in each li while still keeping accessibility in tact. But that's out of the scope of this ticket.

src/features/collapsible-sidebar/CollapsibleSidebar.scss
joshmarnold: Screen Shot 2022-05-06 at 9 44 54 AM

src/features/collapsible-sidebar/CollapsibleSidebar.scss
jcamden: Thanks for your great demo of this today. I now understand the problem, and I like your solution.

src/features/collapsible-sidebar/CollapsibleSidebar.scss
jcamden: Great write up. This answered all my questions. Just want to note that [it's best not to commit commented code](https://kentcdodds.com/blog/please-dont-commit-commented-out-code). I just learned that as part of the header review :) 

In terms of the solutions, solution 1 seems like simplest, but instead of commenting it out, can we modify the `position` style of `.bdl-CollapsibleSidebar-wrapper` for small screen-widths only and hide it behind .is-responsive-web?

joshmarnold: Yeah, that will work. Helps encapsulate these changes. Although, position on the wrapper element is not really necessary. But for more safety, lets put position: initial in the small-screen media query

jcamden:
You made a great observation that the `CollapsibleSidebar` menu item actions overlap the menu items themselves during the transition due to their absolute positioning. I think you also hit the nail on the head when you suggested converting those menu items to flexboxes. That would definitely solve that problem, and possibly many others. 

That work is very probably outside of the scope of this ticket, unless that conversion would affect that way that we are fullscreening this component on small screen-widths. It stands to reason that the flexboxes would suffer similar recalculation issues to width: 100%, but I don't actually know that. Do you happen to know that. or is perhaps something we can look into? If we still need to do a calc with vw even with flexbox `li`s (menu items), then let's just create a follow-up ticket to take care of the menu item actions overlap issue. Otherwise, I guess now would be the time :)

src/features/collapsible-sidebar/CollapsibleSidebar.scss
```diff
@@ -111,6 +111,25 @@ $left-sidebar-z-index: 40 !default;
...
+        .bdl-CollapsibleSidebar-wrapper {
+            position: initial;
+        }
+
+        .bdl-CollapsibleSidebar.is-expanded {
+            width: 100%;
+
+            .bdl-CollapsibleSidebar-nav ul {
+                // On expand, set ul to fullscreen width with no transition for performance.
+                // Use vw because relative parent (custom-scrollbar) is not full screen width
```
jcamden, line 125: I find this comment just a little confusing. Could we say something like, "calc a static width because dynamic calculations for width:100% during transitions created performance issues."

src/features/collapsible-sidebar/CollapsibleSidebar.scss
```diff
@@ -111,6 +111,25 @@ $left-sidebar-z-index: 40 !default;
...
+                width: calc(100vw - (#{$collapsible-side-bar-h-padding} * 2));
+                // width: 100%;
```
jcamden, line 127: The KCD commented code thing :)

joshmarnold:
I don't think looking that up is necessary for this task (not just ticket). Setting the ul to a fixed width works perfectly in my opinion and we don't need it to expand with the parent. 

Using flexbox is just pure speculation on my part. Maybe even a table would work? The main idea is that we try to use a browser optimized element (if that's what you call them or if that even exists?)

src/features/collapsible-sidebar/CollapsibleSidebar.scss
joshmarnold: Oops, I missed that. Pushing again

src/features/collapsible-sidebar/CollapsibleSidebar.scss
joshmarnold: Yeah, that works for me. Changed it to almost what you said

src/features/collapsible-sidebar/CollapsibleSidebar.scss
```diff
@@ -111,6 +111,24 @@ $left-sidebar-z-index: 40 !default;
     }
 }
 
+@include breakpoint($small-screen) {
+    .Body.is-responsive-web {
+        .bdl-CollapsibleSidebar-wrapper {
+            position: initial;
```
tjuanitas, line 117: I understand from this [thread](https://github.com/box/box-ui-elements/pull/2882#discussion_r865391641) there's choppiness if we set `width: 100%;` for `.bdl-CollapsibleSidebar-nav ul`.

But in combination with the new calc solution for `.bdl-CollapsibleSidebar-nav ul`, are we able to set `width: 100%` for the wrapper now?

Since `CollapsibleSidebar` is used in other apps such as Admin Console and Dev Console, it'd be nice if this change could make it responsive there as well. Thus, I don't think we can rely on the `position` of the parent element (in this case `in-app-messenger-provider`).

joshmarnold: Okay, thanks for mentioning this. Take a look at the latest commits. I ended up with a solution where the `wrapper` was able to keep its position=relative. All I had to do was set `.bdl-CollapsibleSidebar.is-expanded` to width=100vw.

joshmarnold:
@jcamden We might want to look into making the small-screen size to something larger. In chrome dev tools, set your screen to be an actual phone. I'm noticing that all of them have screen sizes larger than our small-screen size (374px).

jcamden:
> @jcamden We might want to look into making the small-screen size to something larger. In chrome dev tools, set your screen to be an actual phone. I'm noticing that all of them have screen sizes larger than our small-screen size (374px) so the mobile media queries will never be hit. Can you double check this?

That's a question for Design, namely Jared Comis in [#responsive-web](https://box.slack.com/archives/C01ED9N6UP6). I think it is very likely that he made these changes at the breakpoint which he intended; but I like your idea of double-checking the design with him.

src/features/collapsible-sidebar/CollapsibleSidebar.scss
```diff
@@ -111,6 +111,19 @@ $left-sidebar-z-index: 40 !default;
     }
 }
 
+@include breakpoint($small-screen) {
```
jcamden, line 114: Just a reminder, we gotta change this to `$medium-screen`---thanks to your observation that almost no devices are within our small screen-width breakpoint these days 👍

src/features/collapsible-sidebar/CollapsibleSidebar.scss
```diff
@@ -12,7 +12,7 @@ $left-sidebar-z-index: 40 !default;
 .bdl-CollapsibleSidebar-wrapper {
     position: relative;
     flex-shrink: 0;
-    width: $collapsible-side-bar-width;
+    width: $collapsible-side-bar-collapsed-width;
```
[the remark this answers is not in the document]
joshmarnold: No longer relevant due to recent commits. Thanks

src/features/collapsible-sidebar/CollapsibleSidebar.scss
```diff
@@ -12,7 +12,7 @@ $left-sidebar-z-index: 40 !default;
 .bdl-CollapsibleSidebar-wrapper {
     position: relative;
     flex-shrink: 0;
-    width: $collapsible-side-bar-collapsed-width;
+    width: $collapsible-side-bar-width;
```
joshmarnold, line 15: Leaving this the same as current master

src/features/collapsible-sidebar/CollapsibleSidebar.scss
```diff
@@ -23,7 +23,7 @@ $left-sidebar-z-index: 40 !default;
     z-index: $left-sidebar-z-index;
     display: flex;
     flex-direction: column;
-    width: 100%;
+    width: $collapsible-side-bar-collapsed-width;
```
joshmarnold, line 26: Leaving this the same as current master

src/features/collapsible-sidebar/CollapsibleSidebar.scss
```diff
@@ -111,16 +111,9 @@ $left-sidebar-z-index: 40 !default;
...
-    .Body.is-responsive-web {
-        .bdl-CollapsibleSidebar.is-expanded {
-            width: 100vw;
-
-            .bdl-CollapsibleSidebar-nav ul {
-                // using fixed width due to performance issues during expand transition
-                width: calc(100vw - (#{$collapsible-side-bar-h-padding} * 2));
-            }
-        }
+@include large-size {
```
joshmarnold, line 114: I sorted media queries by width precedence since our current method of targeting sizes uses max-width instead of expact widths like 200-400. It's my opinion that we should correct how widths are targeted in the media-queries file so that we don't have tons of queries in the document and need to worry about order precedence

src/features/collapsible-sidebar/CollapsibleSidebar.scss
```diff
@@ -133,8 +139,34 @@ $left-sidebar-z-index: 40 !default;
...
+            display: none;
+        }
+
+        .bdl-CollapsibleSidebar-nav ul {
+            // using fixed width due to performance issues during expand transition
+            width: calc(100vw - (#{$collapsible-side-bar-h-padding} * 2));
+        }
+
+        .bdl-CollapsibleSidebar {
+            border: none;
```
jcamden, line 160: Does this mean that the sidebar doesn't have a border even when it is expanded?

src/features/collapsible-sidebar/CollapsibleSidebar.scss
jcamden: Are you're saying that if we have multiple active media queries (due to a mix of old breakpoints and new ones) whichever is smaller should win? If that is the case, I don't know the answer to that---whether the precedence should be based on size, or whether the newer stuff should always override, or whether it would have to be a case-by-case thing.

jcamden: Ah, do the old ones use `max-width`?

jcamden: The old `large-size` looks like this:
```
@mixin large-size {
    @media (min-width: #{$medium-width + 1px}) {
        @content;
    }
}
```
Seems like a very confusing situation with some old media queries using `min-width` and new ones using `max-width`.

joshmarnold: In these cases, it will take by order (whoever comes last)

src/features/collapsible-sidebar/CollapsibleSidebar.scss
joshmarnold: For mobile screens only, yeah. No need for a border at all on the sidebar when on mobile because it's either 0 width or 100% screen width.

src/features/collapsible-sidebar/CollapsibleSidebar.scss
```diff
@@ -111,6 +111,12 @@ $left-sidebar-z-index: 40 !default;
     }
 }
 
+@include large-size {
```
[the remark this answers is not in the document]
joshmarnold: [I commented my reasoning as you were typing this, ha. Going to resolve this one. ](https://github.com/box/box-ui-elements/pull/2882#discussion_r870293191)

src/features/collapsible-sidebar/CollapsibleSidebar.scss
jcamden: I am also thinking about non-overlapping styles. On the one hand, sometimes you may want styles to "cascade" inclusively across breakpoints (so like if large specifies block and blue, and small specifies red, then small turns out block and red). This can get pretty confusing, imo, when you have both min-width and max-width, styles bubbling up and down, as it were. The fact that you can't inherit styles from other media queries might be seen as a drawback to exact media queries---but personally, I would rather explicitly set the styles twice than be super confused about what is getting inherited from where.

src/features/collapsible-sidebar/CollapsibleSidebar.scss
jcamden: That is a great point.

src/features/collapsible-sidebar/CollapsibleSidebar.scss
```diff
@@ -133,8 +139,34 @@ $left-sidebar-z-index: 40 !default;
...
-        margin-left: $collapsible-side-bar-h-padding + $bdl-grid-unit * 2 - 1px /* border */;
+@include breakpoint($medium-screen) {
+    .Body.is-responsive-web {
+        .bdl-CollapsibleSidebar-wrapper,
+        .bdl-CollapsibleSidebar {
+            width: 0;
+        }
+
+        .bdl-CollapsibleSidebar-footer,
+        .bdl-CollapsibleSidebar-nav ul {
```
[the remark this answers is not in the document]
jcamden: Let's look around our scss files and see if a precedent exists.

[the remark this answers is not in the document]
jcamden: I was just looking at ContentExplorer.scss, and it looks like lot of classes use `display:flex`, but they are not grouped together for that style. That's one example for us to think about.

jcamden (dismissed):

jcamden (dismissed):

src/features/collapsible-sidebar/CollapsibleSidebar.js
```diff
@@ -92,11 +92,10 @@ class CollapsibleSidebar extends React.Component<Props> {
             'bdl-CollapsibleSidebar',
             className,
         );
-        const wrapperClasses = classNames('bdl-CollapsibleSidebar-wrapper', { hidden: isHidden });
```
tjuanitas, line 95: What's the reason for removing `isHidden`?

@jcamden Are the changes in this PR still needed? https://github.com/box/box-ui-elements/pull/2844
If not, we should probably remove the prop, `aria-hidden`, and the tests associated with `isHidden`.

joshmarnold: isHidden was being used before to show or not show the sidebar which led to issues because we need the sidebar to be controlled by css-only since it has transitions. 

@jcamden Do we need this aria-hidden attribute on the wrapper element?

jcamden: Hey @tjuanitas , thanks for commenting! To put a finer point on what Josh said above, unless I misunderstand, we cannot show or hide CollapsibleSidebar via css-only. Previously, we used JS to conditionally add a `hidden` class to the sidebar to set `display:none`. Josh's approach leverages the existing `isExpanded` class (which is, of course, conditionally added via JS) to set the width of the sidebar, fixed width by default, and `100vw` on mobile screen widths. The width of the sidebar is also set to `0vw` by default on mobile screen widths, so that it toggles between hidden and fullscreen, without the use of `width: 100%` or `display:none`, since both of those styles interfered with transitions.

The implications of this are interesting. It turns out that I didn't need to add `aria-hidden={true}` for `display:none`. Evidently, it is set by default. Funny enough, now that we are using `0vw`, I believe we _do_ need to manually set `aria-hidden={true}`. @tjuanitas , you are 100% correct that the we need to make changes to the tests that depend on `isHidden`. Thank you so much for catching that!

src/features/collapsible-sidebar/CollapsibleSidebar.scss
```diff
@@ -133,8 +139,34 @@ $left-sidebar-z-index: 40 !default;
     }
 }
 
-@include large-size {
-    .bdl-CollapsibleSidebar-logo {
-        margin-left: $collapsible-side-bar-h-padding + $bdl-grid-unit * 2 - 1px /* border */;
+@include breakpoint($medium-screen) {
+    .Body.is-responsive-web {
```
tjuanitas, line 143: Just checking, do we intend to remove `.is-responsive-web` in the future?

Also, is `.Body` necessary? Could it be removed for less specificity?

src/features/collapsible-sidebar/CollapsibleSidebar.js
tjuanitas: Gotcha. So then the `isHidden` prop is now solely used for setting `aria-hidden`, correct? I think the tests may be fine at the moment but let's update the prop comment:
`/** Optionally apply hidden class (__visibility.scss) and "aria-hidden": "true" to CollapsibleSidebar wrapper */`

src/features/collapsible-sidebar/CollapsibleSidebar.scss
joshmarnold: @tjuanitas I also followed convention with the `.Body.is-responsive-web`

At some point we’ll want to rollout responsiveness as default. So the .body.is-responsive-web is not really meant to stay forever. We would clean it up like the .body.deja-blue at some point.

src/features/collapsible-sidebar/CollapsibleSidebar.js
joshmarnold: @tjuanitas We don't use the visibility.scss file here. The sidebar is shown with a transition via its width which is either 0 or "full screen". 

In that case, the comment should be updated to 
`/** Optionally apply "aria-hidden": "true" to CollapsibleSidebar wrapper */`

tjuanitas: Yeah sorry I meant let's update [this line](https://github.com/box/box-ui-elements/blob/f722e7d59a7522ade6ec2bb0e80a8847bffd3028/src/features/collapsible-sidebar/CollapsibleSidebar.js#L40) to reflect the new usage of the prop. The wording you used is good with me.

tjuanitas: Out of curiosity, is there a way we could set `aria-hidden` automatically (and reliably) based on detecting the screen size and using the `expanded` prop? That way we don't need to use `isHidden` since I imagine a lot of developers would forget to set this prop.

joshmarnold: @tjuanitas That's what's happening upstream. 

`const isHidden = (size === VIEW_SIZE_TYPE.small || size === VIEW_SIZE_TYPE.medium) && !isLeftNavForcedOpen;`

joshmarnold: This is possible with event listeners and lifecycle methods. But that would be out of the scope of this ticket

tjuanitas (approved):

src/features/collapsible-sidebar/CollapsibleSidebar.scss
```diff
@@ -133,8 +139,34 @@ $left-sidebar-z-index: 40 !default;
...
+        }
+
+        .bdl-CollapsibleSidebar-footer {
+            display: none;
+        }
+
+        .bdl-CollapsibleSidebar-nav ul {
+            display: none;
+            // using fixed width due to performance issues during expand transition
+            width: calc(100vw - (#{$collapsible-side-bar-h-padding} * 2));
```
jcamden, line 155: I know you didn't come up with them, but these variable names could be better. `Sidebar` should be `sidebar` instead of `side-bar` and something like `padding-x` and `padding-y` seems much more clear than `h-padding` and `v-padding`. I actually didn't know what `h-padding` was without the context of `v-padding`---maybe because it is 2:50AM, or maybe because it's not a great name.

jcamden (approved):

src/features/collapsible-sidebar/CollapsibleSidebar.scss
tjuanitas: Agreed. Also looks like one of the variable names has a typo: $colla**s**psible-side-bar-logo-v-margin. We can address this in a future PR.

tjuanitas:
Force merging this Semantic Pull Request is down and passes other checks.

PR #2884, opened 2022-05-06
fix(modal): Relocated action buttons to the bottom on small screens

msorianoBox:
Hi guys! In order to have the modals to match the look of the mockups, the button action bar had to be relocated to the bottom of the screen, and css adjustments have been made for small screen sizes only.

Screen Shot 2022-05-06 at 11 47 27


src/components/modal/Modal.scss
```diff
@@ -153,6 +153,15 @@
         .modal-close-button {
             right: 10px;
         }
+
+        .modal-actions {
+            position: fixed;
```
jstoffan, line 158: It doesn't seem ideal to have nested fixed position elements. Can we utilize flexbox or relative/absolute positioning, instead?

jstoffan (changes requested):

src/components/modal/Modal.scss
```diff
@@ -153,6 +153,15 @@
...
+            right: 0;
+            bottom: 0;
+            width: 100%;
+            padding: 10px 16px;
+            background-color: #fff;
```
tjuanitas, line 163:
```suggestion
            background-color: $white;
```

src/components/modal/Modal.scss
```diff
@@ -153,6 +153,15 @@
         .modal-close-button {
             right: 10px;
         }
+
+        .modal-actions {
+            position: fixed;
+            right: 0;
+            bottom: 0;
+            width: 100%;
+            padding: 10px 16px;
```
tjuanitas, line 162: Could use `$bdl-grid-unit` for these values.

src/components/modal/Modal.scss
msorianoBox: Hi Jared! I've been looking for an alternate way to do this trying not u sed the position: fixed, but been unable to come with a simple solution like this one. I know the fixed button container is nested and that might not seem ideal, but this would only be for the modal in small screens. If you have some other options I'll appreciate it.

tjuanitas: @msorianoBox  settting `.modal-content` to `flex-grow: 1; position: relative;` should do the trick for the container part^

jstoffan: @msorianoBox and I went over this on a call. The main issue is that `.modal-actions` can be nested to an arbitrary depth within `.modal-content`, so we can't rely on flexbox or relative positioning at the component level. Apart from refactoring Modal's layout styles, I'm not sure there's a better option than nested fixed positions right now.

jstoffan (approved):
@msorianoBox, I'm approving to unblock but we should address @tjuanitas's comments before merging.

src/components/modal/Modal.scss
tjuanitas: Ah yeah you're right.

Question about using `position: fixed`:
For a modal that has content (`.modal-content`) that expands the entire height of the screen, wouldn't a fixed position for `.modal-actions` block/cover the bottom of that content? I think we might need to add a padding to `.modal-content` that takes into account the height of `.modal-actions`.

msorianoBox: @tjuanitas Done :D

src/components/modal/Modal.scss
```diff
@@ -144,6 +144,10 @@
             padding: 0;
         }
 
+        .modal-content {
+            padding-bottom: 10 * $bdl-grid-unit;
+        }
```
tjuanitas, line 149: I'm not sure if this is avoidable without JS but this will add a padding even if there's no `.modal-actions` component. E.g. a modal that has content greater than the height but does not have action buttons will have extra bottom padding. Do we need to check with design if that's alright?

tjuanitas: Also should this be `padding-bottom: 15 * $bdl-grid-unit;`? I think the `.modal-actions` with the new padding is around 60px right?

msorianoBox: We can check with design, but this extra padding doesn't affect the modal display since the only case that content can scroll is the Move/Copy modal and then it's necessary. In any other case, the padding is not enough to make content scroll.

msorianoBox: The button bar height is 40px :D

jstoffan: Given that the rest of the styles don't conform to the grid, I would suggest hardcoding the pixel values to stay consistent.

tjuanitas: @jstoffan Sounds good.

@msorianoBox We also have 10px padding above and below it I think we'll need to account for. Otherwise there will still be some content that's cut off.

Screen Shot 2022-05-19 at 3 05 37 PM

src/components/modal/Modal.scss
```diff
@@ -153,6 +157,15 @@
         .modal-close-button {
             right: 10px;
         }
+
+        .modal-actions {
+            position: fixed;
+            right: 0;
+            bottom: 0;
+            width: 100%;
+            padding: 2.5 * $bdl-grid-unit 4 * $bdl-grid-unit;
```
jstoffan, line 166: Let's use the numeric values here, too, since they're basically copied from other styles in this file.

src/components/modal/Modal.scss
```diff
@@ -144,6 +144,10 @@
             padding: 0;
         }
 
+        .modal-content {
+            padding-bottom: 60px;
```
tjuanitas, line 148: @msorianoBox Sorry about this! I think my math/process was incomplete and it should be `78px`?

The buttons are 40px with 10px on the top and 10px on the bottom = 60px total. So the height of `modal-actions` that's located at the bottom of the screen is 60px.

But since we have `margin-top: 30px` on `modal-actions` in the normal view (the non-responsive view and I assume we want to include this similar margin/design), the end of the content would be `90px` from the bottom of the screen.

We would add this padding to `modal-content` so that there's no content behind the `modal-actions` bar and there's enough white space between the buttons and the content. But we also need to take into account the bottom padding that's already included for `modal-dialog`. So that would be 90px minus 12px = 78px.

Would love a second opinion on this though. It's also very possible that I'm overthinking this :sweat_smile:.

tjuanitas: Also happy with just leaving this as-is.

msorianoBox: In fact, I was testing with some overflowing content and 40 px worked perfectly to display all of the content without cropping. I went for the 60 px, since you requested and even with the extra white space looked good. I really think that there's no need for anymore extra margin, but if you want, I'll add it :D

tjuanitas (approved):

PR #2890, opened 2022-05-13
fix(header-flyout): default resp behavior should be under resp class

ivanthai:
* fixed an issue where some responsive behavior would come through even if it was disabled in props

### DIFF
--- a/src/features/header-flyout/HeaderFlyout.js
+++ b/src/features/header-flyout/HeaderFlyout.js
@@ -18,8 +18,6 @@ type Props = FlyoutProps & {
     footer?: React.Element<any>,
     /** What content to display in the header */
     header?: React.Element<any>,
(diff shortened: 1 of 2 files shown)


src/features/header-flyout/HeaderFlyout.js
```diff
@@ -18,8 +18,6 @@ type Props = FlyoutProps & {
     footer?: React.Element<any>,
     /** What content to display in the header */
     header?: React.Element<any>,
-    /** Enable responsive behaviors of Flyout */
-    isResponsive?: boolean,
```
ivanthai, line 22: Removed this here because the prop is getting passed through via `{...rest}`

tjuanitas (approved):
lgtm!

tjuanitas:
Force merging since Semantic Pull Request is down and passes other checks.

PR #2896, opened 2022-05-17
chore(action): Replaced semantic PR with an action version

greg-in-a-box:
Replaced existing Semantic Pull Request GH app config with Semantic Pull Request GH action config

ref: https://github.com/amannn/action-semantic-pull-request


.github/workflows/semantic.yml
```diff
@@ -0,0 +1,17 @@
...
+            - opened
+            - edited
+            - synchronize
+
+jobs:
+    main:
+        name: Validate
+        runs-on: ubuntu-latest
+        steps:
+            - uses: amannn/action-semantic-pull-request@v4
```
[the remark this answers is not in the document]
greg-in-a-box: Good question, Is there a process for that ?

tjuanitas (approved):
LGTM. Do we have to set up any sort of configuration settings?

.github/semantic.yml
```diff
@@ -1,2 +0,0 @@
-titleAndCommits: true
-anyCommit: true
```
tjuanitas, line 2: Do we need to add these settings for the GH Action?

greg-in-a-box: Not that I am aware of, should work out of the box with the configuration I have in there

PR #2897, opened 2022-05-17
feat(collapsible-sidebar): Move mobile SB down and hide right side header content on expand

joshmarnold:


src/features/collapsible-sidebar/CollapsibleSidebar.scss
```diff
@@ -143,6 +143,11 @@ $left-sidebar-z-index: 40 !default;
     .Body.is-responsive-web {
         .bdl-CollapsibleSidebar-wrapper {
             width: 0;
+            margin-top: 60px;
```
jcamden, line 146: Apparently, if the number of pixels for any spacing related style is divisible by four, we have to use `bdlGridUnit`---in this case, `bdlGridUnit * 15`

jcamden (approved):

src/features/collapsible-sidebar/CollapsibleSidebar.scss
joshmarnold: Decided this isn't necessary

jcamden, line 146: Encapsulation :) We don't want CollapsibleSidebar to be concerned with its external layout. We should apply the margin in EUA for sure.

jcamden (changes requested):

PR #2899, opened 2022-05-18
refactor(typescript): Migrate Datalistitem to typescript

greg-in-a-box:


src/components/datalist-item/__tests__/DatalistItem.test.tsx
```diff
@@ -1,66 +1,44 @@
 import React from 'react';
 import sinon from 'sinon';
-
+import { shallow } from 'enzyme';
 import DatalistItem from '..';
 
 const sandbox = sinon.sandbox.create();
-
```
tjuanitas, line 7: All the newlines in this file are removed. I think we might want to keep these to separate tests / logical groupings of code and help readability.

tjuanitas:
Does this file need to be updated as well?
https://github.com/greg-in-a-box/box-ui-elements/blob/master/scripts/styleguide.config.js

src/components/datalist-item/__tests__/DatalistItem.test.tsx
```diff
@@ -1,66 +1,44 @@
 import React from 'react';
 import sinon from 'sinon';
-
+import { shallow } from 'enzyme';
```
tjuanitas, line 3: This is weird, did the tests even work before when this wasn't imported?

src/components/datalist-item/__tests__/DatalistItem.test.tsx
```diff
@@ -1,66 +1,44 @@
...
-
     describe('render()', () => {
         test('should render default component', () => {
             const child = 'Test';
             const wrapper = shallow(<DatalistItem>{child}</DatalistItem>);
-
             expect(wrapper.is('li')).toBe(true);
             expect(wrapper.hasClass('datalist-item')).toBe(true);
-            expect(wrapper.prop('id')).toBeTruthy();
+            expect(wrapper.prop('id')).toMatch(new RegExp('^datalistitem?'));
```
tjuanitas, line 17: I don't know how often we use regular expressions but I think we'd prefer to use the literal version here.
```suggestion
            expect(wrapper.prop('id')).toMatch(/^datalistitem?/);
```

src/components/datalist-item/__tests__/DatalistItem.test.tsx
greg-in-a-box: No idea, my IDE(intellij) flagged it as missing, I dont know how this was working before.

src/components/contact-datalist-item/ContactDatalistItem.js
```diff
@@ -2,7 +2,7 @@
 import * as React from 'react';
 
 import Avatar from '../avatar';
-import DatalistItem from '../datalist-item';
```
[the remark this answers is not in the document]
greg-in-a-box: not sure why those were changed but ill change them back

src/components/datalist-item/__tests__/DatalistItem.test.tsx
```diff
@@ -1,66 +1,44 @@
...
-            const role = 'listitem';
-            const wrapper = shallow(
-                <DatalistItem id={id} role={role}>
-                    Test
-                </DatalistItem>,
-            );
-
-            expect(wrapper.prop('id')).not.toEqual(id);
-            expect(wrapper.prop('role')).toEqual('option');
-        });
```
tjuanitas, line 61: Are we sure we don't need this?

I think it's testing the order of the props:
`<li {...itemProps} className={classes} id={this.id} role="option" aria-selected={isSelected}>`
so that `itemProps` is not placed after `id={this.id} role="option"`.

src/components/datalist-item/DatalistItem.tsx
```diff
@@ -0,0 +1,75 @@
...
+            'datalist-item',
+            {
+                'is-active': isActive,
+            },
+            className,
+        );
+        const itemProps = omit(rest, ['closeDropdown', 'setActiveItemID']);
+        // required aria props are added dynamically
+
+        /* eslint-disable jsx-a11y/role-has-required-aria-props */
```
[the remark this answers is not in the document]
greg-in-a-box: That comment should have been probably inside of the eslint ignores.  When it was converted it probably didnt detect a line of code below it so it moved it up

src/components/datalist-item/__tests__/DatalistItem.test.tsx
greg-in-a-box: TS should prevent a dev from setting fields that arent in the props interface, also they somehow get into itemProps, they get overwritten since its being set after the spread.

greg-in-a-box: Also when I tried to leave it in there, I was getting TS errors that I was trying to set fields that werent in the interface which was why I had to remove them and decided to combine the assertions with the render test at the top

tjuanitas (approved):

PR #2900, opened 2022-05-19
fix(annotations): annotations pagination loading issue

karelee7:
Fix incorrect query param for annotations pagination

**Example of annotations flow with correct query parameters:**
initial-call
pa
<img width=
ge-2-call" src="https://user-images.githubusercontent.com/95440409/169380712-713d865a-2303-47b2-af61-c2a9796a6fea.png">
last-call

### DIFF
--- a/src/api/MarkerBasedAPI.js
+++ b/src/api/MarkerBasedAPI.js
@@ -9,7 +9,7 @@ import type { ElementsErrorCallback } from '../common/types/api';
 
 type Params = {
     limit: number,
-    marker: string,
+    next_marker?: string,
 };
 
 type Data = {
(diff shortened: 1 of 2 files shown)


src/api/MarkerBasedAPI.js
```diff
@@ -9,7 +9,7 @@ import type { ElementsErrorCallback } from '../common/types/api';
 
 type Params = {
     limit: number,
-    marker: string,
+    next_marker?: string,
```
jstoffan, line 12: This base class is used by several different subclasses, so this would likely cause other issues. Is `marker` the typical convention for our other services? If so, it seems like the underlying service need to be updated to use that, instead?

jstoffan (changes requested):

src/api/MarkerBasedAPI.js
karelee7: Ohh I see, hm..not sure if it's the typical convention for other services, but for annotations, it should be `next_marker` instead of `marker` (also, for annotations, the `next_marker` param should only be passed if there are annotations over the limit)

jstoffan: I understand. Can we update the service to follow the [standard](https://developer.box.com/guides/api-calls/pagination/marker-based/), since pagination is not currently working and the endpoint is not yet public?

PR #2903, opened 2022-05-23
fix(flyout): unrevert responsive flyout changes and fix safari UI issue

ivanthai:
* Un-reverted the revert commit d0dae8adee7a5bbc00ff57646a4de1e733f0c900 which reverted all Flyout changes made to add responsive behaviors
* Fixed an issue that would cause a visible UI issue on Safari. 

The meaningful and unreviewed change is the second commit to this PR: https://github.com/box/box-ui-elements/pull/2903/commits/b67c0869ebbcc82f4026dbf24311837a03541043

**Observed issue on Safari**
Screen Shot 2022-05-23 at 3 47 26 PM

Basically, the issue was caused by two main things...
1) TetherComponent second child's visibility state must be handled by the component state. I added the FlyoutContext outside of the isVisible check which would instead cause the Overlay (second child of TetherComponent) to render - see screenshot below. 
2) Normally the first cause would be OK because TetherComponent renders the content off the screen...however the HelpMenu in EUA overrode the CSS position = absolute which directly conflicted with TetherComponent. Instead of rendering off screen, an opaque CSS transform would be applied and would be visible. I believe the reason why we are seeing this in Safari only is because of a difference of how they render items with CSS transform. I'm not exactly familiar with the specific issue with their browser (if it is an issue)

**Cause**
Screen Shot 2022-05-23 at 3 44 49 PM
Screen Shot 2022-05-23 at 3 48 27 PM

**Expected**
Correction: Overlay elements should and no longer appear by default anymore
Screen Shot 2022-05-23 at 3 54 32 PM
Screen Shot 2022-05-23 at 3 56 29 PM

**Fix**
Original
```
// Flyout.js - render()
<TetherComponent {...tetherProps}>
    {React.cloneElement(overlayButton, overlayButtonProps)}
    {React.cloneElement(overlayContent, overlayProps)}
</TetherComponent>
```
Bad Changes
```
<TetherComponent {...tetherProps}>
    {React.cloneElement(overlayButton, overlayButtonProps)}
    <FlyoutContext.Provider value={{ closeOverlay: this.closeOverlay }}> // not controlled by isVisible state and will always render the CSS transform issue
        {isVisible ? React.cloneElement(overlayContent, overlayProps) : null}
    </FlyoutContext.Provider>
</TetherComponent>
```
Good changes
```
<TetherComponent {...tetherProps}>
    {React.cloneElement(overlayButton, overlayButtonProps)}
    {isVisible && (
        <FlyoutContext.Provider value={{ closeOverlay: this.closeOverlay }}>
            {React.cloneElement(overlayContent, overlayProps)}
        </FlyoutContext.Provider>
    )}
</TetherComponent>
```


mickr:
Thank you for the detailed description of the issue and fix

tjuanitas (approved):

PR #2904, opened 2022-05-24
feat(tooltip, radar-animation, targeting): add hasLowVisualPriority prop…

kkuliczkowski-box:
… for Tooltip and RadarAnimation; add shouldAct param to useOnclickBody hook

### Tooltip, RadarAnimation
`Tooltip` and `RadarAnimation` components had a constant high `z-index`, which resulted in displaying those elements above other "high level" elements such as `Modal` or `DropdownMenu`. This caused an unwanted situation where user would open a Modal and a Tooltip not associated to it would still be visible and obstructing the contents of the Modal.

Proposed solution is a new prop called `hasLowVisualPriority`. Setting it to `true` indicates that this particular instance of Tooltip (or RadarAnimation) has low visual priority and should not be displayed over Modal and Menu elements.
By default this prop is set to `false` to not break any existing user experience.

### useOnclickBody
This hook was used to detect clicks outside of a particular element but had a flaw by adding event listener with `useCapture` param not set to `true`. This caused this hook to not work as intended when a clicked element stopped event propagation.

Proposed solution is to set `useCapture` property to `true` when adding event listener and add additional hook parameter called `shouldAct`. This is a function that is executed every time a click is detected by the listener and passes event to the hook caller. It can then be validated (based on the passed event) whether to execute the function set as `onClick` parameter.


src/styles/variables.json
```diff
@@ -1,110 +1,124 @@
 {
-    "white": "#fff",
```
megansmith-box, line 2: Did anything in this file change? Can we avoid this large diff?

kkuliczkowski-box: Yes, I've added 2 variables in [src/styles/constants/_layout.scss](https://github.com/box/box-ui-elements/pull/2904/files#diff-a53963adad1686c603addefb455563c7244331db4bef3dd85399c29401105cea) and I believe that "variables" files are auto generated. I haven't made any changes to them manually.

PR #2909, opened 2022-05-26
feat(radar-animation): add tetherElementClassName prop

kkuliczkowski-box:
Added `tetherElementClassName` prop to `RadarAnimation` component so that it's possible to add a custom class name to the tether element component, the same way it is now implemented in `Tooltip` component.

### DIFF
--- a/src/components/radar/RadarAnimation.tsx
+++ b/src/components/radar/RadarAnimation.tsx
@@ -70,6 +70,8 @@ export interface RadarAnimationProps {
     offset?: string;
     /** Where to position the radar relative to the wrapped component */
     position: RadarAnimationPosition;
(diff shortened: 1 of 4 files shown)


src/components/radar/RadarAnimation.tsx
```diff
@@ -70,6 +70,8 @@ export interface RadarAnimationProps {
     offset?: string;
     /** Where to position the radar relative to the wrapped component */
     position: RadarAnimationPosition;
+    /** A CSS class for the tether element component */
+    tetherElementClassName?: string;
```
megansmith-box, line 74: I believe we should add this new prop to `components/radar/RadarAnimation.js.flow` as well

tjuanitas (approved):

PR #2910, opened 2022-05-26
fix(modal): Adjusted modal close button markup and styles

msorianoBox:
Adjusted the Markup so the close button is on the same div as the modal title and is aligned on small screens.


src/components/modal/ModalDialog.js
```diff
@@ -116,8 +116,8 @@ class ModalDialog extends React.Component<Props> {
                     <h2 className="modal-title" id={`${this.modalID}-label`}>
                         {title}
                     </h2>
+                    <>{this.renderCloseButton()}</>
```
tjuanitas, line 119: I think we can remove the fragment component.

src/components/modal/ModalDialog.js
```diff
@@ -116,8 +116,8 @@ class ModalDialog extends React.Component<Props> {
...
                 </div>
-                {this.renderCloseButton()}
```
tjuanitas, line 120: What's the reason for moving this?

src/components/modal/Modal.scss
```diff
@@ -151,7 +151,9 @@
         }
 
         .modal-close-button {
+            top: initial;
```
tjuanitas, line 154: minor but `initial` isn't supported in IE11. Although I'm not sure if we still care to support it since this is for responsive design and IE11 is near EOL.

tjuanitas (commented):
Can you add screenshots for the before and after changes? If we modify the `top` property, are we able to avoid making changes to the `line-height` and `{this.renderCloseButton()}`?

src/components/modal/ModalDialog.js
msorianoBox: The Fragment is there so it can render inside the title Div, otherwise the build breaks.

src/components/modal/ModalDialog.js
msorianoBox: So we can unify the header title and the close button in the same div. Currently the Close button was located outside the header along with the rest of the modal content.

tjuanitas: Gotcha, wasn't sure if this was correct semantically but I think it should be fine.

However, it looks like there are some `modal-header` usages in EUA that have `display: none`. Moving the close button inside the header will now hide the button for these modals (e.g. AnnotationsEngagementModal.scss, PaywallUpsellModal.scss).

I think we'll have to keep this on the outside so those modals still have a close button but no header. Otherwise, we'll have to update the modal stylesheets in the parent apps (EUA, AC, Dev Console).

src/components/modal/ModalDialog.js
tjuanitas: I think omitting it should be fine since the method is returning a jsx button. This is how we currently have it on `master`:
Screen Shot 2022-05-27 at 12 41 48 PM

src/components/modal/Modal.scss
tjuanitas: I think `top: auto` would be a good replacement as that should be the initial value in this case.

src/components/modal/Modal.scss
```diff
@@ -155,7 +155,9 @@
...
             right: 10px;
+            line-height: 0;
```
tjuanitas, line 160: Out of curiosity, are you testing this property on a specific modal? For general modals, I'm not sure if this would have an effect.

src/components/modal/Modal.scss
msorianoBox: Done

src/components/modal/ModalDialog.js
msorianoBox: Done

src/components/modal/Modal.scss
```diff
@@ -144,8 +144,23 @@
             padding: 0;
         }
 
+        .modal-header-container {
+            position: fixed;
```
tjuanitas, line 148: Since `.modal-dialog` has `position: relative;`, could we utilize that and use `absolute` positioning?

Not using `fixed` would be ideal if possible.

src/components/modal/Modal.scss
```diff
@@ -144,8 +144,23 @@
...
+            top: 0;
+            left: 0;
+            z-index: 2;
```
tjuanitas, line 151: Is this `z-index` necessary? If we use `absolute` positioning, can this be removed?

src/components/modal/Modal.scss
```diff
@@ -144,8 +144,23 @@
...
+            display: flex;
```
tjuanitas, line 152: Since `.modal-header-container` only has one child, do we need this style?

src/components/modal/ModalDialog.js
```diff
@@ -112,12 +112,14 @@ class ModalDialog extends React.Component<Props> {
...
-                    <h2 className="modal-title" id={`${this.modalID}-label`}>
-                        {title}
-                    </h2>
+                <div className="modal-header-container">
+                    <div className="modal-header">
+                        <h2 className="modal-title" id={`${this.modalID}-label`}>
+                            {title}
+                        </h2>
+                    </div>
+                    {this.renderCloseButton()}
```
tjuanitas, line 121: Since the close button has `absolute` positioning, it seems odd to put this inside of an element that also has `fixed` / `absolute` positioning since the two elements would not have an effect on each other.

**However**, if the purpose of moving this close button inside the `modal-header-container` is to vertically align the button with the title, then this change does make sense.

In that case, I think it would be better to change [this line](https://github.com/box/box-ui-elements/blob/173fb3328488e2001f99ef16fcaa0ca001459025/src/components/modal/Modal.scss#L175) to `height: 100%;`.

tjuanitas (commented):
What's the desired design if there's no header but there's a close button?

src/components/modal/Modal.scss
tjuanitas: If we're using the `display: flex` to vertically align the `.modal-header`, we should add `align-items: center` and remove the top padding for `.modal-header`.

This will make it more maintainable than keeping the padding of `.modal-header` in sync with the height of `.modal-header-container`.

src/components/modal/ModalDialog.js
greathmaster: This was the intent, to align the close button and the title. I've updated it to `height: 100%`.

src/components/modal/Modal.scss
greathmaster: Added `align-items: center` to `.modal-header-container` and also removed top padding from `.modal-header`.

src/components/modal/Modal.scss
greathmaster: Managed to use `position: absolute` remove the z-index after adjusting some of the other styles. 

Basically, the `.modal-content` effectively has a `margin-top: 64px` and `margin-bottom: 60px` which correspond to the height of the `.modal-header-container` and the `.modal-actions` footer respectively.

src/components/modal/Modal.scss
greathmaster: Found a way to use `position: absolute` instead.

greathmaster:
The work done on this PR was incorporated into this PR:

https://github.com/box/box-ui-elements/pull/2946

Closing this PR as the above was already merged.

PR #2911, opened 2022-05-27
refactor(typescript): Migrated ContactDataListItem to typescript

greg-in-a-box:


src/components/contact-datalist-item/ContactDatalistItem.tsx
```diff
@@ -0,0 +1,99 @@
...
+    avatarUrl: string | null | undefined;
+}
+
+class ContactDatalistItem extends React.PureComponent<ContactDatalistItemProps, ContactDatalistItemState> {
+    constructor(props: ContactDatalistItemProps) {
+        super(props);
+        this.state = { avatarUrl: undefined };
+    }
+
+    isMounted = false;
```
tjuanitas, line 30: Do we need to specify a type boolean?

src/components/contact-datalist-item/ContactDatalistItem.tsx
```diff
@@ -0,0 +1,99 @@
+import * as React from 'react';
+import Avatar from '../avatar';
+import DatalistItem from '../datalist-item';
+
+import { ContactType } from '../../features/unified-share-modal/types';
+import { SuggestedPillType } from '../pill-selector-dropdown/types';
+
+import './ContactDatalistItem.scss';
+
+interface ContactDatalistItemProps {
```
tjuanitas, line 10: I think we need to add `export` here.

src/components/contact-datalist-item/ContactDatalistItem.tsx
tjuanitas: Looking at other components, it looks like we don't really specify the type for properties of a class. But is that weird not to?

.eslintrc.js
```diff
@@ -62,6 +63,7 @@ module.exports = {
                 'camelcase': 'error',
                 'no-shadow': 'off',
                 '@typescript-eslint/no-shadow': ['error'],
+                "@typescript-eslint/no-unused-expressions": 'error', // moves eslint no-unused-expressions to a typescript error
```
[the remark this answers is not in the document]
greg-in-a-box: I had to turn off the one above it because it throwing an error for the TS files, if I had both on then we wouldnt be able to resolve the eslint error on the TS files even though the tslint error is gone.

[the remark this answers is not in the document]
greg-in-a-box: what we can do is add another section for js files specifically

src/components/contact-datalist-item/ContactDatalistItem.tsx
greg-in-a-box: I think there was a type there at one point but a ts error made me remove it because its assumed

tjuanitas: Gotcha, yeah it looks like other components omit the type for these properties so it should be fine here as well.

.eslintrc.js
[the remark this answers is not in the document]
tjuanitas: I'm indifferent here. Since we're committed to migrating everything to TS, the JS files will eventually be caught by this rule.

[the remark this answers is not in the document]
greg-in-a-box: Also there shouldnt be anything new as JS but as TS

tjuanitas (approved):

PR #2912, opened 2022-05-30
feat(assets): update assets to include Canvas icon

pjzabinski:


src/components/preview/previewIcons.ts
```diff
@@ -59,6 +60,7 @@ register(IconFileAudio, vars.bdlPurpleRain, [
     'wav',
     'wma',
 ]);
+register(IconFileCanvas, '#F1811A', ['boxcanvas']);
```
jstoffan, line 63: Can we use `vars.bdlOrange`?

jstoffan (dismissed):

src/components/preview/previewIcons.ts
```diff
@@ -5,6 +5,7 @@
  */
 import * as vars from '../../styles/variables';
 import IconFileAudio from '../../icon/content/FileAudio32';
+import IconFileCanvas from '../../icon/content/FileCanvas32';
 import IconFileBoxNote from '../../icon/content/FileBoxNote32';
```
tjuanitas, line 9: minor but we should sort these imports.

```suggestion
import IconFileBoxNote from '../../icon/content/FileBoxNote32';
import IconFileCanvas from '../../icon/content/FileCanvas32';
```

src/icons/file-icon/FileIcon.js.flow
```diff
@@ -2,6 +2,7 @@
 import * as React from 'react';
 
 import IconFileAudio from '../../icon/content/FileAudio32';
+import IconFileCanvas from '../../icon/content/FileCanvas32';
```
tjuanitas, line 5: nit: the usages of the icons are sorted alphabetically in this file and `src/icons/file-icon/FileIcon.tsx`.

src/icons/file/README.md
```diff
@@ -23,6 +23,10 @@ const icons = [
         name: 'FileAudio32',
         component: require('../../icon/content/FileAudio32').default,
     },
+    {
+        name: 'FileCanvas32',
+        component: require('../../icon/content/FileCanvas32').default,
+    },
```
tjuanitas, line 29: same comment here about the sorting.

src/icons/item-icon/ItemIcon.tsx
```diff
@@ -39,6 +40,7 @@ import IconFolderPersonal from '../../icon/content/FolderPersonal32';
 const itemIconTable = {
     audio: IconFileAudio,
     bookmark: BookmarkIcon,
+    boxcanvas: IconFileCanvas,
```
tjuanitas, line 43: should this be just `canvas` instead of `boxcanvas`?

the other icon names that are `IconFileName` have the pattern:
`name: IconFileName`

```suggestion
    canvas: IconFileCanvas,
```

tjuanitas (commented):
Since the design asset files are `FileCanvas[...]`, I think it would be consistent to go with `canvas` instead of `boxcanvas`.

jstoffan (dismissed):

src/icons/item-icon/ItemIcon.tsx
tjuanitas: @jstoffan I think that would be good pattern to use as that's how we handle the Google icons and it makes it clear how the icon "belongs" to a specific brand.

@pjzabinski if you decide to incorporate Jared's suggestion (`IconFileCanvas` ==> `IconFileBoxCanvas`), we should remember to make the same name change across the other files: `src/icons/file-icon/FileIcon.tsx`, `src/components/preview/previewIcons.ts`, etc.

Also would mean you would have to re-sort some of the those usages (sorry bout that!)

src/icons/file/README.md
```diff
@@ -27,6 +27,10 @@ const icons = [
         name: 'FileBoxNote32',
         component: require('../../icon/content/FileBoxNote32').default,
     },
+    {
+        name: 'FileCanvas32',
```
jstoffan, line 31: Should this be `FileBoxCanvas32`? Should the file itself also be renamed?

jstoffan (approved):

src/icons/file/README.md
jstoffan: Do we need to update the design asset to follow the new naming convention?

tjuanitas: Yeah this area I'm a little conflicted so it's more of a compromise for me.

Ideally, we should be using the same names across the different repos: design-assets, box-ui-elements, and the monolith. But because they're owned/maintained by different groups, it seems unlikely we'll be able to stay in sync.

With the exception of `FileBoxNote32`, I think the design team has been pretty consistent with their naming structure. For example, I'm seeing the following pattern: `File[product][size]`

- `FileExcel32`
- `FileSheets32`
- `FileKeynote32`
- `FilePhotoshop32`

In these cases, the file names do not include the company name which makes me lean towards `FileCanvas32` being correct here from the design team's perspective. Otherwise, we would need to rename all the files in the design-assets repo to follow the `File[brand][product][size]` pattern.

When these files are imported into BUIE is where we've seemed to stray from a standard. I think adding the company/brand in the name adds more clarity about the icon but we've only done it with the Google icons so maybe it was correct the other way? The Apple, Adobe, and Microsoft icons only have the product name.

The inconsistencies seem to stem from this table: https://github.com/box/box-ui-elements/blob/master/src/icons/item-icon/ItemIcon.tsx#L39-L73.
Because of the usages of these key values in the monolith, we can't (easily) rename the keys so it looks like we've been mirroring the variable names after them.

E.g. we currently have
```
'excel-spreadsheet': IconFileExcelSpreadsheet,
'google-slides': IconFileGoogleSlides,
```
but it should probably be
```
'excel-spreadsheet': IconFileExcel,
'google-slides': IconFileSlides,
```

I'm okay with merging this as-is and then I can address all the inconsistencies from the BUIE side in a future PR.

PR #2917, opened 2022-06-05
feat(preview): add responsive behavior to preview drawer

greathmaster:
Summary of changes:

for screen sizes less than 768px...
- moved sidebar drawer to bottom 
- animated sidebar drawer open and close
- shows greyscale version of Box Sign icon
- hides the additional apps section

Demo
Responsive-Preview-Drawer


src/elements/content-sidebar/SidebarResponsiveness.scss
```diff
@@ -0,0 +1,154 @@
...
+    @include breakpoint($medium-screen) {
+        .bcs-SidebarNavSign-icon {
+            display: none;
+        }
+
+        .u-md-bcs-SidebarNavSign-icon {
+            display: inline;
+            width: 20px;
+            height: 20px;
+            margin-top: 2px; //for visual alignment
```
ConradJChan, line 17: Should these be applied to the base style above instead of only for the medium breakpoint?

src/elements/content-sidebar/SidebarResponsiveness.scss
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        .bcpr {
+            .bcpr-body {
+                position: static;
```
ConradJChan, line 22: Does this `static` position affect the way annotations are rendered on top of preview?

src/elements/content-sidebar/SidebarResponsiveness.scss
```diff
@@ -0,0 +1,154 @@
...
+                flex-direction: column;
+            }
+
+            .bcpr-container {
+                position: static;
+            }
+        }
+
+        .preview-body {
+            height: 100%;
```
jstoffan, line 32: Relative height is notoriously difficult to work with, since it has to either have a fixed-height parent or go all the way up to the root of the document. Should we use flexbox or absolute positioning, instead?

src/elements/content-sidebar/SidebarResponsiveness.scss
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        .be {
+            height: 100%;
+
+            .bcs-activity-feed {
```
jstoffan, line 38: Can these styles be moved to the components they affect?

src/elements/content-sidebar/SidebarResponsiveness.scss
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        .be.bcs {
+            flex-direction: column;
+            width: 100%;
+            min-width: unset;
+            max-width: unset;
+            height: 100%;
+            max-height: 48px;
+            -webkit-transition: max-height .5s ease-in-out 0s;
```
jstoffan, line 54: We should have autoprefixer, making this unnecessary.

src/elements/content-sidebar/SidebarResponsiveness.scss
```diff
@@ -0,0 +1,154 @@
...
+
+            .bcs-scroll-content {
+                width: unset;
+            }
+        }
+
+        .be.bcs.bcs-is-open {
+            position: absolute;
+            bottom: 0;
+            z-index: 2;
```
jstoffan, line 70: Is this needed? If so, does it work properly with all other floating/tethered elements (annotations, tooltips, modals, etc.)?

src/elements/content-sidebar/SidebarResponsiveness.scss
```diff
@@ -0,0 +1,154 @@
...
+
+        .bdl-AdditionalTabs > *:not(.bdl-MoreOptions) {
+            display: none;
+        }
+
+        .bcs .bcs-SidebarNav-overflow .bdl-AdditionalTabs {
+            padding: 0;
+        }
+
+        .bcs .bcs-SidebarNav-overflow::after,
```
jstoffan, line 112: Let's revisit the amount of specificity we're using and also consolidate our style rules for the same components using SCSS nesting.

src/elements/content-sidebar/SidebarResponsiveness.scss
```diff
@@ -0,0 +1,154 @@
...
+        .bcs .bcs-SidebarNav-overflow::before {
+            display: none;
+        }
+
+        .bcs .bcs-SidebarNav-footer {
+            flex: 1 1 48px;
```
jstoffan, line 118: What is the source of this value? Should it be a variable?

.storybook/preview-body.html
```diff
@@ -1,6 +1,7 @@
 <style>
     #root {
         padding: 16px;
+        position: relative;
```
jstoffan, line 4: Is this needed? Will customers using Preview need to add this to their stylesheets for the new styles to work?

src/elements/content-sidebar/SidebarNavSign.tsx
```diff
@@ -55,6 +56,7 @@ export function SidebarNavSign({ blockedReason, intl, status, targetingApi, ...r
             <Tooltip isDisabled={isTargeted} position={TooltipPosition.MIDDLE_LEFT} text={tooltipMessage}>
                 <PlainButton aria-label={label} className={buttonClassName} isDisabled={isSignDisabled} {...rest}>
                     <BoxSign28 className="bcs-SidebarNavSign-icon" />
+                    <Sign16 className="u-md-bcs-SidebarNavSign-icon" />
```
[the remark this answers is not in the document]
greathmaster: This was requested by Product to change the Box Sign icon to the grayscale version when it shifts to the horizontal drawer on small screens.

src/elements/content-sidebar/additional-tabs/AdditionalTab.js
```diff
@@ -101,6 +101,7 @@ class AdditionalTab extends React.PureComponent<Props, State> {
         const className = classNames('bdl-AdditionalTab', {
             'bdl-is-hidden': isLoading,
             'bdl-is-disabled': isDisabled,
+            'bdl-MoreOptions': id && id < 0,
```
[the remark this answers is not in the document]
greathmaster: Done.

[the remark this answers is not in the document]
greathmaster: Going to go with @jstoffan suggestion of `bdl-is-overflow`

src/elements/content-sidebar/SidebarResponsiveness.scss
greathmaster: The greyscale Sign icon is only meant to show up on medium screen sizes and below, so it's going to be hidden on anything bigger. Functionally, it would end up being the same. I'll go ahead and update it if it's more clear that way.

src/elements/content-sidebar/SidebarResponsiveness.scss
greathmaster: Annotations are unaffected by the `position: static`:
Screen Shot 2022-06-29 at 1 44 10 PM

src/elements/content-sidebar/SidebarResponsiveness.scss
greathmaster: Understood. Going to remove vendor prefixes.

src/elements/content-sidebar/SidebarResponsiveness.scss
greathmaster: The `z-index` is needed so the drawer displays over the Preview Header when expanded.

Screen Shot 2022-06-29 at 3 12 30 PM

src/elements/content-sidebar/SidebarResponsiveness.scss
```diff
@@ -0,0 +1,154 @@
...
+                width: unset;
+            }
+        }
+
+        .be.bcs.bcs-is-open {
+            position: absolute;
+            bottom: 0;
+            z-index: 2;
+            height: 100%;
+            max-height: 100%;
```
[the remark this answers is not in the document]
greathmaster: The `max-height` is being used as the transition property for the drawer animation. Using it just on `height` would cause the drawer to animate when resized.

src/elements/content-sidebar/SidebarResponsiveness.scss
```diff
@@ -0,0 +1,154 @@
...
+
+        .bcs .bcs-NavButton,
+        .bcs-SidebarNav-secondary .btn-plain,
+        .bdl-AdditionalTab {
+            width: 48px;
+            height: 48px;
+            margin-right: 12px;
+        }
+
+        .bdl-AdditionalTabs > *:not(.bdl-MoreOptions) {
```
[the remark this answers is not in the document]
greathmaster: Updated to use `.bdl-AdditionalTab` instead.

src/elements/content-sidebar/SidebarResponsiveness.scss
```diff
@@ -0,0 +1,154 @@
...
+            min-width: unset;
+            max-width: unset;
+            height: 100%;
+            max-height: 48px;
+            -webkit-transition: max-height .5s ease-in-out 0s;
+            transition: max-height .5s ease-in-out 0s;
+        }
+
+        .bcs-content {
+            width: unset;
```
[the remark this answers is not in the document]
greathmaster: Yes. Also updated `auto` in couple other places as well.

src/elements/content-sidebar/SidebarResponsiveness.scss
```diff
@@ -0,0 +1,154 @@
...
+            position: absolute;
+            bottom: 0;
+            z-index: 2;
+            height: 100%;
+            max-height: 100%;
+            -webkit-transition: max-height .5s ease-in-out 0s;
+            transition: max-height .5s ease-in-out 0s;
+        }
+
+        .bcs {
```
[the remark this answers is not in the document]
greathmaster: Removed the nesting.

src/elements/content-sidebar/SidebarResponsiveness.scss
```diff
@@ -0,0 +1,154 @@
+@import '../common/variables';
+
+.u-md-bcs-SidebarNavSign-icon {
```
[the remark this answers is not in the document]
greathmaster: Seems I misunderstood how the `u-` prefix is used. I've updated the classname to `bcs-SidebarNavSign-icon--greyscale` instead.

src/elements/content-sidebar/SidebarResponsiveness.scss
greathmaster: Instead of using `height: 100%` I've updated it to use `flex: basis: 100%` instead.

src/elements/content-sidebar/SidebarResponsiveness.scss
```diff
@@ -0,0 +1,154 @@
...
+            .bcpr-container {
+                position: static;
+            }
+        }
+
+        .preview-body {
+            height: 100%;
+        }
+
+        .be {
```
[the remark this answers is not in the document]
greathmaster: Removed the `height: 100%` for the `.be` class as it was unnecessary.

src/elements/content-sidebar/SidebarResponsiveness.scss
greathmaster: Moved to `ActivityFeed.scss`

src/elements/content-sidebar/SidebarResponsiveness.scss
greathmaster: Consolidated styles and reduced specificity in many places.

src/elements/content-sidebar/ContentSidebar.js
```diff
@@ -39,6 +39,7 @@ import '../common/fonts.scss';
 import '../common/base.scss';
 import '../common/modal.scss';
 import './ContentSidebar.scss';
+import './SidebarResponsiveness.scss';
```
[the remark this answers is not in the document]
greathmaster: Managed to move all the styles into their appropriate files so deleting `SidebarResponsive.scss`.

.storybook/preview-body.html
greathmaster: Adding `position: relative` here is to anchor the `position: absolute` used further down.

I'm unsure if customers will need to add this to their stylesheets.

For that reason, I've updated the code to remove it from `#root` and instead add `position: relative` to the `.be.bcpr` selector. This will leave the `preview-body.html` file unchanged.

This looks okay in Storybook (image below). However, in EUA, we add an additional header with the filename (image below).

To address this I will need to submit a PR to EUA that unsets the `be.bcpr` selector and instead adds the `position: relative` to the `.preview-container`. Seems like a better solution than adding `position: relative` to the `#root` element.

**Storybook**
Screen Shot 2022-07-01 at 6 26 15 PM

**EndUserApp would need a couple additional styles.**
Screen Shot 2022-07-01 at 6 08 06 PM

src/elements/content-sidebar/SidebarResponsiveness.scss
```diff
@@ -0,0 +1,154 @@
...
+
+.is-responsive-web {
+    @include breakpoint($medium-screen) {
+        .bcs-SidebarNavSign-icon {
+            display: none;
+        }
+
+        .u-md-bcs-SidebarNavSign-icon {
+            display: inline;
+            width: 20px;
```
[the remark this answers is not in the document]
greathmaster: I went ahead and did this, but could I get some clarification on why this is better than specifying in CSS?

Seems setting it in JSX makes it hard to change or override the style later?

src/elements/content-sidebar/SidebarResponsiveness.scss
greathmaster: Did some more testing and seems we don't need this property. I've removed it.

src/elements/content-sidebar/SidebarResponsiveness.scss
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        .bcpr {
+            .bcpr-body {
+                position: static;
+                flex-direction: column;
+            }
+
+            .bcpr-container {
+                position: static;
```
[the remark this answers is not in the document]
greathmaster: I've tested this across several formats and it seems to be okay. Additionally, from what I have tested, Annotations render correctly.

src/elements/content-sidebar/SidebarResponsiveness.scss
```diff
@@ -0,0 +1,154 @@
...
+            }
+        }
+
+        .preview-body {
```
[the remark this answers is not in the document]
greathmaster: Personally, I don't think we should move the styles to the Preview SDK because, the additional styles are in support of the Sidebar moving to the bottom on small screens. My thinking is it's better to keep all the functionality in one repo rather than splitting it.

src/elements/content-sidebar/SidebarNav.scss
```diff
@@ -63,3 +63,88 @@
...
+        .bcs-NavButton::before {
+            right: 0;
+            bottom: 0;
+            left: 0;
+            height: 2px;
+        }
+
+        .bcs-SidebarNav-secondary .btn-plain,
+        .bdl-AdditionalTab {
+            width: 48px;
```
jstoffan, line 97: This seems to be a common measurement. Can it be moved to a local variable?

src/elements/content-sidebar/SidebarNav.scss
```diff
@@ -63,3 +63,88 @@
...
+        .bcs-SidebarNav-tabs {
+            flex-direction: row;
+        }
+
+        .bcs-SidebarNav-main {
+            display: flex;
+            flex-direction: row;
+        }
+
+        .bcs-NavButton::before {
```
jstoffan, line 88: Can we group this with and move after the other `.bcs-NavButton` styles? Can we also move those styles to the SidebarNavButton component stylesheet?

src/elements/content-sidebar/SidebarNav.scss
```diff
@@ -63,3 +63,88 @@
...
+        .bcs-SidebarNav-secondary .btn-plain,
+        .bdl-AdditionalTab {
+            width: 46px;
+            margin-right: 0;
+        }
+    }
+
+    @include breakpoint($medium-screen) {
+        .bcs-SidebarNav,
+        .bcs-SidebarNav-tabs {
```
jstoffan, line 79: Can we add `.bcs-SidebarNav-main` to the list of classes here?

src/elements/content-sidebar/SidebarNav.scss
```diff
@@ -63,3 +63,88 @@
         }
     }
 }
+
+.is-responsive-web {
+    @include breakpoint($small-screen) {
+        .bcs .bcs-NavButton,
+        .bcs-SidebarNav-secondary .btn-plain,
+        .bdl-AdditionalTab {
+            width: 46px;
```
jstoffan, line 72: Why did we lose a couple pixels here?

src/elements/content-sidebar/SidebarNav.scss
```diff
@@ -63,3 +63,88 @@
...
+            }
+
+            &::after,
+            &::before {
+                display: none;
+            }
+        }
+
+        .bcs-NavButton {
+            width: 48px;
```
jstoffan, line 124: Duplicated styles from above. Should they be moved to a mixin?

src/elements/content-sidebar/SidebarNav.scss
```diff
@@ -63,3 +63,88 @@
...
+        }
+
+        .bcs-SidebarNav-secondary .btn-plain,
+        .bdl-AdditionalTab {
+            width: 48px;
+            height: 48px;
+            margin-right: 12px;
+        }
+
+        .bdl-AdditionalTabs > .bdl-AdditionalTab:not(.bdl-is-overflow),
```
jstoffan, line 102: Is the extra specificity of a direct descendent selector needed here? We don't nest these components.

src/elements/content-sidebar/SidebarNav.scss
```diff
@@ -63,3 +63,88 @@
...
+        .bdl-AdditionalTabs > .bdl-AdditionalTabPlaceholder {
+            display: none;
+        }
+
+        .bdl-AdditionalTabs > :last-child,
```
jstoffan, line 107: What are we selecting here?

src/elements/content-sidebar/SidebarNav.scss
```diff
@@ -63,3 +63,88 @@
...
+            margin-right: 12px;
+
+            &::before {
+                top: 48px;
+                width: auto;
+            }
+        }
+
+        .bcs-SidebarNav-footer {
+            .btn-plain.bdl-SidebarToggleButton {
```
jstoffan, line 135: Can these styles be moved to the SidebarToggleButton stylesheet?

src/elements/content-sidebar/ContentSidebar.scss
```diff
@@ -28,3 +28,26 @@ $sidebarWidth: $sidebarTabsWidth + $sidebarContentWidth;
...
+            max-height: 48px;
+            transition: max-height .5s ease-in-out 0s;
+
+            &.bcs-is-open {
+                position: absolute;
+                bottom: 0;
+                z-index: 2;
+                flex-basis: 100%;
+                max-height: 100%;
+                transition: max-height .5s ease-in-out 0s;
```
jstoffan, line 49: Do we need to re-declare the transition here?

src/elements/content-sidebar/SidebarNav.scss
jstoffan: It looks like I asked this and it was answered [previously](https://github.com/box/box-ui-elements/pull/2917#discussion_r907874219). Can we add a brief comment in the code to avoid confusion in the future?

src/elements/content-sidebar/SidebarNavSign.tsx
```diff
@@ -55,6 +56,7 @@ export function SidebarNavSign({ blockedReason, intl, status, targetingApi, ...r
             <Tooltip isDisabled={isTargeted} position={TooltipPosition.MIDDLE_LEFT} text={tooltipMessage}>
                 <PlainButton aria-label={label} className={buttonClassName} isDisabled={isSignDisabled} {...rest}>
                     <BoxSign28 className="bcs-SidebarNavSign-icon" />
+                    <Sign16 width={20} height={20} className="bcs-SidebarNavSign-icon--greyscale" />
```
tjuanitas, line 59: nit: majority of spelling in this repo is `gray` vs `grey`

```suggestion
                    <Sign16 width={20} height={20} className="bcs-SidebarNavSign-icon--grayscale" />
```

src/elements/content-sidebar/ContentSidebar.scss
```diff
@@ -28,3 +28,26 @@ $sidebarWidth: $sidebarTabsWidth + $sidebarContentWidth;
...
+            min-width: 0;
+            max-width: none;
+            max-height: 48px;
+            transition: max-height .5s ease-in-out 0s;
+
+            &.bcs-is-open {
+                position: absolute;
+                bottom: 0;
+                z-index: 2;
+                flex-basis: 100%;
```
tjuanitas, line 47: Is setting `flex-basis` needed here since it's being set above?

src/elements/content-sidebar/SidebarNav.scss
greathmaster: Moved it to a local variable `$sidebarTabResponsiveSize` and updated properties.

src/elements/content-sidebar/SidebarNav.scss
greathmaster: Consolidated `.bcs-NavButton` styles, and moved it to `SidebarNavButton.scss`

src/elements/content-sidebar/SidebarNav.scss
greathmaster: Added a comment to clarify.

src/elements/content-sidebar/SidebarNav.scss
greathmaster: Tested this...seems like it isn't. Reduced specificity.

src/elements/content-sidebar/SidebarNav.scss
greathmaster: The last child in the `bdl-AdditionalTabs` is the ... button for More Apps. While the last child in `bcs-SidebarNav-secondary` is the Sign button. Setting the `margin-right: 0` for last child of both of these avoids pushing out the footer button:

#### Without setting `margin: 0`
Screen Shot 2022-07-13 at 5 13 17 PM

src/elements/content-sidebar/SidebarNav.scss
greathmaster: Maybe...The regular (non-responsive) `.bcs-SidebarNav-footer` style is already in `SidebarNav.scss`:

https://github.com/box/box-ui-elements/blob/master/src/elements/content-sidebar/SidebarNav.scss#L52-L64

So I think we should either move both of them or keep both in `SidebarNav.scss`.

src/elements/content-sidebar/ContentSidebar.scss
greathmaster: Hmm...actually we don't.

src/elements/content-sidebar/SidebarNavSign.tsx
greathmaster: Got it. Going with `gray`

src/elements/content-sidebar/ContentSidebar.scss
greathmaster: Yeah. You're right. We don't need it again. Removed.

src/elements/content-sidebar/SidebarNav.scss
greathmaster: Not sure this is still applicable, since it was moved to `SideNavButton.scss`?

src/elements/content-preview/ContentPreview.scss
```diff
@@ -75,3 +75,22 @@
...
+
+        .bcpr {
+            .bcpr-body {
+                position: static;
+                flex-direction: column;
+            }
+
+            .bcpr-container {
+                position: static;
+            }
```
tjuanitas, line 93: question: Is this for modifying the position of the navigation buttons (`.bcpr-navigate-left` and `.bcpr-navigate-right`) to make them vertically centered with the screen as opposed to the preview container when the sidebar is on the bottom?

If not, what's the reason for this change?

src/elements/content-preview/ContentPreview.scss
```diff
@@ -75,3 +75,22 @@
...
+
+.is-responsive-web {
+    @include breakpoint($medium-screen) {
+        .be.bcpr {
+            position: relative;
+        }
+
+        .bcpr {
+            .bcpr-body {
+                position: static;
```
tjuanitas, line 87: Is `position: static` needed?

For screen sizes between 700px (`.be-is-small`) and 768px, the position is `static` by default.

For screen sizes 700px and less (i.e. when `.be-is-small` is applied), `.bcpr-body` has `position: relative;` ([from this](https://github.com/box/box-ui-elements/blob/master/src/elements/content-preview/ContentPreview.scss#L63-L66)) but because `.bcpr` and `.bcpr-body` are the same sizes, this shouldn't have an effect.

tjuanitas: Alternatively, I think we could set `.bcpr-body` to `position: relative;` here and then remove these lines: https://github.com/box/box-ui-elements/blob/3b8e7767d3b88ec926e99c278f37843c1292ade1/src/elements/content-preview/ContentPreview.scss#L81-L83

This allows us in the future to remove this: https://github.com/box/box-ui-elements/blob/3b8e7767d3b88ec926e99c278f37843c1292ade1/src/elements/content-preview/ContentPreview.scss#L64-L66

greathmaster: The main reason for applying `position: static` was to ensure the `position: absolute` from `aside` had a clear path to anchor on the `position relative` of 

`.be.bcpr { 
     position: relative; 
 } `

I tried the above and in Storybook it seems the Preview drawer doesn't fully cover the filename header when extended.

localhost_6061__path=_story_elements-contentpreview--preview-with-annotations

tjuanitas (approved):

tjuanitas:
@Mergifyio unqueue

tjuanitas:
@Mergifyio refresh

PR #2919, opened 2022-06-06
build(node): Bump to lts/fermium

greg-in-a-box:
Bumped to node `14.17.3`
Bumped docker image used in circleci to `cypress/base:14.17.3`
Updated package.json engines node ranges
Updated doc specifiying node version


DEVELOPING.md
```diff
@@ -4,7 +4,7 @@ Our development setup assumes a LINUX/BSD environemnt.
 
 ## Project Setup
 
-1. Install Node v12.
+1. Install Node v14.17.3.
```
tjuanitas, line 7: Do we need to specify the exact version or could it be `Node v14`?

tjuanitas (commented):
Have we tested that the `build` / `build:[...]` and `start` / `start:[...]` script commands still work?

Would this have an impact when linking in parent apps (e.g. `yarn link box-ui-elements`)?

.nvmrc
```diff
@@ -1 +1 @@
-lts/erbium
+14.17.3
```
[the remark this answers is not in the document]
tjuanitas: Since this is the `.nvmrc` file, isn't this more of the "preferred" node version to use for the project instead of the lowest supported version? As long as it's within the engine range it should be fine right?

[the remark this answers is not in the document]
greg-in-a-box: `lts/fermium` points to `14.19.3` and apparently docker doesnt have a cypress image for https://hub.docker.com/r/cypress/base/tags?page=1&name=14.19.3

greg-in-a-box:
> Have we tested that the `build` / `build:[...]` and `start` / `start:[...]` script commands still work?
> 
> Would this have an impact when linking in parent apps (e.g. `yarn link box-ui-elements`)?



> Have we tested that the `build` / `build:[...]` and `start` / `start:[...]` script commands still work?
> 
> Would this have an impact when linking in parent apps (e.g. `yarn link box-ui-elements`)?

Yea specifying the version is really for the circleci docker image that is being used.
If a dev goes above right below 15 or below 14 it should work the same still but see a warning to switch to in between the ranges. if a dev goes to 16 though, thats where we start seeing some issues.

`yarn build`, `yarn start` and `yarn start:storybook` seem to be working without errors.  I am able to see the styleguide still and storybook

for linking 
I tried the following:
`nvm use` in both terminals of admin console and buie 
in buie terminal:
`yarn build:npm`
`yarn link`
in admin console terminal:
`yarn link box-ui-elements`
`yarn build:dev`
only 1 warning came up which was related to pikaday revolving around moment but that is normal when you dont have moment installed and you're using pikaday
other than that checked my devvm and it seems to be fine.

package.json
```diff
@@ -118,7 +118,7 @@
         }
     },
     "engines": {
-        "node": ">=10.x",
+        "node": ">=14.0.0 <15.0.0",
```
tjuanitas, line 121: super nitpicky: I think there should be an added space
```suggestion
        "node": ">=14.0.0 < 15.0.0",
```

greg-in-a-box: Actually, there isnt supposed to be one based on the example here )https://docs.npmjs.com/cli/v7/configuring-npm/package-json#engines)

tjuanitas (approved):

PR #2921, opened 2022-06-07
feat(preview): update freemium preview sidebar for upsell

wlyhdg:
present the winning treatment preview sidebar for when upgrading is applicable
Screen Shot 2022-06-07 at 12 03 55 PM

### DIFF
--- a/src/elements/content-sidebar/versions/VersionsSidebarContainer.js
+++ b/src/elements/content-sidebar/versions/VersionsSidebarContainer.js
@@ -296,7 +296,6 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
                 onPreview={this.handleActionPreview}
                 onPromote={this.handleActionPromote}
                 onRestore={this.handleActionRestore}
-                onUpgradeClick={onUpgradeClick}
                 parentName={parentName}
                 {...this.state}
             />
--- a/src/elements/content-sidebar/versions/VersionsSidebarContainer.js
+++ b/src/elements/content-sidebar/versions/VersionsSidebarContainer.js
@@ -38,18 +38,10 @@ type Props = {
     onVersionRestore: VersionActionCallback,
     parentName: string,
     versionId?: string,
-    versionUpsellExperience: ?string,
-};
-
-const UPSELL_TYPE = {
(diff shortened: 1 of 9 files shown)


src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -37,14 +37,8 @@ type Props = {
     onVersionPromote: VersionActionCallback,
     onVersionRestore: VersionActionCallback,
     parentName: string,
+    showUpsellExperience: boolean,
```
jstoffan, line 40: Is this still needed or can we use the existence of `onUpgradeClick` to determine if the static sidebar should be shown?

jstoffan (approved):
Thanks for simplifying this code path. In the future, it can be helpful to include a screenshot and description so folks can quickly assess the new behaviors.

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
wlyhdg: @jstoffan onUpgradeClick is being passed as `onUpgradeClick: showUpsellExperience && onUpgradeClick` so I do think that our safeguard would still apply. this is a good simplification but do you think this syntax is clear enough for a user (not having the showUpsellExperience explicitly and just using onUpgradeClick)? if so can make the updates!

wlyhdg:
> Thanks for simplifying this code path. In the future, it can be helpful to include a screenshot and description so folks can quickly assess the new behaviors.

thank you, will do! going to add with a few updates

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
jstoffan: Yeah, I think it should be fine. We're likely the only ones using these props and `onUpgradeClick` isn't used for anything else currently. Best to keep the API as small as possible, I think.

jstoffan (dismissed):

jstoffan (approved):

rdacanay (dismissed):

jstoffan (approved):

PR #2924, opened 2022-06-08
bdrvgljrlgjjltfircnrujrdfvruejuriijbtlrcvttdrefactor(typescript): Migrate ProgressBar component to typescript

apieceofbart:


src/components/progress-bar/ProgressBar.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+
+import './ProgressBar.scss';
+
+type Props = {
```
greg-in-a-box, line 5:
```suggestion
export interface ProgressBarProps {
```

src/components/progress-bar/ProgressBar.tsx
```diff
@@ -0,0 +1,23 @@
...
+    className?: string;
+    /** A number between 0 and 100 inclusive representing the width percentage */
+    progress?: number;
+};
+
+const ProgressBar = ({ className = '', progress = 0 }: Props) => {
```
greg-in-a-box, line 11:
```suggestion
const ProgressBar = ({ className = '', progress = 0 }: ProgressBarProps) => {
```

src/components/progress-bar/__tests__/ProgressBar.test.tsx
```diff
@@ -13,7 +14,8 @@ describe('components/progress-bar/ProgressBar', () => {
         const component = renderComponent({
             progress: 20,
         });
-        expect(component.find('.progress-bar').prop('style').width).toEqual(expected.width);
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        expect(component.find('.progress-bar').prop('style')!.width).toEqual(expected.width);
```
greg-in-a-box, line 18:
```suggestion
        expect(component.find('.progress-bar').props().style?.width).toEqual(expected.width);
```

src/components/progress-bar/__tests__/ProgressBar.test.tsx
```diff
@@ -28,11 +30,13 @@ describe('components/progress-bar/ProgressBar', () => {
...
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
             component
                 .find('.progress-bar')
-                .prop('className')
+                .prop('className')!
                 .indexOf(expected.className) !== -1,
         ).toBeTruthy();
-        expect(component.find('.progress-bar').prop('style').width).toEqual(expected.width);
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        expect(component.find('.progress-bar').prop('style')!.width).toEqual(expected.width);
```
greg-in-a-box, line 40:
```suggestion
        expect(component.find('.progress-bar').props().style?.width).toEqual(expected.width);
```

src/components/progress-bar/__tests__/ProgressBar.test.tsx
```diff
@@ -28,11 +30,13 @@ describe('components/progress-bar/ProgressBar', () => {
             progress: 20,
         });
         expect(
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
             component
                 .find('.progress-bar')
-                .prop('className')
+                .prop('className')!
                 .indexOf(expected.className) !== -1,
```
greg-in-a-box, line 37:
```suggestion
                .props()
                .className?.indexOf(expected.className) !== -1,
```

greg-in-a-box (changes requested):
Left some suggestions
Highlights:

- we should convert `type` to `interface` and `export` it 
- we should try to avoid using lint ignores it possible, for example I have left suggestions on avoiding `@typescript-eslint/no-non-null-assertion` 
- we should also use `props().property` instead of using `prop('property')` since using the `props().property` usage does become type safe while it is possible that if `property` is deleted and `prop('property')` is being checked to be `undefined`, this will still pass.

src/components/progress-bar/__tests__/ProgressBar.test.tsx
```diff
@@ -28,11 +30,13 @@ describe('components/progress-bar/ProgressBar', () => {
             progress: 20,
         });
         expect(
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
greg-in-a-box, line 33: should be able to remove this after the change below

src/components/progress-bar/ProgressBar.tsx
apieceofbart: how interface is better than type and we should we export it now?
please note there `type` is used in current implementations

apieceofbart:
> we should also use props().property instead of using prop('property') since using the props().property usage does become type safe while it is possible that if property is deleted and prop('property') is being checked to be undefined, this will still pass.

Are you suggesting 'styles' would be removed from the html spec?! I better check a lot of code then..

apieceofbart:
Just for you to understand, last time I did one of those refactorings, I changed a lot of code (I think for the better) but I was told that the migration should be minimal so it's done faster. So I'm deliberately not changing anything unless I need to. 
You guys could make up your mind, otherwise I'm wasting a lot of time - it's not really encouraging

src/components/progress-bar/ProgressBar.tsx
greg-in-a-box: Thats a good question, it specified in the documentation for TS migration but I do not know why because I joined after the decision.

greg-in-a-box:
> > we should also use props().property instead of using prop('property') since using the props().property usage does become type safe while it is possible that if property is deleted and prop('property') is being checked to be undefined, this will still pass.
> 
> I don't think `styles` or `className` are going to be removed anytime soon but I see your point. Any particular reason why one works and the other don't? Seems like a bug in enzyme typings.

Not sure if its a bug with `prop("property")` but its what I noticed when I was reevaluating the TS migration guide,  my guess its because its trying to do something like `props["property"]` which doesnt have a type available for it on their side

src/components/index.ts
```diff
@@ -90,3 +90,6 @@ export * from './section';
 
 export { default as TimeInput } from './time-input';
 export * from './time-input';
+
+export { default as ProgressBar } from './progress-bar';
+export * from './progress-bar';
```
tjuanitas, line 95: It looks like most of this file is sorted alphabetically so we can try to add the entry after `PrimaryButton`: https://github.com/box/box-ui-elements/blob/master/src/components/index.ts#L1-L74

src/components/progress-bar/__tests__/ProgressBar.test.tsx
```diff
@@ -1,3 +1,4 @@
+import { shallow } from 'enzyme';
 import React from 'react';
```
tjuanitas, line 2: We should have `import React from 'react';` as the first import.

tjuanitas, line 2: We should have `import React from 'react';` as the first import.

tjuanitas (commented):
Our TS migration runbook also says to add a `index.js.flow` file but I'm not sure if that's still needed.

src/components/progress-bar/__tests__/ProgressBar.test.tsx
apieceofbart: we should have prettier to do that for use in the first place...

src/components/index.ts
apieceofbart: "most of it" :) let's sort them all then

apieceofbart:
> Not sure if its a bug with prop("property") but its what I noticed when I was reevaluating the TS migration guide, my guess its because its trying to do something like props["property"] which doesnt have a type available for it on their side

I mean technically there's no reason why one would work and the other don't. Perhaps some performance issues, to be honest I'm not very keen on finding out since enzyme is a thing of past now

src/components/progress-bar/ProgressBar.tsx
apieceofbart: We won't avoid type aliases because that's the only to do unions, for example so it's not like we're going to be consistent anyway. I would enforce one over other tbh. 
I've change to interface as you wished.

src/components/progress-bar/__tests__/ProgressBar.test.tsx
apieceofbart: fixed

src/components/index.ts
apieceofbart: fixed

src/components/progress-bar/__tests__/ProgressBar.test.tsx
greg-in-a-box: Made to ticket to config prettier to do this

tjuanitas (approved):
Thanks! LGTM

greg-in-a-box (approved):

PR #2926, opened 2022-06-09
feat(modal): default ugrade inline notice to top of share modal

wlyhdg:
remove treatment code updates we don't plan to release (remove visuals related to `showNewUpgradeText`; keep winning visuals related to `showUpgradeInlineNotice`)

upgrading user will see:
Screen Shot 2022-06-08 at 4 42 36 PM

### DIFF
--- a/src/features/unified-share-modal/UnifiedShareForm.js
+++ b/src/features/unified-share-modal/UnifiedShareForm.js
@@ -575,32 +575,28 @@ class UnifiedShareForm extends React.Component<USFProps, State> {
     }
 
     renderUpgradeLinkDescription() {
-        const { openUpgradePlanModal = () => {}, showNewUpgradeText = false, trackingProps = {} } = this.props;
+        const { openUpgradePlanModal = () => {}, trackingProps = {} } = this.props;
         const { inviteCollabsEmailTracking = {} } = trackingProps;
         const { upgradeLinkProps = {} } = inviteCollabsEmailTracking;
 
         return (
             <div className="upgrade-description">
                 <UpgradeBadge />
-                {showNewUpgradeText ? (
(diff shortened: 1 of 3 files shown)


src/features/unified-share-modal/UnifiedShareForm.js
```diff
@@ -575,32 +575,28 @@ class UnifiedShareForm extends React.Component<USFProps, State> {
...
-        const { openUpgradePlanModal = () => {}, showNewUpgradeText = false, trackingProps = {} } = this.props;
+        const { openUpgradePlanModal = () => {}, trackingProps = {} } = this.props;
         const { inviteCollabsEmailTracking = {} } = trackingProps;
         const { upgradeLinkProps = {} } = inviteCollabsEmailTracking;
 
         return (
             <div className="upgrade-description">
                 <UpgradeBadge />
-                {showNewUpgradeText ? (
-                    this.renderCollaboratorMessage('external_collab_newcopy_upgrade_cta')
```
wlyhdg, line 586: still using this function to render message for winning treatment code (it was the same copy, different places). Indifferent about keeping the function or moving it more inline but it feels cleaner this way so I left it. lmk what you think.

tclancy (approved):

src/features/unified-share-modal/UnifiedShareForm.js
```diff
@@ -575,32 +575,28 @@ class UnifiedShareForm extends React.Component<USFProps, State> {
     }
 
     renderUpgradeLinkDescription() {
```
jackiejou, line 577: since we are rolling inline Notice, we should also remove this function and always show renderUpgradeInlineNotice.
showUpgradeInlineNotice can also be removed

PR #2932, opened 2022-06-15
feat(content-explorer): Add responsive behavior

greathmaster:
Add responsive behavior to the ContextExplorerModal and ContentExplorer used in the MoveCopyModal

Content-Explorer-Responsive

### Summary of Changes
- Added `AutoSizer` to ReactVirtualized's `Table` component.
- Added `isResponsive` prop to conditionally add styles and make the responsive version opt-in.


src/features/content-explorer/content-explorer-modal/ContentExplorerModal.js
```diff
@@ -25,20 +26,26 @@ const ContentExplorerModal = ({
...
     ...rest
 }: Props) => (
     <Modal
         title={title}
-        className={classNames('content-explorer-modal', className)}
+        className={classNames(
+            'content-explorer-modal',
+            isResponsive ? 'content-explorer-modal--isResponsive' : '',
+            className,
+        )}
```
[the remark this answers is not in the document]
tjuanitas: tangent: does anybody know how we would handle conflicts for two components with the same name and using SUIT?

There are technically two ContentExplorer components:
`src/elements/content-explorer/ContentExplorer.js`
`src/features/content-explorer/content-explorer/ContentExplorer.js`

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -37,6 +37,9 @@ class ContentExplorer extends Component {
         initialFoldersPath: FoldersPathPropType.isRequired,
         /** Initial items that will show up as selected */
         initialSelectedItems: PropTypes.object,
+        /** Whether to use the responsive version */
+        isResponsive: PropTypes.bool,
+
```
tjuanitas, line 42: nit: extra line

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -453,7 +457,11 @@ class ContentExplorer extends Component {
         return (
             // eslint-disable-next-line jsx-a11y/no-static-element-interactions, jsx-a11y/click-events-have-key-events
             <div
-                className={classNames('content-explorer', className)}
+                className={classNames(
+                    'content-explorer',
+                    isResponsive ? 'content-explorer--isResponsive' : '',
+                    className,
+                )}
```
tjuanitas, line 464: Same comment as the modal component.

```suggestion
                className={classNames('content-explorer', className, {
                    'bdl-ContentExplorer--responsive': isResponsive,
                })}
```

src/features/content-explorer/content-explorer/ContentExplorerActionButtons.js
```diff
@@ -88,7 +88,7 @@ const ContentExplorerActionButtons = ({
     };
 
     return (
-        <div className="content-explorer-action-buttons-container" {...actionButtonsProps}>
+        <div className="modal-actions" {...actionButtonsProps}>
```
tjuanitas, line 91: I'm not sure if these classes are swappable.

There are nested styles in `content-explorer-action-buttons-container` that are not in `modal-actions` such as `status-message` and `content-explorer-copy-button`:
https://github.com/box/box-ui-elements/blob/master/src/features/content-explorer/content-explorer/ContentExplorer.scss#L83-L103

The `position: relative;` is also important here so we'd need to confirm that none of the child components that are positioned `absolute` are effected.

Let's also verify the impact in the parent apps that might have repo specific styles for this class.

src/features/content-explorer/item-list/ItemList.js
```diff
@@ -5,6 +5,7 @@ import getProp from 'lodash/get';
 
 import Column from '@box/react-virtualized/dist/commonjs/Table/Column';
 import Table from '@box/react-virtualized/dist/commonjs/Table';
+import AutoSizer from '@box/react-virtualized/dist/commonjs/AutoSizer';
```
tjuanitas, line 8: super nit: we could move this import above `Column` so that the `Table` related components are grouped together and sorted.

src/features/content-explorer/item-list/ItemList.js
```diff
@@ -188,8 +190,30 @@ const ItemList = ({
...
+    TableComponent = addAutoSizer(TableComponent);
+
     return (
-        <div className={classNames('content-explorer-item-list table', className)}>
+        <div
+            className={classNames(
+                'content-explorer-item-list table',
+                isResponsive ? 'content-explorer-item-list--isResponsive' : '',
+                className,
+            )}
```
tjuanitas, line 215:
```suggestion
            className={classNames('content-explorer-item-list table', className, {
                'bdl-ItemList--responsive': isResponsive,
            })}
```

I'm not 100% sure about the SUIT name here but seems correct?

src/features/content-explorer/item-list/ItemList.js
```diff
@@ -188,8 +190,30 @@ const ItemList = ({
         };
     }
 
+    const addAutoSizer = WrappedComponent => {
+        return props => {
+            return isResponsive ? (
+                <div style={{ flex: 1 }}>
+                    <AutoSizer>
+                        {({ width: w, height: h }) => <WrappedComponent {...props} width={w} height={h} />}
```
tjuanitas, line 198: I think it's okay to use the full names here instead `w` and `h`.
```suggestion
                        {({ height, width }) => <WrappedComponent height={height} width={width} {...props} />}
```

src/features/content-explorer/item-list/ItemList.js
```diff
@@ -188,8 +190,30 @@ const ItemList = ({
         };
     }
 
+    const addAutoSizer = WrappedComponent => {
+        return props => {
+            return isResponsive ? (
+                <div style={{ flex: 1 }}>
```
tjuanitas, line 196: Would it work if we pass a `style` prop to `AutoSizer`? I'm wondering if we could avoid adding this `div` if it's not necessary.

src/features/content-explorer/item-list/ItemList.js
```diff
@@ -188,8 +190,30 @@ const ItemList = ({
         };
     }
 
+    const addAutoSizer = WrappedComponent => {
```
tjuanitas, line 193: Thoughts on possibly calling this `withAutoSizer`?  I'm unsure if we tend to use `add` for components although I'm not great on knowing the specifics of when to use different prefixes

src/features/content-explorer/content-explorer-modal/ContentExplorerModal.js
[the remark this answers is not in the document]
greathmaster: Simplified and updated style to use SUIT naming.

src/features/content-explorer/content-explorer/ContentExplorer.js
greathmaster: Removed.

src/features/content-explorer/content-explorer/ContentExplorer.js
greathmaster: Simplified and updated style to use SUIT naming.

src/features/content-explorer/item-list/ItemList.js
greathmaster: Done.

src/features/content-explorer/item-list/ItemList.js
greathmaster: I tried removing the `<div>` and adding the `style` directly to `<AutoSizer />`. It seems to create a side effect of increasing the height upon screen resize:

### With the `<div>`
Move-Copy-AutoSizer-with-div

### Without the `<div>`
Move-Copy-Modal-AutoSizer-no-div

The modal is normally `300px` in height, without the `<div>` after resize, it expands the height.

src/features/content-explorer/item-list/ItemList.js
greathmaster: I went ahead and updated it to `withAutoSizer()`. Initially, I hesitated because it's an internal method and not a usable function by other components. However, it is acting like an HOC so from a code readability standpoint makes sense to use `withAutoSizer()`.

src/features/content-explorer/item-list/ItemList.js
greathmaster: Using standard `width` and `height` actually gave me linting errors:
Screen Shot 2022-07-07 at 3 52 37 PM

src/features/content-explorer/content-explorer/ContentExplorerActionButtons.js
greathmaster: Did some research on switching these two. Seems the `.status-message` is used by the MultiSelect variant of the ContentExplorer. It displays the the number of items selected.

Screen Shot 2022-07-08 at 9.25.40 AM

The other class `.content-explorer-copy-button` sets the `margin-right: 0`. Which is setting the margin on the right most item. The `.modal-actions:` already applies this as:

.modal-actions .btn:last-of-type {
    margin-right: 0;
}
from (https://github.com/box/box-ui-elements/blob/master/src/components/modal/Modal.scss#L117)

I'd consider switching to using `.modal-actions` as a "responsiveness change" since we want the buttons to be on the bottom on small screens. For that reason I've made it conditional to the `isResponsive` prop so it's use will be strictly opt-in.

src/features/content-explorer/item-list/ItemList.js
```diff
@@ -188,8 +189,31 @@ const ItemList = ({
...
+    TableComponent = withAutoSizer(TableComponent);
+
     return (
-        <div className={classNames('content-explorer-item-list table', className)}>
+        <div
+            className={classNames(
+                'content-explorer-item-list table',
+                // isResponsive ? 'content-explorer-item-list--isResponsive' : '',
+                className,
+                { 'bdl-ItemList--responsive': isResponsive },
```
tjuanitas, line 214: I've been thinking on how we want to handle this class name since it's likely `bdl-ItemList` would conflict with a similarly named component in our parent apps.

While it doesn't exactly following the convention, I think we should try `bdl-ContentExplorerItemList--responsive` with the intention of renaming this component to something else in the future.

src/features/content-explorer/item-list/ItemList.js
```diff
@@ -204,14 +203,15 @@ const ItemList = ({
...
 
-    TableComponent = addAutoSizer(TableComponent);
+    TableComponent = withAutoSizer(TableComponent);
 
     return (
         <div
             className={classNames(
                 'content-explorer-item-list table',
-                isResponsive ? 'content-explorer-item-list--isResponsive' : '',
+                // isResponsive ? 'content-explorer-item-list--isResponsive' : '',
```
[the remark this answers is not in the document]
greathmaster: Removed :)

src/features/content-explorer/item-list/ItemList.js
greathmaster: Updated to the new className.

PR #2946, opened 2022-07-09
fix(modal): Adjust modal header elements and enable scrolling action

greathmaster:
### Overview
This is a continuation of this PR: https://github.com/box/box-ui-elements/pull/2910/files

The original author is unable to continue working on the PR, so I've taken over further development. This PR incorporates the author's prior commits.

### Summary
- Adjusted styles to remove `z-index` and `position: fixed`. 
- Adjusted header and footer margins
- Now internal content can be properly scrolled

### Demo
Shows margin, padding, content and then shows scrolling action.
Modal-Header-Changes

### DIFF
--- a/src/components/modal/Modal.scss
+++ b/src/components/modal/Modal.scss
@@ -144,8 +144,23 @@
             padding: 0;
         }
 
+        .modal-header-container {
+            position: fixed;
+            top: 0;
+            left: 0;
+            z-index: 2;
+            display: flex;
+            width: 100%;
+            height: 64px;
+            background-color: $white;
+        }
+
+        .modal-header {
+            padding: 20px 16px;
+        }
+
         .modal-content {
-            padding-bottom: 60px;
+            margin: 60px 0;
         }
 
         .modal-dialog {
@@ -155,9 +170,9 @@
         }
 
         .modal-close-button {
-            top: initial;
+            top: 0;
             right: 10px;
-            line-height: 0;
+            padding: 20px 0;
         }
 
         .modal-actions {
--- a/src/components/modal/Modal.scss
+++ b/src/components/modal/Modal.scss
@@ -155,7 +155,9 @@
         }
 
         .modal-close-button {
+            top: initial;
             right: 10px;
+            line-height: 0;
         }
 
         .modal-actions {
--- a/src/components/modal/Modal.scss
+++ b/src/components/modal/Modal.scss
@@ -5,7 +5,6 @@
 /**************************************
  * Modal
  **************************************/
-
 @keyframes popup_bounce_in {
     0% {
         transform: scale3d(.1, .1, 1);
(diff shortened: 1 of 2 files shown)


src/components/modal/Modal.scss
```diff
@@ -5,6 +5,8 @@
 /**************************************
  * Modal
  **************************************/
+$responsive-modal-header-height: 64px;
+$responsive-modal-actions-height: 60px;
```
tjuanitas, line 9: Can we move this inside the `@include` block since it's specific to those styles?
https://github.com/box/box-ui-elements/blob/96ee050f3928afef1644d48b2bea296a052914cd/src/components/modal/Modal.scss#L143

src/components/modal/Modal.scss
```diff
@@ -144,25 +146,50 @@
...
         .modal-dialog {
+            height: 100%;
             min-height: 100%;
-            padding: 22px 16px 12px 16px;
+            padding: 0;
             border-radius: 0;
         }
 
         .modal-close-button {
+            top: 2px; //for visual alignment
```
tjuanitas, line 181: What are you thoughts on adding `display: flex` and `align-items: center` with `top: 0`? I think this could avoid the pixel adjustment for visual alignment.

src/components/modal/Modal.scss
```diff
@@ -144,25 +146,50 @@
...
-            padding-bottom: 60px;
+            margin-top: $responsive-modal-header-height;
+            margin-bottom: $responsive-modal-actions-height;
+            padding: 24px 16px 12px 16px;
+            overflow: auto;
         }
 
         .modal-dialog {
+            height: 100%;
             min-height: 100%;
```
tjuanitas, line 175: I think we can remove `min-height` now that we're specifying `height`.

src/components/modal/Modal.scss
```diff
@@ -144,25 +146,50 @@
...
 
+        .modal-header-container {
+            position: absolute;
+            top: 0;
+            left: 0;
+            display: flex;
+            align-items: center;
+            width: 100%;
+            height: $responsive-modal-header-height;
+            background-color: $white;
```
tjuanitas, line 157: With the added `margin-top` for `modal-content`, I don't think we need the background-color anymore since there shouldn't be anything under the header.

src/components/modal/Modal.scss
greathmaster: Done

src/components/modal/Modal.scss
greathmaster: You're right, it's not needed anymore. I've removed it.

src/components/modal/Modal.scss
greathmaster: Removed `min-height`.

src/components/modal/Modal.scss
greathmaster: ### With adding `flex` and `top: 0`
Screen Shot 2022-07-12 at 11 01 17 PM

### With `top: 2px`
Screen Shot 2022-07-12 at 11 02 49 PM

Seems like it's still slightly off. However, I'm wondering if I'm optimizing at the 1% level. I'm open to using the `flex` version. It's probably better for long term maintainability. I've made the change in the update, but let me know what you think.

tjuanitas: Ahh interesting, I didn't expect there to be differences. In this case, go with your best judgement and feel free to change it back to `top: 2px` if it works better.

I'm curious how it'll display on Windows computers. If you have Parallels, this could be a good test to help sway your decision one way or the another as it's likely to be similarly misaligned by a couple pixels. I've noticed that our 'Lato' font is rendered differently by OS environment and has additional bottom padding on Windows computers.

For example:

**Chrome on macOS**
Screen Shot 2022-07-13 at 10 17 06 AM

**Chrome on Windows**
Screen Shot 2022-07-13 at 10 16 39 AM

tjuanitas (approved):
Looks good! Feel free to adjust the `top: 2px` for visual alignment if you think it works better a different way.

src/components/modal/Modal.scss
```diff
@@ -5,7 +5,6 @@
 /**************************************
  * Modal
  **************************************/
-
```
tjuanitas, line 8: nit: there's a removed line break but it's minor and can be merged as-is so only address this if you end up making a fix for something else.

PR #2950, opened 2022-07-13
feat(comment): added Comment component (WIP)

kkuliczkowski-box:
Added `Comment` component, which is one of the set of components that are being created for _Threaded Replies_ project.
**This component is work in progress.**

Main goal of this PR is the creation of `ActionDropdown` component.

Further PRs will enhance the `Comment` component.


src/components/comment/Comment.tsx
```diff
@@ -0,0 +1,81 @@
...
+    ];
+
+    return (
+        <Media className={classNames('bdl-ThreadedComment', className)}>
+            <Media.Figure>
+                <Avatar size="large" />
+            </Media.Figure>
+            <Media.Body>
+                <ActionDropdown items={actions} />
+                <div className="bdl-ThreadedComment-author">John Doe</div>
```
MateuszMamczarz, line 74: I know it's still WIP, but do we want to leave it this way or use props instead?

src/components/comment/Comment.scss
```diff
@@ -0,0 +1,3 @@
+.bdl-ThreadedComment-author {
+    font-weight: bold;
+}
```
jankowiakdawid, line 3: This class is duplicated in `ActionDropdown.scss`. Do we need to do that?

kkuliczkowski-box: Yup, that's a mistake :)

src/components/comment/Comment.tsx
```diff
@@ -0,0 +1,81 @@
...
+    userCanResolve?: boolean;
+    userId: string;
+    userName: string;
+}
+
+function Comment({ className }: CommentProps) {
+    const actions: ActionDropdownItem[] = [
+        {
+            icon: <Checkmark16 />,
+            text: 'Resolve',
```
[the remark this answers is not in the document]
kkuliczkowski-box: Sure, this will be added as a proper i18n message but in the later stages.

src/components/comment/Comment.tsx
kkuliczkowski-box: Fixed

src/components/comment/ActionDropdown.tsx
```diff
@@ -0,0 +1,36 @@
...
+    items: ActionDropdownItem[];
+}
+
+function ActionDropdown({ items }: ActionDropdownProps) {
+    if (!items.length) {
+        return null;
+    }
+
+    return (
+        <Media.Menu data-testid="ThreadedComment-actionDropdown" className="bdl-ThreadedComment-actionDropdown">
```
jstoffan, line 23: SUIT naming is component-specific, so this would be `bdl-ActionDropdown`.

src/components/comment/ActionDropdown.scss
```diff
@@ -0,0 +1,3 @@
+.bdl-ThreadedComment-actionDropdown .icon-ellipsis {
```
jstoffan, line 1: It looks like the intention is for this component to not be aware of its children, but this style breaks that encapsulation.

jstoffan (changes requested):
How does this differ from the existing `Comment` component used by the Activity Sidebar? How will we resolve the discrepancy in experience and capabilities between those two?

src/components/comment/Comment.tsx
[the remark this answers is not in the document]
kkuliczkowski-box: Done

src/components/comment/Comment.tsx
```diff
@@ -0,0 +1,81 @@
...
+        },
+    ];
+
+    return (
+        <Media className={classNames('bdl-ThreadedComment', className)}>
+            <Media.Figure>
+                <Avatar size="large" />
+            </Media.Figure>
+            <Media.Body>
+                <ActionDropdown items={actions} />
```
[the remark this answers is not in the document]
kkuliczkowski-box: I wanted ActionDropdown to be a part of Comment component, so I've know refactored how it's used and moved the logic behind checking which menu items to display to ActionDropdown.

src/components/comment/Comment.tsx
```diff
@@ -0,0 +1,81 @@
...
+            text: 'Delete',
+            onClick: () => {
+                // eslint-disable-next-line no-console
+                console.log('delete');
+            },
+        },
+    ];
+
+    return (
+        <Media className={classNames('bdl-ThreadedComment', className)}>
```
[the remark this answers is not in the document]
kkuliczkowski-box: - className fixed
- I've seen that `classNames` utility has been used widely in BUIE and here I want to give the ability to add custom classes. Most probably in the future there will be other classes dynamically added based on props (still WIP).

src/components/comment/ActionDropdown.scss
kkuliczkowski-box: I've refactored the `ActionDropdown` and it's treated as part of the `Comment` component broken up from it. And the button with ellipsis icon is a static part of it that needed a slight design change.

src/components/comment/ActionDropdown.tsx
kkuliczkowski-box: From the Comment's point of view the `ActionDropdown` is just a part of it. It's not a separate component and will not be used outside of `Comment`.

src/components/comment/ActionDropdown.tsx
```diff
@@ -0,0 +1,48 @@
...
+    onResolve?: (commentId: string) => void;
+}
+
+function ActionDropdown({ commentId, onDelete, onEdit, onResolve }: ActionDropdownProps) {
+    if (!onDelete && !onEdit && !onResolve) {
+        return null;
+    }
+
+    return (
+        <Media.Menu data-testid="Comment-actionDropdown" className="bdl-Comment-actionDropdown">
```
jstoffan, line 25: SUIT naming is largely concerned with defining hierarchy and scope. Here, we'd want to use `bdl-ActionDropdown` or rename the component to `CommentActionDropdown` and use `bdl-CommentActionDropdown`.

src/components/comment/ActionDropdown.tsx
```diff
@@ -0,0 +1,48 @@
...
+            {onResolve && (
+                <MenuItem data-testid="ActionDropdownItem-resolve" onClick={() => onResolve(commentId)}>
```
jstoffan, line 27: Do we need `data-testid` attributes for each of these elements? It seems like we should be able to select them based on their user-facing text.

src/components/comment/Comment.scss
```diff
@@ -0,0 +1,18 @@
+@import '../../styles/variables';
+
+.bdl-Comment {
+    font-size: .875rem;
```
jstoffan, line 4: We don't typically use `rem` for our font sizes. Can we use the `bdl-fontSize` variables?

src/components/comment/Comment.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+import './Comment.scss';
+
+export enum CommentType {
+    COMMENT = 'comment',
+    ANNOTATION = 'annotation',
+}
+
+export interface CommentMention {
+    userName: string;
```
jstoffan, line 15: Can we alphabetize the attributes of our types and props?

src/components/comment/Comment.tsx
```diff
@@ -0,0 +1,56 @@
...
+    return (
+        <div className={classNames('bdl-Comment', className)}>
+            <Media>
+                <Media.Figure>
+                    <Avatar className="bdl-Comment-avatar" />
+                </Media.Figure>
+                <Media.Body>
+                    <ActionDropdown commentId={id} onDelete={onDelete} onEdit={onEdit} onResolve={onResolve} />
+                    <div className="bdl-Comment-author">{userName}</div>
+                    <div className="bdl-Comment-date">Today at 9:42 AM</div>
```
jstoffan, line 48: I assume this is placeholder?

src/components/comment/ActionDropdown.scss
```diff
@@ -0,0 +1,3 @@
+.bdl-Comment-actionDropdown .icon-ellipsis {
```
jstoffan, line 1: Where is `icon-ellipsis` used?

src/components/comment/ActionDropdown.tsx
kkuliczkowski-box: I've renamed it to `CommentActionDropdown` with appropriate className.

src/components/comment/ActionDropdown.tsx
kkuliczkowski-box: That's right. I've removed `data-testid` attributes and updated tests.

src/components/comment/Comment.scss
kkuliczkowski-box: Sure, I haven't noticed `bdl-fontSize` variables. There wasn't one for 12 px, so I've created it. I'm not sure if there is any particular naming convention though (I've named it `$bdl-fontSize-medium`).

src/components/comment/Comment.tsx
kkuliczkowski-box: Done

src/components/comment/Comment.tsx
kkuliczkowski-box: Correct

src/components/comment/ActionDropdown.scss
kkuliczkowski-box: It's the icon used in `MediaMenu` component.
I could theoretically extend `MediaMenu` with optional prop of icon size and modify it directly there but I'm not sure if that's worth doing for just one use case.

kkuliczkowski-box: Done

jstoffan (approved):
LGTM, but please update the PR title before merging.

PR #2952, opened 2022-07-20
feat(dropdown-menu): Add responsive dropdown menu behaviors + menu header composability

ivanthai:
### Changes
- Responsive behaviors to `DropdownMenu` by adding prop `isResponsive = true` (opt-in)
- `MenuHeader` as a new child component under `Menu` to improve composability
- `ScrollContainer` full-screen support under `Menu`
- `Submenu` appears on top and anchored to the bottom of the triggering element

 ### Notes about this PR
- Most of changes worth reviewing are in the **menu** module and `DropdownMenu` scss
- About half the files are updated jest snapshots
- Half of the menu components are just corresponding flow files

### Context
After building different prototypes of the responsive version of `DropdownMenu` including a prototype that used `Flyout`, I landed on adding customized responsive behaviors for `DropdownMenu`. `MenuHeader` and `MenuContext` are new components that are similar to `OverlayHeader` and `FlyoutContext` respectively, but catered specifically for the responsive menu use case. This not only improves composability with `Menu` elements, but also simplifies the implementation compared to building a wrapping component or using `Flyout`/`Overlay`. Using `Flyout` and `Overlay` complicated a lot of things due to the custom functionality of `DropdownMenu`

### Current limitations - Tabbing navigation to `MenuHeader`
Tabbing navigation from the `Menu` -> `MenuHeader` elements (e.g. `CloseButton` at the top right) is not functional. However, tabbing, pressing esc, and clicking the `CloseButton` while in the `Menu` all have similar outcomes which is to close the `Menu` so this limitation should be acceptable. Upcoming work will look at adding support for tab navigation to the `MenuHeader` which may include buttons, but we may decide not to include additional buttons in favor of adding it as menu items instead.

### GIF Demos
**Full-screen `DropdownMenu` (`isResponsive = true`)**

responsive-dropdown-menu-demo

**Keyboard navigation**

responsive-dropdown-menu-demo-keyboard-nav

**Dynamic resizing**

responsive-dropdown-menu-demo-resizer

**Scrolling (Extra menu item padding added to trigger scroll for this gif)**

responsive-dropdown-demo-scroll

**`Submenu`**

responsive-dropdown-menu-demo-submenu

**`MenuHeader` <> `Menu` Composability**

image


src/components/dropdown-menu/DropdownMenu.scss
```diff
@@ -17,6 +17,33 @@
 
     z-index: $menu-z-index;
     margin-top: 5px;
-    perspective: 500px;
-    perspective-origin: center top;
```
ivanthai, line 21: Removed this because it is unnecessary and conflicts with the responsive styling

src/components/dropdown-menu/DropdownMenu.scss
```diff
@@ -17,6 +17,50 @@
...
+}
+
+@include breakpoint($medium-screen) {
+    .bdl-DropdownMenu--responsive {
+        &.dropdown-menu-element {
+            right: 0;
+            bottom: 0;
+            left: 0;
```
greathmaster, line 27: What is the reason for including both `right` and `left` here?

src/components/dropdown-menu/DropdownMenu.scss
```diff
@@ -17,6 +17,50 @@
...
+            margin-top: 0;
+        }
+
+        &.dropdown-menu-enabled {
+            // cancels out the tether inline styling without having to set enabled=false
+            transform: none !important;
```
greathmaster, line 33: Can you elaborate on the alternative here? ie setting `enabled=false`?

src/components/dropdown-menu/DropdownMenu.scss
```diff
@@ -17,6 +17,50 @@
...
+            margin: 0;
+            padding-top: $bdl-header-height + $bdl-grid-unit * 3;
+            background-color: $bdl-gray-02;
+            border: none;
+            border-radius: 0;
+            box-shadow: none;
+
+            .scroll-container {
+                height: calc(100vh - #{$bdl-header-height});
+                max-height: calc(100vh - #{$bdl-header-height});
```
greathmaster, line 47: Do we need `max-height` here if it's being set to the same value as `height`?

src/components/menu/Menu.tsx
```diff
@@ -199,7 +201,7 @@ class Menu extends React.Component<MenuProps> {
         const { menuItemEl }: { menuItemEl?: HTMLElement | null } =
             event.target instanceof Node ? this.getMenuItemElFromEventTarget(event.target) : {};
 
-        if (!menuItemEl) {
+        if (!menuItemEl || (event.target as Element).className.includes('submenu-target')) {
```
greathmaster, line 204: Can you elaborate on what this checking? Checking the `className` seems fragile to me, but wondering if it's okay here if we are not going to change it.

src/components/menu/MenuHeader.scss
```diff
@@ -0,0 +1,19 @@
+@import '../../styles/variables';
+
+.bdl-MenuHeader {
+    position: absolute;
+    top: 0;
+    right: 0;
+    left: 0;
```
greathmaster, line 7: Similar to the above. Do we need to use both `right` and `left`?

src/components/dropdown-menu/DropdownMenu.scss
ivanthai: yeah, kinda pointless. I could have instead unset max-height

src/components/menu/Menu.tsx
ivanthai: This is to not close the menu when the menu element is the triggering submenu element. This is not okay on mobile since you don't have hover events, just touch events.

src/components/dropdown-menu/DropdownMenu.scss
ivanthai: This is a workaround to cancel out the styling in CSS. The tether library applies a transform override with `!important `which makes it impossible to override without specifying another `!important` override. This limited the JS changes I needed to make which I felt was cleaner than to have to use the media query library for that one change

src/components/dropdown-menu/DropdownMenu.scss
ivanthai: To enforce left and right bounds of the container.  The alternative is to add it to aria-menu instead

src/components/menu/Menu.tsx
ivanthai: I mean, I think there is probably a better solution here...I'll take a look at this again.

src/components/menu/MenuHeader.scss
ivanthai: No, but I think I left both in there on accident

ivanthai: Oops, I meant to say I do need both because the element does not fully span the full width. There is the option of adding `width: 100%`, but I think that comes with other issues.

src/components/dropdown-menu/DropdownMenu.scss
ivanthai: Moved this CSS styling to `aria-menu` to `display: fixed` like we do in `Flyout`

src/components/menu/Menu.tsx
ivanthai: I ended up moving this logic to `SubmenuItem`. Basically if the submenu triggering element is being clicked, prevent propagation/default

src/components/menu/Menu.scss
```diff
@@ -28,6 +28,10 @@
     .lnk:focus {
         text-decoration: none;
     }
+
+    .bdl-MenuHeader {
+        display: none;
+    }
```
tjuanitas, line 34: When is `MenuHeader` visible?

src/components/menu/MenuHeader.scss
```diff
@@ -0,0 +1,19 @@
+@import '../../styles/variables';
+
+.bdl-MenuHeader {
+    position: absolute;
```
tjuanitas, line 4: What ancestor element are we positioning this to? Can we rely on that position?

greathmaster (commented):
I think after addressing the points @tjuanitas has brought up, we should be good. 

Also, general question...has the functionality been tested in all the browsers?

Additionally, I think in a separate PR we should show how to use it in the Storybook documentation.

src/components/menu/Menu.scss
ivanthai: At the moment, DropdownMenu

src/components/menu/MenuHeader.scss
ivanthai: Yes, since MenuHeader is used only in full-screen mode. I moevd it into the DropdownMenu.scss instead to restrict it to that particular component

src/components/dropdown-menu/DropdownMenu.scss
```diff
@@ -17,6 +17,41 @@
...
+
+@include breakpoint($medium-screen) {
+    .bdl-DropdownMenu--responsive {
+        &.dropdown-menu-enabled {
+            // cancels out the tether inline styling without having to set enabled=false
+            transform: none !important;
+        }
+
+        .aria-menu:not(.submenu) {
+            position: fixed;
```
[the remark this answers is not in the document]
ivanthai: Yeah, fixed this

src/components/menu/MenuHeader.scss
```diff
@@ -0,0 +1,11 @@
+@import '../../styles/variables';
+
+.bdl-MenuHeader {
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
```
tjuanitas, line 6: Does this `justify-content: space-between` have an effect?

I'm thinking that because `.bdl-MenuHeader-content` has `flex-grow: 1`, that this possibly wouldn't do anything because there would be no space between `.bdl-MenuHeader-content` and the `CloseButton`.

src/components/menu/MenuHeader.scss
tjuanitas: Is there a use for `MenuHeader` outside of `DropdownMenu`?

In other words, if this component is specific to `DropdownMenu`, should we move it inside of that folder?

ivanthai: At the moment, I'm not 100% sure where else we will be making changes. ContextMenu which is used in a similar context to DropdownMenu in EUA also uses Menu and may need the MenuHeader. This is something we might look at next quarter.

src/components/menu/MenuHeader.scss
ivanthai: yeah you're right. I just tried it and it's removed now

tjuanitas (approved):

greathmaster (dismissed):

Machoper (approved):

ConradJChan (commented):
Curious about the submenu - seems like having the submenu render as normal might be limiting in terms of having nested submenus. Would it be more consistent if the submenu was stretched to cover the viewport as well?

ivanthai:
@ConradJChan at the moment, nested submenu's are not supported and there weren't plans on doing so at the time of the original implementation. There currently isn't any use case for nest submenus

ConradJChan (approved):

Machoper (approved):

PR #2954, opened 2022-07-21
feat(button): add responsive behavior to button

Machoper:
Allow parents that consume the Button component to configure the responsive behavior, so that for screen sizes less than 768px:

- If **isResponsive=false** OR **icon is undefined**, button content will not change (see share button below)
nonresponsive_or_no_responsivecontent

- If **isResponsive=true** AND **icon is defined**, button content will change to the icon set (see share button below)
text + icon -> icon (**showIcon=true**)
text_and_icon_to_icon
text -> icon (**showIcon=false**)
text_to_icon

### DIFF
--- a/src/components/button/Button.tsx
+++ b/src/components/button/Button.tsx
@@ -21,6 +21,8 @@ export interface ButtonProps {
     isDisabled?: boolean;
     /** whether the button is loading or not */
     isLoading: boolean;
+    /** toggles responsive behavior */
+    isResponsive?: boolean;
     /** whether the button is selected or not */
     isSelected?: boolean;
     /** onClick handler for the button */
@@ -29,6 +31,8 @@ export interface ButtonProps {
     setRef?: Function;
     /** size of the button */
     size?: 'large';
+    /** whether to show icon when the button is not responsive or on large screens */
+    showIcon?: boolean;
     /** whether to show a radar */
     showRadar: boolean;
     /** type for the button */
@@ -41,6 +45,8 @@ class Button extends React.Component<ButtonProps> {
     static defaultProps = {
         className: '',
         isLoading: false,
+        isResponsive: false,
+        showIcon: true,
         showRadar: false,
         type: ButtonType.SUBMIT,
     };
@@ -64,10 +70,12 @@ class Button extends React.Component<ButtonProps> {
             icon,
             isDisabled,
             isLoading,
+            isResponsive,
             isSelected,
             setRef,
             size,
             type,
+            showIcon,
             showRadar,
             ...rest
         } = this.props;
@@ -83,7 +91,8 @@ class Button extends React.Component<ButtonProps> {
                 'is-loading': isLoading,
                 'is-selected': isSelected,
                 'bdl-btn--large': size === 'large',
-                'bdl-has-icon': !!icon,
+                'bdl-show-icon': !!icon && showIcon,
+                'bdl-Button--responsive': isResponsive && !!icon,
             },
             className,
         );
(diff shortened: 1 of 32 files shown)


src/components/button/Button.tsx
```diff
@@ -21,10 +21,14 @@ export interface ButtonProps {
...
     /** whether the button is loading or not */
     isLoading: boolean;
+    /** toggles responsive behavior */
+    isResponsive?: boolean;
     /** whether the button is selected or not */
     isSelected?: boolean;
     /** onClick handler for the button */
     onClick?: Function;
+    /** component to show when breakpoint is reached */
+    responsiveContent?: React.ReactNode;
```
tjuanitas, line 31: Does it work to combine the component from this prop with `children`?

tjuanitas: i.e. in the parent of this component, it's setting `isResponsive` and `responsiveContent`. Does it make sense that when the values of `isResponsive` and `responsiveContent` is supposed to change, it will instead swap the components passed through `children`?

This way we won't need to update separate props.

Machoper: Thanks for reviewing! Not sure I understand correctly here, but if we were to combine `responsiveContent` with `children` we would probably need to use something like `useMediaQuery`, which might not be appropriate as I haven't seen it being used anywhere else in this repo. Or we could simply discard this change and let parents fully control what they want to show on different screen sizes by passing in dynamic `children`.
But for now, I've removed `responsiveContent` and reused the existing `icon`. Also made it backward compatible, so it wouldn't affect existing parents as long as `isResponsive` is false.

src/components/button/Button.tsx
```diff
@@ -29,6 +31,8 @@ export interface ButtonProps {
     setRef?: Function;
     /** size of the button */
     size?: 'large';
+    /** whether to show icon when the button is not responsive or on large screens */
+    showIcon?: boolean;
```
ivanthai, line 35: This works but is a little awkward because it seems to conflict with `isResponsive`. Mostly because this has no effect if `showIcon` is false and `isResponsive` is true. I wonder if it would be appropriate to have a fallback element prop..but I suppose that was what `responsiveContent` was for which I believe we moved away from

ivanthai (approved):
Works for me except the awkwardness I mentioned. Acceptable to me though

src/styles/common/_buttons.scss
```diff
@@ -198,3 +202,15 @@ button svg {
...
+    .btn.bdl-Button--responsive {
+        .btn-content {
+            display: none;
+        }
+
+        .bdl-btn-icon {
+            display: block;
+        }
+    }
+}
```
tjuanitas, line 216: Do we know how many buttons in EUA will use these new props and styles?

What are your thoughts on putting the styles in the parent component like this?
```
// ParentButton.scss

.ParentButton {
    .bdl-btn-icon {
        display: none;
    }
}

.is-responsive-web {
        .ParentButton {
            .btn-content {
                display: none;
            }

            .bdl-btn-icon {
                display: block;
            }
        }
    }
}
```

This way you wouldn't need to pass additional props to the component.

src/components/button/Button.tsx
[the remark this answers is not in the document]
Machoper: agreed. Can't really think of a better way tho.

src/styles/common/_buttons.scss
Machoper: AFAIK only the share button and potentially the create new item button will use these new props and styles, so I think it makes sense to put the styles in parent components.

Machoper:
Since there's really not an easy way to add a generalized yet customizable responsive behavior to `Button` without introducing any breaking changes or weirdness, I will just close this PR and make responsive changes in EUA instead.

PR #2956, opened 2022-07-26
feat(content-explorer): Add responsive behavior

greathmaster:
#### Overview
Note: This fixes the original commit: https://github.com/box/box-ui-elements/pull/2932.

That was reverted in the commit: https://github.com/box/box-ui-elements/pull/2955

#### Reason for revert
The original commit introduced a bug that caused folders with many subfolders or items to behave incorrectly. When an item was selected, and the user scrolled down until the selected item was no longer visible and then selected another item, the file/folder list would scroll back to the top of the file/folder list.

#### Solution
This PR reintroduces the prior commit, but refactors how the `AutoSizer` component is added to `ItemList` component. It moves the HOC from being an internal function, to being an external one.

#### Commits
* feat(content-explorer): Add responsive behavior

* feat(content-explorer): Add responsive behavior revisions


src/features/content-explorer/item-list/ItemList.js
```diff
@@ -188,9 +203,30 @@ const ItemList = ({
...
+    //                     {({ width: w, height: h }) => <WrappedComponent {...props} width={w} height={h} />}
+    //                 </AutoSizer>
+    //             </div>
+    //         ) : (
+    //             <WrappedComponent {...props} />
+    //         );
+    //     };
+    // };
+
+    // TableComponent = withAutoSizer(TableComponent);
```
tjuanitas, line 222: Let's remove this before we merge

tjuanitas (commented):
Looks good but there's some debugging code we can remove.

src/features/content-explorer/item-list/ItemList.js
greathmaster: Oops...Sorry about that. Fixed.

tjuanitas (approved):

PR #2963, opened 2022-08-03
fix(dropdown-menu): fix menu header close btn and text positioning

ivanthai:
* CloseButton is now properly anchored to the top right as opposed it being pushed off screen with a long title
* MenuHeader now takes in title and subtitle props
* Title element will truncate overflow text with ellipsis

image


src/components/menu/MenuHeader.tsx
```diff
@@ -23,8 +25,11 @@ const MenuHeader = ({ children, className, ...rest }: MenuHeaderProps) => {
             role="presentation"
             {...rest}
         >
-            <div className="bdl-MenuHeader-content">{children}</div>
-            <CloseButton onClick={closeMenu} />
+            <div className="bdl-MenuHeader-content">
+                <div className="bdl-MenuHeader-title">{title}</div>
+                <div className="bdl-MenuHeader-subtitle">{subtitle}</div>
```
Machoper, line 30: I think it's better to add null/undefined checks for both `title` and `subtitle` here.

src/components/menu/MenuHeader.tsx
```diff
@@ -7,13 +7,15 @@ import MenuContext from './MenuContext';
 import './MenuHeader.scss';
 
 export interface MenuHeaderProps {
-    /** children - menu section header content */
-    children?: Array<React.ReactChild> | React.ReactChild;
```
Machoper, line 11: Might be a good idea to keep `children` as well in case we need to customize the menu header in the future.

src/components/menu/MenuHeader.tsx
ivanthai: agreed.

Machoper (approved):

PR #2964, opened 2022-08-03
fix(button): diabled button with theme shows custom color in all states

wt-lin:
## Description
- When custom theme is provided to Button, the disabled Button still shows default blue in different status, e.g., hover, active.
- From the [scss file](https://github.com/box/box-ui-elements/blob/master/src/styles/common/_buttons.scss#L175), if a button has no `is-disabled` or `bdl-is-disabled` class, the original primary color will be applied.

## Screenshots
11b3547c6e51e91a3ec83d2a24a27f53


src/components/button/Button.tsx
```diff
@@ -82,6 +82,7 @@ class Button extends React.Component<ButtonProps> {
                 'is-disabled': isDisabled,
                 'is-loading': isLoading,
                 'is-selected': isSelected,
+                'bdl-is-disabled': isDisabled,
```
greg-in-a-box, line 85: Can you elaborate further on why we need a second `is-disabled` classname?

greg-in-a-box (commented):
Left a comment

src/components/button/Button.tsx
wt-lin: I checked the commit log, and found this.
https://github.com/swfree/box-ui-elements/pull/2

A "bdl-" namespace was added to solve the conflicting classnames.
There's a [script](https://github.com/box/box-ui-elements/blob/master/scripts/utils/bdlClassnameManager) for this, but it doesn't cover all the case.

PR #2969, opened 2022-08-03
chore(storybook): Add DatalistItem story

tjuanitas:
Screen Shot 2022-08-03 at 1 49 59 PM


src/components/datalist-item/DatalistItem.stories.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react';
+
+// @ts-ignore JS Import
```
greg-in-a-box, line 3: I am wondering if we still need this `ts-ignore` or not

greg-in-a-box (approved):

PR #2970, opened 2022-08-03
chore(storybook): add story for Flyout

benjamin-shen:
image
image

image
image

image
image


src/components/flyout/Flyout.stories.tsx
```diff
@@ -0,0 +1,98 @@
...
+import Button from '../button';
+import IconHelp from '../../icons/general/IconHelp';
+// @ts-ignore JS import
+import TextInput from '../text-input';
+import PlainButton from '../plain-button';
+import PrimaryButton from '../primary-button';
+// @ts-ignore JS import
+import TextArea from '../text-area';
+
+// @ts-ignore JS import
```
greg-in-a-box, line 13: Check if we need these lines in the code or not, if you dont see any errors or warnings, remove them

greg-in-a-box (changes requested):

src/components/flyout/Flyout.stories.tsx
benjamin-shen: There is an error that looks like this
image

greg-in-a-box (approved):

PR #2972, opened 2022-08-03
fix(button): diabled button with theme shows custom color in all states

wt-lin:
## Description
- When custom theme is provided to Button, the disabled Button still shows default blue in different status, e.g., hover, active.
- From the [scss file](https://github.com/box/box-ui-elements/blob/master/src/styles/common/_buttons.scss#L175), if a button has no `is-disabled` or `bdl-is-disabled` class, the original primary color will be applied.

## Screenshots
11b3547c6e51e91a3ec83d2a24a27f53


src/components/button/Button.tsx
```diff
@@ -82,6 +82,7 @@ class Button extends React.Component<ButtonProps> {
                 'is-disabled': isDisabled,
                 'is-loading': isLoading,
                 'is-selected': isSelected,
+                'bdl-is-disabled': isDisabled,
```
wt-lin, line 85: @greg-in-a-box

Regarding your question,
I checked the commit log, and found this.
https://github.com/swfree/box-ui-elements/pull/2

A "bdl-" namespace was added to solve the conflicting classnames.
There's a [script](https://github.com/box/box-ui-elements/blob/master/scripts/utils/bdlClassnameManager) for this, but it doesn't cover all cases.

I was struggling to replace `is-disabled` with `bdl-is-disabled`, but it would be a breaking change here. Or should I just remove [bdl-is-disabled](https://github.com/box/box-ui-elements/blob/master/src/styles/common/_buttons.scss#L175)? What do you think?

tjuanitas: I don't think you would want to straight replace `is-disabled` unless you were going to immediately update the parents apps that reference this class.

E.g. EUA, Admin Console, etc. have SCSS files that use this class, so that would break their UI when you bump the BUIE version in those apps.

greg-in-a-box (approved):
Coexistence of `is-disabled` and `bdl-is-disabled` should be fine for now till a migration occurs

PR #2974, opened 2022-08-04
fix(menu-item): fix menu item disabled tabbing

karelee7:
Fix disabled menu item by making it reachable via tabbing

**Text read by the voiceover:**
> “Download PDF, menu item”
> “Download PDF is not supported by this file”

### DIFF
--- a/src/components/menu/Menu.scss
+++ b/src/components/menu/Menu.scss
@@ -50,7 +50,7 @@
         font-size: 10px;
     }
 
-    &[aria-disabled] {
+    &[aria-readonly] {
         cursor: default;
         opacity: .4;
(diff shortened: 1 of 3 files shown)


ConradJChan (commented):
Do all MenuItem have a tooltip? Potentially there could be a use case to have aria-disabled as well when there is no tooltip and we don't want the MenuItem to be tabbable

karelee7:
Closed because unnecessary; `ctrl + option + (right) arrow is actually the 'next item' command in voiceover, which is the primary command used to traverse the page. that reads both static and interactive content (unlike the tab key, which only reads interactive content)`

PR #2980, opened 2022-08-10
fix(mediaquery): change type of size to string

ahorowitz123:
BREAKING CHANGE: the previous type was boolean. This could break type
checkers


ivanthai (approved):
thanks for fixing this. We've been referring to size as string as well in other checks in EUA

joshmarnold (approved):

PR #2982, opened 2022-08-11
feat(replies): Update Annotations API to support threaded replies

wpiesiak:
Modify Annotations (src/api/Annotations.js) to implement Threaded Replies requirements

- modify getAnnotations function
New param: `shouldFetchReplies?: boolean`
if true add url param `fields=replies` 

- modify getAnnotation function
New param: `shouldFetchReplies?: boolean`
if true add url param `fields=replies` 

- modify updateAnnotation function
New param status?: OPEN|RESOLVED|DELETED
Include status field in PUT request

- create getAnnotationReplies function
Returns an array of replies for a given annotation id.
 
- create createAnnotationReply function
Creates a reply for an annotation
Request body:
```
{ message: “reply message” }
```

### DIFF
--- a/src/api/Annotations.js
+++ b/src/api/Annotations.js
@@ -2,20 +2,25 @@
 import merge from 'lodash/merge';
 import {
     ERROR_CODE_CREATE_ANNOTATION,
+    ERROR_CODE_CREATE_REPLY,
     ERROR_CODE_DELETE_ANNOTATION,
     ERROR_CODE_EDIT_ANNOTATION,
     ERROR_CODE_FETCH_ANNOTATION,
     ERROR_CODE_FETCH_ANNOTATIONS,
+    ERROR_CODE_FETCH_REPLIES,
     PERMISSION_CAN_CREATE_ANNOTATIONS,
     PERMISSION_CAN_DELETE,
     PERMISSION_CAN_EDIT,
     PERMISSION_CAN_VIEW_ANNOTATIONS,
+    PERMISSION_CAN_RESOLVE,
 } from '../constants';
 import MarkerBasedApi from './MarkerBasedAPI';
+
 import type {
     Annotation,
     AnnotationPermission,
     Annotations as AnnotationsType,
+    AnnotationStatus,
     NewAnnotation,
 } from '../common/types/annotations';
 import type { BoxItemPermission } from '../common/types/core';
@@ -28,6 +33,10 @@ export default class Annotations extends MarkerBasedApi {
 
     getUrlForId(annotationId: string) {
         return `${this.getUrl()}/${annotationId}`;
+    }
+
+    getUrlWithRepliesForId(annotationId: string) {
+        return `${this.getUrlForId(annotationId)}/replies`;
     }
 
     createAnnotation(
@@ -74,24 +83,39 @@ export default class Annotations extends MarkerBasedApi {
         fileId: string,
         annotationId: string,
         permissions: AnnotationPermission,
-        message: string,
+        payload: { message?: string, status?: AnnotationStatus },
         successCallback: (annotation: Annotation) => void,
         errorCallback: (e: ElementsXhrError, code: string) => void,
     ): void {
         this.errorCode = ERROR_CODE_EDIT_ANNOTATION;
+        const { message, status } = payload;
 
-        try {
-            this.checkApiCallValidity(PERMISSION_CAN_EDIT, permissions, fileId);
-        } catch (e) {
-            errorCallback(e, this.errorCode);
-            return;
+        if (message) {
+            try {
+                this.checkApiCallValidity(PERMISSION_CAN_EDIT, permissions, fileId);
+            } catch (e) {
+                errorCallback(e, this.errorCode);
+                return;
+            }
         }
 
-        const requestData = { data: { description: { message } } };
+        if (status) {
+            try {
+                this.checkApiCallValidity(PERMISSION_CAN_RESOLVE, permissions, fileId);
+            } catch (e) {
+                errorCallback(e, this.errorCode);
+                return;
+            }
+        }
 
         this.put({
             id: fileId,
-            data: requestData,
+            data: {
+                data: {
+                    description: message ? { message } : undefined,
+                    status,
+                },
+            },
             errorCallback,
             successCallback,
             url: this.getUrlForId(annotationId),
@@ -128,8 +152,71 @@ export default class Annotations extends MarkerBasedApi {
         permissions: BoxItemPermission,
         successCallback: (annotation: Annotation) => void,
         errorCallback: (e: ElementsXhrError, code: string) => void,
+        shouldFetchReplies?: boolean,
     ): void {
         this.errorCode = ERROR_CODE_FETCH_ANNOTATION;
+
+        try {
+            this.checkApiCallValidity(PERMISSION_CAN_VIEW_ANNOTATIONS, permissions, fileId);
+        } catch (e) {
+            errorCallback(e, this.errorCode);
+            return;
+        }
+
+        const requestData = shouldFetchReplies ? { params: { fields: 'replies' } } : undefined;
+
+        this.get({
+            id: fileId,
+            errorCallback,
+            successCallback,
+            url: this.getUrlForId(annotationId),
+            requestData,
+        });
+    }
+
+    getAnnotations(
+        fileId: string,
+        fileVersionId?: string,
+        permissions: BoxItemPermission,
+        successCallback: (annotations: AnnotationsType) => void,
+        errorCallback: (e: ElementsXhrError, code: string) => void,
+        limit?: number,
+        shouldFetchAll?: boolean,
+        shouldFetchReplies?: boolean,
+    ): void {
+        this.errorCode = ERROR_CODE_FETCH_ANNOTATIONS;
+
+        try {
+            this.checkApiCallValidity(PERMISSION_CAN_VIEW_ANNOTATIONS, permissions, fileId);
+        } catch (e) {
+            errorCallback(e, this.errorCode);
+            return;
+        }
+
+        const requestData = {
+            file_id: fileId,
+            file_version_id: fileVersionId,
+            ...(shouldFetchReplies ? { fields: 'replies' } : null),
+        };
+
+        this.markerGet({
+            id: fileId,
+            errorCallback,
+            limit,
+            requestData,
+            shouldFetchAll,
+            successCallback,
+        });
+    }
+
+    getAnnotationReplies(
+        fileId: string,
+        annotationId: string,
+        permissions: BoxItemPermission,
+        successCallback: (annotation: Annotation) => void,
+        errorCallback: (e: ElementsXhrError, code: string) => void,
+    ): void {
+        this.errorCode = ERROR_CODE_FETCH_REPLIES;
 
         try {
             this.checkApiCallValidity(PERMISSION_CAN_VIEW_ANNOTATIONS, permissions, fileId);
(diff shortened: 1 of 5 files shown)


src/api/Annotations.js
```diff
@@ -164,16 +187,85 @@ export default class Annotations extends MarkerBasedApi {
...
+        errorCallback,
+    }: {
+        annotationId: string,
+        errorCallback: (e: ElementsXhrError, code: string) => void,
+        fileId: string,
+        permissions: BoxItemPermission,
+        shouldFetchReplies?: boolean,
+        successCallback: (annotation: Annotation) => void,
+    }): void {
+        this.errorCode = ERROR_CODE_FETCH_ANNOTATION;
```
jstoffan, line 220: Should we have new error codes that are reply-specific?

src/api/Annotations.js
```diff
@@ -164,16 +187,85 @@ export default class Annotations extends MarkerBasedApi {
...
+        annotationId: string,
+        errorCallback: (e: ElementsXhrError, code: string) => void,
+        fileId: string,
+        message: string,
+        permissions: BoxItemPermission,
+        shouldFetchReplies?: boolean,
+        successCallback: (annotation: Annotation) => void,
+    }): void {
+        this.errorCode = ERROR_CODE_CREATE_ANNOTATION;
+        this.errorCode = ERROR_CODE_CREATE_ANNOTATION;
```
jstoffan, line 254: Duplicate statement?

src/api/Annotations.js
```diff
@@ -164,16 +187,85 @@ export default class Annotations extends MarkerBasedApi {
...
+    }: {
+        annotationId: string,
+        errorCallback: (e: ElementsXhrError, code: string) => void,
+        fileId: string,
+        message: string,
+        permissions: BoxItemPermission,
+        shouldFetchReplies?: boolean,
+        successCallback: (annotation: Annotation) => void,
+    }): void {
+        this.errorCode = ERROR_CODE_CREATE_ANNOTATION;
```
jstoffan, line 253: Should this be a reply-specific error code?

src/api/Annotations.js
```diff
@@ -30,6 +31,10 @@ export default class Annotations extends MarkerBasedApi {
         return `${this.getUrl()}/${annotationId}`;
     }
 
+    getUrlWithRepliesForId(annotationId: string) {
+        return `${this.getUrl()}/${annotationId}/replies`;
```
jstoffan, line 35: Should this use `this.getUrlForId` rather than duplicating its logic?

src/api/Annotations.js
```diff
@@ -164,16 +187,85 @@ export default class Annotations extends MarkerBasedApi {
             return;
         }
 
+        const requestData = {
+            file_id: fileId,
+            file_version_id: fileVersionId,
+            ...(shouldFetchReplies ? { fields: 'replies' } : null),
```
tjuanitas, line 193: minor: I think this could be simplified a little:
```suggestion
            ...(shouldFetchReplies && { fields: 'replies' }),
```

src/api/Annotations.js
```diff
@@ -70,14 +75,23 @@ export default class Annotations extends MarkerBasedApi {
         });
     }
 
-    updateAnnotation(
-        fileId: string,
+    updateAnnotation({
```
tjuanitas, line 78: unlikely: but can we confirm we're not using this directly in our parent apps? in case this would cause breaking changes

src/api/Annotations.js
wpiesiak: I've tried it and I am afraid flow does not like it: 
`boolean [1] is not an object.`

src/api/Annotations.js
wpiesiak: Thanks for paying attention to that - I've checked and didn't found any usage

src/api/Annotations.js
```diff
@@ -87,7 +101,13 @@ export default class Annotations extends MarkerBasedApi {
             return;
         }
 
-        const requestData = { data: { description: { message } } };
+        const requestData = { data: {} };
+
+        if (message) {
```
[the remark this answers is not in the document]
wpiesiak: It does not. I've assumed current use cases would change one or another value, but after reconsideration I think there is no reason to restrict it in that way

src/api/Annotations.js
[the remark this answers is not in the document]
wpiesiak: I found few other examples when we passing object as a parameter (eg. Comments.js). I think in this use case, with 7 parameters, including 2 optional ones, it would help with readability.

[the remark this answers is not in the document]
jstoffan: Perhaps, but we have other many other methods, even in this file (i.e., `getAnnotations`) that use quite a few parameters; we should keep our signatures consistent. Would moving `message` and `status` into a `payload` object help?

jstoffan:
@wpiesiak, we generally want the PR title and main commit message to summarize the purpose of the change. In this case, maybe `feat(replies): Update Annotations API to support threaded replies`?

src/api/Annotations.js
```diff
@@ -70,24 +78,49 @@ export default class Annotations extends MarkerBasedApi {
...
+                this.checkApiCallValidity(PERMISSION_CAN_RESOLVE, permissions, fileId);
+            } catch (e) {
+                errorCallback(e, this.errorCode);
+                return;
+            }
+        }
+
+        const requestData = {
+            data: {
+                ...(message ? { description: { message } } : null),
```
jstoffan, line 120: Minor, but I believe Axios removes properties with a value of `undefined`. Can we simplify to the below:

```
data: {
    description: message,
    status,
}

File: src/api/Annotations.js [hunk 9]
```diff
@@ -138,11 +172,14 @@ export default class Annotations extends MarkerBasedApi {
             return;
         }
 
+        const requestData = shouldFetchReplies ? { fields: 'replies' } : null;
```
jstoffan, line 175: Does this need to be `null`? We should generally only use `null` values if there is a difference between it and `undefined`.

src/api/Annotations.js
```diff
@@ -164,16 +192,84 @@ export default class Annotations extends MarkerBasedApi {
...
+        annotationId,
+        permissions,
+        successCallback,
+        errorCallback,
+    }: {
+        annotationId: string,
+        errorCallback: (e: ElementsXhrError, code: string) => void,
+        fileId: string,
+        permissions: BoxItemPermission,
+        shouldFetchReplies?: boolean,
```
jstoffan, line 222: Can this be removed if it's unused?

src/api/Annotations.js
```diff
@@ -164,16 +192,84 @@ export default class Annotations extends MarkerBasedApi {
...
-                file_id: fileId,
-                file_version_id: fileVersionId,
-            },
+            requestData,
             shouldFetchAll,
             successCallback,
         });
     }
+
+    getAnnotationReplies({
```
jstoffan, line 211: Can we use parameters rather than an object to stay consistent with the other methods in this class?

src/api/Annotations.js
```diff
@@ -164,16 +192,84 @@ export default class Annotations extends MarkerBasedApi {
...
+        successCallback,
+        errorCallback,
+        message,
+    }: {
+        annotationId: string,
+        errorCallback: (e: ElementsXhrError, code: string) => void,
+        fileId: string,
+        message: string,
+        permissions: BoxItemPermission,
+        shouldFetchReplies?: boolean,
```
jstoffan, line 255: Unused?

src/api/Annotations.js
```diff
@@ -74,24 +82,39 @@ export default class Annotations extends MarkerBasedApi {
...
+                return;
+            }
+        }
 
         this.put({
             id: fileId,
-            data: requestData,
+            data: {
+                data: {
+                    description: { message },
```
jstoffan, line 114: I didn't realize that `description` accepted an object here. We should probably do `description: message ? { message } : undefined`, instead.

src/constants.js
```diff
@@ -160,6 +160,7 @@ export const PERMISSION_CAN_SET_SHARE_ACCESS = 'can_set_share_access';
 export const PERMISSION_CAN_SHARE = 'can_share';
 export const PERMISSION_CAN_UPLOAD = 'can_upload';
 export const PERMISSION_CAN_VIEW_ANNOTATIONS = 'can_view_annotations';
+export const PERMISSION_CAN_RESOLVE = 'can_resolve';
```
jstoffan, line 163: Can we keep these permission exports alphabetized?

src/api/Annotations.js
```diff
@@ -138,11 +162,14 @@ export default class Annotations extends MarkerBasedApi {
...
         }
 
+        const requestData = shouldFetchReplies ? { fields: 'replies' } : undefined;
+
         this.get({
             id: fileId,
             errorCallback,
             successCallback,
             url: this.getUrlForId(annotationId),
+            requestData,
```
jstoffan, line 172: Have we tested all this new behavior and does this work as expected? I think we need to pass `params` here, instead.

src/api/__tests__/Annotations.test.js
```diff
@@ -102,36 +112,59 @@ describe('api/Annotations', () => {
...
+            { can_resolve: false, can_edit: true },
+        ])('should reject with an error code for calls with invalid permissions %s', permissions => {
+            const errorCallback = jest.fn();
+            const successCallback = jest.fn();
+            const payload = { message: 'hello', status: 'resolved' };
+            annotations.updateAnnotation('12345', '67890', permissions, payload, successCallback, errorCallback);
+
             expect(errorCallback).toBeCalledWith(expect.any(Error), ERROR_CODE_EDIT_ANNOTATION);
-            expect(annotations.put).not.toBeCalled();
+            expect(annotations.get).not.toBeCalled();
```
jstoffan, line 167: Should this still be `put`?

src/api/__tests__/Annotations.test.js
```diff
@@ -175,6 +208,24 @@ describe('api/Annotations', () => {
                 errorCallback,
                 successCallback,
                 url: 'https://api.box.com/2.0/undoc/annotations/abc',
+                requestData: undefined,
```
jstoffan, line 211: Is this expected?

src/api/__tests__/Annotations.test.js
```diff
@@ -222,4 +295,62 @@ describe('api/Annotations', () => {
...
+            expect(errorCallback).toBeCalledWith(expect.any(Error), ERROR_CODE_FETCH_REPLIES);
+            expect(annotations.get).not.toBeCalled();
+        });
+    });
+
+    describe('createAnnotationReply()', () => {
+        const errorCallback = jest.fn();
+        const successCallback = jest.fn();
+        const message = 'Hello';
+        test('should format its parameters and call the post method', () => {
```
jstoffan, line 334: Can we add whitespace above the test declarations?

src/api/__tests__/Annotations.test.js
wpiesiak: It is, as I did [extract requestData to a variable](https://github.com/box/box-ui-elements/pull/2982#discussion_r948145772). 
Another approach would be use previous notation for requestData in this method, to not include it in params at all. 
`...(shouldFetchReplies ? { requestData: { params: { fields: 'replies' } } } : null),`

src/api/Annotations.js
wpiesiak: Thanks! Indeed we should pass `param` here

src/api/__tests__/Annotations.test.js
wpiesiak: It is - I did extract [requestData to a variable ](https://github.com/box/box-ui-elements/pull/2982#discussion_r948145772). 
Another way would be to change notation to previous one, like: 
`...(shouldFetchReplies ? { requestData: { params: fields: 'replies' } } } : null),`
to not include requestData in get method at all

src/api/Annotations.js
jstoffan: Do we still need to update this to be `params`?

jstoffan (commented):
@wpiesiak, it's very helpful to push individual commits to address feedback, as it makes reviewing cycles much faster.

src/api/Annotations.js
wpiesiak: [get method ](https://github.com/box/box-ui-elements/blob/master/src/api/Base.js#L239 )is receiving `requestData` property, although we need to have `params` property in it to fetch additional fields

jstoffan (approved):

PR #3000, opened 2022-08-18
feat(content-sidebar): make sidebar title optional

kkuliczkowski-box:
This is a change required for Threaded Replies and will be utilized in future development.

### DIFF
--- a/src/elements/content-sidebar/ActivitySidebar.js
+++ b/src/elements/content-sidebar/ActivitySidebar.js
@@ -55,6 +55,7 @@ type ExternalProps = {
     getUserProfileUrl?: GetProfileUrlCallback,
     hasReplies?: boolean,
     hasTasks?: boolean,
+    hasTitle?: boolean,
     hasVersions?: boolean,
     onCommentCreate: Function,
     onCommentDelete: (comment: Comment) => any,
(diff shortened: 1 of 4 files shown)


ConradJChan (commented):
These changes LGTM, but curious what the overall plan for this title will be since it seems like it will need to be more configurable than just hiding or showing the title

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -116,6 +117,7 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         getAnnotationsPath: noop,
         hasReplies: false,
         hasTasks: true,
+        hasTitle: true,
```
jstoffan, line 120: I don't think we want to expose this option at the top-level. In this case, I'd suggest using the `features` prop via `isFeatureEnabled`, such as `isFeatureEnabled('activityFeed.filter.enabled')`

kkuliczkowski-box: I think that activity feed filter is not connected directly to hiding the title. I can imagine that in future when threads will be introduced in Preview title should still be visible alongside the filter. Or do you think those two (title and filter) are mutually exclusive?

I'm implementing the filter itself in a different branch and will be utilizing `features` functionality to control it.

kkuliczkowski-box: As agreed I've removed `hasTitle` prop and title visibility will be controlled with `activityFeed.filter` feature (implemented in separate branch).

jstoffan (approved):

PR #3003, opened 2022-08-22
feat(threaded-replies): create base of ActivityThread component

wpiesiak:
create base of ActivityThread component to support threads for Comments and Annotation. 

The role of this component is to group an activity (Comment or Annotation) with it’s replies

Props:
- replies?: Comment[]
- repliesTotal: number
- onReplyCreate?: Function
- onGetReplies?: Function

UI should be wrapped with ActivityCard component so that the entire thread is treated as a card
UI will render the main item based on it’s type using Comment or AnnotationActivity

### DIFF
--- a/src/common/types/annotations.js
+++ b/src/common/types/annotations.js
@@ -53,6 +53,7 @@ export type Annotation = {
     repliesTotal?: number,
     status?: AnnotationStatus,
     target: Target,
+    total_reply_count?: number,
     type: 'annotation',
 };
 
--- a/src/common/types/annotations.js
+++ b/src/common/types/annotations.js
@@ -50,7 +50,6 @@ export type Annotation = {
     modified_by: User,
     permissions: AnnotationPermission,
     replies?: Array<Comment>,
-    repliesTotal?: number,
     status?: AnnotationStatus,
     target: Target,
     total_reply_count?: number,
--- a/src/common/types/annotations.js
+++ b/src/common/types/annotations.js
@@ -1,7 +1,7 @@
 // @flow
 
 import type { BoxItemVersionMini, Reply, User } from './core';
-import type { ActionItemError } from './feed';
+import type { ActionItemError, Comment } from './feed';
 
 export type Page = {
     type: 'page',
@@ -49,7 +49,8 @@ export type Annotation = {
     modified_at: string,
     modified_by: User,
     permissions: AnnotationPermission,
-    replies?: Array<Reply>,
+    replies?: Array<Comment>,
+    repliesTotal?: number,
     status?: AnnotationStatus,
     target: Target,
     type: 'annotation',
--- a/src/common/types/feed.js
+++ b/src/common/types/feed.js
@@ -39,10 +39,14 @@ type Comment = {
     is_reply_comment?: boolean,
     message?: string,
     modified_at: string,
+    parent?: {
+        id: string,
+        type: 'comment' | 'annotation',
+    },
     permissions: BoxCommentPermission,
     replies?: Array<Comment>,
-    repliesTotal?: number,
     tagged_message: string,
+    total_reply_count?: number,
     type: 'comment',
 };
 
--- a/src/common/types/feed.js
+++ b/src/common/types/feed.js
@@ -40,6 +40,8 @@ type Comment = {
     message?: string,
     modified_at: string,
     permissions: BoxCommentPermission,
+    replies?: Array<Comment>,
+    repliesTotal?: number,
     tagged_message: string,
     type: 'comment',
 };
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
@@ -94,11 +94,11 @@ const ActiveState = ({
                         return (
                             <ActivityItem
                                 key={item.type + item.id}
-                                data-testid={hasReplies ? 'activity-thread' : 'comment'}
+                                data-testid="comment"
                                 isFocused={isFocused}
                                 ref={refValue}
                             >
-                                <ActivityThread hasReplies={hasReplies}>
+                                <ActivityThread hasReplies={hasReplies} data-testid="activity-thread">
                                     <Comment
                                         {...item}
                                         currentUser={currentUser}
@@ -169,12 +169,12 @@ const ActiveState = ({
                         return (
                             <ActivityItem
                                 key={item.type + item.id}
-                                className={!hasReplies && 'bcs-activity-feed-annotation-activity'}
-                                data-testid={hasReplies ? 'activity-thread' : 'annotation-activity'}
+                                className="bcs-activity-feed-annotation-activity"
+                                data-testid="annotation-activity"
                                 isFocused={isFocused}
                                 ref={refValue}
                             >
-                                <ActivityThread hasReplies={hasReplies}>
+                                <ActivityThread hasReplies={hasReplies} data-testid="activity-thread">
                                     <AnnotationActivity
                                         currentUser={currentUser}
                                         getAvatarUrl={getAvatarUrl}
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
@@ -4,6 +4,7 @@
  */
 import * as React from 'react';
 import getProp from 'lodash/get';
+import ActivityThread from './ActivityThread';
 import ActivityItem from './ActivityItem';
 import AppActivity from '../app-activity';
 import AnnotationActivity from '../annotations';
@@ -33,6 +34,7 @@ type Props = {
     getAvatarUrl: GetAvatarUrlCallback,
     getMentionWithQuery?: Function,
     getUserProfileUrl?: GetProfileUrlCallback,
+    hasReplies?: boolean,
     items: FeedItems,
     mentionSelectorContacts?: SelectorItems<>,
     onAnnotationDelete?: ({ id: string, permissions: AnnotationPermission }) => void,
@@ -57,6 +59,7 @@ const ActiveState = ({
     approverSelectorContacts,
     currentFileVersionId,
     currentUser,
+    hasReplies = false,
     items,
     mentionSelectorContacts,
     getMentionWithQuery,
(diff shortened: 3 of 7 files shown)


src/common/types/annotations.js
```diff
@@ -47,7 +47,8 @@ export type Annotation = {
     modified_at: string,
     modified_by: User,
     permissions: AnnotationPermission,
-    replies?: Array<Reply>,
+    replies?: Array<Comment>,
+    repliesTotal?: number,
```
jstoffan, line 51: Nearly all V2 API fields use snake_case naming (see the other fields on this object), so this would be `replies_total`.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -85,6 +88,35 @@ const ActiveState = ({
...
+                        >
+                            <ActivityThread
+                                currentFileVersionId={currentFileVersionId}
+                                getAvatarUrl={getAvatarUrl}
+                                isCurrentVersion={isCurrentVersion}
+                                item={item}
+                                onAnnotationDelete={onAnnotationDelete}
+                                onAnnotationEdit={onAnnotationEdit}
+                                onAnnotationSelect={onAnnotationSelect}
+                                // onAnnotationReplyCreate={() => {}} // TODO will be implemented in next PR
```
jstoffan, line 110: We generally want to avoid commented code, even if it's temporary. Can we omit these lines?

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -85,6 +88,35 @@ const ActiveState = ({
...
+                            className="bcs-activity-thread"
+                            data-testid="activity-thread"
+                            isFocused={isFocused}
+                            ref={refValue}
+                        >
+                            <ActivityThread
+                                currentFileVersionId={currentFileVersionId}
+                                getAvatarUrl={getAvatarUrl}
+                                isCurrentVersion={isCurrentVersion}
+                                item={item}
```
jstoffan, line 106: It looks like the other activity types spread `{...item}` onto their props. Should we do the same here for consistency?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityThread.js
```diff
@@ -0,0 +1,101 @@
...
+
+const ActivityThread = ({
+    currentUser,
+    getAvatarUrl,
+    getMentionWithQuery,
+    getUserProfileUrl,
+    isCurrentVersion,
+    item,
+    mentionSelectorContacts,
+    // eslint-disable-next-line no-unused-vars
```
jstoffan, line 47: Can we omit these props for now?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityThread.js
```diff
@@ -0,0 +1,101 @@
...
+    onGetAnnotationReplies = noop,
+    // eslint-disable-next-line no-unused-vars
+    onGetCommentReplies = noop,
+    translations,
+}: Props) => {
+    return (
+        <ActivityCard className="bcs-ActivityThread">
+            {item.type === 'comment' && (
+                <Comment
```
jstoffan, line 65: It doesn't look like `Comment` and `AnnotationActivity` share much behavior. Can we move these conditions up to `ActiveState` and allow `ActivityThread` to accept them as a `children` prop?

src/elements/content-sidebar/activity-feed/activity-feed/__tests__/ActivityThread.test.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import { shallow, ShallowWrapper } from 'enzyme';
+import Comment from '../../comment/Comment';
+import AnnotationActivity from '../../annotations/AnnotationActivity';
+import ActivityThread from '../ActivityThread.js';
+import ActivityCard from '../../ActivityCard';
+
+describe('src/elements/content-sidebar/activity-feed/activity-feed/ActivityThread', () => {
+    // eslint-disable-next-line flowtype/no-types-missing-file-annotation
```
jstoffan, line 9: We should try to avoid disabling linting rules whenever possible. Should this file be marked with a Flow annotation

src/common/types/feed.js
```diff
@@ -40,6 +40,8 @@ type Comment = {
     message?: string,
     modified_at: string,
     permissions: BoxCommentPermission,
+    replies?: Array<Comment>,
+    repliesTotal?: number,
```
ConradJChan, line 44: Will the new reply be exactly the `Comment` type as it is today or will the type be slightly different? Also, we could consider creating a common type for this to be used in `Comment` and `Annotation`

src/elements/content-sidebar/activity-feed/activity-feed/ActivityThread.js
ConradJChan: +1, it may be easier to use and extend in the future (for Tasks maybe) if we make this `ActivityThread` component more composable, something like 
```
<ActivityThread>
    <Comment />
</ActivityThread>
```
where it's only responsibility is really to show the reply thread in it's compact form or expanded form

src/common/types/feed.js
```diff
@@ -39,8 +39,14 @@ type Comment = {
     is_reply_comment?: boolean,
     message?: string,
     modified_at: string,
+    parent?: {
+        id: string,
+        type: 'commnet' | 'annotation',
```
jstoffan, line 44: Typo.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -161,6 +193,32 @@ const ActiveState = ({
...
+                        if (hasReplies) {
+                            return (
+                                <ActivityItem
+                                    key={item.type + item.id}
+                                    className="bcs-activity-thread"
+                                    data-testid="activity-thread"
+                                    isFocused={isFocused}
+                                    ref={refValue}
+                                >
+                                    <ActivityThread {...item}>
```
jstoffan, line 205: Do we need to spread `{...item}` here or can it be omitted?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityThread.js
```diff
@@ -0,0 +1,20 @@
+// @flow
+import * as React from 'react';
+import ActivityCard from '../ActivityCard';
+
+import type { Comment } from '../../../../common/types/feed';
+
+type Props = {
+    children: React.Node,
+    id: string,
```
jstoffan, line 9: It looks like most of these props are unused right now. Can they be removed?

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
wpiesiak: It can be omitted currently and added in next PR 👍

src/common/types/feed.js
wpiesiak: Every reply will be basically a comment from api perspective

jstoffan (approved):

src/common/types/feed.js
```diff
@@ -39,8 +39,14 @@ type Comment = {
     is_reply_comment?: boolean,
     message?: string,
     modified_at: string,
+    parent?: {
+        id: string,
+        type: 'comment' | 'annotation',
+    },
     permissions: BoxCommentPermission,
+    replies?: Array<Comment>,
     tagged_message: string,
```
jstoffan, line 48: Will these unused fields be included in the new schema? Or should we be using a new "Comment" type for these flows?

jstoffan: I don't think these fields are included in our new backend schema for these objects, though. We should have accurate types, even if it's a bit more work.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -88,6 +91,35 @@ const ActiveState = ({
 
                 switch (item.type) {
                     case 'comment':
+                        if (hasReplies) {
```
ConradJChan, line 94: I'm wondering if it would be cleaner if we deferred the `hasReplies` check in the `ActivityThread` component

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -91,28 +94,31 @@ const ActiveState = ({
                         return (
                             <ActivityItem
                                 key={item.type + item.id}
-                                className="bcs-activity-feed-comment"
-                                data-testid="comment"
+                                className={hasReplies ? 'bcs-activity-thread' : 'bcs-activity-feed-comment'}
```
jstoffan, line 97: It doesn't look like `.bcs-activity-thread` is used at the moment. Will we need different styles for the ActivityItem component in the future?

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -91,28 +94,31 @@ const ActiveState = ({
...
+                                data-testid={hasReplies ? 'activity-thread' : 'comment'}
```
jstoffan, line 98: Should this be on the `ActivityThread` component, instead?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityThread.js
```diff
@@ -0,0 +1,17 @@
...
+import ActivityCard from '../ActivityCard';
+
+type Props = {
+    children: React.Node,
+    hasReplies: boolean,
+};
+
+const ActivityThread = ({ children, hasReplies }: Props) => {
+    if (hasReplies) {
+        return <ActivityCard className="bcs-ActivityThread">{children}</ActivityCard>;
```
jstoffan, line 12: It looks like this will create a nested ActivityCard output, since the `Comment` component's root is also ActivityCard. Is that desirable? It also doesn't look like `.bcs-ActivityThread` is used at the moment. Will it be in the future?

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
wpiesiak: Probably we won't need to style this ActivityItem differently. I was just following the pattern in the component, but noticed  'bcs-activity-feed-comment' is not used neither

src/elements/content-sidebar/activity-feed/activity-feed/ActivityThread.js
wpiesiak: I am aware of it, but I think it would be best to treat ActivityThread as ActivityCard containing main comment with replies and wrapping it up. I think it shouldn't be a problem to have nested ActivityCard for parent comment and each reply. Probably it would just need style fixes for the spacing for the parent component but IMO it would look like more consistent.

wpiesiak: And I would keep the className - it will be used in next [PR](https://github.com/box/box-ui-elements/pull/3010/files)

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
jstoffan: I ask because ActivityItem already has a `data-testid` that we can use. If we want to select on ActivityThread, as well, we can add a `data-testid` there.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -164,25 +169,28 @@ const ActiveState = ({
                         return (
                             <ActivityItem
                                 key={item.type + item.id}
-                                className="bcs-activity-feed-annotation-activity"
-                                data-testid="annotation-activity"
+                                className={!hasReplies && 'bcs-activity-feed-annotation-activity'}
```
jstoffan, line 172: Can we remove this logic? The class seems unrelated since it's on a parent of the ActivityThread.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -164,25 +169,28 @@ const ActiveState = ({
...
+                                data-testid={hasReplies ? 'activity-thread' : 'annotation-activity'}
```
jstoffan, line 173: Again, this should be on the `ActivityThread`, as the name implies.

src/common/types/feed.js
jstoffan: Did we come to a resolution on this question?

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
wpiesiak: No I get it 😄 Sounds good - fixed

jstoffan (approved):

PR #3007, opened 2022-08-24
feat: added accordion collapsible component

assadiandre:
An accordion-type collapsible that is very similar in functionality to the existing collapsible component, but is styled differently. The accordion collapsible is used in Box Forms, and it may be used in Sign and Docgen as well.

Screen Shot 2022-08-24 at 3 26 24 PM


greg-in-a-box:
Different approach will be taken

PR #3012, opened 2022-08-29
feat(annotation-activity): add resolve functionality

wpiesiak:
Adjust AnnotationActivity component to support resolve functionality


src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
```diff
@@ -30,11 +30,13 @@ type Props = {
     getMentionWithQuery?: (searchStr: string) => void,
     getUserProfileUrl?: GetProfileUrlCallback,
     isCurrentVersion: boolean,
+    isResolvingEnabled?: boolean,
```
jstoffan, line 33: Can we control this server-side so that `can_resolve` is always false if the feature is disabled?

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityMenu.js
```diff
@@ -4,36 +4,48 @@ import classNames from 'classnames';
 import { FormattedMessage } from 'react-intl';
 import DropdownMenu from '../../../../components/dropdown-menu';
 import IconEllipsis from '../../../../icons/general/IconEllipsis';
-import messages from './messages';
 import Pencil16 from '../../../../icon/line/Pencil16';
 import PlainButton from '../../../../components/plain-button';
 import Trash16 from '../../../../icon/fill/Trash16';
+import Checkmark16 from '../../../../icon/fill/Checkmark16';
```
jstoffan, line 10: Can we keep the imports alphabetized?

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
wpiesiak: It is meant to based on `hasReplies` prop from ActivitySidebar. For server-side decision handling, we would need to send this data anyway

jstoffan: I don't know if that really answered the question. Replies and resolve aren't directly related, they just happen to be being built at the same time. Can we make the decision about showing these elements based solely on the `can_resolve` permission, instead?

src/common/types/feed.js
```diff
@@ -2,6 +2,7 @@
 import type { MessageDescriptor } from 'react-intl';
 import type { BoxItemPermission, BoxItemVersion, Reply, User } from './core';
 import type { Annotation, AnnotationPermission, Annotations } from './annotations';
+import { COMMENT_STATUS_OPEN, COMMENT_STATUS_RESOLVED } from '../../constants';
```
tjuanitas, line 5: nit: I think we would move this above the `import type` section

src/constants.js
```diff
@@ -451,6 +451,10 @@ export const TASK_EDIT_MODE_EDIT: 'EDIT' = 'EDIT';
 /* ----------------- Task Validation ---------------- */
 export const TASK_MAX_GROUP_ASSIGNEES: 250 = 250;
 
+/* ------------------ Comment (and Annotation) Statuses ----------------- */
+export const COMMENT_STATUS_OPEN: 'open' = 'open';
+export const COMMENT_STATUS_RESOLVED: 'resolved' = 'resolved';
+
```
tjuanitas, line 457: Does it make sense to call this `FEED_ITEM_STATUS_[...]` so it aligns to with the `type FeedItemStatus`: https://github.com/box/box-ui-elements/blob/d0f669b4ad4323c5ea6f65151a1269f750d32f4b/src/common/types/feed.js#L27

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
tjuanitas: It would help simplify some logic where `isResolvingEnabled && canResolve` is being used. How is `isResolvingEnabled` determined when this component is called?

src/constants.js
jstoffan: @tjuanitas, this is due to feedback from a prior PR. The "feed" is an experience rather than a resource, so referring to it in our API layer feels a bit backward. Using "comment" here is also more consistent with the other existing annotation/task/version constants in this file.

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
wpiesiak: That idea make sense - can_resolve permission will be handled server-side 👍

src/constants.js
wpiesiak: it has been discussed here: https://github.com/box/box-ui-elements/pull/3004#discussion_r955121343

src/elements/content-sidebar/activity-feed/annotations/__tests__/AnnotationActivityMenu.test.js
```diff
@@ -31,8 +31,25 @@ describe('elements/content-sidebar/ActivityFeed/annotations/AnnotationActivityMe
         expect(wrapper.exists('[data-testid="edit-annotation-activity"]')).toBe(true);
     });
 
+    test('should render the resolve annotation activity menu item if canResolve is true for unresolved item', () => {
```
jstoffan, line 34: Should we have a test that validates neither option is shown if `can_resolve` is false?

jstoffan (approved):

Issue #2976, opened 2022-08-08
SearchForm search button is not interactive


rlyonbox:
Reported by Boxer, created on their behalf: "Nothing happens when clicking on the magnifying glass in the [ContextExplorerModal]"

Digging into the code a little, this resolves to the `type` attribute on the button which renders the magnifying glass being `button` instead of `submit` on this line: https://github.com/box/box-ui-elements/blob/ef674ae9124b74126ae11b729e074cf8658c15a5/src/components/search-form/SearchForm.js#L168

Resolution is trivial, 1 line change, but I'm not sure the provenance of using `type=button` in the first place.

greg-in-a-box:
Agree with changing the button `type` to a `submit`.  The unit test needs to be updated to simulate clicking on the button and not invoking the submit event directly https://github.com/box/box-ui-elements/blob/ef674ae9124b74126ae11b729e074cf8658c15a5/src/components/search-form/__tests__/SearchForm.test.js#L66

PR #2999, opened 2022-08-18
feat(content-sidebar): initial Threaded Replies related changes

kkuliczkowski-box:
These are the initial changes for Threaded Replies.

* Option to disable features in `ContentSidebar`
* Added `hasReplies` flag to `ActivitySidebarProps` for future development

PR #3013, opened 2022-08-29
fix(search-form): fix search button not invoking form submit

QingyuChai:
The search button was previously un-responsive. Simple fix from `type="button"` to `type="submit"`.

In response to https://github.com/box/box-ui-elements/issues/2976#issuecomment-1211240114:

Simulating click only triggers the onClick callback, and does not actually simulate actually clicking on the component, and does not fire the "submit" event.  Per the discussion here:

https://github.com/enzymejs/enzyme/issues/308

A proposed solution is to use `simulate("submit", ...)`, which is currently implemented. Other users mentioned using something like `wrapper.find('[type="submit"]').get(0).click();`, which seems to be not supported in our codebase. 

Edit: Trevor pointed out that a fellow boxer, Brett Lamy, actually participated in that discussion


QingyuChai:
Force pushed as status checks got stuck - retriggered them with no-change commit

tjuanitas (commented):
Can we verify in the parent apps if they're using this component that there's no issue if we make this switch? I think there are usages in AC but EUA seems to use its own version of this component.

Would we be able to use this approach for simulating the `onSubmit` on the button? https://github.com/enzymejs/enzyme/issues/308#issuecomment-291604063

It would most likely be a separate test so we can still test the form separately but we'd copy and swap these lines for a button: https://github.com/box/box-ui-elements/blob/ef674ae9124b74126ae11b729e074cf8658c15a5/src/components/search-form/__tests__/SearchForm.test.js#L65-L66

src/components/search-form/__tests__/SearchForm.test.js
```diff
@@ -67,6 +67,23 @@ describe('components/search-form/SearchForm', () => {
...
+        const onSubmitSpy = sinon.spy();
+        const event = {
+            target: {
+                elements: [
+                    {
+                        value: 'cheese',
+                    },
+                ],
+            },
+        };
```
QingyuChai, line 80: Have considered moving this out of the test since it's duplicated, not sure if necessary

greg-in-a-box: if its duplicated among the `onSubmit` test then I would consider putting them into the same `describe` then

src/components/search-form/__tests__/SearchForm.test.js
```diff
@@ -67,6 +67,23 @@ describe('components/search-form/SearchForm', () => {
...
+        const wrapper = mount(<SearchForm onSubmit={onSubmitSpy} placeholder="search" value="cheese" />);
+        const searchButton = wrapper.find('form').find('button.search-button');
+        searchButton.simulate('submit', event);
```
greg-in-a-box, line 83: Does this still pass if the button type was still `button`?

src/components/search-form/__tests__/SearchForm.test.js
```diff
@@ -67,6 +67,23 @@ describe('components/search-form/SearchForm', () => {
...
+        const event = {
+            target: {
+                elements: [
+                    {
+                        value: 'cheese',
+                    },
+                ],
+            },
+        };
+        const wrapper = mount(<SearchForm onSubmit={onSubmitSpy} placeholder="search" value="cheese" />);
```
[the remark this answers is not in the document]
QingyuChai: Shallow doesn't work, because the form itself is mounted as a component
```
<LoadableSearchActions
    isLoading={isLoading}
    loadingIndicatorProps={{
        className: 'search-form-loading-indicator',
    }}
/>
```

src/components/search-form/__tests__/SearchForm.test.js
```diff
@@ -67,6 +67,23 @@ describe('components/search-form/SearchForm', () => {
...
+        const searchButton = wrapper.find('form').find('button.search-button');
```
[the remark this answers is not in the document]
QingyuChai: Yup, makes sense!

src/components/search-form/__tests__/SearchForm.test.js
QingyuChai: It does, you're right. I will make sure to explicitly test for the "submit" type then

src/components/search-form/__tests__/SearchForm.test.js
```diff
@@ -67,6 +67,23 @@ describe('components/search-form/SearchForm', () => {
         sinon.assert.calledWithMatch(onSubmitSpy, 'cheese', event);
     });
 
+    test('should call onsubmit on search icon submit event', () => {
+        const onSubmitSpy = sinon.spy();
```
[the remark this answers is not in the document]
tjuanitas: Might be able to use `expect.objectContaining`. There's an example on using it with events: https://jestjs.io/docs/expect#expectobjectcontainingobject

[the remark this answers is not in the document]
QingyuChai: oh that's cool, didn't know this existed. Confirmed it works, thanks!

src/components/search-form/__tests__/SearchForm.test.js
```diff
@@ -61,10 +60,30 @@ describe('components/search-form/SearchForm', () => {
...
+        afterEach(() => {
+            onSubmitMock.mockReset();
+        });
+
+        test('should call onsubmit on form submit event', () => {
+            const wrapper = mount(<SearchForm onSubmit={onSubmitMock} placeholder="search" value="cheese" />);
+            const form = wrapper.find('form');
+
+            form.simulate('submit', event);
+            expect(onSubmitMock).toHaveBeenCalledTimes(1);
```
tjuanitas, line 74: Could we update this and the test below to `toBeCalledTimes`?  https://github.com/box/box-ui-elements/pull/2999#discussion_r951632495
```suggestion
            expect(onSubmitMock).toBeCalledTimes(1);
```

tjuanitas (commented):
one more comment otherwise looks good to me!

tjuanitas (approved):
:tada:

PR #3016, opened 2022-08-30
refactor(current-user): Hoist fetching currentUser

greg-in-a-box:
- Had to move fetching the `currentUser` higher so that both `ActivitySidebar` and `withSidebarAnnotations` were able to get the `currentUser` data without having the consumer retrieve the `currentUser` themselves.
- `currentUser` and `currentUserError` is only available via `props` in `ActivitySidebar` instead of `state`.
- Tests related to fetching the currentUser has been moved to `Sidebar`
- Movement of code blocks is the result of prettier reordering the lifecycle methods

Settings:
`showAnnotations` set to `true` - displays annotations
`showAnnotationsControls` set to `true` - displays annotations controls
`contentSideBarProps.features.activityFeed.annotations.enabled` set to `true` - displays annotation comments in activity sidebar
All annotations props relating to showing and controls still work the same for both legacy and v4 annotations.
`showAnnotationsDrawingCreate` set to `true` - enabled drawing an annotation

Test:
Must supply BoxAnnotations 
via `ContentPreview`
- [x] Can create drawn annotation 
- [x] Can leave comment with drawn annotation
- [x] Can create rectangle annotation
- [x] Can leave comment with rectangle annotation
- [x] Drawn annotation comment appears in `ActivitySidebar`
- [x] Rectangle annotation comment appears in `ActivitySidebar`
- [x] Can leave comment in `ActivitySidebar`

via `preview.js`
- [x] Can create drawn annotation 
- [x] Can leave comment with drawn annotation
- [x] Can create rectangle annotation
- [x] Can leave comment with rectangle annotation
- [x] Drawn annotation comment appears in `ActivitySidebar`
- [x] Rectangle annotation comment appears in `ActivitySidebar`
- [x] Can leave comment in `ActivitySidebar`


src/elements/content-preview/ContentPreview.js
```diff
@@ -184,6 +186,34 @@ const LoadableSidebar = AsyncLoad({
 });
 
 class ContentPreview extends React.PureComponent<Props, State> {
+    static defaultProps = {
```
greg-in-a-box, line 189: When I formatted the file it moved this to the top automatically so I assume some where in our configuration its requiring components to have `defaultProps` at the top

src/elements/content-preview/ContentPreview.js
```diff
@@ -289,10 +295,34 @@ class ContentPreview extends React.PureComponent<Props, State> {
...
         logger.onReadyMetric({
             endMarkName: MARK_NAME_JS_READY,
         });
+
+        if (this.props.useV4BoxAnnotations && !this.props.boxAnnotations && !global.BoxAnnotations) {
+            importV4Annotations().then(() => {
+                const features = {
+                    discoverability: true,
+                    drawing: true,
+                };
```
greg-in-a-box, line 309: I am uncertain if I should be setting this to true for the consumer or if the consumer should have control of this

src/elements/common/async.js
```diff
@@ -0,0 +1,3 @@
+const importV4Annotations = () => import(/* webpackChunkName: "box-annotations" */ 'box-annotations');
```
jstoffan, line 1: Does this work for the pre-built bundles and self-hosted assets? We use code splitting like this elsewhere in the app, but we add the `webpackMode: "lazy"` directive, which is [replaced conditionally|]https://github.com/box/box-ui-elements/blob/master/scripts/webpack.config.js#L159 based on which asset we're building. We likely need to add it here, too, though it will have an impact on load performance in those use-cases.

src/elements/content-sidebar/Sidebar.js
```diff
@@ -288,4 +381,4 @@ class Sidebar extends React.Component<Props, State> {
 }
 
 export { Sidebar as SidebarComponent };
-export default flow([withFeatureConsumer, withRouter])(Sidebar);
+export default flow([withAPIContext, withFeatureConsumer, withRouter])(Sidebar);
```
jstoffan, line 384: Minor, but if we're adding a HOC here we could create a `withCurrentUser` HOC instead to encapsulate the lazy-loading logic.

src/elements/content-preview/ContentPreview.js
```diff
@@ -289,10 +295,34 @@ class ContentPreview extends React.PureComponent<Props, State> {
             // eslint-disable-next-line react/no-unused-state
             prevFileIdProp: fileId,
         };
+
         const { logger } = props;
         logger.onReadyMetric({
             endMarkName: MARK_NAME_JS_READY,
         });
+
+        if (this.props.useV4BoxAnnotations && !this.props.boxAnnotations && !global.BoxAnnotations) {
```
jstoffan, line 304: We typically destructure all our objects rather than using direct accessors.

src/elements/content-sidebar/Sidebar.js
```diff
@@ -236,6 +326,8 @@ class Sidebar extends React.Component<Props, State> {
             'bcs-is-open': isOpen,
         });
 
+        const { currentUser, currentUserError } = this.state;
```
megansmith-box, line 329: Why do we need to fetch and get currentUser from state rather than props?

greg-in-a-box: `ActivitySidebar` needs a `currentUser` to post comments. When `ContentPreview` is used in the webapp, the `currentUser` is provided based on who is logged in verse using the token and file info to assume the `currentUser`.  So if the `currentUser` is retrieved separately and passed via props then its not fetched at all.

src/elements/common/async.js
greg-in-a-box: I added the `webpackMode` and set it to `lazy` and retested with the HOC and it is working via pre-built.  Though, I just tried using the `dist/preview.js` file directly and it doesnt seem to load box-annotations into the global object.  Do you have any suggestions on this?

src/elements/content-sidebar/__tests__/Sidebar.test.js
```diff
@@ -1,30 +1,66 @@
 import React from 'react';
 import { shallow } from 'enzyme';
-import LocalStore from '../../../utils/LocalStore';
 import {
+    SidebarComponent,
     SIDEBAR_FORCE_KEY,
     SIDEBAR_FORCE_VALUE_CLOSED,
     SIDEBAR_FORCE_VALUE_OPEN,
     SidebarComponent as Sidebar,
```
tjuanitas, line 8: SidebarComponent is already being imported as Sidebar?

src/elements/content-sidebar/__tests__/Sidebar.test.js
```diff
@@ -1,30 +1,66 @@
...
     beforeEach(() => {
         LocalStore.mockClear();
     });
 
+    describe('componentDidMount()', () => {
+        let wrapper;
+        let instance;
+        currentUser = {
+            id: '123',
+        };
```
tjuanitas, line 46: do we need this?

src/elements/content-sidebar/__tests__/Sidebar.test.js
```diff
@@ -1,30 +1,66 @@
...
+        beforeEach(() => {
+            jest.spyOn(SidebarComponent.prototype, 'fetchCurrentUser');
+            wrapper = getWrapper({
+                currentUser,
+            });
+            instance = wrapper.instance();
+        });
+
+        afterEach(() => {
```
tjuanitas, line 55: there's only one test in this describe, do we need the `beforeEach` and `afterEach`?

src/elements/content-sidebar/__tests__/Sidebar.test.js
```diff
@@ -1,30 +1,66 @@
...
+    SidebarComponent,
     SIDEBAR_FORCE_KEY,
     SIDEBAR_FORCE_VALUE_CLOSED,
     SIDEBAR_FORCE_VALUE_OPEN,
     SidebarComponent as Sidebar,
 } from '../Sidebar';
+import messages from '../../common/messages';
+import LocalStore from '../../../utils/LocalStore';
+
+const { defaultErrorMaskSubHeaderMessage, currentUserErrorHeaderMessage } = messages;
```
tjuanitas, line 13: nit: could you sort these?

src/elements/content-sidebar/__tests__/ActivitySidebar.test.js
```diff
@@ -85,15 +84,10 @@ describe('elements/content-sidebar/ActivitySidebar', () => {
     describe('componentDidMount()', () => {
         let wrapper;
         let instance;
-        currentUser = {
-            id: '123',
-        };
         beforeEach(() => {
```
tjuanitas, line 87: nit: could you add a line break to separate these?

package.json
```diff
@@ -332,6 +333,7 @@
         "@box/react-virtualized": "9.22.3-rc-box.2",
         "@hapi/address": ">=2.0.0 <2.1.2",
         "axios": "^0.25.0",
+        "box-annotations": "^4.1.1-beta.1",
```
jstoffan, line 336: Can we clarify how we intend to handle this dependency in the future? Will this package eventually be moved to [box-content-preview](https://github.com/box/box-content-preview/blob/master/package.json#L38) to replace the existing `box-annotations` dependency there? Do we intend to continue supporting two versions of annotations?

src/elements/common/async.js
jstoffan: It's probably a matter of fussing with the webpack config so that it creates `dist/box-annotations.js`. Does it need to be added to the `entries` map in webpack.config.js?

src/elements/content-preview/ContentPreview.js
```diff
@@ -751,13 +768,15 @@ class ContentPreview extends React.PureComponent<Props, State> {
     loadPreview = async (): Promise<void> => {
         const {
             annotatorState: { activeAnnotationId } = {},
+            boxAnnotations,
             enableThumbnailsSidebar,
             fileOptions,
             onAnnotatorEvent,
             onAnnotator,
             previewExperiences,
             showAnnotationsControls,
```
jstoffan, line 777: I recall that multiple booleans, including this one, need to be `true` for the annotations controls to show up and work as expected. How do all of these annotations-related booleans work together? Do we need `useV4BoxAnnotations` at all or can we use `showAnnotationsControls` and check for the absence of `boxAnnotations` and lazy-load it if needed?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -342,12 +311,11 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
 
     updateTaskAssignment = (taskId: string, taskAssignmentId: string, status: TaskCollabStatus): void => {
-        const { file, api, onTaskAssignmentUpdate } = this.props;
-        const { currentUser = {} } = this.state;
+        const { api, currentUser, file, onTaskAssignmentUpdate } = this.props;
 
         const successCallback = () => {
             this.feedSuccessCallback();
-            onTaskAssignmentUpdate(taskId, taskAssignmentId, status, currentUser.id);
+            onTaskAssignmentUpdate(taskId, taskAssignmentId, status, currentUser?.id);
```
jstoffan, line 318: Is it possible for `currentUser` to be undefined. Is that desirable or should this throw? Also, we don't see much optional chaining in BUIE, at least in top-level Elements. Does this present any issues for the build?

src/elements/content-preview/ContentPreview.js
```diff
@@ -751,6 +754,7 @@ class ContentPreview extends React.PureComponent<Props, State> {
     loadPreview = async (): Promise<void> => {
         const {
             annotatorState: { activeAnnotationId } = {},
+            boxAnnotations,
```
jstoffan, line 757: Does this need to be destructured still?

src/elements/content-preview/ContentPreview.js
```diff
@@ -194,6 +221,8 @@ class ContentPreview extends React.PureComponent<Props, State> {
 
     api: API;
 
+    boxAnnotations: Object;
```
jstoffan, line 224: Is this instance property still needed?

src/elements/common/current-user/withCurrentUser.tsx
```diff
@@ -0,0 +1,142 @@
...
+    currentUser?: User;
+    currentUserError?: Errors;
+};
+
+export type WithCurrentUserComponent<P> = React.ComponentClass<P & WithCurrentUserProps>;
+
+export default function withCurrentUser<P extends object>(
+    WrappedComponent: React.ComponentType<P & WithCurrentUserProps>,
+): WithCurrentUserComponent<P> {
+    class ComponentWithCurrentUser extends React.Component<P & WithCurrentUserProps, CurrentUserState> {
```
jstoffan, line 40: I assume we stuck with a class to reduce the effort involved in refactoring, but a function component would be nice someday. 😄

src/elements/common/current-user/withCurrentUser.tsx
```diff
@@ -0,0 +1,142 @@
...
+        };
+
+        render() {
+            const { currentUser, currentUserError } = this.state;
+            return <WrappedComponent {...this.props} currentUser={currentUser} currentUserError={currentUserError} />;
+        }
+    }
+
+    const displayName = WrappedComponent.displayName || WrappedComponent.name || 'Component';
+    ComponentWithCurrentUser.displayName = `WithCurrentUser(${displayName})`;
```
jstoffan, line 139: Should this use a lower-case `w`?

src/elements/common/current-user/withCurrentUser.tsx
```diff
@@ -0,0 +1,142 @@
...
+         * @param {Error} code - the code for the error
+         * @param {Object} contextInfo - the context info for the error
+         * @return {void}
+         */
+        errorCallback = (error: ElementsXhrError, code: string, contextInfo: Object = {}): void => {
+            /* eslint-disable no-console */
+            console.error(error);
+            /* eslint-enable no-console */
+
+            // eslint-disable-next-line react/prop-types
```
jstoffan, line 68: Is this still needed?

src/elements/common/current-user/withCurrentUser.tsx
jstoffan: To clarify: I don't think a conversion is needed as part of this PR, given the surface area of the change.

src/elements/content-preview/ContentPreview.js
greg-in-a-box: the webapp uses this to use v4 annotations, so it still needs to be here

src/elements/common/current-user/withCurrentUser.tsx
greg-in-a-box: I was following this https://github.com/box/box-ui-elements/blob/d0d0132795f84329df95b002620b4d0ace21f51a/src/elements/common/annotator-context/withAnnotations.tsx#L184 do you want me to change it to a lower case ?

src/elements/content-preview/ContentPreview.js
greg-in-a-box: probably not, going to remove it

src/elements/content-preview/ContentPreview.js
greg-in-a-box: Spoke with product about the required of a flag and we settled with no flag and if BoxAnnotations object is passed at all it will just assume v4 annotations and the rest of the existing flags will still work as intended.

package.json
greg-in-a-box: removed the dependency after speaking with product

src/elements/content-preview/ContentPreview.js
jstoffan: Has this file actually changed apart from formatting? It looks like the logical changes have been removed?

src/elements/common/current-user/withCurrentUser.tsx
jstoffan: If that's how we have it elsewhere, let's leave it be. It's not super important.

jstoffan (dismissed):

jstoffan (approved):

tjuanitas (approved):
this PR is just focused on extracting currentUser right? could we update the title?

PR #3019, opened 2022-09-01
feat(content-sidebar): updated annotations logic for threaded replies

kkuliczkowski-box:
- Updated logic to fetch Annotations with replies when `hasReplies` flag is true
- Added logic to change Annotation's status
-- propagated status change handler down to `AnnotationActivity` (it's use will be implemented in future PR)

### DIFF
--- a/src/api/Feed.js
+++ b/src/api/Feed.js
@@ -63,6 +63,7 @@ import type {
     Comments,
     FeedItem,
     FeedItems,
+    FeedItemStatus,
     Task,
     Tasks,
 } from '../common/types/feed';
@@ -165,7 +166,8 @@ class Feed extends Base {
     updateAnnotation = (
         file: BoxItem,
         annotationId: string,
-        text: string,
+        text?: string,
+        status?: FeedItemStatus,
         permissions: AnnotationPermission,
         successCallback: (annotation: Annotation) => void,
         errorCallback: ErrorCallback,
@@ -173,19 +175,29 @@ class Feed extends Base {
         if (!file.id) {
             throw getBadItemError();
         }
+        if (!text && !status) {
+            throw new Error('Missing text or status!');
+        }
 
         this.annotationsAPI = new AnnotationsAPI(this.options);
 
         this.file = file;
         this.errorCallback = errorCallback;
 
-        this.updateFeedItem({ message: text, isPending: true }, annotationId);
+        const feedItemChanges = {};
+        if (text) {
+            feedItemChanges.message = text;
+        }
+        if (status) {
+            feedItemChanges.status = status;
+        }
+        this.updateFeedItem({ ...feedItemChanges, isPending: true }, annotationId);
 
         this.annotationsAPI.updateAnnotation(
             this.file.id,
             annotationId,
             permissions,
-            { message: text },
+            feedItemChanges,
             (annotation: Annotation) => {
                 this.updateFeedItem(
                     {
@@ -302,11 +314,13 @@ class Feed extends Base {
         {
             shouldShowAnnotations = false,
             shouldShowAppActivity = false,
+            shouldShowReplies = false,
             shouldShowTasks = true,
             shouldShowVersions = true,
         }: {
             shouldShowAnnotations?: boolean,
             shouldShowAppActivity?: boolean,
+            shouldShowReplies?: boolean,
             shouldShowTasks?: boolean,
             shouldShowVersions?: boolean,
         } = {},
@@ -329,7 +343,9 @@ class Feed extends Base {
         this.file = file;
         this.errors = [];
         this.errorCallback = onError;
-        const annotationsPromise = shouldShowAnnotations ? this.fetchAnnotations(permissions) : Promise.resolve();
+        const annotationsPromise = shouldShowAnnotations
+            ? this.fetchAnnotations(permissions, shouldShowReplies)
+            : Promise.resolve();
         const versionsPromise = shouldShowVersions ? this.fetchVersions() : Promise.resolve();
         const currentVersionPromise = shouldShowVersions ? this.fetchCurrentVersion() : Promise.resolve();
         const commentsPromise = this.fetchComments(permissions);
@@ -359,7 +375,7 @@ class Feed extends Base {
         });
     }
 
-    fetchAnnotations(permissions: BoxItemPermission): Promise<?Annotations> {
+    fetchAnnotations(permissions: BoxItemPermission, shouldFetchReplies?: boolean): Promise<?Annotations> {
         this.annotationsAPI = new AnnotationsAPI(this.options);
         return new Promise(resolve => {
             this.annotationsAPI.getAnnotations(
@@ -368,6 +384,9 @@ class Feed extends Base {
                 permissions,
                 resolve,
                 this.fetchFeedItemErrorCallback.bind(this, resolve),
+                undefined,
+                undefined,
+                shouldFetchReplies,
             );
         });
     }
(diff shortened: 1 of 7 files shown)


src/api/Feed.js
```diff
@@ -165,27 +166,38 @@ class Feed extends Base {
...
+        text?: string,
+        status?: FeedItemStatus,
         permissions: AnnotationPermission,
         successCallback: (annotation: Annotation) => void,
         errorCallback: ErrorCallback,
     ): void => {
         if (!file.id) {
             throw getBadItemError();
         }
+        if (!text && !status) {
```
jstoffan, line 178: What's the user experience in this case?

src/api/Feed.js
```diff
@@ -165,27 +166,38 @@ class Feed extends Base {
...
+        }
 
         this.annotationsAPI = new AnnotationsAPI(this.options);
 
         this.file = file;
         this.errorCallback = errorCallback;
 
-        this.updateFeedItem({ message: text, isPending: true }, annotationId);
+        const feedItemChanges = {};
+        if (text) {
```
jstoffan, line 188: Do we need these checks at this layer or can we defer validation to the API, as we do now?

src/api/__tests__/Feed.test.js
```diff
@@ -569,6 +585,21 @@ describe('api/Feed', () => {
             expect(annotationItems instanceof Promise).toBeTruthy();
             expect(feed.annotationsAPI.getAnnotations).toBeCalled();
         });
+
+        test('when called with shouldFetchReplies = true, should return a promise and call the annotations api with shouldFetchReplies param as true', () => {
+            const annotationItems = feed.fetchAnnotations({ can_edit: true }, true);
+            expect(annotationItems instanceof Promise).toBeTruthy();
+            expect(feed.annotationsAPI.getAnnotations).toBeCalledWith(
+                expect.anything(),
```
jstoffan, line 593: Can we increase the value of this test with more specificity in our assertions?

src/api/Feed.js
```diff
@@ -165,27 +166,38 @@ class Feed extends Base {
...
+        status?: FeedItemStatus,
         permissions: AnnotationPermission,
         successCallback: (annotation: Annotation) => void,
         errorCallback: ErrorCallback,
     ): void => {
         if (!file.id) {
             throw getBadItemError();
         }
+        if (!text && !status) {
+            throw new Error('Missing text or status!');
```
jstoffan, line 179: Should this also use `getBadItemError`? Is this message shown to the user? If so, should it be localized in the component rather than passed back here?

src/api/__tests__/Feed.test.js
```diff
@@ -1367,31 +1398,81 @@ describe('api/Feed', () => {
...
             expect(feed.file).toEqual(file);
         });
 
-        test('should updateFeedItem with isPending set to true', () => {
-            feed.updateFeedItem = jest.fn();
-            feed.updateAnnotation(file, annotationId, text, permissions, successCallback, errorCallback);
+        describe('should updateFeedItem with isPending set to true', () => {
+            test.each`
+                testText     | testStatus   | expected                                                 | description
+                ${'hello'}   | ${undefined} | ${{ message: 'hello', isPending: true }}                 | ${'only text'}
```
jstoffan, line 1429: We generally try to avoid having the description or test summary in the test.each table. Should these just be separate tests?

src/elements/content-sidebar/__tests__/ActivitySidebar.test.js
```diff
@@ -358,32 +358,50 @@ describe('elements/content-sidebar/ActivitySidebar', () => {
...
-            ${true}            | ${false}           | ${true}      | ${true}         | ${true}             | ${false}            | ${true}       | ${true}
-            ${true}            | ${true}            | ${true}      | ${true}         | ${true}             | ${true}             | ${true}       | ${true}
-            ${false}           | ${true}            | ${true}      | ${true}         | ${false}            | ${true}             | ${true}       | ${true}
-            ${false}           | ${false}           | ${true}      | ${true}         | ${false}            | ${false}            | ${true}       | ${true}
-            ${true}            | ${true}            | ${true}      | ${false}        | ${true}             | ${true}             | ${true}       | ${false}
-            ${true}            | ${false}           | ${true}      | ${false}        | ${true}             | ${false}            | ${true}       | ${false}
+            annotationsEnabled | appActivityEnabled | repliesEnabled | tasksEnabled | versionsEnabled | expectedAnnotations | expectedAppActivity | expectedReplies | expectedTasks | expectedVersions
```
jstoffan, line 361: I may be missing something, but it doesn't seem like these variables interact with each other. Is this level of permutation testing necessary or can we simplify to asserting that `expectedAnnotations === annotationsEnabled`?

src/api/Feed.js
kkuliczkowski-box: Currently there is no user experience. This situation would be abnormal just as `!file.id` above. It only throws an error. We could update the UI with running `this.updateCommentErrorCallback` for both of the cases just as it's done when annotationsAPI calls errorCallback. What do you think?

src/api/Feed.js
kkuliczkowski-box: I believe this is needed because `this.updateFeedItem` is called below which updates the annotation temporarily (when the api request is not yet resolved) overwriting it's properties with the properties of the passed `updates` object. Before only message could be updated so the new message was displayed along with the loading indicator. Now when only `status` could be changed then `message` would be `undefined` so during this "loading time" the text would disappear.

src/api/Feed.js
kkuliczkowski-box: As mentioned in the previous comment this message is currently not shown to the user.

I was thinking before if this should be the `badItemError` and decided against because this check verifies the updated values of `message` and `status` and those are not yet part of the item. However I'm not strongly against changing it if you think it qualifies as the `badItemError`.

src/api/__tests__/Feed.test.js
kkuliczkowski-box: I don't think it makes sense to break it into separate tests as they would practically be copies. I can however remove the `description` column and modify the description of test to include the used values like so:
`'given text=$testText and status=$testStatus'`
What do you think?

src/elements/content-sidebar/__tests__/ActivitySidebar.test.js
kkuliczkowski-box: I agree, these do not depend on each other. When I've added the `tasksEnabled` option I was suggested to include all permutations :) (https://github.com/box/box-ui-elements/pull/2999#discussion_r949551034). I'll shorten it and propose to leave only 2 lines so that each option is checked against true and false like that:
```
annotationsEnabled | appActivityEnabled | repliesEnabled | tasksEnabled | versionsEnabled | expectedAnnotations | expectedAppActivity | expectedReplies | expectedTasks | expectedVersions
${false}           | ${false}           | ${false}       | ${false}     | ${false}        | ${false}            | ${false}            | ${false}        | ${false}      | ${false}
${true}            | ${true}            | ${true}        | ${true}      | ${true}         | ${true}             | ${true}             | ${true}         | ${true}       | ${true}
```
What do you think?

jstoffan: Sounds good to me.

src/api/__tests__/Feed.test.js
jstoffan: Sounds good.

src/api/Feed.js
jstoffan: Got it, thanks for clarifying.

jstoffan (approved):
Remaining feedback is minor enough to be optional.

src/api/__tests__/Feed.test.js
kkuliczkowski-box: Done

src/elements/content-sidebar/__tests__/ActivitySidebar.test.js
kkuliczkowski-box: Done

jstoffan (approved):

PR #3025, opened 2022-09-07
feat(content-sidebar): updated comments logic for threaded replies

kkuliczkowski-box:
- Updated Comments related logic to use `ThreadedComments` api when `hasReplies` flag is true
-- fetching
-- updating
-- deleting
- Added logic to change Comment status
-- propagated status change handler down to `Comment` component (it's use will be implemented in future PR)


src/api/Feed.js
```diff
@@ -395,9 +402,23 @@ class Feed extends Base {
      * Fetches the comments for a file
      *
      * @param {Object} permissions - the file permissions
+     * @param {boolean} shouldFetchReplies - indicator for fetchhing comment with replies
```
MateuszMamczarz, line 405: Typo: `fetchhing`

src/api/ThreadedComments.js
```diff
@@ -18,12 +18,24 @@ import {
...
 
 class ThreadedComments extends MarkerBasedApi {
+    /**
+     * Formats comment data for use in components.
+     *
+     * @param {Comment} comment - An individual comment entry from the API
+     * @return {Comment} Updated comment
+     */
+    format(comment: Comment): Comment {
+        return {
```
MateuszMamczarz, line 33: Should we also format replies?

kkuliczkowski-box: Good point, I will actually add that in a future PR that will cover logic changes related to replies.

src/api/Feed.js
```diff
@@ -348,7 +355,7 @@ class Feed extends Base {
             : Promise.resolve();
         const versionsPromise = shouldShowVersions ? this.fetchVersions() : Promise.resolve();
         const currentVersionPromise = shouldShowVersions ? this.fetchCurrentVersion() : Promise.resolve();
-        const commentsPromise = this.fetchComments(permissions);
+        const commentsPromise = this.fetchComments(permissions, shouldShowReplies);
```
jstoffan, line 358: Should we consider creating new methods rather than having the conditions in the inner method?

```
const commentsPromise = shouldShowReplies ? this.fetchThreadedComments : this.fetchComments;
```

src/api/Feed.js
```diff
@@ -682,15 +704,32 @@ class Feed extends Base {
...
-            errorCallback: (e: ElementsXhrError, code: string) => {
-                this.deleteCommentErrorCallback(e, code, commentId);
-            },
-        });
+        const successCallbackFn = this.deleteFeedItem.bind(this, commentId, successCallback);
+        const errorCallbackFn = (e: ElementsXhrError, code: string) => {
+            this.deleteCommentErrorCallback(e, code, commentId);
+        };
+
+        if (shouldUseThreadedComments) {
```
jstoffan, line 712: We're using `shouldFetchReplies` above. Can we keep the boolean name consistent for future searches?

src/api/Feed.js
```diff
@@ -661,19 +682,20 @@ class Feed extends Base {
...
      * @param {Function} successCallback - the function which will be called on success
      * @param {Function} errorCallback - the function which will be called on error     *
      * @return {void}
      */
     deleteComment = (
         file: BoxItem,
         commentId: string,
-        permissions: BoxItemPermission,
+        permissions: BoxCommentPermission,
+        shouldUseThreadedComments?: boolean,
```
jstoffan, line 695: This is a breaking change for this method, since parameter order is important. Do we use this API anywhere else? Would creating a new set of methods for the threaded comments be cleaner?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -411,7 +413,57 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+            hasReplies,
+            successCallback,
+            errorCallback,
+        );
+
+        // need to load the pending item
+        this.fetchFeedItems();
+    };
+
+    updateCommentStatus = (
```
jstoffan, line 432: Does this need to be a separate handler or can we pass down `this.updateComment` to the components, instead? It looks like all the underlying logic supports either/or text/status.

src/api/__tests__/Feed.test.js
```diff
@@ -239,6 +239,111 @@ jest.mock('../Comments', () =>
     })),
 );
 
+jest.mock('../ThreadedComments', () =>
+    jest.fn().mockImplementation(() => ({
+        getComments: jest.fn().mockReturnValue({
+            entries: [
```
jstoffan, line 245: I've left previous feedback about this topic: Can we move these long mocks into a separate file so the test suite is easier to read?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
```diff
@@ -235,6 +244,7 @@ class ActivityFeed extends React.Component<Props, State> {
         const { isInputOpen } = this.state;
         const hasAnnotationCreatePermission = getProp(file, ['permissions', PERMISSION_CAN_CREATE_ANNOTATIONS], false);
         const hasCommentPermission = getProp(file, 'permissions.can_comment', false);
+        const hasResolvePermission = getProp(file, ['permissions', PERMISSION_CAN_RESOLVE], false);
```
jstoffan, line 247: Should we have a test for this logic?

src/api/Feed.js
kkuliczkowski-box: I've changed approach and, as suggested, created separate methods for all operations.

src/api/Feed.js
kkuliczkowski-box: After creating separate methods this is outdated.

src/api/Feed.js
kkuliczkowski-box: We are using this API only in `ActivitySidebar` and I've updated it to support the addition of the `shouldUseThreadedComments` parameter, but I agree, it's safer to split all methods related to comment operations since they're using a different API endpoints.

src/elements/content-sidebar/ActivitySidebar.js
kkuliczkowski-box: I've removed it and integrated into `updateComment` method. I've also updated the use of this handler by the `Comment` component.

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
kkuliczkowski-box: I've removed this as it actually was a wrong approach. The `can_resolve` permission is on a comment (item) level, not file, so it should only be checked in the `Comment` itself (similar to how it's done with `can_edit` or `can_delete` permissions). This will be implemented in a future PR.

src/api/Feed.js
```diff
@@ -395,9 +402,23 @@ class Feed extends Base {
...
-    fetchComments(permissions: BoxItemPermission): Promise<?Comments> {
+    fetchComments(permissions: BoxItemPermission, shouldFetchReplies?: boolean): Promise<?Comments> {
+        if (shouldFetchReplies) {
+            this.threadedCommentsAPI = new ThreadedCommentsAPI(this.options);
+            return new Promise(resolve => {
+                this.threadedCommentsAPI.getComments({
+                    errorCallback: this.fetchFeedItemErrorCallback.bind(this, resolve),
+                    fileId: this.file.id,
+                    permissions,
+                    repliesCount: 1,
```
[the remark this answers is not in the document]
kkuliczkowski-box: Just in case, I've checked this with planned End User Web changes and indeed `1` will be the default value. Removed.

jstoffan (approved):

PR #3026, opened 2022-09-08
feat(content-insights): Migration of components

ConradJChan:


src/features/content-insights/ContentInsightsSummary.scss
```diff
@@ -0,0 +1,25 @@
+@import '../../styles/variables';
+
+.ContentInsightsSummary {
+    display: flex;
+    flex-direction: column;
+
+    .ContentInsightsSummary-header {
+        position: relative;
+        margin: $bdl-grid-unit * 7.5 0 $bdl-grid-unit * 2.5;
```
karelee7, line 9: minor: should we use explicit margin styling rather than the shorthand to be more clear? (i.e. margin-top, margin-left, etc.)

src/features/content-insights/CompactCount.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { injectIntl, IntlShape } from 'react-intl';
+import classNames from 'classnames';
+
+import numAbbr from '../../utils/numAbbr';
+
+interface Props {
+    className?: string;
```
karelee7, line 8: minor: not sure if it's needed, but should we provide a default value (i.e. '') for className in the case that one isn't provided?

src/features/content-insights/ContentInsightsSummary.tsx
```diff
@@ -0,0 +1,41 @@
...
+import GraphCardPreviewsSummary from './GraphCardPreviewsSummary';
+import OpenContentInsightsButton from './OpenContentInsightsButton';
+import { GraphData } from './types';
+
+import './ContentInsightsSummary.scss';
+
+interface Props {
+    graphData: GraphData;
+    isLoading: boolean;
+    onCtaClick: () => void;
```
karelee7, line 13: minor: could we rename this to just onClick since there's no other on click functions in this component? (unless this is the pattern we want to follow for naming onClick props)

src/features/content-insights/ContentAnalyticsErrorState.scss
```diff
@@ -0,0 +1,14 @@
+@import '../../styles/variables';
+
+.ContentAnalyticsErrorState {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+
+    .ContentAnalyticsErrorState-text {
```
karelee7, line 8: minor: do we need to nest this inside of the "ContentAnalyticsErrorState" styling? I vaguely remember something about how this style of class naming (SUIT naming I think it was called) is so that we don't have to nest things too deeply

src/features/content-insights/ContentInsightsSummary.tsx
```diff
@@ -0,0 +1,41 @@
...
+import ContentInsightsSummaryGhostState from './ContentInsightsSummaryGhostState';
+import GraphCardPreviewsSummary from './GraphCardPreviewsSummary';
+import OpenContentInsightsButton from './OpenContentInsightsButton';
+import { GraphData } from './types';
+
+import './ContentInsightsSummary.scss';
+
+interface Props {
+    graphData: GraphData;
+    isLoading: boolean;
```
JChan106, line 12: does this need an isError prop

ConradJChan: It will eventually, I just want to migrate first and we can do a follow on to add the error experience

src/features/content-insights/ContentInsightsSummary.tsx
```diff
@@ -0,0 +1,41 @@
...
+    onCtaClick: () => void;
+    previousPeriodCount: number;
+    totalCount: number;
+}
+
+export default function ContentInsightsSummary({
```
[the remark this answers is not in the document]
ConradJChan: I think arrow function vs named function doesn't matter too much but you're right, I think exporting on a separate line is more consistent with the rest of the codebase

src/features/content-insights/HeaderWithCount.tsx
```diff
@@ -0,0 +1,22 @@
...
+import CompactCount from './CompactCount';
+
+import './HeaderWithCount.scss';
+
+interface Props {
+    title: string;
+    totalCount?: number;
+}
+
+function isNumber(count?: number): count is number {
```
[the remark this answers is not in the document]
ConradJChan: Yeah it's a typescript type guard so that typescript knows `totalCount` is a number in subsequent branches of logic

src/features/content-insights/CompactCount.tsx
ConradJChan: `classNames` gracefully handles falsey values when generating the class name

src/features/content-insights/ContentAnalyticsErrorState.scss
ConradJChan: Good catch, I should be able to flatten this

src/features/content-insights/ContentAnalyticsErrorState.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react';
+import { FormattedMessage } from 'react-intl';
+
+import HatWand140 from '../../illustration/HatWand140';
+import messages from './messages';
+
+import './ContentAnalyticsErrorState.scss';
+
+const ContentAnalyticsErrorState = () => (
```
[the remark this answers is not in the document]
ConradJChan: The type should be able to be inferred

src/features/content-insights/ContentInsightsSummary.scss
ConradJChan: It's possibly but the `margin` shorthand notations are fairly commonly used so I think it should be ok this way as well

greg-in-a-box (approved):
I only noticed a few things that were inconsistent but not enough to not approve the PR since I assume this is a bulk migration from somewhere else

nits: 
mixed usage of the following
1. `import * as React from 'react'` or `import React from 'react'`
2. sometimes using the the full classname instead of a `&`  in a `scss` file to indicate for a parent child style
```
.something {
    ...styles
}
.something-header {
    ...headerStyles
}

vs

.something {
    ...styles    

    &-header{
        ...headerStyles
    }
}
```

3. function vs arrow function for components

src/features/content-insights/_variables.scss
```diff
@@ -0,0 +1,13 @@
+@use "sass:color";
+
+@import '../../styles/variables';
+
+$ActivityList-event-spaceBetween: $bdl-grid-unit * 3;
+$ActivityList-header-spaceBetween: $bdl-grid-unit * 2;
+$ActivityList-eventGroupTitle-spaceAbove: $bdl-grid-unit * 4;
```
tjuanitas, line 7: what are these used for? is `ActivityList` in a separate PR?

tjuanitas: I think SUIT naming would use the property name instead of the value: https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md#variables

so `$ActivityList-event-justifyContent` although I don't think our repo currently follows a strict pattern for SCSS variables.

src/features/content-insights/charts/bar/Bar.tsx
```diff
@@ -0,0 +1,48 @@
...
+    React.useEffect(() => {
+        setStyle({
+            backgroundColor: color,
+            [cssProperty]: adjustedSize === 0 ? '2px' : `${adjustedSize}%`,
+        });
+    }, [adjustedSize, color, cssProperty]);
+
+    return (
+        <div
+            className={classNames('ca-Bar', { 'is-horizontal': isHorizontal })}
```
tjuanitas, line 38: this has a `ca` namespace, are we using that anywhere else? do we want to apply that for the rest of these components?

tjuanitas (commented):
where is this migrating from?

ConradJChan:
> where is this migrating from?

It's migrating from the webapp code

src/features/content-insights/CompactCount.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { injectIntl, IntlShape } from 'react-intl';
+import classNames from 'classnames';
+
+import numAbbr from '../../utils/numAbbr';
+
+interface Props {
```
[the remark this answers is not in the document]
ConradJChan: Is there a convention that we need to export props types? Seeing as how I'm not referencing the Prop types outside of the component currently, I didn't see the need

src/features/content-insights/ContentInsightsSummaryGhostState.tsx
```diff
@@ -0,0 +1,18 @@
...
+import Ghost from '../../components/ghost';
+
+import GraphCardGhostState from './GraphCardGhostState';
+
+import './ContentInsightsSummaryGhostState.scss';
+
+const ContentInsightsSummaryGhostState = () => (
+    <div className="ContentInsightsSummaryGhostState">
+        <GraphCardGhostState />
+        <div className="ContentInsightsSummaryGhostState-cta">
```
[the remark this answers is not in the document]
ConradJChan: Abbreviation of "call to action"

src/features/content-insights/ContentInsightsSummary.scss
```diff
@@ -0,0 +1,25 @@
+@import '../../styles/variables';
+
+.ContentInsightsSummary {
+    display: flex;
+    flex-direction: column;
+
+    .ContentInsightsSummary-header {
```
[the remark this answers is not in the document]
ConradJChan: Good catch, I think this is not used anymore

src/features/content-insights/ContentInsightsSummary.scss
```diff
@@ -0,0 +1,25 @@
...
+        position: relative;
+        margin: $bdl-grid-unit * 7.5 0 $bdl-grid-unit * 2.5;
+        padding-bottom: $bdl-grid-unit * 2.5;
+        color: $bdl-gray;
+        font-size: $bdl-fontSize;
+        border-bottom: 1px solid $bdl-gray-10;
+    }
+
+    .crawler {
```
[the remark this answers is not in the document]
ConradJChan: Good catch, I think this is not used anymore

src/features/content-insights/OpenContentInsightsButton.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+import Button, { ButtonType } from '../../components/button';
+
+import messages from './messages';
+
+interface Props {
+    onClick: () => void;
+}
+
+const OpenContentInsightsButton = ({ onClick }: Props) => {
```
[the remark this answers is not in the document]
ConradJChan: It will also be used in the webapp code

src/features/content-insights/_variables.scss
ConradJChan: Good catch, I think this can be removed

src/features/content-insights/charts/bar/Bar.tsx
ConradJChan: We aren't and we know there's some inconsistencies in the css naming right now but I don't really want to change too much from the original migrated components in this PR

src/features/content-insights/CompactCount.tsx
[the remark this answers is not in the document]
tjuanitas: our runbook for migrating components to TS  has a step for `Rename "Props" to "ComponentProps" in the new TSX files. Use a exported typescript interface for this.`

although, if we don't expect to use this prop type outside of this file, then I don't see a reason to make this update.

When the webapp is migrated to TypeScript, will it need to reference this type?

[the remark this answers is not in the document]
ConradJChan: Don't believe it will be referenced either

src/features/content-insights/charts/bar/Bar.tsx
tjuanitas: fair enough. since it shouldn't be used outside of the webapp, it should be safe to leave and address the inconsistencies at a later time

src/features/content-insights/TrendPill.scss
```diff
@@ -0,0 +1,33 @@
...
+.TrendPill {
+    display: flex;
+    align-items: center;
+    color: $bdl-gray-80;
+    font-weight: bold;
+    background-color: $bdl-gray-05;
+}
+
+.TrendPill-trend {
+    position: relative;
```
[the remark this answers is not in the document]
ConradJChan: Yeah, it affects the relative placement of the trend arrow

tjuanitas (approved):

src/features/content-insights/ContentAnalyticsErrorState.tsx
```diff
@@ -0,0 +1,18 @@
...
+    <div className="ContentAnalyticsErrorState" data-testid="ContentAnalyticsErrorState">
```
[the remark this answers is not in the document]
ConradJChan: tbh, I'm not sure there doesn't seem to be consistency within the `features` folder

greg-in-a-box (approved):

PR #3030, opened 2022-09-12
feat(tooltip): Set aria-hidden attribute on tooltip

joshmarnold:


src/components/tooltip/Tooltip.tsx
```diff
@@ -103,6 +103,8 @@ export type TooltipProps = {
     tetherElementClassName?: string;
     /** Text to show in the tooltip */
     text?: React.ReactNode;
+    /** Sets aria-hidden attribute on tooltip */
+    ariaHidden?: boolean;
```
tjuanitas, line 107: Could we add this into the list alphabetically?

tjuanitas (commented):
What's the use case for this? If we're just hiding the tooltip, could it be achieved using the `isShown` prop?

there was a lot of work done to make sure tooltips are accessible, can we confirm with web accessibility that this won't cause any issues?

let's makes sure it won't cause conflicts to meet these WCAG requirements:
https://github.com/box/box-ui-elements/pull/2544
https://github.com/box/box-ui-elements/pull/2818

joshmarnold:
@tjuanitas This feature was at the request of an accessibility member due to a case where we want this accessibility feature disabled. `isShown` will not work to disable this attribute as its used to toggle the showing of the tooltip and not set accessibility features.

tjuanitas (approved):

PR #3033, opened 2022-09-13
fix(activity-sidebar): omit retention field fetched by versions api

karelee7:
Omit retention from fields fetched by versions api in the activity feed for faster loading

**Currently with the retention field (takes ~35 seconds for the activity sidebar to load):**
Screen Shot 2022-09-13 at 12 45 40 PM

**Removing the retention field (takes ~15 seconds for the activity sidebar to load):**
Screen Shot 2022-09-13 at 12 49 09 PM


src/utils/fields.js
```diff
@@ -126,6 +126,25 @@ const PREVIEW_FIELDS_TO_FETCH = [
     FIELD_IS_DOWNLOAD_AVAILABLE,
 ];
 
+// Fields needed to get versions for a file in activity feed
+const FEED_FILE_VERSIONS_FIELDS_TO_FETCH = [
+    FIELD_AUTHENTICATED_DOWNLOAD_URL,
```
jstoffan, line 131: @karelee7, can we add a very brief comment next to the particularly expensive fields?

src/api/Feed.js
```diff
@@ -445,7 +448,14 @@ class Feed extends Base {
         this.versionsAPI = new VersionsAPI(this.options);
 
         return new Promise(resolve => {
-            this.versionsAPI.getVersions(this.file.id, resolve, this.fetchFeedItemErrorCallback.bind(this, resolve));
+            this.versionsAPI.getVersions(
+                this.file.id,
+                resolve,
+                this.fetchFeedItemErrorCallback.bind(this, resolve),
+                DEFAULT_FETCH_START,
```
jstoffan, line 455: Minor, but you could pass `undefined` here to get the same effect.

jstoffan (commented):
Great job to both you and @Jeff-Meadows. Let's make sure all the Version History sidebar features work after these changes.

karelee7:
Yes! The Version History sidebar looks good! (although because it's using these expensive fields, it's slow) I can choose to preview different versions, download different versions, change the current version, delete/restore versions, and I can successfully "go back" to the current version via that toolbar/header that pops up on the screen when you preview a previous version.

src/utils/fields.js
```diff
@@ -126,6 +126,25 @@ const PREVIEW_FIELDS_TO_FETCH = [
     FIELD_IS_DOWNLOAD_AVAILABLE,
 ];
 
+// Fields needed to get versions for a file in activity feed
+const FEED_FILE_VERSIONS_FIELDS_TO_FETCH = [
+    FIELD_AUTHENTICATED_DOWNLOAD_URL, // Increases load time for the activity feed
```
jstoffan, line 131: I think we want to remove these fields as well, but we should comment their use on `FILE_VERSIONS_FIELDS_TO_FETCH`.

karelee7: Ah yes, I was going to do that in a separate PR; just in case QA finds any issues that we missed with just removing the retention field. But I can include it here if that's what we want to do!

jstoffan: Let's do it all together, since the changes will roll into a single BUIE release anyway.

jstoffan (approved):
🚀

src/utils/fields.js
```diff
@@ -139,14 +156,14 @@ const FILE_VERSION_FIELDS_TO_FETCH = [
...
     FIELD_EXTENSION,
     FIELD_IS_DOWNLOAD_AVAILABLE,
     FIELD_MODIFIED_AT,
     FIELD_MODIFIED_BY,
     FIELD_NAME,
-    FIELD_PERMISSIONS,
+    FIELD_PERMISSIONS, // Expensive field to fetch
     FIELD_RESTORED_AT,
     FIELD_RESTORED_BY,
     FIELD_RETENTION,
```
ConradJChan, line 169: Does this need a comment about being an expensive field as well?

ConradJChan (commented):
Are there any unit tests that need to be updated?

src/utils/fields.js
karelee7: Ah yeah, good catch!

src/api/Feed.js
```diff
@@ -445,7 +446,14 @@ class Feed extends Base {
         this.versionsAPI = new VersionsAPI(this.options);
 
         return new Promise(resolve => {
-            this.versionsAPI.getVersions(this.file.id, resolve, this.fetchFeedItemErrorCallback.bind(this, resolve));
+            this.versionsAPI.getVersions(
```
ConradJChan, line 449: How about a unit test for the `Feed.js` to capture the updated fields requested?

src/api/__tests__/Feed.test.js
```diff
@@ -623,6 +624,18 @@ describe('api/Feed', () => {
                 expect(versionItems instanceof Promise).toBeTruthy();
                 expect(feed.versionsAPI.getVersions).toBeCalled();
             });
+
+            test('should call the versions api with the correct fields', () => {
+                feed.fetchVersions();
+                expect(feed.versionsAPI.getVersions).toBeCalledWith(
+                    '12345',
```
ConradJChan, line 631: Should this be `file.id`?

jstoffan (approved):

ConradJChan (approved):

PR #3034, opened 2022-09-13
feat(content-insights): Add error state to ContentInsightsSummary

JChan106:
In relation to https://git.dev.box.net/Box/EndUserApp/pull/8776

### DIFF
--- a/src/features/content-insights/ContentInsightsSummary.tsx
+++ b/src/features/content-insights/ContentInsightsSummary.tsx
@@ -11,16 +11,16 @@ import './ContentInsightsSummary.scss';
 
 interface Props {
     graphData: GraphData;
-    isError: AxiosError | null;
+    error: AxiosError | null;
     isLoading: boolean;
     onClick: () => void;
     previousPeriodCount: number;
     totalCount: number;
 }
 
-const ContentInsightsSummary = ({ isError, graphData, isLoading, previousPeriodCount, onClick, totalCount }: Props) => {
+const ContentInsightsSummary = ({ error, graphData, isLoading, previousPeriodCount, onClick, totalCount }: Props) => {
     const renderContentAnalyticsSummary = () => {
-        if (isError) {
+        if (error) {
             return <ContentAnalyticsErrorState />;
         }
 
--- a/src/features/content-insights/ContentInsightsSummary.tsx
+++ b/src/features/content-insights/ContentInsightsSummary.tsx
@@ -1,5 +1,4 @@
 import * as React from 'react';
-import { AxiosError } from 'axios';
 
 import ContentAnalyticsErrorState from './ContentAnalyticsErrorState';
 import ContentInsightsSummaryGhostState from './ContentInsightsSummaryGhostState';
@@ -11,7 +10,7 @@ import './ContentInsightsSummary.scss';
 
 interface Props {
     graphData: GraphData;
-    error: AxiosError | null;
+    error: Object | null;
     isLoading: boolean;
     onClick: () => void;
     previousPeriodCount: number;
--- a/src/features/content-insights/ContentInsightsSummary.tsx
+++ b/src/features/content-insights/ContentInsightsSummary.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 
+import ContentAnalyticsErrorState from './ContentAnalyticsErrorState';
 import ContentInsightsSummaryGhostState from './ContentInsightsSummaryGhostState';
 import GraphCardPreviewsSummary from './GraphCardPreviewsSummary';
 import OpenContentInsightsButton from './OpenContentInsightsButton';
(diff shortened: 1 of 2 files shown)


src/features/content-insights/ContentInsightsSummary.tsx
```diff
@@ -9,17 +10,24 @@ import './ContentInsightsSummary.scss';
 
 interface Props {
     graphData: GraphData;
+    isError: boolean;
     isLoading: boolean;
     onClick: () => void;
```
karelee7, line 15: minor: I think you can also type it as `Function`; although, both patterns seem to exist so not sure which one is more correct ¯\_(ツ)_/¯

src/features/content-insights/ContentInsightsSummary.tsx
```diff
@@ -9,17 +10,24 @@ import './ContentInsightsSummary.scss';
 
 interface Props {
     graphData: GraphData;
+    isError: boolean;
```
ConradJChan, line 13: I'm wondering if this should be an object because we'll need to differentiate the permission error vs API error scenarios

src/features/content-insights/ContentInsightsSummary.tsx
ConradJChan: I think this type is actually more specific and better than a general `Function`

src/features/content-insights/ContentInsightsSummary.tsx
JChan106: Good point. We use AxiosError in EUA, should we try and use the same typings maybe? Or we can be less explicit and use just an object for now.

ConradJChan: Yeah seems like `AxiosError` exists in BUIE as well so let's go with that for now

src/features/content-insights/__tests__/ContentInsightsSummary.test.tsx
```diff
@@ -38,9 +41,28 @@ describe('features/content-insights/ContentInsightsSummary', () => {
...
-        test('should render correctly when isLoading is false', () => {
+        test('should render correctly when isLoading is false and error is null', () => {
+            mockError.mockReturnValueOnce(null);
+            getWrapper();
+
+            expect(screen.queryByTestId('ContentAnalyticsErrorState')).toBeNull();
+            expect(screen.queryByTestId('GraphCardGhostState')).toBeNull();
+            expect(screen.getByLabelText(localize(messages.previewGraphLabel.id))).toBeVisible();
+        });
+        test('should show the error state when isError is true', () => {
```
karelee7, line 52: Should this say `error` instead of `isError`? Seems like isError was updated/removed from the props?

src/features/content-insights/ContentInsightsSummary.tsx
```diff
@@ -8,18 +10,27 @@ import { GraphData } from './types';
...
     totalCount: number;
 }
 
-const ContentInsightsSummary = ({ graphData, isLoading, previousPeriodCount, onClick, totalCount }: Props) => (
-    <div className="ContentInsightsSummary">
-        {isLoading ? (
-            <ContentInsightsSummaryGhostState />
-        ) : (
+const ContentInsightsSummary = ({ error, graphData, isLoading, previousPeriodCount, onClick, totalCount }: Props) => {
+    const errorStatus = !!error && !!error.response ? error.response.status : null;
```
ConradJChan, line 22: I don't think we need to overthink this at the moment -- just convert the error to a `isError` and show the error state. In a follow up we can probably pass the error into `ContentAnalyticsErrorState` and have that affect the message displayed

src/features/content-insights/ContentInsightsSummary.tsx
```diff
@@ -9,17 +11,24 @@ import './ContentInsightsSummary.scss';
 
 interface Props {
     graphData: GraphData;
+    isError: AxiosError | null;
```
ConradJChan, line 14: Can we make this `error` and derive `isError` in the component below?

src/features/content-insights/__tests__/ContentInsightsSummary.test.tsx
```diff
@@ -17,11 +17,14 @@ const mockData = [
     { start: 7, previewsCount: 1, type: 'day' },
 ] as GraphData;
 
+const mockError = jest.fn();
```
ConradJChan, line 20: Does this need to be a mock? Can we pass the `error` as appropriate into the `getWrapper`?

JChan106: We would have to mock an axios error with different fields, can't pass in a regular error

src/features/content-insights/ContentInsightsSummary.tsx
```diff
@@ -9,17 +11,24 @@ import './ContentInsightsSummary.scss';
 
 interface Props {
     graphData: GraphData;
+    error: AxiosError | null;
```
ConradJChan, line 14: I think this actually isn't an `AxiosError`, could we just type it as `Object`?

src/features/content-insights/__tests__/ContentInsightsSummary.test.tsx
ConradJChan: See my comment about the error type, that should likely simplify the typing situation

src/features/content-insights/ContentInsightsSummary.tsx
```diff
@@ -9,17 +10,24 @@ import './ContentInsightsSummary.scss';
 
 interface Props {
     graphData: GraphData;
+    error: Object | null;
```
ConradJChan, line 13: Can we change this to `error?: Object`?

ConradJChan (approved):

PR #3038, opened 2022-09-19
feat(content-sidebar): added replies related logic for threaded replies

kkuliczkowski-box:
- Created `Feed` method to create Annotation or Comment reply
- Created handler in `ActivitySidebar`
-- Passed that handler to `ActivityThread`
- Added missing helper functions


src/api/utils.js
```diff
@@ -0,0 +1,30 @@
...
+    };
+
+    if (comment.replies && comment.replies.length) {
+        formattedComment.replies = comment.replies.map(formatComment);
+    }
+
+    return formattedComment;
+};
+
+export default {
```
JChan106, line 28: Do we normally export default and also a named export?

src/api/Annotations.js
```diff
@@ -24,9 +25,26 @@ import type {
...
 
 export default class Annotations extends MarkerBasedApi {
+    /**
+     * Formats annotation replies' comment data for use in components.
+     *
+     * @param {Annotation} annotation - An individual annotation entry from the API
+     * @return {Annotation} Updated annotation
+     */
+    format(annotation: Annotation): Annotation {
+        if (annotation.replies && annotation.replies.length) {
```
ConradJChan, line 38: Could we re-arrange this as a guard clause?
```
if (!annotation.replies || !annotation.replies.length) {
  return annotation;
}
```

src/api/Annotations.js
```diff
@@ -24,9 +25,26 @@ import type {
...
+import type { Comment, FeedItemStatus, ThreadedComments as ThreadedCommentsType } from '../common/types/feed';
 
 export default class Annotations extends MarkerBasedApi {
+    /**
+     * Formats annotation replies' comment data for use in components.
+     *
+     * @param {Annotation} annotation - An individual annotation entry from the API
+     * @return {Annotation} Updated annotation
+     */
+    format(annotation: Annotation): Annotation {
```
ConradJChan, line 37: Should we rename this to `formatReplies`?

src/api/Feed.js
```diff
@@ -1178,6 +1180,58 @@ class Feed extends Base {
...
+        const date = new Date().toISOString();
+        const pendingReply: Comment = {
+            created_at: date,
+            created_by: currentUser,
+            modified_at: date,
+            isPending: true,
+            ...commentBase,
+        };
+
+        const getItemWithPendingReply = <T: { replies?: Array<Comment> }>(item: T, reply: Comment): T => {
```
ConradJChan, line 1208: Does this need to be inlined in `addPendingReply`?

src/api/Feed.js
```diff
@@ -1220,6 +1274,53 @@ class Feed extends Base {
...
+    /**
+     * Callback for failed creation of a reply.
+     *
+     * @param {Object} e - The axios error
+     * @param {string} code - the error code
+     * @param {string} parentId - ID of the parent feed item
+     * @param {string} id - ID of the feed item to update
+     * @return {void}
+     */
+    createReplyErrorCallback = (e: ElementsXhrError, code: string, parentId: string, id: string) => {
```
ConradJChan, line 1315: Can we use `error` instead of `e`?

src/api/Feed.js
```diff
@@ -1265,6 +1366,55 @@ class Feed extends Base {
...
 
+    /**
+     * Replace a reply of feed item with new comment data.
+     *
+     * @param {Object} replyUpdates - New data to be applied to the reply.
+     * @param {string} parentId - ID of the parent feed item.
+     * @param {string} id - ID of the reply to replace.
+     * @return {void}
+     */
+    updateReplyItem = (replyUpdates: Object, parentId: string, id: string) => {
```
ConradJChan, line 1377: Should this also set `isPending` on the reply when updating the reply message?

src/api/Feed.js
```diff
@@ -1178,6 +1180,58 @@ class Feed extends Base {
...
+        if (cachedItems) {
+            const updatedFeedItems = cachedItems.items.map(item => {
+                if (item.id === parentId && item.type === 'comment') {
+                    return getItemWithPendingReply<Comment>(item, pendingReply);
+                }
+                if (item.id === parentId && item.type === 'annotation') {
+                    return getItemWithPendingReply<Annotation>(item, pendingReply);
+                }
+                return item;
+            });
```
ConradJChan, line 1227: Do we need the typed helper method? Could it be something like
```
cachedItems.items.map(item => {
    if (item.id !== parentId) {
        return item;
    }

    return {
        ...item,
        replies: item.replies ? [...item.replies, pendingReply] : [pendingReply],
    }
})
```

Similarly, for the `updateReplyItem` method

src/api/Feed.js
```diff
@@ -1369,6 +1519,77 @@ class Feed extends Base {
...
+        };
+
+        this.file = file;
+        this.errorCallback = errorCallback;
+        this.addPendingReply(parentId, currentUser, commentData);
+
+        const successCallbackFn = (comment: Comment) => {
+            this.createReplySuccessCallback(comment, parentId, uuid, successCallback);
+        };
+        const errorCallbackFn = (e: ErrorResponseData, code: string) => {
```
karelee7, line 1564: Is there a better name for this instead of `e`? Perhaps something a bit more descriptive?

src/api/Annotations.js
```diff
@@ -166,7 +186,9 @@ export default class Annotations extends MarkerBasedApi {
         this.get({
             id: fileId,
             errorCallback,
-            successCallback,
+            successCallback: (annotation: Annotation) => {
```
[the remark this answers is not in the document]
kkuliczkowski-box: I've created `annotationSuccessCallback` method that is now used in both `updateAnnotation` and `getAnnotation`.

src/api/utils.js
kkuliczkowski-box: I wanted to keep named export as it seems to make more sense when importing particular utilities, but current ESLint rules enforce using default export (`import/prefer-default-export`).

src/api/Annotations.js
kkuliczkowski-box: Makes sense. Renamed.

src/api/Feed.js
kkuliczkowski-box: It is used only in `addPendingReply` method and its purpose is nothing more then be a helper function for this method. Do you think I should move it outside of the Class?

src/api/Feed.js
kkuliczkowski-box: Sure, renamed.

src/api/Feed.js
kkuliczkowski-box: This sets whatever is passed in `replyUpdates` param. It works in similar way as `updateFeedItem`. In future PR I will create a handler for updating replies that will utilize `updateReplyItem` and will also pass `isPending: true`.

src/api/Feed.js
```diff
@@ -63,12 +63,14 @@ import type {
     AppActivityItems,
     BoxCommentPermission,
     Comment,
+    CommentFeedItemType,
```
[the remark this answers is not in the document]
kkuliczkowski-box: Shouldn't we keep alphabetical order?

src/api/Feed.js
kkuliczkowski-box: I've renamed it to `error`.

src/api/Annotations.js
```diff
@@ -115,7 +133,7 @@ export default class Annotations extends MarkerBasedApi {
                 },
             },
             errorCallback,
-            successCallback,
+            successCallback: this.annotationSuccessCallback.bind(this, successCallback),
```
ConradJChan, line 136: Should we bind this once in the constructor?

src/api/utils.js
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Formats comment data (including replies) for use in components.
+ *
+ * @param {Comment} comment - An individual comment entry from the API
+ * @return {Comment} Updated comment
+ */
+export const formatComment = (comment: Comment): Comment => {
+    const formattedComment = {
+        ...comment,
+        tagged_message: comment.message,
```
ConradJChan, line 18: Curious as to why the comments endpoint doesn't return a `tagged_message` field?

src/api/Annotations.js
kkuliczkowski-box: I'm not sure how would this work. I need to bind the `successCallback` (which is given as param in `updateAnnotation`) as the first param in order to invoke it in `annotationSuccessCallback`.

src/api/utils.js
kkuliczkowski-box: I believe that was the decision made by API Working Group to abandon the legacy tagged_message field. Not using it in the new API for Threaded Comments was also advocated by @jstoffan.

src/api/Feed.js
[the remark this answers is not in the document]
karelee7: Ah yes you're right, you can ignore this comment!

src/api/Feed.js
```diff
@@ -1220,6 +1274,53 @@ class Feed extends Base {
...
+
+        if (!this.isDestroyed()) {
+            successCallback(commentData);
+        }
+    };
+
+    /**
+     * Callback for failed creation of a reply.
+     *
+     * @param {Object} e - The axios error
```
karelee7, line 1309: should this be updated to `error`?

karelee7: Should the param type be `ElementsXhrError`? Or does `Object` suffice??

kkuliczkowski-box: That's correct. I've updated JSDoc so that it's aligned with Flow.

ConradJChan (approved):

PR #3040, opened 2022-09-20
feat(content-sidebar): Add SidebarContentInsights for Details Sidebar

ConradJChan:
Screen Shot 2022-09-20 at 3 27 59 PM


karelee7 (approved):
Looks good to me!

scripts/webpack.config.js
```diff
@@ -91,7 +91,7 @@ function getConfig(isReactExternalized) {
                     // For webpack dev build perf we want to exlcude node_modules unless we want to support legacy browsers like IE11
                     exclude: shouldIncludeAllSupportedBrowsers
                         ? /@babel(?:\/|\\{1,2})runtime|pikaday|core-js/
-                        : /node_modules/,
+                        : /node_modules\/(?!@box\/cldr-data)/, // Exclude node_modules except for @box/cldr-data which is needed for styleguidist
```
jstoffan, line 94: What about this particular change necessitated this new configuration?

src/elements/content-sidebar/SidebarContentInsights.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { FormattedMessage } from 'react-intl';
+
+import ContentInsightsSummary from '../../features/content-insights/ContentInsightsSummary';
+import { GraphData } from '../../features/content-insights/types';
+// @ts-ignore Module is written in Flow
+import messages from '../common/messages';
+// @ts-ignore Module is written in Flow
+import SidebarSection from './SidebarSection';
+// @ts-ignore Module is written in Flow
+// eslint-disable-next-line import/no-named-as-default, import/no-named-as-default-member
```
jstoffan, line 12: We should ideally try to avoid ever disabling our linting rules. It looks like we have other components that import this module without issue. What was the reasoning here?

src/elements/content-sidebar/SidebarContentInsights.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+interface ContentInsights {
+    error?: Object;
+    graphData: GraphData;
+    isLoading: boolean;
+    previousPeriodCount: number;
+    totalCount: number;
+}
+
+interface Props {
```
jstoffan, line 23: Should we export these interfaces?

src/elements/content-sidebar/SidebarContentInsights.tsx
```diff
@@ -0,0 +1,67 @@
...
+}: Props) => {
+    const [hasMounted, setHasMounted] = React.useState(false);
+    const { graphData, isLoading, previousPeriodCount, totalCount } = contentInsights;
+
+    React.useEffect(() => {
+        if (hasMounted) {
+            return;
+        }
+
+        onContentInsightsMount();
```
jstoffan, line 50: What was the reasoning for exposing the internal lifecycle here, again?

src/elements/content-sidebar/SidebarContentInsights.tsx
```diff
@@ -0,0 +1,67 @@
...
+// @ts-ignore Module is written in Flow
+import messages from '../common/messages';
+// @ts-ignore Module is written in Flow
+import SidebarSection from './SidebarSection';
+// @ts-ignore Module is written in Flow
+// eslint-disable-next-line import/no-named-as-default, import/no-named-as-default-member
+import withErrorHandling from './withErrorHandling';
+
+interface ContentInsights {
+    error?: Object;
```
jstoffan, line 16: Do we have a better type for this object?

scripts/webpack.config.js
ConradJChan: No other code in the repo had actually called [numAbbr](https://github.com/box/box-ui-elements/blob/master/src/utils/numAbbr.ts)

src/elements/content-sidebar/SidebarContentInsights.tsx
ConradJChan: So that the Content Insights summary in the details sidebar can initiate a call to fetch the data if none has yet been cached

jstoffan: Got it, thanks. Should this be handled in `DetailsSidebar` like it is for access stats? It also supports a `refresh` method, that does something similar on-demand.

src/elements/content-sidebar/SidebarContentInsights.tsx
ConradJChan: The issue appears to be with the tooling, eslint seems to choke on the `type` import in `withErrorHandling` and it seems like I can work around it by moving the file to a `hocs` folder and providing an `index.js` but refactoring  that code just to solve this typescript/flow import issue seems like overkill. I'll try to add a comment to describe why these linting rules are disabled

src/elements/content-sidebar/SidebarContentInsights.tsx
ConradJChan: Sure, let me explore this option

scripts/webpack.config.js
ConradJChan: It increases the bundle size for the sidebar.js from `1.57MB` to `1.61.MB` which seems ok. It should work with both versions of react-intl because it shortcuts both and uses `IntlMessageFormat` to abbreviate the number and when we fully support react-intl v3 we can actually migrate over to using `<FormattedNumber />` with the `compact` notation

jstoffan (dismissed):

jstoffan (commented):
Do we need to add a call to `refresh`?

ConradJChan:
> Do we need to add a call to `refresh`?

There's no use case for it currently, but it's something we can keep in mind for the future if required

jstoffan (approved):

PR #3042, opened 2022-09-21
feat(content-insights): Add permission error state for 403 errors

JChan106:
Adds error state for permission errors

Updated error states: 

image
image

image
image

image
image

### DIFF
--- a/src/features/content-insights/ContentAnalyticsErrorState.scss
+++ b/src/features/content-insights/ContentAnalyticsErrorState.scss
@@ -6,6 +6,27 @@
     align-items: center;
     justify-content: center;
     height: 100%;
+
+    &.small {
+        svg {
+            width: 56px;
+            height: 56px;
+        }
+    }
+
+    &.medium {
+        svg {
+            width: 80px;
+            height: 80px;
+        }
+    }
+
+    &.large {
+        svg {
+            width: 95px;
+            height: 95px;
+        }
+    }
 }
 
 .ContentAnalyticsErrorState-text,
--- a/src/features/content-insights/ContentAnalyticsErrorState.scss
+++ b/src/features/content-insights/ContentAnalyticsErrorState.scss
@@ -4,9 +4,12 @@
     display: flex;
     flex-direction: column;
     align-items: center;
+    justify-content: center;
+    height: 100%;
 }
 
-.ContentAnalyticsErrorState-text {
+.ContentAnalyticsErrorState-text,
+.ContentAnalyticsPermissionError-text {
     width: 100%;
     margin-top: $bdl-grid-unit * 5;
     font-size: $bdl-fontSize--dejaBlue;
--- a/src/features/content-insights/ContentAnalyticsErrorState.scss
+++ b/src/features/content-insights/ContentAnalyticsErrorState.scss
@@ -6,31 +6,10 @@
     align-items: center;
     justify-content: center;
     height: 100%;
-
-    &.small {
-        svg {
-            width: 56px;
-            height: 56px;
-        }
-    }
-
-    &.medium {
-        svg {
-            width: 80px;
-            height: 80px;
-        }
-    }
-
-    &.large {
-        svg {
-            width: 95px;
-            height: 95px;
-        }
-    }
 }
 
 .ContentAnalyticsErrorState-text,
-.ContentAnalyticsPermissionError-text {
+.ContentAnalyticsErrorState-text--permission {
     width: 100%;
     margin-top: $bdl-grid-unit * 5;
     font-size: $bdl-fontSize--dejaBlue;
--- a/src/features/content-insights/ContentAnalyticsErrorState.tsx
+++ b/src/features/content-insights/ContentAnalyticsErrorState.tsx
@@ -2,45 +2,52 @@ import * as React from 'react';
 import { FormattedMessage } from 'react-intl';
 
 import PropTypes from 'prop-types';
-import HatWand140 from '../../illustration/HatWand140';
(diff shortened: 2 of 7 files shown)


src/features/content-insights/ContentInsightsSummary.tsx
```diff
@@ -42,4 +44,8 @@ const ContentInsightsSummary = ({ error, graphData, isLoading, previousPeriodCou
     return <div className="ContentInsightsSummary">{renderContentAnalyticsSummary()}</div>;
 };
 
+ContentInsightsSummary.propTypes = {
```
ConradJChan, line 47: Do we need to add this if we already have typescript for typing?

src/features/content-insights/ContentInsightsSummary.tsx
```diff
@@ -20,7 +21,8 @@ interface Props {
 const ContentInsightsSummary = ({ error, graphData, isLoading, previousPeriodCount, onClick, totalCount }: Props) => {
     const renderContentAnalyticsSummary = () => {
         if (error) {
-            return <ContentAnalyticsErrorState />;
+            const isPermissionError = error.status === 403;
+            return <ContentAnalyticsErrorState isPermissionError={isPermissionError} />;
```
ConradJChan, line 25: Could we pass the error directly into `ContentAnalyticsErrorState` component and let it figure out what to show?

ConradJChan:
The Permission error icon looks pretty small, is that the desired size?

src/features/content-insights/types.ts
```diff
@@ -23,3 +23,11 @@ export type MetricsData = {
 export type GraphDatum = TimeRange & MetricsData;
 
 export type GraphData = Array<GraphDatum>;
+
+export type ResponseError = Error & {
```
ConradJChan, line 27: Does it need to include the `Error` type?

JChan106: The error type has fields `message` and `stack`, I found both those fields to exist when taking a look at a 500 error I made in the accessstats api

src/features/content-insights/ContentInsightsSummary.tsx
JChan106: Yeah we can do that

src/features/content-insights/ContentInsightsSummary.tsx
JChan106: I was getting this `"status" is missing in props validation` warning. I found this to be the only way to remove it

src/features/content-insights/ContentInsightsSummary.tsx
JChan106: > The Permission error icon looks pretty small, is that the desired size?

Mocks seem to have it ranging from 42 - 72 depending on where its at, currently default is 56

ConradJChan:
As discussed, let's update the icon and handle the two different sizes in this repository

src/features/content-insights/ContentAnalyticsErrorState.tsx
```diff
@@ -1,18 +1,59 @@
...
+import { ResponseError } from './types';
 
 import './ContentAnalyticsErrorState.scss';
 
-const ContentAnalyticsErrorState = () => (
-    <div className="ContentAnalyticsErrorState" data-testid="ContentAnalyticsErrorState">
-        <HatWand140 data-testid="ContentAnalyticsErrorState-image" />
-        <div className="ContentAnalyticsErrorState-text">
-            <FormattedMessage {...messages.contentAnalyticsErrorText} />
+const SIZE_MAP = {
```
ConradJChan, line 11: Instead of javascript, can we apply the styles via CSS in each use case? i.e. in `ContentInsightsSummary.scss` style it to be the small size, and in the flyout and modal, style it as appropriate

src/features/content-insights/ContentAnalyticsErrorState.tsx
```diff
@@ -1,18 +1,59 @@
...
+}
+
+const ContentAnalyticsErrorState = ({ error, size = 'small' }: Props) => {
+    const renderErrorMessage = () => {
+        const isPermissionError = error.status === 403;
+
+        if (isPermissionError) {
+            return (
+                <div
+                    className="ContentAnalyticsPermissionError-text"
```
tjuanitas, line 28: I think you would update the descendant or use a modifier in this case:

```suggestion
                    className="ContentAnalyticsErrorState-permissionText"
```
```suggestion
                    className="ContentAnalyticsErrorState-text--permission"
```

src/features/content-insights/ContentAnalyticsErrorState.scss
```diff
@@ -4,9 +4,33 @@
...
+            height: 80px;
+        }
+    }
+
+    &.large {
+        svg {
+            width: 95px;
+            height: 95px;
+        }
+    }
```
ConradJChan, line 29: Are these still needed?

src/features/content-insights/ContentAnalyticsErrorState.tsx
```diff
@@ -1,18 +1,49 @@
...
+                    <FormattedMessage {...messages.contentAnalyticsPermissionError} />
+                </div>
+            );
+        }
+
+        return (
+            <div className="ContentAnalyticsErrorState-text" data-testid="ContentAnalyticsErrorState-text">
+                <FormattedMessage {...messages.contentAnalyticsErrorText} />
+            </div>
+        );
```
ConradJChan, line 34: Do we need separate divs or can we change which formatted message is used?

karelee7 (approved):

PR #3043, opened 2022-09-22
feat(content-sidebar): added missing replies logic for threaded replies

kkuliczkowski-box:
- Created `Feed` methods for
-- fetching replies
-- updating reply
-- deleting reply
- Created handlers in `ActivitySidebar`
-- Passed handlers to `ActivityThread`

### DIFF
--- a/src/api/Feed.js
+++ b/src/api/Feed.js
@@ -1289,25 +1289,23 @@ class Feed extends Base {
         parentId: string,
         successCallback: (id: string, parentId: string) => void = noop,
     ) => {
-        const cachedItems = this.getCachedItems(this.file.id);
-        if (cachedItems) {
-            const feedItems = cachedItems.items.map(item => {
-                if (item.id !== parentId) {
-                    return item;
-                }
-                if (item.type === 'annotation') {
-                    return getItemWithFilteredReplies<Annotation>(item, id);
-                }
-                if (item.type === 'comment') {
-                    return getItemWithFilteredReplies<Comment>(item, id);
-                }
+        const cachedItems = this.getCachedItems(this.file.id) || { items: [], errors: [] };
+        const feedItems = cachedItems.items.map(item => {
+            if (item.id !== parentId) {
                 return item;
-            });
-            this.setCachedItems(this.file.id, feedItems);
-
-            if (!this.isDestroyed()) {
-                successCallback(id, parentId);
             }
+            if (item.type === 'annotation') {
+                return getItemWithFilteredReplies<Annotation>(item, id);
+            }
+            if (item.type === 'comment') {
+                return getItemWithFilteredReplies<Comment>(item, id);
+            }
+            return item;
+        });
+        this.setCachedItems(this.file.id, feedItems);
+
+        if (!this.isDestroyed()) {
+            successCallback(id, parentId);
         }
     };
 
--- a/src/api/Feed.js
+++ b/src/api/Feed.js
@@ -239,6 +239,33 @@ class Feed extends Base {
         this.feedErrorCallback(true, e, code);
     };
 
+    /**
+     * Error callback for updating a reply
+     *
+     * @param {ElementsXhrError} error - the error returned by the API
+     * @param {string} code - the error code
+     * @param {string} id - the id of the reply (comment)
+     * @param {string} parentId - the id of either the parent item (an annotation or comment)
+     * @return {void}
+     */
+    updateReplyErrorCallback = (error: ElementsXhrError, code: string, id: string, parentId: string) => {
+        this.updateReplyItem(this.createFeedError(messages.commentUpdateErrorMessage), parentId, id);
+        this.feedErrorCallback(true, error, code);
+    };
+
+    /**
+     * Error callback for fetching replies
+     *
+     * @param {ElementsXhrError} error - the error returned by the API
+     * @param {string} code - the error code
+     * @param {string} id - the id of either an annotation or comment
+     * @return {void}
+     */
+    fetchRepliesErrorCallback = (error: ElementsXhrError, code: string, id: string) => {
+        this.updateFeedItem(this.createFeedError(messages.repliesFetchErrorMessage), id);
+        this.feedErrorCallback(true, error, code);
+    };
+
     deleteAnnotation = (
         file: BoxItem,
         annotationId: string,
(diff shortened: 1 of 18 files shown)


src/api/Feed.js
```diff
@@ -1726,6 +1898,64 @@ class Feed extends Base {
...
+     * @return {void}
+     */
+    updateReply = (
+        file: BoxItem,
+        id: string,
+        parentId: string,
+        text: string,
+        hasMention?: boolean,
+        permissions: BoxCommentPermission,
+        successCallback: Function,
```
JChan106, line 1921: Should we use `() => void` typing?

src/api/Feed.js
```diff
@@ -1128,6 +1261,45 @@ class Feed extends Base {
...
+            return { ...item, replies };
+        };
+
+        if (cachedItems) {
+            const feedItems = cachedItems.items.map(item => {
+                if (item.id !== parentId) {
+                    return item;
+                }
+                if (item.type === 'annotation') {
+                    return getItemWithFilteredReplies<Annotation>(item, id);
```
jstoffan, line 1288: This looks similar to several other methods/helpers in this class. How much logic is left remaining after this change to complete our feature set? I think a refactor may be in order because this file is getting pretty enormous.

src/api/Feed.js
```diff
@@ -439,6 +466,58 @@ class Feed extends Base {
...
+
+        const successCallbackFn = (comments: ThreadedCommentsType) => {
+            this.updateFeedItem({ isRepliesLoading: false, replies: comments.entries }, itemId);
+            successCallback(comments.entries);
+        };
+        const errorCallbackFn = (e: ErrorResponseData, code: string) => {
+            this.fetchRepliesErrorCallback(e, code, itemId);
+        };
+
+        if (itemType === 'annotation') {
```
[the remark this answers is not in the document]
kkuliczkowski-box: I think that's a very good idea, however I believe it should be done later as a separate change. Because it will affect many files, it might obscure the goal of this change. Also I think it needs to be done thoroughly to replace all instances of hardcoded types (all of them) and because of lack of time, should not be done right now.

src/elements/content-sidebar/activity-feed/activity-feed/ActivityThread.js
```diff
@@ -45,6 +60,8 @@ const ActivityThread = ({
     onGetReplies = noop,
     onReplyDelete = noop,
     onReplyEdit = noop,
+    parentId,
```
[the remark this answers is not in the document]
kkuliczkowski-box: That's correct, both `parentId` and `parentType` were not used besides being passed in the handler functions. Refactored.

src/api/Feed.js
```diff
@@ -439,6 +466,58 @@ class Feed extends Base {
...
+     * @param {BoxItem} file - The file to which the comment or annotation belongs to
+     * @param {string} itemId - ID of the comment or annotation
+     * @param {CommentFeedItemType} itemType - Type of the comment or annotation
+     * @param {Function} successCallback
+     * @param {ErrorCallback} errorCallback
+     * @return {void}
+     */
+    fetchReplies(
+        file: BoxItem,
+        itemId: string,
```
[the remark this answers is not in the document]
kkuliczkowski-box: These were usually referred to as "feedItems", but in this case we only expect a comment or annotation. I have created a type `CommentFeedItemType` before for the type property that can only be `comment` or `annotation`. To be consistent I've renamed arguments to `commentFeedItemId` and `commentFeedItemType`.

src/api/Feed.js
```diff
@@ -1128,6 +1261,45 @@ class Feed extends Base {
...
+     * Deletes a reply from the cache
+     *
+     * @param {string} id - The id of the feed item to be deleted
+     * @param {string} parentId - The id of the parent feed item
+     * @param {Function} successCallback - function to be called after the delete
+     */
+    deleteReplyItem = (id: string, parentId: string, successCallback: Function = noop) => {
+        const cachedItems = this.getCachedItems(this.file.id);
+
+        const getItemWithFilteredReplies = <T: { replies?: Array<Comment> }>(item: T, replyId: string): T => {
```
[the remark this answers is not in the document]
kkuliczkowski-box: I've extracted it outside of the class. Did the same with similar `getItemWithPendingReply` helper function.

src/api/Feed.js
```diff
@@ -1128,6 +1261,45 @@ class Feed extends Base {
...
+            if (!item.replies) {
+                return item;
+            }
+            const replies = item.replies.filter(reply => reply.id !== replyId);
+            return { ...item, replies };
+        };
+
+        if (cachedItems) {
+            const feedItems = cachedItems.items.map(item => {
```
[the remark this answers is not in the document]
kkuliczkowski-box: The separate cache for replies seems a good idea, but I'm not sure if it would be a benefit at this point.

This is my understanding: I think that introducing separate cache for replies would be a complication at this point as the replies are already nested in the initial server response when fetching comments or annotations (last reply is in the replies array) and it's already inserted into the feed items cache. I would need to modify the `feedItems` method and go through all the feed items in order to fill initial value of the `repliesCache`.  In addition I would need to modify the logic of returning the cached feed items by going through all feed items and inserting the replies from the `repliesCache` so that the returned collection can be set as a new value of `feedItems` state (in `ActivitySidebar`). As this would need to be done on every change in the replies (`feedItems` state needs to change to visualize the change in UI) I don't think there will be difference vs the current implementation because mapping the whole collection would need to happen every time state would need to be updated anyway.
I think that having a separate cache for replies would make sense if we would also store the replies as a separate state.

Please let me know if my understanding is incorrect and you had a different solution in mind.

src/api/Feed.js
kkuliczkowski-box: I agree, `Feed` got pretty big and future refactoring is a good idea. Actually this PR is the last set of new methods and I don't anticipate adding more. Some changes and fixes might be needed when finalizing and testing FE, BFF and Domain Service takes place very shortly, but we shouldn't add anything new.

src/api/Feed.js
```diff
@@ -1128,6 +1277,40 @@ class Feed extends Base {
...
+        const cachedItems = this.getCachedItems(this.file.id);
+        if (cachedItems) {
+            const feedItems = cachedItems.items.map(item => {
+                if (item.id !== parentId) {
+                    return item;
+                }
+                if (item.type === 'annotation') {
+                    return getItemWithFilteredReplies<Annotation>(item, id);
+                }
+                if (item.type === 'comment') {
```
jstoffan, line 1301: We should probably have a constant for these type values.

src/api/Feed.js
```diff
@@ -1128,6 +1277,40 @@ class Feed extends Base {
...
+     * @param {string} parentId - The id of the parent feed item
+     * @param {Function} successCallback - function to be called after the delete
+     */
+    deleteReplyItem = (
+        id: string,
+        parentId: string,
+        successCallback: (id: string, parentId: string) => void = noop,
+    ) => {
+        const cachedItems = this.getCachedItems(this.file.id);
+        if (cachedItems) {
```
jstoffan, line 1293: Minor, but it seems like we could remove some nesting here with a change to the line above:

```
const cachedItems = this.getCachedItems(this.file.id) || [];

File: src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js [hunk 10]
```diff
@@ -305,7 +319,10 @@ class ActivityFeed extends React.Component<Props, State> {
                             onAppActivityDelete={onAppActivityDelete}
                             onCommentDelete={hasCommentPermission ? onCommentDelete : noop}
                             onCommentEdit={hasCommentPermission ? onCommentUpdate : noop}
+                            onShowReplies={onShowReplies}
```
jstoffan, line 322: Minor, but can we alphabetize these props?

jstoffan (dismissed):

src/api/Feed.js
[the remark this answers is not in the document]
jstoffan: I think your understanding is correct. Ideally, the components would be updated to retrieve the correct collection of replies based on the feedItemId passed into them rather than denormalizing the state again. This pattern is very common in [Redux](https://redux.js.org/usage/structuring-reducers/normalizing-state-shape) but can be valuable outside of it, as well.

Since this is the last change we anticipate to the Feed API, we don't need to tackle this now. I do think refactoring this area in the future and normalizing the state would help pull apart some of the spaghetti we're seeing, though.

jstoffan (approved):

src/api/Feed.js
kkuliczkowski-box: The default `cachedItems` actually needs to be `{ items: [], errors: [] }`, but I have made he change.

jstoffan (approved):

PR #3048, opened 2022-09-23
fix(infinite-scroll): fix documentation and prop name mismatch

QingyuChai:
One of the props for `infinite-scroll` is called `scrollContainerNode`, but is sometimes referred to as `scrollContainerRef`. Fixed for consistency.

### DIFF
--- a/src/components/infinite-scroll/InfiniteScroll.md
+++ b/src/components/infinite-scroll/InfiniteScroll.md
@@ -58,7 +58,7 @@ const hasMore = true;
       isLoading={isLoading}
       onLoadMore={onLoadMore}
       useWindow={false}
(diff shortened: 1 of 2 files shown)


QingyuChai:
should I update `InfiniteScroll.js.flow` as well?

tjuanitas (approved):

tjuanitas:
> should I update `InfiniteScroll.js.flow` as well?

If you want / have extra time. It's not super important though.

QingyuChai:
> > should I update `InfiniteScroll.js.flow` as well?
> 
> If you want / have extra time. It's not super important though.

It's probably not worth it, honestly. Thanks!

PR #3049, opened 2022-09-24
fix(notification): fix tooltip overlapping notification banner

karelee7:
Fix tooltip overlapping notification banner

Screen Shot 2022-09-23 at 4 48 09 PM


src/styles/constants/_layout.scss
```diff
@@ -2,7 +2,7 @@
 $modal-dialog-z-index: 160 !default;
 $menu-z-index: 170 !default;
 $submenu-z-index: 171 !default;
-$notifications-wrapper-z-index: 180 !default;
+$notifications-wrapper-z-index: 191 !default;
```
ConradJChan, line 5: minor: Seems like these variables are sorted by z-index value, when you get a chance, could we fix the ordering? Similar comment on the JS equivalent below.

Also having it be 1 more than the overlay/tooltip doesn't leave much room for anything else in between. It seems like we've used increments of 10 for the other step ups

karelee7: Sure! I'll update this

PR #3050, opened 2022-09-26
fix(content-explorer): Added overflow style to DeleteConfirmationDialog

greg-in-a-box:
Screen Shot 2022-09-26 at 4 31 03 PM


src/elements/content-explorer/DeleteConfirmationDialog.scss
```diff
@@ -0,0 +1,5 @@
+.be-modal.be-modal-delete {
+    .be-modal-dialog-content {
```
tjuanitas, line 2: I think we'll eventually want to upgrade the elements to use our internal modal. But this is consistent to how we're doing the other modals in the elements so I think we're good to keep.

tjuanitas (approved):

PR #3052, opened 2022-09-27
feat(content-sidebar): Add error support for SidebarContentInsights

ConradJChan:
# Screenshot

Screen Shot 2022-09-26 at 5 17 12 PM

### DIFF
--- a/src/elements/content-sidebar/SidebarContentInsights.scss
+++ b/src/elements/content-sidebar/SidebarContentInsights.scss
@@ -1,0 +1,12 @@
+.bcs-SidebarContentInsights {
+    .ContentAnalyticsErrorState {
+        svg {
+            margin-top: 25px;
+        }
+    }
+
+    .ContentAnalyticsErrorState-text,
+    .ContentAnalyticsErrorState-text--permission {
+        margin-bottom: 30px;
+    }
+}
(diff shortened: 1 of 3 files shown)


src/elements/content-sidebar/SidebarContentInsights.scss
```diff
@@ -0,0 +1,12 @@
+.bcs-SidebarContentInsights {
+    .ContentAnalyticsErrorState {
+        svg {
+            margin-top: 25px;
+        }
+    }
+
+    .ContentAnalyticsErrorState-text,
+    .ContentAnalyticsErrorState-text--permission {
```
jstoffan, line 9: Modifier classes should be added to elements alongside the base class. Does the element affected in this case not have `ContentAnalyticsErrorState-text` applied?

src/elements/content-sidebar/SidebarContentInsights.scss
```diff
@@ -0,0 +1,12 @@
...
+        margin-bottom: 30px;
```
jstoffan, line 10: Should we add padding to the container, instead?

ConradJChan: Good idea, fixing it

jstoffan (approved):

PR #3057, opened 2022-09-27
fix(infinite-scroll): fix unscrollable page can't trigger loadMore

QingyuChai:
Our `infinite-scroll` component was adapted from `jaredpalmer/react-simple-infinite-scroll`. There was a bug at the time where if the initial content fits into screen, we would never be able to trigger the `loadMore` action, as it is impossible to scroll. Here is the fix they applied: 

https://github.com/jaredpalmer/react-simple-infinite-scroll/pull/10/files

This PR applies their fix, updates test cases accordingly, and switches them from sinon to jest. Furthermore, `enzyme-adapter-react-17` was bumped up, to ensure more consistent timing for lifecycle function calls, as well as ensure unmounting happens before the next test starts.

Note: bumping the enzyme adapter version does not break anything


src/components/infinite-scroll/InfiniteScroll.tsx
```diff
@@ -93,6 +93,9 @@ function InfiniteScroll({
             container.addEventListener('resize', resizeHandler);
         }
 
+        // calls onContainerScroll until page becomes scrollable
+        onContainerScroll(params);
```
QingyuChai, line 97: we don't want to throttle this, because if useEffect gets triggered again, and this doesn't get called due to a throttle, it will stop loading

tjuanitas: what about a debounce function? would that be appropriate instead of a throttle?
https://lodash.com/docs/4.17.15#debounce

src/components/infinite-scroll/InfiniteScroll.tsx
```diff
@@ -93,6 +93,9 @@ function InfiniteScroll({
             container.addEventListener('resize', resizeHandler);
         }
 
+        // calls onContainerScroll until page becomes scrollable
```
tjuanitas, line 96: I think this comment could be a little more explicit about why we need to call onContainerScroll.

maybe `// loads more content until page becomes scrollable` ?

src/components/infinite-scroll/__tests__/InfiniteScroll.test.tsx
```diff
@@ -16,154 +13,108 @@ const propsList: InfiniteScrollProps = {
     useWindow: true,
     onLoadMore: mockOnLoadMore,
     threshold,
-};
-
-const getSentinel = () => {
-    const nodes = document.querySelectorAll('[data-testid="sentinel"]');
-    return nodes[nodes.length - 1];
+    throttle: 0, // tests run too fast - keep it 0 for sanity
```
tjuanitas, line 16: what does it mean the tests are running too fast?

src/components/infinite-scroll/__tests__/InfiniteScroll.test.tsx
```diff
@@ -16,154 +13,108 @@ const propsList: InfiniteScrollProps = {
...
-
-        const sentinel = getSentinel();
-        sandbox.stub(sentinel, 'getBoundingClientRect').returns({ top: 500 + (threshold - 1) } as DOMRect);
-
-        scrollContainer.dispatchEvent(new Event('scroll'));
-        expect(mockOnLoadMore).toHaveBeenCalledTimes(1);
-    });
+            // update prop to trigger useEffect
+            component.setProps({ throttle: 1 });
+            expect(mockOnLoadMore).toHaveBeenCalledTimes(1);
```
tjuanitas, line 90: could we change this to `toBeCalledTimes(1)`?

tjuanitas: as well as the others from `toHaveBeen[...]` to `toBe[...]`

QingyuChai: yup! didn't notice as I was modifying the original tests

src/components/infinite-scroll/__tests__/InfiniteScroll.test.tsx
QingyuChai: So what happens is that the function doesn't end up getting called because the `scroll` event is instantaneous. Some of the tests take like 3-4 ms to execute, so even having the throttle be 1ms causes inconsistency.

src/components/infinite-scroll/InfiniteScroll.tsx
QingyuChai: I think in general it doesn't make sense to throttle this one, since it only get's called once per useeffect, which only happens upon prop changes. In case data fetches very fast, this function being throttled means we stop fetching. It's a bit hard to explain, I'm happy to hop on a call regarding this tomorrow

src/components/infinite-scroll/__tests__/InfiniteScroll.test.tsx
tjuanitas: ahhh gotcha. this is so the function isn't throttled between separate tests, right? i.e. the first test will complete very quickly but when the second test runs, the `onContainerScroll` is still on "cool down" (due to the throttle and the first test completing fast)

tjuanitas: I think we can improve this comment to provide more clarity. maybe we could include how the 0 is to prevent throttling between tests?

QingyuChai: Actually, this was causing throttling within the test. I realized that we don't need this anymore tho - this was needed previously because the `onContainerScroll` function that loads content until the page becomes scrollable was sharing the same throttle. That's also one of the reasons I removed the throttle here:

https://github.com/box/box-ui-elements/pull/3057#discussion_r981809121

tjuanitas (approved):

PR #3067, opened 2022-09-30
fix(dropdown): add scroll behavior when fullscreen

greathmaster:
Add scrolling behavior to Dropdown in fullscreen:

Dropdown-scroll


src/components/dropdown-menu/DropdownMenu.scss
```diff
@@ -20,6 +20,8 @@
 }
 
 @include breakpoint($medium-screen) {
+    $responsive-dropdown-header-height: 64px;
```
greg-in-a-box, line 23: i see `bdl-header-height` as a variable in the variables, do you think we can use this instead of hardcoding 64?

src/components/dropdown-menu/DropdownMenu.scss
```diff
@@ -28,15 +30,15 @@
...
             // cancels out the tether inline styling without having to set enabled=false
             transform: none !important;
         }
 
         .aria-menu:not(.submenu) {
-            position: relative;
             height: 100%;
-            margin: 0;
-            padding-top: $bdl-header-height + $bdl-grid-unit * 3;
+            margin: $responsive-dropdown-header-height 0 0 0;
```
greg-in-a-box, line 40: Are we able to do something this instead
```suggestion
            margin-top: $bdl-header-height ;
```

greg-in-a-box (changes requested):

greg-in-a-box (approved):

PR #3068, opened 2022-10-03
fix(notification): fix variable ordering and value

karelee7:
Fix z index variable ordering and adjust value


karelee7:
Per comment on https://github.com/box/box-ui-elements/pull/3049

JChan106 (approved):

PR #3069, opened 2022-10-03
feat(selectable): change single-click from select to navigate behind FF

jcamden:
LVE FF enabled:
navigate-on-click

LVE FF disabled:
select-on-click

I tried two other approaches in EUA:
1. creating an alternate `withItemSelection` and changing `onDoubleClick` to `onClick`
    - This was not enough, as the crucial change happens here where something is assigned to `onRowClick`
    - We can still do the above in conjunction with these changes, but probably the best would be to wait until we redact the FF and then simply change `onDoubleClick` to `onClick` both here (`makeSelectable`) and there (`withItemSelection`).
2. wrapping `TableRow` in `ItemLink`, but that road was hard to hoe, and maybe not the best idea (not sure how that could affect the interactivity of `TableRow` descendants.

Please note that this change does not accomplish what is required for Grid View to navigate on click.

Also, I'm not sure how to test this, off the top of my head.

### DIFF
--- a/src/components/table/makeSelectable.js
+++ b/src/components/table/makeSelectable.js
@@ -22,8 +22,10 @@ function makeSelectable(BaseTable) {
             className: PropTypes.string,
             /** Array of unique IDs of the items in the table. Each item should be a string or number, in the order they appear in the table. */
             data: PropTypes.array.isRequired,
+            isListViewEnhancementsEnabled: PropTypes.bool,
             /** Called when focus changes. `(focusedIndex: number) => void` */
             onFocus: PropTypes.func,
+            onItemDoubleClick: PropTypes.func,
             /** Called when selection changes. `(selectedItems: Array<string> | Array<number> | Set<string> | Set<number>) => void` */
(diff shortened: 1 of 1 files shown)


ahorowitz123 (commented):
Hmm I thought we had decided to go the route of making a new HOC. I'm not entirely following the reasoning behind using this approach instead.

Also, GridView does use the `handleRowClick` function so I'm confused why this also doesn't work there

ahorowitz123 (approved):

ahorowitz123 (changes requested):
On second thought, lets move this logic into EUA by swapping functionality there

greg-in-a-box:
@jcamden @ahorowitz123 can this PR be closed if the logic will be provided in EUA ?

jcamden:
> @jcamden @ahorowitz123 can this PR be closed if the logic will be provided in EUA ?

Hi @greg-in-a-box , sorry that I missed this! Yes, this branch can be closed.

PR #3070, opened 2022-10-04
chore(annotation-thread): Remove unneeded hasMention

jankowiakdawid:
We're not using this param for new Threaded Replies comments


src/api/Feed.js
```diff
@@ -1768,7 +1764,6 @@ class Feed extends Base {
      * @param {BoxItem} file - The file to which the task is assigned
      * @param {string} commentId - Comment ID
      * @param {string} text - the comment text
-     * @param {boolean} hasMention - true if there is an @mention in the text
```
jstoffan, line 1771: Was this accidentally removed? It looks like we're still using this parameter in `updateComment` and several other methods.

jankowiakdawid: Yes, definitely. I need to bring it back.

jstoffan (approved):

PR #3076, opened 2022-10-05
feat(content-sidebar): added communication with use of AnnotatorContext

kkuliczkowski-box:
- Updated `AnnotatorContext` with publishing methods
- Updated logic to act on the `AnnotatorState` changes in `withSidebarAnnotations`
- Created `useAnnotatorEvents` hook for acting on actions


src/elements/common/annotator-context/useAnnotatorEvents.ts
```diff
@@ -0,0 +1,68 @@
...
+            onAnnotationDeleteEnd(annotation.id);
+        } else if (action === Action.UPDATE_START && annotation && onAnnotationUpdateStart) {
+            onAnnotationUpdateStart(annotation);
+        } else if (action === Action.UPDATE_END && annotation && onAnnotationUpdateEnd) {
+            onAnnotationUpdateEnd(annotation);
+        }
+    }
+
+    // Take action only if annotation became active in Sidebar
+    if (origin === 'sidebar' && action === Action.SET_ACTIVE && onSidebarAnnotationSelected) {
```
wpiesiak, line 54: why is it important to narrow it down only for the action taken from Sidebar?

kkuliczkowski-box: This will be used by Canvas to call their function that centers screen on the annotation and it should be done when Annotation is selected in Sidebar. The same action will be published when selecting from `AnnotationThread` (in order to also select that annotation in Sidebar) and I don't think Canvas should act at that time.

PR #3079, opened 2022-10-06
feat(annotation-thread): Truncate the message in Comment / AnnotationActivity

jankowiakdawid:
We want to truncate long messages to only first 7 lines. When any massage is longer we'll truncate it and show button to show more.

#### Message truncated as it would look like in IE
Screenshot 2022-10-06 at 22 59 14

#### Message truncated as it would look like in modern browsers
Screenshot 2022-10-06 at 23 01 40

#### Full message
Screenshot 2022-10-06 at 22 59 30

### DIFF
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
@@ -187,6 +187,7 @@ const ActiveState = ({
                                             can_reply: getProp(item.permissions, 'can_reply', false),
                                             can_resolve: getProp(item.permissions, 'can_resolve', false),
                                         }}
+                                        shouldTruncate
                                         translations={translations}
                                     />
                                 </ActivityThread>
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
@@ -187,7 +187,7 @@ const ActiveState = ({
                                             can_reply: getProp(item.permissions, 'can_reply', false),
                                             can_resolve: getProp(item.permissions, 'can_resolve', false),
                                         }}
-                                        shouldTruncate
+                                        allowCollapse
                                         translations={translations}
                                     />
                                 </ActivityThread>
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
@@ -187,7 +187,6 @@ const ActiveState = ({
                                             can_reply: getProp(item.permissions, 'can_reply', false),
                                             can_resolve: getProp(item.permissions, 'can_resolve', false),
                                         }}
-                                        allowCollapse
                                         translations={translations}
                                     />
                                 </ActivityThread>
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActivityThreadReplies.js
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActivityThreadReplies.js
@@ -51,7 +51,6 @@ const ActivityThreadReplies = ({
                 <Comment
                     key={lastReply.type + lastReply.id}
                     {...lastReply}
-                    allowCollapse
                     currentUser={currentUser}
                     getAvatarUrl={getAvatarUrl}
                     getMentionWithQuery={getMentionWithQuery}
@@ -67,7 +66,6 @@ const ActivityThreadReplies = ({
                     <Comment
                         key={reply.type + reply.id}
                         {...reply}
-                        allowCollapse
                         currentUser={currentUser}
                         getAvatarUrl={getAvatarUrl}
                         getMentionWithQuery={getMentionWithQuery}
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActivityThreadReplies.js
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActivityThreadReplies.js
@@ -59,7 +59,7 @@ const ActivityThreadReplies = ({
                     onDelete={onDelete}
                     onEdit={onEdit}
                     permissions={getReplyPermissions(lastReply)}
-                    shouldTruncate
+                    allowCollapse
                     translations={translations}
                 />
             ) : (
@@ -75,7 +75,7 @@ const ActivityThreadReplies = ({
                         onDelete={onDelete}
                         onEdit={onEdit}
                         permissions={getReplyPermissions(reply)}
-                        shouldTruncate
+                        allowCollapse
                         translations={translations}
                     />
                 ))
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActivityThreadReplies.js
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActivityThreadReplies.js
@@ -51,6 +51,7 @@ const ActivityThreadReplies = ({
                 <Comment
                     key={lastReply.type + lastReply.id}
                     {...lastReply}
+                    allowCollapse
                     currentUser={currentUser}
                     getAvatarUrl={getAvatarUrl}
                     getMentionWithQuery={getMentionWithQuery}
@@ -59,7 +60,6 @@ const ActivityThreadReplies = ({
                     onDelete={onDelete}
                     onEdit={onEdit}
                     permissions={getReplyPermissions(lastReply)}
-                    allowCollapse
                     translations={translations}
                 />
             ) : (
@@ -67,6 +67,7 @@ const ActivityThreadReplies = ({
                     <Comment
                         key={reply.type + reply.id}
                         {...reply}
+                        allowCollapse
                         currentUser={currentUser}
                         getAvatarUrl={getAvatarUrl}
                         getMentionWithQuery={getMentionWithQuery}
@@ -75,7 +76,6 @@ const ActivityThreadReplies = ({
                         onDelete={onDelete}
                         onEdit={onEdit}
                         permissions={getReplyPermissions(reply)}
-                        allowCollapse
                         translations={translations}
                     />
                 ))
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActivityThreadReplies.js
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActivityThreadReplies.js
@@ -59,6 +59,7 @@ const ActivityThreadReplies = ({
                     onDelete={onDelete}
                     onEdit={onEdit}
                     permissions={getReplyPermissions(lastReply)}
+                    shouldTruncate
                     translations={translations}
                 />
             ) : (
@@ -74,6 +75,7 @@ const ActivityThreadReplies = ({
                         onDelete={onDelete}
                         onEdit={onEdit}
                         permissions={getReplyPermissions(reply)}
+                        shouldTruncate
                         translations={translations}
                     />
                 ))
--- a/src/elements/content-sidebar/activity-feed/comment/Comment.js
+++ b/src/elements/content-sidebar/activity-feed/comment/Comment.js
@@ -22,18 +22,20 @@ import ActivityStatus from '../common/activity-status';
 import CommentForm from '../comment-form';
 import { COMMENT_STATUS_OPEN, COMMENT_STATUS_RESOLVED, PLACEHOLDER_USER } from '../../../../constants';
 import messages from './messages';
+import { withFeatureConsumer, isFeatureEnabled } from '../../../common/feature-checking';
 import type { GetAvatarUrlCallback, GetProfileUrlCallback } from '../../../common/flowTypes';
 import type { Translations } from '../../flowTypes';
 import type { SelectorItems, User } from '../../../../common/types/core';
 import type { ActionItemError, BoxCommentPermission, FeedItemStatus } from '../../../../common/types/feed';
+import type { FeatureConfig } from '../../../common/feature-checking';
 import './Comment.scss';
 
 type Props = {
-    allowCollapse?: boolean,
     created_at: string | number,
     created_by: User,
     currentUser?: User,
     error?: ActionItemError,
+    features: FeatureConfig,
     getAvatarUrl: GetAvatarUrlCallback,
     getMentionWithQuery?: Function,
     getUserProfileUrl?: GetProfileUrlCallback,
@@ -114,13 +116,13 @@ class Comment extends React.Component<Props, State> {
 
     render(): React.Node {
         const {
-            allowCollapse,
             created_by,
             created_at,
             permissions = {},
             id,
             isPending,
             error,
+            features,
             tagged_message = '',
             translatedTaggedMessage,
             translations,
@@ -134,6 +136,7 @@ class Comment extends React.Component<Props, State> {
             onEdit,
             status,
         } = this.props;
+        const allowCollapse = isFeatureEnabled(features, 'activityFeed.collapsableMessages.enabled');
         const { isConfirmingDelete, isEditing, isInputOpen } = this.state;
         const canDelete = permissions.can_delete;
         const canEdit = onEdit !== noop && permissions.can_edit;
@@ -273,4 +276,5 @@ class Comment extends React.Component<Props, State> {
     }
 }
 
-export default Comment;
+export { Comment as CommentComponent };
+export default withFeatureConsumer(Comment);
--- a/src/elements/content-sidebar/activity-feed/comment/Comment.js
+++ b/src/elements/content-sidebar/activity-feed/comment/Comment.js
@@ -254,7 +254,7 @@ class Comment extends React.Component<Props, State> {
                             />
                         ) : (
                             <ActivityMessage
-                                allowCollapse={allowCollapse}
+                                canCollapse={allowCollapse}
                                 id={id}
                                 isEdited={(isEdited: boolean)}
                                 tagged_message={tagged_message}
--- a/src/elements/content-sidebar/activity-feed/comment/Comment.js
+++ b/src/elements/content-sidebar/activity-feed/comment/Comment.js
@@ -69,7 +69,6 @@ type State = {
 
 class Comment extends React.Component<Props, State> {
     static defaultProps = {
-        allowCollapse: false,
         onDelete: noop,
         onEdit: noop,
     };
--- a/src/elements/content-sidebar/activity-feed/comment/Comment.js
+++ b/src/elements/content-sidebar/activity-feed/comment/Comment.js
@@ -22,12 +22,10 @@ import ActivityStatus from '../common/activity-status';
 import CommentForm from '../comment-form';
 import { COMMENT_STATUS_OPEN, COMMENT_STATUS_RESOLVED, PLACEHOLDER_USER } from '../../../../constants';
 import messages from './messages';
-import { withFeatureConsumer, isFeatureEnabled } from '../../../common/feature-checking';
 import type { GetAvatarUrlCallback, GetProfileUrlCallback } from '../../../common/flowTypes';
 import type { Translations } from '../../flowTypes';
 import type { SelectorItems, User } from '../../../../common/types/core';
 import type { ActionItemError, BoxCommentPermission, FeedItemStatus } from '../../../../common/types/feed';
-import type { FeatureConfig } from '../../../common/feature-checking';
 import './Comment.scss';
 
 type Props = {
@@ -35,7 +33,6 @@ type Props = {
     created_by: User,
     currentUser?: User,
     error?: ActionItemError,
-    features: FeatureConfig,
     getAvatarUrl: GetAvatarUrlCallback,
     getMentionWithQuery?: Function,
     getUserProfileUrl?: GetProfileUrlCallback,
@@ -121,7 +118,6 @@ class Comment extends React.Component<Props, State> {
             id,
             isPending,
             error,
-            features,
             tagged_message = '',
             translatedTaggedMessage,
             translations,
@@ -135,7 +131,6 @@ class Comment extends React.Component<Props, State> {
             onEdit,
             status,
         } = this.props;
-        const allowCollapse = isFeatureEnabled(features, 'activityFeed.collapsableMessages.enabled');
         const { isConfirmingDelete, isEditing, isInputOpen } = this.state;
         const canDelete = permissions.can_delete;
         const canEdit = onEdit !== noop && permissions.can_edit;
@@ -256,7 +251,6 @@ class Comment extends React.Component<Props, State> {
                             />
                         ) : (
                             <ActivityMessage
-                                canCollapse={allowCollapse}
                                 id={id}
                                 isEdited={(isEdited: boolean)}
                                 tagged_message={tagged_message}
@@ -275,5 +269,4 @@ class Comment extends React.Component<Props, State> {
     }
 }
 
-export { Comment as CommentComponent };
-export default withFeatureConsumer(Comment);
+export default Comment;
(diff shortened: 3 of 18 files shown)


src/elements/content-sidebar/activity-feed/common/activity-message/ActivityMessage.scss
```diff
@@ -23,6 +23,27 @@
     color: $bdl-box-blue;
 }
 
+.bcs-ActivityMessage-truncated {
+    /* webkit-line-clamp only works in combinatio with -webkit-box */
+    /* stylelint-disable-next-line value-no-vendor-prefix */
+    display: -webkit-box;
+    max-height: 140px;
```
karelee7, line 30: Should we store this px amount in a variable? Or make a note about how we got 140px exactly?

src/elements/content-sidebar/activity-feed/common/activity-message/__tests__/ActivityMessage.test.js
```diff
@@ -136,4 +136,14 @@ describe('elements/content-sidebar/ActivityFeed/common/activity-message', () =>
         expect(wrapper.state('isTranslation')).toBe(false);
         expect(wrapper.state('isLoading')).toBe(false);
     });
+
+    test('should have TruncatableMessage when `shouldTruncate` is passed', () => {
```
karelee7, line 140: Should we also add a test for when it shouldn't be truncated?

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -179,6 +179,7 @@ const ActiveState = ({
                                             can_reply: getProp(item.permissions, 'can_reply', false),
                                             can_resolve: getProp(item.permissions, 'can_resolve', false),
                                         }}
+                                        shouldTruncate
```
bfoxx1906, line 182: So it looks like shouldTruncate doesn't determine whether or not a comment is truncated it determines if truncation should be possible. shouldTruncate could be true but truncation may never happen because the messages aren't long enough right? If this is the case then It might be more intuitive to name this something along the lines of **allowTruncation**.

jankowiakdawid: Exactly! 💯 Great suggestion! I double checked in designs, and I've noticed they use different name for this `Collapse` so I think I'll go with full rename for the component.
But will definitely change this prop to `allowCollapsing`

src/elements/content-sidebar/activity-feed/common/activity-message/TruncatableMessage.js
```diff
@@ -0,0 +1,39 @@
...
+
+export default function TruncatableMessage({ children }: Props) {
+    const [isTruncated, setTruncated] = React.useState(true);
+    const [shouldTruncate, setShouldTrucate] = React.useState(false);
+    const messageContainer = React.useRef<HTMLDivElement | null>(null);
+
+    React.useLayoutEffect(() => {
+        if (messageContainer.current) {
+            const { clientHeight, scrollHeight } = messageContainer.current;
+            setShouldTrucate(clientHeight !== scrollHeight || clientHeight > 140);
```
jstoffan, line 19: We're currently duplicating this value in the stylesheet. Can we share the value between the component and stylesheet? If not, should we apply this one style rule via inline styles, instead?

src/elements/content-sidebar/activity-feed/common/activity-message/TruncatableMessage.js
```diff
@@ -0,0 +1,39 @@
...
+            <div
+                className={classNames({
+                    'bcs-ActivityMessage-truncated': isTruncated,
+                })}
+                ref={messageContainer}
+            >
+                {children}
+            </div>
+
+            {shouldTruncate ? (
```
jstoffan, line 34: Minor, but we should be able to use `&&` here, instead.

src/elements/content-sidebar/activity-feed/common/activity-message/__tests__/ActivityMessage.test.js
```diff
@@ -136,4 +136,14 @@ describe('elements/content-sidebar/ActivityFeed/common/activity-message', () =>
...
+        const commentText = {
+            tagged_message: 'How u doing @[2030326577:Young Jeezy]?',
+        };
+
+        const wrapper = shallow(<ActivityMessage id="123" {...commentText} shouldTruncate />);
+
+        expect(wrapper.find('TruncatableMessage').length).toEqual(1);
```
jstoffan, line 147: Minor, but we typically prefer `expect(wrapper.exists('TruncatableMessage')).toBe(true)`

src/elements/content-sidebar/activity-feed/common/activity-message/ToggleMoreLessButton.js
```diff
@@ -0,0 +1,24 @@
...
+
+import messages from './messages';
+
+type Props = {
+    isMore: boolean,
+    onClick: Function,
+};
+
+const ToggleMoreLessButton = ({ onClick, isMore }: Props): React.Node => (
+    <PlainButton className="bcs-ActivityMessage-toggleMoreLess" onClick={onClick}>
```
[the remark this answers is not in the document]
jankowiakdawid: Definitely!

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
tjuanitas: this is minor but `allow` isn't a common prefix we use for booleans. Would `canCollapse` suffice?

src/elements/content-sidebar/activity-feed/common/activity-message/ActivityMessage.scss
```diff
@@ -23,6 +23,27 @@
     color: $bdl-box-blue;
 }
 
+.bcs-ActivityMessage-truncated {
+    /* webkit-line-clamp only works in combinatio with -webkit-box */
```
tjuanitas, line 27: typo
```suggestion
    /* webkit-line-clamp only works in combination with -webkit-box */
```

src/elements/content-sidebar/activity-feed/common/activity-message/ActivityMessage.scss
```diff
@@ -23,6 +23,27 @@
...
+    /* stylelint-disable-next-line value-no-vendor-prefix */
+    display: -webkit-box;
+    max-height: 140px;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    overflow-wrap: break-word;
+    -webkit-line-clamp: 7;
+    -webkit-box-orient: vertical;
```
tjuanitas, line 35: I'm not familiar with these or `display: -webkit-box`. what does this do? does this work in other browsers (Firefox, Safari)?

src/elements/content-sidebar/activity-feed/common/activity-message/TruncatableMessage.js
```diff
@@ -0,0 +1,39 @@
...
+import ToggleMoreLessButton from './ToggleMoreLessButton';
+import './ActivityMessage.scss';
+
+type Props = {
+    children: React.Node,
+};
+
+export default function TruncatableMessage({ children }: Props) {
+    const [isTruncated, setTruncated] = React.useState(true);
+    const [shouldTruncate, setShouldTrucate] = React.useState(false);
```
tjuanitas, line 13: I think this is being renamed but in case not:
```suggestion
    const [shouldTruncate, setShouldTruncate] = React.useState(false);
```

src/elements/content-sidebar/activity-feed/common/activity-message/messages.js
```diff
@@ -12,6 +12,16 @@ const messages = defineMessages({
         defaultMessage: 'Translate',
         description: 'Translate button for translating comment',
     },
+    activityMessageSeeMore: {
+        id: 'be.contentSidebar.activityFeed.common.seeMore',
+        defaultMessage: 'See more',
+        description: 'See more button for showing whole long message',
+    },
+    activityMessageSeeLess: {
+        id: 'be.contentSidebar.activityFeed.common.seeLess',
```
tjuanitas, line 21: minor: the `id`s for the messages above have a `Message` suffix

`id: 'be.contentSidebar.activityFeed.common.seeMoreMessage',`
`id: 'be.contentSidebar.activityFeed.common.seeLessMessage',`

src/elements/content-sidebar/activity-feed/common/activity-message/ActivityMessage.scss
jankowiakdawid: It allows to define after how many lines text should be cut, and adds `...` at the end
It is standardized for legacy support

https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp
https://caniuse.com/css-line-clamp

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -179,6 +179,7 @@ const ActiveState = ({
                                             can_reply: getProp(item.permissions, 'can_reply', false),
                                             can_resolve: getProp(item.permissions, 'can_resolve', false),
                                         }}
+                                        allowCollapse
```
jstoffan, line 182: Is this desired? It will immediately reflect in the live Preview experience for long comments.

src/elements/content-sidebar/activity-feed/activity-feed/ActivityThreadReplies.js
```diff
@@ -74,6 +75,7 @@ const ActivityThreadReplies = ({
                         onDelete={onDelete}
                         onEdit={onEdit}
                         permissions={getReplyPermissions(reply)}
+                        allowCollapse
```
jstoffan, line 78: Minor, but we typically try to keep our props, destructuring, etc. alphabetized when possible.

src/elements/content-sidebar/activity-feed/common/activity-message/ActivityMessage.js
```diff
@@ -5,10 +5,12 @@ import LoadingIndicator from '../../../../../components/loading-indicator';
 import formatTaggedMessage from '../../utils/formatTaggedMessage';
 import ShowOriginalButton from './ShowOriginalButton';
 import TranslateButton from './TranslateButton';
+import CollapsableMessage from './CollapsableMessage';
```
jstoffan, line 8: Alphabetize?

src/styles/constants/_layout.scss
```diff
@@ -21,3 +21,7 @@ $bdl-grid-unit-px: 4 !default; // unitless grid unit (defined in pixels for manu
 
 // Headers
 $bdl-header-height: 64px;
+
+// Message
+$message-line-clamp: 7;
+$message-max-height: $message-line-clamp * 20px; // line-height for message is 20px with font-size of 13px, mostly for IE11
```
jstoffan, line 27: I understand we created these variables to share them between the component/stylesheet. However, these are global values. Do we intend to use them elsewhere? What happens if other experiences have different font sizes and/or line heights? It may be preferable to move the values into the JS component and use them via inline styles. What do you think?

src/elements/content-sidebar/activity-feed/common/activity-message/ToggleMoreLessButton.js
```diff
@@ -0,0 +1,24 @@
...
+
+import messages from './messages';
+
+type Props = {
+    isMore: boolean,
+    onClick: Function,
+};
+
+const ToggleMoreLessButton = ({ isMore, onClick }: Props): React.Node => (
+    <PlainButton type="button" className="bcs-ActivityMessage-toggleMoreLess" onClick={onClick}>
```
jstoffan, line 19: Minor, but can we move `type` to the end?

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
jankowiakdawid: Without it the top-level comments won't be collapsible. Is there a check that I can put inside so that it will be true only for Threaded Comments? like:
```suggestion
                                        allowCollapse={hasReplies}
```
?

I'll bring it to the team tomorrow.

src/styles/constants/_layout.scss
jankowiakdawid: I need some time to think this through. I don't like to use inline styles.

tjuanitas: I would also hesitate to put them in this file/directory. at minimum they would need a `bdl-` prefix.

since they only have one usage at the moment, is it okay to have them as local variables in `ActivityMessage.scss`?

src/elements/content-sidebar/activity-feed/common/activity-message/ToggleMoreLessButton.js
jankowiakdawid: Can we have lint rule for this?

src/elements/content-sidebar/activity-feed/common/activity-message/ActivityMessage.js
jankowiakdawid: Can we have lint rule for this?

src/styles/constants/_layout.scss
jankowiakdawid: I wanted them in `ActivityMessage.scss` but the `:export {}` syntax didn't worked. Then I found that we have build script for generating js files with sass variables.
I think I'll go with `styles`.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
jankowiakdawid: @kkuliczkowski-box suggested using features for this, that way it would be enabled for Canvas only for now.
Similarly to this
https://github.com/box/box-ui-elements/blob/master/src/elements/content-sidebar/ActivitySidebar.js#L892

I found that I can use https://github.com/box/box-ui-elements/blob/master/src/elements/common/feature-checking/FeatureContext.js

I decided to put that in single place in `Comment.js`. What do you think?
I have implementation ready, but still some tests are failing before I commit.

jankowiakdawid: here's the commit https://github.com/box/box-ui-elements/pull/3079/commits/f807afc79df5c9a09992de997043b61474462183

src/elements/content-sidebar/activity-feed/common/activity-message/ActivityMessage.js
tjuanitas: Yeah we have an internal ticket to investigate updating our prettier config. It's sort of waiting on a third party plugin and would need a review from security as well. But it's something that definitely should be added to the repo.

src/elements/content-sidebar/activity-feed/common/activity-message/ToggleMoreLessButton.js
jstoffan: Sounds fantastic. The issue we've had in the past with new rules like that is that there's lots of non-conforming code that needs to be updated.

src/elements/content-sidebar/activity-feed/common/activity-message/CollapsableMessage.js
```diff
@@ -0,0 +1,43 @@
...
+import * as React from 'react';
+import classNames from 'classnames';
+import ToggleMoreLessButton from './ToggleMoreLessButton';
+import './ActivityMessage.scss';
+
+type Props = {
+    children: React.Node,
+};
+
+// this is for backward support for IE11
```
jstoffan, line 11: We can remove this logic, if this is the case, since new features do not have to support IE11.

src/elements/content-sidebar/activity-feed/comment/Comment.js
```diff
@@ -249,6 +255,7 @@ class Comment extends React.Component<Props, State> {
                             />
                         ) : (
                             <ActivityMessage
+                                canCollapse={allowCollapse}
```
jstoffan, line 258: Should this be handled within `ActivityMessage` so annotation messages also get the same experience?

src/elements/content-sidebar/activity-feed/comment/Comment.js
```diff
@@ -66,6 +69,7 @@ type State = {
 
 class Comment extends React.Component<Props, State> {
     static defaultProps = {
+        allowCollapse: false,
```
jstoffan, line 72: Is this needed?

jankowiakdawid: No

src/elements/content-sidebar/activity-feed/common/activity-message/CollapsableMessage.js
```diff
@@ -0,0 +1,43 @@
+// @flow
+import * as React from 'react';
+import classNames from 'classnames';
+import ToggleMoreLessButton from './ToggleMoreLessButton';
+import './ActivityMessage.scss';
+
+type Props = {
+    children: React.Node,
+};
```
JChan106, line 9: Do we need to define children in the props? I think React components come with children built into the props

src/elements/content-sidebar/activity-feed/common/activity-message/CollapsableMessage.js
```diff
@@ -0,0 +1,43 @@
...
+        }
+    }, []);
+
+    return (
+        <>
+            <div
+                className={classNames({
+                    'bcs-ActivityMessage-collapsed': isCollapsed,
+                })}
+                style={{ maxHeight: isCollapsed ? MAX_MESSAGE_HEIGHT : 'none' }}
```
JChan106, line 32: Do we need to use inline styling if we use `isCollapsed` in the classname?

src/elements/content-sidebar/activity-feed/common/activity-message/ToggleMoreLessButton.js
```diff
@@ -0,0 +1,24 @@
...
+import * as React from 'react';
+import { FormattedMessage } from 'react-intl';
+
+import PlainButton from '../../../../../components/plain-button';
+
+import messages from './messages';
+
+type Props = {
+    isMore: boolean,
+    onClick: Function,
```
JChan106, line 15: Can we type this with the arrow syntax `() => void` to be more explicit to what the props and return value are

src/elements/content-sidebar/activity-feed/common/activity-message/CollapsableMessage.js
```diff
@@ -0,0 +1,43 @@
...
+import './ActivityMessage.scss';
+
+type Props = {
+    children: React.Node,
+};
+
+// this is for backward support for IE11
+const MAX_MESSAGE_HEIGHT = 140;
+
+export default function CollapsableMessage({ children }: Props) {
```
JChan106, line 14: Should we export default inline or at the end of files?

src/elements/content-sidebar/activity-feed/comment/Comment.js
jankowiakdawid: I mean, I don't know. It can. Probably would be better for the users.

src/elements/content-sidebar/activity-feed/common/activity-message/CollapsableMessage.js
jankowiakdawid: Actually, we don't need it anymore. It was for backward compatibility with IE11. But as @jstoffan pointed out it's not required.

src/elements/content-sidebar/activity-feed/common/activity-message/CollapsableMessage.js
jankowiakdawid: We need. https://flow.org/en/docs/react/children/

src/elements/content-sidebar/activity-feed/common/activity-message/ActivityMessage.scss
```diff
@@ -13,18 +13,49 @@
...
+    /* stylelint-disable-next-line value-no-vendor-prefix */
+    display: -webkit-box;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    overflow-wrap: break-word;
+    -webkit-line-clamp: 7;
+    -webkit-box-orient: vertical;
+}
+
+.btn-plain.bcs-ActivityMessage-translate,
```
jstoffan, line 36: Minor, but this can be simplified:

```
.bcs-ActivityMessage-toggleMoreLess,
.bcs-ActivityMessage-translate {
    &.btn-plain {
        // ... styles
    }
}

File: src/elements/content-sidebar/activity-feed/common/activity-message/__tests__/ActivityMessage.test.js [hunk 25]
```diff
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { mount, shallow } from 'enzyme';
 
-import ActivityMessage from '../ActivityMessage';
+import { ActivityMessage } from '../ActivityMessage';
```
jstoffan, line 4: Should we be testing the default export by wrapping it in a FeaturesProvider, instead? It's what other components will use.

src/elements/content-sidebar/activity-feed/common/activity-message/ActivityMessage.scss
jankowiakdawid: I didn't knew exact syntax for that, but I knew you'll tell me 😉

src/elements/content-sidebar/activity-feed/common/activity-message/ActivityMessage.scss
```diff
@@ -13,18 +13,49 @@
...
+.btn-plain.bcs-ActivityMessage-toggleMoreLess {
     display: block;
+    margin-left: -2px;
     color: $bdl-box-blue;
-}
+    font-weight: bold;
+    // to align with rest of text despite invisible border
+    line-height: $bdl-line-height - 4px;
+    border: 2px solid transparent;
```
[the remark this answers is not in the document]
jankowiakdawid: No. If we won't have this transparent border then when user will interact with the button, button will move due to appearing border. My solution to that was to always have border but keep it invisible. This lead to more white space around those button which I then needed to subtract.

src/elements/content-sidebar/activity-feed/common/activity-message/CollapsableMessageToggle.js
```diff
@@ -0,0 +1,24 @@
...
+
+import messages from './messages';
+
+type Props = {
+    isMore: boolean,
+    onClick: () => void,
+};
+
+const CollapsableMessageToggle = ({ isMore, onClick }: Props): React.Node => (
+    <PlainButton className="bcs-ActivityMessage-toggleMoreLess" onClick={onClick} type="button">
```
jstoffan, line 19: Minor, but it's best to pass this in from the `ActivityMessage` component. SUIT naming asks for each component to have its own namespace. Apologies for missing this on prior reviews.

jstoffan (dismissed):

src/elements/content-sidebar/activity-feed/common/activity-message/CollapsableMessageToggle.js
jankowiakdawid: I will read more on SUIT naming. For now, I leave it as it is, because I see similar naming in other components inside `./activity-message`. Can it be that all the names are in context of `ActivityMessage`  which is a single place of styling for all those small components that were just extracted from `ActivityMessage` but aren't standalone on their own?

jstoffan: We may have made similar mistakes in other components, but that doesn't mean it's correct. Each component is supposed to have its own top-level name and hyphens are used to indicate descendent elements within a component. Since this is a separate file, it should have a separate class name. Either way, this issue alone isn't enough to block the merge.

jstoffan (approved):

PR #3081, opened 2022-10-07
feat(content-sidebar): updated annotation events communication

kkuliczkowski-box:
- Updated `AnnotatorContext` with new emit methods
- Updated logic that acts on the `AnnotatorState` changes in `withSidebarAnnotations`
- Created `useAnnotatorEvents` hook for handling events


src/elements/common/annotator-context/useAnnotatorEvents.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    const emitUpdateAnnotationStartEvent = (annotation: Object) => {
+        emitUpdateAnnotationEvent(eventEmitter, annotation, Status.PENDING);
+    };
+    const emitUpdateAnnotationEndEvent = (annotation: Object) => {
+        emitUpdateAnnotationEvent(eventEmitter, annotation, Status.SUCCESS);
+    };
+
+    const annotationSidebarSelectedListener = (annotationId: string) => {
+        if (onSidebarAnnotationSelected) {
```
jstoffan, line 42: Can we simplify this logic by setting these to `noop` functions? That way, we can pass them directly to the `addListener` calls.

src/elements/common/annotator-context/useAnnotatorEvents.ts
```diff
@@ -0,0 +1,91 @@
...
+export interface UseAnnotatorEventsProps {
+    eventEmitter: EventEmitter;
+    onAnnotationDeleteEnd?: (annotationId: string) => void;
+    onAnnotationDeleteStart?: (annotationId: string) => void;
+    onAnnotationUpdateEnd?: (annotation: Object) => void;
+    onAnnotationUpdateStart?: (annotation: Object) => void;
+    onSidebarAnnotationSelected?: (annotationId: string) => void;
+}
+
+const emitUpdateAnnotationEvent = (eventEmitter: EventEmitter, annotation: Object, status: Status) => {
```
jstoffan, line 14: Should this be declared inside the hook to stay consistent with the others? That way, it doesn't need `eventEmitter` passed to it.

src/elements/content-sidebar/withSidebarAnnotations.js
```diff
@@ -130,6 +134,31 @@ export default function withSidebarAnnotations(
...
+        updateAnnotation() {
+            const {
+                annotatorState: { action, annotation },
+                api,
+                file,
+                isOpen,
+                location,
+            } = this.props;
+
+            const feedAPI = api.getFeedAPI(false);
```
jstoffan, line 146: This method duplicates quite a bit of logic from `addAnnotation`. Can we simplify/reuse the common parts?

src/elements/content-sidebar/withSidebarAnnotations.js
```diff
@@ -69,7 +69,11 @@ export default function withSidebarAnnotations(
             const isTransitioningToAnnotationPath = activeAnnotationId && !isAnnotationsPath;
             const prevFileVersionId = getProp(prevMatch, 'params.fileVersionId');
 
-            if (annotation && prevAnnotation !== annotation) {
+            if (action === 'update_start' || action === 'update_end') {
+                this.updateAnnotation();
+            }
+
+            if ((action === 'create_start' || action === 'create_end') && annotation && prevAnnotation !== annotation) {
                 this.addAnnotation();
```
jstoffan, line 77: Is it possible to combine add/update into a single method? It could detect whether the annotation already exists in cache and, if so, skip creating it anew? That would let us completely ignore `action` in this lifecycle step.

src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThread.js
```diff
@@ -27,6 +28,7 @@ type Props = {
     className: string,
     clientName: string,
     currentUser: User,
+    eventEmitter?: EventEmitter,
```
jstoffan, line 31: It doesn't look like this prop is used in this component. Can we remove this?

src/elements/content-sidebar/__tests__/withSidebarAnnotations.test.js
```diff
@@ -249,6 +265,75 @@ describe('elements/content-sidebar/withSidebarAnnotations', () => {
...
+                const instance = wrapper.instance();
+
+                instance.updateAnnotation();
+
+                const expectedAnnotationData = { ...annotation, isPending: expectedIsPending };
+                expect(feedAPI.updateFeedItem).toBeCalledWith(expectedAnnotationData, annotation.id);
+            },
+        );
+
+        test.each`
```
jstoffan, line 294: If we can avoid duplicating the logic, we can avoid duplicating the tests.

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -75,20 +77,39 @@ export default function withAnnotations<P extends object>(
...
+        };
+
+        emitAnnotationUpdateEvent = (annotation: Object, isStartEvent = false) => {
+            const { annotator } = this;
+
+            if (!annotator) {
+                return;
+            }
+
+            const event = isStartEvent ? 'sidebar.annotations_update_start' : 'sidebar.annotations_update';
```
jstoffan, line 100: Why do we need the `sidebar` prefix here but not elsewhere?

src/elements/common/annotator-context/__tests__/useAnnotatorEvents.test.tsx
```diff
@@ -0,0 +1,158 @@
...
+import useAnnotatorEvents from '../useAnnotatorEvents';
+import { Status } from '../types';
+
+describe('src/elements/common/annotator-context/useAnnotatorEvents', () => {
+    const mockAddListener = jest.fn();
+    const mockEmit = jest.fn();
+    const mockRemoveListener = jest.fn();
+
+    afterEach(() => {
+        jest.clearAllMocks();
```
jstoffan, line 11: Minor, but we should be able to omit this line.

src/elements/content-sidebar/withSidebarAnnotations.js
kkuliczkowski-box: I've extracted the logic for refreshing Activity Sidebar into a `refreshActivitySidebar` method.

src/elements/content-sidebar/withSidebarAnnotations.js
kkuliczkowski-box: If it was the only piece of logic to be added here I think your suggestion would make a lot of sense, but we will need to add support for 2 more actions in the coming days. One for deleting annotations and another for simply adding an annotation to the cache (and state) - when the annotation is added in `AnnotationThread` it needs to be "copied" to sidebar (annotation will not need to be created in sidebar as it's done in Preview integration). Having that in mind I believe that adding a clear distinction of which logic to follow based on action will be a lot more readable here.

src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThread.js
kkuliczkowski-box: I've included it here for future reference as another PR will be created shortly. `AnnotationThread` will reuse the `useAnnotatorEvents` hook and thus it needs to pass eventEmitter object to it.

src/elements/content-sidebar/__tests__/withSidebarAnnotations.test.js
kkuliczkowski-box: Right, once I've extracted the logic for refreshing Activity Sidebar, there is just one instance of that test.

src/elements/common/annotator-context/withAnnotations.tsx
kkuliczkowski-box: This one is a bit tricky. I would love to use the sidebar namespace for all events that originate from sidebar, however it needs to be backwards compatible with Preview integration and I can't change the event name for removing the annotation (`annotations_remove`). I also can't add the common namespace for all events from outside of sidebar, because of the same reason - Preview emits events `annotations_active_change` and `annotations_create`. Current implementation also uses a different event name for changing the active annotation. When the change is done in sidebar it's `annotations_active_set` and when in Preview it's `annotations_active_change`. This is a good candidate for refactoring in future when there's more time so that the naming is unified for all parties.

jstoffan (approved):

PR #3082, opened 2022-10-07
feat(gridview): Make hot key config for grid view

ahorowitz123:
- Refactor hotkeys to split into shared, list view, and grid view
- Add functionality for shift keyboard navigation on grid view

### DIFF
--- a/src/components/table/__tests__/makeSelectable.test.js
+++ b/src/components/table/__tests__/makeSelectable.test.js
@@ -315,6 +315,89 @@ describe('components/table/makeSelectable', () => {
         });
     });
 
+    describe('handleShiftKeyDownForGridRow()', () => {
+        afterEach(() => {
+            shiftSelect.mockReset();
+        });
+
+        test('should be no-op when target is the boundary and already selected', () => {
+            const wrapper = getWrapper({
+                selectedItems: ['a'],
+                onSelect: sandbox.mock().never(),
+            });
+            wrapper.setState({ focusedIndex: 0 });
+
+            wrapper.instance().handleShiftKeyDownForGridRow(0, 0);
+            expect(shiftSelect).not.toHaveBeenCalled();
+        });
+
+        test('should select target when it is not already selected', () => {
+            const selectedItems = [];
+            const focusedIndex = 1;
+            const rowIndex = 0;
+            const anchorIndex = 1;
+
+            const wrapper = getWrapper({
+                selectedItems,
+            });
+            wrapper.setState({ focusedIndex });
+            const instance = wrapper.instance();
+            instance.previousIndex = focusedIndex;
+            shiftSelect.mockImplementation(() => new Set([0]));
+            instance.onSelect = (funcSelectedItems, funcFocusedIndex) => {
+                expect(funcSelectedItems.equals(new Set(['a']))).toBe(true);
+                expect(funcFocusedIndex).toEqual(0);
+            };
+
+            instance.handleShiftKeyDownForGridRow(0, 0);
+            expect(shiftSelect).toHaveBeenCalledWith(Set(selectedItems), focusedIndex, rowIndex, anchorIndex);
+        });
+
+        test('should deselect source when both source and target are selected', () => {
+            const selectedItems = ['a', 'b'];
+            const focusedIndex = 0;
+            const rowIndex = 1;
+            const anchorIndex = 0;
+
+            const wrapper = getWrapper({
+                selectedItems,
+            });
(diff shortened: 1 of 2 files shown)


src/components/table/makeSelectable.js
```diff
@@ -98,118 +102,259 @@ function makeSelectable(BaseTable) {
...
+        getSharedHotkeyConfigs = () => {
+            const { hotkeyType } = this.props;
+            return [
+                new HotkeyRecord({
+                    key: 'shift+x',
+                    description: <FormattedMessage {...messages.shiftXDescription} />,
+                    handler: () => {
+                        const { focusedIndex } = this.state;
+
+                        if (focusedIndex === undefined) {
```
jcamden, line 114: It's interesting that `clearFocus` sets focusedIndex to `undefined` instead of `null`. Some people think that it was a mistake to include both in JS.

jcamden, line 114: It's interesting that `clearFocus` sets `focusedIndex` to `undefined` instead of `null`. Some people think that it was a mistake to include both in JS.

src/components/table/makeSelectable.js
```diff
@@ -98,118 +102,259 @@ function makeSelectable(BaseTable) {
             }
         };
 
+        getSharedHotkeyConfigs = () => {
```
jcamden, line 105: Good call to make the shared things shared at this stage :)

src/components/table/makeSelectable.js
```diff
@@ -326,6 +471,61 @@ function makeSelectable(BaseTable) {
             this.onSelect(selectedItems.add(data[focusedIndex]), newFocusedIndex);
         };
 
+        isContinuation = (selectedItemIndecies, sourceIndex, targetIndex) => {
```
jcamden, line 474: Nits: this name is a bit ambiguous. I wasn't sure what it meant. Could we do something like `isContiguousSelection`? Also, `selectedItemIndecies` => `selectedItemIndices`. (One does not create an application to automate publication indexing without finally learning how to spell indices 😂)

src/components/table/makeSelectable.js
```diff
@@ -326,6 +471,61 @@ function makeSelectable(BaseTable) {
...
+        handleShiftKeyDownForGrid = newFocusedIndex => {
+            const { data, loadedData, selectedItems } = this.getProcessedProps();
+            const { focusedIndex } = this.state;
+
+            const dataSize = data.length;
+            const targetIndex = newFocusedIndex < 0 ? 0 : Math.min(newFocusedIndex, dataSize - 1);
+            const isSourceSelected = selectedItems.has(data[focusedIndex]);
+            const isTargetSelected = selectedItems.has(data[targetIndex]);
+
+            // if data is not loaded, we don't want it to be able to be selected
```
jcamden, line 493: Thanks for the helpful comment here. What does it mean when the data is not loaded? Is that ghosted item?

src/components/table/__tests__/makeSelectable.test.js
```diff
@@ -313,6 +315,176 @@ describe('components/table/makeSelectable', () => {
...
+        const { isContinuation } = instance;
+        test('returns true if source is less than target and source - 1 is selected', () => {
+            expect(isContinuation(Set([0]), 1, 2)).toEqual(true);
+        });
+
+        test('returns false if source is less than target and source - 1 is not selected', () => {
+            expect(isContinuation(Set(), 1, 2)).toEqual(false);
+        });
+
+        test('returns true if source is grater than target and source + 1 is selected', () => {
```
jcamden, line 330: nit: ye olde "grater"

jcamden (dismissed):
Great, exhaustive testing. Some nits, but basically good to go.

src/components/table/__tests__/makeSelectable.test.js
ahorowitz123: my arch nemesis

src/components/table/makeSelectable.js
ahorowitz123: Yeah in the case of `FilePage` it is a ghosted item. i didn't want to put any reference of ghosted items in here though since this is BUIE and this HOC is supposed to be reusable

src/components/table/makeSelectable.js
jcamden: Hmm, I would have expected clearFocus to assign `null`, but maybe I don't understand how `null` is intended to be used.

jcamden (approved):
Great stuff.

PR #3083, opened 2022-10-07
feat(classification): update security controls watermarking copy

bxie-box:
Before
BeforeWatermarkingCopy

After
AfterWatermarkingCopy


jpfranco (approved):
LGTM. Just a note that, since the text is live, might be preferable to wait for translations before bumping.

bxie-box:
Will do!

PR #3085, opened 2022-10-11
chore(feed-items): introduce constants for Feed Items types

wpiesiak:
Introduce constants for Feed Items types and replace in all use cases

### DIFF
--- a/src/api/Annotations.js
+++ b/src/api/Annotations.js
@@ -8,7 +8,6 @@ import {
     ERROR_CODE_FETCH_ANNOTATION,
     ERROR_CODE_FETCH_ANNOTATIONS,
     ERROR_CODE_FETCH_REPLIES,
-    FEED_ITEM_TYPE_VERSION,
     PERMISSION_CAN_CREATE_ANNOTATIONS,
     PERMISSION_CAN_DELETE,
     PERMISSION_CAN_EDIT,
@@ -119,7 +118,7 @@ export default class Annotations extends MarkerBasedApi {
             },
             file_version: {
                 id: fileVersionId,
-                type: FEED_ITEM_TYPE_VERSION,
+                type: 'file_version',
             },
         };
 
--- a/src/api/Annotations.js
+++ b/src/api/Annotations.js
@@ -8,6 +8,8 @@ import {
     ERROR_CODE_FETCH_ANNOTATION,
     ERROR_CODE_FETCH_ANNOTATIONS,
     ERROR_CODE_FETCH_REPLIES,
+    FEED_ITEM_TYPE_COMMENT,
+    FEED_ITEM_TYPE_VERSION,
     PERMISSION_CAN_CREATE_ANNOTATIONS,
     PERMISSION_CAN_DELETE,
     PERMISSION_CAN_EDIT,
@@ -63,7 +65,7 @@ export default class Annotations extends MarkerBasedApi {
         // We don't have entries when updating/creating an annotation
         if (!data.entries) {
             // Check if the response is a comment (result of createAnnotationReply)
-            if (data.type && data.type === 'comment') {
+            if (data.type && data.type === FEED_ITEM_TYPE_COMMENT) {
                 this.successCallback(formatComment(data));
                 return;
             }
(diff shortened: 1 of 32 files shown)


MateuszMamczarz (approved):
😮 👏 🥇

src/api/Versions.js
```diff
@@ -249,7 +250,7 @@ class Versions extends OffsetBasedAPI {
             data: {
                 data: {
                     id: versionId,
-                    type: 'file_version',
```
jstoffan, line 252: I don't think this is desired, since versions are used outside of the Activity Feed experience. The API layer and types should drive the experiences, not the other way around.

src/api/Annotations.js
```diff
@@ -63,7 +65,7 @@ export default class Annotations extends MarkerBasedApi {
         // We don't have entries when updating/creating an annotation
         if (!data.entries) {
             // Check if the response is a comment (result of createAnnotationReply)
-            if (data.type && data.type === 'comment') {
+            if (data.type && data.type === FEED_ITEM_TYPE_COMMENT) {
```
jstoffan, line 68: Same comment here as versions, since annotations are used in other contexts.

jstoffan (changes requested):
The data access layer should not be aware of how it's being used, at least in most cases.

src/api/Annotations.js
```diff
@@ -118,7 +119,7 @@ export default class Annotations extends MarkerBasedApi {
             },
             file_version: {
                 id: fileVersionId,
-                type: 'file_version',
+                type: FEED_ITEM_TYPE_VERSION,
```
jstoffan, line 122: Can we revert this change or use a non-feed related constant? The Annotations API module is used outside of the context of Feed.

src/api/__tests__/Comments.test.js
```diff
@@ -28,7 +33,7 @@ describe('api/Comments', () => {
 
     describe('successHandler()', () => {
         const comment = {
-            type: 'comment',
+            type: FEED_ITEM_TYPE_COMMENT,
```
jstoffan, line 36: Can we revert these changes to the Comments tests?

src/api/__tests__/Versions.test.js
```diff
@@ -220,7 +220,7 @@ describe('api/Versions', () => {
                 const requestData = {
                     data: {
                         id: versionId,
-                        type: 'file_version',
+                        type: FEED_ITEM_TYPE_VERSION,
```
jstoffan, line 223: Revert?

src/common/types/annotations.js
```diff
@@ -61,7 +62,7 @@ export type Annotation = {
     status?: FeedItemStatus,
     target: Target,
     total_reply_count?: number,
-    type: 'annotation',
+    type: typeof FEED_ITEM_TYPE_ANNOTATION,
```
jstoffan, line 65: Can we reverse the flow of types here? This file should be the source of truth for annotations-related types rather than the other way around.

jstoffan (changes requested):
We still have a lot of Feed references in areas they shouldn't be in.

src/elements/content-sidebar/versions/StaticVersionSidebar.js
```diff
@@ -32,7 +33,7 @@ const StaticVersionsSidebar = ({ isLoading, onUpgradeClick, parentName }: Props)
         return {
             id: versionNumber,
             version_number: versionNumber,
-            type: 'file_version',
+            type: FEED_ITEM_TYPE_VERSION,
```
jstoffan, line 36: Can we revert, as this sidebar component isn't related to Activity Feed?

src/common/types/tasks.js
```diff
@@ -8,6 +8,7 @@ import type {
     MarkerPaginatedCollection,
     SelectorItems,
 } from './core';
+import { FEED_ITEM_TYPE_TASK } from '../../constants';
```
jstoffan, line 11: Can we revert? Common types should be the source of truth, rather than using Feed in them.

src/elements/content-sidebar/activity-feed/annotation-thread/useRepliesAPI.js
```diff
@@ -65,7 +66,7 @@ const useRepliesAPI = ({
         const replyData = {
             id: uuid,
             tagged_message: message,
-            type: 'comment',
+            type: FEED_ITEM_TYPE_COMMENT,
```
jstoffan, line 69: I think we plan to extract this component from the feed directory at some point, right? Should we keep them separate?

jstoffan (approved):

PR #3088, opened 2022-10-12
feat(content-sidebar): extended annotation events communication

kkuliczkowski-box:
- Added option to emit Activity Sidebar Filter Change event
- Extended `useAnnotatorEvents` hook with
-- listening for Activity Sidebar Filter Change event
-- emit event functions for annotation delete and add actions
- Updated `withAnnotations` and `withSidebarAnnotations` to support delete action

### DIFF
--- a/src/elements/common/annotator-context/withAnnotations.tsx
+++ b/src/elements/common/annotator-context/withAnnotations.tsx
@@ -14,6 +14,7 @@ export type ActiveChangeEventHandler = (event: ActiveChangeEvent) => void;
 
 export type ComponentWithAnnotations = {
     emitActiveAnnotationChangeEvent: (id: string | null) => void;
+    emitActivitySidebarFilterChangeEvent: (status: string) => void;
     emitAnnotationRemoveEvent: (id: string, isStartEvent?: boolean) => void;
     emitAnnotationUpdateEvent: (annotation: Object, isStartEvent?: boolean) => void;
     getAction: (eventData: AnnotationActionEvent) => Action;
@@ -76,6 +77,16 @@ export default function withAnnotations<P extends object>(
             }
 
             annotator.emit('annotations_active_set', id);
+        };
+
+        emitActivitySidebarFilterChangeEvent = (status: string) => {
+            const { annotator } = this;
+
+            if (!annotator) {
+                return;
+            }
+
+            annotator.emit('sidebar.activity_sidebar_filter_change', status);
         };
 
         emitAnnotationRemoveEvent = (id: string, isStartEvent = false) => {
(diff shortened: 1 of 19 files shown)


src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -77,6 +79,16 @@ export default function withAnnotations<P extends object>(
...
         };
 
+        emitActivitySidebarFilterChangeEvent = (status: string) => {
+            const { annotator } = this;
+
+            if (!annotator) {
+                return;
+            }
+
+            annotator.emit('sidebar.activity_sidebar_filter_change', status);
```
jstoffan, line 89: Can we move this to a callback prop on `ActivitySidebar`, instead? It's not really directly related to annotations.

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -147,6 +163,18 @@ export default function withAnnotations<P extends object>(
             });
         };
 
+        handleAnnotationDelete = (eventData: AnnotationActionEvent) => {
+            const { annotation = null, error = null, meta = null } = eventData;
+
+            this.setState({
+                ...this.state,
```
jstoffan, line 170: Minor, but state should be merged automatically without this spread.

src/elements/common/annotator-context/withAnnotations.tsx
kkuliczkowski-box: I could do that but how would we pass the `annotator` object to `ActivitySidebar`?

kkuliczkowski-box: Refactored as discussed.

src/elements/common/annotator-context/withAnnotations.tsx
```diff
@@ -147,6 +163,18 @@ export default function withAnnotations<P extends object>(
             });
         };
 
+        handleAnnotationDelete = (eventData: AnnotationActionEvent) => {
+            const { annotation = null, error = null, meta = null } = eventData;
```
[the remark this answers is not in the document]
kkuliczkowski-box: I don't think we can do it as we need to preserve `eventData` object in order to pass it to `getDeleteAction` method.

src/elements/content-sidebar/ActivitySidebarFilter.js
```diff
@@ -21,13 +21,11 @@ type ActivitySidebarFilterProps = {
 
 const ACTIVITY_FEED_ITEM_ALL = 'all';
 const ACTIVITY_FEED_ITEM_OPEN = COMMENT_STATUS_OPEN;
-const ACTIVITY_FEED_ITEM_RESOLVED = COMMENT_STATUS_RESOLVED;
```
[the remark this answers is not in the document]
kkuliczkowski-box: It's vaguely related as this change is connected to the whole communication between different parts topic. This particular change was the result of discussion whether the filter change to Resolved should have any effect in Canvas and in the end Product decided to remove this option altogether. This was so small that I've decided to include in this PR.

PR #3089, opened 2022-10-12
feat(annotation-activity): hide AnnotationActivityLink

wpiesiak:
- show AnnotationActivityLink base on `hasVerions` ActivitySidebar param
- move logic responsible for AnnotationActivityLink message to `AnnotationActivityLink`
- rewrite `AnnotationActivityLink.test.js` to react-testing-library


src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.scss
```diff
@@ -42,15 +42,6 @@
 .bcs-AnnotationActivity-timestamp {
     display: flex;
     white-space: nowrap;
-
-    .bcs-ActivityTimestamp {
-        &::after {
-            margin-right: 8px;
```
karelee7, line 48: Do we typically use `$bdl-grid-unit` for margin amounts?

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityLink.tsx
```diff
@@ -38,7 +55,7 @@ const AnnotationActivityLink = ({
     };
 
     const handleMouseDown = (event: React.MouseEvent<HTMLButtonElement>) => {
-        if (isDisabled) {
+        if (isFileVersionUnavailable) {
```
karelee7, line 58: Can we do something like `!fileVersion`? Or do we have to specifically check for `null`?

src/elements/content-sidebar/activity-feed/annotations/__tests__/AnnotationActivity.test.js
```diff
@@ -127,30 +127,18 @@ describe('elements/content-sidebar/ActivityFeed/annotations/AnnotationActivity',
...
-        });
-        expect(activityLink.prop('isDisabled')).toBe(true);
-    });
-
     test('should render commenter as a link', () => {
         const wrapper = getWrapper();
         expect(wrapper.find('UserLink').prop('name')).toEqual(mockActivity.item.created_by.name);
     });
 
+    test('should hide AnnotationActivityLink if hasVersions is false', () => {
```
karelee7, line 135: Should we also add a test case for when `hasVersions` is true? Or would the `AnnotationActivityLink` still be hidden in that case?

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityLink.tsx
```diff
@@ -4,28 +4,45 @@ import noop from 'lodash/noop';
...
     ...rest
-}: AnnotationActivityLinkProps): JSX.Element => {
+}: AnnotationActivityLinkProps): JSX.Element | null => {
+    if (shouldHideLink) {
+        return null;
+    }
+
+    const isFileVersionUnavailable = fileVersion == null;
+
+    const linkMessage = isCurrentVersion ? messages.annotationActivityPageItem : messages.annotationActivityVersionLink;
```
JChan106, line 40: Lines 40 - 45 build on each other, we could encapsulate them into a function. Might be easier to read with if statements and guard clauses too

src/elements/content-sidebar/activity-feed/annotations/messages.ts
```diff
@@ -0,0 +1,52 @@
+import { defineMessages } from 'react-intl';
+
+const messages = defineMessages({
+    annotationActivityDeleteMenuItem: {
+        id: 'be.contentSidebar.activityFeed.annotationActivity.annotationActivityDeleteMenuItem',
```
JChan106, line 5: I see some of these ids have `annotationActivity` in the path and some don't. If it's in the same file they all should have the same path right? I've seen we generally follow the filepath so wouldn't the path be `be.contentSidebar.activityFeed.annotations`

wpiesiak: It is basically copied content from `messages.flow.js` to be able to use it in typescript component (I've found this pattern in `src/components/badge` - let me know if there is a better way to handle translation messages in flow and ts at the same time). I would rather not change ids of translations which are already in the codebase for some time.

JChan106 (approved):
lgtm

PR #3093, opened 2022-10-17
feat(content-sidebar): added annotation replies events communication

kkuliczkowski-box:
Added logic supporting Annotation replies related events communication.

### DIFF
--- a/src/elements/common/annotator-context/__tests__/useAnnotatorEvents.test.tsx
+++ b/src/elements/common/annotator-context/__tests__/useAnnotatorEvents.test.tsx
@@ -28,6 +28,80 @@ describe('src/elements/common/annotator-context/useAnnotatorEvents', () => {
         return renderHook(() => useAnnotatorEvents({ ...props, eventEmitter: mockEventEmitter }));
     };
 
+    test('should remove all listeneres on cleanup', () => {
+        const mockOnSidebarAnnotationSelected = jest.fn();
+        const mockOnAnnotationDeleteEnd = jest.fn();
+        const mockOnAnnotationDeleteStart = jest.fn();
+        const mockOnAnnotationUpdateEnd = jest.fn();
+        const mockOnAnnotationUpdateStart = jest.fn();
+        const mockOnAnnotationReplyAddEnd = jest.fn();
+        const mockOnAnnotationReplyAddStart = jest.fn();
+        const mockOnAnnotationReplyDeleteEnd = jest.fn();
+        const mockOnAnnotationReplyDeleteStart = jest.fn();
+        const mockOnAnnotationReplyUpdateEnd = jest.fn();
+        const mockOnAnnotationReplyUpdateStart = jest.fn();
+
+        const { unmount } = getHook({
+            onSidebarAnnotationSelected: mockOnSidebarAnnotationSelected,
+            onAnnotationDeleteEnd: mockOnAnnotationDeleteEnd,
+            onAnnotationDeleteStart: mockOnAnnotationDeleteStart,
+            onAnnotationUpdateEnd: mockOnAnnotationUpdateEnd,
+            onAnnotationUpdateStart: mockOnAnnotationUpdateStart,
+            onAnnotationReplyAddEnd: mockOnAnnotationReplyAddEnd,
+            onAnnotationReplyAddStart: mockOnAnnotationReplyAddStart,
+            onAnnotationReplyDeleteEnd: mockOnAnnotationReplyDeleteEnd,
+            onAnnotationReplyDeleteStart: mockOnAnnotationReplyDeleteStart,
+            onAnnotationReplyUpdateEnd: mockOnAnnotationReplyUpdateEnd,
+            onAnnotationReplyUpdateStart: mockOnAnnotationReplyUpdateStart,
+        });
+
+        unmount();
+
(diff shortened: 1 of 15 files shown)


src/elements/common/annotator-context/__tests__/useAnnotatorEvents.test.tsx
```diff
@@ -80,6 +154,94 @@ describe('src/elements/common/annotator-context/useAnnotatorEvents', () => {
         expect(mockAnnotationUpdateEnd).toBeCalledWith(mockAnnotation);
     });
 
+    test('should call onAnnotationReplyAddStart when proper event is emitted', () => {
+        const mockAnnotationReplyAddStart = jest.fn();
+        const mockEventData = { annotationId: '123', reply: { tagged_message: 'abc' }, requestId: 'comment_456' };
```
JChan106, line 159: Seems like annotationID is '123' in every test, you could make that a variable at the top level. Or you can make a mock object which includes the annotationID and just modify the object in each test

src/elements/common/annotator-context/types.ts
```diff
@@ -37,6 +44,14 @@ export type GetMatchPath = (location?: Location) => match<MatchParams> | null;
 export interface AnnotatorContext {
     emitActiveAnnotationChangeEvent?: (id: string) => void;
     emitAnnotationRemoveEvent?: (id: string, isStartEvent?: boolean) => void;
+    emitAnnotationReplyCreateEvent?: (
+        reply: Object,
```
JChan106, line 48: do we know the the typing of `reply`?

src/elements/common/annotator-context/types.ts
```diff
@@ -60,7 +75,8 @@ export interface Metadata {
 }
 
 export interface AnnotationActionEvent {
-    annotation?: object;
+    annotation?: Object;
```
JChan106, line 78: Do we need to type out these fields?

src/elements/common/annotator-context/types.ts
```diff
@@ -37,6 +44,14 @@ export type GetMatchPath = (location?: Location) => match<MatchParams> | null;
 export interface AnnotatorContext {
     emitActiveAnnotationChangeEvent?: (id: string) => void;
     emitAnnotationRemoveEvent?: (id: string, isStartEvent?: boolean) => void;
+    emitAnnotationReplyCreateEvent?: (
+        reply: Object,
+        requestId: string,
+        annotationId: string,
+        isStartEvent?: boolean,
+    ) => void;
+    emitAnnotationReplyDeleteEvent?: (id: string, annotationId: string, isStartEvent?: boolean) => void;
```
JChan106, line 53: Do we full capitalize id in camelcased names? Something like `annotationID`? Not sure if we try to push for this in BUIE

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -393,14 +397,18 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+        const { api, emitAnnotationReplyDeleteEvent, file } = this.props;
 
+        emitAnnotationReplyDeleteEvent(id, parentId, true);
         api.getFeedAPI(false).deleteReply(
             file,
             id,
             parentId,
             permissions,
-            this.feedSuccessCallback,
+            () => {
```
JChan106, line 408: should we move all of these crud callbacks into separate functions instead of inlining in a call

src/elements/content-sidebar/withSidebarAnnotations.js
```diff
@@ -150,6 +199,43 @@ export default function withSidebarAnnotations(
...
+
+                if (!annotationItem) {
+                    return;
+                }
+
+                // Check if the parent annotation has the reply currently visible
+                const replyItem = annotationItem.replies.find(({ id }) => id === replyId);
+                if (replyItem) {
+                    feedAPI.deleteReplyItem(replyId, annotationId);
+                } else if (annotationItem.total_reply_count !== undefined) {
```
JChan106, line 230: could total_reply_count be 0? If so, the logic below would make a negative I believe

src/elements/common/annotator-context/types.ts
kkuliczkowski-box: It's an object containing annotation updates. It may contain any number of its properties. I didn't want to change the already used approach.

src/elements/common/annotator-context/types.ts
kkuliczkowski-box: I've checked and I haven't found in BUIE any case of the fully capitalized ID in variable names specifically for annotation or comment and just a few examples of such variables in other areas. Having in mind that it's currently broadly used with camel case, I don't think it makes sense to change all those instances at this point.

src/elements/content-sidebar/ActivitySidebar.js
kkuliczkowski-box: Makes sense. I've created separate onSuccess methods.

src/elements/content-sidebar/withSidebarAnnotations.js
kkuliczkowski-box: Good point, I've updated this condition.

JChan106 (approved):

PR #3094, opened 2022-10-17
fix(content-sidebar): ThreadedComments api; handling reply edit logic

kkuliczkowski-box:
- Fixed handling reply edit logic to include `hasMention` attribute in the handler ultimately passed to `Comment` component
- Fixed double `file_id` url param being created in GET requests in `ThreadedComments` API

### DIFF
--- a/src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadContent.js
+++ b/src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadContent.js
@@ -9,6 +9,7 @@ import useAnnotationAPI from './useAnnotationAPI';
 
 import type { BoxItem, SelectorItems, User } from '../../../../common/types/core';
 import type { ErrorContextProps } from '../../../../common/types/api';
+import type { BoxCommentPermission, FeedItemStatus } from '../../../../common/types/feed';
 
 import './AnnotationThreadContent.scss';
 
@@ -54,6 +55,16 @@ const AnnotationThreadContent = ({
         errorCallback: onError,
     });
 
+    const onReplyEditHandler = (
+        id: string,
+        text: string,
+        status?: FeedItemStatus,
+        hasMention?: boolean,
+        replyPermissions: BoxCommentPermission,
+    ) => {
+        handleEditReply(id, text, status, replyPermissions);
+    };
+
     return (
         <ActivityThread
             hasReplies
@@ -64,7 +75,7 @@ const AnnotationThreadContent = ({
             mentionSelectorContacts={mentionSelectorContacts}
             onReplyCreate={handleCreateReply}
             onReplyDelete={handleDeleteReply}
-            onReplyEdit={handleEditReply}
+            onReplyEdit={onReplyEditHandler}
             replies={replies}
             repliesTotalCount={replies.length}
         >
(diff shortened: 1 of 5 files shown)


jankowiakdawid (dismissed):

src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadContent.js
```diff
@@ -54,6 +55,16 @@ const AnnotationThreadContent = ({
         errorCallback: onError,
     });
 
+    const onReplyEditHandler = (
+        id: string,
+        text: string,
+        status?: FeedItemStatus,
+        hasMention?: boolean,
```
jstoffan, line 62: It looks like this value is unused. Can we omit it?

src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadContent.js
```diff
@@ -54,6 +55,16 @@ const AnnotationThreadContent = ({
...
+        replyPermissions: BoxCommentPermission,
+    ) => {
+        handleEditReply(id, text, status, replyPermissions);
```
jstoffan, line 65: Minor, but it looks like the naming got a bit twisted up with `useAnnotationAPI`. We'd typically expect callback props to use "on" while internal handlers use "handle", so this is inverted.

src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadContent.js
```diff
@@ -9,6 +9,7 @@ import useAnnotationAPI from './useAnnotationAPI';
 
 import type { BoxItem, SelectorItems, User } from '../../../../common/types/core';
 import type { ErrorContextProps } from '../../../../common/types/api';
+import type { BoxCommentPermission, FeedItemStatus } from '../../../../common/types/feed';
```
jstoffan, line 12: Minor, but can we alphabetize these imports?

jstoffan (dismissed):
Do we need any new tests?

src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadContent.js
kkuliczkowski-box: I've actually needed to create this internal handler (and add `hasMention` argument) as in the end it's passed to a `Comment` component's `onEdit` prop, which is called with `hasMention` as 4th argument. `ThreadedReplies` api actually does not require `hasMention` so it's being abandoned here and not used in `handleEditReply` call.

src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadContent.js
kkuliczkowski-box: I agree, the naming of those handleXYZ functions in `useAnnotationAPI` could be changed to onXYZ, but I think it should be done in a separate PR. Here I only wanted to fix the issue we had with passing `handleEditReply` (without the `hasMention` as 4th argument) directly to  `onReplyEdit` prop of `ActivityThread`.

jstoffan (approved):

PR #3095, opened 2022-10-17
fix(preview-sidebar): fix version not updating on small screens

greathmaster:
## Overview
Currently, if a version is selected in the Preview sidebar and the Preview sidebar is closed, it reverts to the latest version of the document.

This creates problems for small screens where the Preview sidebar is moved to the bottom and becomes a vertical sliding drawer. When an old version is selected and the drawer is closed, a user cannot view the old version because, it reverts to the latest version of the document.

This PR fixes the issue by removing the version reset performed when the drawer is closed.

Note this would also affect the regular desktop experience. If a user closes the Preview Sidebar after selecting the version, it would remain on that version instead of changing back to the latest version.

### Before
Preview-Drawer-Before

### After
Preview-Drawer-After

### DIFF
--- a/src/elements/content-sidebar/versions/VersionsSidebarContainer.js
+++ b/src/elements/content-sidebar/versions/VersionsSidebarContainer.js
@@ -95,11 +95,6 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
         }
     }
 
-    componentWillUnmount() {
-        // Reset the current version id since the wrapping route is no longer active
-        this.props.onVersionChange(null);
-    }
-
     handleActionDelete = (versionId: string): Promise<void> => {
         this.setState({ isLoading: true });
 
--- a/src/elements/content-sidebar/versions/__tests__/VersionsSidebarContainer.test.js
+++ b/src/elements/content-sidebar/versions/__tests__/VersionsSidebarContainer.test.js
@@ -58,17 +58,6 @@ describe('elements/content-sidebar/versions/VersionsSidebarContainer', () => {
         });
     });
 
-    describe('componentWillUnmount', () => {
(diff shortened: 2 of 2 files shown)


jstoffan:
@greathmaster, has the desktop change been approved by the relevant stakeholders?

greathmaster (commented):
I've asked the relevant stakeholders and they are okay with the the behavior change on desktop experiences.

src/elements/content-sidebar/versions/VersionsSidebarContainer.js
```diff
@@ -95,11 +95,6 @@ class VersionsSidebarContainer extends React.Component<Props, State> {
         }
     }
 
-    componentWillUnmount() {
```
bfoxx1906, line 98: Hey so are there any possible side effects of removing this code completely.  If we can remove it then it looks like it may have been unnecessary code in the first place.

jstoffan: It also means that switching to other sidebars, like "Details" or "Metadata", will not revert to the current version preview. The only mechanism to do so will be the button in the header.

Have we heard back from the relevant stakeholders that this change is acceptable even on larger screen sizes?

greathmaster: Have approval for the version behavior change, ie to show the previous version after closing the right sidebar on large screens. However, let me circle back after checking about the "Details" and "Metadata" experiences.

jstoffan: @greathmaster, understood. I would think the Next/Previous buttons to navigate forward and backward through previews in a folder may also be affected.

PR #3097, opened 2022-10-18
feat(annotation-thread): introduce useAnnotationThread and handle useAnnotatorEvents

wpiesiak:
Created `useAnnotationThread` as an additional logic layer connecting `useAnnotationAPI`, `useRepliesAPI` and `useAnnotatorEvents` for specific use in `AnnotationThread` component. 

Implement `useAnnotatorEvents` in `useAnnotationThread` to handle events emitted in the Sidebar

Updated useAnnotationAPI to handle only API logic (additional functionalities moved to `useAnnotationThread`)


src/elements/content-sidebar/activity-feed/annotation-thread/useAnnotationAPI.js
```diff
@@ -68,14 +71,15 @@ const useAnnotationAPI = ({
     annotationId,
     api,
     currentUser,
+    eventEmitter,
```
jstoffan, line 74: We should really try to isolate the event pattern as much as possible and not let it leak into other areas. Can we orchestrate this at the common layer for both these hooks?

src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadContent.js
```diff
@@ -73,8 +73,8 @@ const AnnotationThreadContent = ({
             isAlwaysExpanded
             isRepliesLoading={isLoading}
             mentionSelectorContacts={mentionSelectorContacts}
-            onReplyCreate={handleCreateReply}
-            onReplyDelete={handleDeleteReply}
+            onReplyCreate={handleReplyCreate}
+            onReplyDelete={handleReplyDelete}
             onReplyEdit={onReplyEditHandler}
```
kkuliczkowski-box, line 78: This handler has now inconsistent naming with other ones.

src/elements/content-sidebar/activity-feed/annotation-thread/useAnnotationThread.js
```diff
@@ -0,0 +1,241 @@
...
+    const [isLoading, setIsLoading] = React.useState<boolean>(true);
+
+    // handling events from Sidebar
+    const onAnnotationDeleteStart = (id: string) => {
+        if (annotation !== undefined && id === annotationId) {
+            setAnnotation(prevAnnotation => ({ ...prevAnnotation, isPending: true }));
+        }
+    };
+
+    const onAnnotationUpdateStart = (id: string) => {
```
kkuliczkowski-box, line 93: This function duplicates `onAnnotationDeleteStart`. I think it could be just a single one named `setAnnotationPending` or sth similar and it can be passed to both hook "Start" props callbacks.

src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadContent.js
```diff
@@ -40,31 +42,18 @@ const AnnotationThreadContent = ({
...
-        handleEdit,
-        handleStatusChange,
-        handleDelete,
-        handleCreateReply,
-        handleDeleteReply,
-        handleEditReply,
-    } = useAnnotationAPI({
+        annotationActions: { handleAnnotationStatusChange, handleAnnotationDelete, handleAnnotationEdit },
+        repliesActions: { handleReplyEdit, handleReplyCreate, handleReplyDelete },
+    } = useAnnotationThread({
```
jstoffan, line 47: We should reconsider this approach in the future, when time permits. It seems like the underlying state should live in the consuming application rather than in the component.

jstoffan (approved):

PR #3099, opened 2022-10-18
feat(icons): add addl props to Lock16 icon for customization

ivanthai:
Add color prop to Lock16

Screen Shot 2022-10-18 at 4 10 43 PM


ivanthai:
Overriding colors via scss

PR #3103, opened 2022-10-20
fix(content-sidebar): reply operations interactions

kkuliczkowski-box:
- Added `modifyFeedItemRepliesCountBy` method for changing replies count
- Used `modifyFeedItemRepliesCountBy` in the create and delete reply actions
- Refactored `withSidebarAnnotations` to also use modifyFeedItemRepliesCountBy
- Updated `updateAnnotation` and `updateThreadedComment` methods to not update the feed item with `replies` and `total_reply_count` as their values are not included in the server response

### DIFF
--- a/src/api/Feed.js
+++ b/src/api/Feed.js
@@ -220,7 +220,8 @@ class Feed extends Base {
             annotationId,
             permissions,
             feedItemChanges,
-            (annotation: Annotation) => {
+            // Do not update replies and total_reply_count props as their current values are not included in the response
+            ({ replies, total_reply_count, ...annotation }: Annotation) => {
                 this.updateFeedItem(
                     {
                         ...annotation,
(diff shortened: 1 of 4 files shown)


src/api/Feed.js
```diff
@@ -505,9 +507,12 @@ class Feed extends Base {
 
         this.updateFeedItem({ isRepliesLoading: true }, commentFeedItemId);
 
-        const successCallbackFn = (comments: ThreadedCommentsType) => {
-            this.updateFeedItem({ isRepliesLoading: false, replies: comments.entries }, commentFeedItemId);
-            successCallback(comments.entries);
+        const successCallbackFn = ({ entries }: ThreadedCommentsType) => {
```
JChan106, line 510: If this is only using the entries, should we pass in entries field instead of the object?

JChan106 (approved):

src/api/Feed.js
kkuliczkowski-box: In this case we're using only entries but I think the API should return the whole response including marker data as this is the common behaviour in all API methods.

PR #3106, opened 2022-10-21
feat(annotation-thread): Emit events when Annotation is created

jankowiakdawid:
Some of the events are handled in #3097

### DIFF
--- a/src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadCreate.js
+++ b/src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadCreate.js
@@ -1,9 +1,14 @@
 // @flow
 import React from 'react';
 import classNames from 'classnames';
+import uniqueId from 'lodash/uniqueId';
+
+import type EventEmitter from 'events';
+
 import API from '../../../../api/APIFactory';
 import CommentForm from '../comment-form';
 import { getBadUserError } from '../../../../utils/error';
+import useAnnotatorEvents from '../../../common/annotator-context/useAnnotatorEvents';
 
 import type { Annotation, Target } from '../../../../common/types/annotations';
 import type { BoxItem, SelectorItems, User } from '../../../../common/types/core';
@@ -14,6 +19,7 @@ import './AnnotationThreadCreate.scss';
 type Props = {
     api: API,
     currentUser: User,
+    eventEmitter: EventEmitter,
     file: BoxItem,
     getAvatarUrl: string => Promise<?string>,
     getMentionWithQuery: (searchStr: string) => void,
@@ -28,6 +34,7 @@ const AnnotationThreadCreate = ({
     currentUser,
     getAvatarUrl,
     getMentionWithQuery,
+    eventEmitter,
     file,
     handleCancel,
     mentionSelectorContacts,
@@ -36,16 +43,21 @@ const AnnotationThreadCreate = ({
     target,
 }: Props) => {
     const [isPending, setIsPending] = React.useState(false);
+    const events = useAnnotatorEvents({ eventEmitter });
 
     const handleCreate = (text: string): void => {
         const { id, permissions = {}, file_version = {} } = file;
+        const requestId = uniqueId('annotation_');
         setIsPending(true);
+        events.emitAddAnnotationStartEvent({ text }, requestId);
+
         if (!currentUser) {
             throw getBadUserError();
         }
 
         const successCallback = (annotation: Annotation) => {
             onAnnotationCreate(annotation);
+            events.emitAddAnnotationEndEvent(annotation, requestId);
         };
 
         const payload = {
--- a/src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadCreate.js
+++ b/src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadCreate.js
@@ -43,13 +43,12 @@ const AnnotationThreadCreate = ({
     target,
 }: Props) => {
     const [isPending, setIsPending] = React.useState(false);
-    const events = useAnnotatorEvents({ eventEmitter });
+    const { emitAddAnnotationStartEvent, emitAddAnnotationEndEvent } = useAnnotatorEvents({ eventEmitter });
 
     const handleCreate = (text: string): void => {
         const { id, permissions = {}, file_version = {} } = file;
         const requestId = uniqueId('annotation_');
         setIsPending(true);
-        events.emitAddAnnotationStartEvent({ text }, requestId);
 
         if (!currentUser) {
             throw getBadUserError();
@@ -57,7 +56,7 @@ const AnnotationThreadCreate = ({
 
         const successCallback = (annotation: Annotation) => {
             onAnnotationCreate(annotation);
-            events.emitAddAnnotationEndEvent(annotation, requestId);
+            emitAddAnnotationEndEvent(annotation, requestId);
         };
 
         const payload = {
@@ -65,6 +64,7 @@ const AnnotationThreadCreate = ({
             target,
         };
 
+        emitAddAnnotationStartEvent({ text }, requestId);
         api.getAnnotationsAPI(false).createAnnotation(
             id,
             file_version.id,
--- a/src/elements/content-sidebar/activity-feed/annotation-thread/__tests__/AnnotationThreadContent.test.js
+++ b/src/elements/content-sidebar/activity-feed/annotation-thread/__tests__/AnnotationThreadContent.test.js
@@ -4,7 +4,6 @@ import { render } from '@testing-library/react';
 import { IntlProvider } from 'react-intl';
 import AnnotationThreadContent from '../AnnotationThreadContent';
 import { annotation, user } from '../../../../../__mocks__/annotations';
-import useAnnotatorEvents from '../../../../common/annotator-context/useAnnotatorEvents';
 
 import commonMessages from '../../../../common/messages';
 import messages from '../messages';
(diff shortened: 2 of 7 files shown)


src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadCreate.js
```diff
@@ -36,16 +43,21 @@ const AnnotationThreadCreate = ({
     target,
 }: Props) => {
     const [isPending, setIsPending] = React.useState(false);
+    const events = useAnnotatorEvents({ eventEmitter });
```
kkuliczkowski-box, line 46: Minor, but you could potentially do destructuring assignment of `emitAddAnnotationStartEvent` and `emitAddAnnotationEndEvent` here.

src/elements/content-sidebar/activity-feed/annotation-thread/__tests__/AnnotationThreadCreate.test.js
```diff
@@ -93,6 +107,24 @@ describe('elements/content-sidebar/activity-feed/annotation-thread/AnnotationThr
...
+        mockCreateAnnotation = jest.fn((fileId, fileVersionId, payload, filePermissions, successCallback) => {
+            successCallback(annotation);
+        });
+        const props = getDefaultProps();
+        const { getByText } = getWrapper(props);
+        const { eventEmitter } = props;
+
+        fireEvent.click(getByText('Post'));
+
+        expect(useAnnotatorEvents).toHaveBeenCalledWith({ eventEmitter });
```
kkuliczkowski-box, line 120: In BUIE it's common to use `ToBeCalledWith` instead of `toHaveBeenCalledWith`.
Also applies to other use cases below and in `useAnnotationThread.test.js`

src/elements/content-sidebar/activity-feed/annotation-thread/useAnnotationThread.js
```diff
@@ -97,12 +97,16 @@ const useAnnotationThread = ({
         }
     };
 
-    useAnnotatorEvents({
+    const events = useAnnotatorEvents({
```
kkuliczkowski-box, line 100: Minor, but you could potentially do destructuring assignment of `emitAnnotationActiveChangeEvent` here.

src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadCreate.js
```diff
@@ -36,16 +43,21 @@ const AnnotationThreadCreate = ({
...
 
     const handleCreate = (text: string): void => {
         const { id, permissions = {}, file_version = {} } = file;
+        const requestId = uniqueId('annotation_');
         setIsPending(true);
+        events.emitAddAnnotationStartEvent({ text }, requestId);
```
jstoffan, line 52: I thought we made `requestId` optional? It's actually coming from Redux in box-annotations, so maybe it's not necessary? I suspect it has a different format, as well.

src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadCreate.js
jstoffan: I agree with @kkuliczkowski-box, we should destructure the handlers here.

src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadCreate.js
```diff
@@ -36,16 +43,21 @@ const AnnotationThreadCreate = ({
...
+
         if (!currentUser) {
             throw getBadUserError();
         }
 
         const successCallback = (annotation: Annotation) => {
             onAnnotationCreate(annotation);
```
jstoffan, line 59: It really seems like the consuming component of `AnnotationThread` should be managing the event emitter rather than passing it down to every individual component. It's something we should look to refactor in the future.

src/elements/content-sidebar/activity-feed/annotation-thread/__tests__/AnnotationThreadContent.test.js
```diff
@@ -32,17 +32,26 @@ jest.mock('../useAnnotationThread', () => {
     }));
 });
 
+jest.mock('../../../../common/annotator-context/useAnnotatorEvents', () => {
+    const mockedEmitAnnotationActiveChangeEvent = jest.fn();
```
jstoffan, line 36: Minor, but this can be inlined.

src/elements/content-sidebar/activity-feed/annotation-thread/__tests__/AnnotationThreadContent.test.js
```diff
@@ -32,17 +32,26 @@ jest.mock('../useAnnotationThread', () => {
...
+
+    return jest.fn(() => ({
```
jstoffan, line 38: Minor, but we can use an implicit return here.

src/elements/content-sidebar/activity-feed/annotation-thread/__tests__/AnnotationThreadContent.test.js
```diff
@@ -32,17 +32,26 @@ jest.mock('../useAnnotationThread', () => {
     }));
 });
 
+jest.mock('../../../../common/annotator-context/useAnnotatorEvents', () => {
```
jstoffan, line 35: Should we declare this mock in `common/annotator-context/__mocks__/useAnnotatorEvents` so it can be reused?

src/elements/content-sidebar/activity-feed/annotation-thread/__tests__/AnnotationThreadContent.test.js
```diff
@@ -32,17 +32,26 @@ jest.mock('../useAnnotationThread', () => {
...
+    }));
+});
+
 describe('elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadContent', () => {
     const mockGetAvatarUrl = jest.fn(() => Promise.resolve());
+    const mockEventEmitter = {};
 
-    const defaultProps = {
-        annotationId: mockedAnnotation.id,
+    const defaultProps = () => ({
```
jstoffan, line 47: Can we rename this to `getDefaultProps` now that it's a helper function?

src/elements/content-sidebar/activity-feed/annotation-thread/__tests__/AnnotationThreadCreate.test.js
```diff
@@ -22,12 +24,22 @@ jest.mock('../../comment-form', () => props => {
...
+        emitAddAnnotationEndEvent: mockedEmitAddAnnotationEndEvent,
+    }));
+});
 
 describe('elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadCreate', () => {
     const mockHandleCancel = jest.fn();
     const mockOnAnnotationCreate = jest.fn();
     const mockOnError = jest.fn();
     let mockCreateAnnotation = jest.fn();
+    const mockEventEmitter = {};
```
jstoffan, line 42: We typically want to group out `let` and `const` with each other if they're colocated.

src/elements/content-sidebar/activity-feed/annotation-thread/useAnnotationThread.js
```diff
@@ -97,12 +97,16 @@ const useAnnotationThread = ({
...
         onAnnotationDeleteStart: setAnnotationPending,
         onAnnotationUpdateEnd,
         onAnnotationUpdateStart: setAnnotationPending,
     });
 
+    React.useEffect(() => {
+        events.emitAnnotationActiveChangeEvent(annotationId, fileId);
+    }, [annotationId, events, fileId]);
```
jstoffan, line 108: Is this stable? I think `events` has a lot in it and could end up changing somewhat often?

src/elements/content-sidebar/activity-feed/annotation-thread/__tests__/useAnnotationThread.test.js
```diff
@@ -235,5 +236,13 @@ describe('src/elements/content-sidebar/activity-feed/useAnnotationThread', () =>
 
             expect(result.current.annotation).toEqual({ ...updatedAnnotation, isPending: false });
         });
+
+        test('Should emit annotation active change event when used', () => {
+            const event = useAnnotatorEvents();
+
+            getHook();
+
+            expect(event.emitAnnotationActiveChangeEvent).toHaveBeenCalled();
```
jstoffan, line 245: Is this desirable if nothing has actually changed from one hook call to the next?

src/elements/content-sidebar/activity-feed/annotation-thread/__tests__/AnnotationThreadContent.test.js
jankowiakdawid: it was so easy to mock here. I thing creating common mock would be too much hustle for just 2 places. If that would be used in more places then that would be worth while

src/elements/content-sidebar/activity-feed/annotation-thread/useAnnotationThread.js
jankowiakdawid: I think you're right. We don't need events there. Maybe just `annotationId` and `fileId`. We want to emit this event only once when Annotation is displayed

src/elements/content-sidebar/activity-feed/annotation-thread/__tests__/useAnnotationThread.test.js
jankowiakdawid: Actually we call tested hook (`useAnnotationThread`) only once here. So it's desirable for the event to be called once. I'll try to improve the test to make it more clear.

jankowiakdawid:
We keep this PR on hold for now. First we want to take `eventEmmiter` out of child components in a separate PR.

PR #3107, opened 2022-10-23
fix(tooltips): Error tooltips were being truncated on smaller screens

greathmaster:
# Overview
Error tooltips on the Tasks Modal and PillSelector component were being truncated on smaller screens

# Summary of Changes
- Repositioned the error tooltips to stay 'inside' the containing element

# Demo

### Fullscreen (small screen)
Screen Shot 2022-10-23 at 10 33 39

### Regular (desktop size)
Screen Shot 2022-10-23 at 10 51 09

### Fullscreen (small screen)
Screen Shot 2022-10-23 at 10 29 07

### Regular (desktop size)
Screen Shot 2022-10-23 at 10 29 56


src/components/pill-selector-dropdown/PillSelector.js
```diff
@@ -205,7 +205,7 @@ class PillSelector extends React.Component<Props, State> {
         };
 
         return (
-            <Tooltip isShown={hasError} text={error || ''} position="middle-right" theme="error">
+            <Tooltip isShown={hasError} text={error || ''} position="bottom-right" theme="error">
```
tjuanitas, line 208: Is this confirmed with design that all PillSelectors will be changed to bottom-right?

tjuanitas: we can add a `errorPosition` / `errorTooltipPosition` prop if needed

greathmaster: Yes. Just in case, I went back and reviewed our conversation. The one difference is a slight preference for using `bottom-left` instead of `bottom-right`. I've gone ahead and made the change.

tjuanitas (approved):

bfoxx1906 (approved):

PR #3108, opened 2022-10-23
fix(modal): Modal header for long filenames was displaying incorrectly

greathmaster:
# Overview
Modal title was overflowing the header for long filenames. The title is suppose to be limited to 2 lines on small screens.

# Summary of Changes

- Adds `line-clamp` mixin
- Adds above mixin to fullscreen `modal-title` class

# Demo

## Before
Screen Shot 2022-10-23 at 12 36 12 PM

## After
Screen Shot 2022-10-23 at 12 32 42 PM


src/styles/mixins/_typography.scss
```diff
@@ -42,3 +42,30 @@
...
     letter-spacing: .3px;
 }
+
+@mixin target-ie11 {
+    @media all and (-ms-high-contrast: none), (-ms-high-contrast: active) {
+        @content;
+    }
+}
+
+@mixin line-clamp($max-lines: 2, $line-height: 1.5) {
```
tjuanitas, line 52: New mixins seem to use a `bdl` prefix so I think we should add that

tjuanitas, line 52: I was thinking something like `bdl-lineClamp` but maybe it'll be better to have a more descriptive name like `bdl-truncateMultiline` or `bdl-limitTextLines` or some other variation?

src/styles/mixins/_typography.scss
```diff
@@ -42,3 +42,30 @@
...
+    display: -webkit-box; /* stylelint-disable-line value-no-vendor-prefix */
+    overflow: hidden;
+    line-height: $line-height;
+    white-space: normal;
+    overflow-wrap: break-word;
+
+    /* autoprefixer: ignore next */
```
tjuanitas, line 59: is this needed? it looks like it might've been fixed https://github.com/postcss/autoprefixer/issues/1220

greathmaster: My understanding is it is still needed because we introduced the EUA version last year, which was after this fix.

src/styles/mixins/_typography.scss
greathmaster: We are already using this `line-clamp` mixin in EUA.

My hope was I could...

- introduce it in BUIE since we have some strings that need it
- import it back to EUA and update to pull it from BUIE
- remove it from EUA to keep a single source of truth.

My understanding is people have been using the mixin as `line-clamp` already.  So renaming might complicate things.

tjuanitas: Gotcha. I think it would be good to give it a namespace (same with `target-ie11`) so there's no naming conflicts when used in parent apps and it's more apparent/explicit that this mixin is coming from BUIE.

src/styles/mixins/_typography.scss
tjuanitas: ah I see. the fix was released in 9.6 and EUA is at 10 so I'm not sure if it was originally needed

tjuanitas (approved):

PR #3109, opened 2022-10-24
feat(make-selectable): expose selectToggle function

ahorowitz123:


jcamden (approved):

jcamden (approved):

src/components/table/makeSelectable.js
```diff
@@ -581,6 +585,7 @@ function makeSelectable(BaseTable) {
                         className={classNames(className, 'is-selectable')}
                         focusedIndex={focusedIndex}
                         focusedItem={focusedItem}
+                        onCheckboxClick={this.onCheckboxClick}
```
tjuanitas, line 588: What's the use case for this? I think this prop name may be too specific since checkboxes aren't a normal characteristic of a table / table row.

Does it make sense to call this something like `onSelectToggle` / `onRowSelectToggle`?

PR #3116, opened 2022-10-27
feat(insights): Added support for Advanced Content Insights

marcoartaviaq:
- Added a new component that handles ACI activation and deactivation.
Screenshot 2022-11-23 at 3 43 44 PM
- Unified Share Modal (USM): Added the option to toggle on/off Advanced Content Insights for a specific file.
Screenshot 2022-11-23 at 3 41 48 PM
- Modal Dialog: Added the option to show a back button (and a callback) inside the modal.
Screenshot 2022-11-23 at 3 44 38 PM
- Charts: Added the capability to show a label in each bar and return the bar’s offset position while hovering over it. 
Screenshot 2022-11-23 at 3 44 55 PM
- Content Preview: Send Advanced Content Insights options to Box Content Preview.


src/features/unified-share-modal/UnifiedShareForm.js
```diff
@@ -744,6 +750,20 @@ class UnifiedShareForm extends React.Component<USFProps, State> {
                         />
                     )}
 
+                    {isAdvancedContentInsightsEnabled &&
```
bfoxx1906, line 753: Can we create a const boolean value to represent this condition instead?

src/features/advanced-content-insights/flowTypes.js
```diff
@@ -0,0 +1,22 @@
+// @flow
+export type contentInsightsConfigType = {
```
tjuanitas, line 2: should this be pascal case?

src/features/advanced-content-insights/flowTypes.js
```diff
@@ -0,0 +1,22 @@
...
+    isActive: boolean,
+    requireEmail: false,
+    requireNotification: false,
+};
+
+export type ContentInsightsSectionTypes = {
+    contentInsightsConfig?: contentInsightsConfigType,
+    /** Whether the advanced content insights FF is enabled */
+    isAdvancedContentInsightsEnabled?: boolean,
+    onAdvancedInsightsEmailToggle?: (isEnabled: boolean) => void,
```
tjuanitas, line 12: I see a mixed usage of `AdvancedContentInsights`, `AdvancedInsights`, `ContentInsights`, could we consolidate these?

src/features/advanced-content-insights/ContentInsightsSection.js
```diff
@@ -0,0 +1,145 @@
+// @flow
```
tjuanitas, line 1: since this is a brand new file, could we write this in TypeScript instead of JS + flow?

src/features/advanced-content-insights/ContentInsightsSection.js
```diff
@@ -0,0 +1,145 @@
...
+    item: itemtype | null,
+    onAdvancedInsightsEmailToggle?: (isEnabled: boolean) => void,
+    onAdvancedInsightsNotificationToggle?: (isEnabled: boolean) => void,
+    onAdvancedInsightsToggle?: (isEnabled: boolean) => void,
+    renderOptionsInDropdown?: boolean,
+    submitting: boolean,
+};
+
+type CheckboxWrapperProps = {
+    children: Node,
```
tjuanitas, line 33: if you do `React.Node` does it prevent having to do the extra import of `import type { Node } from 'react';`?

src/features/advanced-content-insights/ContentInsightsSection.js
```diff
@@ -0,0 +1,145 @@
...
+import DropdownMenu from '../../components/dropdown-menu';
+import Gear from '../../icon/line/Gear16';
+import IconInfo from '../../icons/general/IconInfo';
+import { Menu, MenuItem } from '../../components/menu';
+import PlainButton from '../../components/plain-button';
+import Toggle from '../../components/toggle';
+import Tooltip from '../../components/tooltip';
+import { ITEM_TYPE_FILE } from '../../common/constants';
+
+import type { item as itemtype } from '../unified-share-modal/flowTypes';
```
tjuanitas, line 19:
```suggestion
import type { item as ItemType } from '../unified-share-modal/flowTypes';
```

src/features/advanced-content-insights/ContentInsightsSection.js
```diff
@@ -0,0 +1,145 @@
...
+                    </DropdownMenu>
+                )}
+            </>
+        );
+    };
+
+    const renderToggle = () => {
+        //  Render the require email and require notification options
+        return (
+            <div className="shared-link-toggle-row">
```
tjuanitas, line 112: we should use SUIT naming for all new classes:
https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md

src/features/advanced-content-insights/ContentInsightsSection.js
```diff
@@ -0,0 +1,145 @@
...
+                    <DropdownMenu className="ContentInsightsSection-dropdownMenu" isRightAligned>
+                        <PlainButton type="button" data-testid="insights-options-dropdown-toggle">
+                            <Gear />
+                        </PlainButton>
+                        <Menu>{renderCheckboxes()}</Menu>
+                    </DropdownMenu>
+                )}
+            </>
+        );
+    };
```
tjuanitas, line 107: I think it would be cleaner to have `renderCheckboxes` return an array of `Checkbox` components

then you could use a map in `<Menu>...</Menu>` like `checkboxes.map(checkbox => <MenuItem>{checkbox}</MenuItem>` which removes the need for `CheckboxWrapper`

src/features/advanced-content-insights/ContentInsightsSection.js
```diff
@@ -0,0 +1,145 @@
...
+    submitting,
+}: Props) {
+    // TODO: Add extension/type checking, since we only support documents
+    if (!contentInsightsConfig || !item || item.type !== ITEM_TYPE_FILE) {
+        return null;
+    }
+
+    const { isActive, requireEmail, requireNotification } = contentInsightsConfig;
+    const showDescriptionInTooltip = !renderOptionsInDropdown || isActive;
+    const helpLink = !showDescriptionInTooltip && (
```
tjuanitas, line 56: could you add this directly in `advancedContentInsightsDescription`?

tjuanitas (commented):
could you add screenshots of these changes?

I think it would be helpful in the future to split changes like these into smaller PRs

src/components/modal/Modal.scss
```diff
@@ -82,6 +82,20 @@
     margin-top: 20px;
 }
 
+.modal-back-button {
```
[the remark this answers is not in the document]
marcoartaviaq: Updated the description above

src/components/modal/Modal.scss
```diff
@@ -82,6 +82,20 @@
...
+    align-items: center;
+    justify-content: center;
+    padding: 0;
+    background: none;
+    border: 0;
+    cursor: pointer;
+}
+
+.modal-back-button + .modal-title {
+    margin-left: 12px;
```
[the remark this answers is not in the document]
marcoartaviaq: Added the screenshot. According to design, the back button is next to the title, so I think it's better to use relative position.

src/components/modal/ModalDialog.js
```diff
@@ -24,7 +26,9 @@ type Props = {
     closeButtonProps: Object,
     intl: Object,
     modalRef?: Function,
+    onBackClick?: Function,
```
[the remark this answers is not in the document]
marcoartaviaq: Changed. I think it makes sense to use a similar naming pattern.

src/components/modal/ModalDialog.js
```diff
@@ -5,7 +5,9 @@ import omit from 'lodash/omit';
 import uniqueId from 'lodash/uniqueId';
 import { defineMessages, injectIntl } from 'react-intl';
 
+import { bdlGray50 } from '../../styles/variables';
 import IconClose from '../../icon/fill/X16';
+import IconBack from '../../icons/general/IconShow';
```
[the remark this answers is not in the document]
marcoartaviaq: @tjuanitas  I found the icon under the `icon` folder (fill/Arrow16) but it is rotated. I fixed it with CSS, but should I ask the designer for an updated icon?

src/features/advanced-content-insights/ContentInsightsSection.js
```diff
@@ -0,0 +1,145 @@
...
+
+type CheckboxWrapperProps = {
+    children: Node,
+    wrapComponent: boolean,
+};
+
+export const CheckboxWrapper = ({ wrapComponent, children }: CheckboxWrapperProps) =>
+    wrapComponent ? <MenuItem>{children}</MenuItem> : children;
+
+export default function ContentInsightsSection({
```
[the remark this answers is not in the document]
marcoartaviaq: Advanced Content Insights is a different from Content Insights.

src/features/advanced-content-insights/ContentInsightsSection.js
marcoartaviaq: I changed the file's name to AdvancedContentInsightsToggle. We received a last update to drop the options, so it only has the toggle for now.

src/features/advanced-content-insights/ContentInsightsSection.js
marcoartaviaq: Dropped the Advanced Content Insights options because they will be decoupled.

src/features/content-insights/charts/bar/BarChart.tsx
```diff
@@ -66,7 +72,8 @@ function BarChart({
                 <Bar
                     key={datum[labelAccessor]}
                     direction={direction}
-                    onMouseEnter={() => handleBarMouseEnter({ datum })}
+                    label={showAxisLabel && datum[labelAccessor]}
```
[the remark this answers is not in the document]
marcoartaviaq: The labelAccesor is the key used on the datum to indicate the prop that contains the value of the label. This should be localized, but it is a task for the parent app to send it already localized.

src/features/content-insights/charts/bar/Bar.scss
```diff
@@ -11,6 +11,7 @@ $ca-Bar-borderRadius: 10px;
 }
 
 .ca-Bar {
+    position: relative;
```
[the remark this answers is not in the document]
marcoartaviaq: It was a specific case on parent app. Removed it for consistency and added the specific css on parent

src/features/content-insights/charts/bar/Bar.tsx
```diff
@@ -11,12 +11,20 @@ const DEFAULT_SIZE = 50;
 interface Props {
     color?: string;
     direction?: Direction;
-    onMouseEnter?: () => void;
+    label?: string | null;
```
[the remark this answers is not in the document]
marcoartaviaq: Yes, it should be false. Updated

src/features/advanced-content-insights/AdvancedContentInsightsToggle.tsx
```diff
@@ -0,0 +1,58 @@
...
+import './AdvancedContentInsightsToggle.scss';
+
+interface Props {
+    hasDescriptionInTooltip?: boolean;
+    isActive: boolean;
+    onChange?: (isEnabled: boolean) => void;
+    submitting: boolean;
+}
+
+const AdvancedContentInsightsToggle = ({
```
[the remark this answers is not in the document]
marcoartaviaq: This component is also used on a parent app. That's why I separated it into its own folder. We will also need to add more components for the feature in the future.

src/features/advanced-content-insights/AdvancedContentInsightsToggle.tsx
```diff
@@ -0,0 +1,58 @@
...
+// @ts-ignore flow import
+import Toggle from '../../components/toggle';
+import Tooltip from '../../components/tooltip';
+import './AdvancedContentInsightsToggle.scss';
+
+interface Props {
+    hasDescriptionInTooltip?: boolean;
+    isActive: boolean;
+    onChange?: (isEnabled: boolean) => void;
+    submitting: boolean;
```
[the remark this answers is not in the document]
marcoartaviaq: I changed it to be 'isDisabled' to make it clearer. We want to disable the toggle, for example, if the configuration is loading or if we are changing it (that was the submit part).

src/features/unified-share-modal/UnifiedShareModal.scss
```diff
@@ -312,6 +313,14 @@
         }
     }
 
+    .shared-link-checkgroup-row {
```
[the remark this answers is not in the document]
marcoartaviaq: Thanks. It's not longer used since we dropped the options.

src/features/unified-share-modal/UnifiedShareForm.js
```diff
@@ -744,6 +755,19 @@ class UnifiedShareForm extends React.Component<USFProps, State> {
                         />
                     )}
 
+                    {showContentInsightsToggle && (
+                        <>
+                            <div className="bdl-SharedLinkSection-separator" />
+                            <div className="shared-link-toggle-row">
```
tjuanitas, line 761: similar to the separator, I think we should avoid using a class that was intended for another component.

are there any specific styles we need from `shared-link-toggle-row`? it looks like it uses the top margin, any other styles we need?
maybe we could create a generic `bdl-UnifiedShareForm-row` or `bdl-UnifiedShareForm-toggle` that has only the bare styles.

tjuanitas (commented):
it's looking good!

src/features/advanced-content-insights/AdvancedContentInsightsToggle.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { FormattedMessage } from 'react-intl';
+import messages from './messages';
+import InfoBadge16 from '../../icon/fill/InfoBadge16';
+// @ts-ignore flow import
+import Toggle from '../../components/toggle';
+import Tooltip from '../../components/tooltip';
+import './AdvancedContentInsightsToggle.scss';
+
+interface Props {
+    hasDescriptionInTooltip?: boolean;
```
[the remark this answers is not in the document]
marcoartaviaq: Changed to the suggested name. I prefer the shorter version, but I'm not sure if it's clear that it would only be the description inside that tooltip though.

src/features/advanced-content-insights/flowTypes.js
```diff
@@ -0,0 +1,14 @@
...
+export type AdvancedContentInsightsType = {
+    /** Whether the advanced content insights FF is enabled */
+    enabled: boolean,
+    isActive: boolean,
+    onChange?: (isEnabled: boolean) => void,
+    ownerEId?: number,
+    userId?: number,
+};
+
+export type AdvancedContentInsightsUSProps = {
```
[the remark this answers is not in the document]
marcoartaviaq: Changed. And yes, you are right. That type is only used on USM.

src/elements/content-preview/ContentPreview.js
```diff
@@ -216,6 +219,10 @@ class ContentPreview extends React.PureComponent<Props, State> {
     };
 
     static defaultProps = {
+        advancedContentInsights: {
+            enabled: false,
+            config: {},
+        },
```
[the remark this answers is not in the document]
marcoartaviaq: Changed since we dropped all the config options and have only the isActive option

src/features/unified-share-modal/UnifiedShareModal.scss
```diff
@@ -255,7 +255,8 @@
         justify-content: space-between;
         margin-top: 10px;
 
-        .share-toggle-container {
+        .share-toggle-container,
+        .AdvancedContentInsightsToggle {
```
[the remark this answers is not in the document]
marcoartaviaq: These styles will be only used on USM.

src/features/content-insights/charts/bar/Bar.tsx
```diff
@@ -11,12 +11,20 @@ const DEFAULT_SIZE = 50;
 interface Props {
     color?: string;
     direction?: Direction;
-    onMouseEnter?: () => void;
+    label?: string | boolean;
```
tjuanitas, line 14: this is my fault, this should really only be a `string`. could we do the following?

1. let's remove the boolean, it's odd that it could be both a string and boolean
```suggestion
    label?: string;
```
.
2. change this from `label = false,` to just `label,` and let it be undefined
https://github.com/box/box-ui-elements/blob/d8315b512e89b7bc4539d93a6f5c988712b789ba/src/features/content-insights/charts/bar/Bar.tsx#L25
.
3. change this to `hasAxisLabel ? datum[labelAccessor] : ''`
https://github.com/box/box-ui-elements/blob/d8315b512e89b7bc4539d93a6f5c988712b789ba/src/features/content-insights/charts/bar/BarChart.tsx#L75

sorry bout that!

src/features/content-insights/charts/bar/__tests__/Bar.test.tsx
```diff
@@ -70,5 +84,48 @@ describe('features/content-insights/charts/bar/Bar', () => {
...
         });
+
+        test('should return the bar offset onMouseEnter', () => {
+            const elementOffset = {
+                bottom: 0,
+                height: 50,
+                left: 50,
+                right: 0,
+                top: 20,
+                toJSON: jest.fn(),
```
tjuanitas, line 95: what's `toJSON` from? is this needed?

marcoartaviaq: Hi @tjuanitas. I added because it was giving some typescript errors because the prop was missing. I updated it to make it consistent on how we are mocking the DOMRect on the repo

src/features/advanced-content-insights/flowTypes.js
```diff
@@ -0,0 +1,9 @@
+// @flow
+export type AdvancedContentInsightsType = {
+    /** Whether the advanced content insights FF is enabled */
+    enabled: boolean,
+    isActive: boolean,
+    onChange?: (isEnabled: boolean) => void,
+    ownerEId?: number,
+    userId?: number,
```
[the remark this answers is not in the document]
tjuanitas: @marcoartaviaq what are `ownerEId` and `userId` used for?

tjuanitas (approved):
had one last question but looks good to me! thanks for working through the feedback!

src/features/advanced-content-insights/flowTypes.js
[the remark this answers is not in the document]
marcoartaviaq: Sorry, I forgot to answer this. These two props are sent from a parent app that would be used on a children's app (box content preview). On that children's app, we use them if present.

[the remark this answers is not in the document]
tjuanitas: oh so these props are only being used for Preview? if another app was going to use Advanced Content Insights, would they use these props as well?

does USM use it in the config? https://github.com/box/box-ui-elements/blob/d8315b512e89b7bc4539d93a6f5c988712b789ba/src/features/unified-share-modal/flowTypes.js#L323

we might want to rework how we're passing some of the props instead of putting them in one object. since `enabled` is only used as a feature flip, then it would eventually be removed right? that means that `isActive` would ideally be the only thing prop needed to toggle Advanced Content Insights?

src/features/advanced-content-insights/AdvancedContentInsightsToggle.tsx
```diff
@@ -0,0 +1,55 @@
...
+import Toggle from '../../components/toggle';
+import Tooltip from '../../components/tooltip';
+
+import messages from './messages';
+
+import './AdvancedContentInsightsToggle.scss';
+
+interface Props {
+    hasTooltip?: boolean;
+    isActive?: boolean;
```
tjuanitas, line 15: could we rename `isActive` to `isChecked`? this would align the prop name with goal of renaming the `Toggle` prop to checked as well:
https://github.com/box/box-ui-elements/blob/8fd0931f8b336dc23bc24bfe31e0d7401f0a0aaf/src/components/toggle/Toggle.js#L13

src/features/advanced-content-insights/AdvancedContentInsightsToggle.tsx
```diff
@@ -0,0 +1,55 @@
...
+                        <InfoBadge16 height={14} width={14} />
+                    </div>
+                </Tooltip>
+            )}
+        </>
+    );
+
+    return (
+        <div className="AdvancedContentInsightsToggle">
+            <Toggle
```
tjuanitas, line 43: is this outer `div` necessary?

src/features/advanced-content-insights/AdvancedContentInsightsToggle.scss
```diff
@@ -0,0 +1,12 @@
+.AdvancedContentInsightsToggle {
+    display: flex;
```
tjuanitas, line 2: what's the `display: flex` doing in this case?

src/features/advanced-content-insights/AdvancedContentInsightsToggle.scss
```diff
@@ -0,0 +1,12 @@
...
+
+    .toggle-container {
+        margin-bottom: 0;
+    }
+}
```
tjuanitas, line 7: if we remove the outer div here:
https://github.com/box/box-ui-elements/blob/c2dfb7efa2db910eee4f3b2cac99b79f73f667f9/src/features/advanced-content-insights/AdvancedContentInsightsToggle.tsx#L42-L43

then we could pass in a className prop (`AdvancedContentInsightsToggle`) to `Toggle` and this could simplify to:
```
.AdvancedContentInsightsToggle {
    margin-bottom: 0;
}
```

src/features/advanced-content-insights/AdvancedContentInsightsToggle.scss
```diff
@@ -0,0 +1,12 @@
...
+
+.AdvancedContentInsightsToggle-icon {
+    display: inline;
```
tjuanitas, line 10: is it okay if we change to `inline-block`? I think margin's and padding's are less predictable (maybe don't work?) on inline elements

src/features/unified-share-modal/UnifiedShareForm.js
```diff
@@ -665,11 +666,12 @@ class UnifiedShareForm extends React.Component<USFProps, State> {
     render() {
         // Shared link section props
         const {
+            advancedContentInsightsConfig = {},
```
tjuanitas, line 669: so I don't think we should pass the value of feature flip down to these lower components and we only need the feature flip at the consuming component level. i.e. we could control whether ACI is hidden/shown based on the presence of `onAdvancedContentInsightsToggle`, then we won't need to pass `enabled` in the config.

my thinking is that the only new props we have are:
`isAdvancedContentInsightsChecked` <-- this will control the toggle on/off state
`onAdvancedContentInsightsToggle` <-- this is the onChange of the toggle and will control whether ACI is shown

we could change `showContentInsightsToggle`: https://github.com/box/box-ui-elements/blob/c2dfb7efa2db910eee4f3b2cac99b79f73f667f9/src/features/unified-share-modal/UnifiedShareForm.js#L712
to
```
const showContentInsightsToggle =
    onAdvancedContentInsightsToggle &&
    !isEmailLinkSectionExpanded &&
    !isInviteSectionExpanded &&
    !showCollaboratorList &&
    item?.type === ITEM_TYPE_FILE;
```

tjuanitas (changes requested):
okay sorry about this, I misunderstood how these props were being used by the components. If `enabled` is simply the value of a feature flip, then I don't think we should be using it in these components. We can control / determine the behavior of the components using the presence of other props such as `onChange` functions.

for example, instead of doing this:
```
if (advancedContentInsights.enabled) {
    onAdvancedContentInsightsToggle();
}
```
we could do
```
if (onAdvancedContentInsightsToggle) {
    onAdvancedContentInsightsToggle();
}
```
the consuming / parent component would be the only component that uses the feature flip and would do something like this:
```
<Component
    onAdvancedContentInsightsToggle={isFeatureFlipEnabled && onChange}
/>
```
this allows us to pass in less props and won't require clean up work to remove the feature flip.


also for easier reviewing in the future, I'd recommend splitting a PR like this into separate ones:

- one for the `Modal` changes
- one for adding the `AdvancedContentInsightsToggle` component
- one for the USM changes
- etc.

so it's easier to follow and understand the motivation behind the changes

src/elements/content-preview/ContentPreview.js
```diff
@@ -789,6 +808,7 @@ class ContentPreview extends React.PureComponent<Props, State> {
         }
 
         const previewOptions = {
+            advancedContentInsights,
```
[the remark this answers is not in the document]
marcoartaviaq: Sure @tjuanitas . Preview will use the enabled, isActive, ownerEId and userId. Since preview needs to know if AdvancedContentInsights is enabled (but is not using the onChange method), I'll prefer to keep the enabled prop.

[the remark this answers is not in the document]
tjuanitas: what about if `enabled` is false, then the parent component doesn't pass anything for the `advancedContentInsights` prop?

```
// in some parent component in the webapp

return (
  <ContentPreview
    advancedContentInsights={isFeatureFlipEnabled ? propsForPreview : undefined}
  />
);
```

my concern with `enabled` is that these lower components shouldn't need to know the state of a feature flip. after the feature has been released, how should these components behave? will we always need to pass `enabled` even if the feature is released and then removed?

src/features/advanced-content-insights/AdvancedContentInsightsToggle.tsx
tjuanitas: yeah I think we might be doing too much with the current `type AdvancedContentInsightsType`. I think it would be nice to have them as separate types with the USM flowTypes being:
```
// https://github.com/box/box-ui-elements/blob/master/src/features/unified-share-modal/flowTypes.js

type AdvancedContentInsightsTypes = {
    isAdvancedContentInsightsChecked?: boolean,
    onAdvancedContentInsightsToggle?: Function,
};
```
with these changes: https://github.com/box/box-ui-elements/pull/3116/files#r1046624678

src/elements/content-preview/ContentPreview.js
```diff
@@ -384,6 +388,15 @@ class ContentPreview extends React.PureComponent<Props, State> {
         if (haveExperiencesChanged && this.preview && this.preview.updateExperiences) {
             this.preview.updateExperiences(previewExperiences);
         }
+
+        if (advancedContentInsights) {
+            const { advancedContentInsights: prevContentInsightsOptions } = prevProps;
+            const haveContentInsightsChanged = prevContentInsightsOptions !== advancedContentInsights;
```
tjuanitas, line 394: I'm not entirely sure but would `haveContentInsightsChanged` always be true? since they're objects, it would compare references and so would always be different right?

tjuanitas (commented):
@marcoartaviaq that was my fault, I should've caught on with how these changes were connected earlier

src/features/advanced-content-insights/flowTypes.js
```diff
@@ -0,0 +1,6 @@
+// @flow
+export type AdvancedContentInsightsType = {
+    isActive: boolean,
+    ownerEId?: number,
+    userId?: number,
```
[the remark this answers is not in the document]
marcoartaviaq: Moved it to Content Preview props and make them non-optional

src/elements/content-preview/ContentPreview.js
marcoartaviaq: @tjuanitas you are right. I was comparing before only the isActive prop, but I changed it and didn't do the deep comparison...My bad! But I think it seems we have the same issue with haveExperiencesChanged in this file. Should I do something?

tjuanitas: @marcoartaviaq nice catch!

I'm discussing with @megansmith-box and getting her opinion on it. it sounds logical that if some object from `prevProps` is unchanged, then it would use the same object reference. although from testing it, that doesn't seem to be the case so the `haveExperiencesChanged` calculation could've been an oversight.

I would leave `haveExperiencesChanged` as-is for now just in case and we can address it in a separate PR

src/elements/content-preview/ContentPreview.js
```diff
@@ -384,6 +392,14 @@ class ContentPreview extends React.PureComponent<Props, State> {
         if (haveExperiencesChanged && this.preview && this.preview.updateExperiences) {
             this.preview.updateExperiences(previewExperiences);
         }
+
+        if (advancedContentInsights) {
+            const { advancedContentInsights: prevContentInsightsOptions } = prevProps;
+            const haveContentInsightsChanged = !isEqual(prevContentInsightsOptions, advancedContentInsights);
```
tjuanitas, line 398: these two lines could be moved up (and combined) to the top of this method:

```
        const { advancedContentInsights, previewExperiences, token } = this.props;
        const { advancedContentInsights: prevAdvancedContentInsights, previewExperiences: prevPreviewExperiences, token: prevToken } = prevProps;
        const { currentFileId } = this.state;
        const hasFileIdChanged = prevState.currentFileId !== currentFileId;
        const hasTokenChanged = prevToken !== token;
        const haveExperiencesChanged = prevPreviewExperiences !== previewExperiences;
        const haveContentInsightsChanged = !isEqual(prevAdvancedContentInsights, advancedContentInsights);
```

src/elements/content-preview/ContentPreview.js
```diff
@@ -384,6 +392,14 @@ class ContentPreview extends React.PureComponent<Props, State> {
...
+            if (haveContentInsightsChanged && this.preview && this.preview.updateContentInsightsOptions) {
+                this.preview.updateContentInsightsOptions(advancedContentInsights);
```
tjuanitas, line 400: do we want to call this if `advancedContentInsights` updates from the previous config to `null` or `undefined`?

marcoartaviaq: No @tjuanitas . It won't be a possible scenario at this moment so I would left it as-is.

tjuanitas (approved):
looks great! thank you for your patience and working through the feedback! much appreciated

src/features/advanced-content-insights/AdvancedContentInsightsToggle.tsx
```diff
@@ -0,0 +1,59 @@
...
+import Toggle from '../../components/toggle';
+import Tooltip from '../../components/tooltip';
+
+import messages from './messages';
+
+import './AdvancedContentInsightsToggle.scss';
+
+interface Props {
+    hasTooltip?: boolean;
+    isAdvancedContentInsightsChecked?: boolean;
```
tjuanitas, line 15: this prop could be `isChecked` since `AdvancedContentInsights` is part of the component name so it's not needed to include it in the prop name

note: the change would only be for this component. it still makes sense to keep it in UnifiedShareForm

tjuanitas (commented):
took another pass through since I saw the PR was still open

src/features/advanced-content-insights/AdvancedContentInsightsToggle.tsx
tjuanitas: thanks! one more place sorry:
https://github.com/box/box-ui-elements/blob/bf909deaf9d8bd7428b99327634dddd0b83a8834/src/features/unified-share-modal/UnifiedShareForm.js#L762
`isChecked={isAdvancedContentInsightsChecked}`

marcoartaviaq: @tjuanitas my bad! Updated

tjuanitas (approved):

src/elements/content-preview/__tests__/ContentPreview.test.js
```diff
@@ -1253,4 +1316,37 @@ describe('elements/content-preview/ContentPreview', () => {
             },
         );
     });
+
+    describe('getThumbnail()', () => {
```
bfoxx1906, line 1320: Should we add a test to ensure that the thumbnail has the properties that we are expecting. I notice we check for null but it might be helpful to test that the thumbnail is going to be returned properly.

src/features/unified-share-modal/UnifiedShareForm.js
```diff
@@ -706,6 +709,13 @@ class UnifiedShareForm extends React.Component<USFProps, State> {
             !showCollaboratorList &&
             allShareRestrictionWarning;
 
+        const showContentInsightsToggle =
+            onAdvancedContentInsightsToggle &&
```
bfoxx1906, line 713: Parts of this conditional are the exact same as the ones in the statement above. I'm guessing that combination of conditions has greater meaning in the context of the form than they do individually with regard to whether or not we want to show those elements. Do you think it would be a good idea to extract those common conditions out into their own function? It might make the code more readable and easy to understand.

src/elements/content-preview/__tests__/ContentPreview.test.js
marcoartaviaq: Hi, for the thumbnail We are testing that the viewer.getThumbnail is called. So for unit testing I think is the way to go. That viewer function will need to unit test if the thumbnail is being returned properly. Also, this function is going to be removed before going GA since we are going to switch the thumbnail provider (not gonna use PDFjs)

src/features/unified-share-modal/UnifiedShareForm.js
marcoartaviaq: Hey @bfoxx1906 , @tjuanitas . I added some logic regarding this comment @bfoxx1906 added. I agree with him that it would be better to make that condition more readable. PTAL and please fell free to share your thoughts

marcoartaviaq: Hi @tjuanitas and @bfoxx1906 . Bumping this to get final feedback for this change. Thanks

src/features/unified-share-modal/UnifiedShareForm.js
```diff
@@ -699,22 +702,21 @@ class UnifiedShareForm extends React.Component<USFProps, State> {
         const { sharedLinkTracking, sharedLinkEmailTracking } = trackingProps;
         const { isEmailLinkSectionExpanded, isInviteSectionExpanded, showCollaboratorList } = this.state;
 
-        // Only show the restriction warning on the main page of the USM where the email and share link option is available
-        const showShareRestrictionWarning =
-            !isEmailLinkSectionExpanded &&
-            !isInviteSectionExpanded &&
-            !showCollaboratorList &&
-            allShareRestrictionWarning;
+        const isMainUSMSectionShown = !isEmailLinkSectionExpanded && !isInviteSectionExpanded && !showCollaboratorList;
```
tjuanitas, line 705: not sure if these are better names but what about: `isInitialStateShown` or `hasExpandedSections` (requires inverting the boolean logic)

"main USM section" sounds a little ambiguous. i.e. what is the main section?

marcoartaviaq: Thanks @tjuanitas . Actually I like the hasExpandedSections better so I updated the code

tjuanitas (approved):

bfoxx1906 (approved):

PR #3118, opened 2022-11-02
feat(content-sidebar): add on fetch file callback function

MateuszMamczarz:
Content sidebar is displaying only when file is fetched successfully - we have onError callback function but we're missing one for success which makes it difficult to add proper loading indicator.


src/elements/content-sidebar/ContentSidebar.js
```diff
@@ -66,6 +66,7 @@ type Props = {
     messages?: StringMap,
     metadataSidebarProps: MetadataSidebarProps,
     onAnnotationSelect?: Function,
+    onFetchFileSuccess?: Function,
```
JChan106, line 69: I think we should start using the arrow function notation for function types to be more explicit. Dont have to change the other props, just the new one if possible

src/elements/content-sidebar/ContentSidebar.js
```diff
@@ -270,6 +271,8 @@ class ContentSidebar extends React.Component<Props, State> {
      * @return {void}
      */
     fetchFileSuccessCallback = (file: BoxItem): void => {
+        const { onFetchFileSuccess }: Props = this.props;
+        onFetchFileSuccess && onFetchFileSuccess();
```
JChan106, line 275: I think if possible its safer to double coalesce `!!onFetchFileSuccess` since it's not a boolean. I've ran into some weird cases when using the `&&` with a function call. Might also be more readable with a conditional instead

src/elements/content-sidebar/__tests__/ContentSidebar.test.js
```diff
@@ -156,18 +156,22 @@ describe('elements/content-sidebar/ContentSidebar', () => {
...
-            wrapper = getWrapper();
+            wrapper = getWrapper({
+                onFetchFileSuccess,
+            });
             instance = wrapper.instance();
             instance.setState = setState;
         });
 
-        test('should set the state with the file and view and then call fetchMetadata', () => {
+        test('should set the state with the file and view and, call onFetchFileSuccess and then call fetchMetadata', () => {
```
JChan106, line 174: Lets also add a case where onFetchFileSuccess isn't passed in

karelee7 (approved):
looks good to me!

PR #3122, opened 2022-11-03
feat(checkbox): allow stopPropagation on onClick events

ahorowitz123:
- Add `shouldStopPropagationOnClick` to `Checkbox` component to allow for `stopPropagation` to be called in an `onClick` event. This allows outside `onClick` events to not be called when the checkbox is clicked

### DIFF
--- a/src/components/checkbox/Checkbox.tsx
+++ b/src/components/checkbox/Checkbox.tsx
@@ -33,6 +33,8 @@ export interface CheckboxProps {
           };
     /** onChange - change callback function that takes the event as the argument */
     onChange?: (e: React.SyntheticEvent<HTMLInputElement, Event>) => string | number | boolean | void;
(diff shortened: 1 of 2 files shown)


joshmarnold (approved):

src/components/checkbox/Checkbox.tsx
```diff
@@ -33,6 +33,8 @@ export interface CheckboxProps {
           };
     /** onChange - change callback function that takes the event as the argument */
     onChange?: (e: React.SyntheticEvent<HTMLInputElement, Event>) => string | number | boolean | void;
+    /** Stops propagation when the checkbox container is clicked */
+    shouldStopPropagationOnClick?: boolean;
```
tjuanitas, line 37: what's the use case for this? it seems odd to control this through a prop.

PR #3123, opened 2022-11-03
fix(content-sidebar): reply operations interactions

kkuliczkowski-box:
Changed the logic of hiding replies in a thread to removing them (leaving just the last one) from the Feed instead of displaying last one based on the internal `isExpanded` state in `ActivityThread`.

This allows to meet the requirement of displaying multiple replies when the thread is not expanded.

### DIFF
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActivityThread.js
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActivityThread.js
@@ -27,6 +27,7 @@ type Props = {
     isAlwaysExpanded?: boolean,
     isRepliesLoading?: boolean,
     mentionSelectorContacts?: SelectorItems<>,
+    onHideReplies?: (lastReply: CommentType) => void,
     onReplyCreate?: (text: string) => void,
     onReplyDelete?: ({ id: string, permissions: BoxCommentPermission }) => void,
     onReplyEdit?: (
@@ -54,6 +55,7 @@ const ActivityThread = ({
     isAlwaysExpanded = false,
     isRepliesLoading,
     mentionSelectorContacts,
+    onHideReplies = noop,
     onReplyCreate,
     onReplyDelete = noop,
     onReplyEdit = noop,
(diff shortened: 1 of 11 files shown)


src/elements/content-sidebar/activity-feed/activity-feed/ActivityThread.js
```diff
@@ -62,16 +64,44 @@ const ActivityThread = ({
     repliesTotalCount = 0,
     translations,
 }: Props) => {
-    const [isExpanded, setIsExpanded] = React.useState(isAlwaysExpanded);
+    const { length: repliesLength } = replies;
```
JChan106, line 67: can replies length be 0?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityThread.js
```diff
@@ -62,16 +64,44 @@ const ActivityThread = ({
     repliesTotalCount = 0,
```
JChan106, line 64: What happens if this is defaulted to 0? Can we add tests for this edge case

src/elements/content-sidebar/activity-feed/activity-feed/ActivityThread.js
```diff
@@ -62,16 +64,44 @@ const ActivityThread = ({
...
+                    className="bcs-ActivityThread-toggle"
+                    onClick={onShowReplies}
+                    type="button"
+                    data-testid="activity-thread-button"
+                >
+                    <FormattedMessage values={{ repliesToLoadCount }} {...messages.showReplies} />
+                </PlainButton>
+            );
+        }
+        if (repliesTotalCount > 1 && repliesTotalCount === repliesLength) {
```
JChan106, line 91: So if replies total count is 1, we wont show any button? We should add a test for this too

src/elements/content-sidebar/activity-feed/activity-feed/ActivityThreadReplies.js
```diff
@@ -47,37 +43,21 @@ const ActivityThreadReplies = ({
 
     return (
         <div className="bcs-ActivityThreadReplies" data-testid="activity-thread-replies">
-            {!isExpanded ? (
+            {replies.map((reply: CommentType) => (
                 <Comment
-                    key={lastReply.type + lastReply.id}
-                    {...lastReply}
+                    key={reply.type + reply.id}
```
JChan106, line 48: can we use template literals for this? Seems more flexible and readable

src/elements/content-sidebar/activity-feed/activity-feed/__tests__/ActivityThread.test.js
```diff
@@ -35,16 +39,28 @@ describe('src/elements/content-sidebar/activity-feed/activity-feed/ActivityThrea
...
+        const onHideReplies = jest.fn();
+        const lastReply = cloneDeep(repliesMock[repliesMock.length - 1]);
+        const { getByText } = getWrapper({ onHideReplies });
+
+        const button = getByText(localize(messages.hideReplies.id));
+        expect(button).toBeVisible();
+        fireEvent.click(button);
+
+        expect(onHideReplies).toBeCalledWith(lastReply);
     });
```
JChan106, line 64: Lets add some edge cases mentioned above where the total is 0 or the number of replies is 0 and vise versa

src/elements/content-sidebar/activity-feed/activity-feed/ActivityThread.js
kkuliczkowski-box: That is correct, when there are no replies the length is 0. Theoretically `onHideRepliesHandler` shouldn't be called when `repliesLength = 0,` but I've added additional check in `onHideRepliesHandler`.

src/elements/content-sidebar/activity-feed/activity-feed/ActivityThread.js
kkuliczkowski-box: It's 0 when there are no replies. In that case the button is not rendered. I've updated tests to cover that scenario.

src/elements/content-sidebar/activity-feed/activity-feed/ActivityThread.js
kkuliczkowski-box: Correct. There was a test covering that, but I've made it now clearer to see.

src/elements/content-sidebar/activity-feed/activity-feed/__tests__/ActivityThread.test.js
kkuliczkowski-box: I've modified the test for not rendering button with multiple cases. This test actually covers the rendering of button and calling `onHideReplies` on click. I've also updated the description so that it's clearer.

JChan106 (approved):
lgtm

PR #3124, opened 2022-11-03
feat(menu): Allow stop propagation on menu click

ahorowitz123:
Add prop `shouldStopPropagationOnClick` to allow `stopPropagation` to be called on menu click

### DIFF
--- a/src/components/menu/Menu.tsx
+++ b/src/components/menu/Menu.tsx
@@ -51,6 +51,7 @@ interface MenuProps {
     setRef?: Function;
     /** shouldOutlineFocus - whether the focused menu item should have an outline */
     shouldOutlineFocus?: boolean;
+    shouldStopPropagationOnClick?: boolean;
     /** tabIndex - indicator of whether the menu is focusable */
     tabIndex?: number;
 }
@@ -202,6 +203,12 @@ class Menu extends React.Component<MenuProps> {
     };
 
     handleClick = (event: React.MouseEvent<HTMLUListElement, MouseEvent>) => {
+        const { shouldStopPropagationOnClick } = this.props;
+
+        if (shouldStopPropagationOnClick) {
+            event.stopPropagation();
+        }
+
         const { menuItemEl }: { menuItemEl?: HTMLElement | null } =
             event.target instanceof Node ? this.getMenuItemElFromEventTarget(event.target) : {};
 
--- a/src/components/menu/Menu.tsx
+++ b/src/components/menu/Menu.tsx
@@ -51,6 +51,7 @@ interface MenuProps {
     setRef?: Function;
     /** shouldOutlineFocus - whether the focused menu item should have an outline */
     shouldOutlineFocus?: boolean;
+    /** shouldStopPropagationOnClick - whether we should stop propagation on menu click */
     shouldStopPropagationOnClick?: boolean;
     /** tabIndex - indicator of whether the menu is focusable */
     tabIndex?: number;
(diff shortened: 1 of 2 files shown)


tjuanitas (commented):
what is the use case for this? I left comments on a similar PR: https://github.com/box/box-ui-elements/pull/3122

is this intended to be added to more components?

ahorowitz123:
> what is the use case for this? I left comments on a similar PR: #3122
> 
> is this intended to be added to more components?

The use case for this is in the new grid view implementation we have an outer div that has an onclick function which navigates to a different page. Therefore, when every menu item was being clicked, it was triggering the navigation to a different page.



> what is the use case for this? I left comments on a similar PR: #3122
> 
> is this intended to be added to more components?

The use case here is that we have an outer div which navigates to a different page on click. Therefore, when a menu item is clicked, we navigate to another page.

Do you think it makes more sense to just take in an `onClick` prop and allow the callers to define whether they want to use `stopPropagation` inside of their own `onClick` functions? Strangely, I notice that the `Menu` component here does have an `onClick` prop but its never called

src/components/menu/Menu.tsx
```diff
@@ -202,6 +204,12 @@ class Menu extends React.Component<MenuProps> {
     };
 
     handleClick = (event: React.MouseEvent<HTMLUListElement, MouseEvent>) => {
+        const { shouldStopPropagationOnClick } = this.props;
+
+        if (shouldStopPropagationOnClick) {
+            event.stopPropagation();
```
mickr, line 210: Can this handled in onClose that gets passed into the props and called on L202?

ahorowitz123: From my understanding and testing it has to be inside of an onClick function or else it won't stop bubbled up onClick functions from being called

ahorowitz123: Based on what @tjuanitas said here and in my other PR I think it might make most sense to just wrap the components in divs/spans in EUA and handle it there

src/components/menu/Menu.tsx
```diff
@@ -51,6 +51,8 @@ interface MenuProps {
     setRef?: Function;
     /** shouldOutlineFocus - whether the focused menu item should have an outline */
     shouldOutlineFocus?: boolean;
+    /** shouldStopPropagationOnClick - whether we should stop propagation on menu click */
+    shouldStopPropagationOnClick?: boolean;
```
tjuanitas, line 55: > The use case here is that we have an outer div which navigates to a different page on click. Therefore, when a menu item is clicked, we navigate to another page.
> 
> Do you think it makes more sense to just take in an `onClick` prop and allow the callers to define whether they want to use `stopPropagation` inside of their own `onClick` functions? Strangely, I notice that the `Menu` component here does have an `onClick` prop but its never called

That seems like it's more of an issue with the outer div than with these components. If more components (overlays, tooltips, banners, etc.) are added "above" this outer div, would all of those need this new prop as well?

is it possible to add a check in the outer div's onClick that verifies the click event is coming from the currentTarget instead of elements in which the event is propagating to it?

ahorowitz123: That was our initial solution but there are many elements inside the div where do want the outer div click event to trigger. There's only 2 elements where we don't so it made more sense to just single those out instead

src/components/menu/Menu.tsx
mickr: I think that would be best case and leave it up to the consumer to stopPropagation

PR #3125, opened 2022-11-03
fix(menu): fix long title in menu header

Machoper:
## Summary of change
Apply new `bdl-lineClamp` mixin to menu header title.

## Demo
More options menu in full screen
### Before:
before

### After:
after

### DIFF
--- a/src/components/menu/MenuHeader.scss
+++ b/src/components/menu/MenuHeader.scss
@@ -23,11 +23,9 @@
     @include bdl-lineClamp;
 
     margin: 0;
-    overflow: hidden;
     color: $bdl-gray;
     font-weight: bold;
     font-size: 16px;
-    text-overflow: ellipsis;
 }
 
 .bdl-MenuHeader-subtitle {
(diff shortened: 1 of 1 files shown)


src/components/menu/MenuHeader.scss
```diff
@@ -20,6 +20,8 @@
 }
 
 .bdl-MenuHeader-title {
+    @include bdl-lineClamp;
```
tjuanitas, line 23: is there a conflict with using `line-clamp` and `text-overflow: ellipsis` together? does it just add ellipsis to the last line of the line clamp?

Machoper: There's no conflict between the two but `text-overflow` doesn't have any effect when being used together with `bdl-lineClamp` either regardless of the value set. So I removed `text-overflow: ellipsis` as well as `overflow: hidden`, which is already assigned in `bdl-lineClamp`.

`bdl-lineClamp` automatically adds an ellipsis to the last line of text.

tjuanitas (approved):

PR #3126, opened 2022-11-03
feat(snapshots): Enabled visual-tests and fixed broken image snapshots

greg-in-a-box:


src/components/draggable-list/DraggableList.scss
```diff
@@ -13,4 +13,8 @@
     .draggable-list-handle {
         display: flex;
     }
+
+    .checkbox-container {
+        margin: inherit;
+    }
```
tjuanitas, line 19: what's the reason for this?

tjuanitas: this would affect components in parent apps that rely on the existing margin

tjuanitas (commented):
trial, beta, and upgrade badges don't have snapshot previews?

PR #3129, opened 2022-11-07
feat(threaded-replies): fix styling for resolve

MateuszMamczarz:
Fix visual issues:
- adjust margin for `resolved` pill
- change fill for `X` icon to be consistent with other icons
Screen Shot 2022-11-07 at 18 22 48

### DIFF
--- a/src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityMenu.scss
+++ b/src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityMenu.scss
@@ -1,0 +1,5 @@
+@import '../../../common/variables';
+
+.bcs-AnnotationActivityMenu-unresolveAnnotation path {
+    fill: $bdl-gray;
+}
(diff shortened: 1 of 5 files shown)


wpiesiak (approved):

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityMenu.scss
```diff
@@ -0,0 +1,5 @@
+@import '../../../common/variables';
+
+.bcs-AnnotationActivityMenu-unresolveAnnotation path {
+    fill: $bdl-gray;
```
jstoffan, line 4: Why is this needed? It looks like all the other icons use `bdl-gray-50` for their paths.

jstoffan:
@MateuszMamczarz, it looks like the trash icons are different between the comment and annotation menus. Can we unify the icons and make them all the same color rather than as one-offs?

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityMenu.scss
MateuszMamczarz: All icons for comments are using `bdl-gray`, for annotations trash was different, but now it's the same as for comments. With this change this is how the whole menu looks like:
Screen Shot 2022-11-09 at 22 38 32

jstoffan (approved):

PR #3132, opened 2022-11-08
feat(annotations): hide AnnotationActivityLink based on hasVersions prop

wpiesiak:
show / hide AnnotationActivityLink based on hasVerions ActivitySidebar property

### DIFF
--- a/src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityLink.scss
+++ b/src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityLink.scss
@@ -6,13 +6,6 @@
     color: $bdl-box-blue;
     font-weight: bold;
 
-    &::before {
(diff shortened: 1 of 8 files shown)


src/elements/content-sidebar/activity-feed/annotations/AnnotationActivityLink.scss
```diff
@@ -6,6 +6,13 @@
     color: $bdl-box-blue;
     font-weight: bold;
 
+    &::before {
```
jstoffan, line 9: I think this change makes sense, but can we move it back to the `AnnotationActivity` stylesheet? In theory, the AnnotationActivityLink component could be reused elsewhere and may have a different layout/experience.

jstoffan (approved):

PR #3134, opened 2022-11-09
feat(content-sidebar): added support for active comment reply

kkuliczkowski-box:
Added support for handling links with active item that is a reply. This is to support mentioning users directly in a reply. The link that user will receive in notification email will have only type and id of the feed item. This change covers that scenario by doing additional request for that replie's details (in order to get id of its parent). Logic can be seen in the diagram:

handling-active-comment-reply

**Please note that this is a temporary solution done based on time constraints presented by Product. Final solution would be to change the link in notification email to also include `parentId`, thus making this additional request mentioned above not needed.**

### DIFF
--- a/src/api/Feed.js
+++ b/src/api/Feed.js
@@ -463,6 +463,51 @@ class Feed extends Base {
     }
 
     /**
+     * Fetches a comment for a file
+     *
+     * @param {BoxItem} file - The file to which the comment belongs to
+     * @param {string} commentId - comment id
+     * @param {Function} successCallback
+     * @param {ErrorCallback} errorCallback
+     * @return {Promise} - the file comments
+     */
+    fetchThreadedComment(
+        file: BoxItem,
+        commentId: string,
+        successCallback: (comment: Comment) => void,
+        errorCallback: ErrorCallback,
+    ): Promise<?Comment> {
+        const { id, permissions } = file;
+        if (!id || !permissions) {
+            throw getBadItemError();
+        }
+
+        this.threadedCommentsAPI = new ThreadedCommentsAPI(this.options);
+        return new Promise(resolve => {
+            this.threadedCommentsAPI.getComment({
+                commentId,
+                errorCallback,
+                fileId: id,
+                permissions,
+                successCallback: this.fetchThreadedCommentSuccessCallback.bind(this, resolve, successCallback),
+            });
+        });
+    }
+
+    /**
+     * Callback for successful fetch of a comment
+     *
+     * @param {Function} resolve - resolve function
+     * @param {Function} successCallback - success callback
+     * @param {Comment} comment - comment data
+     * @return {void}
+     */
+    fetchThreadedCommentSuccessCallback = (resolve: Function, successCallback: Function, comment: Comment): void => {
+        successCallback(comment);
+        resolve();
+    };
+
+    /**
      * Fetches the comments with replies for a file
      *
      * @param {Object} permissions - the file permissions
--- a/src/api/ThreadedComments.js
+++ b/src/api/ThreadedComments.js
@@ -229,6 +229,45 @@ class ThreadedComments extends MarkerBasedApi {
     }
 
     /**
+     * API for fetching comment
+     *
+     * @param {string} commentId - comment id
+     * @param {string} fileId - the file id
+     * @param {BoxItemPermission} permissions - the permissions for the file
+     * @param {Function} successCallback - the success callback
+     * @param {Function} errorCallback - the error callback
+     * @returns {void}
+     */
+    getComment({
+        commentId,
+        errorCallback,
+        fileId,
+        permissions,
+        successCallback,
+    }: {
+        commentId: string,
+        errorCallback: (e: ElementsXhrError, code: string) => void,
+        fileId: string,
+        permissions: BoxItemPermission,
+        successCallback: (comment: Comment) => void,
+    }): void {
+        this.errorCode = ERROR_CODE_FETCH_COMMENTS;
+        try {
+            this.checkApiCallValidity(PERMISSION_CAN_COMMENT, permissions, fileId);
+        } catch (e) {
+            errorCallback(e, this.errorCode);
+            return;
+        }
+
+        this.get({
+            id: fileId,
+            errorCallback,
+            successCallback,
+            url: this.getUrlForId(commentId),
+        });
+    }
+
+    /**
      * API for fetching comments
      *
      * @param {string} fileId - the file id
--- a/src/api/ThreadedComments.js
+++ b/src/api/ThreadedComments.js
@@ -12,6 +12,7 @@ import {
     ERROR_CODE_CREATE_COMMENT,
     ERROR_CODE_UPDATE_COMMENT,
     ERROR_CODE_DELETE_COMMENT,
+    ERROR_CODE_FETCH_COMMENT,
     ERROR_CODE_FETCH_COMMENTS,
     PERMISSION_CAN_RESOLVE,
     ERROR_CODE_FETCH_REPLIES,
@@ -251,7 +252,7 @@ class ThreadedComments extends MarkerBasedApi {
         permissions: BoxItemPermission,
         successCallback: (comment: Comment) => void,
     }): void {
-        this.errorCode = ERROR_CODE_FETCH_COMMENTS;
+        this.errorCode = ERROR_CODE_FETCH_COMMENT;
         try {
             this.checkApiCallValidity(PERMISSION_CAN_COMMENT, permissions, fileId);
         } catch (e) {
--- a/src/elements/content-sidebar/ActivitySidebar.js
+++ b/src/elements/content-sidebar/ActivitySidebar.js
@@ -50,11 +50,13 @@ import type {
     AnnotationPermission,
     BoxCommentPermission,
     Comment,
+    CommentFeedItem,
     CommentFeedItemType,
+    FocusableFeedItem,
     FocusableFeedItemType,
-    FeedItem,
     FeedItems,
     FeedItemStatus,
+    FeedItemType,
 } from '../../common/types/feed';
 import type { ErrorContextProps, ElementsXhrError } from '../../common/types/api';
 import type { WithLoggerProps } from '../../common/types/logging';
@@ -695,35 +697,70 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
             hasTasks: shouldShowTasks,
             hasVersions: shouldShowVersions,
         } = this.props;
+        const { activeFeedEntryId, activeFeedEntryType, feedItems: feedItemsState } = this.state;
         const shouldShowAppActivity = isFeatureEnabled(features, 'activityFeed.appActivity.enabled');
         const shouldShowAnnotations = isFeatureEnabled(features, 'activityFeed.annotations.enabled');
+        const shouldFetchRepliesOnLoad =
+            shouldShowReplies && !feedItemsState && activeFeedEntryId && activeFeedEntryType === FEED_ITEM_TYPE_COMMENT;
 
         api.getFeedAPI(shouldDestroy).feedItems(
             file,
             shouldRefreshCache,
-            this.fetchFeedItemsSuccessCallback,
+            shouldFetchRepliesOnLoad ? this.fetchRepliesForFeedItems : this.fetchFeedItemsSuccessCallback,
             this.fetchFeedItemsErrorCallback,
             this.errorCallback,
             { shouldShowAnnotations, shouldShowAppActivity, shouldShowReplies, shouldShowTasks, shouldShowVersions },
         );
     }
 
+    fetchRepliesForFeedItems = (feedItems: FeedItems) => {
+        const { activeFeedEntryId } = this.props;
+
+        this.getActiveFeedEntryData(feedItems)
+            .then(({ id, type }) => {
+                if (
+                    !id ||
+                    !type ||
+                    !activeFeedEntryId ||
+                    !this.isItemTypeComment(type) ||
+                    this.isActiveEntryInFeed(feedItems, activeFeedEntryId)
+                ) {
+                    this.fetchFeedItemsSuccessCallback(feedItems, id, type);
+                    return;
+                }
+
+                // $FlowFixMe: type has been checked to be either 'comment' or 'annotation' by this.isItemTypeComment(type)
+                this.getFeedItemsWithReplies(feedItems, id, type)
+                    .then(updatedItems => {
+                        this.fetchFeedItemsSuccessCallback(updatedItems, id, type);
+                    })
+                    .catch((error: ElementsXhrError) => {
+                        this.fetchFeedItemsErrorCallback(feedItems, [error]);
+                    });
+            })
+            .catch((error: ElementsXhrError) => {
+                this.fetchFeedItemsErrorCallback(feedItems, [error]);
+            });
+    };
+
     /**
      * Handles a successful feed API fetch
      *
      * @private
      * @param {Array} feedItems - the feed items
+     * @param {string} activeFeedEntryId - active feed entry id
+     * @param {FocusableFeedItemType} activeFeedEntryType - active feed entry type
      * @return {void}
      */
-    fetchFeedItemsSuccessCallback = (feedItems: FeedItems): void => {
+    fetchFeedItemsSuccessCallback = (
+        feedItems: FeedItems,
+        activeFeedEntryId?: string,
+        activeFeedEntryType?: FocusableFeedItemType,
+    ): void => {
         const {
-            activeFeedEntryId,
-            activeFeedEntryType,
             file: { id: fileId },
-            hasReplies: shouldShowReplies,
             logger,
         } = this.props;
-        const { feedItems: feedItemsState } = this.state;
 
         mark(MARK_NAME_DATA_READY);
 
@@ -738,34 +775,12 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
             );
         }
 
-        const baseChanges = { feedItems, activityFeedError: undefined };
-
-        // If activeFeedEntryId is not set or
-        // feed items are already in state (we want to load the active entry (reply) data only if the feed items are loaded for the first time) or
-        // replies should not be loaded at all or
-        // active feed entry type is other than comment,
-        // set feedItems state without any additional actions
-        if (
-            !activeFeedEntryId ||
-            feedItemsState ||
-            !shouldShowReplies ||
-            activeFeedEntryType !== FEED_ITEM_TYPE_COMMENT
-        ) {
-            this.setState({ ...baseChanges });
-            return;
-        }
-
-        this.getActiveFeedEntryData(feedItems)
-            .then(({ id, type }) => {
-                this.setState({
-                    ...baseChanges,
-                    activeFeedEntryId: id,
-                    activeFeedEntryType: type,
-                });
-            })
-            .catch((error: ElementsXhrError) => {
-                this.fetchFeedItemsErrorCallback(feedItems, [error]);
-            });
+        this.setState({
+            activeFeedEntryId: activeFeedEntryId || this.state.activeFeedEntryId,
+            activeFeedEntryType: activeFeedEntryType || this.state.activeFeedEntryType,
+            feedItems,
+            activityFeedError: undefined,
+        });
     };
 
     /**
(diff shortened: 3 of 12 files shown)


src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
```diff
@@ -110,23 +125,38 @@ class ActivityFeed extends React.Component<Props, State> {
             currentUser: prevCurrentUser,
             feedItems: prevFeedItems,
         } = prevProps;
-        const { feedItems: currFeedItems, activeFeedEntryId } = this.props;
-        const { isInputOpen: prevIsInputOpen } = prevState;
-        const { isInputOpen: currIsInputOpen } = this.state;
+        const { feedItems: currFeedItems, activeFeedEntryId, onShowReplies = noop } = this.props;
+        const { activeEntry, isInputOpen: prevIsInputOpen } = prevState;
+        const { activeEntry: prevActiveEntry, isInputOpen: currIsInputOpen } = this.state;
```
jstoffan, line 130: Is this inverted? It looks like this should be `currActiveEntry` and the line above should be `prevActiveEntry`.

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
```diff
@@ -110,23 +125,38 @@ class ActivityFeed extends React.Component<Props, State> {
             currentUser: prevCurrentUser,
             feedItems: prevFeedItems,
         } = prevProps;
-        const { feedItems: currFeedItems, activeFeedEntryId } = this.props;
-        const { isInputOpen: prevIsInputOpen } = prevState;
-        const { isInputOpen: currIsInputOpen } = this.state;
+        const { feedItems: currFeedItems, activeFeedEntryId, onShowReplies = noop } = this.props;
```
jstoffan, line 128: Let's use defaultProps for the default `noop` value here.

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
```diff
@@ -110,23 +125,38 @@ class ActivityFeed extends React.Component<Props, State> {
...
         const didLoadFeedItems = prevFeedItems === undefined && currFeedItems !== undefined;
         const hasInputOpened = currIsInputOpen !== prevIsInputOpen;
         const hasActiveFeedEntryIdChanged = activeFeedEntryId !== prevActiveFeedEntryId;
+        const hasActiveEntryChanged = activeEntry?.id !== prevActiveEntry?.id;
 
         if ((hasLoaded || hasMoreItems || didLoadFeedItems || hasInputOpened) && activeFeedEntryId === undefined) {
             this.resetFeedScroll();
         }
 
-        if (didLoadFeedItems || hasActiveFeedEntryIdChanged) {
```
jstoffan, line 127: Do we still need to check `didLoadFeedItems`? If not, won't this fail in most circumstances?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
```diff
@@ -88,12 +99,16 @@ type Props = {
...
 
 type State = {
+    activeEntry?: FeedItem | null,
     isInputOpen: boolean,
+    isReplyDataLoading: boolean,
 };
 
 class ActivityFeed extends React.Component<Props, State> {
     state = {
+        activeEntry: undefined,
```
jstoffan, line 109: We should avoid storing derived state from props in this way, since it can lead to synchronization issues. Let's retrieve the entry as needed using the value from `props.activeFeedEntryId`.

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
```diff
@@ -160,6 +190,61 @@ class ActivityFeed extends React.Component<Props, State> {
         return feedItems.length === 0 || (feedItems.length === 1 && feedItems[0].type === ItemTypes.fileVersion);
     };
 
+    getActiveEntry = (callback: (activeEntry: FeedItem | null, repliesNeedToBeLoaded?: boolean) => void) => {
```
jstoffan, line 193: Can we use a promise or async/await instead of introducing callbacks?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
```diff
@@ -160,6 +190,61 @@ class ActivityFeed extends React.Component<Props, State> {
...
+        if (!activeFeedEntryId || !Array.isArray(feedItems)) {
+            return callback(null);
+        }
+
+        const firstLevelItem = feedItems.find(
+            ({ id, type }) => id === activeFeedEntryId && type === activeFeedEntryType,
+        );
+        // If the active entry is a first level Feed item, return it
+        if (firstLevelItem) {
+            return callback(firstLevelItem);
```
jstoffan, line 205: We generally want to avoid multiple returns, as it makes a function very difficult to debug and reason about. This function is also _very_ complicated and probably needs to be decomposed.

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
```diff
@@ -160,6 +190,61 @@ class ActivityFeed extends React.Component<Props, State> {
...
+            return callback(null);
+        }
+
+        const firstLevelItemWithActiveReply = feedItems.find(item => {
+            if (
+                (item.type !== FEED_ITEM_TYPE_ANNOTATION && item.type !== FEED_ITEM_TYPE_COMMENT) ||
+                !item.replies ||
+                !Array.isArray(item.replies)
+            ) {
+                return false;
```
jstoffan, line 219: Should this return undefined?

src/api/ThreadedComments.js
```diff
@@ -228,6 +228,45 @@ class ThreadedComments extends MarkerBasedApi {
...
+        permissions,
+        successCallback,
+    }: {
+        commentId: string,
+        errorCallback: (e: ElementsXhrError, code: string) => void,
+        fileId: string,
+        permissions: BoxItemPermission,
+        successCallback: (comment: Comment) => void,
+    }): void {
+        this.errorCode = ERROR_CODE_FETCH_COMMENTS;
```
jstoffan, line 254: Should this use a new error code?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
```diff
@@ -160,6 +190,61 @@ class ActivityFeed extends React.Component<Props, State> {
...
+            }
+            return item.replies.some(({ id }) => id === activeFeedEntryId);
+        });
+        // If the active entry is within present replies of any first level Feed items (if it's a comment or annotation),
+        // return that first level Feed item
+        if (firstLevelItemWithActiveReply) {
+            return callback(firstLevelItemWithActiveReply);
+        }
+
+        if (!getComment) {
```
jstoffan, line 229: Is this possible with the current implementation? It looks like the method is always defined.

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
```diff
@@ -43,6 +49,11 @@ type Props = {
     file: BoxItem,
     getApproverWithQuery?: Function,
     getAvatarUrl: GetAvatarUrlCallback,
+    getComment?: (
```
jstoffan, line 52: Should we handle this case in `ActivitySidebar.fetchFeedItems` rather than inside ActivityFeed?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
```diff
@@ -298,7 +378,7 @@ class ActivityFeed extends React.Component<Props, State> {
                     }}
                     className="bcs-activity-feed-items-container"
                 >
-                    {isLoading && (
+                    {(isLoading || isReplyDataLoading) && (
```
jstoffan, line 381: The combination of loading state here indicates again that this should probably occur when the feed items are being fetched rather than separately. If not, though, should this move into the `hasLoaded` method?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
```diff
@@ -313,8 +393,8 @@ class ActivityFeed extends React.Component<Props, State> {
                     {!isEmpty && !isLoading && (
                         <ActiveState
                             {...activityFeedError}
-                            activeFeedEntryId={activeFeedEntryId}
-                            activeFeedEntryType={activeFeedEntryType}
+                            activeFeedEntryId={activeEntry?.id}
```
jstoffan, line 396: Let's revert these changes given that we want to avoid derived state.

jstoffan (changes requested):

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
```diff
@@ -160,6 +190,61 @@ class ActivityFeed extends React.Component<Props, State> {
...
+            activeFeedEntryId,
+            ({ parent }) => {
+                this.setState({ isReplyDataLoading: false });
+                const parentItem = feedItems.find(({ id }) => id === parent?.id) || null;
+                callback(parentItem, true);
+            },
+            () => {
+                this.setState({ isReplyDataLoading: false });
+                callback(null);
+            },
```
tjuanitas, line 244: where is `getComment` initially from passed from? i'm curious how the order of these args are determined.

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
tjuanitas, line 396: is there a reason we call the state `activeEntry` instead of `activeFeedEntry`?

kkuliczkowski-box:
I've changed the approach and followed @jstoffan's advice and moved the whole logic of fetching the reply data to `ActivitySidebar`. Thanks to that we don't need this additional "loading" step and the reply data is being loaded before the feed items are displayed. Because all of the changes done in `ActivityFeed` have been reverted, I haven't responded to most of the comments.

I'm struggling with Flow check at the moment and will continue working on it.

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
kkuliczkowski-box: I think It can return `false` as this is the return statement of the `find` callback function. I've copied that piece of logic into `ActivitySidebar`.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -152,20 +158,31 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
-        const { logger } = this.props;
+        const { activeFeedEntryId, activeFeedEntryType, logger } = this.props;
 
         mark(MARK_NAME_DATA_LOADING);
 
         logger.onReadyMetric({
             endMarkName: MARK_NAME_JS_READY,
         });
-        this.state = {};
+        this.state = { activeFeedEntryId, activeFeedEntryType, approverSelectorContacts: [] };
```
jstoffan, line 168: We should almost always avoid duplicating our props or any prop-derived value in state, as it can easily lead to synchronization issues. Let's update the logic to use the values directly from props.

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
kkuliczkowski-box: That makes a lot of sense. I've moved the logic for fetching reply data to `ActivitySidebar` but I've placed it in `fetchFeedItemsSuccessCallback` as I didn't want to change the original purpose of `fetchFeedItems. It seemed better to put it in `fetchFeedItemsSuccessCallback` as I'm taking actions based on the success of the `fetchFeedItems`.

src/api/Feed.js
```diff
@@ -462,6 +462,51 @@ class Feed extends Base {
...
+
+    /**
+     * Callback for successful fetch of a comment
+     *
+     * @param {Function} resolve - resolve function
+     * @param {Function} successCallback - success callback
+     * @param {Comment} comment - comment data
+     * @return {void}
+     */
+    fetchThreadedCommentSuccessCallback = (resolve: Function, successCallback: Function, comment: Comment): void => {
```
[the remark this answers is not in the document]
kkuliczkowski-box: I needed to keep `comment` as the last param because I'm binding this method in `fetchThreadedComment` method.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -717,7 +738,34 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+            activeFeedEntryType !== FEED_ITEM_TYPE_COMMENT
+        ) {
+            this.setState({ ...baseChanges });
+            return;
+        }
+
+        this.getActiveFeedEntryData(feedItems)
+            .then(({ id, type }) => {
+                this.setState({
+                    ...baseChanges,
```
jstoffan, line 761: I'm a bit concerned about race conditions here. What happens if `fetchFeedItems` is called again while this promise is still in progress? Would we wind up with stale feed items? It seems like we'd want to handle this conditional promise chain prior to the success callback being invoked?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -717,7 +738,34 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
             );
         }
 
-        this.setState({ feedItems, activityFeedError: undefined });
+        const baseChanges = { feedItems, activityFeedError: undefined };
+
+        // If activeFeedEntryId is not set or
```
jstoffan, line 743: This comment is largely describing the logic below. We should use comments to explain "why" something is written if it's not obvious in the code.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -827,6 +875,149 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+                if (
+                    type === FEED_ITEM_TYPE_ANNOTATION ||
+                    type === FEED_ITEM_TYPE_COMMENT ||
+                    type === FEED_ITEM_TYPE_TASK
+                ) {
+                    resolve({ id, type });
+                    return;
+                }
+                resolve({ id: activeFeedEntryId, type: activeFeedEntryType });
+                return;
```
jstoffan, line 958: We should avoid multiple returns, as it makes logic very difficult to debug and reason about. We're doing quite a bit of duplicate work in this function and it would be great to refactor it for simplicity.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -827,6 +875,149 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+
+    /**
+     * Fetches comment data
+     *
+     * @param {string} id - id of the comment
+     * @param {Function} onSuccess - the success callback
+     * @param {ErrorCallback} onError -the error callback
+     * @return {void}
+     */
+    getComment = (
```
jstoffan, line 1011: Minor, but is this needed? Can we call `fetchThreadedComment` directly, instead?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
```diff
@@ -127,6 +140,20 @@ class ActivityFeed extends React.Component<Props, State> {
...
+            onShowReplies &&
+            activeFeedEntryId &&
+            activeFeedEntryType &&
+            (activeFeedEntryType === FEED_ITEM_TYPE_COMMENT || activeFeedEntryType === FEED_ITEM_TYPE_ANNOTATION) &&
+            hasLoaded &&
+            activeFeedEntryReplyId &&
+            currFeedItems &&
+            !isItemInFeed(currFeedItems, activeFeedEntryReplyId)
+        ) {
+            onShowReplies(activeFeedEntryId, activeFeedEntryType);
```
jstoffan, line 155: This seems very indirect. Can we load the replies if we detect that they're needed within the `fetchFeedItems` chain?

src/elements/content-sidebar/ActivitySidebar.js
kkuliczkowski-box: I understand your point but in this case I don't see any alternative really. The `activeFeedEntryId` and `activeFeedEntryType` props represent the values of url params. Previously they were simply passed to `ActivityFeed`, where logic for displaying "selected state" lives.

Now with the possibility of a reply id being set as `activeFeedEntryId` we need to check if it is somewhere within feed items. If it turns out that it's a reply we need to pass its parent as `activeFeedEntryId` as the parent needs to be "selected/highlighted". In that case I'm updating values in state and pass them to `ActivityFeed`. If no action is taken. Moreover if the reply id is not found within feed items, we need to fetch for that reply's data in order to find its parent and then set its id in state. 

Ultimately this is a temporary solution until we get the `partentId` as an url param. With that we would not need to store the active item id in state as it would simply be either the `itemId` or the `parentId`.

src/elements/content-sidebar/ActivitySidebar.js
kkuliczkowski-box: These additional asynchronous actions (fetching reply data and its parent replies) will take place only once when feed items are not yet loaded. From what I can tell, during that time fetchFeedItems will not be called as the whole Activity Sidebar is in the "loading state".

I've also moved that additional actions logic outside of `fetchFeedItemsSuccessCallback`.

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
kkuliczkowski-box: Yes, that was a wrong approach. I've reverted the `ActivityFeed` completely and fetching for replies is now a part of the `fetchFeedItems` chain during the first load of `ActivitySidebar`.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -530,11 +536,32 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
 
+        const feedAPI = api.getFeedAPI(false);
         feedAPI.file = file;
+
+        // Detect if replies are being hidden and activeFeedEntryId belongs to a reply
+        // that is in currently being updated parent, in order to disable active item
+        if (
+            activeFeedEntryId &&
+            replies.length === 1 &&
+            feedItems.find(
```
jstoffan, line 554: Should this be `.some`? It looks we do a form of this check in quite a few places. Would it be meaningful to move it to a utility function?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -530,11 +536,32 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+        }
 
+        const feedAPI = api.getFeedAPI(false);
         feedAPI.file = file;
+
+        // Detect if replies are being hidden and activeFeedEntryId belongs to a reply
+        // that is in currently being updated parent, in order to disable active item
+        if (
+            activeFeedEntryId &&
+            replies.length === 1 &&
```
jstoffan, line 553: Would this cause any issues with threads that are exactly one reply long?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -671,26 +698,68 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+                    !this.isItemTypeComment(type)
+                ) {
+                    this.fetchFeedItemsSuccessCallback(feedItems);
+                    return;
+                }
+
+                const parentType: CommentFeedItemType =
+                    type === FEED_ITEM_TYPE_COMMENT ? FEED_ITEM_TYPE_COMMENT : FEED_ITEM_TYPE_ANNOTATION;
+
+                this.getFeedItemsWithReplies(feedItems, id, parentType)
```
jstoffan, line 754: Can we flatten out this promise to avoid needing the duplicate `.catch`?

```
this.getActiveFeedEntryData(feedItems)
    .then(({ id, type }) => {
        if (
            !id ||
            !type ||
            this.isActiveEntryInFeed(feedItems, activeFeedEntryId) ||
            !this.isItemTypeComment(type)
        ) {
            return Promise.resolve(feedItems);
        }

        const parentType: CommentFeedItemType =
            type === FEED_ITEM_TYPE_COMMENT ? FEED_ITEM_TYPE_COMMENT : FEED_ITEM_TYPE_ANNOTATION;

        return this.getFeedItemsWithReplies(feedItems, id, parentType)
    })
    .then(updatedItems => this.fetchFeedItemsSuccessCallback(updatedItems))
    .catch(error => this.fetchFeedItemsErrorCallback(feedItems, [error]));

jstoffan (approved):
LGTM as a temporary solution. We'll want to revisit this code in the near future, though.

wpiesiak (approved):

src/elements/content-sidebar/ActivitySidebar.js
kkuliczkowski-box: Right, I've changed to `some` and also noticed that I could actually use a method (`getCommentFeedItemByReplyId`) I've created earlier, that already has logic to find the parent based on reply id and have used it now.
I've merged this PR because of time constraints but opened a new one for these minor fixes: https://github.com/box/box-ui-elements/pull/3148

src/elements/content-sidebar/ActivitySidebar.js
kkuliczkowski-box: It should not as threads can be "hidden" only if there are 2 or more replies. Hiding replies basically means setting replies array to just one entry (the last reply).

kkuliczkowski-box:
I've created a new PR in order to address the latest comments: https://github.com/box/box-ui-elements/pull/3148

PR #3144, opened 2022-11-18
feat: Add option to display header in ContentExplorerModalContainer

apieceofbart:
What are we trying to do here?
We're trying to show the custom header for the table. 
Because by default it won't be visible because table header height is 0 I am also introducing tableHeaderHeight prop in order to make the header visible. 
I am also adding `headerRowRenderer` that we can pass through to the Table component that will allow to create a custom table headers.
If this is not clear enough we could introduce another prop `disableHeader` that would control the header visibility (you'd still need the height though). 
This is a non breaking change. 

image

### DIFF
--- a/src/features/content-explorer/item-list/ItemList.js
+++ b/src/features/content-explorer/item-list/ItemList.js
@@ -137,6 +137,8 @@ const ItemList = ({
     onItemDoubleClick,
     onItemNameClick,
     onLoadMoreItems,
+    headerHeight,
+    headerRenderer,
     itemIconRenderer,
     itemNameLinkRenderer,
     itemButtonRenderer,
@@ -145,8 +147,6 @@ const ItemList = ({
     width,
     height,
     rowHeight = DEFAULT_ROW_HEIGHT,
-    headerHeight,
-    headerRowRenderer,
 }) => {
     const getRow = ({ index }) => items[index];
 
@@ -221,6 +221,8 @@ const ItemList = ({
                 headerClassName="table-header-item"
                 width={width}
                 height={height}
+                headerHeight={headerHeight}
+                headerRowRenderer={headerRenderer}
                 rowHeight={rowHeight}
                 rowCount={items.length}
                 onRowClick={onItemClick}
@@ -228,8 +230,6 @@ const ItemList = ({
                 rowGetter={getRow}
                 rowRenderer={renderRow}
                 noRowsRenderer={noItemsRenderer}
-                headerHeight={headerHeight}
-                headerRowRenderer={headerRowRenderer}
                 {...tableProps}
             >
                 <Column
@@ -286,6 +286,8 @@ ItemList.propTypes = {
     onItemDoubleClick: PropTypes.func,
     onItemNameClick: PropTypes.func,
     onLoadMoreItems: PropTypes.func,
+    headerHeight: PropTypes.number,
+    headerRenderer: PropTypes.func,
     itemIconRenderer: PropTypes.func,
     itemNameLinkRenderer: PropTypes.func,
     itemButtonRenderer: PropTypes.func,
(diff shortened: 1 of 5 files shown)


src/features/content-explorer/messages.js
```diff
@@ -143,6 +143,11 @@ const messages = defineMessages({
         description: "Results label for number of items on list when it's just 1",
         id: 'boxui.contentExplorer.result',
     },
+    folderNameTableHeader: {
```
tjuanitas, line 146: this could probably be column header instead of table header

```suggestion
    folderNameColumnHeader: {
```

src/features/content-explorer/item-list/ItemList.js
```diff
@@ -287,6 +293,7 @@ ItemList.propTypes = {
     width: PropTypes.number.isRequired,
     height: PropTypes.number.isRequired,
     rowHeight: PropTypes.number,
+    tableHeaderHeight: PropTypes.number,
```
tjuanitas, line 296: since we call the other prop just `rowHeight`, could we call this `headerHeight`?

tjuanitas (commented):
I notice that the header in the mocks also include an icon and a dynamic number (number of items?), how would this be addressed?

I think this might a use case for using `itemRowRenderer`. My thinking is that you could pass in the table headers as the first index of `items`. then in your custom renderer, render a custom header for the first index and return `defaultTableRowRenderer` for all other indexes.
would that work?

also, let's remove the link to figma in the PR description since this is public and the mocks are for internal purposes. we could add it to the post in slack.

ref:
https://github.com/bvaughn/react-virtualized/blob/master/source/Table/defaultRowRenderer.js
https://github.com/bvaughn/react-virtualized/blob/master/source/Table/defaultHeaderRowRenderer.js

apieceofbart:
> I notice that the header in the mocks also include an icon and a dynamic number (number of items?), how would this be addressed?
> 
> I think this might be a use case for using `itemRowRenderer`. My thinking is that you could pass in the table headers as the first index of `items`. then in your custom renderer, render a custom header for the first index and return `defaultTableRowRenderer` for all other indexes. would that work?
> 
> also, let's remove the link to figma in the PR description since this repo is public and the mocks are for internal purposes. we could add it to the post in slack.
> 
> ref: https://github.com/bvaughn/react-virtualized/blob/master/source/Table/defaultRowRenderer.js https://github.com/bvaughn/react-virtualized/blob/master/source/Table/defaultHeaderRowRenderer.js

Icon and dynamic number won't be implemented since we don't have the count data and I think(?) we can sort as well. So I was just planning to do the minimum. 
Anyway, I was able to implement pretty similar results using `headerRowRenderer` that we can pass to Table component. It's ugly as hell and requires a lot of React.cloneElement (since I don't want to harcode the classNames or styles that are used inside the Table component) but it works.

So if you don't mind I will remove the column labels, will keep the `headerHeight` prop and add `headerRowRenderer` and that should be it.

src/features/content-explorer/item-list/ItemList.js
apieceofbart: sure

src/features/content-explorer/messages.js
apieceofbart: will remove completely so doesn't matter now

src/features/content-explorer/messages.js
```diff
@@ -143,6 +143,11 @@ const messages = defineMessages({
...
+        defaultMessage: 'Folder Name',
```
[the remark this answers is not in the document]
apieceofbart: will remove completely

tjuanitas (approved):

menaczar (approved):

SkyisAakash (approved):

megansmith-box (approved):

SkyisAakash (approved):

PR #3147, opened 2022-11-22
fix(replies): fix total_replies_count value update after delete reply

wpiesiak:
`total_replies_count` property on parent was not updated properly after deleting its reply


jankowiakdawid (approved):

jstoffan (approved):
Do we have a test to prevent the defect from recurring?

PR #3149, opened 2022-11-22
fix(dropdown-menu): fix positioning for responsive dropdown menu

ivanthai:
### Overview
There are a few bug fixes here related to a few bugs we encountered with DropdownMenu for full-screen which includes

- Position menu correctly when in full-screen - Added an additional css class and JS logic to properly position the menu above the bottom viewport
- Remove unneccesary scrollbar

### Demo
**Bad menu positioning and jerking motion**

buie-dropdown-menu-rpositioning-bug

**Fix**

buie-jarring-menu-bug-fix

### DIFF
--- a/src/components/dropdown-menu/DropdownMenu.scss
+++ b/src/components/dropdown-menu/DropdownMenu.scss
@@ -60,7 +60,7 @@
                 bottom: auto;
                 left: auto;
 
-                &.is-inline-bottom {
+                &.is-right-bottom-aligned {
                     top: auto;
                     bottom: 100%;
                 }
--- a/src/components/dropdown-menu/DropdownMenu.scss
+++ b/src/components/dropdown-menu/DropdownMenu.scss
@@ -54,13 +54,15 @@
                 box-shadow: $bdl-header-box-shadow;
             }
 
-            .submenu {
+            .submenu:not(.is-hidden) {
                 top: 100%;
                 right: 0;
+                bottom: auto;
                 left: auto;
 
-                &.is-hidden {
-                    display: none;
+                &.is-inline-bottom {
+                    top: auto;
+                    bottom: 100%;
                 }
             }
         }
--- a/src/components/dropdown-menu/DropdownMenu.scss
+++ b/src/components/dropdown-menu/DropdownMenu.scss
@@ -36,7 +36,7 @@
         .aria-menu:not(.submenu) {
             height: 100%;
             margin-top: $bdl-header-height;
-            overflow: scroll;
+            overflow-x: auto;
             background-color: $bdl-gray-02;
             border: none;
             border-radius: 0;
--- a/src/components/menu/SubmenuItem.tsx
+++ b/src/components/menu/SubmenuItem.tsx
@@ -129,11 +129,11 @@ class SubmenuItem extends React.Component<SubmenuItemProps, SubmenuItemState> {
         }
     };
 
-    closeSubmenu = debounce(() => {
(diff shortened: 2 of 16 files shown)


src/components/dropdown-menu/DropdownMenu.scss
```diff
@@ -54,13 +54,15 @@
                 box-shadow: $bdl-header-box-shadow;
             }
 
-            .submenu {
+            .submenu:not(.is-hidden) {
```
tjuanitas, line 57: what happens if we remove `:not(.is-hidden)`?

I'm wondering if the extra specificity here will take precedence: https://github.com/box/box-ui-elements/blob/00fc868b7f67566927eaaa11a6c5f7eb24438961/src/components/menu/SubmenuItem.scss#L20-L23

src/components/dropdown-menu/DropdownMenu.scss
```diff
@@ -36,7 +36,7 @@
         .aria-menu:not(.submenu) {
             height: 100%;
             margin-top: $bdl-header-height;
-            overflow: scroll;
+            overflow-x: auto;
```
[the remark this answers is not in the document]
ivanthai: There should not be a case for horizontal scrolling for a full screen menu. I think I want the scrollbar to appear when there are more elements that can fit on screen

src/components/menu/SubmenuItem.tsx
```diff
@@ -82,10 +83,12 @@ class SubmenuItem extends React.Component<SubmenuItemProps, SubmenuItemState> {
             submenuTriggerElBounding.right + submenuElBounding.width > rightBoundaryElementBounding.right;
         const isBottomAligned =
             submenuTriggerElBounding.top + submenuElBounding.height > bottomBoundaryElementBounding.bottom;
-
+        const isInlineBottom =
```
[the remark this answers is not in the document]
ivanthai: it is hovering over the menu and is aligned at the bottom. Maybe a better name would be inside bottom aligned?

src/components/dropdown-menu/DropdownMenu.scss
ivanthai: It doesn't take precedence

ivanthai: These styles only really apply when the submenu is not hidden anyways and it was easier to just omit is-hidden to not have to deal with style overrides

src/components/menu/SubmenuItem.tsx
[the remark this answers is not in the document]
ivanthai: I renamed it right-bottom-aligned

tjuanitas (approved):
looks good to me but had a final clarification on the name

src/components/dropdown-menu/DropdownMenu.scss
```diff
@@ -54,13 +54,15 @@
...
                 top: 100%;
                 right: 0;
+                bottom: auto;
                 left: auto;
 
-                &.is-hidden {
-                    display: none;
+                &.is-right-bottom-aligned {
+                    top: auto;
+                    bottom: 100%;
```
tjuanitas, line 65: I'm curious why `isRightBottomAligned` a better name when there doesn't seem to be any styles related to the horizontal position?

ivanthai: This actually depends on the outer style

Machoper (approved):

PR #3153, opened 2022-11-28
chore(template): Use ready to merge label instead of merge button

greg-in-a-box:
Added PR template to highlight that people should use the ready-to-merge label instead of clicking the merge button.  Adding the label, adds their PR to the merge train for safe merging.


tjuanitas (commented):
the `ready-to-merge` label can only be applied by those under webapp. how should we handle for developers that don't have access?

QingyuChai (approved):

PR #3157, opened 2022-11-29
feat(annotation-thread): added reply form, UI and other minor fixes

kkuliczkowski-box:
- Added reply form that is separated from the `ActivityThread`
- Improved UI to meet the design requirements
- Added `getUserProfileUrl` prop (and support for it) to `AnnotationThread`  in order to display linked user names

### DIFF
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActivityThread.scss
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActivityThread.scss
@@ -20,6 +20,10 @@
         padding: 0;
     }
 
+    .bcs-AnnotationActivity-menu {
+        top: 0;
+    }
+
     .bcs-Comment-timestamp {
         margin-bottom: 16px;
 
--- a/src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThread.scss
+++ b/src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThread.scss
@@ -5,11 +5,14 @@
     background-color: $white;
 
     .bcs-ActivityThread {
+        box-shadow: none;
+    }
+
+    .bcs-ActivityThread-content {
         position: relative;
         max-height: 528px;
         margin-bottom: $bdl-grid-unit * 4;
         overflow: auto;
-        box-shadow: none;
 
         .bcs-AnnotationActivity-menu {
             top: $sidebarActivityFeedSpacingVertical;
--- a/src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThread.scss
+++ b/src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThread.scss
@@ -1,7 +1,6 @@
 @import '../../../../styles/variables';
 
 .AnnotationThread {
-    max-width: 340px;
-    max-height: 600px;
+    min-width: $bdl-grid-unit * 85;
     background-color: $white;
 }
--- a/src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThread.scss
+++ b/src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThread.scss
@@ -1,6 +1,23 @@
-@import '../../../../styles/variables';
+@import '../../../common/variables';
 
 .AnnotationThread {
-    min-width: $bdl-grid-unit * 85;
+    min-width: 340px;
     background-color: $white;
+
+    .bcs-ActivityThread {
+        position: relative;
+        max-height: 528px;
+        margin-bottom: $bdl-grid-unit * 4;
+        overflow: auto;
+        box-shadow: none;
+
+        .bcs-AnnotationActivity-menu {
+            top: $sidebarActivityFeedSpacingVertical;
+        }
+    }
+
+    .bcs-ActivityThreadReplyForm-toggle,
+    .bcs-ActivityThreadReplyForm-toggle:hover {
+        margin: 0 $sidebarActivityFeedSpacingHorizontal $sidebarActivityFeedSpacingVertical;
+    }
 }
--- a/src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadContent.js
+++ b/src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadContent.js
@@ -57,8 +57,6 @@ const AnnotationThreadContent = ({
         filePermissions: permissions,
         errorCallback: onError,
     });
-    // eslint-disable-next-line no-console
-    console.log('annotation', annotation);
 
     return (
         <>
--- a/src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadContent.js
+++ b/src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadContent.js
@@ -60,26 +60,27 @@ const AnnotationThreadContent = ({
 
     return (
         <>
-            <ActivityThread
-                hasReplies
-                getAvatarUrl={getAvatarUrl}
-                getMentionWithQuery={getMentionWithQuery}
-                getUserProfileUrl={getUserProfileUrl}
-                isAlwaysExpanded
-                isRepliesLoading={isLoading}
-                mentionSelectorContacts={mentionSelectorContacts}
-                onReplyDelete={handleReplyDelete}
-                onReplyEdit={handleReplyEdit}
-                replies={replies}
-                repliesTotalCount={replies.length}
-            >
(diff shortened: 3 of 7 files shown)


src/elements/content-sidebar/activity-feed/annotation-thread/useAnnotationThread.js
```diff
@@ -98,6 +99,7 @@ const useAnnotationThread = ({
     };
 
     useAnnotatorEvents({
+        eventEmitter,
```
kkuliczkowski-box, line 102: I'm passing `eventEmitter` here in order for the current state of code to work, but I will create a seperate ticket for the refactoring of events handling as suggested by @jstoffan here: https://github.com/box/box-ui-elements/pull/3106#discussion_r1019439252

src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThread.scss
```diff
@@ -1,7 +1,6 @@
 @import '../../../../styles/variables';
 
 .AnnotationThread {
-    max-width: 340px;
-    max-height: 600px;
+    min-width: $bdl-grid-unit * 85;
```
jstoffan, line 4: We can probably revert this unless `85` is a meaningful value for our grid.

src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadContent.scss
```diff
@@ -3,3 +3,22 @@
 .AnnotationThreadContent-loading {
     padding: $bdl-grid-unit * 5 0;
 }
+
+.AnnotationThread {
```
jstoffan, line 7: Should this be in `AnnotationThread.scss`?

src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadContent.scss
```diff
@@ -3,3 +3,22 @@
...
+    .bcs-ActivityThread {
+        position: relative;
+        max-height: $bdl-grid-unit * 132;
```
jstoffan, line 10: Should we share this value with the other activity types? How are we keeping the threads and reply forms consistent?

src/elements/content-sidebar/activity-feed/annotation-thread/useAnnotationThread.js
```diff
@@ -155,8 +157,10 @@ const useAnnotationThread = ({
     });
 
     React.useEffect(() => {
-        handleFetch({ annotationId, successCallback: handleFetchAnnotationSuccess });
-    }, [annotationId, handleFetch]);
+        if (!annotation || annotation.id !== annotationId) {
```
jstoffan, line 160: Minor, but can we convert this to a guard clause?

src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadContent.js
```diff
@@ -55,39 +59,49 @@ const AnnotationThreadContent = ({
...
+                    />
+                )}
+            </ActivityThread>
             {annotation && (
-                <AnnotationActivity
-                    getAvatarUrl={getAvatarUrl}
-                    currentUser={currentUser}
-                    isCurrentVersion
-                    item={annotation}
+                <ActivityThreadReplyForm
```
jstoffan, line 98: Will we need to replicate this composition in each usage of `ActivityThread`? Should `ActivityThreadReplyForm` be used within that component, instead?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityThread.scss
```diff
@@ -20,6 +20,10 @@
         padding: 0;
     }
 
+    .bcs-AnnotationActivity-menu {
+        top: 0;
```
jstoffan, line 24: What's the purpose behind this change?

src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThread.scss
kkuliczkowski-box: I don't think it's a meaningful value. I thought that the `$bdl-grid-unit` is a "base" grid size value and the width of elements should be connected to it. I've changed to a static value.

src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadContent.scss
kkuliczkowski-box: Makes sense. I've moved it.

src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadContent.scss
kkuliczkowski-box: This value sets the maximum height of the Activity Thread only when used in `AnnotationThread`. The reasoning behind it is that when Annotation is displayed in a "bubble" over the Canvas Board we don't want the thread to take too much vertical space. Similar limitation is not needed in other activity types that are displayed in Sidebar.
I've also changed the value to a static one.

src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadContent.js
kkuliczkowski-box: We could potentially refactor how the `ActivityThread` is used in Sidebar. Right now the `ActivityThreadReplyForm` is built as part of `ActivityThread` and is rendered if `onReplyCreate` handler is passed.

Here the `ActivityThreadReplyForm` needs to be displayed below the `ActivityThread` as the activity thread's height needs to be limited, but the form should always be visible.

src/elements/content-sidebar/activity-feed/activity-feed/ActivityThread.scss
kkuliczkowski-box: When Annotation is rendered within the `ActivityThread` styling is a bit different and this change is to compensate that, so that the menu button is displayed in the same position.

As we haven't tested Annotations in Canvas before, we haven't caught that before.

src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadContent.js
jstoffan: I see. I think we could add a div around `children`, `renderButton()`, and the replies section [here](https://github.com/box/box-ui-elements/blob/master/src/elements/content-sidebar/activity-feed/activity-feed/ActivityThread.js#L113) that we can use to set the max-height and overflow values on. It's debatable, though, so not a blocking issue.

src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThread.scss
```diff
@@ -1,7 +1,23 @@
...
 
 .AnnotationThread {
-    max-width: 340px;
-    max-height: 600px;
+    min-width: 340px;
     background-color: $white;
+
+    .bcs-ActivityThread {
+        position: relative;
+        max-height: 528px;
```
jstoffan, line 9: Is this being derived from something? Is it `600px` minus some other value(s)? If so, we should have those values as variables, ideally.

jstoffan (approved):

src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThread.scss
kkuliczkowski-box: I don't think it's based on some other value.

As discussed, I've added `.bcs-ActivityThread-content` wrapper in the `ActivityThread`, so that the form is not used outside of it.

src/elements/content-sidebar/activity-feed/annotation-thread/AnnotationThreadContent.js
```diff
@@ -55,39 +58,43 @@ const AnnotationThreadContent = ({
...
-            isAlwaysExpanded
-            isRepliesLoading={isLoading}
-            mentionSelectorContacts={mentionSelectorContacts}
-            onReplyCreate={handleReplyCreate}
-            onReplyDelete={handleReplyDelete}
-            onReplyEdit={handleReplyEdit}
-            replies={replies}
-            repliesTotalCount={replies.length}
-        >
+        <>
```
jstoffan, line 61: Should this be reverted?

jstoffan (approved):

wpiesiak (approved):

PR #3167, opened 2022-12-06
fix(collaborator-avatars): open link in same window if inside iframe

Machoper:


src/features/collaborator-avatars/CollaboratorListItem.js
```diff
@@ -38,7 +39,7 @@ const CollaboratorListItem = (props: Props) => {
                 <Link
                     href={profileURL || `/profile/${userID}`}
                     rel="noopener"
-                    target="_blank"
+                    target={isInsideIframe() ? '_self' : '_blank'}
```
tjuanitas, line 42: what's the current behavior if it's inside an iframe vs not in an iframe? is this considered a bug?

would there be a use case to use the current behavior? i'm wondering how it'll impact current usages since it may be expected now. e.g. if a customer is using ContentSharing inside of their own iframe

what if we pass `target` as a prop in `usernameProps` which is in `trackingProps`?

src/utils/iframe.js
```diff
@@ -4,6 +4,15 @@
  * @author Box
  */
 
+/**
+ * Check if the current window is inside an iframe
+ *
+ * @return {boolean}
+ */
+export function isInsideIframe() {
+    return window.top !== window.self;
```
tjuanitas, line 13: does this solution work for all browsers we support?

tjuanitas: seems like it should work

for reference I found an Azure implementation:
https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/src/utils/BrowserUtils.ts#L43-L45

Machoper:
Closing the PR as this is part of a wider issue, which we will fix at a later time.

PR #3175, opened 2022-12-12
feat(content-explorer): Allow selecting empty selection in ContentExplorer

apieceofbart:
In Admin console we would like to support a case for choosing empty selection in the ContentExplorer. 
The reason for this is we keep the selection in local state and pass it to the ContentExplorer as initial selection. Therefore it is possible user would like to unselect everything and save this. Currently "Save" (choose) button is disabled if the selection is empty. 
We propose to add a boolean prop `allowChoosingEmptySelection` that when true would allow for selecting even when selection is empty.


src/features/content-explorer/content-explorer-modal-container/ContentExplorerModalContainer.js
```diff
@@ -13,6 +13,8 @@ class ContentExplorerModalContainer extends Component {
          * Each column has to be a Column element
          */
         additionalColumns: PropTypes.arrayOf(PropTypes.element),
+        /**  Should user be able to choose empty selection in MULTI_SELECT mode, defaults to false */
+        allowChoosingEmptySelection: PropTypes.bool,
```
greg-in-a-box, line 17: Can we change the name of this prop using a similar convention to other props (e.g. `isSelectAllAllowed`) in this component? I think `isNoSelectionAllowed` might be more appropriate

src/features/content-explorer/content-explorer/ContentExplorerActionButtons.js
```diff
@@ -46,9 +47,13 @@ const ContentExplorerActionButtons = ({
...
 
-        if (onChooseClick && chosenItems.length > 0) {
+        if (onChooseClick && (chosenItems.length > 0 || allowChoosingEmptySelection)) {
             onChooseClick(chosenItems);
         }
+
+        // if (onChooseClick && chosenItems.length > 0) {
+        //     onChooseClick(chosenItems);
+        // }
     };
```
greg-in-a-box, line 58: I think its safe to say that we can delete these comments

src/features/content-explorer/content-explorer/__tests__/ContentExplorer.test.js
```diff
@@ -240,6 +240,22 @@ describe('features/content-explorer/content-explorer/ContentExplorer', () => {
                 expect(wrapper.find('ContentExplorerActionButtons').prop('areButtonsDisabled')).toBe(true);
             });
         });
+
+        test('should render enabled action buttons in MULTI_SELECT mode if selection is empty but allowChoosingEmptySelection is true', () => {
```
greg-in-a-box, line 244:
```suggestion
        test('should render with action buttons enabled in MULTI_SELECT mode if there is no selection made when isNoSelectionAllowed is true', () => {
```

src/features/content-explorer/content-explorer-modal-container/ContentExplorerModalContainer.js
```diff
@@ -13,6 +13,8 @@ class ContentExplorerModalContainer extends Component {
          * Each column has to be a Column element
          */
         additionalColumns: PropTypes.arrayOf(PropTypes.element),
+        /**  Should user be able to choose empty selection in MULTI_SELECT mode, defaults to false */
```
greg-in-a-box, line 16:
```suggestion
        /**  Allow users to choose no selections in MULTI_SELECT mode, defaults to false */
```

greg-in-a-box (changes requested):
left some feedback

src/features/content-explorer/content-explorer/ContentExplorerActionButtons.js
apieceofbart: haha yes, thank you

src/features/content-explorer/content-explorer-modal-container/ContentExplorerModalContainer.js
apieceofbart: I don't mind `isSelectNoneAllowed` but I think the difference between this and `isSelectAllAllowed` is that there's also **choose** action involved - not only selecting.

greg-in-a-box (approved):

tjuanitas (approved):

PR #3176, opened 2022-12-13
feat(annotation): Add icon indicator for AnnotationActivity

jankowiakdawid:
This PR adds an icon indicator to an `AnnotationActivity` item (similarly to `Task` component).

Screenshot 2022-12-13 at 17 09 18
(I don't know what happened with the upload picture, but if you click it it shows)


src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
```diff
@@ -23,6 +23,8 @@ import type { Annotation, AnnotationPermission, FeedItemStatus } from '../../../
 import type { GetAvatarUrlCallback, GetProfileUrlCallback } from '../../../common/flowTypes';
 import type { SelectorItems, User } from '../../../../common/types/core';
 
+import IconAnnotation from '../../../../icons/two-toned/IconTaskGeneral';
```
jankowiakdawid, line 26: This is not the correct Icon, designers started the work on adding the right one to the assets and refreshing it along the way. To parallelize the process, I used the one from `Task` to get the review on code changes sooner.

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
```diff
@@ -138,8 +140,9 @@ const AnnotationActivity = ({
                         'bcs-is-pending': isPending || error,
                     })}
                 >
-                    <Media.Figure>
+                    <Media.Figure className="bcs-AnnotationActivity-avatar">
                         <Avatar getAvatarUrl={getAvatarUrl} user={createdByUser} />
```
jstoffan, line 144: It looks like `Avatar` already uses a `Badgeable` component for the "external" icon internally. Can we expand the `Avatar` API slightly to allow a badge to be passed in, instead? That way we don't have to restyle it every time it's used and a mixin wouldn't be necessary.

jankowiakdawid: I think this would require slight changes to the border of the badge, at least for the medium size. From what I see in the designs (and code) for medium there should be `2px` border instead of `1px`.

jankowiakdawid: I that in this commit https://github.com/box/box-ui-elements/pull/3176/commits/affc7b8925a3dc13612406e384ba9fc43e837e3b

Not exactly sure if that's what you had in mind.
I've added it to storybook and that's how it looks like
(I don't know why the image doesn't show, it's available after click)

Screenshot 2022-12-15 at 15 45 21

I'm not sure if it's worth the effort, there are slight visual regressions because it was hard to have the same placement for icons of various size.

src/components/avatar/Avatar.tsx
```diff
@@ -67,13 +86,17 @@ function Avatar({ avatarUrl, className, name, id, isExternal, shouldShowExternal
         avatar = <UnknownUserAvatar className="avatar-icon" />;
     }
 
+    let badge = null;
+    if (shouldShowExternal && isExternal) {
+        badge = <GlobeBadge16 className="bdl-Avatar-badge bdl-Avatar-externalBadge" />;
+    } else if (badgeIcon) {
+        badge = React.cloneElement(badgeIcon, {
```
jstoffan, line 93: Rather than cloning the element, can we add another wrapping div around it and apply the styles there?

src/elements/content-sidebar/activity-feed/task-new/AvatarGroupAvatar.js
```diff
@@ -40,13 +40,19 @@ const AvatarGroupAvatar = React.memo<Props>(({ user, status, getAvatarUrl, class
...
-            status={status}
-            className={`bcs-AvatarGroupAvatar-statusIcon ${camelCase(status)}`}
-            height={12}
-            width={12}
-            title={<FormattedMessage {...messages.taskAssignmentCompleted} />}
+        <Avatar
+            badgeIcon={
+                <StatusIcon
+                    className={`${camelCase(status)}`}
+                    height={12}
```
jstoffan, line 47: Can we standardize the size of the activity type badges? Can we use 20/20 for this one?

src/components/avatar/Avatar.tsx
```diff
@@ -67,13 +86,17 @@ function Avatar({ avatarUrl, className, name, id, isExternal, shouldShowExternal
         avatar = <UnknownUserAvatar className="avatar-icon" />;
     }
 
+    let badge = null;
+    if (shouldShowExternal && isExternal) {
+        badge = <GlobeBadge16 className="bdl-Avatar-badge bdl-Avatar-externalBadge" />;
```
jstoffan, line 91: It's probably safest to style this separately and leave `bdl-Avatar-badge` off of it.

jstoffan (commented):
I think it's worth pursuing for the sake of having a standardized experience. It looks like we're already pretty close, we just need to fix one badge size?

src/components/avatar/Avatar.tsx
jankowiakdawid: `bdl-Avatar-badge` reuses most generic part of the styling there were in `bdl-Avatar-externalBadge`. I've reused it for other badges but didn't wanted to mess with more specific styling.

I guess I could do it like so:
```css
.bdl-Avatar-badge,
.bdl-Avatar-externalBadge {
    background-color: $white;
    border-color: $white;
    border-style: solid;
    border-radius: 50%;
}
```

and then this line would be
```suggestion
        badge = <GlobeBadge16 className="bdl-Avatar-externalBadge" />;
```

src/components/avatar/Avatar.tsx
jankowiakdawid: Yes, we can. I usually try to avoid creating `div`'s cause too many of them will mage page slow.

src/elements/content-sidebar/activity-feed/annotations/AnnotationActivity.js
jankowiakdawid: Talked with designers and the decision was to go with badge icon sizes being related to Avatar size, as in design system. Also, I didn't touch the external badge styling.

jankowiakdawid:
This is how the Avatar with badge looks like now

Screenshot 2022-12-19 at 14 20 00
(I don't get why GH won't show the picture, you can click on it, and it'll send you to the picture)

src/components/avatar/Avatar.scss
```diff
@@ -81,6 +97,20 @@
                 left: calc(100% - 10px);
             }
         }
+
+        &.avatar--iconBadge {
+            line-height: 0; // .badgeable-container has `line-height: 1 which evaluates to 13px which is too much for small badge
+
+            @include Badgeable-bottomRightBadge {
```
jstoffan, line 104: Is this mixin call needed? It should be inherited by the styles above in the base class?

jstoffan (approved):
LGTM, so long as we're okay with this change affecting Preview right away.

wpiesiak (approved):

PR #3177, opened 2022-12-13
fix(content-picker): add accessible names to Footer

MarieRogers:

### DIFF
--- a/src/elements/content-picker/Footer.js
+++ b/src/elements/content-picker/Footer.js
@@ -56,60 +56,61 @@ const Footer = ({
     children,
     renderCustomActionButtons,
     showSelectedButton,
-}: Props) => (
-    <footer className="bcp-footer">
-        <div className="bcp-footer-left">
-            {showSelectedButton && !isSingleSelect && (
-                <Button className="bcp-selected" onClick={onSelectedClick} type="button">
-                    <FormattedMessage
-                        className="bcp-selected-count"
-                        {...messages.selected}
-                        values={{ count: selectedCount }}
-                    />
-                    {hasHitSelectionLimit && (
-                        <span className="bcp-selected-max">
-                            (<FormattedMessage {...messages.max} />)
-                        </span>
-                    )}
-                </Button>
-            )}
-        </div>
-        <div className="bcp-footer-right">
-            {children}
+}: Props) => {
+    const cancelMessage = intl.formatMessage(messages.cancel);
+    const chooseMessage = intl.formatMessage(messages.choose);
+    return (
+        <footer className="bcp-footer">
+            <div className="bcp-footer-left">
+                {showSelectedButton && !isSingleSelect && (
+                    <Button className="bcp-selected" onClick={onSelectedClick} type="button">
+                        <FormattedMessage
+                            className="bcp-selected-count"
+                            {...messages.selected}
+                            values={{ count: selectedCount }}
+                        />
+                        {hasHitSelectionLimit && (
+                            <span className="bcp-selected-max">
+                                (<FormattedMessage {...messages.max} />)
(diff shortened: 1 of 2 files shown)


src/elements/content-picker/Footer.js
```diff
@@ -104,4 +112,4 @@ const Footer = ({
     </footer>
 );
 
-export default Footer;
+export default injectIntl(Footer);
```
QingyuChai, line 115: Hi Marie! You might need to do something along the lines of:
```javascript
export { Footer as FooterBase };
export default injectIntl(Footer);
```
And in the tests, you then `import { FooterBase as Footer }` and mock the `injectIntl()` function

QingyuChai: smt like here:
https://github.com/box/box-ui-elements/blob/3e358f9aca921c04adb7e60cd14f584001dd38c0/src/components/breadcrumb/__tests__/Breadcrumb.test.js

MarieRogers: The tests pass without this, is it necessary? Not sure if you were suggesting this when the tests were failing because of incorrect GH setup or if this helps to test the intl specifically

QingyuChai: I've seen stuff previously where the intl env isn't setup for tests, so they would fail if you don't mock them. Seems to not be the case for BUIE

src/elements/content-picker/Footer.js
```diff
@@ -53,55 +56,61 @@ const Footer = ({
...
+                        {hasHitSelectionLimit && (
+                            <span className="bcp-selected-max">
+                                (<FormattedMessage {...messages.max} />)
+                            </span>
+                        )}
+                    </Button>
+                )}
+            </div>
+            <div className="bcp-footer-right">
+                {children}
```
tjuanitas, line 82: weird line break they had before

tjuanitas (approved):
:tada:

QingyuChai (dismissed):

src/elements/content-picker/__tests__/Footer.test.js
```diff
@@ -25,6 +25,15 @@ describe('elements/content-picker/Footer', () => {
             expect(wrapper.find('.footer-child').length).toBe(1);
         });
 
+        test('should render Footer tooltips with aria-label', () => {
```
tjuanitas, line 28: these `aria-label`s are added to the buttons and not the tooltips

src/elements/content-picker/__tests__/Footer.test.js
```diff
@@ -25,6 +25,15 @@ describe('elements/content-picker/Footer', () => {
...
+            const tooltipButtonWrapper = getWrapper()
+                .find('Tooltip')
+                .find('Button');
```
tjuanitas, line 31: you could probably call this `buttons = getWrapper()` since tooltipButton might imply a button inside of the tooltip itself

I think you could also get away with removing `.find('Tooltip')`

src/elements/content-picker/__tests__/Footer.test.js
```diff
@@ -25,6 +25,15 @@ describe('elements/content-picker/Footer', () => {
...
+
+            expect(tooltipButtonWrapper.at(0).prop('aria-label')).toEqual('Cancel');
+            expect(tooltipButtonWrapper.at(1).prop('aria-label')).toEqual('Choose');
```
tjuanitas, line 34: we could use `toBe` instead of `toEqual` for primitive values since it's doesn't need the deep equality provided by `toEqual`

tjuanitas (approved):

QingyuChai (approved):

PR #3178, opened 2022-12-13
fix(sub-header): add accessibility label to SortButton

QingyuChai:
Normal source:
image

Hover source:
image

### DIFF
--- a/src/elements/common/sub-header/SortButton.js
+++ b/src/elements/common/sub-header/SortButton.js
@@ -5,19 +5,22 @@
  */
 
 import React from 'react';
-import { FormattedMessage } from 'react-intl';
+import { injectIntl } from 'react-intl';
 import Button from '../../../components/button';
 import IconSort from '../../../icons/general/IconSort';
 import messages from '../messages';
 import Tooltip from '../Tooltip';
 import './SortButton.scss';
 
-const SortButton = (props: ?Object) => (
-    <Tooltip text={<FormattedMessage {...messages.sort} />}>
-        <Button className="be-btn-sort" type="button" {...props}>
-            <IconSort />
-        </Button>
-    </Tooltip>
-);
+const SortButton = ({ intl, ...rest }: ?Object) => {
+    const sortMessage = intl.formatMessage(messages.sort);
+    return (
+        <Tooltip text={sortMessage}>
+            <Button aria-label={sortMessage} className="be-btn-sort" type="button" {...rest}>
+                <IconSort />
+            </Button>
+        </Tooltip>
+    );
+};
 
-export default SortButton;
+export default injectIntl(SortButton);
--- a/src/elements/common/sub-header/SortButton.js
+++ b/src/elements/common/sub-header/SortButton.js
@@ -6,13 +6,21 @@
 
 import React from 'react';
 import { injectIntl } from 'react-intl';
+import type { IntlShape } from 'react-intl';
+
 import Button from '../../../components/button';
 import IconSort from '../../../icons/general/IconSort';
-import messages from '../messages';
 import Tooltip from '../Tooltip';
+
+import messages from '../messages';
+
 import './SortButton.scss';
 
-const SortButton = ({ intl, ...rest }: ?Object) => {
+type Props = {
+    intl: IntlShape,
+};
+
+const SortButton = ({ intl, ...rest }: Props) => {
     const sortMessage = intl.formatMessage(messages.sort);
     return (
         <Tooltip text={sortMessage}>
--- a/src/elements/common/sub-header/SortButton.js
+++ b/src/elements/common/sub-header/SortButton.js
@@ -31,4 +31,5 @@ const SortButton = ({ intl, ...rest }: Props) => {
     );
 };
 
+export { SortButton as SortButtonBase };
 export default injectIntl(SortButton);
--- a/src/elements/common/sub-header/__tests__/Sort.test.js
+++ b/src/elements/common/sub-header/__tests__/Sort.test.js
@@ -7,7 +7,7 @@ import SortButton from '../SortButton';
 import messages from '../../messages';
 import { SORT_ASC, SORT_DESC } from '../../../../constants';
 
-describe('elements/SubHeader/Sort', () => {
+describe('elements/common/sub-header/Sort', () => {
     test('should render a button and menu with 4 menu items', () => {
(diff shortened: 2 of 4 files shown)


src/elements/common/sub-header/SortButton.js
```diff
@@ -5,19 +5,22 @@
...
 import './SortButton.scss';
 
-const SortButton = (props: ?Object) => (
-    <Tooltip text={<FormattedMessage {...messages.sort} />}>
-        <Button className="be-btn-sort" type="button" {...props}>
-            <IconSort />
-        </Button>
-    </Tooltip>
-);
+const SortButton = ({ intl, ...props }: ?Object) => {
```
tjuanitas, line 15: small: I would do rest instead of props
```suggestion
const SortButton = ({ intl, ...rest }: ?Object) => {
```

tjuanitas (dismissed):

src/elements/common/sub-header/SortButton.js
QingyuChai: good point!

tjuanitas (approved):

tjuanitas (dismissed):

greg-in-a-box (dismissed):

greg-in-a-box (dismissed):
@QingyuChai  Add a test/assertion to check for the existence of the aria label that was added

src/elements/common/sub-header/__tests__/SortButton.test.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import { shallow } from 'enzyme';
```
greg-in-a-box, line 2: I dont think we need this imported because its global

src/elements/common/sub-header/__tests__/SortButton.test.js
```diff
@@ -0,0 +1,25 @@
...
+import { SortButtonBase } from '../SortButton';
```
tjuanitas, line 3: we could put the named import last after the default ones (specifically for this component section of imports)

src/elements/common/sub-header/__tests__/SortButton.test.js
```diff
@@ -0,0 +1,25 @@
...
+import Button from '../../../../components/button';
+import IconSort from '../../../../icons/general/IconSort';
+
+const intlMock = {
+    formatMessage: jest.fn().mockReturnValue('Sort'),
+};
+
+describe('Elements/SubHeader/SortButton', () => {
```
tjuanitas, line 11:
```suggestion
describe('elements/common/sub-header/SortButton', () => {
```

src/elements/common/sub-header/__tests__/SortButton.test.js
```diff
@@ -0,0 +1,25 @@
...
+    test('should render IconSort', () => {
+        const wrapper = getWrapper();
+        expect(wrapper.exists(IconSort)).toBe(true);
+    });
+
+    test('should have aria-label "sort"', () => {
+        const wrapper = getWrapper();
+        const button = wrapper.find(Button);
+        expect(button.prop('aria-label')).toBe('Sort');
+        expect(button.prop('aria-describedBy')).toBeFalsy();
```
tjuanitas, line 23: the test above for IconSort uses `.toBe(true)`

might change this to match `.toBe(false)`

src/elements/common/sub-header/__tests__/SortButton.test.js
```diff
@@ -0,0 +1,25 @@
...
+
+describe('Elements/SubHeader/SortButton', () => {
+    const getWrapper = () => shallow(<SortButtonBase intl={intlMock} />);
+
+    test('should render IconSort', () => {
+        const wrapper = getWrapper();
+        expect(wrapper.exists(IconSort)).toBe(true);
+    });
+
+    test('should have aria-label "sort"', () => {
```
tjuanitas, line 19: if you're putting the text in quotation marks, then I would capitalize to `Sort` so it's the exact text

src/elements/common/sub-header/__tests__/SortButton.test.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import { SortButtonBase } from '../SortButton';
+import Button from '../../../../components/button';
+import IconSort from '../../../../icons/general/IconSort';
+
+const intlMock = {
+    formatMessage: jest.fn().mockReturnValue('Sort'),
+};
```
tjuanitas, line 9: not sure if needed if this is global: https://github.com/box/box-ui-elements/blob/master/scripts/jest/mocks/lib-intl-mock.js

src/elements/common/sub-header/__tests__/SortButton.test.js
jpan-box, line 23: should describedBy be described**b**y?

src/elements/common/sub-header/__tests__/SortButton.test.js
QingyuChai: I'll update the other 2 tests as well

src/elements/common/sub-header/__tests__/SortButton.test.js
QingyuChai: I dont think that will work - I'm checking for aria-describedBy to be undefined, null, or an empty string, and I think they will fill the `toBe(false)` condition

src/elements/common/sub-header/__tests__/SortButton.test.js
QingyuChai: I don't think this is global, both VSCode and `yarn test /path/to/test` throw errors - I could import it but I'd rather just use a custom mock

src/elements/common/sub-header/__tests__/SortButton.test.js
QingyuChai: eagle eye Jason

src/elements/common/sub-header/__tests__/Sort.test.js
```diff
@@ -7,7 +7,7 @@ import SortButton from '../SortButton';
 import messages from '../../messages';
 import { SORT_ASC, SORT_DESC } from '../../../../constants';
 
-describe('elements/SubHeader/Sort', () => {
+describe('elements/common/sub-header/Sort', () => {
```
tjuanitas, line 10: love it!

tjuanitas (approved):

greg-in-a-box (approved):

karelee7 (approved):

PR #3179, opened 2022-12-13
fix(sharedlinksettings): fix expiration timestamp

ahorowitz123:
- Correctly format incoming date so that it gets set to UTC midnight. This is what happens on the folder settings page and is expected from the backend functions that both add the value to the DB and fetch it from the DB


src/features/shared-link-settings-modal/ExpirationSection.js
```diff
@@ -29,6 +29,7 @@ const ExpirationSection = ({
     const datepicker = (
         <div>
             <DatePicker
+                dateFormat="utcTime"
```
tjuanitas, line 32: I'd be a little cautious since `SharedLinkSettingsModal` is used in the ContentSharing element and Canvas, probably other apps as well. there might be an expectation now that this returns a date in the local timezone so that this change could break those usages while fixing EUA.

tjuanitas, line 32: the default date is using the browser's timezone: https://github.com/box/box-ui-elements/blob/1ab5eb11137ba471039f1b6c4bc8119358bf1de0/src/features/shared-link-settings-modal/SharedLinkSettingsModal.js#L146

tjuanitas (commented):
you might consider converting the date value (which comes from the DB in UTC) to the local timezone before passing into `SharedLinkSettingsModal` as `expirationTimestamp`

src/features/shared-link-settings-modal/ExpirationSection.js
ahorowitz123: Ok thanks for the heads up. I will confirm in these places as well. The issue here is that the backend (monolith) for link settings expects it in this format so the current implementation is broken and showing inconsistencies in other parts of the app (folder settings)

ahorowitz123:
> you might consider converting the date value (which comes from the DB in UTC) to the local timezone before passing into `SharedLinkSettingsModal` as `expirationTimestamp`

The main issue here is that the value in the DB is incorrect. And is set differently than the one in folder settings which is correct

src/features/shared-link-settings-modal/ExpirationSection.js
ahorowitz123: @tjuanitas just pushed an update to make this set by the `SharedLinkSettingsModal` component so I can set this through EUA and keep the changes isolated

src/features/shared-link-settings-modal/__tests__/ExpirationSection.test.js
```diff
@@ -95,5 +95,14 @@ describe('features/shared-link-settings-modal/ExpirationSection', () => {
 
             expect(wrapper.prop('inputProps')['data-prop']).toEqual('input');
         });
+
+        test('should pass dateFormat to DatePicker', () => {
```
jcamden, line 99: Is there any way to test this (and the tests in `SharedLinkSettingsModal`) via the UI instead of making assertions about internals? IMO, it's okay and even ideal for, e.g., Sonarqube to complain about ostensible gaps in coverage in order for us to write non-brittle tests that solely define user-facing functionality.

jcamden (dismissed):
Congratulations on sorting this out! Seemed like a tough one. My only suggestions/question is for us to limit our testing to assertions about user-facing functionality. Not a big enough deal to request changes, so I'm approving here. Might also be good if we could get @tjuanitas to give it another look too, if he's available.

src/features/shared-link-settings-modal/SharedLinkSettingsModal.js
```diff
@@ -212,8 +214,8 @@ class SharedLinkSettingsModal extends Component {
         this.setState({ isPasswordEnabled: event.target.checked });
     };
 
-    onExpirationDateChange = date => {
-        this.setState({ expirationDate: date, expirationError: undefined });
+    onExpirationDateChange = (date, formattedDate) => {
+        this.setState({ expirationDate: new Date(formattedDate), expirationError: undefined });
```
tjuanitas, line 218: could we confirm that existing uses are unaffected by this change?

does this work as expected for the different `dateFormat` options? (i.e. are all the `formattedDate` values valid to be passed into `new Date()`?)

tjuanitas (commented):
this looks good to me but I would test and confirm that the change to `onExpirationDateChange` is returning the same value as before for the unchanged usages

jcamden (approved):
Really great work.

mickr (approved):

PR #3180, opened 2022-12-14
fix(item-action): add aria-label to plainbutton

jpan-box:
**After**
* aria-label added, matching hover text
* aria-describedby removed
Screenshot 2022-12-13 at 4 56 16 PM
Screenshot 2022-12-14 at 10 56 01 AM

### DIFF
--- a/src/elements/content-uploader/ItemAction.js
+++ b/src/elements/content-uploader/ItemAction.js
@@ -1,6 +1,6 @@
 /**
  * @flow
- * @file Item action component
+ * @file Item action component displayed on the upload toast, e.g. cancel/resume
  */
 
 import React from 'react';
@@ -50,6 +50,7 @@ const ItemAction = ({
     let icon = <IconClose />;
     let tooltip;
     const { code } = error;
+    const { formatMessage } = intl;
 
     if (isFolder && status !== STATUS_PENDING) {
         return null;
@@ -96,12 +97,14 @@ const ItemAction = ({
             </PrimaryButton>
         );
     }
+    const isDisabled = status === STATUS_STAGED;
+    const tooltipText = tooltip && formatMessage(tooltip);
 
     return (
         <div className="bcu-item-action">
             {tooltip ? (
-                <Tooltip position="top-left" text={intl.formatMessage(tooltip)}>
-                    <PlainButton onClick={onClick} type="button" isDisabled={status === STATUS_STAGED}>
+                <Tooltip position="top-left" text={tooltipText}>
+                    <PlainButton aria-label={tooltipText} isDisabled={isDisabled} onClick={onClick} type="button">
                         {icon}
                     </PlainButton>
                 </Tooltip>
--- a/src/elements/content-uploader/__tests__/ItemAction.test.js
+++ b/src/elements/content-uploader/__tests__/ItemAction.test.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import noop from 'lodash/noop';
 import { shallow } from 'enzyme';
+import PlainButton from '../../../components/plain-button';
 import { ItemActionForTesting as ItemAction } from '../ItemAction';
 import {
     ERROR_CODE_UPLOAD_FILE_SIZE_LIMIT_EXCEEDED,
@@ -15,7 +16,7 @@ describe('elements/content-uploader/ItemAction', () => {
     const getWrapper = props =>
         shallow(
             <ItemAction
-                intl={{ formatMessage: data => <span {...data} /> }}
+                intl={{ formatMessage: data => data.defaultMessage }}
                 onClick={noop}
                 status={STATUS_PENDING}
                 {...props}
@@ -75,5 +76,14 @@ describe('elements/content-uploader/ItemAction', () => {
 
         expect(wrapper.exists('PrimaryButton')).toBe(true);
         expect(wrapper.exists('PlainButton')).toBe(false);
+    });
+
+    test('should have aria-label "Cancel this upload" when status is pending', () => {
+        const wrapper = getWrapper({
+            status: STATUS_PENDING,
+        });
+        const plainButton = wrapper.find(PlainButton);
+        expect(plainButton.prop('aria-label')).toBe('Cancel this upload');
+        expect(plainButton.prop('aria-describedby')).toBeFalsy();
     });
 });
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -66,7 +66,7 @@
         "lint:js": "eslint --max-warnings=0 .",
         "lint:ts": "tsc && eslint --ext=.tsx,.ts --max-warnings=0 .",
         "lint:css": "stylelint \"src/**/*.scss\" --syntax scss",
-        "prebuild:es": "npm-run-all build:variables copy:styles ts:defs copy:flow",
```
tjuanitas, line 69: let's not commit this

src/elements/content-uploader/ItemAction.js
```diff
@@ -101,7 +101,12 @@ const ItemAction = ({
         <div className="bcu-item-action">
             {tooltip ? (
                 <Tooltip position="top-left" text={intl.formatMessage(tooltip)}>
-                    <PlainButton onClick={onClick} type="button" isDisabled={status === STATUS_STAGED}>
+                    <PlainButton
+                        onClick={onClick}
+                        type="button"
+                        isDisabled={status === STATUS_STAGED}
+                        aria-label={intl.formatMessage(tooltip)}
```
tjuanitas, line 108: nit: could we alphasort these props?

src/elements/content-uploader/__tests__/ItemAction.test.js
```diff
@@ -15,7 +16,7 @@ describe('elements/content-uploader/ItemAction', () => {
     const getWrapper = props =>
         shallow(
             <ItemAction
-                intl={{ formatMessage: data => <span {...data} /> }}
+                intl={{ formatMessage: data => data.defaultMessage }}
```
tjuanitas, line 19: good catch!

tjuanitas (commented):
looks great overall! small things

could we include similar changes for `ItemRemove` as well?

src/elements/content-uploader/__tests__/ItemAction.test.js
jpan-box: it was didi :D

package.json
jpan-box: ok!

src/elements/content-uploader/ItemAction.js
jpan-box: yes, i need to think about things that way!

tjuanitas: don't worry bout it, we'll add a prettier config to handle this

jpan-box:
Planning on handling ItemRemove in a separate PR!

src/elements/content-uploader/ItemAction.js
```diff
@@ -101,7 +101,12 @@ const ItemAction = ({
         <div className="bcu-item-action">
             {tooltip ? (
                 <Tooltip position="top-left" text={intl.formatMessage(tooltip)}>
-                    <PlainButton onClick={onClick} type="button" isDisabled={status === STATUS_STAGED}>
+                    <PlainButton
+                        aria-label={intl.formatMessage(tooltip)}
```
tjuanitas, line 105: not necessary to change but could destruct `intl` near the top of the function:

`const { formatMessage } = intl;`

tjuanitas (approved):
LGTM

src/elements/content-uploader/ItemAction.js
```diff
@@ -101,7 +101,12 @@ const ItemAction = ({
...
+                        isDisabled={status === STATUS_STAGED}
```
karelee7, line 106: super minor and not absolutely necessary! But for readability, should we put this in a variable rather than inline it? (haven't seen this inlining pattern for props that often elsewhere)

jpan-box: I'm all for making things more readable! I only see my perspective so it's helpful to hear where things can be changed :D Just to make sure, do you mean aria-label should be defined as a variable or this `status === STATUS_STAGED`?

src/elements/content-uploader/ItemAction.js
jpan-box: did it!

src/elements/content-uploader/ItemAction.js
jpan-box: I think I did both with @tjuanitas 's help!

tjuanitas (approved):
lgtm

QingyuChai (approved):
l g t m

src/elements/content-uploader/ItemAction.js
```diff
@@ -50,6 +50,7 @@ const ItemAction = ({
     let icon = <IconClose />;
     let tooltip;
     const { code } = error;
+    const { formatMessage } = intl;
```
JChan106, line 53: can we still use `intl.formatMessage` instead of extracting it out?

src/elements/content-uploader/__tests__/ItemAction.test.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import noop from 'lodash/noop';
 import { shallow } from 'enzyme';
+import PlainButton from '../../../components/plain-button';
```
JChan106, line 4: Should we add test cases for null or undefined tooltip text?

src/elements/content-uploader/ItemAction.js
QingyuChai: I think Jason did that as a response to https://github.com/box/box-ui-elements/pull/3180#discussion_r1048812307

tjuanitas: probably could go either way now. originally intended when there were multiple usages of `intl.formatMessage`

src/elements/content-uploader/__tests__/ItemAction.test.js
jpan-box: yes, i wanted to but i think i should avoid it for now since:
* i don't know how we would end up with an undefined tooltip - only if there was no status?
* should keep this PR just to the a11y bits. We've got a big pile of components to go through and update so we want to do a pass with all just the a11y stuff first.

src/elements/content-uploader/ItemAction.js
karelee7: Ah I meant the `status === STATUS_STAGED`, but it sounds like you've got it! :)

JChan106 (approved):

PR #3181, opened 2022-12-14
fix(itemremove): add aria-label

jpan-box:
WIP - Don't merge yet

After:
* Removed aria-describedby
* Added aria-label to match tooltip text
Screenshot 2022-12-14 at 3 39 03 PM


src/elements/content-uploader/ItemRemove.js
```diff
@@ -15,9 +16,9 @@ import type { UploadItem, UploadStatus } from '../../common/types/upload';
 type Props = {
     onClick: (item: UploadItem) => void,
     status: UploadStatus,
-};
+} & InjectIntlProvidedProps;
 
-const ItemRemove = ({ onClick, status }: Props) => {
+const ItemRemove = ({ onClick, status, intl }: Props) => {
```
tjuanitas, line 21: mini
```suggestion
const ItemRemove = ({ intl, onClick, status }: Props) => {
```

src/elements/content-uploader/ItemRemove.js
```diff
@@ -4,7 +4,8 @@
  */
 
 import React from 'react';
-import { FormattedMessage } from 'react-intl';
+import { injectIntl } from 'react-intl';
+import type { InjectIntlProvidedProps } from 'react-intl';
```
tjuanitas, line 8: are we able to move this `import type` at the bottom of this list just before `import type { UploadItem, UploadStatus } [...]`?

to keep the type imports grouped together

src/elements/content-uploader/ItemRemove.js
```diff
@@ -33,13 +34,20 @@ const ItemRemove = ({ onClick, status }: Props) => {
 
     return (
         <div className="bcu-item-action">
-            <Tooltip position="top-left" text={<FormattedMessage {...messages.remove} />}>
-                <PlainButton onClick={onClick} type="button" isDisabled={status === STATUS_STAGED} {...resin}>
+            <Tooltip position="top-left" text={intl.formatMessage(messages.remove)}>
+                <PlainButton
+                    aria-label={intl.formatMessage(messages.remove)}
+                    isDisabled={status === STATUS_STAGED}
```
tjuanitas, line 40: you could update this to include the similar feedback from the ItemAction PR

src/elements/content-uploader/__tests__/ItemRemove.test.js
```diff
@@ -0,0 +1,30 @@
...
+import { ItemRemoveBase as ItemRemove } from '../ItemRemove';
+
+import { STATUS_ERROR, STATUS_IN_PROGRESS, STATUS_STAGED } from '../../../constants';
+
+describe('elements/content-uploader/ItemRemove', () => {
+    test.each`
+        status
+        ${STATUS_IN_PROGRESS}
+        ${STATUS_STAGED}
+        ${STATUS_ERROR}
```
tjuanitas, line 15: there's another syntax you can use since this table only has one column

I think possibly this:
`test.each([STATUS_ERROR, STATUS_IN_PROGRESS, STATUS_STAGED])`

src/elements/content-uploader/__tests__/ItemRemove.test.js
```diff
@@ -0,0 +1,30 @@
...
+import { shallow } from 'enzyme';
+import noop from 'lodash/noop';
+
+import PlainButton from '../../../components/plain-button';
+import { ItemRemoveBase as ItemRemove } from '../ItemRemove';
+
+import { STATUS_ERROR, STATUS_IN_PROGRESS, STATUS_STAGED } from '../../../constants';
+
+describe('elements/content-uploader/ItemRemove', () => {
+    test.each`
```
tjuanitas, line 11: do we need this `test.each`? shouldn't the tooltip always appear regardless of the status?

src/elements/content-uploader/ItemRemove.js
jpan-box: ESLint didn't like it... would this be something i can modify in the eslint config or is it worth ignoring the eslint warning?

Screenshot 2022-12-19 at 2 53 01 PM

src/elements/content-uploader/ItemRemove.js
jpan-box: tanks!

src/elements/content-uploader/ItemRemove.js
jpan-box: yes i can!

src/elements/content-uploader/ItemRemove.js
tjuanitas: eh it's probably fine how it is

src/elements/content-uploader/__tests__/ItemRemove.test.js
jpan-box: Hmm yeah, that's a good point. In that case shouldn't I test all possible statuses? Or just pick one?

tjuanitas: I would omit it altogether here since the status doesn't have an affect on the tooltip and their behaviors are independent

tjuanitas: if you wanted to test statuses, such as if `isDisabled` on `status === STATUS_STAGED`, then this would be a separate test

tjuanitas: I think the main difference between this component and the previous one (ItemAction) is that the tooltip text will change for the different statuses. but in this one, the text will always be "Remove"

tjuanitas: for the shallow wrapper, you could set any status. as we add more tests, then we could have a `getWrapper` implementation that would use the most common status but it still wouldn't affect the aria-label

jpan-box: sounds good, will do!!

src/elements/content-uploader/__tests__/ItemRemove.test.js
```diff
@@ -0,0 +1,25 @@
...
+import { ItemRemoveBase as ItemRemove } from '../ItemRemove';
+
+import { STATUS_IN_PROGRESS } from '../../../constants';
+
+describe('elements/content-uploader/ItemRemove', () => {
+    test('should have aria-label "Remove" and no aria-describedby', () => {
+        const wrapper = shallow(
+            <ItemRemove
+                intl={{ formatMessage: data => data.defaultMessage }}
+                isResumableUploadsEnabled
```
tjuanitas, line 15: is `isResumableUploadsEnabled` part of this component?

src/elements/content-uploader/__tests__/ItemRemove.test.js
```diff
@@ -0,0 +1,25 @@
...
+                onClick={noop}
+                status={STATUS_IN_PROGRESS} // Use any status. The label does not change based on status.
```
tjuanitas, line 17: feel free to keep this but it's also okay to omit comments if they could be inferred from the component logic

jpan-box: i did this for me - i am less than stellar at inferring logic :sadparrot:

src/elements/content-uploader/__tests__/ItemRemove.test.js
jpan-box: oops no it is not, that is leftover from testing. will remove

JChan106 (approved):
lgtm

PR #3187, opened 2022-12-19
feat(tooltip): support error tooltip position

megansmith-box:

### DIFF
--- a/src/components/select-button/SelectButton.js
+++ b/src/components/select-button/SelectButton.js
@@ -3,18 +3,23 @@ import * as React from 'react';
 import classNames from 'classnames';
 
 import Tooltip from '../tooltip';
+import type { Position } from '../tooltip';
 import './SelectButton.scss';
 
 type Props = {
     children?: React.Node,
     className: string,
     error?: React.Node,
+    errorTooltipPosition?: Position,
     isDisabled: boolean,
 };
 
 const SelectButton = React.forwardRef<Props, HTMLButtonElement>(
-    ({ children, className = '', error, isDisabled = false, ...rest }: Props, ref) => (
-        <Tooltip isShown={!!error} position="middle-right" text={error} theme="error">
+    (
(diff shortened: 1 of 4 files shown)


src/components/select-button/SelectButton.js
```diff
@@ -2,19 +2,21 @@
...
     error?: React.Node,
+    errorTooltipPosition?: Position,
     isDisabled: boolean,
 };
 
 const SelectButton = React.forwardRef<Props, HTMLButtonElement>(
-    ({ children, className = '', error, isDisabled = false, ...rest }: Props, ref) => (
-        <Tooltip isShown={!!error} position="middle-right" text={error} theme="error">
+    ({ children, className = '', error, errorTooltipPosition, isDisabled = false, ...rest }: Props, ref) => (
+        <Tooltip isShown={!!error} position={errorTooltipPosition || 'middle-right'} text={error} theme="error">
```
tjuanitas, line 19: I see we do `position={errorTooltipPosition || 'middle-right'}` in other places as well but i'm curious why not we don't default `errorTooltipPosition` in the function parameters

src/components/select-button/SelectButton.js
```diff
@@ -2,19 +2,21 @@
 import * as React from 'react';
 import classNames from 'classnames';
 
+import type { Position } from '../tooltip';
```
tjuanitas, line 5: minor but could move this after the Tooltip component

tjuanitas (approved):

src/components/select-button/SelectButton.js
megansmith-box: No reason in particular, I see it both ways in the code. I like the default function parameter suggestion!

tjuanitas (approved):

QingyuChai (dismissed):
lgtm! I think one of Trevor's comments is still unaddressed though

this one: https://github.com/box/box-ui-elements/pull/3187#discussion_r1052737897

QingyuChai (approved):
thx for the fix!

tjuanitas (approved):

PR #3189, opened 2022-12-20
fix(content-explorer): add aria-label to grid slider

MarieRogers:


src/components/grid-view/GridViewSlider.js
```diff
@@ -40,6 +40,7 @@ const GridViewSlider = ({ columnCount, gridMaxColumns, gridMinColumns, intl, max
                     <IconMinusThin color={bdlGray50} width={14} height={14} />
                 </PlainButton>
                 <input
+                    aria-label={intl.formatMessage(messages.gridViewColumnSize)}
```
tjuanitas, line 43: not necessary but there could be an opportunity to destruct `intl` at the top of the function:
`const { formatMessage } = intl;`
since `formatMessage` is being used in several places

src/components/grid-view/__tests__/GridViewSlider.test.js
```diff
@@ -2,9 +2,36 @@ import React from 'react';
...
 import GridViewSlider from '../GridViewSlider';
 
+const intl = {
+    formatMessage: jest.fn(),
+};
+
 describe('components/grid-view/GridViewSlider', () => {
     test('should render()', () => {
-        const wrapper = shallow(<GridViewSlider columnCount={4} maxColumnCount={7} onChange={() => {}} />);
+        const wrapper = mount(
```
tjuanitas, line 11: if possible could we keep the `shallow`? `mount` comes with more overhead that's not always needed

if we add `export { GridViewSlider as GridViewSliderBase };` in `GridViewSlider.js`, then I think we could still use shallow?

src/components/grid-view/__tests__/GridViewSlider.test.js
```diff
@@ -2,9 +2,36 @@ import React from 'react';
...
+                .find('PlainButton')
+                .at(0)
+                .prop('aria-label'),
+        ).toBe('Decrease column size');
+        expect(
+            wrapper
+                .find('PlainButton')
+                .at(1)
+                .prop('aria-label'),
+        ).toBe('Increase column size');
```
tjuanitas, line 35: I would move these to a separate test specific for testing `aria-label`s

src/components/grid-view/__tests__/GridViewSlider.test.js
MarieRogers: Hmm I tried this in so many different forms and couldn't get it to work. I was wondering if maybe because it's a const and not a class it doesn't work in the same way?

tjuanitas: I'm thinking something like this:

```
// inside of GridViewSlider.js
export { GridViewSlider as GridViewSliderBase };


// inside of GridViewSlider.test.js
import { GridViewSliderBase as GridViewSlider } from '../GridViewSlider';
```

and I think you'll need to change the `const intl` on line 5 to something like:
```
const intl = {
    formatMessage: message => message.defaultMessage,
};
```

src/components/grid-view/__tests__/GridViewSlider.test.js
```diff
@@ -1,10 +1,44 @@
...
+const getWrapper = () =>
+    shallow(
+        <GridViewSlider
+            columnCount={4}
+            maxColumnCountcolumnCount={4}
+            gridMaxColumns={7}
+            gridMinColumns={2}
+            intl={intl}
+            maxColumnCount={7}
+            onChange={() => {}}
```
tjuanitas, line 18: smallest of thing: this could be sorted

src/components/grid-view/__tests__/GridViewSlider.test.js
```diff
@@ -1,10 +1,44 @@
...
+
+const intl = {
+    formatMessage: message => message.defaultMessage,
+};
+
+const getWrapper = () =>
+    shallow(
+        <GridViewSlider
+            columnCount={4}
+            maxColumnCountcolumnCount={4}
```
tjuanitas, line 13: 404 `maxColumnCountcolumnCount`

tjuanitas (approved):

megansmith-box (approved):

karelee7 (approved):

PR #3193, opened 2022-12-20
fix(usm): add aria-label to expiration icon

MarieRogers:


src/features/collaborator-avatars/CollaboratorAvatarItem.js
```diff
@@ -44,14 +46,14 @@ const CollaboratorAvatarItem = (props: Props) => {
         ) : (
             <Avatar name={name || '-'} {...rest} isExternal={isExternalCollab} shouldShowExternal={allowBadging} />
         );
-
+    const expirationMessage = intl.formatMessage(messages.expirationText);
     const expirationBadge =
         allowBadging && expiration && expiration.executeAt ? (
             <Tooltip
                 position="middle-right"
                 text={<FormattedMessage {...messages.expirationTooltipText} values={{ date: expiration.executeAt }} />}
```
QingyuChai, line 54: can you update the tooltip `text` prop to use an `intl.formatMessage` instead of `<FormattedMessage />`? The latter creates a <span>, and is likely the reason why the `describedBy` doesn't disappear

src/features/collaborator-avatars/CollaboratorAvatarItem.js
```diff
@@ -45,13 +47,12 @@ const CollaboratorAvatarItem = (props: Props) => {
             <Avatar name={name || '-'} {...rest} isExternal={isExternalCollab} shouldShowExternal={allowBadging} />
         );
 
+    const expirationToolTipMessageValue = expiration && expiration.executeAt ? { date: expiration.executeAt } : null;
```
megansmith-box, line 50: [nit] Tooltip with lowercase t, below too

src/features/collaborator-avatars/CollaboratorAvatarItem.js
```diff
@@ -45,13 +47,12 @@ const CollaboratorAvatarItem = (props: Props) => {
...
+    const expirationToolTipMessage = intl.formatMessage(messages.expirationTooltipText, expirationToolTipMessageValue);
```
tjuanitas, line 51: could this be?
```suggestion
    const expirationToolTipMessage = intl.formatMessage(messages.expirationTooltipText, { date: expiration?.executeAt });
```

src/features/collaborator-avatars/__tests__/CollaboratorAvatarItem.test.js
```diff
@@ -1,9 +1,12 @@
 import * as React from 'react';
 
-import CollaboratorAvatarItem from '../CollaboratorAvatarItem';
+import { CollaboratorAvatarItemBase as CollaboratorAvatarItem } from '../CollaboratorAvatarItem';
 
 describe('features/collaborator-avatars/CollaboratorAvatarItem', () => {
-    const getWrapper = (props = {}) => shallow(<CollaboratorAvatarItem id={1} name="foo bar" {...props} />);
+    const intl = {
```
tjuanitas, line 6: minor: could we move this above the `describe`? that was the pattern we've been doing for the other ones

tjuanitas (approved):

src/features/collaborator-avatars/CollaboratorAvatarItem.js
```diff
@@ -45,13 +47,13 @@ const CollaboratorAvatarItem = (props: Props) => {
             <Avatar name={name || '-'} {...rest} isExternal={isExternalCollab} shouldShowExternal={allowBadging} />
         );
 
+    const expirationTooltipMessage = intl.formatMessage(messages.expirationTooltipText, {
+        date: expiration?.executeAt,
```
megansmith-box, line 51: If this is undefined, do we want to show a different Tooltip Message?

src/features/collaborator-avatars/__tests__/CollaboratorAvatarItem.test.js
```diff
@@ -1,9 +1,13 @@
 import * as React from 'react';
 
-import CollaboratorAvatarItem from '../CollaboratorAvatarItem';
+import { CollaboratorAvatarItemBase as CollaboratorAvatarItem } from '../CollaboratorAvatarItem';
+
+const intl = {
+    formatMessage: message => message.defaultMessage,
```
megansmith-box, line 6: Should we use jest?
```
const intl = { formatMessage: jest.fn().mockImplementation((message) => message.defaultMessage) };
```

src/features/collaborator-avatars/CollaboratorAvatarItem.js
MarieRogers: It will not even show up if it's undefined
`const expirationBadge =
        allowBadging && expiration && expiration.executeAt ? (`

src/features/collaborator-avatars/__tests__/CollaboratorAvatarItem.test.js
MarieRogers: Trevor had suggested this approach in a different PR, is there a reason why would we go with jest here instead?

https://github.com/box/box-ui-elements/pull/3189#discussion_r1052819182

megansmith-box: I suppose the reason would be consistency, I see it everywhere else in the code use jest. But if manual mocking works, that's ok too!

megansmith-box (approved):

src/features/collaborator-avatars/__tests__/CollaboratorAvatarItem.test.js
tjuanitas: @megansmith-box yeah that might've been a good idea to use jest. I saw some other tests do manual mocking so I didn't really think of using `mockImplementation` but maybe we could migrate all those to use jest in the future

PR #3197, opened 2022-12-27
fix(content-explorer): add aria label for MoreOptions button

QingyuChai:


src/elements/content-explorer/MoreOptions.js
```diff
@@ -1,13 +1,16 @@
 // @flow
 import React from 'react';
-import { FormattedMessage } from 'react-intl';
+import { FormattedMessage, injectIntl } from 'react-intl';
```
benjamin-shen, line 3: Should we replace all the `FormattedMessage`s with `intl.formatMessage()`? Like in the other accessibility PRs

benjamin-shen (commented):
I think something like "add aria-label" instead of "add button role" might be more accurate in the pr title

src/elements/content-explorer/MoreOptions.js
QingyuChai: I think that when they're in the JSX directly instead of as a prop, it's safer to keep them as `FormattedMessage`, because it creates a `<span />` wrapper. Removing it could potentially affect some styles.

benjamin-shen (approved):

src/elements/content-explorer/__tests__/MoreOptions.test.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
```
QingyuChai, line 1: skeleton tests - tracking untested components in a deveco ticket

karelee7 (approved):

PR #3198, opened 2022-12-28
feat(.gitignore): update with vscode history support

QingyuChai:
just adding support for the vscode history extension

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *~~bak
 .DS_Store
 .idea
+.history
 _site
 __diff_output__
 dist


tjuanitas (commented):
not sure if we want to include this since it's for a specific IDE and extension and i don't know how common it's used. have you tried using a global `.gitignore` using `core.excludesFile`?

QingyuChai:
> not sure if we want to include this since it's for a specific IDE and extension and i don't know how common it's used. have you tried using a global `.gitignore` using `core.excludesFile`?

Did not know it existed! Thanks haha, enjoy your day off

PR #3203, opened 2023-01-04
fix(preview-sidebar): fix preview versions behavior

Machoper:
### This PR is built on https://github.com/box/box-ui-elements/pull/3095, so I'm borrowing the same description with a few small changes:

Currently, if a version is selected in the Preview sidebar and the Preview sidebar is closed, it reverts to the latest version of the document.

This creates problems for small screens where the Preview sidebar is moved to the bottom and becomes a vertical sliding drawer. When an old version is selected and the drawer is closed, a user cannot view the old version because it reverts to the latest version of the document.

This PR fixes the issue by **adjusting the logic to reset the current version only when the user navigates away from version history**.

Note this would also affect the regular desktop experience. If a user closes the Preview Sidebar after selecting the version, it would remain on that version instead of changing back to the latest version (approved by relevant stakeholders). **This would not affect the current behavior when switching to other sidebars like "Details" or "Metadata"**.

#### small screen
before_small
#### large screen
before_large

#### small screen
after_small
#### large screen
after_large


tjuanitas (approved):
looks good from my testing

one thing to note: there seems to be an existing bug where if you go to an older file version and click the Next/Previous buttons, the version number of the previous file is still in the URL for the new file. but this is not a regression from this change

src/constants.js
```diff
@@ -396,6 +396,9 @@ export const SIDEBAR_VIEW_METADATA: 'metadata' = 'metadata';
 export const SIDEBAR_VIEW_ACTIVITY: 'activity' = 'activity';
 export const SIDEBAR_VIEW_VERSIONS: 'versions' = 'versions';
 
+/* ------------------ Sidebar Path ---------------------- */
+export const SIDEBAR_PATH_VERSIONS = '/:sidebar(activity|details)/versions/:versionId?';
```
jstoffan, line 400: Can we store this in a module within the `content-sidebar` directory?

src/elements/content-sidebar/Sidebar.js
```diff
@@ -111,8 +112,18 @@ class Sidebar extends React.Component<Props, State> {
...
         }
+
+        // Reset the current version id if the wrapping versions route is no longer active
+        if (onVersionChange && this.getVersionsMatchPath(prevLocation) && !this.getVersionsMatchPath(location)) {
+            onVersionChange(null);
+        }
     }
 
+    getVersionsMatchPath = (location: Location) => {
+        const pathname = getProp(location, 'pathname', '');
```
jstoffan, line 123: Do we need `getProp` here? I think `location` is always defined.

src/elements/content-sidebar/Sidebar.js
```diff
@@ -111,8 +112,18 @@ class Sidebar extends React.Component<Props, State> {
             this.setForcedByLocation();
             this.setState({ isDirty: true });
         }
+
+        // Reset the current version id if the wrapping versions route is no longer active
+        if (onVersionChange && this.getVersionsMatchPath(prevLocation) && !this.getVersionsMatchPath(location)) {
```
jstoffan, line 117: ContentSidebar already does a lot. I'm wondering if this would be better suited to live in `SidebarPanels`, since that's where we manage all of this already.

src/elements/content-sidebar/__tests__/SidebarPanels.test.js
```diff
@@ -144,4 +149,21 @@ describe('elements/content-sidebar/SidebarPanels', () => {
...
+        expect(onVersionChange).not.toBeCalled();
+
+        wrapper.setProps({ location: { pathname: '/activity/versions/123' }, isOpen: false });
+        expect(onVersionChange).not.toBeCalled();
+
+        wrapper.setProps({ location: { pathname: '/activity/versions/456' }, isOpen: true });
+        expect(onVersionChange).not.toBeCalled();
+
+        wrapper.setProps({ location: { pathname: '/details' } });
+        expect(onVersionChange).lastCalledWith(null);
```
jstoffan, line 167: Would `.toBeCalledWith(null)` work here or is the number and order of calls important?

src/elements/content-sidebar/__tests__/SidebarPanels.test.js
```diff
@@ -144,4 +149,21 @@ describe('elements/content-sidebar/SidebarPanels', () => {
             expect(instance.versionsSidebar.current.refresh).toHaveBeenCalledWith();
         });
     });
+
+    test('should reset the current version if the versions route is no longer active', () => {
+        const onVersionChange = jest.fn();
+        const wrapper = getWrapper({ location: { pathname: '/activity' }, onVersionChange });
+
+        wrapper.setProps({ location: { pathname: '/activity/versions/123' }, isOpen: true });
```
jstoffan, line 157: Can we add assertions for `/activity/versions` and `/details/versions`, as well?

src/elements/content-sidebar/__tests__/SidebarPanels.test.js
```diff
@@ -11,18 +11,23 @@ jest.mock('../SidebarUtils');
...
+                file={{ id: '1234' }}
+                hasActivity
+                hasDetails
+                hasMetadata
+                hasSkills
+                hasVersions
+                isOpen
+                {...rest}
+            />,
+            {
```
jstoffan, line 24: Curious, what was the reason for this change?

src/elements/content-sidebar/SidebarPanels.js
```diff
@@ -102,6 +105,21 @@ class SidebarPanels extends React.Component<Props, State> {
         this.setState({ isInitialized: true });
     }
 
+    componentDidUpdate(prevProps: Props): void {
+        const { location, onVersionChange }: Props = this.props;
+        const { location: prevLocation }: Props = prevProps;
```
jstoffan, line 110: Are the `: Props` explicit types needed here? I would think the type would be inferred from the variables.

src/elements/content-sidebar/__tests__/SidebarPanels.test.js
Machoper: It's from an earlier attempt to run a few assertions in series and I forgot to update it afterwards.
I just rearranged the tests and change it to `.toBeCalledWith(null)`.

src/elements/content-sidebar/__tests__/SidebarPanels.test.js
Machoper: `setProps()` can only be called on the root, which was `MemoryRouter` in the previous setup. So, this is more of a workaround to call `setProps()` on `SidebarPanels` and thus trigger `componentDidUpdate()`.

jstoffan (approved):
Nice job. 👍

src/elements/content-sidebar/SidebarPanels.js
Machoper: No, they are not needed. I copy-pasted from `Sidebar.js`. Deleted!

src/elements/content-sidebar/__tests__/SidebarPanels.test.js
```diff
@@ -144,4 +149,38 @@ describe('elements/content-sidebar/SidebarPanels', () => {
             expect(instance.versionsSidebar.current.refresh).toHaveBeenCalledWith();
         });
     });
+
+    describe('componentDidUpdate', () => {
+        const onVersionChange = jest.fn();
```
tjuanitas, line 154: just to confirm, the call counts are reset between the tests right?

tjuanitas (approved):
lgtm

PR #3205, opened 2023-01-09
fix(content-uploader): fix uploadinput keyboard accessibility

QingyuChai:
make the upload button tabbable, and keyboard interactable


src/elements/content-uploader/UploadInput.js
```diff
@@ -20,19 +20,29 @@ const UploadInput = ({
...
+        if (e.key === 'Enter' || e.key === ' ') {
+            hiddenInputRef.current?.click();
+        }
+    };
 
+    return inputLabel ? (
+        <>
+            <label className={inputLabelClass} tabIndex={0} onKeyDown={onKeyDown}>
+                {inputLabel}
+                <input
```
QingyuChai, line 36: note: tabIndex={0} on the input doesn't work, because it's nested within the label

src/elements/content-uploader/UploadInput.js
```diff
@@ -15,17 +15,27 @@ type Props = {
...
+    isFolderUpload = false,
+    isMultiple = true,
+}: Props) => {
+    const inputRef = React.useRef(null);
+
+    const onKeyDown = e => {
+        if (e.key === 'Enter' || e.key === ' ') {
+            if (inputRef.current) {
+                inputRef.current.click();
+            }
```
QingyuChai, line 30: Flow is not compatible with optional chaining `inputRef.current?.click()` so this will have to be used until we update Flow

QingyuChai: Also, I think it's more readable to have this as a separate if statement, rather than doing
`if ((e.key === 'Enter || e.key === ' ') && inputRef.current) { ... }`

tjuanitas (approved):

JChan106 (approved):
lgtm

PR #3209, opened 2023-01-11
fix(content-explorer): create default shared link when none exists

QingyuChai:
Customer reported issue where they, as a Viewer or Viewer uploader, are able to create shared links on the website, but unable to do so through the Content Explorer. After some investigation, there seems to be an extra permission check for creating shared links, where the user also needs permission to change the access level for that link. When creating a default access shared link, that permission is not actually needed, so an overwrite was added to the API call. 

Furthermore, when fetching info for a file that does not have a shared link, a shared link should have been generated upon clicking the share button. That functionality is documented, but was previously unimplemented.

This PR aims to address those points and updated tests


src/elements/content-explorer/ContentExplorer.js
```diff
@@ -1277,8 +1277,26 @@ class ContentExplorer extends Component<Props, State> {
     handleSharedLinkSuccess = (newItem: BoxItem) => {
```
QingyuChai, line 1277: This function seemed to be missing some functionality - the comments indicate it creates a shared link using enterprise defaults, but that part was implemented before

src/api/Item.js
```diff
@@ -333,7 +334,9 @@ class Item extends Base {
             this.successCallback = successCallback;
             this.errorCallback = errorCallback;
 
-            this.validateRequest(id, permissions);
+            // if we use the default access level, we don't need permission to set the access level
+            const skipCanSetShareAccess = access === undefined;
+            this.validateRequest(id, permissions, skipCanSetShareAccess);
```
QingyuChai, line 339: When a Viewer or a Viewer Uploader creates a shared link using enterprise default settings, they do not need the `can_set_share_access` permission. Thus, we need to skip this permission check using a boolean

test/integration/content-explorer/ContentExplorer.int.test.js
```diff
@@ -101,6 +101,13 @@ describe('ContentExplorer', () => {
             allowed_shared_link_access_levels: ['collaborators', 'open', 'company'],
             shared_link: null,
         });
+        cy.route('PUT', '**/files/319004423111', {
```
QingyuChai, line 104: We are doing one more request when clicking the share button, AND shared link does not already exist. This is the mock for that request

src/elements/content-explorer/__tests__/ContentExplorer.test.js
```diff
@@ -559,6 +559,50 @@ describe('elements/content-explorer/ContentExplorer', () => {
...
+        });
+
+        afterEach(() => {
+            getApiShareMock.mockClear();
+            getApiMock.mockClear();
+            updateCollectionMock.mockClear();
+        });
+
+        test('should create shared link when it doesnt exist yet', () => {
+            instance.handleSharedLinkSuccess({ ...boxItem, shared_link: null });
```
QingyuChai, line 590: I understand testing like this is an anti-pattern, but invoking this function naturally requires a very large amount of setup

src/api/Item.js
```diff
@@ -333,7 +334,9 @@ class Item extends Base {
             this.successCallback = successCallback;
             this.errorCallback = errorCallback;
 
-            this.validateRequest(id, permissions);
+            // if we use the default access level, we don't need permission to set the access level
+            const skipCanSetShareAccess = access === undefined;
```
bfoxx1906, line 338: what if access is null?

src/elements/content-explorer/__tests__/ContentExplorer.test.js
bfoxx1906: These are unit tests and not integration tests so I don't mind favoring simplicity here as long as the tests are comprehensive. So I think this is ok.

src/api/Item.js
QingyuChai: if access is null, it is treated as a "remove shared link" action, which requires the `can_set_share_access` permission

bfoxx1906: ok cool

bfoxx1906 (approved):

bfoxx1906 (approved):

PR #3211, opened 2023-01-12
fix(usm): add aria-label to tooltip close button

QingyuChai:
Adding the aria-label directly to the tooltip requires updating at least 20 tests, and 150+ snapshots. Trevor recommended the idea of a separate component for the close button for the tooltip, which seems to work! The existing BUIE close button won't work, due to a mismatched icon size, as well as no support for localization without a wrapper.

image

### DIFF
--- a/src/components/tooltip/CloseButton.tsx
+++ b/src/components/tooltip/CloseButton.tsx
@@ -18,4 +18,5 @@ const CloseButton = ({ intl, onClick }: Props) => (
     </PlainButton>
 );
 
+export { CloseButton as CloseButtonBase };
 export default injectIntl(CloseButton);
--- a/src/components/tooltip/CloseButton.tsx
+++ b/src/components/tooltip/CloseButton.tsx
@@ -1,0 +1,21 @@
+import * as React from 'react';
+import { injectIntl, IntlShape } from 'react-intl';
+
+import IconClose from '../../icon/fill/X16';
+import PlainButton from '../plain-button';
+
+// @ts-ignore flow import
+import messages from '../../common/messages';
+
+type Props = {
+    intl: IntlShape;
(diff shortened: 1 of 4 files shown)


src/components/tooltip/__tests__/__snapshots__/Tooltip.test.tsx.snap
```diff
@@ -247,16 +247,9 @@ exports[`components/tooltip/Tooltip render() should render with close button if
     role="tooltip"
   >
     hi
-    <PlainButton
-      className="tooltip-close-button"
+    <CloseButton
```
benjamin-shen, line 250: Should we add a new snapshot test for this component?

benjamin-shen (commented):
LGTM, left one comment

src/components/tooltip/__tests__/__snapshots__/Tooltip.test.tsx.snap
QingyuChai: think we're leaning away from snapshots for new tests, so I'll add some manual tests! good ide atho

benjamin-shen (approved):
Nice :)

src/components/tooltip/CloseButton.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+// @ts-ignore flow import
+import messages from '../../common/messages';
+
+type Props = {
+    intl: IntlShape;
+    onClick: (event: React.SyntheticEvent<HTMLButtonElement, Event>) => void;
+};
+
+const CloseButton = ({ intl, onClick }: Props) => (
```
tjuanitas, line 15: it's out of the scope of the original issue but in the future we should look into if we could combine these two components: https://github.com/box/box-ui-elements/tree/master/src/components/close-button

tjuanitas (approved):

PR #3214, opened 2023-01-17
feat(checkbox): make focus behavior configurable

jcamden:
These changes allow more control over focus styles by moving focus styles to a separate class and by declaring `onFocus` prop. These changes are needed in order to move focus styles to a click-target wrapper around the checkbox while maintaining the checkbox input as the actual focus. Here is an example usage:

Screenshot 2023-01-17 at 12 01 25 PM

### DIFF
--- a/src/components/checkbox/Checkbox.scss
+++ b/src/components/checkbox/Checkbox.scss
@@ -4,6 +4,12 @@
  * Checkbox
  **************************************/
 
+.Checkbox--isFocused {
+    &:focus + span::before {
+        border: 2px solid $primary-color;
+    }
+}
+
 .checkbox-label {
     position: relative;
     display: inline-flex;
@@ -52,7 +58,7 @@
         }
 
         &:focus + span::before {
-            border: 2px solid $primary-color;
+            border: none;
         }
 
         &:disabled {
--- a/src/components/checkbox/Checkbox.scss
+++ b/src/components/checkbox/Checkbox.scss
@@ -4,12 +4,6 @@
  * Checkbox
  **************************************/
 
-.Checkbox--isFocused {
-    &:focus + span::before {
-        border: 2px solid $primary-color;
-    }
-}
-
 .checkbox-label {
     position: relative;
     display: inline-flex;
@@ -58,7 +52,7 @@
         }
 
         &:focus + span::before {
-            border: none;
+            border: 2px solid $primary-color;
         }
 
         &:disabled {
--- a/src/components/checkbox/Checkbox.tsx
+++ b/src/components/checkbox/Checkbox.tsx
@@ -13,6 +13,8 @@ export interface CheckboxProps {
     description?: React.ReactNode;
     /** fieldLabel - label for the field shown on top of the checkbox */
     fieldLabel?: React.ReactNode;
+    /** hideFocusStyles - whether the focus styles are hidden or not */
+    hideFocusStyles?: boolean;
     /** hideLabel - whether the checkbox label is hidden or not */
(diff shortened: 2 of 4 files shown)


src/components/checkbox/Checkbox.tsx
```diff
@@ -13,6 +13,8 @@ export interface CheckboxProps {
     description?: React.ReactNode;
     /** fieldLabel - label for the field shown on top of the checkbox */
     fieldLabel?: React.ReactNode;
+    /** hideFocusStyles - whether the focus styles are hidden or not */
+    hideFocusStyles?: boolean;
```
tjuanitas, line 17: this looks like it might be a very specific use case. do we anticipate other apps would use this prop for checkboxes?

if this is for a specific component, would you be able to set the `border: none;` in that component's stylesheet?
```
&:focus + span::before {
    border: none;
}
```

src/components/checkbox/Checkbox.scss
```diff
@@ -4,6 +4,12 @@
  * Checkbox
  **************************************/
 
+.Checkbox--isFocused {
```
tjuanitas, line 7: for SUIT naming, this would probably be `.bdl-Checkbox-input.bdl-is-focused`:
- `bdl-Checkbox-input` since the class would be on the descendant element
- `bdl-is-focused` as state should be a separate class

src/components/checkbox/Checkbox.scss
```diff
@@ -52,7 +58,7 @@
         }
 
         &:focus + span::before {
-            border: 2px solid $primary-color;
+            border: none;
```
tjuanitas, line 61: would this style always take priority because the specificity is higher (not 100% sure though) than the `.Checkbox--isFocused` selector?

src/components/checkbox/Checkbox.tsx
jcamden: Yes, we can override the default styles from the component where this is consumer in EUA. It sounds like you think that's the best route, so I'll make that change---unless you say otherwise :)

src/components/checkbox/Checkbox.scss
jcamden: Ah, I wondered about this. Thank you very much for giving the answer there 👍

src/components/checkbox/Checkbox.scss
jcamden: That is a very good point, and come to think of it, I do not think I have not checked that the focus styles are indeed there without the prop. Probably should have written a story for that. But since I think we'll go the route of overriding from EUA, I will revert these changes to styles.

jcamden: (Pending verification that I can override them properly from EUA :) )

src/components/checkbox/Checkbox.tsx
```diff
@@ -67,10 +70,11 @@ const Checkbox = ({
             <input
                 aria-describedby={description ? `description_${inputID}` : ''}
                 checked={isChecked}
-                className={inputClassName}
+                className={classNames(inputClassName)}
```
tjuanitas, line 73: we can remove `classNames`
```suggestion
                className={inputClassName}
```

tjuanitas (approved):
small thingy

src/components/checkbox/Checkbox.tsx
jcamden: Oh, right.

tjuanitas (approved):

shahzadaziz (approved):

PR #3216, opened 2023-01-19
fix(shared-link-section): accessibility tooltip

megansmith-box:
Screen Shot 2023-01-25 at 5 20 42 PM

Screen Shot 2023-01-26 at 4 52 53 PM

Description:
The tooltip for the "Create shared link" is not presented to a screen reader. That is because it is associated to the `<div with tabindex="0>`, but since `<div>` does not support aria-describedby, the tooltip will not be read reliably by screen reader users.

Element Source
```
<div class="share-toggle-container tooltip-target tooltip-abutted tooltip-abutted-top tooltip-element-attached-bottom tooltip-element-attached-left tooltip-target-attached-top tooltip-target-attached-left" tabindex="0">
...
<input class="toggle-simple-input" name="toggle" type="checkbox">
```
Recommendation:

Move aria-describedby to the switch <input> and ensure that it is read reliably by screen readers.

### DIFF
--- a/src/features/unified-share-modal/SharedLinkSection.js
+++ b/src/features/unified-share-modal/SharedLinkSection.js
@@ -487,18 +487,16 @@ class SharedLinkSection extends React.Component<Props, State> {
         }
 
         const toggleComponent = (
-            <div className="share-toggle-container">
-                <Toggle
-                    isDisabled={!isToggleEnabled}
-                    isOn={isSharedLinkEnabled}
-                    label={linkText}
-                    name="toggle"
-                    onChange={onToggleSharedLink}
-                    ref={ref => {
-                        this.toggleRef = ref;
-                    }}
-                />
-            </div>
+            <Toggle
+                isDisabled={!isToggleEnabled}
+                isOn={isSharedLinkEnabled}
+                label={linkText}
+                name="toggle"
+                onChange={onToggleSharedLink}
+                ref={ref => {
+                    this.toggleRef = ref;
+                }}
+            />
         );
 
         if (!submitting) {
--- a/src/features/unified-share-modal/UnifiedShareModal.scss
+++ b/src/features/unified-share-modal/UnifiedShareModal.scss
@@ -255,14 +255,10 @@
         justify-content: space-between;
         margin-top: 10px;
 
-        .share-toggle-container {
+        .toggle-container {
             flex: 0 1 80%;
             margin-bottom: 0;
             overflow-y: hidden;
-
-            .toggle-container {
-                margin-bottom: 0;
-            }
 
             .toggle-simple-label {
                 color: $bdl-gray;
--- a/src/components/toggle/Toggle.js
+++ b/src/components/toggle/Toggle.js
@@ -21,6 +21,12 @@ type Props = {
     onBlur?: (e: SyntheticInputEvent<HTMLInputElement>) => any,
     /** change callback function called with event as the argument */
     onChange?: (e: SyntheticInputEvent<HTMLInputElement>) => any,
+    /** focus callback function called with event as the argument */
+    onFocus?: (e: SyntheticInputEvent<HTMLInputElement>) => any,
+    /** mouse enter callback function called with event as the argument */
+    onMouseEnter?: (e: SyntheticInputEvent<HTMLDivElement>) => any,
+    /** mouse leave callback function called with event as the argument */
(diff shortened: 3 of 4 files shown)


src/features/unified-share-modal/SharedLinkSection.js
```diff
@@ -457,6 +458,7 @@ class SharedLinkSection extends React.Component<Props, State> {
         const canAddSharedLink = this.canAddSharedLink(isSharedLinkEnabled, item.grantedPermissions.itemShare);
         const canRemoveSharedLink = this.canRemoveSharedLink(isSharedLinkEnabled, canChangeAccessLevel);
         const isToggleEnabled = (canAddSharedLink || canRemoveSharedLink) && !submitting;
+        const toggleDescriptionId = uniqueId('toggle-description');
```
QingyuChai, line 461: I'm worried that this will create a new unique ID on every render. Any chance we can use something like:
`const [toggleDescriptionId] = React.useState(uniqueId('toggle-description'))`

src/features/unified-share-modal/SharedLinkSection.js
```diff
@@ -502,6 +505,9 @@ class SharedLinkSection extends React.Component<Props, State> {
         );
 
         if (!submitting) {
+            const getText = text => {
+                <span id={toggleDescriptionId}>{text}</span>;
+            };
```
QingyuChai, line 510: Just wondering, why are we wrapping the text with a span? I think according to Tooltip's implementation, the `text` prop should actually be a string

QingyuChai: nevermind, it seems to be currently set as type ReactNode, but it is used as a string here:
https://github.com/box/box-ui-elements/blob/master/src/components/tooltip/Tooltip.tsx#L254

I think we prefer to use strings though, so we don't have duplicate aria labels (aria-label and aria-desribedBy)

https://github.com/box/box-ui-elements/blob/master/src/components/tooltip/Tooltip.tsx#L289

src/features/unified-share-modal/SharedLinkSection.js
megansmith-box: This is not a functional component, i referenced other examples and moved it to the top!

QingyuChai (approved):
lgtm!

src/components/tooltip/Tooltip.tsx
```diff
@@ -285,7 +288,7 @@ class Tooltip extends React.Component<TooltipProps, State> {
         }
 
         if (showTooltip) {
-            if (!isLabelMatchingTooltipText || childAriaLabel === undefined) {
+            if (!id && (!isLabelMatchingTooltipText || childAriaLabel === undefined)) {
```
tjuanitas, line 291: why do we need this?

tjuanitas, line 291: it looks like the original issue might be caused because of the wrapping `div` around `Toggle`: https://github.com/box/box-ui-elements/blob/cbe6c855a8c396000624de2d41c3e55643ccfefb/src/features/unified-share-modal/SharedLinkSection.js#L493-L494
it's not adding `aria-describedby` to the correct component

it's possible that removing the wrapping div and moving the appropriate styles from `share-toggle-container` to `toggle-container` would fix the issue

megansmith-box: To avoid duplicate aria-describeby labels

src/features/unified-share-modal/UnifiedShareModal.scss
```diff
@@ -255,7 +255,7 @@
         justify-content: space-between;
         margin-top: 10px;
 
-        .share-toggle-container {
+        .toggle-container {
```
tjuanitas, line 258: the `.toggle-container` a few lines below (line 263) is not needed

src/features/unified-share-modal/UnifiedShareModal.scss
```diff
@@ -255,15 +255,11 @@
...
             flex: 0 1 80%;
```
tjuanitas, line 259: these flex styles still work as expected when moving to `toggle-container` right?

tjuanitas (approved):
we should ask QA to test this toggle still works the same way but looks great!

QingyuChai (approved):

src/features/unified-share-modal/UnifiedShareModal.scss
megansmith-box: yup!

PR #3217, opened 2023-01-23
fix(usm): simplified sharing tooltip at high zoom

jpan-box:
Make simplified sharing visible at high zoom
by moving from middle-left to bottom-center

Should not merge until we are confident tab order may flow between the tooltip and the primary modal.

QUESTION: The updated snapshots reflect the intended change, HOWEVER, I am wondering why the Tooltip component still shows up in some of these snapshots where I expect the tooltip to NOT render at all, i.e. when isShown prop is NOT true?

Screenshot 2023-01-19 at 4 11 59 PM

Screenshot 2023-01-19 at 4 12 10 PM

### DIFF
--- a/src/features/unified-share-modal/UnifiedShareForm.js
+++ b/src/features/unified-share-modal/UnifiedShareForm.js
@@ -482,7 +482,7 @@ class UnifiedShareForm extends React.Component<USFProps, State> {
             className: 'usm-ftux-tooltip',
             // don't want ftux tooltip to show if the recommended sharing tooltip callout is showing
             isShown: !recommendedSharingTooltipCalloutName && shouldRenderFTUXTooltip && showCalloutForUser,
-            position: 'middle-left',
+            position: 'bottom-center',
             showCloseButton: true,
             text: ftuxTooltipText,
             theme: 'callout',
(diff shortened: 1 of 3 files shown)


src/features/unified-share-modal/UnifiedShareForm.js
```diff
@@ -482,7 +482,7 @@ class UnifiedShareForm extends React.Component<USFProps, State> {
             className: 'usm-ftux-tooltip',
             // don't want ftux tooltip to show if the recommended sharing tooltip callout is showing
             isShown: !recommendedSharingTooltipCalloutName && shouldRenderFTUXTooltip && showCalloutForUser,
-            position: 'middle-left',
+            position: 'bottom-center',
```
tjuanitas, line 485: it's possible that with a long tooltip text, the problem is now moved to the bottom of the modal i.e. the tooltip becomes off-screen at the bottom.

might not be reproducible with english but could be with other languages.

tjuanitas: but we don't really have a better solution at the moment besides removing it altogether

src/features/unified-share-modal/__tests__/UnifiedShareForm.test.js
```diff
@@ -106,6 +106,7 @@ describe('features/unified-share-modal/UnifiedShareForm', () => {
                 shouldRenderFTUXTooltip: true,
                 showCalloutForUser: true,
             });
+            expect(wrapper.find('.usm-ftux-tooltip').prop('position')).toBe('bottom-center');
```
tjuanitas, line 109: not sure the best practice on this but I've generally seen us use the component name unless we need the extra specificity of the class

tjuanitas (approved):
lgtm

jpan-box:
Closed / Won't Fix

PR #3218, opened 2023-01-23
fix(react-virtualized): Remove role for Table_Grid_innerScrollContainer

tjuanitas:
Removes the default role "row" for Grid_innerScrollContainer when used inside of Table. The innerScrollContainer contains a group of rows and is not a row itself.

The solution follows the approach we're using for the main webapp.


QingyuChai (approved):
l g t m

greg-in-a-box (approved):

PR #3224, opened 2023-01-26
fix(itemlist): Remove header section when list is empty

SkyisAakash:

### DIFF
--- a/src/features/content-explorer/item-list/ItemList.js
+++ b/src/features/content-explorer/item-list/ItemList.js
@@ -210,6 +210,15 @@ const ItemList = ({
         };
     }
 
+    const rowCount = items.length;
+    const headerProps =
+        rowCount > 0
+            ? {
+                  headerHeight,
+                  headerRowRenderer: headerRenderer,
+              }
+            : {};
+
     return (
         <div
             className={classNames('content-explorer-item-list table', className, {
@@ -221,8 +230,6 @@ const ItemList = ({
                 headerClassName="table-header-item"
                 width={width}
                 height={height}
-                headerHeight={headerHeight}
-                headerRowRenderer={headerRenderer}
                 rowHeight={rowHeight}
                 rowCount={items.length}
                 onRowClick={onItemClick}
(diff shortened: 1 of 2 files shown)


SkyisAakash:
Before:
Screenshot 2023-01-26 at 9 33 11 AM

SkyisAakash:
After:
Screenshot 2023-01-26 at 9 43 02 AM

src/features/content-explorer/item-list/ItemList.js
```diff
@@ -210,6 +210,15 @@ const ItemList = ({
...
     }
 
+    const rowCount = items.length;
+    const headerProps =
+        rowCount > 0
+            ? {
+                  headerHeight,
+                  headerRowRenderer: headerRenderer,
+              }
+            : {};
```
tjuanitas, line 220: maybe this could utilize the `tableProps`:
```suggestion
    if (items.length > 0) {
        tableProps.headerHeight = headerHeight;
        tableProps.headerRowRenderer = headerRenderer;
    }
```

src/features/content-explorer/item-list/ItemList.js
```diff
@@ -231,6 +238,7 @@ const ItemList = ({
                 rowRenderer={renderRow}
                 noRowsRenderer={noItemsRenderer}
```
tjuanitas, line 239: have you tried using this `noRowsRenderer` prop?

SkyisAakash: We already use it for contentExplorer. https://github.com/box/box-ui-elements/blob/master/src/features/content-explorer/content-explorer/ContentExplorer.js#L529
But as seen in screenshot, header is displayed on top of contents of `noRowsRenderer`

tjuanitas: ahh gotcha!

maybe we could add a check to if the `if` statement that if `noRowsRenderer` is not passed, then display the table as normal. something like 

```
if (!noItemsRenderer || items.length > 0) {
    [...]
}
```
that way usages of `ItemList` that don't have a custom `noRowsRenderer` will continue to display the headers as normal

SkyisAakash: Makes sense yes, added that

tjuanitas (approved):
lgtm!

menaczar (approved):

apieceofbart (approved):

PR #3226, opened 2023-01-31
fix(usm): tooltip should be dialog

megansmith-box:

### DIFF
--- a/src/components/tooltip/Tooltip.tsx
+++ b/src/components/tooltip/Tooltip.tsx
@@ -84,6 +84,8 @@ export type DefaultTooltipProps = {
 export type TooltipProps = {
     /** Sets aria-hidden attribute on tooltip */
     ariaHidden?: boolean;
+    /** Sets aria-label attribute on tooltip */
+    ariaLabel?: string;
     /** An HTML element to append the tooltip container into (otherwise appends to body) */
     bodyElement?: HTMLElement;
     /** A React element to put the tooltip on */
@@ -244,6 +246,7 @@ class Tooltip extends React.Component<TooltipProps, State> {
     render() {
         const {
             ariaHidden,
+            ariaLabel,
             bodyElement,
             children,
             className,
@@ -377,6 +380,7 @@ class Tooltip extends React.Component<TooltipProps, State> {
                     role={role}
                     aria-live="polite"
                     aria-hidden={ariaHidden || isLabelMatchingTooltipText}
+                    aria-labelledby={ariaLabel}
                     data-testid="bdl-Tooltip"
                 >
                     {tooltipInner}
@@ -386,6 +390,7 @@ class Tooltip extends React.Component<TooltipProps, State> {
             <div
                 aria-live="polite"
                 aria-hidden={ariaHidden || isLabelMatchingTooltipText}
+                aria-labelledby={ariaLabel}
                 className={classes}
                 data-testid="bdl-Tooltip"
                 id={this.tooltipID}
--- a/src/components/tooltip/Tooltip.tsx
+++ b/src/components/tooltip/Tooltip.tsx
@@ -5,6 +5,8 @@ import getProp from 'lodash/get';
 import TetherComponent from 'react-tether';
 
 import TetherPosition from '../../common/tether-positions';
+// @ts-ignore flow import
+import FocusTrap from '../focus-trap';
 import CloseButton from './CloseButton';
 
 import './Tooltip.scss';
@@ -88,6 +90,8 @@ export type TooltipProps = {
     children: React.ReactChild;
     /** A CSS class for the tooltip */
     className?: string;
+    /** Allow the component to maintain focus until escaped */
+    isFocusTrapped?: boolean;
     /** Forces the tooltip to be shown or hidden (useful for errors) */
     isShown?: boolean;
     /** Whether to add tabindex=0.  Defaults to `true` */
@@ -132,12 +136,14 @@ class Tooltip extends React.Component<TooltipProps, State> {
     }
 
     componentDidUpdate(prevProps: TooltipProps, prevState: State) {
-        const isControlled = this.isControlled();
-
-        // Reset wasClosedByUser state when isShown transitions from false to true
-        if (isControlled) {
+        if (this.isControlled()) {
             if (!prevProps.isShown && this.props.isShown) {
+                // Reset wasClosedByUser state when isShown transitions from false to true
                 this.setState({ wasClosedByUser: false });
+                document.addEventListener('keydown', this.handleKeyDown, true);
+            }
+            if (!prevState.wasClosedByUser && this.state.wasClosedByUser) {
+                document.removeEventListener('keydown', this.handleKeyDown, true);
             }
         } else {
             if (!prevState.isShown && this.state.isShown) {
@@ -215,7 +221,11 @@ class Tooltip extends React.Component<TooltipProps, State> {
     handleKeyDown = (event: KeyboardEvent) => {
         if (event.key === 'Escape') {
             event.stopPropagation();
-            this.setState({ isShown: false });
+            if (this.isControlled()) {
+                this.setState({ isShown: false, wasClosedByUser: true });
+            } else {
+                this.setState({ isShown: false });
+            }
         }
         this.fireChildEvent('onKeyDown', event);
     };
@@ -240,6 +250,7 @@ class Tooltip extends React.Component<TooltipProps, State> {
             constrainToScrollParent,
             constrainToWindow,
             isDisabled,
+            isFocusTrapped,
             isTabbable = true,
             offset,
             position = TooltipPosition.TOP_CENTER,
@@ -344,6 +355,12 @@ class Tooltip extends React.Component<TooltipProps, State> {
                 {withCloseButton && <CloseButton onClick={this.closeTooltip} />}
             </>
         );
+        let role;
+        if (isFocusTrapped) {
+            role = 'dialog';
+        } else if (theme !== TooltipTheme.ERROR) {
+            role = 'tooltip';
+        }
 
         const tooltip = stopBubble ? (
             <div
@@ -357,7 +374,7 @@ class Tooltip extends React.Component<TooltipProps, State> {
                 role="presentation"
             >
                 <div
-                    role={theme === TooltipTheme.ERROR ? undefined : 'tooltip'}
+                    role={role}
                     aria-live="polite"
                     aria-hidden={ariaHidden || isLabelMatchingTooltipText}
                     data-testid="bdl-Tooltip"
@@ -374,7 +391,7 @@ class Tooltip extends React.Component<TooltipProps, State> {
                 id={this.tooltipID}
                 onMouseEnter={this.handleMouseEnter}
                 onMouseLeave={this.handleMouseLeave}
-                role={theme === TooltipTheme.ERROR ? undefined : 'tooltip'}
+                role={role}
             >
                 {tooltipInner}
             </div>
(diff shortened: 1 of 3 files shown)


src/components/tooltip/Tooltip.tsx
```diff
@@ -88,6 +88,8 @@ export type TooltipProps = {
     children: React.ReactChild;
     /** A CSS class for the tooltip */
     className?: string;
+    /** Component maintain focus within the tooltip until escaped */
+    focusTrapComponent?: React.ElementType;
```
QingyuChai, line 92: not sure I understand this part, is the focused trapped within the tooltip? Or is it trapped within this component called `focusTrapComponent`?

src/components/tooltip/Tooltip.tsx
```diff
@@ -138,16 +140,15 @@ class Tooltip extends React.Component<TooltipProps, State> {
         if (isControlled) {
             if (!prevProps.isShown && this.props.isShown) {
                 this.setState({ wasClosedByUser: false });
-            }
-        } else {
-            if (!prevState.isShown && this.state.isShown) {
-                // capture event so that tooltip closes before any other floating components that can be closed by
-                // "Escape" key(e.g. Modal, Menu, etc.)
```
QingyuChai, line 145: could we leave this comment?

src/components/tooltip/Tooltip.tsx
```diff
@@ -380,6 +390,25 @@ class Tooltip extends React.Component<TooltipProps, State> {
             </div>
         );
 
+        if (shouldTrapFocus) {
+            tooltip = (
```
tjuanitas, line 394: it looks like `shouldTrapFocus` cannot be used with `stopBubble`, is that okay? do we need to wrap the existing value?

```
if (shouldTrapFocus) {
    tooltip = (
        <FocusTrap aria-modal shouldDefaultFocus>{tooltip}</FocusTrap>
    );
}
```

src/components/tooltip/Tooltip.tsx
```diff
@@ -5,6 +5,8 @@ import getProp from 'lodash/get';
 import TetherComponent from 'react-tether';
 
 import TetherPosition from '../../common/tether-positions';
+// @ts-ignore flow import
+import FocusTrap from '../focus-trap';
```
tjuanitas, line 9: should we limit this to GuideTooltip or could we see a use case for all tooltips?

are all GuideTooltips considered a dialog?

src/components/tooltip/Tooltip.tsx
```diff
@@ -96,6 +98,8 @@ export type TooltipProps = {
     offset?: string;
     /** Function called if the user manually dismisses the tooltip - only applies if showCloseButton is true */
     onDismiss?: () => void;
+    /** allow the component to maintain focus until escaped */
+    shouldTrapFocus?: boolean;
```
tjuanitas, line 102: is `isFocusTrapped` appropriate? I think it would be good to avoid `should`

src/components/tooltip/Tooltip.tsx
```diff
@@ -96,6 +98,8 @@ export type TooltipProps = {
     offset?: string;
     /** Function called if the user manually dismisses the tooltip - only applies if showCloseButton is true */
     onDismiss?: () => void;
+    /** allow the component to maintain focus until escaped */
```
tjuanitas, line 101: smallest of all smalls
```suggestion
    /** Allow the component to maintain focus until escaped */
```
because the other comments are capitalized

src/components/tooltip/Tooltip.tsx
megansmith-box: I don't think it's limited to GuideTooltips, the tooltip that the issue arose on was not a Guide Tooltip. I don't know if all GuideTooltips are considered dialogs, but I feel like it makes sense to enable this FocusTrap on a case by case basis, rather than enable it for all. What do you think?

src/components/tooltip/Tooltip.tsx
megansmith-box: yes, that's a good point, I will modify

src/components/tooltip/Tooltip.tsx
tjuanitas: oh you're right, i got my tooltips mixed up and thought all FTUX tooltips were GuideTooltips

tjuanitas: although I think it's a little weird that USM isn't using the GuideTooltip component

QingyuChai (dismissed):
lgtm! I think ideally, we'd want to test this feature as well, but I'm not sure if that's possible here

src/components/tooltip/Tooltip.tsx
```diff
@@ -138,16 +142,17 @@ class Tooltip extends React.Component<TooltipProps, State> {
...
-            if (prevState.isShown && !this.state.isShown) {
-                document.removeEventListener('keydown', this.handleKeyDown, true);
-            }
+        }
+
+        if (!prevState.isShown && this.state.isShown) {
+            // capture event so that tooltip closes before any other floating components that can be closed by
+            // "Escape" key(e.g. Modal, Menu, etc.)
+            document.addEventListener('keydown', this.handleKeyDown, true);
+        }
```
tjuanitas, line 153: moving this outside of the `else` would add a duplicate event listener if it also enters the above `if` block

src/components/tooltip/Tooltip.tsx
```diff
@@ -96,6 +98,8 @@ export type TooltipProps = {
     offset?: string;
     /** Function called if the user manually dismisses the tooltip - only applies if showCloseButton is true */
     onDismiss?: () => void;
+    /** Allow the component to maintain focus until escaped */
+    isFocusTrapped?: boolean;
```
tjuanitas, line 102: nit: alphasort

src/components/tooltip/Tooltip.tsx
```diff
@@ -213,9 +218,15 @@ class Tooltip extends React.Component<TooltipProps, State> {
     };
 
     handleKeyDown = (event: KeyboardEvent) => {
+        const isControlled = this.isControlled();
+
         if (event.key === 'Escape') {
             event.stopPropagation();
-            this.setState({ isShown: false });
+            if (isControlled) {
```
tjuanitas, line 225: could call `this.Controlled()` directly so the variable isn't needed

src/components/tooltip/Tooltip.tsx
```diff
@@ -243,6 +254,7 @@ class Tooltip extends React.Component<TooltipProps, State> {
             isTabbable = true,
             offset,
             position = TooltipPosition.TOP_CENTER,
+            isFocusTrapped,
```
tjuanitas, line 257: nit: alphasort

src/components/tooltip/Tooltip.tsx
```diff
@@ -374,16 +387,24 @@ class Tooltip extends React.Component<TooltipProps, State> {
...
+            </FocusTrap>
+        ) : (
+            tooltip
+        );
+
         return (
             <TetherComponent ref={this.tetherRef} {...tetherProps}>
                 {React.cloneElement(React.Children.only(children) as React.ReactElement, componentProps)}
-                {showTooltip && tooltip}
+                {showTooltip && focusedTooltip}
```
tjuanitas, line 407: might be okay to inline here assuming it's still readable
```
{showTooltip &&
    (isFocusTrapped ? (
        <FocusTrap aria-modal shouldDefaultFocus>
            {tooltip}
        </FocusTrap>
    ) : (
        tooltip
    ))}
```

src/components/tooltip/Tooltip.tsx
```diff
@@ -357,7 +370,7 @@ class Tooltip extends React.Component<TooltipProps, State> {
                 role="presentation"
             >
                 <div
-                    role={theme === TooltipTheme.ERROR ? undefined : 'tooltip'}
+                    role={theme === TooltipTheme.ERROR ? undefined : role}
```
tjuanitas, line 373: might be able to combine the `theme === TooltipTheme.ERROR` part into the new `role` variable. although could also keep it like this if it's easier to understand

megansmith-box: I was having a hard time finding a cleaner way to do this, trying to avoid nested ternary and if statements and lets... do you have any suggestion to make it cleaner?

src/components/tooltip/Tooltip.tsx
megansmith-box: I believe GuideTooltips are usually multi-step tooltips

src/components/tooltip/Tooltip.tsx
tjuanitas: the cleanest way i could think of was
```
let role;
if (isFocusedTrapped) {
    role = 'dialog';
}
else if (theme !== TooltipTheme.ERROR) {
    role = 'tooltip';
}
```
but i don't know if this is an improvement

src/components/tooltip/Tooltip.tsx
```diff
@@ -138,16 +142,16 @@ class Tooltip extends React.Component<TooltipProps, State> {
...
-            if (prevState.isShown && !this.state.isShown) {
-                document.removeEventListener('keydown', this.handleKeyDown, true);
-            }
+        } else if (!prevState.isShown && this.state.isShown) {
+            // capture event so that tooltip closes before any other floating components that can be closed by
+            // "Escape" key(e.g. Modal, Menu, etc.)
+            document.addEventListener('keydown', this.handleKeyDown, true);
+        }
+
+        if ((prevState.isShown && !this.state.isShown) || (!prevState.wasClosedByUser && this.state.wasClosedByUser)) {
```
tjuanitas, line 153: I might be misunderstanding but I think `prevState.isShown && !this.state.isShown` might also have a similar edge case where this `if` can be entered for a controlled tooltip because the `handleMouseEnter` and `handleMouseLeave` methods: https://github.com/box/box-ui-elements/blob/a603d18f75872864b3b5e822c0d9228f79c5c3fd/src/components/tooltip/Tooltip.tsx#L199-L200

i.e. the tooltip is visible through the `isShown` prop but hovering the mouse could update the state to remove the event listener. not 100% sure though

megansmith-box: You are right! Thank you for clarifying, I wasn't understanding completely before.

src/components/tooltip/Tooltip.tsx
megansmith-box: Yeah, I'm indifferent! I suppose it's nice it's all in one place

QingyuChai (approved):

src/components/tooltip/Tooltip.tsx
```diff
@@ -215,7 +221,11 @@ class Tooltip extends React.Component<TooltipProps, State> {
     handleKeyDown = (event: KeyboardEvent) => {
         if (event.key === 'Escape') {
             event.stopPropagation();
-            this.setState({ isShown: false });
+            if (this.isControlled()) {
+                this.setState({ wasClosedByUser: true });
+            } else {
+                this.setState({ isShown: false });
+            }
```
tjuanitas, line 228: thoughts on keeping `this.setState({ isShown: false });` to always set?

even though we don't use the internal state for controlled inputs, the state is still kept in sync through the other methods: `handleMouseEnter`, `handleMouseLeave`, etc. so it might be useful in the future that it's all still in sync.

```
            this.setState({ isShown: false });
            if (this.isControlled()) {
                this.setState({ wasClosedByUser: true });
            }
```

megansmith-box: That sounds good

tjuanitas (approved):

QingyuChai (approved):
l g t m

PR #3227, opened 2023-02-01
fix(content-explorer): Add svg titles to icons

benjamin-shen:
The `title` prop was undefined before this PR. When it is defined, the `aria-labelledby` attribute is added to the `svg`, which points to a `<title>` child component.

image

image

## Reference:
- QuickSearchItem feature (very similar `title` code, not refactored): https://github.com/box/box-ui-elements/blob/b7ecaabe124277aacc7a8c352e9daefb8af3e157/src/features/quick-search/QuickSearchItem.js#L76-L94
- ItemList feature (uses `ItemListIcon`, which was refactored in this PR): https://github.com/box/box-ui-elements/blob/f52ad0e513f7253e3ba4b6276e250f7126046cd1/src/features/content-explorer/item-list/ItemList.js#L52-L57


src/elements/common/item/iconCellRenderer.js
```diff
@@ -5,28 +5,48 @@
...
+                />
+            );
         case TYPE_WEBLINK:
-            return <BookmarkIcon height={dimension} width={dimension} />;
+            title = <FormattedMessage {...messages.bookmark} />;
+            return <BookmarkIcon height={dimension} width={dimension} title={title} />;
         default:
-            throw new Error('Unsupported item type!');
+            title = <FormattedMessage {...messages.file} />;
+            return <FileIcon dimension={dimension} title={title} />;
```
benjamin-shen, line 46: Inspired by `ItemListIcon.js`

src/elements/common/messages.js
```diff
@@ -980,6 +980,31 @@ const messages = defineMessages({
...
+        defaultMessage: 'Bookmark',
+    },
+    collaboratedFolder: {
+        id: 'be.collaboratedFolder',
+        description: 'Icon title for a Box item of type folder that has collaborators',
+        defaultMessage: 'Collaborated Folder',
+    },
+    externalFolder: {
+        id: 'be.externalFolder',
+        description: "Icon title for a Box item of type folder that has collaborators outside of the user's enterprise",
```
megansmith-box, line 995:
```suggestion
        description: 'Icon title for a Box item of type folder that has collaborators outside of the user's enterprise',
```

src/common/constants.js
```diff
@@ -1,7 +1,9 @@
 // @flow
-const ITEM_TYPE_FILE: 'file' = 'file';
-const ITEM_TYPE_FOLDER: 'folder' = 'folder';
-const ITEM_TYPE_WEBLINK: 'web_link' = 'web_link';
+import { TYPE_FILE, TYPE_FOLDER, TYPE_WEBLINK } from '../constants';
+
+const ITEM_TYPE_FILE = TYPE_FILE;
```
megansmith-box, line 4: Can we always directly import from src/constants or what is the reason for re-exporting the constants here?

src/elements/common/item/iconCellRenderer.js
```diff
@@ -5,28 +5,48 @@
...
         default:
-            throw new Error('Unsupported item type!');
+            title = intl.formatMessage(messages.file);
+            return <FileIcon dimension={dimension} title={title} />;
     }
 }
 
-export default (dimension: number = 32): Function => ({ rowData }: { rowData: BoxItem }) => (
-    <div className="be-item-icon">{getIcon(dimension, rowData)}</div>
+export default (intl: IntlShape, dimension: number = 32): Function => ({ rowData }: { rowData: BoxItem }) => (
```
megansmith-box, line 50: can we wrap this method with `injectIntl` to avoid passing as param?

src/common/constants.js
benjamin-shen: I don't think there is an intentional reason why there are redundant constants here, the only difference is the variable names. I can refactor them in this PR

src/elements/common/item/iconCellRenderer.js
benjamin-shen: Would that work on a non-JSX element? I think intl is being used as a function argument rather than a prop, but I can try it out

src/elements/common/messages.js
tjuanitas: I think you'll need to escape the apostrophe

src/common/constants.js
tjuanitas: I think @greg-in-a-box and I looked into this once and it was related to types? might want to double check if we need to keep it the way it was

src/features/quick-search/QuickSearchItem.js
```diff
@@ -4,6 +4,7 @@ import classNames from 'classnames';
 import { injectIntl, FormattedMessage } from 'react-intl';
 import { RecordOf } from 'immutable';
 
+import { TYPE_FILE, TYPE_FOLDER, TYPE_WEBLINK } from '../../constants';
```
tjuanitas, line 7: could you move this to right above the `messages` import? we usually place the `constants` import near the bottom

src/features/content-explorer/messages.js
```diff
@@ -1,11 +1,6 @@
 import { defineMessages } from 'react-intl';
 
 const messages = defineMessages({
-    bookmark: {
```
tjuanitas, line 4: are we moving these because we plan to use them for ContentPicker?

src/features/content-explorer/item-list/ItemListName.js
```diff
@@ -2,15 +2,15 @@ import PropTypes from 'prop-types';
 import React from 'react';
 
 import PlainButton from '../../../components/plain-button';
+import { TYPE_FOLDER } from '../../../constants';
```
tjuanitas, line 5: could we move this out of this section? we don't follow this too well but it'd be good to group similar imports together. e.g. grouping them like:
```
// base/root imports
// external libs
// internal libs
// components
// types
// constants, messages
// styles
```
some of the sections could be swapped around but that's how I've generally seen them

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -11,9 +11,9 @@ import ContentExplorerSelectAll from './ContentExplorerSelectAll';
 import ItemList from '../item-list';
 import { ContentExplorerModePropType, FoldersPathPropType, ItemsPropType } from '../prop-types';
 import ContentExplorerModes from '../modes';
-import ItemTypes from '../item-types';
 
 import './ContentExplorer.scss';
+import { TYPE_FOLDER } from '../../../constants';
```
tjuanitas, line 16: similar here: I would put this above the styles import

src/elements/common/item/iconCellRenderer.js
tjuanitas: yeah I agree with @megansmith-box. some of the changes like in `ItemList.js` felt odd

src/common/constants.js
greg-in-a-box: https://stackoverflow.com/a/64145916/18705593

greg-in-a-box: I cant remember what was going on here, I think there was a type issue but that was because we were trying to convert to TS probably

src/features/content-explorer/item-list/ItemListName.js
benjamin-shen: Yup. I think most of these were automatic imports, there's probably a way to configure it somehow 🤔

src/elements/common/messages.js
benjamin-shen: Prettier seems to complain 😿 
image

src/features/content-explorer/messages.js
benjamin-shen: I moved them out since they're not used in `features/` anymore (after the `ItemListIcon.js` refactor). I also didn't see ContentPicker in `features/`

tjuanitas: ah okay i see what's happening now. I thought this was happening in the elements version. I don't think we should be importing things from `/elements` into `/features`. we probably should keep these separate and almost think of them as separate packages

tjuanitas: @greg-in-a-box thoughts on importing things from the `/elements` folder? my gut is telling me we should avoid that so things do not become dependent on `/elements`

src/common/constants.js
benjamin-shen: I reverted this specific file just to be safe. I looked at some of the git blames and it looks like it might cause flow-related issues in consuming libraries

src/features/content-explorer/messages.js
benjamin-shen: Yeah, I'm with you. I decided to go with it anyway since I saw it done in a couple other places:
image

And it's a tradeoff with duplicated code. If we want to avoid importing, then I'll go ahead and undo some of the refactor in `ItemListIcon.js`

src/elements/common/item/iconCellRenderer.js
```diff
@@ -5,28 +5,52 @@
...
+            return <FileIcon dimension={dimension} title={title} />;
     }
 }
 
-export default (dimension: number = 32): Function => ({ rowData }: { rowData: BoxItem }) => (
-    <div className="be-item-icon">{getIcon(dimension, rowData)}</div>
-);
+export default (dimension: number = 32) =>
+    injectIntl(({ intl, rowData }: { intl: IntlShape, rowData: BoxItem }) => (
+        <div className="be-item-icon">{getIcon(intl, rowData, dimension)}</div>
```
megansmith-box, line 55: Is it possible to inject getIcon with intl?

src/features/content-explorer/messages.js
tjuanitas: I think you can keep the changes. I thought it was a lot more (like all the messages as well) but it seems it's only for `getIcon` inside of `ItemListIcon`

src/elements/common/messages.js
megansmith-box: Ah ok!

src/elements/content-picker/ItemList.js
```diff
@@ -175,4 +176,4 @@ const ItemList = ({
     );
 };
 
-export default ItemList;
```
tjuanitas, line 178: is this still needed?

src/elements/common/item/iconCellRenderer.js
benjamin-shen: I think so. I turned `getIcon` into an explicit component (renamed to `IconCell`) instead of a function.

There is a slight difference, which is that the `ItemListIcon` component hierarchy now includes the `IconCell` component. So in the unit tests, `shallow(ItemListicon)` would see `IconCell` instead of `FileIcon`/`FolderIcon`/`BookmarkIcon`.

src/elements/content-picker/ItemList.js
benjamin-shen: Nope, whoops missed it 🤣

src/components/breadcrumb/Breadcrumb.js
```diff
@@ -84,5 +84,5 @@ const Breadcrumb = ({
     );
 };
 
-export { Breadcrumb as BreadcrumbCore };
+export { Breadcrumb as BreadcrumbBase };
```
megansmith-box, line 87: Why this change?

src/elements/common/item/__tests__/IconCell.test.js
```diff
@@ -0,0 +1,94 @@
...
+        });
+
+        test('should render correct bookmark icon', () => {
+            const rowData = { type: 'web_link' };
+            const wrapper = getWrapper({ rowData });
+
+            expect(wrapper.name()).toEqual('BookmarkIcon');
+            expect(wrapper.prop('title')).toEqual('Bookmark');
+
+            expect(wrapper).toMatchSnapshot();
```
megansmith-box, line 91: Do you think these snapshots are necessary?

src/features/content-explorer/item-list/__tests__/ItemListIcon.test.js
```diff
@@ -33,33 +37,37 @@ describe('features/content-explorer/item-list/ItemListIcon', () => {
...
         test('should render correct bookmark icon', () => {
-            const wrapper = renderComponent({ type: 'web_link' });
+            const rowData = { type: 'web_link' };
+            const wrapper = renderComponent(rowData);
 
-            expect(wrapper.find('BookmarkIcon').length).toBe(1);
-            expect(wrapper.prop('title')).toBeTruthy();
+            expect(wrapper.find('IconCell').length).toBe(1);
+            expect(wrapper.prop('rowData')).toEqual(expect.objectContaining(rowData));
+            expect(wrapper).toMatchSnapshot();
```
megansmith-box, line 70: I don't think these snapshots test anything different than directly checking the props?

src/features/content-explorer/item-list/ItemListIcon.js
```diff
@@ -1,40 +1,12 @@
...
-        }
-        case ItemTypes.FILE:
-            return <FileIcon extension={extension} title={<FormattedMessage {...messages.file} />} />;
-        case ItemTypes.BOOKMARK:
-            return <BookmarkIcon title={<FormattedMessage {...messages.bookmark} />} />;
-        default:
-            // Use generic file icon as fallback
-            return <FileIcon title={<FormattedMessage {...messages.file} />} />;
-    }
+const ItemListIcon = ({ type, extension, hasCollaborations = false, isExternallyOwned = false }) => {
```
megansmith-box, line 7: Can we directly replace ItemListIcon with IconCell in ItemList?

src/elements/common/item/IconCell.js
```diff
@@ -0,0 +1,55 @@
+/**
```
tjuanitas, line 1: let's write this file and the test in TS if possible

src/components/breadcrumb/Breadcrumb.js
benjamin-shen: Just a cleanup. `ComponentBase` is used frequently throughout the codebase, whereas `ComponentCore` is only used here. I found it when I was searching for another component to reference

src/features/content-explorer/item-list/ItemListIcon.js
benjamin-shen: I double checked and it doesn't seem like it's being used internally, so I guess we could. But I could also see us adding props to `IconCell` so that it's more versatile (like how it's done in EUA).

src/elements/common/item/IconCell.js
benjamin-shen: I think it would be a larger refactor outside the scope of this PR, since it has dependencies on files that are not in typescript.

src/features/content-explorer/item-list/__tests__/ItemListIcon.test.js
benjamin-shen: You're totally right. I added them to match the existing snapshot test, but I think we can safely remove all of them.

src/elements/common/item/__tests__/IconCell.test.js
benjamin-shen: They are probably redundant

src/elements/common/item/IconCell.js
megansmith-box: Would it be easily resolved by using the comment Trevor told me about https://github.com/box/box-ui-elements/blob/004407dde579ac5d02f0c040ee0541c4b8189039/src/components/time-input/TimeInput.tsx#L7?

src/features/content-explorer/item-list/ItemListIcon.js
megansmith-box: That makes sense 👍

src/features/content-explorer/item-list/ItemListIcon.js
```diff
@@ -1,40 +1,12 @@
...
+    const rowData = { type, extension, has_collaborations: hasCollaborations, is_externally_owned: isExternallyOwned };
```
megansmith-box, line 8: is there a reason this one is snake case? can we camel case it?

tjuanitas: I think the snake case comes from an API request

src/features/content-explorer/item-list/__tests__/ItemListIcon.test.js
```diff
@@ -33,33 +37,35 @@ describe('features/content-explorer/item-list/ItemListIcon', () => {
...
 
         test('should render correct file icon', () => {
-            const extension = 'boxnote';
-            const wrapper = renderComponent({ type: 'file', extension });
+            const rowData = { type: 'file', extension: 'boxnote' };
+            const wrapper = renderComponent(rowData);
 
-            expect(wrapper.find('FileIcon').length).toBe(1);
-            expect(wrapper.prop('extension')).toEqual(extension);
-            expect(wrapper.prop('title')).toBeTruthy();
```
tjuanitas, line 55: do we not need to check the `title` anymore?

megansmith-box (approved):
looks good to me

src/elements/common/item/IconCell.js
benjamin-shen: Cool! The ts-ignore works for most of the issues.

But unfortunately the component also depends on a flow type `BoxItem`. That would need to be migrated to TS first.

src/features/content-explorer/item-list/ItemListIcon.js
tjuanitas: oh wait this already returned and transformed from snake case to camelcase. @megansmith-box is right, why do we transform it back to snake?

tjuanitas: nvm again it looks like it's to satisfy the BoxItem type

tjuanitas (dismissed):

src/elements/common/item/__tests__/IconCell.test.js
```diff
@@ -0,0 +1,88 @@
...
+
+describe('elements/common/item/IconCell', () => {
+    const getWrapper = props => shallow(<IconCell intl={intl} {...props} />);
+
+    describe('render()', () => {
+        test('should render default file icon', () => {
+            const rowData = { type: undefined };
+            const wrapper = getWrapper({ rowData });
+
+            expect(wrapper.name()).toEqual('FileIcon');
```
tjuanitas, line 17: small thing but these `toEqual`s could be `toBe`s

src/features/content-explorer/item-list/__tests__/ItemListIcon.test.js
benjamin-shen: It no longer has the `title` prop, only `rowData`. The `title` logic is now tested in [IconCell.test.ts](https://github.com/box/box-ui-elements/pull/3227/files/387dcd50fab1a887f2e2e72a1ade7c0b4130e37e#diff-b96c51318ab384cd29d726558833527aae42433a1620d61e045ec91763d1c170).

tjuanitas (approved):

megansmith-box (approved):

src/elements/common/item/iconCellRenderer.js
```diff
@@ -1,32 +1,15 @@
 /**
  * @flow
- * @file Function to render the icon table cell
```
JChan106, line 3: Do we need to add a different description for the file now?

JChan106 (approved):

src/elements/common/item/iconCellRenderer.js
tjuanitas: @benjamin-shen remember to address this if needed

benjamin-shen: Ooh thanks, good catch!! I added it back here: https://github.com/box/box-ui-elements/pull/3238, could you approve @JChan106 ?

PR #3228, opened 2023-02-02
fix: Improve color contrast for placeholder text

tjuanitas:
The placeholder text color of inputs/textareas does not meet a 4.5:1 color contrast ratio. The issue affects all inputs but the design team recommended only updating the specific usages in question since it would be a global change. The other usages would be fixed with the migration to the UI toolkit.

Design and accessibility team recommended using #767676 - $bdl-gray-62

**Before - Content Explorer**
Screen Shot 2023-02-01 at 5 44 03 PM

**After - Content Explorer**
Screen Shot 2023-02-01 at 5 43 51 PM

**Before - USM**
Screen Shot 2023-02-01 at 5 46 32 PM

**After - USM**
Screen Shot 2023-02-01 at 5 46 24 PM

### DIFF
--- a/src/elements/common/_variables.scss
+++ b/src/elements/common/_variables.scss
@@ -45,7 +45,8 @@ $sidebarTabResponsiveSize: 48px;
 }
 
 @mixin placeholder {
+    color: $bdl-gray-62;
     font-weight: normal;
     font-family: $font;
-    opacity: .6;
+    opacity: 1; // https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder#opaque_text
 }
--- a/src/features/unified-share-modal/UnifiedShareModal.scss
+++ b/src/features/unified-share-modal/UnifiedShareModal.scss
@@ -216,7 +216,7 @@
         }
 
         .text-area-container textarea::placeholder {
-            color: $bdl-gray-50;
+            color: $bdl-gray-62;
         }
 
         .bdl-PillSelector {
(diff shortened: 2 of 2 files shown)


src/elements/common/_variables.scss
```diff
@@ -45,7 +45,8 @@ $sidebarTabResponsiveSize: 48px;
 }
 
 @mixin placeholder {
+    color: $bdl-gray-62;
     font-weight: normal;
     font-family: $font;
-    opacity: .6;
+    opacity: 1; // https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder#opaque_text
```
tjuanitas, line 51: tldr: some browsers set an opacity less than 100% which affects the text color

greg-in-a-box (approved):

QingyuChai (approved):

src/elements/common/_variables.scss
JChan106: Should we change the opacity for certain browsers then? Would that make the placeholder in the browsers that have 100% opacity different than before?

tjuanitas: the main culprit seems to be Firefox. its default opacity for placeholders is `0.54`: https://hg.mozilla.org/mozilla-central/file/a07aebef20e7/layout/style/forms.css#l180

by setting `opacity: 1;`, all browsers will use the same opacity for placeholder text (i.e. align Firefox with Chrome and Edge)

JChan106 (approved):

PR #3229, opened 2023-02-02
docs(packages): describe supported packages

jpan-box:


QingyuChai (approved):
that looks good to me! We can always add more when they pop up as well

megansmith-box (approved):

PR #3232, opened 2023-02-03
fix(content-explorer): fix itemlist column header role

QingyuChai:
Replaced the `columnHeader` role with a `gridcell` role for empty column headers

The two empty column headers are:
1. above the icon column
2. above the 3 dots and share column
image

### DIFF
--- a/src/components/grid-view/__tests__/__snapshots__/GridView.test.js.snap
+++ b/src/components/grid-view/__tests__/__snapshots__/GridView.test.js.snap
@@ -37,6 +37,7 @@ exports[`components/grid-view/GridView should render() 1`] = `
     flexGrow={1}
     flexShrink={1}
     headerRenderer={[Function]}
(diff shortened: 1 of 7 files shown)


src/components/grid-view/__tests__/__snapshots__/GridView.test.js.snap
```diff
@@ -37,6 +37,7 @@ exports[`components/grid-view/GridView should render() 1`] = `
     flexGrow={1}
     flexShrink={1}
     headerRenderer={[Function]}
+    headerRole="columnheader"
```
QingyuChai, line 40: previously it was hard coded into react-virtualized/table, and now we are adding the hard-coded value as a default prop that can be overridden

megansmith-box (approved):
looks good to me

tjuanitas (approved):

PR #3233, opened 2023-02-06
fix(tooltip): Disable aria-describedby when ariaHidden is true

joshmarnold:
When passing the `ariaHidden` prop to the Tooltip, the `aria-describedby` attribute should also be hidden. This ensures that the description is not presented to assistive technology when the tooltip is meant to be hidden.

### DIFF
--- a/src/components/tooltip/__tests__/__snapshots__/Tooltip.test.tsx.snap
+++ b/src/components/tooltip/__tests__/__snapshots__/Tooltip.test.tsx.snap
@@ -3,7 +3,7 @@
 exports[`components/tooltip/Tooltip render() should match snapshot when stopBubble is set 1`] = `
 <div
   className="tooltip bdl-Tooltip"
-  id="tooltip26"
+  id="tooltip27"
   onClick={[Function]}
   onContextMenu={[Function]}
   onKeyPress={[Function]}


src/components/tooltip/Tooltip.tsx
```diff
@@ -299,7 +299,7 @@ class Tooltip extends React.Component<TooltipProps, State> {
         }
 
         if (showTooltip) {
-            if (!isLabelMatchingTooltipText || childAriaLabel === undefined) {
+            if (!ariaHidden && (!isLabelMatchingTooltipText || childAriaLabel === undefined)) {
```
tjuanitas, line 302: do we need it on the `if` below as well for `aria-errormessage`? if so, we could add `ariaHidden` to the outer `if` statement:
```
if (showTooltip && !ariaHidden) {
    [...]
}
```

tjuanitas, line 302: could we add a test for this?

joshmarnold: @tjuanitas I'm not sure here. I'm not familiar with the bug that was the cause for the `aria-errormessage`

tjuanitas: gotcha but I think it might have a similar issue

my understanding is that if `ariaHidden` is true, then the tooltip will be hidden to screen readers and so the component using `aria-errormessage` (the label, button, etc) will reference a tooltip ID that is not visible/accessible

tjuanitas (approved):
some minor things but it looks fine and I can follow up with someone from accessibility to figure out if `aria-errormessage` should be the same

tjuanitas (approved):

tjuanitas (approved):

tjuanitas (approved):

ahorowitz123 (approved):

PR #3234, opened 2023-02-06
fix: Improve color contrast for input borders

tjuanitas:
Updating the border-color of text fields to meet a 3:1 contrast ratio with the input background. Design's recommendation is to only update the ones in question (i.e. the inputs in the elements) and to update the other components to use UI Toolkit.

**Before**
Screen Shot 2023-02-06 at 11 36 57 AM
Screen Shot 2023-02-06 at 11 39 57 AM

**After**
Screen Shot 2023-02-06 at 11 37 09 AM
Screen Shot 2023-02-06 at 11 40 22 AM

### DIFF
--- a/src/elements/common/_variables.scss
+++ b/src/elements/common/_variables.scss
@@ -48,5 +48,5 @@ $sidebarTabResponsiveSize: 48px;
     color: $bdl-gray-62;
     font-weight: normal;
     font-family: $font;
-    opacity: 1; // https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder#opaque_text
+    opacity: 1; // Firefox sets opacity for placeholder to less than 100%
 }
--- a/src/elements/common/_variables.scss
+++ b/src/elements/common/_variables.scss
@@ -48,5 +48,7 @@ $sidebarTabResponsiveSize: 48px;
     color: $bdl-gray-62;
     font-weight: normal;
     font-family: $font;
-    opacity: 1; // Firefox sets opacity for placeholder to less than 100%
+    // Firefox sets opacity for placeholder to less than 100%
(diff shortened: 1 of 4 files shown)


src/elements/common/_variables.scss
```diff
@@ -48,5 +48,5 @@ $sidebarTabResponsiveSize: 48px;
     color: $bdl-gray-62;
     font-weight: normal;
     font-family: $font;
-    opacity: 1; // https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder#opaque_text
+    opacity: 1; // Firefox sets opacity for placeholder to less than 100%
```
QingyuChai, line 51: hmm, should we consider keeping the link as well?

src/elements/common/base.scss
```diff
@@ -12,6 +12,20 @@
     line-height: 20px;
     letter-spacing: .3px;
 
+    // TODO: Remove when border-color in src/styles/_inputs.scss is accessible
```
megansmith-box, line 15: When will this be? We can't we update it there now?

tjuanitas: tbh I'm not sure if we'll ever update that stylesheet before we migrate everything to ui-toolkit.

The reason we're not modifying it now is because it'll be a global change and design recommended to only update the ones in question. the other inputs would be fixed with the migration to ui-toolkit. I think this is because updating the border now won't align with the current text color and placeholder color for those inputs. i.e. the new border color will look weird with the old text color so we'd need to update all of them together with all the different interaction states

QingyuChai (approved):
lgtm

src/elements/common/base.scss
megansmith-box: That makes sense, thanks for explaining!

megansmith-box (approved):

JChan106 (approved):
lgtm

PR #3236, opened 2023-02-06
fix: Update box-shadow for inputs

tjuanitas:
Confirmed with the designer of design-system that box-shadows for input should be the following:
image (2)
and there is no value change for the different interaction states.

**Before**
Screen Shot 2023-02-06 at 3 39 50 PM

**After**
Screen Shot 2023-02-06 at 3 40 05 PM

### DIFF
--- a/src/styles/_inputs.scss
+++ b/src/styles/_inputs.scss
@@ -5,17 +5,16 @@
     color: $bdl-gray;
     border: 1px solid $bdl-gray-20;
     border-radius: $bdl-border-radius-size-med;
-    box-shadow: inset 0 1px 1px rgba(0, 0, 0, .1);
+    box-shadow: inset 0 2px 4px rgba(0, 0, 0, .1);
     transition: border-color linear .15s, box-shadow linear .1s;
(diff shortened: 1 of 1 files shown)


src/styles/_inputs.scss
```diff
@@ -5,17 +5,16 @@
...
     border-radius: $bdl-border-radius-size-med;
-    box-shadow: inset 0 1px 1px rgba(0, 0, 0, .1);
+    box-shadow: inset 0 2px 4px rgba(0, 0, 0, .1);
     transition: border-color linear .15s, box-shadow linear .1s;
     -webkit-font-smoothing: antialiased;
 }
 
 @mixin box-inputs-hover {
-    box-shadow: inset 0 1px 1px rgba(0, 0, 0, .15);
+    box-shadow: inset 0 2px 4px rgba(0, 0, 0, .1);
```
tjuanitas, line 14: keeping this mixin so it does not break current usages

QingyuChai (approved):

megansmith-box (approved):

PR #3241, opened 2023-02-08
feat(content explorer empty state): conditionally remove text

pusch-e:
Child of: https://github.com/box/box-ui-elements/pull/3237

Removes this text, only on the searching page, if isIncludeSubfoldersAllowed is true:
image

### DIFF
--- a/src/features/content-explorer/content-explorer/ContentExplorer.js
+++ b/src/features/content-explorer/content-explorer/ContentExplorer.js
@@ -43,6 +43,8 @@ class ContentExplorer extends Component {
         customInput: PropTypes.func,
         /** Any extra items in the header to the right of the search input (and new folder button) */
         headerActionsAccessory: PropTypes.node,
+        /** Whether the user can see the Include Subfolders toggle with breadcrumb updated design */
+        isIncludeSubfoldersAllowed: PropTypes.bool,
         /** Initial path of folders. The last folder in the array is the current folder. */
         initialFoldersPath: FoldersPathPropType.isRequired,
         /** Initial items that will show up as selected */
@@ -395,9 +397,18 @@ class ContentExplorer extends Component {
 
     renderItemListEmptyState = () => {
         const { foldersPath, isInSearchMode } = this.state;
+        const { isIncludeSubfoldersAllowed } = this.props;
         const isViewingSearchResults = isInSearchMode && foldersPath.length === 1;
+        const isOnInitialModalPage = foldersPath.length === 1 && foldersPath[0].id === '0';
+        console.log(foldersPath);
 
-        return <ContentExplorerEmptyState isSearch={isViewingSearchResults} />;
+        return (
+            <ContentExplorerEmptyState
+                isIncludeSubfoldersAllowed={isIncludeSubfoldersAllowed}
+                isOnInitialModalPage={isOnInitialModalPage}
+                isSearch={isViewingSearchResults}
+            />
+        );
     };
 
     render() {
--- a/src/features/content-explorer/content-explorer/ContentExplorer.js
+++ b/src/features/content-explorer/content-explorer/ContentExplorer.js
@@ -16,6 +16,12 @@ import { TYPE_FOLDER } from '../../../constants';
 
 import './ContentExplorer.scss';
 
+const INITIAL_FOLDERS_PATH = [
+    {
+        id: '0',
+        name: 'All Files',
+    },
+];
 class ContentExplorer extends Component {
     static propTypes = {
         /** Props for the action buttons container */
@@ -399,8 +405,10 @@ class ContentExplorer extends Component {
         const { foldersPath, isInSearchMode } = this.state;
         const { isIncludeSubfoldersAllowed } = this.props;
         const isViewingSearchResults = isInSearchMode && foldersPath.length === 1;
-        const isOnInitialModalPage = foldersPath.length === 1 && foldersPath[0].id === '0';
-        console.log(foldersPath);
+        const isOnInitialModalPage =
+            foldersPath.length === 1 &&
+            foldersPath[0].id === INITIAL_FOLDERS_PATH[0].id &&
+            foldersPath[0].name === INITIAL_FOLDERS_PATH[0].name;
 
         return (
             <ContentExplorerEmptyState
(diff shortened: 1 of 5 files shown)


src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -395,9 +403,20 @@ class ContentExplorer extends Component {
 
     renderItemListEmptyState = () => {
         const { foldersPath, isInSearchMode } = this.state;
+        const { isIncludeSubfoldersAllowed } = this.props;
         const isViewingSearchResults = isInSearchMode && foldersPath.length === 1;
+        const isOnInitialModalPage =
+            foldersPath.length === 1 &&
+            foldersPath[0].id === INITIAL_FOLDERS_PATH[0].id &&
+            foldersPath[0].name === INITIAL_FOLDERS_PATH[0].name;
```
menaczar, line 411: do we need to check on name here too? Id is unique

pusch-e: It might not be necessary. I just wanted to cover all bases to ensure it was the initial folder path

src/features/content-explorer/content-explorer/__tests__/ContentExplorerEmptyState.test.js
```diff
@@ -19,5 +19,17 @@ describe('features/content-explorer/content-explorer/ContentExplorerEmptyState',
 
             expect(wrapper.find('SearchEmptyState').length).toBe(1);
         });
+
+        test('should not render any text when isSearch is false, isIncludeSubfolders is true, and isOnInitialModalPage is true', () => {
```
SkyisAakash, line 23: We might want to test all possible combinations with `withData` to make sure conditional statements in `ContentExplorerEmptyState` are protected from mistakes.

tjuanitas (commented):
where is the
```
Search for Content

Find content across your enterprise [...]
```
stuff coming from?

is this custom rendered from the parent app? if so, why not pass a prop `noItemsRenderer` into `ContentExplorer` to `ItemList` that will render the prop or `this.renderItemListEmptyState`?

e.g.
```
// inside of the parent app

const noItemsRenderer = () => {...};

return <ContentExplorerModalContainer noItemsRenderer={noItemsRenderer}  />

---

// inside of BUIE: pass noItemsRenderer down to ContentExplorer

// inside of ContentExplorer.js

return (
    [...]
    <ItemList
        noItemsRenderer={noItemsRenderer || this.renderItemListEmptyState}
    />
    [...]
);

```

^ this is all based on the assumption there's some custom rendering happening from the screenshot

pusch-e:
> where is the
> 
> ```
> Search for Content
> 
> Find content across your enterprise [...]
> ```
> 
> stuff coming from?
> 
> is this custom rendered from the parent app? if so, why not pass a prop `noItemsRenderer` into `ContentExplorer` to `ItemList` that will render the prop or `this.renderItemListEmptyState`?
> 
> e.g.
> 
> ```
> // inside of the parent app
> 
> const noItemsRenderer = () => {...};
> 
> return <ContentExplorerModalContainer noItemsRenderer={noItemsRenderer}  />
> 
> ---
> 
> // inside of BUIE: pass noItemsRenderer down to ContentExplorer
> 
> // inside of ContentExplorer.js
> 
> return (
>     [...]
>     <ItemList
>         noItemsRenderer={noItemsRenderer || this.renderItemListEmptyState}
>     />
>     [...]
> );
> ```
> 
> ^ this is all based on the assumption there's some custom rendering happening from the screenshot

This is actually a really good point. I did not look into the ContentExplorerEmpty state as deeply until you mentioned this and it looks like the parent component to this in the admin-console repo (which is the gcmFolderExplorer) has a renderHeaderActionsAccessory() function of its own where it returns the empty state that we want and passes it down to the ContentExplorer as a prop called headerActionsAccessory. 

The ContentExplorerEmptyState component in BUIE is actually getting rendered over the empty state being passed in from the gcmFolderExplorer. 

In fact, when you set renderItemListEmptyState() from the ContentExplorer in BUIE to return null, the gcmFolderExplorer empty state is rendered instead, and it is the exact state that we want. So, I can instead get rid of all of these extra boolean values and just have it render the proper empty state if it exits, and else render the ContentExplorerEmptyState.

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -16,6 +16,12 @@ import { TYPE_FOLDER } from '../../../constants';
 
 import './ContentExplorer.scss';
 
+const INITIAL_FOLDERS_PATH = [
+    {
+        id: '0',
+        name: 'All Files',
```
apieceofbart, line 22: shouldn't this be translated?

tjuanitas:
> The only issue is that we want that text of "there are no subfolders in this folder" to still show in folders that have no contents. So we still need to conditionally check if we are on the first page to remove that text, since that is the only time we want to remove it.

this seems tied to a specific use case. if another team wanted to use the new toggle, they wouldn't have the same custom rendering as `GCMFolderExplorerEmptyState` so hiding this text would mean only the icon is rendered.

using `headerActionsAccessory` to render an empty state I think is doing more than the initial intention of the prop. It was mainly only for buttons/inputs in the header. I think it'd be good to utilize a new prop `noItemsRenderer` and pass the empty state through that. Then add the logic for displaying the text "there are no subfolders in this folder" inside of `GCMFolderExplorerEmptyState`

pusch-e:
> 

I understand now. This is a good point. I will move that logic to a new noItemsRenderer prop and pass that down from gcmFolderExplorer to ContentExplorer and ItemList like you suggested. Thank you for the suggestion, I think that is the cleanest way to do it!

pusch-e:
> > The only issue is that we want that text of "there are no subfolders in this folder" to still show in folders that have no contents. So we still need to conditionally check if we are on the first page to remove that text, since that is the only time we want to remove it.
> 
> this seems tied to a specific use case. if another team wanted to use the new toggle, they wouldn't have the same custom rendering as `GCMFolderExplorerEmptyState` so hiding this text would mean only the icon is rendered.
> 
> using `headerActionsAccessory` to render an empty state I think is doing more than the initial intention of the prop. It was mainly only for buttons/inputs in the header. I think it'd be good to utilize a new prop `noItemsRenderer` and pass the empty state through that. Then add the logic for displaying the text "there are no subfolders in this folder" inside of `GCMFolderExplorerEmptyState`

I understand now. This is a good point. I will move that logic to a new noItemsRenderer prop and pass that down from gcmFolderExplorer to ContentExplorer and ItemList like you suggested. Thank you for the suggestion, I think that is the cleanest way to do it!

pusch-e:
we found a fix for this somewhere else instead

PR #3242, opened 2023-02-08
feat(content-explorer): include subfolders toggle

pusch-e:
Child of: https://github.com/box/box-ui-elements/pull/3237

This includes the changes for the Include Subfolders toggle in ContentExplorer and the corresponding ItemList and ItemListButton changes


src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -164,13 +173,32 @@ class ContentExplorer extends Component {
...
+
+        // If we are inside a selected folder while the include subfolders toggle is on
+        if (includeSubfolders && !itemsAreSelectedAndActionDisabled && this.shouldSelectAndActionDisableItems()) {
+            this.setState({ itemsAreSelectedAndActionDisabled: true, allItemsEnabled: false });
+        }
+        // If we turn include subfolders toggle off while inside a folder that was chosen to include subfolders
+        else if (itemsAreSelectedAndActionDisabled && !this.shouldSelectAndActionDisableItems()) {
+            this.setState({ itemsAreSelectedAndActionDisabled: false });
+        }
+        // If we return back to the folder that our folder chosen for including subfolders is in
```
SkyisAakash, line 193:
```suggestion
        // If we return back to the folder that our folder-chosen-for-including-subfolders is in
```

SkyisAakash: Assuming I got it right. If not, please make it more clear.

pusch-e: No yeah that is what I was going for. Thank you for the suggestion

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -382,17 +430,145 @@ class ContentExplorer extends Component {
...
+
+    disableAllOtherItemButtons = itemToKeepId => {
+        const { handleSelectAllTree, items } = this.props;
+
+        if (this.itemsAreLoading()) {
+            return;
+        }
+
+        items.forEach(item => {
+            item.id === itemToKeepId ? (item.isDisabled = false) : (item.isDisabled = true);
```
SkyisAakash, line 478: `item.isDisabled = item.id !== itemToKeepId`

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -382,17 +430,145 @@ class ContentExplorer extends Component {
...
+                    }
+                });
+            } else if (!allItemsEnabled) {
+                this.setState({ folderToDeepScan: {}, itemsAreSelectedAndActionDisabled: false });
+                this.enableAllItemButtons();
+            }
+        });
+    };
+
+    areOtherItemsDisabled = enabledItemId => {
```
SkyisAakash, line 515: This and some other functions seem like they could be suited for a helper file. Items can also be passed in as array argument. That way we can write test cases to test them thoroughly with all test cases. 
Name could also be little more specific, perhaps something like `areAllItemsDisabledExcept`?

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -382,17 +430,145 @@ class ContentExplorer extends Component {
...
+                item.isLoading
+            ) {
+                return false;
+            }
+            return true;
+        });
+        return result;
+    };
+
+    foldersPresent = () => {
```
SkyisAakash, line 531: Suitable for helper function, name could be more like `isFolderPresent`

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -382,17 +430,145 @@ class ContentExplorer extends Component {
...
+    noFoldersSelected = () => {
+        const { selectedItems } = this.state;
+        const result = Object.values(selectedItems).every(item => {
+            if (item.type === TYPE_FOLDER) {
+                return false;
+            }
+            return true;
+        });
+        return result;
+    };
```
SkyisAakash, line 555: You can use `isFolderPresent` here.

pusch-e: Ah yeah if I make isFoldersPresent take in an iterable then I can use it for both. The difference here is that I am checking if any of the selected items (state.selectedItems) are folders where as foldersPresent is currently checking if any of the displayed items (props.items) are folders

src/features/content-explorer/content-explorer-modal-container/ContentExplorerModalContainer.js
```diff
@@ -54,6 +54,8 @@ class ContentExplorerModalContainer extends Component {
         onSelectItem: PropTypes.func,
         /** Folder is in the process of being created. */
         isCreatingFolder: PropTypes.bool,
+        /** Whether the user can see the Include Subfolders toggle with updated design */
+        isIncludeSubfoldersAllowed: PropTypes.bool,
```
tjuanitas, line 58: how about something like `canIncludeSubfolders` as the prop name? present tense verbs with an `is` prefix sound a little awkward to me

src/features/content-explorer/content-explorer-modal-container/ContentExplorerModalContainer.js
```diff
@@ -54,6 +54,8 @@ class ContentExplorerModalContainer extends Component {
         onSelectItem: PropTypes.func,
         /** Folder is in the process of being created. */
         isCreatingFolder: PropTypes.bool,
+        /** Whether the user can see the Include Subfolders toggle with updated design */
```
tjuanitas, line 57: i think you can remove the `with updated design` part

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -508,12 +692,25 @@ class ContentExplorer extends Component {
                 >
                     {headerActionsAccessory}
                 </ContentExplorerHeaderActions>
-                {isSelectAllAllowed && (
-                    <ContentExplorerSelectAll
-                        numTotalItems={numTotalItems}
-                        isSelectAllChecked={isSelectAllChecked}
+                {isSelectAllAllowed && isIncludeSubfoldersAllowed ? (
```
tjuanitas, line 695: does this mean the new feature can only be used with `isSelectAllAllowed`? could we make them independent of each other?

something like:
```
<div>
    {isIncludeSubfoldersAllowed && <ContentExplorerIncludeSubfolders [...] />}
    {isSelectAllAllowed && <ContentExplorerSelectAll [...] />}
</div>
```
with `ContentExplorerSelectAll` taking a prop to optionally hide the label

this reuses the checkbox without having to add a separate one into `ContentExplorerIncludeSubfolders`

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -525,11 +722,16 @@ class ContentExplorer extends Component {
                     itemButtonRenderer={itemButtonRenderer}
                     itemIconRenderer={itemIconRenderer}
                     itemNameLinkRenderer={itemNameLinkRenderer}
-                    items={items}
+                    items={includeSubfolders && !itemsAreSelectedAndActionDisabled ? this.getFoldersOnly() : items}
```
tjuanitas, line 725: some of this looks very specific to how enterprise-reports will use the toggle but it might not be the same for all use cases. could we make it more general and utilize the item renderers in the parent app?

tjuanitas (commented):
a lot of this might be too specific to enterprise reports. could we try to generalize it (while allowing customization through props) so other apps could potentially use this toggle as well without the specific behaviors?

for example:
I was thinking that the changes in ItemList might be able to utilize the `itemButtonRenderer` prop. in addition, I think it's possible to move the logic for the new methods: `shouldSelectAndActionDisableItems`, `disableAllOtherItemButtons`, `enableAllItemButtons`, etc. into the parent app and update the `items` prop that's passed to ContentExplorer. this way the toggle could fit more use cases e.g. a use case to select multiple folders with the toggle enabled

src/features/content-explorer/content-explorer/ContentExplorer.js
pusch-e: As of right now, I believe the new feature can only be used with isSelectAllAllowed. This could potentially change in the future though

src/features/content-explorer/messages.js
```diff
@@ -118,6 +118,31 @@ const messages = defineMessages({
         description: "Results label for number of items on list when it's just 1",
         id: 'boxui.contentExplorer.result',
     },
+    includeSubfolders: {
+        defaultMessage: 'Include Subfolders',
+        description: 'Label text shown next to the Include Subfolders toggle',
+        id: 'boxui.contentExplorer.includeSubfolders',
+    },
+    includeSubfoldersDefaultInformation: {
+        defaultMessage: 'When enabled, report is run on a single selected folder and all its subfolders.',
```
tjuanitas, line 127: this message is specifying `report` which is specific to reports. could we generalize this? maybe we can pass the tooltip component or message as a prop?

src/features/content-explorer/content-explorer/constants.js
```diff
@@ -0,0 +1,23 @@
...
+    return result;
+};
+
+export const isFolderPresent = items => {
+    for (let i = 0; i < items.length; i += 1) {
+        if (items[i].type === TYPE_FOLDER) {
+            return true;
+        }
+    }
+    return false;
```
tjuanitas, line 18: `return items.some(item => item.type === TYPE_FOLDER);`

src/features/content-explorer/content-explorer/constants.js
```diff
@@ -0,0 +1,23 @@
...
+};
+
+export const itemsAreLoading = items => {
+    return items && items[0] && items[0].isLoading;
```
tjuanitas, line 22: why is it only checking the first item?

src/features/content-explorer/content-explorer/constants.js
```diff
@@ -0,0 +1,23 @@
+import { TYPE_FOLDER } from '../../../constants';
+
+/* ----------------------- Helpers for ContentExplorer ---------------------------- */
+
+export const areAllItemsActionDisabled = items => {
+    const result = items.every(item => {
+        return item.isActionDisabled;
+    });
+    return result;
```
tjuanitas, line 9: this could return directly
```suggestion
    return items.every(item => item.isActionDisabled);
```

src/features/content-explorer/content-explorer/constants.js
```diff
@@ -0,0 +1,23 @@
...
+export const isFolderPresent = items => {
+    for (let i = 0; i < items.length; i += 1) {
+        if (items[i].type === TYPE_FOLDER) {
+            return true;
+        }
+    }
+    return false;
+};
+
+export const itemsAreLoading = items => {
```
tjuanitas, line 21: `isLoadingItems`?

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -56,13 +62,22 @@ class ContentExplorer extends Component {
          * @param {Array} newFoldersPath
          */
         onEnterFolder: PropTypes.func.isRequired,
+        /** Called when the folders path is updated */
+        onFoldersPath: PropTypes.func,
```
tjuanitas, line 66: is this duplicating the use of `onEnterFolder`? do we need both props?

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -156,6 +171,7 @@ class ContentExplorer extends Component {
             foldersPath: props.initialFoldersPath,
             isInSearchMode: false,
             isSelectAllChecked: false,
+            includeSubfolders: false,
```
tjuanitas, line 174: `isIncludeSubfoldersEnabled`?

src/features/content-explorer/content-explorer/ContentExplorerIncludeSubfolders.js
```diff
@@ -0,0 +1,74 @@
...
+ContentExplorerIncludeSubfolders.propTypes = {
+    isFolderPresent: PropTypes.bool,
+    handleIncludeSubfoldersToggle: PropTypes.func,
+    handleSelectAllClick: PropTypes.func,
+    hideSelectAllCheckbox: PropTypes.bool,
+    isSelectAllChecked: PropTypes.bool,
+    noFoldersSelected: PropTypes.bool,
+    numOfSelectedItems: PropTypes.number,
+    isToggleDisabled: PropTypes.bool,
+};
```
tjuanitas, line 72: could we sort these?

src/features/content-explorer/content-explorer/ContentExplorerIncludeSubfolders.js
```diff
@@ -0,0 +1,74 @@
...
+export { ContentExplorerIncludeSubfolders as ContentExplorerIncludeSubfoldersBase };
```
tjuanitas, line 73: really really minor but you can add a line break here?

src/features/content-explorer/content-explorer/ContentExplorerIncludeSubfolders.js
```diff
@@ -0,0 +1,74 @@
...
+            return <FormattedMessage {...messages.includeSubfoldersMultipleFoldersSelected} />;
+        }
+        if (numOfSelectedItems === 1 && noFoldersSelected) {
+            return <FormattedMessage {...messages.includeSubfoldersNoFoldersSelected} />;
+        }
+        if (numOfSelectedItems === 0 && !isFolderPresent) {
+            return <FormattedMessage {...messages.includeSubfoldersNoFoldersToSelect} />;
+        }
+        return <FormattedMessage {...messages.includeSubfoldersDefaultInformation} />;
+    };
```
tjuanitas, line 32: we tend to avoid multiple returns like this unless it's like a guard clause.

could we do something like:

```
let tooltipMessage = messages.includeSubfoldersDefaultInformation;

if (numOfSelectedItems > 1) {
    tooltipMessage = messages.includeSubfoldersMultipleFoldersSelected;
}
if (numOfSelectedItems === 1 && noFoldersSelected) {
    tooltipMessage = messages.includeSubfoldersNoFoldersSelected;
}
if (numOfSelectedItems === 0 && !isFolderPresent) {
    tooltipMessage = messages.includeSubfoldersNoFoldersToSelect;
}

[...]

<Tooltip text={<FormattedMessage {...tooltipMessage} />} />
```

src/features/content-explorer/content-explorer/ContentExplorerIncludeSubfolders.js
```diff
@@ -0,0 +1,74 @@
...
+    isToggleDisabled,
+}) => {
+    const getTooltipMessage = () => {
+        if (numOfSelectedItems > 1) {
+            return <FormattedMessage {...messages.includeSubfoldersMultipleFoldersSelected} />;
+        }
+        if (numOfSelectedItems === 1 && noFoldersSelected) {
+            return <FormattedMessage {...messages.includeSubfoldersNoFoldersSelected} />;
+        }
+        if (numOfSelectedItems === 0 && !isFolderPresent) {
```
tjuanitas, line 28: could these be `else if`s so all of them aren't evaluated?

src/features/content-explorer/content-explorer/ContentExplorerIncludeSubfolders.js
```diff
@@ -0,0 +1,74 @@
...
+                </Tooltip>
+            )}
+        </div>
+    );
+};
+
+ContentExplorerIncludeSubfolders.propTypes = {
+    isFolderPresent: PropTypes.bool,
+    handleIncludeSubfoldersToggle: PropTypes.func,
+    handleSelectAllClick: PropTypes.func,
```
tjuanitas, line 66: when we pass functions as props, we usually use `on` prefix instead of `handle`

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -423,6 +488,7 @@ class ContentExplorer extends Component {
             isChooseButtonLoading,
             isCopyButtonLoading,
             isCreateNewFolderAllowed,
+            canIncludeSubfolders,
```
tjuanitas, line 491: this list looks mostly sorted

src/features/content-explorer/content-explorer/ContentExplorerIncludeSubfolders.js
```diff
@@ -0,0 +1,74 @@
+import PropTypes from 'prop-types';
+import React from 'react';
```
tjuanitas, line 2: can we move react first?

src/features/content-explorer/content-explorer/ContentExplorerIncludeSubfolders.js
```diff
@@ -0,0 +1,74 @@
...
+import { injectIntl, FormattedMessage } from 'react-intl';
+import Checkbox from '../../../components/checkbox/Checkbox';
+import Tooltip from '../../../components/tooltip';
+import Toggle from '../../../components/toggle';
+
+import messages from '../messages';
+import IconInfo from '../../../icons/general/IconInfo';
```
tjuanitas, line 9: does `InfoBadge16` work? https://opensource.box.com/box-ui-elements/storybook/?path=/story/icon-fill-infobadge16--info-badge-16

tjuanitas, line 9: could we group this import with the components above?

src/features/content-explorer/content-explorer/ContentExplorer.scss
```diff
@@ -71,6 +71,31 @@
         }
     }
 
+    .bdl-ContentExplorerIncludeSubfolders {
+        display: inline-block;
+        width: 100%;
+        height: 8 * $bdl-grid-unit;
+        vertical-align: middle;
```
tjuanitas, line 78: `vertical-align` isn't that great with aligning. are we able to use flex-box instead?

src/features/content-explorer/item-list/ItemListButton.js
```diff
@@ -38,6 +45,7 @@ const ItemListButton = ({ contentExplorerMode, id = '', isDisabled = false, isSe
 ItemListButton.propTypes = {
     contentExplorerMode: ContentExplorerModePropType.isRequired,
     id: PropTypes.string,
+    isChecked: PropTypes.bool,
```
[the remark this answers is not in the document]
pusch-e: I am using isChecked in the itemButtonRenderer function in the parent component in admin console because item buttons cannot be both selected and disabled if you are using isDisabled and isSelected

src/features/content-explorer/content-explorer/ContentExplorer.scss
```diff
@@ -71,6 +71,31 @@
...
+
+        .bdl-ContentExplorerIncludeSubfolders-toggle {
+            display: block;
+            float: left;
```
tjuanitas, line 82: `float` is sort of an older solution. are we able to use flexbox with `justify-content: space-between;`?

src/features/content-explorer/item-list/ItemListButton.js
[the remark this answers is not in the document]
pusch-e: There is a test demonstrating this at line 47 in the ItemListButton.test.js

src/features/content-explorer/item-list/ItemList.js
```diff
@@ -86,16 +86,19 @@ const itemNameCellRenderer = rendererParams => {
...
+    return (
+        name && (
+            <ItemListButton
+                contentExplorerMode={contentExplorerMode}
+                id={id}
+                isDisabled={isActionDisabled}
+                isSelected={isItemSelected(id, selectedItems)}
+                name={name}
+            />
+        )
```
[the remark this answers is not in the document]
pusch-e: Oops yeah I can revert it back. I should have made sure that it was the same

src/features/content-explorer/content-explorer/ContentExplorerIncludeSubfolders.js
```diff
@@ -0,0 +1,74 @@
...
+                className="bdl-ContentExplorerIncludeSubfolders-toggle"
+                label={<FormattedMessage {...messages.includeSubfolders} />}
+                isDisabled={isToggleDisabled}
+                onChange={handleIncludeSubfoldersToggle}
+            />
+
+            <Tooltip className="bdl-ContentExplorerIncludeSubfolders-info-tooltip" text={getTooltipMessage()}>
+                <div className="bdl-ContentExplorerIncludeSubfolders-icon-container">
+                    <IconInfo className="bdl-ContentExplorerIncludeSubfolders-icon" height={20} width={20} />
+                </div>
```
tjuanitas, line 46: I don't think you need this outer div

src/features/content-explorer/content-explorer/ContentExplorerIncludeSubfolders.js
```diff
@@ -0,0 +1,74 @@
...
+            </Tooltip>
+
```
tjuanitas, line 48: nit: empty line

src/features/content-explorer/content-explorer/constants.js
pusch-e: Backend returns a hardcoded list of 7 item objects that just have the value {isLoading: true} when the items are loading. So, I just need to check that one of the items is loading. This logic was in BUIE before, I just moved it to a helper function

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -56,13 +62,22 @@ class ContentExplorer extends Component {
...
          *
          * @param {Object} selectedItem
          * @param {number} selectedItemIndex
          */
         onSelectItem: PropTypes.func,
+        /** Called when an item is selected
+         *
+         * @param {Object} selectedItem
+         */
+        onSelectedItems: PropTypes.func,
```
tjuanitas, line 80: why do we need this instead of utilizing `onSelectItem` and `onSelectedClick`?

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -41,6 +45,8 @@ class ContentExplorer extends Component {
         contentExplorerMode: ContentExplorerModePropType.isRequired,
         /** Props that contains the custom search input. Is rendered in header actions */
         customInput: PropTypes.func,
+        /** Called when a user selects a folder to include subfolders for */
+        handleSelectAllTree: PropTypes.func,
```
tjuanitas, line 49: is this being used?

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -508,12 +574,25 @@ class ContentExplorer extends Component {
                 >
                     {headerActionsAccessory}
                 </ContentExplorerHeaderActions>
-                {isSelectAllAllowed && (
-                    <ContentExplorerSelectAll
-                        numTotalItems={numTotalItems}
-                        isSelectAllChecked={isSelectAllChecked}
+                {isSelectAllAllowed && canIncludeSubfolders ? (
```
tjuanitas, line 577: we might be able to replace `canIncludeSubfolders` with `onIncludeSubfoldersToggle` so we're only passing one prop all the way down instead of two.

i.e. instead of passing two props (one to display and one callback), we can pass just the callback and use that to hide or show the toggle

src/features/content-explorer/content-explorer/ContentExplorerIncludeSubfolders.js
pusch-e: Is there a version of this in the primary blue color?

src/features/content-explorer/content-explorer/ContentExplorerIncludeSubfolders.js
```diff
@@ -0,0 +1,74 @@
...
+const ContentExplorerIncludeSubfolders = ({
+    isFolderPresent,
+    handleIncludeSubfoldersToggle,
+    handleSelectAllClick,
+    hideSelectAllCheckbox,
+    isSelectAllChecked,
+    noFoldersSelected,
+    numOfSelectedItems = 0,
+    isToggleDisabled,
```
tjuanitas, line 19: could we sort?

src/features/content-explorer/content-explorer/ContentExplorer.js
tjuanitas: that would be for the enterprise-reports use case though right? that could be handled by passing the appropriate props from the parent

I think it would be worth making them independent. changing it in the future might be a breaking change for other apps that start using this feature

src/features/content-explorer/content-explorer/ContentExplorerIncludeSubfolders.js
tjuanitas: you need to use CSS to target the fill and path to set to the color.

it would be something like:
```
// inside ContentExplorerIncludeSubfolders.js
<InfoBadge16 className="bdl-ContentExplorerIncludeSubfolders-infoIcon" />

// inside ContentExplorerIncludeSubfolders.scss
.bdl-ContentExplorerIncludeSubfolders-infoIcon {
    path {
        fill: $bdl-box-blue;
    }
}
```

src/features/content-explorer/content-explorer/__tests__/ContentExplorer.test.js
```diff
@@ -106,6 +106,113 @@ describe('features/content-explorer/content-explorer/ContentExplorer', () => {
             expect(wrapper.find('ContentExplorerSelectAll').prop('isSelectAllChecked')).toEqual(isSelectAllChecked);
         });
 
+        test('should render ContentExplorerIncludeSubfolders with canIncludeSubfolders = true and isSelectAllAllowed = true', () => {
+            const isSelectAllAllowed = true;
+            const canIncludeSubfolders = true;
+            const wrapper = renderComponent({ canIncludeSubfolders, isSelectAllAllowed });
+
+            expect(wrapper.exists('ContentExplorerIncludeSubfolders')).toBeTruthy();
```
tjuanitas, line 114: when using `exists`, we can be more explicit about the value: `expect(wrapper.exists('ContentExplorerIncludeSubfolders')).toBe(true);`

src/features/content-explorer/content-explorer/__tests__/ContentExplorer.test.js
```diff
@@ -106,6 +106,113 @@ describe('features/content-explorer/content-explorer/ContentExplorer', () => {
...
+        });
+
+        test('should not render ContentExplorerIncludeSubfolders with canIncludeSubfolders = false and isSelectAllAllowed = true', () => {
+            const isSelectAllAllowed = true;
+            const canIncludeSubfolders = false;
+            const wrapper = renderComponent({ canIncludeSubfolders, isSelectAllAllowed });
+
+            expect(wrapper.exists('ContentExplorerSelectAll')).toBeTruthy();
+            expect(wrapper.find('ContentExplorerIncludeSubfolders').length).toBe(0);
```
tjuanitas, line 123: you're using both to `exists` and `find`. you should be able to do:
`expect(wrapper.exists('ContentExplorerIncludeSubfolders')).toBe(false);`

src/features/content-explorer/content-explorer/__tests__/ContentExplorer.test.js
```diff
@@ -106,6 +106,113 @@ describe('features/content-explorer/content-explorer/ContentExplorer', () => {
...
+        });
+
+        test('should not render ContentExplorerIncludeSubfolders nor ContentExplorerSelectAll with canIncludeSubfolders = true and isSelectAllAllowed = false', () => {
+            const isSelectAllAllowed = false;
+            const canIncludeSubfolders = true;
+            const wrapper = renderComponent({ canIncludeSubfolders, isSelectAllAllowed });
+
+            expect(wrapper.find('ContentExplorerSelectAll').length).toBe(0);
+            expect(wrapper.find('ContentExplorerIncludeSubfolders').length).toBe(0);
```
tjuanitas, line 132: these and the tests below switched from using `exist` to `find`/`length`. we should use one approach

src/features/content-explorer/content-explorer/__tests__/ContentExplorer.test.js
```diff
@@ -106,6 +106,113 @@ describe('features/content-explorer/content-explorer/ContentExplorer', () => {
             expect(wrapper.find('ContentExplorerSelectAll').prop('isSelectAllChecked')).toEqual(isSelectAllChecked);
         });
 
+        test('should render ContentExplorerIncludeSubfolders with canIncludeSubfolders = true and isSelectAllAllowed = true', () => {
+            const isSelectAllAllowed = true;
+            const canIncludeSubfolders = true;
+            const wrapper = renderComponent({ canIncludeSubfolders, isSelectAllAllowed });
```
tjuanitas, line 112: you don't need variables here, you can add them directly to the object inline

tjuanitas: same for the tests below

src/features/content-explorer/content-explorer/__tests__/ContentExplorer.test.js
```diff
@@ -106,6 +106,113 @@ describe('features/content-explorer/content-explorer/ContentExplorer', () => {
...
+            expect(wrapper.find('ContentExplorerSelectAll').length).toBe(0);
+            expect(wrapper.find('ContentExplorerIncludeSubfolders').length).toBe(0);
+        });
+
+        test('should pass isFolderPresent to ContentExplorerIncludeSubfolders', () => {
+            const items = [
+                { id: 'item1', name: 'name1', type: 'folder' },
+                { id: 'item2', name: 'name2' },
+            ];
+            const isFolderPresent = true;
```
tjuanitas, line 149: you can add this directly to the `expect`, you don't need the variable

src/features/content-explorer/content-explorer/__tests__/ContentExplorer.test.js
```diff
@@ -106,6 +106,113 @@ describe('features/content-explorer/content-explorer/ContentExplorer', () => {
...
+
+            const wrapper = renderComponent({
+                canIncludeSubfolders: true,
+                isSelectAllAllowed: true,
+                items,
+            });
+            expect(wrapper.find('ContentExplorerIncludeSubfolders').prop('isFolderPresent')).toEqual(isFolderPresent);
```
tjuanitas, line 156: we usually use `toBe` over `toEqual` for primitive values

src/features/content-explorer/content-explorer/__tests__/ContentExplorer.test.js
```diff
@@ -106,6 +106,113 @@ describe('features/content-explorer/content-explorer/ContentExplorer', () => {
...
+        test('should not render ContentExplorerIncludeSubfolders nor ContentExplorerSelectAll with canIncludeSubfolders = false and isSelectAllAllowed = false', () => {
+            const isSelectAllAllowed = false;
+            const canIncludeSubfolders = false;
+            const wrapper = renderComponent({ canIncludeSubfolders, isSelectAllAllowed });
+
+            expect(wrapper.find('ContentExplorerSelectAll').length).toBe(0);
+            expect(wrapper.find('ContentExplorerIncludeSubfolders').length).toBe(0);
+        });
+
+        test('should pass isFolderPresent to ContentExplorerIncludeSubfolders', () => {
```
tjuanitas, line 144: you can omit these tests are just passing props

src/features/content-explorer/messages.js
pusch-e: Pass it all the way down from the parent component?

src/features/content-explorer/content-explorer/ContentExplorer.js
pusch-e: It covers a little more than entering a folder. The parent component needs to know the most up to date folderspath so that it knows where it is in relation to the folder chosen to deep scan with (aka the folder chosen while the toggle is on). onEnterFolder just handles when you click on a folder or a breadcrumb, but not when you go back to the all files modal page

I might be able to use a combination of onEnterFolder and something else to mimic on Folderspath

tjuanitas: that sounds like a bug if `onEnterFolder` is not updating when the user goes back to the root folder. based on the description, it should update whenever the current folder changes

src/features/content-explorer/content-explorer/ContentExplorer.js
pusch-e: Alright I did this instead. I am going to have to make a small chnge to the other PR as well since that uses canIncludeSubfodlers still

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -319,7 +353,7 @@ class ContentExplorer extends Component {
 
         if (item.type === TYPE_FOLDER) {
             this.enterFolder(item);
-        } else {
+        } else if (!item.isActionDisabled) {
```
tjuanitas, line 356: what's item.isActionDisabled vs item.isDisabled?

pusch-e: actionDisabled means that the item's name can still be clicked on to see what is inside the folder. But it's checkbox cannot be interacted with and it cannot be selected.

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -383,14 +415,44 @@ class ContentExplorer extends Component {
...
+        const selectedItemIsNotFolder =
+            this.numOfSelectedItems() === 1 && !isFolderPresent(Object.values(selectedItems));
+        return (
+            isLoadingItems(items) ||
+            (!isIncludeSubfoldersEnabled &&
+                !(selectedItemIsFolder && !isFolderPresent(items)) &&
+                ((!areAllItemsActionDisabled(items) && !isFolderPresent(items)) ||
+                    selectedItemIsNotFolder ||
+                    this.numOfSelectedItems() > 1))
+        );
```
tjuanitas, line 455: this logic is hard to follow and as well as could be different by use case. it looks we can determine if the toggle should be disabled in the parent app because the wrapping component should know the `selectedItems` and can track the state of the toggle using `onIncludeSubfoldersToggle`.

we might be able to get away with passing only one prop all the way down called `includeSubfoldersProps` (similar to `searchInputProps`).

the wrapping component should keep it's own state of if the toggle is enabled and the selected items

src/features/content-explorer/content-explorer/ContentExplorer.scss
```diff
@@ -71,6 +71,33 @@
...
 
+    .bdl-ContentExplorerIncludeSubfolders {
+        display: flex;
+        align-items: center;
+        justify-content: space-between;
+        height: 10 * $bdl-grid-unit;
+        margin-top: 20px;
+        margin-bottom: -10px;
+
+        .bdl-ContentExplorerIncludeSubfolders-toggleIconContainer {
```
tjuanitas, line 82: why is this div necessary? shouldn't the `display: flex;` and `align-items: center;` on the outer div be enough to center everything horizontally including the checkbox?

src/features/content-explorer/content-explorer/ContentExplorer.scss
```diff
@@ -71,6 +71,33 @@
         }
     }
 
+    .bdl-ContentExplorerIncludeSubfolders {
+        display: flex;
+        align-items: center;
+        justify-content: space-between;
+        height: 10 * $bdl-grid-unit;
+        margin-top: 20px;
+        margin-bottom: -10px;
```
tjuanitas, line 80: why do we need the negative margin? should the margin-top be decreased?

src/features/content-explorer/content-explorer/ContentExplorer.scss
pusch-e: This was to wrap the info badge icon and toggle together so I can do justify content space between with this div and the checkbox. 
I can alternatively take this div out then margin auto the checkbox to the right

src/features/content-explorer/content-explorer/ContentExplorer.scss
pusch-e: The negative margin was added because the base toggle and checkbox components had a bottom margin of 20px on them. So a negative margin for the outer div helped push everything down to where it should be positioned

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -508,12 +578,23 @@ class ContentExplorer extends Component {
...
+                        tooltipMessageForToggle={tooltipMessageForToggle}
                     />
+                ) : (
+                    isSelectAllAllowed && (
+                        <ContentExplorerSelectAll
+                            numTotalItems={numTotalItems}
+                            isSelectAllChecked={isSelectAllChecked}
+                            handleSelectAllClick={this.handleSelectAllClick}
+                        />
+                    )
```
tjuanitas, line 597: this is how I think we can combine the checkboxes used for both:

```
{includeSubfoldersProps && <ContentExplorerIncludeSubfolders {...includeSubfoldersProps} />
{isSelectAllAllowed &&
        <ContentExplorerSelectAll
                handleSelectAllClick={this.handleSelectAllClick}
                isLabelHidden={includeSubfoldersProps}
                isSelectAllChecked={isSelectAllChecked}
                numTotalItems={numTotalItems}
        />
```

then inside of `ContentExplorerSelectAll`:
```
const ContentExplorerSelectAll = ({ intl, numTotalItems = 0, isLabelHidden = false, isSelectAllChecked, handleSelectAllClick }) => (
    <div className="content-explorer-select-all-container">
        {!isLabelHidden && <label className="content-explorer-select-all-items-counter">
            {numTotalItems === 1 ? (
                <FormattedMessage {...messages.result} values={{ itemsCount: intl.formatNumber(numTotalItems) }} />
            ) : (
                <FormattedMessage {...messages.results} values={{ itemsCount: intl.formatNumber(numTotalItems) }} />
            )}
        </label>
        }
        {!isLabelHidden && <label className="content-explorer-select-all-checkbox-label">
            <FormattedMessage {...messages.selectAll} />
        </label>
        }
        <Checkbox
            hideLabel
            className="content-explorer-select-all-checkbox"
            onChange={handleSelectAllClick}
            isChecked={isSelectAllChecked}
        />
    </div>
);
```

pusch-e: In addition to this we would also have to wrap the Checkbox in a conditional since it is supposed to be hidden when the toggle is turned on. Also, how can we render the tooltip given that it is only supposed to show when we have the toggle

tjuanitas: whoops forgot to add the tooltip:

const ContentExplorerSelectAll = ({ intl, numTotalItems = 0, isLabelHidden = false, isSelectAllChecked, handleSelectAllClick }) => (
    <div className="content-explorer-select-all-container">
        {!isLabelHidden && <label className="content-explorer-select-all-items-counter">
            {numTotalItems === 1 ? (
                <FormattedMessage {...messages.result} values={{ itemsCount: intl.formatNumber(numTotalItems) }} />
            ) : (
                <FormattedMessage {...messages.results} values={{ itemsCount: intl.formatNumber(numTotalItems) }} />
            )}
        </label>
        }
        {!isLabelHidden && <label className="content-explorer-select-all-checkbox-label">
            <FormattedMessage {...messages.selectAll} />
        </label>
        }
        <Tooltip isShown={isLabelHidden ? undefined : false}> // undefined will give it the default hover behavior. false will hide it
            <Checkbox
                hideLabel
                className="content-explorer-select-all-checkbox"
                onChange={handleSelectAllClick}
                isChecked={isSelectAllChecked}
            />
        </Tooltip>
    </div>
);

tjuanitas: the parent component should know if the toggle is enabled or disabled and can change value of prop `isSelectAllAllowed`. this way the two features are kept independent of each other

src/features/content-explorer/content-explorer/ContentExplorer.scss
tjuanitas: instead of the negative margin, could we target the toggle and checkbox classes and reset those within this block?

```
.bdl-ContentExplorerIncludeSubfolders {
    .checkbox-container ,
    .toggle-container {
        margin: 0;
    }
}
```

src/features/content-explorer/content-explorer/ContentExplorer.js
pusch-e: Also, onEnterFolder takes in just a single folder. I need the whole foldersPath and I dont know if I can alter the function to take in the folderspath as well since other components could be using it

src/features/content-explorer/content-explorer/ContentExplorer.scss
tjuanitas: you can leave this for now but I think it might be better to put the info tooltip inside the label prop of the tooltip similar to this component: https://github.com/box/box-ui-elements/blob/master/src/features/advanced-content-insights/AdvancedContentInsightsToggle.tsx#L33-L55

src/features/content-explorer/content-explorer/ContentExplorer.js
pusch-e: There are places where the selectedItems state is altered but these functions are not called. For instance, in the handleSelectAllClick function, the selectedItems in the ContentExplorer state is changed, but no handler is called and I need the parent app to know if those items are selected. So I added the on selectedItems handler that is called in every case that the ContentExplorer's state.selectedItems is changed.
I am not sure why the other handlers you mentioned do not do this, but I was unsure of if I could paste them in these scenarios since I am not entirely sure of where else they are used

pusch-e: Both this and the onFoldersPath handlers that I made were both created under the thought process where I knew I needed to know how both of these were changing in the parent component, and I saw some handlers that already existed in ContentExplorer, but they are not being called in every instance where either the selectedItems or foldersPath is being updated. If it can be verified that the other handlers are safe to add to these other situations where the state is being changed, then I can put those in instead

src/features/content-explorer/content-explorer/ContentExplorer.js
tjuanitas: I think it should be available as the second param:

https://github.com/box/box-ui-elements/blob/3689909192a6626cab6356007bdd29c83b83ebe7/src/features/content-explorer/content-explorer/ContentExplorer.js#L287

https://github.com/box/box-ui-elements/blob/3689909192a6626cab6356007bdd29c83b83ebe7/src/features/content-explorer/content-explorer-modal-container/ContentExplorerModalContainer.js#L172

although it's possible there's a bug somewhere

src/features/content-explorer/content-explorer/ContentExplorer.scss
pusch-e: Ah I see that is a good idea. I would have to wrap the FormattedMessage component and tooltip tg to pass as a label prop, but it would be cleaner

src/features/content-explorer/content-explorer/ContentExplorer.js
tjuanitas: gotcha, I think I'm better understanding now

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -56,13 +58,22 @@ class ContentExplorer extends Component {
...
         /**
          * Called when an item is selected
          *
          * @param {Object} selectedItem
          * @param {number} selectedItemIndex
          */
         onSelectItem: PropTypes.func,
+        /** Called when an item is selected
+         *
+         * @param {Object} selectedItem
```
tjuanitas, line 74: `selectedItems`

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -56,13 +58,22 @@ class ContentExplorer extends Component {
...
+        /** Called when the include subfolders toggle is changed */
+        onIncludeSubfoldersToggle: PropTypes.func,
         /**
          * Called when an item is selected
          *
          * @param {Object} selectedItem
          * @param {number} selectedItemIndex
          */
         onSelectItem: PropTypes.func,
+        /** Called when an item is selected
```
tjuanitas, line 72: this is called whenever the selected items list changes. not just when one single one is selected

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -284,11 +312,11 @@ class ContentExplorer extends Component {
     };
 
     handleItemClick = ({ event, index }) => {
-        const { contentExplorerMode, items, onSelectItem } = this.props;
+        const { contentExplorerMode, onSelectedItems, items, onSelectItem } = this.props;
```
tjuanitas, line 315: these are sorted

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -284,11 +312,11 @@ class ContentExplorer extends Component {
...
         const { selectedItems } = this.state;
         const item = items[index];
 
-        if (item.isDisabled || item.isLoading) {
+        if (item.isDisabled || item.isLoading || item.isActionDisabled) {
```
tjuanitas, line 319: wouldn't this change existing behavior?

i.e. if the parent folder has `isActionDisabled` but the subfolders do not, then we should allow the user to click into that folder to select the subfolders

there's a demo of that here: https://opensource.box.com/box-ui-elements/#/Features?id=content-explorer
with the `Can't choose but can drill in` item

Screen Shot 2023-02-15 at 11 37 11 PM
Screen Shot 2023-02-15 at 11 37 17 PM

tjuanitas: although now I think I might misunderstand how `handleItemClick` is being called

tjuanitas: yup I think i understood this incorrectly. what you have should be right.

pusch-e: I might be a little confused. I thought this would still be fine since this is the handleitemclick and not handleItemNameClick. So users can still click into this folder to see subcontents even if they are action disabled. I thought then a new items array gets passed in as the items prop for ContentExplorer, where then those subcontents will have their own values of action disabled or disabled and they can be selected or not

pusch-e: Oops sorry replied to this a tad too late. Glad we are one same page though

src/features/content-explorer/content-explorer/ContentExplorer.js
pusch-e: Yeah so we can actually pass up the newFoldersPath which is exactly what I want. The only thing is that the folder nor the folders path is being passed up when the user clicks on the Search Results breadcrumb. In this function https://github.com/box/box-ui-elements/blob/master/src/features/content-explorer/content-explorer/ContentExplorerHeaderActions.js#L63-L82
return is called instead of onEnterFolder if the user clicks on the Search Results

tjuanitas: gotcha, I think onFoldersPath is good then

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -164,12 +178,15 @@ class ContentExplorer extends Component {
...
-        const { initialFoldersPath } = this.props;
+        const { initialFoldersPath, onFoldersPath } = this.props;
 
         if (prevInitialFoldersPath !== initialFoldersPath) {
             this.setState({
                 foldersPath: initialFoldersPath,
             });
+            if (onFoldersPath) {
+                onFoldersPath(initialFoldersPath);
+            }
```
tjuanitas, line 189: could we call `handleFoldersPathUpdated` here?

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -259,14 +280,21 @@ class ContentExplorer extends Component {
...
         this.setState(newState);
+        if (onFoldersPath) {
+            onFoldersPath(newState.foldersPath);
+        }
 
         onEnterFolder(enteredFolder, newFoldersPath);
     };
 
     handleFoldersPathUpdated = newFoldersPath => {
+        const { onFoldersPath } = this.props;
```
tjuanitas, line 291: can we rename `onFoldersPath` to `onFoldersPathUpdate` / `onFoldersPathChange`?

src/features/content-explorer/content-explorer-modal/ContentExplorerModal.js
```diff
@@ -19,14 +19,19 @@ type Props = {
...
     itemRowHeight?: number,
     itemRowRenderer?: Function,
     listHeaderHeight?: number,
     listHeaderRenderer?: Function,
+    onFoldersPath: Function,
+    onIncludeSubfoldersToggle: Function,
     onRequestClose?: Function,
     onSelectItem?: (item: Object, index: number) => void,
     onSelectedClick?: () => void,
+    onSelectedItems: Function,
```
tjuanitas, line 32: can we call this onSelectedItemsUpdate or onSelectedItemsChange?

src/features/content-explorer/content-explorer/ContentExplorer.scss
pusch-e: Yes this is much better

src/features/content-explorer/content-explorer/ContentExplorer.js
pusch-e: Alright, I made all of the react changes that you mentioned here for it to work while still being independent. It does work, the only issue is that  the two components have different css and without changing any of the css, the modal looks like this: 
Screen Shot 2023-02-16 at 2 52 18 PM

pusch-e: This is because the containers for the ContentExplorerIncludeSubfodlers component has a flex layout and the ContentExplorerSelectAll component has a display block, each with their own respective heights. Should I mess around with the  original ContennExplorerSelectAll css? Or should I try to make some new css classnames and have them conditionally set on the checkbox based on the isLabelHidden prop?

pusch-e: I am leaning towards the latter at the moment

pusch-e: Screen Shot 2023-02-16 at 3 14 50 PM
Screen Shot 2023-02-16 at 3 15 16 PM



 Got it looking like this at the moment

src/features/content-explorer/content-explorer/ContentExplorerIncludeSubfolders.js
```diff
@@ -0,0 +1,32 @@
...
+                <InfoBadge16 className="bdl-ContentExplorerIncludeSubfolders-icon" fill="blue" />
+            </Tooltip>
+        </div>
+    </div>
+);
+
+ContentExplorerIncludeSubfolders.propTypes = {
+    isToggleDisabled: PropTypes.bool,
+    onIncludeSubfoldersToggle: PropTypes.func,
+    tooltipMessageForToggle: PropTypes.object,
```
tjuanitas, line 28: we can remove toggle word from these props. since `ContentExplorerIncludeSubfolders` is a toggle component, having `toggle` also in the names is unnecessary

`isToggleDisabled` -> `isDisabled`
`onIncludeSubfoldersToggle` -> `onChange`
`tooltipMessageForToggle` -> `tooltipMessage`

src/features/content-explorer/content-explorer/ContentExplorerIncludeSubfolders.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { injectIntl, FormattedMessage } from 'react-intl';
+import Tooltip from '../../../components/tooltip';
+import Toggle from '../../../components/toggle';
+import messages from '../messages';
+import InfoBadge16 from '../../../icon/fill/InfoBadge16';
```
tjuanitas, line 7: minor:
```
import InfoBadge16 from '../../../icon/fill/InfoBadge16';
import Toggle from '../../../components/toggle';
import Tooltip from '../../../components/tooltip';
```

src/features/content-explorer/content-explorer/ContentExplorerIncludeSubfolders.js
```diff
@@ -0,0 +1,32 @@
...
+
+const ContentExplorerIncludeSubfolders = ({ isToggleDisabled, onIncludeSubfoldersToggle, tooltipMessageForToggle }) => (
+    <div className="bdl-ContentExplorerIncludeSubfolders">
+        <div className="bdl-ContentExplorerIncludeSubfolders-toggleIconContainer">
```
tjuanitas, line 11: are both of these divs still necessary? can we remove `bdl-ContentExplorerIncludeSubfolders-toggleIconContainer`?
a single div inside of a div usually isn't needed

src/features/content-explorer/content-explorer/ContentExplorer.scss
```diff
@@ -72,22 +72,56 @@
...
+
+        .bdl-ContentExplorerIncludeSubfolders-icon {
+            margin-left: 120px;
         }
 
-        .content-explorer-select-all-checkbox {
-            float: right;
-            margin-right: -21 * $bdl-grid-unit;
+        .bdl-ContentExplorerIncludeSubfolders-icon path {
+            fill: $bdl-box-blue;
```
tjuanitas, line 124: these blocks can be combined
```
.bdl-ContentExplorerIncludeSubfolders-icon {
    margin-left: 120px;

    path {
        fill: $bdl-box-blue;
    }
}
```

src/features/content-explorer/content-explorer/ContentExplorer.scss
```diff
@@ -72,22 +72,56 @@
...
+        align-items: center;
+        height: 10 * $bdl-grid-unit;
+
+        .bdl-ContentExplorerIncludeSubfolders-toggleIconContainer {
+            display: flex;
+            align-items: center;
+        }
+
+        .bdl-ContentExplorerIncludeSubfolders-icon {
+            margin-left: 120px;
```
tjuanitas, line 120: what is this margin for? shouldn't it be on the toggle or the outer div?

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -304,6 +332,10 @@ class ContentExplorer extends Component {
             newSelectedItems[item.id] = item;
         }
 
+        if (onSelectedItemsUpdate) {
+            onSelectedItemsUpdate(newSelectedItems);
+        }
```
tjuanitas, line 337: any particular reason this is before `setState`? if not, could we move it after since that's the pattern for `onFoldersPathUpdate`?

src/features/content-explorer/content-explorer/ContentExplorer.scss
```diff
@@ -72,22 +72,56 @@
...
+            margin-right: 4.45 * $bdl-grid-unit;
+        }
+    }
+
     .content-explorer-select-all-container {
-        display: block;
+        display: flex;
+        align-items: center;
+        justify-content: space-between;
+        width: 100%;
```
tjuanitas, line 94: somewhat related to this comment: https://github.com/box/box-ui-elements/pull/3242/files#r1113512295

do you need these styles? I think `margin-left: auto` on `content-explorer-select-all-checkbox-label` might be enough

src/features/content-explorer/content-explorer/ContentExplorer.scss
pusch-e: It is because the info badge icon is not a part of the label of the toggle component, so for now it needs the left margin to be positioned correctly with the toggle

src/features/content-explorer/content-explorer/ContentExplorer.scss
pusch-e: Yes, I need the width: 100% or else it does not look correct

src/features/content-explorer/content-explorer/__tests__/ContentExplorerIncludeSubfolders.test.js
```diff
@@ -0,0 +1,53 @@
...
+            expect(toggle.prop('onChange')).toEqual(onChange);
+            expect(toggle.prop('isDisabled')).toEqual(isDisabled);
+        });
+
+        test('should render toggle label correctly', () => {
+            const wrapper = renderComponent();
+
+            const toggle = wrapper.find('Toggle');
+
+            const toggleLabelId = toggle.prop('label').props.id;
```
tjuanitas, line 37:
```
            const wrapper = renderComponent();
            const toggleLabelId = wrapper.find('Toggle').prop('label').props.id;
```

you can even potentially write it directly in the `expect`

src/features/content-explorer/content-explorer/ContentExplorerIncludeSubfolders.js
```diff
@@ -0,0 +1,30 @@
...
+
+const ContentExplorerIncludeSubfolders = ({ isDisabled, onChange, tooltipMessage }) => (
+    <div className="bdl-ContentExplorerIncludeSubfolders">
+        <Toggle
+            className="bdl-ContentExplorerIncludeSubfolders-toggle"
+            label={<FormattedMessage {...messages.includeSubfolders} />}
+            isDisabled={isDisabled}
+            onChange={onChange}
+        />
+        <Tooltip text={<FormattedMessage {...tooltipMessage} />}>
```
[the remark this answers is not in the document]
pusch-e: I'll make it conditionally render

src/features/content-explorer/content-explorer/__tests__/ContentExplorerIncludeSubfolders.test.js
```diff
@@ -0,0 +1,53 @@
...
+
+            expect(toggleLabelId).toEqual('boxui.contentExplorer.includeSubfolders');
```
tjuanitas, line 39: `toBe`

src/features/content-explorer/content-explorer/__tests__/ContentExplorerIncludeSubfolders.test.js
```diff
@@ -0,0 +1,53 @@
...
+        });
+
+        test('should render the info icon Tooltip based on value from tooltipMessage', () => {
+            const tooltipMessage = { test: 'test' };
+            const wrapper = renderComponent();
+            const numOfSelectedItems = 2;
```
tjuanitas, line 45: what is this?

src/features/content-explorer/content-explorer/__tests__/ContentExplorerIncludeSubfolders.test.js
```diff
@@ -0,0 +1,53 @@
...
+            wrapper.setProps({ numOfSelectedItems, tooltipMessage });
```
tjuanitas, line 46: you can pass `tooltipMessage` inside of renderComponent()

src/features/content-explorer/content-explorer/__tests__/ContentExplorerIncludeSubfolders.test.js
```diff
@@ -0,0 +1,53 @@
...
+
+            const toggleTooltip = wrapper.find('Tooltip').at(0);
+
+            expect(toggleTooltip.prop('text').props).toEqual(tooltipMessage);
```
tjuanitas, line 50:
```suggestion
            expect(wrapper.find('Tooltip').prop('text').props).toEqual(tooltipMessage);
```

src/features/content-explorer/content-explorer/ContentExplorer.scss
```diff
@@ -72,22 +72,50 @@
...
+        display: flex;
+        align-items: center;
+
+        .checkbox-container,
+        .toggle-container {
+            margin: 0;
+        }
+
+        .content-explorer-select-all-checkbox {
+            margin-right: 4.45 * $bdl-grid-unit;
```
tjuanitas, line 85: how was this calculated? could we do 18px?

src/features/content-explorer/content-explorer/__tests__/ContentExplorerIncludeSubfolders.test.js
pusch-e: Checking that the tooltip message that is passed into it is rendered on the tooltip

src/features/content-explorer/content-explorer/ContentExplorer.scss
```diff
@@ -72,22 +72,50 @@
...
+            margin-left: auto;
+        }
+    }
+
     .content-explorer-select-all-container {
-        display: block;
-        height: 8 * $bdl-grid-unit;
+        display: flex;
+        align-items: center;
+        width: 100%;
```
tjuanitas, line 93: i think replace `width: 100%` with `flex-grow: 1` might be more appropriate

with `width: 100%` means it should have the width of the parent but what we want is only the width of the remaining space in the parent

src/features/content-explorer/content-explorer-modal/ContentExplorerModal.js
```diff
@@ -55,6 +60,7 @@ const ContentExplorerModal = ({
         <ContentExplorer
             breadcrumbProps={breadcrumbProps}
             customInput={customInput}
+            includeSubfoldersProps={includeSubfoldersProps}
```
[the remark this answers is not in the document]
pusch-e: It does not get passed through unless I include this line. When I exclude this, the prop does not go through

src/features/content-explorer/content-explorer/ContentExplorer.scss
```diff
@@ -72,22 +72,50 @@
...
         .content-explorer-select-all-items-counter {
             margin-left: 2px;
         }
 
         .content-explorer-select-all-checkbox-label {
-            float: right;
-            margin-right: 42px;
+            display: flex;
+            flex: 1;
+            justify-content: flex-end;
```
tjuanitas, line 103: I think I see what's happening and I think this should solve it:

set `margin-left: auto;` here instead of these flex styles
```suggestion
            margin-left: auto;
```

then for `.content-explorer-select-all-checkbox` on line 84, try this:
```
.content-explorer-select-all-checkbox {
    margin-right: ...

    &:first-child {
        margin-left: auto;
    }
}
```

src/features/content-explorer/content-explorer/ContentExplorer.scss
pusch-e: We could do 18 px as well. I just thought it was recommended to use bdl-grid-units since the previous code for ContentExplorerSelectAll uses it

src/features/content-explorer/content-explorer/__tests__/ContentExplorerIncludeSubfolders.test.js
tjuanitas: what about the `numOfSelectedItems`? that's not a prop right?

pusch-e: Yes sorry that is an artifact. I will delete

src/features/content-explorer/content-explorer/ContentExplorer.scss
tjuanitas: okay I see what's happening here now. we can't have this fixed margin because it won't work for different languages. i.e. the text would be different lengths so we can't rely on a fixed margin

we'll have to put the icon into the label like this: https://github.com/box/box-ui-elements/blob/master/src/features/advanced-content-insights/AdvancedContentInsightsToggle.tsx#L33-L55

src/features/content-explorer/content-explorer/ContentExplorer.scss
pusch-e: Yes this also works

src/features/content-explorer/content-explorer-modal/ContentExplorerModal.js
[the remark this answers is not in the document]
tjuanitas: was it removed from line 43 as well as here? i think that should include in the `rest`

tjuanitas (commented):
one last thing. thank you for working through this!

src/features/content-explorer/content-explorer/__tests__/ContentExplorerIncludeSubfolders.test.js
```diff
@@ -0,0 +1,51 @@
...
+
+        test('should render the info icon Tooltip based on value from tooltipMessage', () => {
+            const tooltipMessage = { test: 'test' };
+            const wrapper = renderComponent({ tooltipMessage });
+            const expectedMessage = wrapper
+                .find('Toggle')
+                .dive()
+                .find('Tooltip')
+                .prop('text').props;
+            console.log(expectedMessage);
```
tjuanitas, line 46: console.log

pusch-e: oops

tjuanitas (approved):

menaczar (approved):

PR #3243, opened 2023-02-13
fix(content-picker): add disabled

jpan-box:
Add `disabled` attribute to picker choose button when no items are selected.

Screenshot 2023-02-13 at 6 13 19 PM

### DIFF
--- a/src/elements/content-picker/Footer.js
+++ b/src/elements/content-picker/Footer.js
@@ -59,6 +59,7 @@ const Footer = ({
 }: Props) => {
     const cancelMessage = intl.formatMessage(messages.cancel);
     const chooseMessage = intl.formatMessage(messages.choose);
+    const isChooseButtonDisabled = !selectedCount;
     return (
         <footer className="bcp-footer">
             <div className="bcp-footer-left">
@@ -96,10 +97,11 @@ const Footer = ({
                                 <IconClose height={16} width={16} />
                             </Button>
                         </Tooltip>
-                        <Tooltip isDisabled={!selectedCount} text={chooseButtonLabel || chooseMessage}>
+                        <Tooltip isDisabled={isChooseButtonDisabled} text={chooseButtonLabel || chooseMessage}>
                             <PrimaryButton
                                 aria-label={chooseMessage}
-                                isDisabled={!selectedCount}
+                                disabled={isChooseButtonDisabled} // sets disabled attribute
+                                isDisabled={isChooseButtonDisabled} // used in Button component
                                 onClick={onChoose}
                                 type="button"
                             >
--- a/src/elements/content-picker/__tests__/Footer.test.js
+++ b/src/elements/content-picker/__tests__/Footer.test.js
@@ -25,6 +25,16 @@ describe('elements/content-picker/Footer', () => {
             expect(wrapper.find('.footer-child').length).toBe(1);
         });
 
+        test('should render footer with disabled button', () => {
+            const buttons = getWrapper().find('Button');
+            const chooseButton = buttons.at(1);
+
+            // https://www.w3.org/WAI/ARIA/apg/patterns/button/
+            // When the action associated with a button is unavailable, the button has aria-disabled set to true.
+            expect(chooseButton.html().includes('aria-disabled')).toBe(true);
+            expect(chooseButton.prop('disabled')).toBe(true);
+        });
+
         test('should render Footer buttons with aria-label', () => {
             const buttons = getWrapper().find('Button');
 


src/elements/content-picker/__tests__/Footer.test.js
```diff
@@ -25,6 +25,19 @@ describe('elements/content-picker/Footer', () => {
...
+        test('should render footer with disabled button', () => {
+            // Wrapper with two nodes
+            const buttons = getWrapper().find('Button');
+
+            // Select the 'Choose' Button
+            const chooseButton = buttons.at(1);
+
+            // https://www.w3.org/WAI/ARIA/apg/patterns/button/
+            // When the action associated with a button is unavailable, the button has aria-disabled set to true.
+            expect(chooseButton.html().includes('aria-disabled')).toBe(true);
```
tjuanitas, line 37: does `prop()` work?

src/elements/content-picker/Footer.js
```diff
@@ -102,6 +102,7 @@ const Footer = ({
                                 isDisabled={!selectedCount}
                                 onClick={onChoose}
                                 type="button"
+                                disabled={!selectedCount}
```
tjuanitas, line 105: could we `abc` this?

tjuanitas, line 105: we might want to make an `isChooseButtonDisabled` variable now. i see three usages

src/elements/content-picker/Footer.js
```diff
@@ -102,6 +102,7 @@ const Footer = ({
                                 isDisabled={!selectedCount}
```
QingyuChai, line 102: feels kinda weird that `isDisabled` and `disabled` don't do the same thing, but I think it's too risky combining them, since we've been using `isDisabled` for a while, without setting `disabled` on the html element.
https://github.com/box/box-ui-elements/blob/fef7991a3e75ef8e437e1c2293ed1c92de907861/src/components/button/Button.tsx#L49-L58

QingyuChai (approved):
lgtm!

src/elements/content-picker/__tests__/Footer.test.js
```diff
@@ -25,6 +25,19 @@ describe('elements/content-picker/Footer', () => {
             expect(wrapper.find('.footer-child').length).toBe(1);
         });
 
+        test('should render footer with disabled button', () => {
+            // Wrapper with two nodes
```
tjuanitas, line 29: this comment isn't really necessary and could become outdated if buttons are added or removed

src/elements/content-picker/__tests__/Footer.test.js
jpan-box: it did not, when i printed out props i got only what was explicitly set in Footer
```
    {
      className: 'btn-primary ',
      'aria-label': 'Choose',
      isDisabled: true,
      onClick: [Function: onChoose],
      type: 'button',
      disabled: true,
      children: {
        '$$typeof': Symbol(react.element),
        type: [Function: IconCheck],
        key: null,
        ref: null,
        props: { color: '#fff', height: 16, width: 16 },
        _owner: FiberNode {
          tag: 0,
          key: null,
          elementType: [Function: Footer],
          type: [Function: Footer],
          stateNode: null,
          return: [FiberNode],
          child: [FiberNode],
          sibling: null,
          index: 0,
          ref: null,
          pendingProps: [Object],
          memoizedProps: [Object],
          updateQueue: null,
          memoizedState: null,
          dependencies: null,
          mode: 0,
          flags: 1,
          nextEffect: null,
          firstEffect: [FiberNode],
          lastEffect: [FiberNode],
          lanes: 0,
          childLanes: 1,
          alternate: [FiberNode],
          actualDuration: 0,
          actualStartTime: -1,
          selfBaseDuration: 0,
          treeBaseDuration: 0,
          _debugID: 250,
          _debugSource: null,
          _debugOwner: [FiberNode],
          _debugNeedsRemount: false,
          _debugHookTypes: null
        },
        _store: {}
      },
      isLoading: false,
      showRadar: false
    }

```

i couldn't figure out a way to get at the aria-disabled prop set within button itself

src/elements/content-picker/Footer.js
jpan-box: yarp

jpan-box: yarp

src/elements/content-picker/__tests__/Footer.test.js
jpan-box: yes you're right, these are more or less leftover from me explaining it to myself as i went along

src/elements/content-picker/Footer.js
```diff
@@ -96,10 +97,11 @@ const Footer = ({
                                 <IconClose height={16} width={16} />
                             </Button>
                         </Tooltip>
-                        <Tooltip isDisabled={!selectedCount} text={chooseButtonLabel || chooseMessage}>
+                        <Tooltip isDisabled={isChooseButtonDisabled} text={chooseButtonLabel || chooseMessage}>
                             <PrimaryButton
                                 aria-label={chooseMessage}
-                                isDisabled={!selectedCount}
+                                disabled={isChooseButtonDisabled} // sets disabled attribute
+                                isDisabled={isChooseButtonDisabled} // used in Button component
```
tjuanitas, line 104: lol these comments aren't really necessary either

tjuanitas (approved):

QingyuChai (approved):

src/elements/content-picker/Footer.js
QingyuChai: I think we do need them, because otherwise ppl will be confused about what the difference is

tjuanitas: yeah you're right

PR #3244, opened 2023-02-14
fix: Fix aria-label for BE header search

tjuanitas:
Matches the search inputs `aria-label` and `placeholder` messages

### DIFF
--- a/src/elements/common/header/Header.js
+++ b/src/elements/common/header/Header.js
@@ -25,6 +25,7 @@ type Props = {
 
 // eslint-disable-next-line react/prop-types
 const Header = ({ isHeaderLogoVisible = true, view, isSmall, searchQuery, onSearch, logoUrl, intl }: Props) => {
+    const { formatMessage } = intl;
     const search = ({ currentTarget }: { currentTarget: HTMLInputElement }) => onSearch(currentTarget.value);
     const isFolder = view === VIEW_FOLDER;
     const isSearch = view === VIEW_SEARCH;
@@ -33,10 +34,11 @@ const Header = ({ isHeaderLogoVisible = true, view, isSmall, searchQuery, onSear
             {isHeaderLogoVisible && <Logo isSmall={isSmall} url={logoUrl} />}
             <div className="be-search">
                 <input
-                    aria-label="search"
+                    aria-label={formatMessage(messages.searchPlaceholder)}
+                    data-testid="be-search-input"
                     disabled={!isFolder && !isSearch}
                     onChange={search}
-                    placeholder={intl.formatMessage(messages.searchPlaceholder)}
+                    placeholder={formatMessage(messages.searchPlaceholder)}
                     type="search"
                     value={searchQuery}
                 />
@@ -45,4 +47,5 @@ const Header = ({ isHeaderLogoVisible = true, view, isSmall, searchQuery, onSear
     );
 };
 
+export { Header as HeaderBase };
 export default injectIntl(Header);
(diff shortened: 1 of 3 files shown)


src/elements/common/header/Header.js
```diff
@@ -33,10 +34,11 @@ const Header = ({ isHeaderLogoVisible = true, view, isSmall, searchQuery, onSear
             {isHeaderLogoVisible && <Logo isSmall={isSmall} url={logoUrl} />}
             <div className="be-search">
                 <input
-                    aria-label="search"
+                    aria-label={formatMessage(messages.searchPlaceholder)}
+                    data-testid="be-search-input"
                     disabled={!isFolder && !isSearch}
                     onChange={search}
-                    placeholder={intl.formatMessage(messages.searchPlaceholder)}
+                    placeholder={formatMessage(messages.searchPlaceholder)}
```
QingyuChai, line 41: Since it's being used in two places, I'd prefer we use a variable, but this is fine as well

src/elements/common/header/__tests__/Header.test.js
```diff
@@ -0,0 +1,32 @@
...
+
+    test('renders Logo component when isHeaderLogoVisible is passed', () => {
+        const wrapper = getWrapper({ isHeaderLogoVisible: true });
+        expect(wrapper.find('Logo').exists()).toBe(true);
+    });
+
+    test('renders matching values for aria-label and placeholder attributes', () => {
+        const wrapper = getWrapper();
+        const inputProps = wrapper.find('[data-testid="be-search-input"]').props();
+        expect(inputProps['aria-label']).toBe(inputProps.placeholder);
```
QingyuChai, line 20: minor: you're using using `inputProps['aria-label']`, but `inputProps.placeholder`. They're both fine, but I think we should be consistent

tjuanitas: yeah this was an eslint thing. it prefers dot notation but it can't be used for the aria-label

QingyuChai: oh that's interesting, thanks for the info

src/elements/common/header/__tests__/Header.test.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { shallow } from 'enzyme';
```
JChan106, line 2: Think we're trying to switch away from enzyme, can we use testing-library instead

src/elements/common/header/Header.js
```diff
@@ -45,4 +49,5 @@ const Header = ({ isHeaderLogoVisible = true, view, isSmall, searchQuery, onSear
     );
 };
 
+export { Header as HeaderBase };
```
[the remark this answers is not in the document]
tjuanitas: this is so the base component can be tested without the intl injection stuff

src/elements/common/header/__tests__/Header.test.js
```diff
@@ -0,0 +1,32 @@
...
+        formatMessage: jest.fn().mockImplementation(message => message.defaultMessage),
+    };
+
+    const getWrapper = props => shallow(<Header intl={intl} {...props} />);
+
+    test('renders Logo component when isHeaderLogoVisible is passed', () => {
+        const wrapper = getWrapper({ isHeaderLogoVisible: true });
+        expect(wrapper.find('Logo').exists()).toBe(true);
+    });
+
```
karelee7, line 16: Should we also add a test for the negative case above? (i.e. doesn't render logo when `isHeaderLogoVisible` is not passed or falsey)

karelee7 (approved):
Looks good to me!

QingyuChai (approved):

PR #3246, opened 2023-02-14
fix(content-explorer): enable create folder with trailing space in name

TaoJiang98:

### DIFF
--- a/src/elements/content-explorer/ContentExplorer.js
+++ b/src/elements/content-explorer/ContentExplorer.js
@@ -1213,7 +1213,7 @@ class ContentExplorer extends Component<Props, State> {
         this.setState({ isLoading: true });
         this.api.getFolderAPI().create(
             id,
-            name,
+            name.trim(),
             (item: BoxItem) => {
                 this.refreshCollection();
                 this.select(item);
--- a/src/elements/content-explorer/ContentExplorer.js
+++ b/src/elements/content-explorer/ContentExplorer.js
@@ -1142,7 +1142,7 @@ class ContentExplorer extends Component<Props, State> {
         this.setState({ isLoading: true });
         this.api.getAPI(type).rename(
             selected,
-            name,
+            name.trim(),
             (updatedItem: BoxItem) => {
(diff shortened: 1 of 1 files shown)


src/elements/content-explorer/ContentExplorer.js
```diff
@@ -1213,7 +1213,7 @@ class ContentExplorer extends Component<Props, State> {
         this.setState({ isLoading: true });
         this.api.getFolderAPI().create(
             id,
-            name,
+            name.trim(),
```
TaoJiang98, line 1216: Adding trim() function to remove the trailing whitespace when create button has been clicked. Similar to how EUA handles this situation. Ref: https://git.dev.box.net/Box/EndUserApp/blob/11365198c9e285203372204f25bb8d4e387caf01/src/components/modals/create-folder-modal/components/CreateFolderModal.js#L249

tjuanitas: I noticed there's an existing bug on line 1197, could we add a `trim()` there as well? it doesn't look like that `if` would ever enter because of the `if` right before it.

https://github.com/box/box-ui-elements/blob/6c68ae94753a136b4f00cdd9af116f9ea8b1fae2/src/elements/content-explorer/ContentExplorer.js#L1197

this way it'll be a little similar to rename functionality:
https://github.com/box/box-ui-elements/blob/6c68ae94753a136b4f00cdd9af116f9ea8b1fae2/src/elements/content-explorer/ContentExplorer.js#L1134

QingyuChai:
Note: edited the description to remove jira link

src/elements/content-explorer/ContentExplorer.js
TaoJiang98: resolved!

tjuanitas (approved):
LGTM!

QingyuChai (approved):
lgtm! Just wondering, is there any way to add this to the tests somewhere? Just a check to make sure we are trimming the name?

PR #3248, opened 2023-02-14
fix(content-uploader): make disabled buttons unfocusable

QingyuChai:
Similar fix to this: https://github.com/box/box-ui-elements/pull/3243/files

Essentially: The `isDisabled` prop does not set the `disabled` property, which means that it is still focusable by the keyboard.

Potential idea to have `isDisabled` set the `disabled` property as well, but because it might cause unintended consequences in parent apps we won't do so here.

https://www.w3.org/WAI/ARIA/apg/patterns/button/

image

### DIFF
--- a/src/elements/content-uploader/Footer.js
+++ b/src/elements/content-uploader/Footer.js
@@ -39,7 +39,7 @@ const Footer = ({ isLoading, hasFiles, errorCode, onCancel, onClose, onUpload, f
         <div className="bcu-footer">
             <div className="bcu-footer-left">
                 {onClose ? (
-                    <Button isDisabled={hasFiles} onClick={onClose} type="button">
+                    <Button isDisabled={hasFiles} disabled={hasFiles} onClick={onClose} type="button">
                         <FormattedMessage {...messages.close} />
                     </Button>
                 ) : null}
--- a/src/elements/content-uploader/Footer.js
+++ b/src/elements/content-uploader/Footer.js
@@ -23,6 +23,9 @@ type Props = {
 };
 
 const Footer = ({ isLoading, hasFiles, errorCode, onCancel, onClose, onUpload, fileLimit, isDone }: Props) => {
+    const isCancelButtonDisabled = !hasFiles || isDone;
+    const isUploadButtonDisabled = !hasFiles;
+
     let message;
     switch (errorCode) {
         case ERROR_CODE_UPLOAD_FILE_LIMIT:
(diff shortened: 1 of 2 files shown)


src/elements/content-uploader/Footer.js
```diff
@@ -23,6 +23,10 @@ type Props = {
 };
 
 const Footer = ({ isLoading, hasFiles, errorCode, onCancel, onClose, onUpload, fileLimit, isDone }: Props) => {
+    const isCloseButtonDisabled = hasFiles;
+    const isCancelButtonDisabled = !hasFiles || isDone;
+    const isUploadButtonDisabled = !hasFiles;
```
QingyuChai, line 28: moved the closeButton and uploadButton disabled booleans up as well, for readability. Not necessary though

tjuanitas (approved):

megansmith-box (approved):

src/elements/content-uploader/Footer.js
jpan-box: i dig it

jpan-box (approved):

src/elements/content-uploader/Footer.js
```diff
@@ -36,17 +40,33 @@ const Footer = ({ isLoading, hasFiles, errorCode, onCancel, onClose, onUpload, f
         <div className="bcu-footer">
             <div className="bcu-footer-left">
                 {onClose ? (
-                    <Button isDisabled={hasFiles} onClick={onClose} type="button">
+                    <Button
+                        disabled={isCloseButtonDisabled}
```
JChan106, line 44: what's the difference between disabled and isDisabled?

src/elements/content-uploader/__tests__/Footer.test.js
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import noop from 'lodash/noop';
```
JChan106, line 3: move lodash before enzyme

src/elements/content-uploader/__tests__/Footer.test.js
```diff
@@ -0,0 +1,59 @@
...
+import Button from '../../../components/button/Button';
+import PrimaryButton from '../../../components/primary-button/PrimaryButton';
+import Footer from '../Footer';
```
JChan106, line 6: move footer before primary button

src/elements/content-uploader/Footer.js
QingyuChai: isDisabled sets the styles/classes for the button, while disabled is passed onto the html element <button />, which prevents focus on the element

src/elements/content-uploader/__tests__/Footer.test.js
QingyuChai: I think we try to follow a convention where react packages are at the top, and we alphasort the rest

JChan106 (approved):
lgtm

tjuanitas (approved):

PR #3251, opened 2023-02-17
fix(content-sharing): removed api condition when constructing api

greg-in-a-box:
stories for this element was not working based on the paragraph in the story because it was not regenerating a new API object when the knobs changed.  After further investigation, the API object should always be provided if the props are not provided to the API constructor.  Error handling happens in the functions

### DIFF
--- a/src/elements/content-sharing/ContentSharing.js
+++ b/src/elements/content-sharing/ContentSharing.js
@@ -75,10 +75,10 @@ function ContentSharing({
 
     // Reset the API if necessary
     React.useEffect(() => {
-        if (!api && apiHost && itemID && itemType && token) {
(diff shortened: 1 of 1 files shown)


tjuanitas (approved):

sophiayangyxy (approved):

Machoper (approved):

tjuanitas (changes requested):
slack thread as information that checking `!api` might be necessary (although not the way we're expecting it to be)

in its current form, it looks like this `useEffect` does not do anything because `api` shouldn't be falsey. but removing the `!api` will change the previous behavior and call `setAPI` more, which could introduce a bug

tjuanitas (approved):

Issue #3100, opened 2022-10-19
Update node.js to v18 and npm to v8


ASDAlexey:
We use box-ui-elements as an external lib in our project.
Box lib does not allow us to update our project node.js to v18 and npm to v8. Please update box-ui-elements engines to latest node.js

micahjonas:
Same issue for us. We use box for our software platform and were using box-ui-elements version 15.0.0 as part of that. I wanted to update to version 16.0.0 but it's not possible because we are using node v18 & npm v8.

csutorasr:
The required engines are only needed for development, but not while using the release.

Can you remove that part of the `package.json` before uploading the package?

csutorasr:
I made this PR to remove that part of the `package.json` before uploading to the registry. I need a review on that.

@greg-in-a-box you are show here as a participant. Can you check it or ask any other maintainer to do it?

ASDAlexey:
> ci: remove package.json engines from production build

no need to remove engines, need to update it latest support like:
"node": "^14.20.0 || ^16.13.0 || >=18.10.0"

For that need to check project update all libs in project to support  ^14.20.0 || ^16.13.0 || >=18.10.0"

This feature need not only for development, for release it needed to because we make build based on node version in engines section package.json

All modern libs(angular, react) have section need update this to and check all project tests for all works fine - it is necessary that the box-ui-elements command does this periodically. This is a good tone of libraries
"engines": {
      "node": "^14.20.0 || ^16.13.0 || >=18.10.0",

ASDAlexey:
For update to angular 16 need minimum node.js 16.3(https://github.com/angular/angular/blob/16.0.0-next.2/package.json)
In 1 month angular 16 will be released plz update box.com 
Only because of we use box.com we can't update angular to latest version

andrew-caines:
Is there a plan to update the box-elements-ui to support LTS nodejs and React 18.  Both of those are defacto standard at this point. (both LTS).  Currently, it will not install.

code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR!
npm ERR! While resolving: client@0.0.0
npm ERR! Found: react@18.2.0
npm ERR! node_modules/react
npm ERR!   react@"^18.2.0" from the root project
npm ERR!
npm ERR! Could not resolve dependency:
[...]

PR #3252, opened 2023-02-20
ci: remove package.json engines from production build

csutorasr:
This dependencies are only required for development.

Closes Update node.js to v18 and npm to v8 #3100

### DIFF
--- a/scripts/cut_release.sh
+++ b/scripts/cut_release.sh
@@ -198,6 +198,10 @@ check_npm_login() {
     fi
 }
 
+remove_package_json_engines() {
+    ./scripts/utils/removeEnginesFromPackageJson || return 1
+}
+
 push_new_release() {
     # Check branch being dirty
     check_branch_dirty || return 1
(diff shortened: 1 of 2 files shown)


ASDAlexey:
no need to remove engines, need to update it latest support like:
"node": "^14.20.0 || ^16.13.0 || >=18.10.0"

For that need to check project update all libs in project to support ^14.20.0 || ^16.13.0 || >=18.10.0"

This feature need not only for development, for release it needed to because we make build based on node version in engines section package.json

All modern libs(angular, react) have section need update this to and check all project tests for all works fine - it is necessary that the box-ui-elements command does this periodically. This is a good tone of libraries
"engines": {
"node": "^14.20.0 || ^16.13.0 || >=18.10.0",

Before update lib in my project I always look for engines and if it's ok for me I update to new version

greg-in-a-box:
We cannot simply remove `engines` from the `package.json` nor are we ready to support node LTS.  Supporting LTS is on the roadmap but we dont know when exactly.  When we are ready to support node LTS, we will update `engines`

PR #3253, opened 2023-02-21
fix(scss): Replace all division for dart-sass

TaoJiang98:

### DIFF
--- a/src/components/badgeable/Badgeable.scss
+++ b/src/components/badgeable/Badgeable.scss
@@ -4,7 +4,7 @@
 $badgeable-position-offset: 1px;
 $badgeable-icon-size: 14px;
 
-$badgeable-position: ($badgeable-icon-size/2);
+$badgeable-position: ($badgeable-icon-size * .5);
 
 %badge {
     position: absolute;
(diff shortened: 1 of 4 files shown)


src/styles/common/_buttons.scss
```diff
@@ -56,7 +56,7 @@
 
         .bdl-btn-icon {
             // Icon without text:
-            margin: ($bdl-btn-height - $bdl-btn-icon-size) / 2 - $bdl-btn-border-width;
+            margin: ($bdl-btn-height - $bdl-btn-icon-size) * .5 - $bdl-btn-border-width;
```
TaoJiang98, line 59: Used Sass-Migrator to auto replaced all divisions. For the case "/ 2", it will change it to "* .5". Otherwise, it will change to math.div(*, *).

Ref: https://sass-lang.com/documentation/breaking-changes/slash-div

src/components/badgeable/Badgeable.scss
```diff
@@ -4,7 +4,7 @@
 $badgeable-position-offset: 1px;
 $badgeable-icon-size: 14px;
 
-$badgeable-position: ($badgeable-icon-size/2);
+$badgeable-position: ($badgeable-icon-size * .5);
```
tjuanitas, line 7: not blocking but doesn't look like we need these parenthesis

tjuanitas (approved):
LGTM

greg-in-a-box (approved):

PR #3254, opened 2023-02-21
fix(toggle): add switch role

tjuanitas:
Improves accessibility for the component

https://www.w3.org/WAI/ARIA/apg/patterns/switch/examples/switch-checkbox/


src/components/toggle/Toggle.js
```diff
@@ -53,15 +53,15 @@ const Toggle = React.forwardRef<Props, HTMLInputElement>(
...
+
+        const toggleElements = [
             <div key="toggle-simple-switch" className="toggle-simple-switch" />,
             <div key="toggle-simple-label" className="toggle-simple-label">
                 {label}
             </div>,
         ];
-
         if (isToggleRightAligned) {
-            toggleElements = toggleElements.reverse();
```
tjuanitas, line 64: this reassignment doesn't do anything because `reverse()` mutates the original array so it's assigning it to itself: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reverse#the_reverse_method_returns_the_reference_to_the_same_array

greg-in-a-box (approved):

TaoJiang98 (approved):

PR #3257, opened 2023-02-22
fix(flyout): increase flyout overlay css specificity

boxithai:
This fixes an issue with the responsive overlay CSS being applied other .overlay elements that are within the **Flyout** **Overlay**. One example is when we have a **SelectorDropdown** within a **Flyout**. **SelectorDropdown** defines its own `.overlay` element and the responsive behaviors defined for **FlyoutOverlay** is spilling over to other unintended elements.

The chosen solution to increase the specificity was to add `.flyout-overlay` class name and restrict the CSS to the direct child descendent
`.flyout-overlay > .overlay`

Another accepted approach might have been to rename the **Flyout** and **Overlay** components to match SUIT naming but that comes with inherent risks as there are other CSS classes that override `.overlay`.

Observed Bug
flyout-responsive-css-issue

Fixed
flyout-responsive-fix

### DIFF
--- a/src/components/flyout/Flyout.scss
+++ b/src/components/flyout/Flyout.scss
@@ -32,7 +32,7 @@
             transform: none !important;
         }
 
-        .overlay {
+        .bdl-Overlay > .overlay {
             position: fixed;
             top: 0;
             right: 0;
--- a/src/components/flyout/Overlay.js
+++ b/src/components/flyout/Overlay.js
@@ -1,5 +1,6 @@
 // @flow
 import * as React from 'react';
+import classNames from 'classnames';
 import omit from 'lodash/omit';
 
 import FocusTrap from '../focus-trap';
@@ -35,7 +36,7 @@ class Overlay extends React.Component<Props> {
     render() {
         const { children, className, ...rest } = this.props;
         const overlayProps = omit(rest, ['onClose']);
-        overlayProps.className = className;
+        overlayProps.className = classNames('bdl-Overlay', className);
         overlayProps.handleOverlayKeyDown = this.handleOverlayKeyDown;
         overlayProps.tabIndex = 0;
 
--- a/src/components/flyout/__tests__/Overlay.test.js
+++ b/src/components/flyout/__tests__/Overlay.test.js
@@ -29,7 +29,7 @@ describe('components/flyout/Overlay', () => {
             expect(wrapper.childAt(0).is('FocusTrap')).toBe(true);
             expect(wrapper.find('div.overlay').length).toBe(1);
 
-            expect(wrapper.childAt(0).prop('className')).toEqual('hey');
+            expect(wrapper.childAt(0).prop('className')).toEqual('bdl-Overlay hey');
             expect(wrapper.childAt(0).prop('id')).toEqual('overlay');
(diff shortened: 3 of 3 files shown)


src/components/flyout/Flyout.scss
```diff
@@ -32,7 +32,7 @@
             transform: none !important;
         }
 
-        .overlay {
+        & > .flyout-overlay > .overlay {
```
tjuanitas, line 35: wouldn't this be confusing that there are two `flyout-overlay` classes on separate elements?

src/components/flyout/Flyout.js
```diff
@@ -408,6 +408,7 @@ class Flyout extends React.Component<Props, State> {
         }
 
         const overlayProps = {
+            className: 'flyout-overlay',
```
tjuanitas, line 411: wouldn't styles of the previously defined `flyout-overlay` class now apply to this overlay?

tjuanitas: maybe this should be `overlay-content` / `flyout-overlay-content` since the class is applied to the overlayContent component?

src/components/flyout/Flyout.scss
```diff
@@ -32,7 +32,7 @@
             transform: none !important;
         }
 
-        .overlay {
+        & > .flyout-overlay-content > .overlay {
```
tjuanitas, line 35: does this work?
```suggestion
        .flyout-overlay-content > .overlay {
```

ivanthai: I updated it to add .overlay-content inside of the Overlay class instead. I didn't choose .overlay-wrapper because it conflicted with other css definitions

ivanthai: To answer your question...yes that does work

src/components/flyout/Overlay.js
```diff
@@ -35,7 +36,7 @@ class Overlay extends React.Component<Props> {
     render() {
         const { children, className, ...rest } = this.props;
         const overlayProps = omit(rest, ['onClose']);
-        overlayProps.className = className;
+        overlayProps.className = classNames('overlay-content', className);
```
ivanthai, line 39:
```suggestion
        overlayProps.className = classNames('bdl-Overlay', className);
```

ivanthai: @tjuanitas thoughts on adding this...at the risk of modifying a component to have inconsistent naming styling?

tjuanitas: yeah that change seems appropriate to me

any particular reason the class name was moved from `Flyout.js` to `Overlay.js`?

ivanthai: I decided to move it there because making that change in Flyout seemed more complicated. In there, we duplicate the Overlay child and override the props. Making an update to Overlay's classname will now require me too grab the original className so I can make sure that it is propagated properly

```
const overlayProps = {
            className: classNames('flyout-overlay-content', overlayContent.className), // overlayContent.className doesn't work btw, just using it to prove my point
...
}
```

tjuanitas (approved):

ivanthai (approved):

PR #3262, opened 2023-02-24
fix(selector-dropdown): add dynamic positioning to the dropdown overlay

ivanthai:
### Updates
- Added explanation of the design decision of adding the `isPopperEnabled` prop to preserve the former **SelectorDropdown** behavior
- Added `forwardRef` to **SearchForm** because it is used as a child to **PopperComponent** to **SelectorDropdown**

### Overview
**SelectorDropdown** is not responsive in terms of the dropdown positioning. i.e. when the dropdown doesn't have enough space between the triggering element and the bottom viewport, it doesn't render above the triggering element instead. This change was needed because there is at least one element that doesn't render well in small viewports

### Solution
Utilize react-popper to handle the rendering of the overlay. This required adding forwardRefs to the respective children components. `forwardRef` and `flow` don't get along together well so a few 

To allow backward compatibility with other components that do not want to use **react-popper** which would change the overlay to use use absolute positioning, `isPopperEnabled` was added. This way, components like **TemplateDropdown** can remain compatible with **SelectorDropdown**. **TemplateDropdown** embeds a **PopperComponent** within a **Flyout** which would essentially be embedding an absolute within an absolute without a defined height.

An alternative to adding the `isPopperEnabled` prop is to have **TemplateDropdown** override the behavior using CSS which was a bit simplier, but more hacky

```
// TemplateDropdown.scss
.metadata-instance-editor-template-dropdown-menu {
    .overlay-wrapper {
        // override react-popper
        position: relative ! important;
        transform: none ! important;
...
```

### Demo

**Issue - dropdown renders below the viewport**
image

**Fix - dropdown renders above the triggering element when there is no space**
content-filter-owners-search-fix

**Smallest supported viewport**
image


src/components/pill-selector-dropdown/__tests__/PillSelectorDropdown.test.js
```diff
@@ -45,17 +45,6 @@ describe('components/pill-selector-dropdown/PillSelectorDropdown', () => {
             expect(selectorDropdown.contains(children)).toBe(true);
         });
 
-        test('should render pill selector', () => {
```
ivanthai, line 48: This test was not only causing some problems, but is also unnecessary as this functionality is already tested in `SelectorDropdown.test.js`

src/components/pill-selector-dropdown/PillSelector.js
```diff
@@ -54,6 +55,7 @@ class PillSelector extends React.Component<Props, State> {
         error: '',
         inputProps: {},
         placeholder: '',
+        // $FlowFixMe defaultProps does not propagate to forwardRef components
```
greathmaster, line 58: Curious about this...according to this Flow article:

https://medium.com/flow-type/supporting-react-forwardref-and-beyond-f8dd88f35544

It seems `forwardRef` support was added in version `0.89` and looking at `flow-bin` we're running at least `0.95`?

Anyway we can get Flow to work with the `forwardRef`?

src/components/pill-selector-dropdown/__tests__/PillSelectorDropdown.test.js
greathmaster: Curious on what problems this was causing?

src/components/selector-dropdown/SelectorDropdown.js
```diff
@@ -311,13 +312,13 @@ class SelectorDropdown extends React.Component<Props, State> {
                 onPaste={this.handleInput}
                 ref={this.selectorDropdownRef}
             >
-                {React.cloneElement(selector, { inputProps })}
-                {isOpen && (
-                    <div className={`${OVERLAY_WRAPPER_CLASS} is-visible`}>
+                <PopperComponent placement={PLACEMENT_BOTTOM_START} isOpen={isOpen}>
```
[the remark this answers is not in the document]
ivanthai: It will have the same behavior as the SelectorField elements above - it render on top of the triggering element. There is a min/max height and once it crosses that threshold, it will begin to be clipped by the viewport. This is entirely dependent on the individual max height, but so far I don't see any issues on "Small" viewport sizes (i.e. 400px height)

For example, at under 400px height, you already run into other issues. I guess we don't expect the best experience under 400px height in general
image

src/components/pill-selector-dropdown/__tests__/PillSelectorDropdown.test.js
ivanthai: The test didn't seem to work and I couldn't wrap my head around it and then I realized that it was already tested in the SelectorDropdown so it didn't seem necessary to continue. 

However, I just took a second look at it. I found a bug in this test which I didn't realize at first and it was causing me headaches. It showed up after modifying PIllSelector to return a forwardRef component. I just fixed the bug and added the test back in there.

ivanthai: I guess it was a good exercise!

src/components/pill-selector-dropdown/PillSelector.js
ivanthai: That article is helpful. Thanks. I managed to get it to work!

greathmaster (dismissed):

src/components/pill-selector-dropdown/PillSelector.js
```diff
@@ -26,12 +26,13 @@ type Props = {
     getPillClassName?: (option: Option) => string,
     /** Function to retrieve the image URL associated with a pill */
     getPillImageUrl?: (data: { id: string | number, [key: string]: any }) => string,
+    innerRef?: React.Ref<any>,
     inputProps: Object,
     onInput: Function,
     onRemove: Function,
     onSuggestedPillAdd?: Function,
     placeholder: string,
-    selectedOptions: Array<Object> | List<Object>,
```
tjuanitas, line 34: why is this removed?

src/components/selector-dropdown/SelectorDropdown.js
```diff
@@ -311,13 +312,13 @@ class SelectorDropdown extends React.Component<Props, State> {
...
+                    {React.cloneElement(selector, { inputProps })}
+                    <div className="SelectorDropdown-overlay-wrapper">
```
tjuanitas, line 317: descendant element would be camelcase `SelectorDropdown-overlayWrapper`

but maybe wrapper isn't necessary: `SelectorDropdown-overlay`

src/components/selector-dropdown/SelectorDropdown.scss
```diff
@@ -13,7 +13,7 @@
         @include bdl-Overlay-container;
     }
 
-    .overlay-wrapper,
```
tjuanitas, line 16: removing this isn't safe. Admin Console has some usages like `.SelectorDropdown .overlay-wrapper`.

you would need to audit and fix those in the parent apps if this is merged

src/components/pill-selector-dropdown/PillSelector.js
ivanthai: More specifically, only Array<Object> was being removed. Flow was complaining of ambiguity of the type of the value assigned to selectedOptions in defaultProps. 

```
Could not decide which case to select. Since case 1 [1] may work but if it doesn't case 2 [2] looks promising too. To
fix add a type annotation to unknown element type of empty array [3].

 [1][2] 57│     selectedOptions: Array<Object> | List<Object>,
          :
        67│         error: '',
        68│         inputProps: {},
        69│         placeholder: '',
    [3] 70│         selectedOptions: [],
        71│         validator: () => true,
        72│     };
        73│

```

Additionally, there are functions called on selectedOptions that are only supported in List and not in Array.

```
Property size is missing in array type [1].

 [1] 35│     selectedOptions: Array<Object>,
       :
     79│     getNumSelected = (): number => {
     80│         const { selectedOptions } = this.props;
     81│
     82│         return typeof selectedOptions.size === 'number' ? selectedOptions.size : selectedOptions.length;
     83│     };
     84│
     85│     getPillsByKey = (key: string): Array<any> => {

```

src/components/selector-dropdown/SelectorDropdown.scss
ivanthai: thanks for catching that. I added it back in

src/features/metadata-instance-editor/TemplateDropdown.js
```diff
@@ -38,15 +38,6 @@ type State = {
     templates: Array<MetadataTemplate>,
 };
 
-const InputContainer = ({ inputProps = {}, ...rest }: { inputProps?: Object }) => (
```
ivanthai, line 41: I removed this because it didn't seem to serve any purpose being wrapped inside of a container/functional component

src/components/selector-dropdown/SelectorDropdown.scss
tjuanitas: whoops, I commented this on the wrong line. I meant the `${OVERLAY_WRAPPER_CLASS} is-visible` part needs to stay. so it'll be something like:

```
<div className={`SelectorDropdown-overlay ${OVERLAY_WRAPPER_CLASS} is-visible`}>
```

but the line in this file could actually be removed. i.e. replace `.overlay-wrapper` with `.SelectorDropdown-overlay` here

tjuanitas: this is because the `overlay-wrapper` is being targeted in an Admin Console stylesheet, so we can't remove the classname from the element itself

ivanthai: ah that makes sense

src/components/conditional-wrapper/__tests__/ConditionalWrapper.test.js
```diff
@@ -0,0 +1,24 @@
...
+                otherProps: 'other-props',
+            };
+
+            const wrapper = shallow(
+                <ConditionalWrapper {...props}>
+                    <div />
+                </ConditionalWrapper>,
+            );
+
+            expect(wrapper).toMatchSnapshot();
```
tjuanitas, line 21: we're trying to avoid screenshots cause they're expensive. can we instead test for the correct components?

src/components/conditional-wrapper/__tests__/ConditionalWrapper.test.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import ConditionalWrapper from '../ConditionalWrapper';
+
+describe('components/in-app-messenger/core/components/ConditionalWrapper', () => {
```
tjuanitas, line 6: can we fix this path?

src/components/selector-dropdown/SelectorDropdown.scss
```diff
@@ -13,8 +13,9 @@
...
+    .SelectorDropdown-overlay,
     .overlay {
+        position: relative;
```
tjuanitas, line 18: what's this for? is something using `position: absolute;`?

src/components/selector-dropdown/SelectorDropdown.js
```diff
@@ -23,6 +26,8 @@ type Props = {
     dividerIndex?: number,
     /** Options to keep the results always open */
     isAlwaysOpen?: boolean,
+    /** Option to disable dynamic positioning with popper */
+    isPopperDisabled?: boolean,
```
tjuanitas, line 30: should we call this `hasDynamicPosition`?

src/components/selector-dropdown/SelectorDropdown.scss
ivanthai: The overlay css adds position: absolute to the parent of the PopperComponent which breaks the react-popper usage so I'm defining it here.

src/components/conditional-wrapper/ConditionalWrapper.js
```diff
@@ -0,0 +1,17 @@
...
+ * conditional wrapper will disable the wrapper and only return the children if isDisabled is true
+ */
+type Props = {
+    children: React.Node,
+    isDisabled?: boolean,
+    wrapper: any,
+    wrapperProps?: Object,
+};
+
+const ConditionalWrapper = ({ children, isDisabled, wrapper, wrapperProps = {} }: Props) =>
```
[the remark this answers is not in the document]
Machoper: True, I think it's better to just modify the specific wrapper being used rather than introducing a component like this.

I removed the component and added a new prop `hasDynamicPosition` to `PopperComponent` instead, which optionally appends the computed style to the popper element.

src/components/selector-dropdown/SelectorDropdown.scss
Machoper: I moved this to `PillSelectorDropdown` as it seems to be the only place we need `position: relative;`. 

@ivanthai please feel free to make changes if that's not the case.

src/components/selector-dropdown/SelectorDropdown.js
Machoper: I renamed and updated the corresponding code.

Machoper (dismissed):

src/components/pill-selector-dropdown/PillSelector.js
```diff
@@ -289,4 +304,11 @@ class PillSelector extends React.Component<Props, State> {
     }
 }
 
+export { PillSelectorBase };
+
+const PillSelector = React.forwardRef<Config, PillSelectorBase>((props: Config, ref: React.Ref<any>) => (
```
tjuanitas, line 309: I think `PillSelectorBase` should be `HTMLSpanElement`:
```
React.forwardRef<Config, HTMLSpanElement>
```

src/components/pill-selector-dropdown/PillSelector.js
```diff
@@ -289,4 +304,11 @@ class PillSelector extends React.Component<Props, State> {
...
+    <PillSelectorBase {...props} innerRef={ref} />
+));
+PillSelector.displayName = 'PillSelector';
```
tjuanitas, line 312: is the `displayName` missing if we remove this line?

it'd be great if we could do this:
```
class PillSelector extends [...]

[...]

export { PillSelector as PillSelectorBase };
export default React.forwardRef[...]
    <PillSelector {...props} innerRef={ref} />
));
```
since that's the common pattern for this repo

src/components/search-form/SearchForm.js
```diff
@@ -226,4 +242,12 @@ class SearchForm extends React.Component<Props, State> {
     }
 }
 
-export default injectIntl(SearchForm);
+const SearchFormBaseIntl = injectIntl(SearchFormBase);
+export { SearchFormBaseIntl };
+
+const SearchForm = React.forwardRef<Config, SearchFormBase>((props: Config, ref: React.Ref<any>) => (
```
tjuanitas, line 248: `React.forwardRef<Config, HTMLDivElement>`?

src/components/pill-selector-dropdown/PillSelector.js
```diff
@@ -26,12 +26,13 @@ type Props = {
...
     /** Function to retrieve the image URL associated with a pill */
     getPillImageUrl?: (data: { id: string | number, [key: string]: any }) => string,
+    innerRef?: React.Ref<any>,
     inputProps: Object,
     onInput: Function,
     onRemove: Function,
     onSuggestedPillAdd?: Function,
     placeholder: string,
-    selectedOptions: Array<Object> | List<Object>,
+    selectedOptions: List<Object>,
```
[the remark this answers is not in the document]
ivanthai: `return typeof selectedOptions.size === 'number' ? selectedOptions.size : selectedOptions.length;`
 
size() is not supported on Array. This was reported by flow as well

src/components/pill-selector-dropdown/PillSelector.js
ivanthai: `displayName` will be missing without it. Otherwise the snapshots will show something like `<ForwardRef>` instead of `<PillSelector>`

ivanthai: This will break existing tests that rely on finding a component by displayName. 

`expect(wrapper.is('SearchForm')).toBe(true);`

tjuanitas: gotcha, makes sense. thanks for checking!

src/components/selector-dropdown/SelectorDropdown.js
```diff
@@ -311,13 +324,17 @@ class SelectorDropdown extends React.Component<Props, State> {
                 onPaste={this.handleInput}
                 ref={this.selectorDropdownRef}
             >
-                {React.cloneElement(selector, { inputProps })}
-                {isOpen && (
-                    <div className={`${OVERLAY_WRAPPER_CLASS} is-visible`}>
+                <PopperComponent
```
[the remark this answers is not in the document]
ivanthai: Yeah it seems to work fine

src/components/pill-selector-dropdown/PillSelector.scss
```diff
@@ -122,6 +122,10 @@
             line-height: $bdl-line-height + $bdl-grid-unit * 2;
         }
     }
+
+    .SelectorDropdown-overlay > ul.overlay {
```
tjuanitas, line 126: any particular reason for this extra specificity?

tjuanitas (approved):
left a question but looks good to me!

src/components/search-form/SearchForm.js
```diff
@@ -54,13 +55,27 @@ type Props = {
     value?: string,
 };
 
+type DefaultProps = {
+    action: string,
+    method: 'get' | 'post',
+    name: string,
+    queryParams: Object,
```
JChan106, line 62: Are we able to use `{}` notation to be more explicit?

src/components/pill-selector-dropdown/PillSelector.scss
ivanthai: There can be a case where a descendent can be an overlay. This is why I needed to specify this rule

tjuanitas:
fyi I think some of the changes got reverted back in the last push

tjuanitas (approved):

JChan106 (approved):
lgtm

PR #3265, opened 2023-02-27
fix(metadata): fix template dropdown truncation issue

Machoper:
The metadata template dropdown uses the Flyout component which flips the overlay upward if it exceeds the bottom of the viewport, making the dropdown look truncated.

To fix this issue, I added a new prop `contrainToWindowWithPin` to Flyout. When set to true, it pins the overlay to edges on small screens and makes good use of available space.

The main use case is in the Embed Widget 2.0 with the smallest recommended dimensions of 330px x 400px. Anything < 400px in height is out of our consideration.

### Before
metadata_template_before

### After
metadata_template_after

### DIFF
--- a/src/components/flyout/Flyout.js
+++ b/src/components/flyout/Flyout.js
@@ -120,6 +120,10 @@ export type FlyoutProps = {
      */
     constrainToWindow?: boolean,
     /**
+     * Sets tether constrain to window with pin
+     */
+    constrainToWindowWithPin?: boolean,
+    /**
      * Toggles responsive behavior
      */
     isResponsive?: boolean,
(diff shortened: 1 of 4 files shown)


tjuanitas (approved):
looks fine since we're already doing the same approach in `DropdownMenu`

in the future, we could consider a `constraints` prop to allow further customization: https://github.com/shipshapecode/tether/blob/master/docs/intro.md#constraints

i.e. rename the current `constraints` variable to `tetherConstraints`, add a `constraints` prop, and then in `tetherProps`:
```
const tetherProps: Object = {
        ...
        constraints: constraints || tetherConstraints,
        ...
}
```
but maybe for another day

ivanthai (approved):

PR #3267, opened 2023-03-01
feat(input-with-copy): add node for settings button

hkisel-box:
Possible usages (how it looks like with additional buttons and without):
don't pay a lot of attention on content and paddings in these 2 new buttons, you should provide yours to these Nodes, that's just an examples.

Screenshot 2023-03-02 at 18 39 06

Screenshot 2023-03-02 at 18 40 43

Screenshot 2023-03-02 at 18 42 05

how we will use it in Relay:
Screenshot 2023-03-02 at 18 41 06

_I have also just realised that this "copy" button has different paddings_ :thinking_face:

@tjuanitas am I right that it should be possible simply by passing `className` to `buttonProps`?

### DIFF
--- a/src/components/text-input-with-copy-button/TextInputWithCopyButton.js
+++ b/src/components/text-input-with-copy-button/TextInputWithCopyButton.js
@@ -16,7 +16,7 @@ const defaultCopyText = <FormattedMessage {...messages.copy} />;
 const defaultCopiedText = <FormattedMessage {...messages.copied} />;
 
 type Props = {
-    /** Node for settings button */
+    /** Array of nodes for additional buttons */
     additionalButtons?: Array<React.Node>,
     /** Set the focus to input when component loads */
     autofocus?: boolean,
--- a/src/components/text-input-with-copy-button/TextInputWithCopyButton.js
+++ b/src/components/text-input-with-copy-button/TextInputWithCopyButton.js
@@ -28,6 +28,8 @@ type Props = {
     /** Label displayed for the text input */
     // TODO: Make label required
     label?: React.Node,
+    /** Node for settings button */
+    nodeForButton?: React.Node,
     /** onFocus handler for the input el */
     onCopySuccess?: Function,
     /** Function called when link is copied by keyboard or button */
@@ -216,6 +218,7 @@ class TextInputWithCopyButton extends React.PureComponent<Props, State> {
         return (
             <div className={wrapperClasses} {...copyEvent}>
                 <TextInput {...inputProps} onFocus={this.handleFocus} />
+                {rest.nodeForButton}
                 {this.renderCopyButton()}
             </div>
         );
--- a/src/components/text-input-with-copy-button/TextInputWithCopyButton.js
+++ b/src/components/text-input-with-copy-button/TextInputWithCopyButton.js
@@ -188,7 +188,7 @@ class TextInputWithCopyButton extends React.PureComponent<Props, State> {
         ) : null;
 
     render() {
-        const { className, additionalButtons, ...rest } = this.props;
+        const { additionalButtons, className, ...rest } = this.props;
         const { copySuccess } = this.state;
         const { isCopyCommandSupported } = this;
 
--- a/src/components/text-input-with-copy-button/TextInputWithCopyButton.js
+++ b/src/components/text-input-with-copy-button/TextInputWithCopyButton.js
@@ -16,6 +16,8 @@ const defaultCopyText = <FormattedMessage {...messages.copy} />;
 const defaultCopiedText = <FormattedMessage {...messages.copied} />;
 
 type Props = {
+    /** Node for settings button */
+    additionalButtons?: Array<React.Node>,
     /** Set the focus to input when component loads */
     autofocus?: boolean,
     /** Default copy button text */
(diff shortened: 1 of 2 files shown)


src/components/text-input-with-copy-button/TextInputWithCopyButton.js
```diff
@@ -216,6 +218,7 @@ class TextInputWithCopyButton extends React.PureComponent<Props, State> {
         return (
             <div className={wrapperClasses} {...copyEvent}>
                 <TextInput {...inputProps} onFocus={this.handleFocus} />
+                {rest.nodeForButton}
```
tjuanitas, line 221: could we destructure the prop at the top of the method?

```
const { className, nodeForButton, ...rest} = this.props;
```

then you can use it directly here without the dot notation

src/components/text-input-with-copy-button/TextInputWithCopyButton.scss
```diff
@@ -27,11 +27,19 @@
...
+
     &.copy-success {
-        .btn,
-        .btn:hover,
-        .btn:active,
-        .btn:active:hover {
+        .btn:last-child,
+        .btn:last-child:hover,
+        .btn:last-child:active,
+        .btn:active:last-child:hover {
```
tjuanitas, line 42: I think we can remove this selector `.btn:active:hover`. it should be the same as the two above it

src/components/text-input-with-copy-button/TextInputWithCopyButton.scss
```diff
@@ -27,11 +27,19 @@
         transition: background-color .5s ease, border-color .5s ease;
     }
 
+    .bdl-Button:not(:last-child),
+    .btn:not(:last-child) {
+        margin: 0;
+        border-left: 0;
+        border-radius: 0;
+        transition: background-color .5s ease, border-color .5s ease;
+    }
```
tjuanitas, line 36: I think we can write these two blocks can be written as:

```
.bdl-Button,
.btn {
    margin: 0;
    border-left: 0;
    border-radius: 0;
}

.bdl-Button:last-child,
.btn:last-child {
    min-width: 80px;
    border-top-right-radius: $bdl-border-radius-size-med;
    border-bottom-right-radius: $bdl-border-radius-size-med;
    transition: background-color .5s ease, border-color .5s ease;
}
```

tjuanitas (commented):
could we add a screenshot to the PR?

hkisel-box:
> could we add a screenshot to the PR?

done! I think it is also good idea to rename `buttonProps` to `copyButtonProps` now, but not sure how to do it - I guess existing property name already in usage through the products

src/components/text-input-with-copy-button/TextInputWithCopyButton.js
```diff
@@ -16,6 +16,8 @@ const defaultCopyText = <FormattedMessage {...messages.copy} />;
 const defaultCopiedText = <FormattedMessage {...messages.copied} />;
 
 type Props = {
+    /** Node for settings button */
```
tjuanitas, line 19: we should update this description to make it more general

src/components/text-input-with-copy-button/TextInputWithCopyButton.js
```diff
@@ -186,7 +188,7 @@ class TextInputWithCopyButton extends React.PureComponent<Props, State> {
         ) : null;
 
     render() {
-        const { className, ...rest } = this.props;
+        const { className, additionalButtons, ...rest } = this.props;
```
tjuanitas, line 191: this is super super minor but could we alphasort these?

```suggestion
        const { additionalButtons, className, ...rest } = this.props;
```

tjuanitas (commented):
> I think it is also good idea to rename buttonProps to copyButtonProps now, but not sure how to do it - I guess existing property name already in usage through the products

good call but you're right that since it's already being used in the parent apps, it's difficult to update prop names. we should leave it for now and audit which components would need to be fixed later

I left some call comments but otherwise looks good to me!

src/components/text-input-with-copy-button/TextInputWithCopyButton.js
hkisel-box: good catch, thanks

tjuanitas (approved):
lgtm!

src/components/text-input-with-copy-button/TextInputWithCopyButton.js
```diff
@@ -216,6 +218,7 @@ class TextInputWithCopyButton extends React.PureComponent<Props, State> {
         return (
             <div className={wrapperClasses} {...copyEvent}>
                 <TextInput {...inputProps} onFocus={this.handleFocus} />
+                {additionalButtons}
```
megansmith-box, line 221: Do we need a default value of null? What does `{undefined}` do here?

tjuanitas: `undefined` will not render anything

megansmith-box: Good to know thanks!

megansmith-box (approved):

src/components/text-input-with-copy-button/TextInputWithCopyButton.scss
```diff
@@ -19,19 +19,25 @@
...
+
+    .bdl-Button:last-child,
+    .btn:last-child {
         min-width: 80px;
         margin: 0;
         border-left: 0;
-        border-top-left-radius: 0;
-        border-bottom-left-radius: 0;
+        border-radius: 0 6px 6px 0;
```
tjuanitas, line 33: @hkisel-box oops minor but we should use the SASS variables for this

not really blocking though but I can fix this later

PR #3273, opened 2023-03-04
fix(content-explorer): Fix color contrast issue for item details

tjuanitas:
**Changes**

- Improve color contrast for the item details text when the row is hovered
  - This is barely noticeable in screenshots
- Remove extra padding before "All Files" when only one breadcrumb item

**Before**
Screen Shot 2023-03-03 at 5 15 31 PM

**After**
Screen Shot 2023-03-03 at 5 15 13 PM

### DIFF
--- a/src/elements/common/breadcrumbs/InlineBreadcrumbs.scss
+++ b/src/elements/common/breadcrumbs/InlineBreadcrumbs.scss
@@ -3,7 +3,7 @@
 .be-inline-breadcrumbs {
     display: flex;
     min-width: 0;
-    color: $bdl-gray-62;
+    color: $bdl-gray-65;
     font-size: 11px;
 
     .be-breadcrumbs {
@@ -12,7 +12,7 @@
 
     .be-breadcrumb {
         &:first-child {
-            min-width: 43px;
+            min-width: auto;
         }
 
         .be & .btn-plain {
--- a/src/elements/common/item/ItemDetails.scss
+++ b/src/elements/common/item/ItemDetails.scss
@@ -2,7 +2,7 @@
 
 .be-item-name {
     .be-item-details {
-        color: $bdl-gray-62;
+        color: $bdl-gray-65;
         font-size: 11px;
     }
 }
--- a/src/elements/content-explorer/ItemList.js
+++ b/src/elements/content-explorer/ItemList.js
@@ -181,7 +181,7 @@ const ItemList = ({
                             />
                             {isSmall ? null : (
                                 <Column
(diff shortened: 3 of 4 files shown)


src/elements/common/breadcrumbs/InlineBreadcrumbs.scss
```diff
@@ -12,7 +12,7 @@
 
     .be-breadcrumb {
         &:first-child {
-            min-width: 43px;
```
tjuanitas, line 15: I'm not sure what this value represents but it was added in this PR: https://github.com/box/box-ui-elements/pull/46

The min-width is adding an extra padding when the text is shorter than the width. there isn't any context from the PR about it's purpose but I tested ContentExplorer and ContentPicker and didn't see any issues with removing this.

It needs to be reset to `auto` because it inherits another min-width that is set to `64px`

QingyuChai (approved):
lgtm

src/elements/common/breadcrumbs/InlineBreadcrumbs.scss
bfoxx1906: Did you test your change with different viewports? I'm always hesitant to change things like this because it was certainly put there on purpose. Additionally, that css mi-width change  is for the first child so is .be-breadcrumb used in any other places that might be impacted by your change?

tjuanitas: yeah I tested with different viewports and view modes and didn't notice any impact.

this change would be limited to the `InlineBreadcrumbs` component which is only used by the `ItemDetails` component used for rendering the file name and additional details. this breadcrumb component is specific for the UI elements and would only be for the details text (folder path).

assuming that the breadcrumb has content, this change will set the width to the same as the length of the text instead of possibly making the width larger than it needs to be.

if the breadcrumb doesn't have content, then this change would remove whitespace. and if this was affecting something, it would be more appropriate to use margin instead.

bfoxx1906: Ok so are we good with really short file names as well? 43px is kind of specific so I'm pretty sure he did it for a reason but then again that was back in 2017.

tjuanitas: yeah this change is an improvement for shorter names:

**Before**
Screen Shot 2023-03-06 at 12 19 33 PM

**After**
Screen Shot 2023-03-06 at 12 20 21 PM

bfoxx1906 (approved):
looks good.

PR #3274, opened 2023-03-04
refactor(text-input-with-copy): Use SASS variables and remove styles

tjuanitas:


src/components/text-input-with-copy-button/TextInputWithCopyButton.scss
```diff
@@ -22,15 +22,13 @@
         margin: 0;
         border-left: 0;
         border-radius: 0;
-        transition: background-color .5s ease, border-color .5s ease;
```
tjuanitas, line 25: the transition is only needed on the copy button

src/components/text-input-with-copy-button/TextInputWithCopyButton.scss
```diff
@@ -22,15 +22,13 @@
...
     }
 
     .bdl-Button:last-child,
     .btn:last-child {
         min-width: 80px;
-        margin: 0;
-        border-left: 0;
```
tjuanitas, line 32: these are redundant since the above block as them

QingyuChai (approved):
lgtm

hkisel-box (approved):

PR #2231, opened 2020-07-14
feat(annotations): Improve annotation activity card interactions

jstoffan:
The activity feed is in kind of an awkward place with regard to code conventions. I tried to change/refactor only what was helpful to this particular enhancement rather than the entire component, so it's a bit of mashup at the moment. I'm open to feedback on it, though.

Todo
- [x] Design review

PR #3279, opened 2023-03-13
feat(threaded-replies): Decouple ActivityCard from Comment

JChan106:
This PR is to decouple the ActivityCard component from the comment component. We have another ticket to decouple ActivityCard from Annotations as well. ActivityThread and ActivityThreadReplies will be going away, but I made changes to that file regardless to track the styling needed when those files get deleted.  These changes will not affect the existing padding for Comments without the new FF changes:
image

There are selectors for ActivityCard in `ActiveState.scss`, however I confirmed those are only for annotations: https://github.com/box/box-ui-elements/pull/2231/files https://github.com/box/box-ui-elements/pull/2508/files 

So I will address them in another PR.

With `newThreadedReplies` and `hasReplies` enabled:
image

With `newThreadedReplies` and `hasReplies` disabled:
image

### DIFF
--- a/src/elements/content-sidebar/ActivitySidebar.js
+++ b/src/elements/content-sidebar/ActivitySidebar.js
@@ -139,6 +139,7 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         emitAnnotationUpdateEvent: noop,
         getAnnotationsMatchPath: noop,
         getAnnotationsPath: noop,
+        hasNewThreadedReplies: false,
         hasReplies: false,
         hasTasks: true,
         hasVersions: true,
--- a/src/elements/content-sidebar/ActivitySidebar.js
+++ b/src/elements/content-sidebar/ActivitySidebar.js
@@ -139,7 +139,6 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         emitAnnotationUpdateEvent: noop,
         getAnnotationsMatchPath: noop,
         getAnnotationsPath: noop,
-        hasNewThreadedReplies: false,
         hasReplies: false,
         hasTasks: true,
         hasVersions: true,
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActivityItem.scss
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActivityItem.scss
@@ -1,7 +1,7 @@
 @import '../../../common/variables';
 
 .bcs-ActivityItem {
-    .bcs-Comment {
+    .bcs-BaseComment {
         padding: $sidebarActivityFeedSpacingVertical $sidebarActivityFeedSpacingHorizontal;
     }
 }
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActivityItem.scss
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActivityItem.scss
@@ -1,0 +1,7 @@
+@import '../../../common/variables';
+
+.bcs-ActivityItem {
+    .bcs-Comment {
+        padding: $sidebarActivityFeedSpacingVertical $sidebarActivityFeedSpacingHorizontal;
+    }
+}
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActivityThread.scss
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActivityThread.scss
@@ -16,11 +16,13 @@ $sidebarActivityFeedSelectedItemBorderWidth: 2px;
         padding-bottom: $sidebarActivityFeedSpacingVertical - $sidebarActivityFeedSelectedItemBorderWidth;
     }
 
-    .bcs-ActivityCard {
+    .bcs-ActivityCard,
+    .bcs-Comment {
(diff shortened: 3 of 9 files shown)


src/elements/content-sidebar/activity-feed/activity-feed/ActivityItem.scss
```diff
@@ -0,0 +1,7 @@
+@import '../../../common/variables';
+
+.bcs-ActivityItem {
+    .bcs-Comment {
+        padding: $sidebarActivityFeedSpacingVertical $sidebarActivityFeedSpacingHorizontal;
```
karelee7, line 5: minor: in the past Conrad mentioned that it was better to explicitly say `padding-left`, `padding-top`, etc., but it's up to you! Just a fyi

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -168,7 +167,7 @@ const BaseComment = (props: Props) => {
     const isResolved = status === COMMENT_STATUS_RESOLVED;
 
     return (
-        <ActivityCard className="bcs-Comment">
+        <div className="bcs-Comment">
```
karelee7, line 170: Can we add a `data-testid`?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityItem.scss
JChan106: Gotcha, not sure if we want to follow that in this case since it's 2 values for all 4 sides. Also we have variables saying `vertical` and `horizontal` makes it more readable. Being more explicit for 3 values+ makes sense.

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
JChan106: We should add data-testid attributes when they are needed. In this case, we can check the actual content of the comment. If we can, we should always test what the user would see with rtl.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -139,6 +139,7 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         emitAnnotationUpdateEvent: noop,
         getAnnotationsMatchPath: noop,
         getAnnotationsPath: noop,
+        hasNewThreadedReplies: false,
```
tjuanitas, line 142: what's this?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityItem.scss
```diff
@@ -0,0 +1,7 @@
+@import '../../../common/variables';
+
+.bcs-ActivityItem {
+    .bcs-Comment {
```
tjuanitas, line 4: I'm not too familiar with Sidebar, is the `Comment` component used in other places? is this nesting necessary? why not move it to `Comment.scss`?

src/elements/content-sidebar/ActivitySidebar.js
JChan106: sorry, rebase accident

src/elements/content-sidebar/activity-feed/activity-feed/ActivityItem.scss
JChan106: Because this padding is specific to `Activity`, we want decouple as much activity related css away from the comment component if possible

bfoxx1906:
Have you verified that everything works properly in Canvas as well?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityThread.scss
```diff
@@ -80,12 +82,14 @@ $sidebarActivityFeedSelectedItemBorderWidth: 2px;
     }
 }
 
-.be .bcs-activity-feed .bcs-activity-feed-items-container .bcs-ActivityItem.bcs-is-focused .bcs-ActivityThread .bcs-ActivityCard {
+.be .bcs-activity-feed .bcs-activity-feed-items-container .bcs-ActivityItem.bcs-is-focused .bcs-ActivityThread .bcs-ActivityCard,
+.be .bcs-activity-feed .bcs-activity-feed-items-container .bcs-ActivityItem.bcs-is-focused .bcs-ActivityThread .bcs-Comment {
```
bfoxx1906, line 86: There's a lot going on here. I would recommend just duplicating the css for now and keeping it separate. I'm willing to sacrifice some brevity for readability in this case especially sense the other stuff will eventually go away.

JChan106: This file will be deleted

karelee7 (approved):
Looks good to me! Seems like one of the e2e tests isn't happy though

JChan106:
> Looks good to me! Seems like one of the e2e tests isn't happy though

Hmm yeah idk why, will take a look

patlm (approved):

tjuanitas (approved):
LGTM. we're planning on completing removing ActivityThread and ActivityThreadReplies right?

JChan106:
> LGTM. we're planning on completing removing ActivityThread and ActivityThreadReplies right?

Yes exactly, we will have replies as children of the parent comment instead.

PR #3280, opened 2023-03-13
feat(threaded-replies): Update activity filter to support more options

patlm:
This PR is to update the ActivitySidebarFilter and ActivitySidebar components to support additional filter options like seen in screenshots below.

`activityFeed.newThreadedReplies` feature is enabled and the sidebar may contain tasks:
image

`activityFeed.newThreadedReplies` feature is enabled and the sidebar will never contain tasks:
image

`activityFeed.filter` feature is enabled and `activityFeed.newThreadedReplies` is not enabled:
image

Which filter options are shown is based on the context that the filter is being used in. This can be seen in screenshots above. Specifically, the context is determined based on `hasTasks` props that are passed to the ActivitySidebar component. Additionally, the conditional check is behind the `activityFeed.newThreadedReplies` feature. If only the `activityFeed.filter` is specified, it uses the previous basic filter options of all and unresolved. Inside the ActivitySidebar is where it determines which filter options should be available.

In this PR, the filtering that occurs is very basic client-side filtering that does not fetch the latest activity when a filter is applied. However, future story/PR will re-evaluate how the filtering is handled.

Lastly, these changes are behind the `activityFeed.newThreadedReplies` feature.

Todo:
- [x] Update/add any needed tests


src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -1103,7 +1107,11 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
             return feedItems;
         }
         return feedItems.filter(item => {
-            return item.status === feedItemsStatusFilter || item.type === FEED_ITEM_TYPE_VERSION;
+            return (
+                item.status === feedItemsStatusFilter ||
+                item.type === FEED_ITEM_TYPE_VERSION ||
+                item.type === feedItemsStatusFilter
```
patlm, line 1113: Adding this last filter option adds basic support for filtering tasks. This doesn't re-fetch the activities each time a new filter is selected. Rather, it filters the current list of activities. New activities are only fetched on page reload.

src/elements/content-sidebar/ActivitySidebarFilter.js
```diff
@@ -9,29 +9,51 @@ import { FormattedMessage } from 'react-intl';
...
+    ACTIVITY_FEED_ITEM_OPEN,
+    ACTIVITY_FEED_ITEM_RESOLVED,
+    ACTIVITY_FEED_ITEM_TASKS,
+    ACTIVITY_FILTER_TITLE_ALL_COMMENTS,
+    COMMENT_STATUS_OPEN,
+    COMMENT_STATUS_RESOLVED,
+} from '../../constants';
 import { Menu, SelectMenuItem } from '../../components/menu';
-import type { FeedItemStatus } from '../../common/types/feed';
+import type { ActivityFilterAllTitle, ActivityFilterStatus } from '../../common/types/feed';
```
patlm, line 22: A new type (`ActivityFilterStatus`) was created to be used with filters because `FeedItemStatus` is used only for comments and annotations to differentiate between open and resolved.

src/elements/content-sidebar/ActivitySidebarFilter.js
```diff
@@ -9,29 +9,51 @@ import { FormattedMessage } from 'react-intl';
...
-    [ACTIVITY_FEED_ITEM_ALL]: { msg: messages.activitySidebarFilterOptionAll, val: undefined },
-    [ACTIVITY_FEED_ITEM_OPEN]: { msg: messages.activitySidebarFilterOptionOpen, val: COMMENT_STATUS_OPEN },
+    activityFilterAllTitle?: ActivityFilterAllTitle,
+    activityFilterOptions?: ActivityFilterStatus[],
+    feedItemStatus?: ActivityFilterStatus,
+    onFeedItemStatusClick: (status?: ActivityFilterStatus) => void,
 };
 
 function ActivitySidebarFilter({
+    activityFilterAllTitle = ACTIVITY_FILTER_TITLE_ALL_COMMENTS,
```
patlm, line 33: This title prop was created to ensure that Canvas could maintain current functionality while providing Preview its needed functionality. For preview, instead of the title "All Comments" that Canvas uses, Preview will need to use the title "All Activity".

An alternative considered was determining the title used by if the filter options included tasks. However, I was concerned this might couple two different things that in the future we may not want as tightly coupled.

i18n/en-US.properties
```diff
@@ -22,8 +22,14 @@ be.activitySidebar.activityFeed.showReplies = See {repliesToLoadCount, plural, o
...
 # One of the dropdown options to select in order to filter comments
 be.activitySidebarFilter.status.all = All Comments
+# One of the dropdown options to select in order to filter activities
+be.activitySidebarFilter.status.allActivity = All Activity
 # One of the dropdown options to select in order to filter comments
 be.activitySidebarFilter.status.open = Unresolved Comments
+# One of the dropdown options to select in order to filter comments
+be.activitySidebarFilter.status.resolved = Resolved Comments
+# One of the dropdown options to select in order to filter activities
+be.activitySidebarFilter.status.task = Tasks
```
JChan106, line 32: Could be something like `One of the dropdown options to select in order to filter tasks`. 

And I think we can make this plural`be.activitySidebarFilter.status.tasks`

src/constants.js
```diff
@@ -463,6 +463,16 @@ export const TASK_MAX_GROUP_ASSIGNEES: 250 = 250;
...
 
+/* ------------------ Activity Filter Statuses ------------------ */
+export const ACTIVITY_FEED_ITEM_ALL: 'all' = 'all';
+export const ACTIVITY_FEED_ITEM_OPEN: 'open' = 'open';
+export const ACTIVITY_FEED_ITEM_RESOLVED: 'resolved' = 'resolved';
+export const ACTIVITY_FEED_ITEM_TASKS: 'task' = 'task';
+
+/* ------------------ Activity Filter All Title Options ------------------ */
+export const ACTIVITY_FILTER_TITLE_ALL_ACTIVITY: 'all_activity' = 'all_activity';
+export const ACTIVITY_FILTER_TITLE_ALL_COMMENTS: 'all_comments' = 'all_comments';
```
JChan106, line 474: Are we going to have an all comments filter? I thought it was only activity

src/constants.js
```diff
@@ -463,6 +463,16 @@ export const TASK_MAX_GROUP_ASSIGNEES: 250 = 250;
 export const COMMENT_STATUS_OPEN: 'open' = 'open';
 export const COMMENT_STATUS_RESOLVED: 'resolved' = 'resolved';
 
+/* ------------------ Activity Filter Statuses ------------------ */
+export const ACTIVITY_FEED_ITEM_ALL: 'all' = 'all';
+export const ACTIVITY_FEED_ITEM_OPEN: 'open' = 'open';
+export const ACTIVITY_FEED_ITEM_RESOLVED: 'resolved' = 'resolved';
+export const ACTIVITY_FEED_ITEM_TASKS: 'task' = 'task';
```
JChan106, line 470: should be plural

src/elements/content-sidebar/messages.js
```diff
@@ -12,11 +12,26 @@ const messages = defineMessages({
...
         defaultMessage: 'Unresolved Comments',
         description: 'One of the dropdown options to select in order to filter comments',
     },
+    activitySidebarFilterOptionResolved: {
+        id: 'be.activitySidebarFilter.status.resolved',
+        defaultMessage: 'Resolved Comments',
+        description: 'One of the dropdown options to select in order to filter comments',
+    },
+    activitySidebarFilterOptionTasks: {
+        id: 'be.activitySidebarFilter.status.task',
```
JChan106, line 31: plural

src/elements/content-sidebar/messages.js
```diff
@@ -12,11 +12,26 @@ const messages = defineMessages({
...
+        defaultMessage: 'Tasks',
+        description: 'One of the dropdown options to select in order to filter activities',
```
JChan106, line 33: `filter tasks`

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -70,6 +72,8 @@ import './ActivitySidebar.scss';
 type ExternalProps = {
     activeFeedEntryId?: string,
     activeFeedEntryType?: FocusableFeedItemType,
+    activityFilterAllTitle?: ActivityFilterAllTitle,
```
JChan106, line 75: Why do we need to pass this down?

src/constants.js
patlm: For filters, there will only be one "All" filter that appears at a time. For Canvas, the title of the filter should be "All Comments" to not break its current design. However, for Preview, the title needs to be "All Activities".

src/elements/content-sidebar/ActivitySidebar.js
patlm: This is related to "All Comments" vs "All Activities" title. There are alternative options to passing it down such as determining the title based on characteristics in ActivitySidebarFilter or ActivitySidebar.

src/constants.js
patlm: This currently is not plural so that it matches the activity type for task feed items. This provides the ability to filter with `item.type === feedItemsStatusFilter` instead of first mapping `feedItemsStatusFilter` to a type. I can see how making it plural could be better but that would also require mapping it to `task` for when filtering based on type.

i18n/en-US.properties
karelee7: Should we include the specific filter name in the comment? i.e. `The all activity dropdown option to select for filtering activities`

src/constants.js
karelee7: I see! In that case, would it help to include a comment about why it's singular here? Would it be better to: 1) use a different variable for the item.type or 2) use the plain string values when filtering via item.type? I'll leave it up to you, but if we keep it as is, maybe a comment would be helpful!

i18n/en-US.properties
```diff
@@ -20,10 +20,16 @@ be.activitySidebar.activityFeed.replyInThread = Reply in thread
 be.activitySidebar.activityFeed.showReplies = See {repliesToLoadCount, plural, one {# reply} other {# replies}}
 # Text to show when a task no longer exists
 be.activitySidebar.activityFeed.taskMissingError = This task no longer exists
-# One of the dropdown options to select in order to filter comments
-be.activitySidebarFilter.status.all = All Comments
-# One of the dropdown options to select in order to filter comments
+# One of the dropdown options to select for filtering activity list in order to filter all activity
```
karelee7, line 23: minor: To make it less wordy, you could say something like `Dropdown option for filtering all activity from activity list`

src/elements/content-sidebar/ActivitySidebarFilter.js
```diff
@@ -9,29 +9,57 @@ import { FormattedMessage } from 'react-intl';
 import DropdownMenu, { MenuToggle } from '../../components/dropdown-menu';
```
karelee7, line 9: Are there any unit tests for this file that we need to update too?

src/elements/content-sidebar/__tests__/ActivitySidebar.test.js
```diff
@@ -1578,19 +1581,68 @@ describe('elements/content-sidebar/ActivitySidebar', () => {
...
+
+            test('when activityFeed.newThreadedReplies is disabled and filter options list is provided', () => {
+                const wrapper = getWrapper({
+                    activityFilterOptions: ['all', 'open', 'resolved', 'tasks'],
+                });
+                const instance = wrapper.instance();
+                expect(instance.renderActivitySidebarFilter()).toBe(null);
+            });
+
+            test('when activityFeed.newThreadedReplies is enabled and no filter options list is provided', () => {
```
karelee7, line 1605: Do we also need to test for `when activityFeed.newThreadedReplies is enabled and filter options list is provided`? Also, would it be easier to group these tests together in a `test.each...` since they're similar?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -1145,15 +1155,34 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+        const activityFilterOptions: ActivityFilterOption[] = [ACTIVITY_FILTER_OPTION_ALL];
+        if (newThreadedRepliesEnabled) {
+            // Determine which filter options to show based on what activity types are available in current context
+            if (hasReplies) {
+                activityFilterOptions.push(ACTIVITY_FILTER_OPTION_OPEN, ACTIVITY_FILTER_OPTION_RESOLVED);
+            }
+            if (hasTasks) {
+                activityFilterOptions.push(ACTIVITY_FILTER_OPTION_TASKS);
+            }
+        } else if (activityFeedFilterEnabled) {
```
JChan106, line 1178: Not sure if these conditionals are correct.
I think regardless of `newThreadedRepliesEnabled`. `activityFeedFilterEnabled` needs to be enabled in order to show the filter options

src/elements/content-sidebar/ActivitySidebarFilter.js
```diff
@@ -9,29 +9,57 @@ import { FormattedMessage } from 'react-intl';
...
     onFeedItemStatusClick,
 }: ActivitySidebarFilterProps) {
+    const isCommentsOnlyFilter = !activityFilterOptions.includes(ACTIVITY_FILTER_OPTION_TASKS);
+
+    const statusMap = {
+        [ACTIVITY_FILTER_OPTION_ALL]: {
+            msg: isCommentsOnlyFilter
+                ? messages.activitySidebarFilterOptionAllComments
+                : messages.activitySidebarFilterOptionAllActivity,
+            val: undefined,
```
JChan106, line 50: would this be `ACTIVITY_FILTER_OPTION_ALL`?

src/elements/content-sidebar/ActivitySidebarFilter.js
```diff
@@ -9,29 +9,57 @@ import { FormattedMessage } from 'react-intl';
...
+    [ACTIVITY_FILTER_OPTION_TASKS]: FEED_ITEM_TYPE_TASK,
 };
 
 function ActivitySidebarFilter({
-    feedItemStatus = ACTIVITY_FEED_ITEM_ALL,
+    activityFilterOptions,
+    feedItemStatus = ACTIVITY_FILTER_OPTION_ALL,
     onFeedItemStatusClick,
 }: ActivitySidebarFilterProps) {
+    const isCommentsOnlyFilter = !activityFilterOptions.includes(ACTIVITY_FILTER_OPTION_TASKS);
```
JChan106, line 43: We can have the ternary here return the all message, so we can keep logic out of the statusMap

JChan106:
Can we also get screenshots for the different menus that appear when `newThreadedReplies` is enabled vs `activityFeedFilterEnabled`? There might be some confusion there.

src/elements/content-sidebar/ActivitySidebar.js
patlm: That makes sense, I could see that being beneficial in the long run. For the filtering, I think an important consideration is that within 1-2 sprints, we plan to implement a different strategy. I added this more as an MVP and bridge until we implement a proper filtering strategy.

I would lean towards adding a comment and leaving it as-is. Then, when updating the filtering strategy, change it to pass in that option (or something similar). I think it might be better to wait until we have more context into how the filtering will work and what would work best with the solution we implement.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -1145,15 +1155,34 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
     };
 
     renderActivitySidebarFilter = () => {
-        const { features } = this.props;
+        const { features, hasReplies, hasTasks } = this.props;
         const { feedItemsStatusFilter } = this.state;
-        const shouldShowActivityFeedFilter = isFeatureEnabled(features, 'activityFeed.filter.enabled');
+        const activityFeedFilterEnabled = isFeatureEnabled(features, 'activityFeed.filter.enabled');
+        const newThreadedRepliesEnabled = isFeatureEnabled(features, 'activityFeed.newThreadedReplies.enabled');
+        const shouldShowActivityFeedFilter =
```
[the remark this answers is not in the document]
patlm: I had a conversation with Brian about this today and definitely agree that it is confusing. The `activityFeedFilterEnabled` is a feature flip used by Canvas for their filtering feature. `newThreadedRepliesEnabled` is the feature flip Preview is using for the new threaded replies feature. I believe the thought process is that this filter option should always be on if the client is using new threaded replies. Also, using `newThreadedRepliesEnabled` will allow Canvas to maintain current functionality while Preview uses our new functionality. However, the `(hasReplies || hasTasks)` should not have been included so I will remove it.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -1145,15 +1155,34 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+            activityFeedFilterEnabled || (newThreadedRepliesEnabled && (hasReplies || hasTasks));
 
         if (!shouldShowActivityFeedFilter) {
             return null;
         }
+
+        const activityFilterOptions: ActivityFilterOption[] = [ACTIVITY_FILTER_OPTION_ALL];
+        if (newThreadedRepliesEnabled) {
+            // Determine which filter options to show based on what activity types are available in current context
+            if (hasReplies) {
```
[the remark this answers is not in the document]
patlm: This was added based on the idea that hasReplies also determines whether or not comments have the ability to be unresolved/resolved. However, after discussing it further with Brian, if the feature flip newThreadedReplies is enabled, it will always have unresolved/resolved comments. Code will be updated.

src/elements/content-sidebar/ActivitySidebarFilter.js
```diff
@@ -9,47 +9,83 @@ import { FormattedMessage } from 'react-intl';
...
 
 function ActivitySidebarFilter({
-    feedItemStatus = ACTIVITY_FEED_ITEM_ALL,
-    onFeedItemStatusClick,
+    activityFilterOptions,
+    feedItemType = ACTIVITY_FILTER_OPTION_ALL,
+    onFeedItemTypeClick,
 }: ActivitySidebarFilterProps) {
+    // The message for all activty is based on whether only comments are in the activityFilterOptions prop
+    const allFilterMessage = activityFilterOptions.every(option =>
```
JChan106, line 44: Can we separate this out into a separate function, and put the array in a variable?

So we can have something like this

`hasOnlyCommentOptions(activityFilterOptions) ? messages.activitySidebarFilterOptionAllComments : messages.activitySidebarFilterOptionAllActivity;`

JChan106 (approved):
lgtm, thanks for the changes!

bfoxx1906 (approved):

PR #3281, opened 2023-03-16
fix(usm): Remove aria-label and add aria-haspopup

tjuanitas:
The `aria-label` is unnecessary and mismatched with the text. Recommendation is to remove and add `aria-haspopup`.

Screen Shot 2023-03-16 at 11 35 38 AM

The "Link Settings" button is the element in question.

### DIFF
--- a/src/features/unified-share-modal/SharedLinkSection.js
+++ b/src/features/unified-share-modal/SharedLinkSection.js
@@ -438,7 +438,7 @@ class SharedLinkSection extends React.Component<Props, State> {
                 >
                     <PlainButton
                         {...sharedLinkSettingsButtonProps}
-                        aria-label={intl.formatMessage(messages.settingsButtonLabel)}
+                        aria-haspopup="dialog"
                         className="shared-link-settings-btn"
                         onClick={onSettingsClick}
                         type="button"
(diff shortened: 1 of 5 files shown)


src/features/unified-share-modal/SharedLinkSection.js
```diff
@@ -438,7 +438,7 @@ class SharedLinkSection extends React.Component<Props, State> {
                 >
                     <PlainButton
                         {...sharedLinkSettingsButtonProps}
-                        aria-label={intl.formatMessage(messages.settingsButtonLabel)}
+                        aria-haspopup="dialog"
```
tjuanitas, line 441: the recommendation is to add this because it opens a new dialog but the actual behavior is determined by the `onSettingsClick` prop which is defined in the parent. so technically it could do something else like open a new page or create a dropdown menu.

however, the current intention (at least in the webapp) is to open a dialog

megansmith-box: would it make  sense to keep both?

tjuanitas: it was recommended to remove `aria-label` because it's unnecessary since the button text already acts as the "label". it also creates a mismatch between what's visible on the element (through the button text) and what the screenreader reads (through `aria-label`) because it uses different messages

megansmith-box (approved):

TaoJiang98 (approved):

PR #3282, opened 2023-03-16
fix(usm): Removing redundant button element

tjuanitas:
The "Manage All" button is wrapped by both a link and a button. This causes it to be focused twice (using keyboard) and is distracting to screen readers. The button is replaced with a span and added `letter-spacing` to mimic the other "link" styled buttons on the main USM component.

**Before**
Screen Shot 2023-03-16 at 4 11 55 PM

**After**
Screen Shot 2023-03-16 at 4 12 08 PM

^ i.e. no difference. verified in Chrome, Safari, and Firefox

### DIFF
--- a/src/features/collaborator-avatars/CollaboratorList.js
+++ b/src/features/collaborator-avatars/CollaboratorList.js
@@ -4,7 +4,6 @@ import { FormattedMessage } from 'react-intl';
 
 import { ModalActions } from '../../components/modal';
 import Button from '../../components/button';
-import PlainButton from '../../components/plain-button';
 import { Link } from '../../components/link';
 import CollaborationBadge from '../../icons/badges/CollaborationBadge';
 import type {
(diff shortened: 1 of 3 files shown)


src/features/collaborator-avatars/CollaboratorList.js
```diff
@@ -51,9 +50,9 @@ class CollaboratorList extends React.Component<Props> {
         const { collaborators, onDoneClick, maxCollaboratorListSize, trackingProps } = this.props;
         const { usernameProps, emailProps, manageLinkProps, viewAdditionalProps, doneButtonProps } = trackingProps;
         const manageAllBtn = (
-            <PlainButton className="manage-all-btn" type="button">
+            <span className="manage-all-btn">
```
tjuanitas, line 53: while this class name is not on a button anymore, it cannot be changed since the classname is used in stylesheets in the main app and the monolith

src/features/unified-share-modal/UnifiedShareModal.scss
```diff
@@ -366,6 +366,7 @@
 
         float: right;
         color: $bdl-box-blue;
+        letter-spacing: normal;
```
megansmith-box, line 369: is this change necessary?

greg-in-a-box (approved):

megansmith-box (approved):

PR #3283, opened 2023-03-20
docs: add Box developer logo

mgrytsai:


README.md
```diff
@@ -1,4 +1,8 @@
-<img width="50%" alt="Box Elements" src="https://repository-images.githubusercontent.com/95743138/c161b500-021b-11ea-8bf9-3aa8776acdec" />
+<p align="center">
+  <img width="50%" alt="Box Elements" src="https://repository-images.githubusercontent.com/95743138/c161b500-021b-11ea-8bf9-3aa8776acdec" />
+  <img width="50%" src="https://github.com/box/sdks/blob/master/images/box-dev-logo.png" alt= “box-dev-logo”>
```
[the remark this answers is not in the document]
tjuanitas: I think since `alt` attributes are used by screen readers, the text should be more human readable:
```
alt="Box Developer logo"
```
\* not entirely sure if we should do `Box Dev` or `Box Developer` though
\** the double quotes used for the `alt` attribute look different than the other double quotes. we should make them the same

tjuanitas (dismissed):

README.md
```diff
@@ -1,4 +1,8 @@
-<img width="50%" alt="Box Elements" src="https://repository-images.githubusercontent.com/95743138/c161b500-021b-11ea-8bf9-3aa8776acdec" />
+<p align="center">
+  <img width="50%" alt="Box Elements logo" src="https://repository-images.githubusercontent.com/95743138/c161b500-021b-11ea-8bf9-3aa8776acdec" />
+  <img width="50%" alt="Box Developer logo" src="https://github.com/box/sdks/blob/master/images/box-dev-logo.png" />
```
benjamin-shen, line 3: [nit] I didn't realize you could link to github.com here. I don't think this will render the image properly in vanilla html. It looks like Github actually makes two network requests for this image (which doubles the image latency).
image

Maybe we should replace it with the raw link or permalink so it points directly to Github's CDN?
- https://raw.githubusercontent.com/box/sdks/master/images/box-dev-logo.png
- https://raw.githubusercontent.com/box/sdks/f55569c6f023a5dfddad96729b87a259c9bf4113/images/box-dev-logo.png

benjamin-shen (dismissed):

README.md
tjuanitas: oh good point

benjamin-shen: I think permalink is better so we can avoid link rot (if `box-dev-logo.png` is moved/renamed/removed from `master` or we rename `master` to `main`, for example). The small tradeoff is just that we'd have to update the link here if the logo changes.

benjamin-shen (approved):

megansmith-box (approved):

PR #3287, opened 2023-03-22
feat(collapsible-sidebar): Add prop to show CollapsibleSidebar collapsed

greathmaster:
### Overview
Currently, if the `CollapsibleSidebar` if the viewport width is greater than `1220px` there is no option to collapse it. This PR adds a prop `narrowItems` to make it collapsible even above `1220px` making the content more visible.

### Current behavior at `1300px`
app hvenkatesh inside-box net_folder_8803997002_s=5j4mo9gvb7mfpvyfjfvj5dxla5ivtjli

### After this PR with `narrowItems = true` at `1300px`
app hvenkatesh inside-box net_folder_8803997002_s=5j4mo9gvb7mfpvyfjfvj5dxla5ivtjli (2)

### DIFF
--- a/src/features/collapsible-sidebar/CollapsibleSidebar.js
+++ b/src/features/collapsible-sidebar/CollapsibleSidebar.js
@@ -39,6 +39,9 @@ type Props = {
 
     /** Optionally apply "aria-hidden": "true" to CollapsibleSidebar wrapper */
     isHidden?: boolean,
+
+    /** Whether to show the collapsed version with only icons */
+    narrowItems?: boolean,
 };
 
 class CollapsibleSidebar extends React.Component<Props> {
@@ -84,10 +87,11 @@ class CollapsibleSidebar extends React.Component<Props> {
     };
 
     render() {
-        const { children, className, expanded, isHidden, htmlAttributes } = this.props;
+        const { children, className, expanded, isHidden, htmlAttributes, narrowItems } = this.props;
         const navClasses = classNames(
             {
                 'is-expanded': expanded,
+                'is-narrow': narrowItems,
             },
             'bdl-CollapsibleSidebar',
             className,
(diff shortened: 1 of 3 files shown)


tjuanitas:
hey @greathmaster! could you add a screenshot of this change?

src/features/collapsible-sidebar/CollapsibleSidebar.js
```diff
@@ -39,6 +39,9 @@ type Props = {
 
     /** Optionally apply "aria-hidden": "true" to CollapsibleSidebar wrapper */
     isHidden?: boolean,
+
+    /** Whether to show the collapsed version with only icons */
```
tjuanitas, line 43: I'm not familiar with how this is different than `expanded=false`. are `expanded` and `narrowItems` mutually exclusive?

src/features/collapsible-sidebar/CollapsibleSidebar.js
```diff
@@ -39,6 +39,9 @@ type Props = {
...
+    narrowItems?: boolean,
```
tjuanitas, line 44: can prefix booleans with `is`?

src/features/collapsible-sidebar/CollapsibleSidebar.js
greathmaster: In `CollapsibleSidebar.scss` we have a media query that sets the sidebar width to the narrow width (`68px`) when the viewport width is under `1220px`. But above `1220px` it shrinks the sidebar with `expanded=false`, but it leaves the width causing the Files List to be displaced:

app hvenkatesh inside-box net_folder_8803997002_s=5j4mo9gvb7mfpvyfjfvj5dxla5ivtjli (3)

https://github.com/box/box-ui-elements/blob/90e7c476bd7f74c3ad15a49c069f5c5e823f8308/src/features/collapsible-sidebar/CollapsibleSidebar.scss#L115-L137

Question...am I re-creating how `expanded=false` should be working but isn't? ie show the collapsed narrow version regardless of screen size? What is the expected behavior of the `expanded` prop?

PR #3288, opened 2023-03-22
feat(threaded-replies): Create Replies component

JChan106:
This PR adds the `Replies` component to contain replies of a comment. This PR was meant to show replies using the new compoenent, and add the ability to show and hide replies. The styling and logic of other Threaded Replies functionalities like creating, editing, deleting, etc. will come in future PRs.

- [x] Unit tests

`hasReplies` enabled, `hasNewThreadedReplies` disabled:
image

`hasReplies` enabled, `hasNewThreadedReplies` enabled:
image


src/elements/content-sidebar/activity-feed/comment/Replies.js
```diff
@@ -0,0 +1,160 @@
...
+    const { length: repliesLength } = replies;
+    const repliesToLoadCount = Math.max(repliesTotalCount - repliesLength, 0);
+
+    const onHideRepliesHandler = () => {
+        if (repliesLength) {
+            onHideReplies(replies[repliesLength - 1]);
+        }
+    };
+
+    const renderButton = () => {
```
jstoffan, line 78: Good rule of thumb: If we're introducing a separate "render" helper, move the contents to a new component, instead.

src/elements/content-sidebar/activity-feed/comment/Replies.js
```diff
@@ -0,0 +1,160 @@
...
+    parentID,
+    replies = [],
+    repliesTotalCount = 0,
+    translations,
+}: Props) => {
+    const { length: repliesLength } = replies;
+    const repliesToLoadCount = Math.max(repliesTotalCount - repliesLength, 0);
+
+    const onHideRepliesHandler = () => {
+        if (repliesLength) {
```
jstoffan, line 73: Minor, but we typically prefer guard clauses.

src/elements/content-sidebar/activity-feed/comment/Replies.js
```diff
@@ -0,0 +1,160 @@
...
+    };
+
+    const renderButton = () => {
+        if (isAlwaysExpanded || isRepliesLoading) {
+            return null;
+        }
+
+        if (repliesTotalCount > repliesLength) {
+            return (
+                <PlainButton
```
jstoffan, line 85: Using separate buttons for a toggle will cause the focus to transfer to the document, which is a poor user experience. Let's use a single button and swap the contents, instead.

src/elements/content-sidebar/activity-feed/comment/Replies.js
```diff
@@ -0,0 +1,160 @@
...
+                {replies.map(reply => {
+                    return (
+                        <li key={`${reply.type}${reply.id}`}>
+                            <BaseComment
+                                {...reply}
+                                currentUser={currentUser}
+                                getAvatarUrl={getAvatarUrl}
+                                getMentionWithQuery={getMentionWithQuery}
+                                getUserProfileUrl={getUserProfileUrl}
+                                id={reply.id}
```
jstoffan, line 142: Prefer destructuring rather than direct property accessors.

src/elements/content-sidebar/activity-feed/comment/Replies.js
```diff
@@ -0,0 +1,160 @@
...
+    if (!hasReplies) {
+        return null;
+    }
+
+    return (
+        <div className="bcs-Replies">
+            {renderButton()}
+            <ul className="bcs-Replies-list">
+                {isRepliesLoading && (
+                    <div className="bcs-Replies-loading" data-testid="replies-loading">
```
jstoffan, line 129: I don't think `<ul><div></div></ul>` is valid markup. Is this structure correct?

src/elements/content-sidebar/activity-feed/comment/Replies.js
```diff
@@ -0,0 +1,160 @@
...
+    };
+
+    if (!hasReplies) {
+        return null;
+    }
+
+    return (
+        <div className="bcs-Replies">
+            {renderButton()}
+            <ul className="bcs-Replies-list">
```
jstoffan, line 127: Is this an unordered list? The replies are sorted by date, correct?

src/elements/content-sidebar/activity-feed/comment/Replies.scss
```diff
@@ -0,0 +1,47 @@
+@import '../../../common/variables';
+
+.bcs-Replies {
+    margin: $bdl-grid-unit * 4.5 0;
+    padding-left: 40px;
```
jstoffan, line 5: What is the `40px` value based on? Should this be using the grid, as well?

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -168,133 +173,147 @@ const BaseComment = (props: Props) => {
...
-                                data-testid="comment-actions-menu"
-                                dropdownProps={{
-                                    onMenuOpen: handleMenuOpen,
-                                    onMenuClose: handleMenuClose,
-                                }}
-                                menuProps={{
-                                    'data-resin-component': ACTIVITY_TARGETS.COMMENT_OPTIONS,
-                                }}
+        <div className={classNames('bcs-BaseComment', { reply: !!parentID, parent: !parentID })}>
+            <div className={classNames('bcs-BaseComment-selectWrapper', { reply: !!parentID, parent: !parentID })}>
```
jstoffan, line 177: SUIT CSS prefers descendent naming, so "wrapper" is atypical. Would this be `bcs-BaseComment-content`?

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -168,133 +173,147 @@ const BaseComment = (props: Props) => {
...
+                                {...translations}
+                                translationFailed={error ? true : null}
+                                getUserProfileUrl={getUserProfileUrl}
+                            />
+                        )}
+                    </Media.Body>
+                </Media>
+                {/* $FlowFixMe */}
+                {error ? <ActivityError {...error} /> : null}
+                {!!children && (
```
jstoffan, line 304: I'm not fully following the component API we're trying to create. It seems likely that we'd want to add additional "child" content to a comment at some point. Should this be rethought and this component decomposed?

jstoffan (changes requested):
This PR is already very large, even without unit tests, which is often a symptom of overly-coupled components. Let's try to better isolate the responsibilities of each component to allow us to break up the changes.

JChan106:
> This PR is already very large, even without unit tests, which is often a symptom of overly-coupled components. Let's try to better isolate the responsibilities of each component to allow us to break up the changes.

Good point, ill make some of the changes mentioned in the PR, and break it up once components/changes are more isolated.

src/elements/content-sidebar/activity-feed/comment/Replies.js
JChan106: Good point, will modify

src/elements/content-sidebar/activity-feed/comment/Replies.js
JChan106: yeah didn't realize it's an ordered list, will change

src/elements/content-sidebar/activity-feed/comment/Replies.scss
JChan106: We would be multiplying by 10 if we use the grid unit, which I have hesitencies about since it's such a large number. Maybe using a variable for this case would be better for readability since it's an uncommon value

src/elements/content-sidebar/activity-feed/comment/Replies.scss
```diff
@@ -0,0 +1,47 @@
+@import '../../../common/variables';
+
+.bcs-Replies {
+    margin: $bdl-grid-unit * 4.5 0;
```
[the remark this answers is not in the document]
JChan106: Yeah I think it should be okay in this case, the difference is not noticeable

src/elements/content-sidebar/activity-feed/comment/Replies.js
```diff
@@ -0,0 +1,160 @@
...
+                                getMentionWithQuery={getMentionWithQuery}
+                                getUserProfileUrl={getUserProfileUrl}
+                                id={reply.id}
+                                isPending={reply.isPending}
+                                mentionSelectorContacts={mentionSelectorContacts}
+                                onDelete={onReplyDelete}
+                                onEdit={onReplyEdit}
+                                onSelect={onReplySelect}
+                                parentID={parentID}
+                                permissions={getReplyPermissions(reply)}
```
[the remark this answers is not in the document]
JChan106: Yeah @bfoxx1906 pointed that out in another PR, I created a Jira for it for modifying the props for BaseComment component. If we want to pass in the item/reply object that would also remove the amount of props we have to pass through as well.

src/elements/content-sidebar/activity-feed/comment/Replies.js
```diff
@@ -0,0 +1,160 @@
...
+const Replies = ({
+    currentUser,
+    getAvatarUrl,
+    getMentionWithQuery,
+    getUserProfileUrl,
+    hasReplies = false,
+    isAlwaysExpanded = false,
+    isRepliesLoading,
+    mentionSelectorContacts,
+    onHideReplies = noop,
```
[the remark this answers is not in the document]
JChan106: You mean having `BaseComment` to have the `Replies` component inside instead of passing it in as children? That would also require the reply callback props to be passed in through `BaseComment` right?

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
JChan106: I think I know what you mean? We could in the future make it a compound component, but I'm not sure if it's necessary for now since we don't know if would even have additional child content to a comment in the future. Replies seems like an obvious `child` of a comment to me, and I think thats why the API in the RFC constructed like this. Maybe @bfoxx1906 has some additional thoughts.

JChan106:
> Cleaned up some linting errors, will be extracting out RepliesToggle into another PR

Extracted out Replies toggle into https://github.com/box/box-ui-elements/pull/3294

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -164,49 +173,55 @@ const ActiveState = ({
...
+                                        permissions={{
+                                            can_delete: getProp(item.permissions, 'can_delete', false),
+                                            can_edit: getProp(item.permissions, 'can_edit', false),
+                                            can_reply: getProp(item.permissions, 'can_reply', false),
+                                            can_resolve: getProp(item.permissions, 'can_resolve', false),
+                                        }}
+                                    >
+                                        <Replies
+                                            {...commentProps}
+                                            hasReplies={hasReplies}
```
bfoxx1906, line 196: Do we need to add hasReplies here? If it renders this doesn't that mean that hasReplies is already true enabled also the hasNewThreadedReplies should make this unnecessary right?

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -295,6 +298,7 @@ const BaseComment = (props: Props) => {
             </Media>
             {/* $FlowFixMe */}
             {error ? <ActivityError {...error} /> : null}
+            {!!children && <div className="bcs-Comment-replies">{children}</div>}
```
bfoxx1906, line 301: Are we not including the reply button in this PR? Is that functionality we are looking to implement later?

src/elements/content-sidebar/activity-feed/comment/Replies.js
```diff
@@ -0,0 +1,92 @@
...
+
+    return (
+        <div className="bcs-Replies">
+            <div className="bcs-Replies-content">
+                {isRepliesLoading && (
+                    <div className="bcs-Replies-loading" data-testid="replies-loading">
+                        <LoadingIndicator />
+                    </div>
+                )}
+                <ol className="bcs-Replies-list">
```
bfoxx1906, line 64: So if this component is used outside of ActivityFeed the ol will actually have list style which is not what we want. If we are going to use the ol here let's specify that style in bcs-Replies-list or use a ul instead. Here's the style for reference https://github.com/box/box-ui-elements/blob/master/src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss#L22

bfoxx1906 (changes requested):

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -156,6 +157,14 @@ const ActiveState = ({
                 const isFocused = item === activeFeedItem;
                 const refValue = isFocused ? activeFeedItemRef : undefined;
                 const itemFileVersionId = getProp(item, 'file_version.id');
+                const commentProps = {
```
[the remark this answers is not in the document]
JChan106: Both BaseComment and Replies have duplicate props in this case, so I created a variable to remove duplication. I named it commentProps since they are props needed by BaseComment

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -164,49 +173,55 @@ const ActiveState = ({
...
+                                        onEdit={onCommentEdit}
+                                        onSelect={onCommentSelectHandler(item.id)}
+                                        permissions={{
+                                            can_delete: getProp(item.permissions, 'can_delete', false),
+                                            can_edit: getProp(item.permissions, 'can_edit', false),
+                                            can_reply: getProp(item.permissions, 'can_reply', false),
+                                            can_resolve: getProp(item.permissions, 'can_resolve', false),
+                                        }}
+                                    >
+                                        <Replies
```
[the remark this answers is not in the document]
JChan106: Pros are that we can have an isolation of props going to Replies instead of feeding all of the reply specific props to BaseComment. Theres some callbacks so there would also be one less level to drill. Another pro is that if we want replies to other activity in the future, we have this readable pattern we can follow in ActiveState. 

Cons are that we have duplication of props in `commentProps`. ActiveState seems a bit bulky as well. There's also concerns on the usage of children in this case. If we want Replies to modify the parentComment it would be harder to do as well.

I didn't realize the intention in the RFC was to have the replies in the Comment. I think Jared had some concerns on the usage of children for Replies as well, so I think I can change it

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
JChan106: Yes we are implementing in a separate PR

bfoxx1906: cool beans

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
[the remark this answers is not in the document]
JChan106: Agreed offline to put Replies in the BaseComment. We added Replies to the BaseComment file to avoid the circular dependency warning.

bfoxx1906 (approved):

src/elements/content-sidebar/activity-feed/activity-feed/ActivityItem.tsx
```diff
@@ -6,11 +6,24 @@ type Props = {
...
 
-function ActivityItem({ children, className, isFocused, ...rest }: Props, ref: React.Ref<HTMLLIElement>) {
+function ActivityItem(
+    { children, className, isFocused, isHoverable = false, hasNewThreadedReplies = false, ...rest }: Props,
+    ref: React.Ref<HTMLLIElement>,
+) {
     return (
-        <li className={classNames('bcs-ActivityItem', className, { 'bcs-is-focused': isFocused })} ref={ref} {...rest}>
+        <li
+            className={classNames('bcs-ActivityItem', className, {
```
karelee7, line 19: For future reference, may be more readable to pull this out into a separate variable!

karelee7 (approved):

PR #3290, opened 2023-03-23
fix(aci-toggle): info badge vertical alignment

arturfrombox:


apieceofbart (approved):

src/features/advanced-content-insights/AdvancedContentInsightsToggle.scss
```diff
@@ -5,4 +5,8 @@
 .AdvancedContentInsightsToggle-icon {
     display: inline-block;
     margin-left: 5px;
+
+    > svg {
+        vertical-align: middle;
+    }
```
tjuanitas, line 11: `vertical-align` has some weird quirks that I feel isn't very intuitive

could we use a flex solution? maybe add `display: flex` and `align-items: center` to `AdvancedContentInsightsToggle-icon` instead?

arturfrombox: hi @tjuanitas in this case we would need also to apply `display: flex;` to the `.toggle-simple-label`. let me know if thats ok ?

tjuanitas: ahh okay, I didn't know that. Yeah, that should be fine! as long as it's contained within this component

so something like this:
```
.AdvancedContentInsightsToggle {
    ...

    .toggle-simple-label {
        display: flex;
    }
}

.AdvancedContentInsightsToggle-icon {
    display: flex;
    align-items: center
    margin-left: 5px;
}
```

arturfrombox:
hi @tjuanitas, i updated this pr. let me know if any more changes is needed.

tjuanitas (approved):
lgtm, thanks!

apieceofbart (approved):

PR #3292, opened 2023-03-25
fix(collapsible-sidebar): Refactor for modularity

greathmaster:
Refactored Collapsible Sidebar to consistently be a collapsible sidebar for larger screen widths instead of switching to a fixed size sidebar. Such customization should be done in client applications.

The prop `wrapperClass` was added to apply CSS styles to the CollapsibleSidebar-wrapper div.

Some styles were removed and moved to client application.


tjuanitas:
Ended up adding specific styles to EUA to not break default functionality for the parent apps (EUA, Admin Console, Dev Console) and easier to keep them in sync. added necessary changes here: https://github.com/box/box-ui-elements/pull/3295

PR #3294, opened 2023-03-29
feat(threaded-replies): Create RepliesToggle component

JChan106:
Splitting up the Replies PR: https://github.com/box/box-ui-elements/pull/3288  for easier review

### DIFF
--- a/src/elements/content-sidebar/activity-feed/comment/RepliesToggle.js
+++ b/src/elements/content-sidebar/activity-feed/comment/RepliesToggle.js
@@ -8,21 +8,14 @@ import PlainButton from '../../../../components/plain-button';
 import './RepliesToggle.scss';
 
 type Props = {
-    isAlwaysExpanded?: boolean,
     onHideReplies: (index: number) => void,
     onShowReplies: () => void,
     repliesShownCount: number,
     repliesTotalCount: number,
 };
 
-const RepliesToggle = ({
-    isAlwaysExpanded = false,
-    onShowReplies,
-    onHideReplies,
-    repliesShownCount,
-    repliesTotalCount,
-}: Props) => {
-    if (isAlwaysExpanded || repliesTotalCount <= 1) {
+const RepliesToggle = ({ onShowReplies, onHideReplies, repliesShownCount, repliesTotalCount }: Props) => {
+    if (repliesTotalCount <= 1) {
         return null;
     }
 
--- a/src/elements/content-sidebar/activity-feed/comment/RepliesToggle.js
+++ b/src/elements/content-sidebar/activity-feed/comment/RepliesToggle.js
@@ -5,6 +5,8 @@ import { FormattedMessage } from 'react-intl';
 import messages from './messages';
 import PlainButton from '../../../../components/plain-button';
 
+import './RepliesToggle.scss';
+
 type Props = {
     isAlwaysExpanded?: boolean,
     isRepliesLoading: boolean,
@@ -39,7 +41,7 @@ const RepliesToggle = ({
     };
 
     return (
-        <PlainButton className="bcs-Replies-toggle" onClick={handleToggle} type="button">
+        <PlainButton className="bcs-RepliesToggle" onClick={handleToggle} type="button">
             <FormattedMessage values={{ repliesToLoadCount }} {...toggleMessage} />
         </PlainButton>
     );
--- a/src/elements/content-sidebar/activity-feed/comment/RepliesToggle.js
+++ b/src/elements/content-sidebar/activity-feed/comment/RepliesToggle.js
@@ -1,0 +1,48 @@
+// @flow
+import * as React from 'react';
+import { FormattedMessage } from 'react-intl';
+
+import messages from './messages';
+import PlainButton from '../../../../components/plain-button';
+
+type Props = {
+    isAlwaysExpanded?: boolean,
+    isRepliesLoading: boolean,
+    onHideReplies: (index: number) => void,
+    onShowReplies: () => void,
+    repliesShownCount: number,
+    repliesTotalCount: number,
+};
+
+const RepliesToggle = ({
+    isAlwaysExpanded = false,
+    isRepliesLoading,
+    onShowReplies,
+    onHideReplies,
+    repliesShownCount,
+    repliesTotalCount,
+}: Props) => {
+    if (isAlwaysExpanded || isRepliesLoading || repliesTotalCount <= 1) {
+        return null;
+    }
+
+    const hasMoreRepliesToShow = repliesTotalCount > repliesShownCount;
+    const toggleMessage = hasMoreRepliesToShow ? messages.showReplies : messages.hideReplies;
+    const repliesToLoadCount = Math.max(repliesTotalCount - repliesShownCount, 0);
+
+    const handleToggle = () => {
+        if (hasMoreRepliesToShow) {
+            onShowReplies();
+        } else if (repliesShownCount) {
+            onHideReplies(repliesShownCount - 1);
+        }
+    };
+
+    return (
+        <PlainButton className="bcs-Replies-toggle" onClick={handleToggle} type="button">
+            <FormattedMessage values={{ repliesToLoadCount }} {...toggleMessage} />
+        </PlainButton>
+    );
+};
+
+export default RepliesToggle;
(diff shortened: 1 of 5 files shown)


src/elements/content-sidebar/activity-feed/comment/RepliesToggle.js
```diff
@@ -0,0 +1,50 @@
...
+
+const RepliesToggle = ({
+    isAlwaysExpanded = false,
+    isRepliesLoading,
+    onShowReplies,
+    onHideReplies,
+    repliesShownCount,
+    repliesTotalCount,
+}: Props) => {
+    if (isAlwaysExpanded || isRepliesLoading || repliesTotalCount <= 1) {
```
karelee7, line 27: Pull logic out into separate variable for readability?

src/elements/content-sidebar/activity-feed/comment/__tests__/RepliesToggle.test.js
```diff
@@ -0,0 +1,82 @@
...
+                onHideReplies={hideReplies}
+                repliesShownCount={shownCount}
+                repliesTotalCount={totalCount}
+                {...props}
+            />
+        </IntlProvider>,
+    );
+
+describe('elements/content-sidebar/ActivityFeed/comment/RepliesToggle', () => {
+    test('should correctly render show toggle', () => {
```
karelee7, line 33: Do we also need tests for when `isAlwaysExpanded` and/or `isRepliesLoading` is true?

src/elements/content-sidebar/activity-feed/comment/RepliesToggle.js
JChan106: Don't think it would improve readability in this case since each variable has its own reason to not render the replies component.

JChan106: Now that you say it, I think I can remove isRepliesLoading since it's essentially the same type of prop as isAlwaysExpanded. We can just have the parent determine the rendering of the toggle based on the loading.

src/elements/content-sidebar/activity-feed/comment/__tests__/RepliesToggle.test.js
```diff
@@ -0,0 +1,91 @@
...
+        const countDifference = totalCount - 1;
```
patlm, line 34: The value of this variable is confusing me a bit. Is the `- 1` because `shownCount` is initially 1? Or is it an index offset or something different?

src/elements/content-sidebar/activity-feed/comment/RepliesToggle.js
```diff
@@ -0,0 +1,48 @@
...
+    repliesShownCount,
+    repliesTotalCount,
+}: Props) => {
+    if (isAlwaysExpanded || repliesTotalCount <= 1) {
+        return null;
+    }
+
+    const hasMoreRepliesToShow = repliesTotalCount > repliesShownCount;
+    const toggleMessage = hasMoreRepliesToShow ? messages.showReplies : messages.hideReplies;
+    const repliesToLoadCount = Math.max(repliesTotalCount - repliesShownCount, 0);
```
bfoxx1906, line 31: Is there a case where repliesTotalCount can actually be less than the repliesShownCount? Also if that value is 0 should we even be showing the toggle message?

src/elements/content-sidebar/activity-feed/comment/RepliesToggle.js
```diff
@@ -0,0 +1,48 @@
...
+};
+
+const RepliesToggle = ({
+    isAlwaysExpanded = false,
+    onShowReplies,
+    onHideReplies,
+    repliesShownCount,
+    repliesTotalCount,
+}: Props) => {
+    if (isAlwaysExpanded || repliesTotalCount <= 1) {
```
bfoxx1906, line 25: parent should probably determine whether this is shown or not

src/elements/content-sidebar/activity-feed/comment/RepliesToggle.js
JChan106: There shouldn't be a case where total count is less than shown count. And if there is, the `repliesTotalCount <= 1` conditional will take care of it. We have a test to capture that case as well, and the toggle is not shown

src/elements/content-sidebar/activity-feed/comment/__tests__/RepliesToggle.test.js
JChan106: Yeah it's because the shownCount is initially 1, so when creating the message it would use the difference: `See ${difference} replies`

src/elements/content-sidebar/activity-feed/comment/RepliesToggle.js
bfoxx1906: We had some questions about how we want to deal with this in our BDD session but we can just leave it as be for now.

src/elements/content-sidebar/activity-feed/comment/RepliesToggle.js
```diff
@@ -0,0 +1,41 @@
...
+
+type Props = {
+    onHideReplies: (index: number) => void,
+    onShowReplies: () => void,
+    repliesShownCount: number,
+    repliesTotalCount: number,
+};
+
+const RepliesToggle = ({ onShowReplies, onHideReplies, repliesShownCount, repliesTotalCount }: Props) => {
+    if (repliesTotalCount <= 1) {
```
bfoxx1906, line 18: shouldn't the parent control whether or not this component is visible or not?

JChan106: In this case, I think checking if the count passes edgecase criteria makes sense. I guess the parent can check if repliesTotalCount is greater than 0, but shouldn't that be a given for rendering the show/reply toggle? This is checking for edgecases, which I don't think the parent should control.

JChan106: If the logic for hide/show was in the parent then it would make sense to move the conditional. But since this component is not a dumb component, we should have some guards for the logic

src/elements/content-sidebar/activity-feed/comment/RepliesToggle.js
```diff
@@ -0,0 +1,41 @@
...
+
+    const hasMoreRepliesToShow = repliesTotalCount > repliesShownCount;
+    const toggleMessage = hasMoreRepliesToShow ? messages.showReplies : messages.hideReplies;
+    const repliesToLoadCount = Math.max(repliesTotalCount - repliesShownCount, 0);
+
+    const handleToggle = () => {
+        if (hasMoreRepliesToShow) {
+            onShowReplies();
+        } else if (repliesShownCount) {
+            onHideReplies(repliesShownCount - 1);
```
bfoxx1906, line 30: Does this match how onHideReplies is currently being used? How do you plan on using repliesShownCount - 1 to actually hide the replies?

JChan106: Yes it's going to be working the same way it's currently being used. The toggle passes back the last index shown, and Replies component will pass back the last reply shown.

karelee7 (approved):
Looks good!

src/elements/content-sidebar/activity-feed/comment/RepliesToggle.js
bfoxx1906: This looks good for now. It matches the current functionality. Let's revisit how we are showing/hiding the replies as we integrate it with the other components and see if there's a way of simplifying that process.

src/elements/content-sidebar/activity-feed/comment/__tests__/RepliesToggle.test.js
```diff
@@ -0,0 +1,82 @@
+// @flow
+
+import React from 'react';
+import { IntlProvider } from 'react-intl';
+import { fireEvent, render, screen } from '@testing-library/react';
+import RepliesToggle from '../RepliesToggle';
+
+jest.mock('react-intl', () => ({
+    ...jest.requireActual('react-intl'),
```
bfoxx1906, line 9: Why do we need to require the actual implementation while just mocking it here? Were you seeing errors by not doing this?

JChan106: Yeah seems like it just can't find any of the messages without this. I believe its throughout the sidebar

bfoxx1906: Ok. This is used elsewhere in the code. Will reach out to Wiola and find out why.

bfoxx1906 (approved):

PR #3298, opened 2023-04-06
feat(threaded-replies): Create Reply button component

JChan106:
This PR includes the CreateReply component which will be used in the Replies component in another PR. This component is a toggle which switches from a button to a Comment form based on the props.

- [x]  Unit Tests

### DIFF
--- a/src/elements/content-sidebar/activity-feed/comment-form/CommentForm.js
+++ b/src/elements/content-sidebar/activity-feed/comment-form/CommentForm.js
@@ -126,11 +126,10 @@ class CommentForm extends React.Component<Props, State> {
         const inputContainerClassNames = classNames('bcs-CommentForm', className, {
             'bcs-is-open': isOpen,
         });
-        const showAvatar = !isEditing && !!user && !!getAvatarUrl;
 
         return (
             <Media className={inputContainerClassNames}>
-                {showAvatar && (
+                {!isEditing && !!user && !!getAvatarUrl && (
                     <Media.Figure className="bcs-CommentForm-avatar">
                         <Avatar getAvatarUrl={getAvatarUrl} user={user} />
                     </Media.Figure>
--- a/src/elements/content-sidebar/activity-feed/comment-form/CommentForm.js
+++ b/src/elements/content-sidebar/activity-feed/comment-form/CommentForm.js
@@ -129,7 +129,7 @@ class CommentForm extends React.Component<Props, State> {
 
         return (
             <Media className={inputContainerClassNames}>
-                {!isEditing && !!user && !!getAvatarUrl && (
+                {!isEditing && !!user && (
                     <Media.Figure className="bcs-CommentForm-avatar">
                         <Avatar getAvatarUrl={getAvatarUrl} user={user} />
                     </Media.Figure>
--- a/src/elements/content-sidebar/activity-feed/comment-form/CommentForm.js
+++ b/src/elements/content-sidebar/activity-feed/comment-form/CommentForm.js
@@ -27,7 +27,7 @@ type Props = {
     contactsLoaded?: boolean,
     createComment?: Function,
     entityId?: string,
-    getAvatarUrl: GetAvatarUrlCallback,
+    getAvatarUrl?: GetAvatarUrlCallback,
     getMentionWithQuery?: Function,
     isDisabled?: boolean,
     isEditing?: boolean,
@@ -40,7 +40,7 @@ type Props = {
     showTip?: boolean,
     tagged_message?: string,
     updateComment?: Function,
-    user: User,
+    user?: User,
 } & InjectIntlProvidedProps;
 
 type State = {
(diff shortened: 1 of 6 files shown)


src/elements/content-sidebar/activity-feed/comment-form/CommentForm.js
```diff
@@ -129,7 +129,7 @@ class CommentForm extends React.Component<Props, State> {
 
         return (
             <Media className={inputContainerClassNames}>
-                {!isEditing && (
+                {!isEditing && !!user && !!getAvatarUrl && (
```
karelee7, line 132: Should we pull this logic out into a separate variable for readability?

src/elements/content-sidebar/activity-feed/comment-form/CommentForm.js
```diff
@@ -27,7 +27,7 @@ type Props = {
     contactsLoaded?: boolean,
     createComment?: Function,
     entityId?: string,
-    getAvatarUrl: GetAvatarUrlCallback,
+    getAvatarUrl?: GetAvatarUrlCallback,
```
karelee7, line 30: Curious, what do we do when there is no avatar url? Do we show anything in place of the avatar?

src/elements/content-sidebar/activity-feed/comment/CreateReply.js
```diff
@@ -0,0 +1,72 @@
+// @flow
+
+import * as React from 'react';
+import { FormattedMessage, injectIntl } from 'react-intl';
+import type { IntlShape } from 'react-intl';
+import noop from 'lodash/noop';
+
+import PlainButton from '../../../../components/plain-button';
+import ArrowArcRight from '../../../../icon/fill/ArrowArcRight';
```
karelee7, line 9: minor: alphabetical order

src/elements/content-sidebar/activity-feed/comment/CreateReply.js
```diff
@@ -0,0 +1,72 @@
...
+import type { SelectorItems } from '../../../../common/types/core';
+
+import CommentForm from '../comment-form';
+
+import messages from './messages';
+import './CreateReply.scss';
+
+type Props = {
+    getMentionWithQuery?: (searchStr: string) => void,
+    intl: IntlShape,
```
karelee7, line 20: Do we have to include this in the props? Could we just import `intl` to use below?

src/elements/content-sidebar/activity-feed/comment/__tests__/CreateReply.test.js
```diff
@@ -0,0 +1,119 @@
...
+    });
+
+    test('should correctly render CreateReply form', () => {
+        getWrapper({ showReplyForm: true });
+
+        expect(screen.getByText(replyInThreadMessage)).toBeVisible();
+        expect(screen.queryByText(replyMessage)).not.toBeInTheDocument();
+    });
+
+    test('should disable Reply button if isDisabled property is true', () => {
```
karelee7, line 58: Should we add a test for when `isDisabled` is false?

src/elements/content-sidebar/activity-feed/comment-form/CommentForm.js
JChan106: Tried doing that but it gives some linting errors

src/elements/content-sidebar/activity-feed/comment-form/CommentForm.js
JChan106: In this case we wouldn't show anything. Previously we would require the cb, but the reply form doesn't need an avatar so I made it optional, and to just not show the avatar if there's no cb passed in.

src/elements/content-sidebar/activity-feed/comment/__tests__/CreateReply.test.js
JChan106: isDisabled is false by default.

src/elements/content-sidebar/activity-feed/comment-form/CommentForm.js
JChan106: Actually looks like `Avatar.js` doesn't require a `getAvatarUrl` in the component. So I will remove that boolean from the conditional.

src/elements/content-sidebar/activity-feed/comment/__tests__/CreateReply.test.js
```diff
@@ -0,0 +1,119 @@
...
+        getWrapper();
+
+        fireEvent.click(screen.getByText(replyMessage));
+        expect(onClick).toBeCalledTimes(1);
+        expect(onSubmit).not.toBeCalled();
+        expect(onFocus).not.toBeCalled();
+        expect(onCancel).not.toBeCalled();
+    });
+
+    test('should not show form when isDisabled', () => {
```
karelee7, line 78: minor: `should not show form when isDisabled` -> `should not show form when isDisabled is true`

karelee7, line 78: When `isDisabled` is false, do we show this form? If so, do we need a test for that??

JChan106: there's a test for that already, when you don't have the `isDisabled` as a prop it defaults to false.

PR #3306, opened 2023-04-14
fix(usm): Order person before icon in collaborator list DOM

tjuanitas:
Recommendation from WebAIM is to order the persons name before the icon in the DOM. I will follow up with PM to confirm this is the approach we want to do.

Solution: move the `info` div before the `CollaboratorAvatarItem` component so that it appears first in the DOM and use `flex-direction: row-reverse` to fix the order in the UI.

### DIFF
--- a/src/features/collaborator-avatars/CollaboratorListItem.scss
+++ b/src/features/collaborator-avatars/CollaboratorListItem.scss
@@ -74,3 +74,7 @@
         }
     }
 }
+
+.bdl-CollaboratorListItem-user {
+    flex-direction: row-reverse;
+}
--- a/src/features/collaborator-avatars/CollaboratorListItem.js
+++ b/src/features/collaborator-avatars/CollaboratorListItem.js
@@ -55,7 +55,11 @@ const CollaboratorListItem = (props: Props) => {
     return (
         <li>
             <div className="collaborator-list-item">
-                <div className="user">
+                <div className="bdl-CollaboratorListItem-user user">
+                    <div className="info">
+                        {userOrGroupNameContent}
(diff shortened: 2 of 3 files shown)


src/features/collaborator-avatars/CollaboratorListItem.scss
```diff
@@ -74,3 +74,7 @@
         }
     }
 }
+
+.bdl-CollaboratorListItem-user {
```
tjuanitas, line 78: I created a new class because the `user` class is on multiple levels in this component.

e.g. `collaborator-list-item > user > info > user`

megansmith-box, line 78: Should we move this within `collaborator-list-item`? Does it make sense to introduce the standard `bdl-CollaboratorListItem` when all the other classes don't follow this pattern in this file?

tjuanitas: this file has super weird classes `info`, `role`, `user`, etc.

normally I think it might be okay to use the current file's pattern but because these are so generic that it'll likely conflict with customer's classes that I think it'd be good to lean towards using new classnames in this case

tjuanitas: > Should we move this within `collaborator-list-item`?

with the new SUIT class, I left it outside of the parent scope block so that it doesn't have the extra specificity

megansmith-box: Ok sounds good

megansmith-box (approved):
The change looks good from my perspective

PR #3307, opened 2023-04-17
feat(aci-toggle): add separate prop for description

arturfrombox:
What was changed?
added separate prop to AdvancedContentInsightsToggle component in order to control description component separately

### DIFF
--- a/src/features/advanced-content-insights/AdvancedContentInsightsToggle.tsx
+++ b/src/features/advanced-content-insights/AdvancedContentInsightsToggle.tsx
@@ -11,6 +11,7 @@ import messages from './messages';
 import './AdvancedContentInsightsToggle.scss';
 
 interface Props {
+    hasDescription?: boolean;
     hasTooltip?: boolean;
     isChecked?: boolean;
     isDisabled: boolean;
@@ -18,12 +19,18 @@ interface Props {
 }
 
 const AdvancedContentInsightsToggle = ({
+    hasDescription,
     hasTooltip = true,
     isChecked = false,
     isDisabled,
     onChange = noop,
 }: Props) => {
-    const description = <FormattedMessage {...messages.advancedContentInsightsDescription} />;
+    const description = hasDescription ? (
+        <FormattedMessage {...messages.advancedContentInsightsDescription} />
+    ) : (
+        undefined
+    );
+    const tooltipText = <FormattedMessage {...messages.advancedContentInsightsTooltipText} />;
     const label = (
         <>
             <FormattedMessage
@@ -32,7 +39,7 @@ const AdvancedContentInsightsToggle = ({
                     : messages.advancedContentInsightsTitleDisabled)}
             />
             {hasTooltip && (
-                <Tooltip text={description}>
+                <Tooltip text={tooltipText}>
                     <div className="AdvancedContentInsightsToggle-icon">
                         <InfoBadge16 height={14} width={14} />
                     </div>
@@ -45,7 +52,7 @@ const AdvancedContentInsightsToggle = ({
         <Toggle
             className="AdvancedContentInsightsToggle"
             data-testid="insights-toggle"
-            description={!hasTooltip && !isChecked && description}
+            description={description}
             isDisabled={isDisabled}
             isOn={isChecked}
             label={label}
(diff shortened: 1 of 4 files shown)


marcoartaviaq (dismissed):

src/features/advanced-content-insights/AdvancedContentInsightsToggle.tsx
```diff
@@ -11,19 +11,26 @@ import messages from './messages';
 import './AdvancedContentInsightsToggle.scss';
 
 interface Props {
+    hasDescription?: boolean;
```
tjuanitas, line 14: are the description and tooltip now independent? does that mean someone could use both at the same time? is that intended?

src/features/advanced-content-insights/messages.ts
```diff
@@ -17,6 +17,12 @@ const messages = defineMessages({
         description: 'Description text about advanced content insights.',
         id: 'boxui.advancedContentInsights.advancedContentInsightsDescription',
     },
+    advancedContentInsightsTooltipText: {
```
tjuanitas, line 20: since this is the same as `advancedContentInsightsDescription`, do we need separate messages? do we anticipate that the messages will diverge in the future?

src/features/advanced-content-insights/AdvancedContentInsightsToggle.tsx
```diff
@@ -45,7 +52,7 @@ const AdvancedContentInsightsToggle = ({
         <Toggle
             className="AdvancedContentInsightsToggle"
             data-testid="insights-toggle"
-            description={!hasTooltip && !isChecked && description}
+            description={description}
```
tjuanitas, line 55: would it work to keep:
`const description = <FormattedMessage ...` ~line 28
and
`<Tooltip text={description}>` ~line 42
the way it was previously and do:
`description={hasDescription && description}`
instead?

src/features/advanced-content-insights/AdvancedContentInsightsToggle.tsx
arturfrombox: yep, we want to have it controlled by separate prop ,but we can add !hasTooltip flag to avoid having both tooltip and description at the same time

src/features/advanced-content-insights/AdvancedContentInsightsToggle.tsx
arturfrombox: did that in anticipation that product will want to change text and possibly to have different messages. but for now message can be shared between tooltip and description

arturfrombox:
Hi @tjuanitas , thanks for reviewing this pr. I updated it so please let me know if more changes is needed.

src/features/advanced-content-insights/AdvancedContentInsightsToggle.tsx
tjuanitas: is there a scenario in which we show neither a tooltip nor description? if not, why do we need the separate prop?

tjuanitas (approved):
left another question but looks fine!

src/features/advanced-content-insights/AdvancedContentInsightsToggle.tsx
arturfrombox: thanks, good question. previously description was based on toggle state (!isChecked), but product requirement is different now and we need to add more conditions for description to show up. that why we need separate prop

marcoartaviaq (approved):

src/features/advanced-content-insights/AdvancedContentInsightsToggle.tsx
tjuanitas: couldn't they still be controlled with one prop? the additional conditions would just be added into the `hasTooltip` logic

for example:
```
// instead of:
<AdvancedContentInsightsToggle
    hasDescription={condition && otherCondition}
    hasTooltip={true}
    ...
/>

// it would be:
<AdvancedContentInsightsToggle
    hasTooltip={!(condition && otherCondition)}
    ...
/>
```

this is with the assumption that if the description conditions is true, then the tooltip is not shown. but if the description is not shown, then the tooltip is rendered

PR #3316, opened 2023-04-27
feat(uaa-integration): Integrate UAA into ActivitySidebar

JChan106:
This PR is to integrate UAA into the activity sidebar. Because of the large change in response for `versions` we will separate that out for another PR as that will likely require a new component. There are some changes that are still needed on the BE for tasks, however I believe it is in a good state for early review.

- [x] Screenshots
- [x] Unit tests?


src/api/Feed.js
```diff
@@ -405,27 +418,57 @@ class Feed extends Base {
...
+            ? this.fetchFileActivities(permissions, [
+                  UAA_ACTIVITY_TYPE_ANNOTATION,
+                  UAA_ACTIVITY_TYPE_APP_ACTIVITY,
+                  UAA_ACTIVITY_TYPE_COMMENT,
+                  UAA_ACTIVITY_TYPE_TASK,
+              ])
+            : Promise.resolve();
+
+        if (shouldUseUAA) {
+            Promise.all([versionsPromise, currentVersionPromise, fileActivitiesPromise]).then(
```
jstoffan, line 431: Lots of nesting here. Can we break each branch of logic into separate methods?

src/elements/content-sidebar/activity-feed/task-new/Task.js
```diff
@@ -71,6 +71,7 @@ type Props = {|
     onEdit?: Function,
     onModalClose?: Function,
     onView?: Function,
+    shouldUseUAA?: boolean,
```
jstoffan, line 74: Can we normalize this in the API layer, instead?

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -190,6 +192,7 @@ const BaseComment = (props: BaseCommentProps) => {
         mentionSelectorContacts,
         translations,
     };
+    const displayMessage = tagged_message || message;
```
jstoffan, line 195: Can we normalize this in the API layer, instead?

src/api/utils.js
```diff
@@ -25,6 +26,51 @@ export const formatComment = (comment: Comment): Comment => {
...
+export const getUAAQueryParams = (fileID: string, activityTypes?: UAAActivityTypes[], enableReplies?: boolean) => {
+    const baseEndpoint = `/file_activities?file_id=${fileID}`;
+    const hasFilteredActivityTypes = !!activityTypes && !!activityTypes.length;
+
+    return `${baseEndpoint}${hasFilteredActivityTypes ? `&activity_types=${activityTypes.join()}` : ''}${
+        enableReplies ? '&enable_replies=true&reply_limit=1' : ''
+    }`;
+};
+
+export const parseFileActivitiesResponseForFeed = (data?: UAAFileActivity[]) => {
```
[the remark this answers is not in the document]
JChan106: Good point, think it could go either way since it is a utils function, but since it's very specific to the FileActivities API currently, I think it's good to have it there

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -190,6 +190,7 @@ const BaseComment = (props: BaseCommentProps) => {
         mentionSelectorContacts,
         translations,
     };
+    // const displayMessage = tagged_message || message;
```
jstoffan, line 193: Can we remove this commented code?

jstoffan (commented):
Looks like we're nearly there.

src/api/FileActivities.js
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    const parsedData = [];
+
+    data.map(item => {
+        if (item.source) {
+            // UAA follows a lowercased enum naming convention, convert to uppercase to align with task api
+            const taskItem = item.source[FILE_ACTIVITY_TYPE_TASK];
+            const commentItem = item.source[FILE_ACTIVITY_TYPE_COMMENT];
+            if (taskItem) {
```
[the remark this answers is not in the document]
JChan106: Might be able to by getting the first item of `item.source.keys()`, but this method seems less readable than using if/else statements.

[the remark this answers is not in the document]
jstoffan: It looks like there's an `item.type` field?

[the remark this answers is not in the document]
JChan106: ah you're right, switch statement might work then, ill try it out

src/api/FileActivities.js
```diff
@@ -0,0 +1,166 @@
...
+        return `${this.getBaseApiUrl()}${getFileActivityQueryParams(id, undefined, true)}`;
+    }
+
+    /**
+     * API URL for filtered file activities
+     *
+     * @param {string} [id] - a box file id
+     * @return {string} base url for files
+     */
+    getFilteredUrl(id?: string, activityTypes?: FileActivityTypes[]): string {
```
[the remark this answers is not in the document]
JChan106: We don't, however I did encounter a linting error when trying to add `activityTypes` params to `getUrl` since the `Base` component is expecting 1 parameter for `getUrl`. Thats why I created a separate method.

src/api/Feed.js
JChan106: Thought it broken down enough after splitting the success callback and error callback. Eventually the FF conditional `shouldUseUAA` will go away, leaving it with the `all` array and the `then` statement. Are you suggesting to create methods for `all` and `then`?

jstoffan: My thought was each branch of the if/else could be its own method, but it's probably not necessary given the short lifetime of the else branch. Feel free to disregard.

tjuanitas:
sorry i'm not familiar but what does UAA stand for?

JChan106:
> sorry i'm not familiar but what does UAA stand for?

Unified Activity API

src/api/Feed.js
```diff
@@ -36,6 +38,10 @@ import {
     TASK_NEW_NOT_STARTED,
     TYPED_ID_FEED_PREFIX,
     TASK_MAX_GROUP_ASSIGNEES,
+    FILE_ACTIVITY_TYPE_ANNOTATION,
```
jstoffan, line 41: Minor, but can we alphabetize these imports?

src/api/Feed.js
```diff
@@ -95,6 +103,75 @@ const getItemWithPendingReply = <T: { replies?: Array<Comment> }>(item: T, reply
...
+    const parsedData = [];
+
+    data.map(item => {
+        if (!item.source) {
+            return item;
+        }
+
+        const source = { ...item.source };
+        // $FlowFixMe
+        const commentItem: Comment = source[FILE_ACTIVITY_TYPE_COMMENT];
```
jstoffan, line 121: Should these be within the `case` statements?

src/api/Feed.js
```diff
@@ -95,6 +103,75 @@ const getItemWithPendingReply = <T: { replies?: Array<Comment> }>(item: T, reply
...
+        // $FlowFixMe
+        const taskItem: Task = source[FILE_ACTIVITY_TYPE_TASK];
+
+        switch (item.type) {
+            case FILE_ACTIVITY_TYPE_TASK:
+                // UAA follows a lowercased enum naming convention, convert to uppercase to align with task api
+                if (!!taskItem.assigned_to && !!taskItem.assigned_to.entries) {
```
jstoffan, line 128: Does BUIE support optional chaining? Is the `!!` needed?

src/api/Feed.js
```diff
@@ -95,6 +103,75 @@ const getItemWithPendingReply = <T: { replies?: Array<Comment> }>(item: T, reply
...
+                    // $FlowFixMe
+                    taskItem.assigned_to.entries.map(entry => {
+                        entry.role = entry.role.toUpperCase();
+                        entry.status = entry.status.toUpperCase();
+
+                        return entry;
```
jstoffan, line 134: We're still mutating the underlying `entry` object. Can we return a new object using destructuring/rest operators?

src/api/Feed.js
```diff
@@ -95,6 +103,75 @@ const getItemWithPendingReply = <T: { replies?: Array<Comment> }>(item: T, reply
...
+                }
+                if (taskItem.status && isString(taskItem.status)) {
+                    taskItem.status = taskItem.status.toUpperCase();
+                }
+                if (taskItem.task_type && isString(taskItem.task_type)) {
+                    taskItem.task_type = taskItem.task_type.toUpperCase();
+                }
+
+                // $FlowFixMe
+                taskItem.created_by = {
```
jstoffan, line 148: Same comment on mutation.

src/api/Feed.js
```diff
@@ -394,38 +478,77 @@ class Feed extends Base {
         this.file = file;
         this.errors = [];
         this.errorCallback = onError;
-        const annotationsPromise = shouldShowAnnotations
-            ? this.fetchAnnotations(permissions, shouldShowReplies)
+        const annotationsPromise =
+            !shouldUseUAA && shouldShowAnnotations
```
jstoffan, line 482: This is really hard to read. If we don't need to make these calls at all, should we have a separate UAA-specific `feedItems` method rather than passing a `shouldUseUAA` flag?

src/api/Feed.js
```diff
@@ -394,38 +478,77 @@ class Feed extends Base {
...
+                    if (!feedItems || !feedItems.length) {
+                        return;
+                    }
+
+                    const fileActivitiesResponse = feedItems[0];
+                    const versionsWithCurrent = currentVersion
+                        ? this.versionsAPI.addCurrentVersion(currentVersion, versions, this.file)
+                        : undefined;
+                    const parsedFeedItems = parseFileActivitiesResponseForFeed(fileActivitiesResponse);
+                    // $FlowFixMe
```
jstoffan, line 531: Why is this needed?

src/api/FileActivities.js
```diff
@@ -0,0 +1,106 @@
...
+const getFileActivityQueryParams = (
+    fileID: string,
+    activityTypes?: FileActivityTypes[] = [],
+    enableReplies?: boolean = true,
+) => {
+    const baseEndpoint = `/file_activities?file_id=${fileID}`;
+    const hasActivityTypes = !!activityTypes && !!activityTypes.length;
+    const enabledRepliesQueryParam = '&enable_replies=true&reply_limit=1';
+
+    return `${baseEndpoint}${hasActivityTypes ? `&activity_types=${activityTypes.join()}` : ''}${
```
jstoffan, line 22: Still a bit tough to see what's going on here. Can we use the `query-string` library to help us? It's already in the dependencies list.

src/api/Feed.js
JChan106: We get a `Unexpected lexical declaration in case block` eslint message

jstoffan: Need to use block syntax for the case statements, which is generally a good idea anyway.

```
case:
    // whatever


case: {
    // whatever
}
```

src/api/FileActivities.js
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        return `${this.getBaseApiUrl()}${getFileActivityQueryParams(id, undefined, true)}`;
+    }
+
+    /**
+     * API URL for filtered file activities
+     *
+     * @param {string} [id] - a box file id
+     * @return {string} base url for files
```
jstoffan, line 46: Missing a param in the JSDocs here.

src/api/Feed.js
JChan106: I believe it does based on other usages in BUIE, linter removes the `!!` after optional chaining too

src/api/Feed.js
JChan106: Would it not be modifying just the cloned source object here?: `const source = { ...item.source };`

src/api/Feed.js
JChan106: Not sure if it's necessary since all of these promises should be going away after `shouldUseUAA` FF is gone. I can add a comment around this to make it easier to read.

src/api/Feed.js
JChan106: Flow is complaining that Comments/Annotations/Tasks, etc. is incompatable with an empty array literal. I've tried allowing an empty array as a parameter in `parseFileActivitiesResponseForFeed` however that did not fix the problem either...

`Cannot call sortFeedItems because:
 • in array element:
    • Either empty array literal [1] is incompatible with Comments [2].
    • Or empty array literal [1] is incompatible with ThreadedComments [3].
    • Or empty array literal [1] is incompatible with Tasks [4].
    • Or empty array literal [1] is incompatible with FileVersions [5].
    • Or empty array literal [1] is incompatible with AppActivityItems [6].
    • Or empty array literal [1] is incompatible with Annotations [7].
 • in array element:
    • Either property total_count is missing in object literal [8] but exists in Comments [2].
    • Or property limit is missing in object literal [8] but exists in ThreadedComments [3].
    • Or property next_marker is missing in object literal [8] but exists in Tasks [4].
    • Or property total_count is missing in object literal [8] but exists in FileVersions [5].
    • Or property total_count is missing in object literal [8] but exists in AppActivityItems [6].
    • Or property limit is missing in object literal [8] but exists in Annotations [7].`

src/api/FileActivities.js
JChan106: Tried out using query-string, but it seems like our version is very limited in functionality. Seems like it cannot use commas as separators for arrays, as well as recognize commas in `stringify` so I couldn't use `join()` alongside it. To make the url easier to read, I put all of the ternaries in variables so the url consists of the baseUrl and queryParam variables.

src/api/Feed.js
JChan106: Did not know this was a thing, interesting, will do

JChan106:
UAA File_activities endpoint is now fixed and I was able to fix some of the Flow type issues.

Those Flow type issues remaining have comments that explain why they are not solved.

- For the taskItem `created_by` Flow issue, I created a tech debt Jira for it to be handled later, the description explains the reasoning: https://jira.inside-box.net/browse/COXP-12037
- For the taskItem `assigned_to` Flow issue, Flow is expecting a string literal because of the TaskNew type, however because we are transforming the field with `toUpperCase()` it is a string. Flow counts this as a mismatch, unsure if this is needed to be fixed, and if so, how?
- For the Sort function Flow issue, the upcoming Jira should not require the use of the sorter since we are removing the separate version call: https://jira.inside-box.net/browse/COXP-11588. And File Activities is already sorted by `created_at` out the box.

src/api/Feed.js
```diff
@@ -95,6 +103,83 @@ const getItemWithPendingReply = <T: { replies?: Array<Comment> }>(item: T, reply
...
 
+const parseFileActivitiesResponseForFeed = (response?: { entries: FileActivity[] }) => {
+    if (!response || !response.entries || !response.entries.length) {
+        return { entries: [] };
+    }
+
+    const data = response.entries;
+    const parsedData = [];
+
+    data.map(item => {
```
jstoffan, line 114: Can we use `const parsedData = data.map(item => {` to get value out of the map call?

src/api/Feed.js
```diff
@@ -95,6 +103,83 @@ const getItemWithPendingReply = <T: { replies?: Array<Comment> }>(item: T, reply
...
+
+                commentItem.tagged_message = commentItem.tagged_message || commentItem.message || '';
+                break;
+            }
+            default: {
+                break;
+            }
+        }
+
+        parsedData.push(...Object.values(source));
```
jstoffan, line 175: Can we `return source` here or is the Object.values call needed?

jstoffan (commented):
Implementation looks good, just needs tests for the new modules.

src/api/Feed.js
JChan106: Can't easily return source here since the key name depends on the type.

JChan106: Actually, I think we can if we make a case for every type. Which might be better since it's more explicit.

src/api/Feed.js
```diff
@@ -394,38 +486,78 @@ class Feed extends Base {
...
-        const annotationsPromise = shouldShowAnnotations
-            ? this.fetchAnnotations(permissions, shouldShowReplies)
-            : Promise.resolve();
+
+        // Using the UAA File Activities endpoint replaces the need for these calls
+        const annotationsPromise =
+            !shouldUseUAA && shouldShowAnnotations
+                ? this.fetchAnnotations(permissions, shouldShowReplies)
+                : Promise.resolve();
+        const commentsPromise = () => {
```
tjuanitas, line 495: why is the `commentsPromise` declared differently than the others? could it have the same syntax as the others and then doesn't need to be invoked on line ~549?

src/api/Feed.js
```diff
@@ -525,6 +657,25 @@ class Feed extends Base {
         });
     }
 
+    /**
+     * Fetches the file activities for a file
+     *
+     * @param {Object} permissions - the file permissions
```
tjuanitas, line 663: this I think?
```suggestion
     * @param {BoxItemPermission} permissions - the file permissions
     * @param {Array<FileActivityTypes>} activityTypes - [...]
```

src/constants.js
```diff
@@ -470,6 +470,16 @@ export const ACTIVITY_FILTER_OPTION_RESOLVED: 'resolved' = 'resolved';
...
 
+/* ------------------ File Activity Types ------------------ */
+export const FILE_ACTIVITY_TYPE_ANNOTATION: 'annotation' = 'annotation';
+export const FILE_ACTIVITY_TYPE_APP_ACTIVITY: 'app_activity' = 'app_activity';
+export const FILE_ACTIVITY_TYPE_COMMENT: 'comment' = 'comment';
+export const FILE_ACTIVITY_TYPE_TASK: 'task' = 'task';
+export const FILE_ACTIVITY_TYPE_VERSION: 'versions' = 'versions';
+
+/* ------------------ File Type ------------------ */
+export const FILE_TYPE_ACTIVITY: 'activity' = 'activity';
```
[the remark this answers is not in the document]
JChan106: Currently it can only be of type `activity` as far as I know. This is coming from the backend. Are you suggesting to name the typing something other than `FileActivityType`?

src/api/FileActivities.js
```diff
@@ -0,0 +1,106 @@
...
+            errorCallback(e, this.errorCode);
+            return;
+        }
+
+        this.get({
+            id: fileId,
+            successCallback,
+            errorCallback,
+            requestData: {
+                ...(repliesCount ? { replies_count: repliesCount } : null),
```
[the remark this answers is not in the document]
JChan106: I believe i've used `&&` before but a little hesitant to use as its not as explicit and i've had unexpected results in the past. In this case it would probably work, but I think if theres no improvement, Id rather use ternary.

src/api/Feed.js
```diff
@@ -95,6 +103,83 @@ const getItemWithPendingReply = <T: { replies?: Array<Comment> }>(item: T, reply
     return { replies: [...replies, reply], ...rest };
 };
 
+const parseFileActivitiesResponseForFeed = (response?: { entries: FileActivity[] }) => {
```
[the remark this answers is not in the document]
JChan106: hm good point, I think `parseFileActivitiesResponse` could work in this case, will change

src/api/Feed.js
```diff
@@ -95,6 +103,83 @@ const getItemWithPendingReply = <T: { replies?: Array<Comment> }>(item: T, reply
...
+
+                        assignedToEntry.role = entry.role.toUpperCase();
+                        assignedToEntry.status = entry.status.toUpperCase();
+
+                        return assignedToEntry;
+                    });
+                    // $FlowFixMe Converting to uppercase makes role and status a string, which is incompatible with string literal
+                    taskItem.assigned_to.entries = assignedToEntries;
+                }
+                if (taskItem.completion_rule && isString(taskItem.completion_rule)) {
```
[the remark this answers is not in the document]
JChan106: Hm I think I was erroring on the side of caution before, I found the typing to be only string so I don't think we need it, will remove

src/api/Feed.js
JChan106: It didn't have a `Promise.resolve()` condition before. It would either fetch ThreadedComments or Comments. In this case, if we have the FF on, we don't want either of the Comment fetches.

src/api/Feed.js
```diff
@@ -95,6 +103,83 @@ const getItemWithPendingReply = <T: { replies?: Array<Comment> }>(item: T, reply
     return { replies: [...replies, reply], ...rest };
 };
 
+const parseFileActivitiesResponseForFeed = (response?: { entries: FileActivity[] }) => {
+    if (!response || !response.entries || !response.entries.length) {
```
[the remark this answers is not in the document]
JChan106: Actually that threw a flow error for some reason. It was saying that response.entries could be undefined in the next line. Will keep as is, unless we have a known work around

src/constants.js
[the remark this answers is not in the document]
tjuanitas: it was specifically this name that confused me: `FILE_TYPE_ACTIVITY` under the `--- File Type ---` section

I would've assumed files types would be something like:
```
export const FILE_TYPE_IMAGE: 'image' = 'image';
export const FILE_TYPE_VIDEO: 'video' = 'video';
// or
export const FILE_TYPE_JPEG: 'jpeg' = 'jpeg';
```

but if this is how it aligns in the backend and others are familiar with it, we can leave this

src/api/Feed.js
tjuanitas: oh I see now

tjuanitas: new thought: do these calls: `commentsPromise`, `annotationsPromise`, `tasksPromise`, and `appActivityPromise` need their declarations changed from the original implementation?

from how I understand, these are only used once in the `Promise.all` (~line 559) which is located in the branch when `if (shouldUseUAA == false)` which means these individual functions won't need to check `!shouldUseUAA` right?

JChan106: I think without `shouldUseUAA` in each of the declarations, the calls would be made still, its just that the promises wouldn't be handled. Having `shouldUseUAA` in each prevents making those extra calls. Monoproxy is broken currently so its hard to check, but I think last I checked, without changing the declarations, the calls were being made.

src/api/FileActivities.js
```diff
@@ -0,0 +1,106 @@
...
+     * API URL for filtered file activities
+     *
+     * @param {string} [id] - a box file id
+     * @param {Array<FileActivityTypes>} activityTypes - optional. Array of File Activity types to filter by, returns all Activity Types if omitted.
+     * @return {string} base url for files
+     */
+    getFilteredUrl(id?: string, activityTypes?: FileActivityTypes[]): string {
+        if (!id) {
+            throw new Error('Missing file id!');
+        }
```
tjuanitas, line 54: should we make `id` required for this function?

and maybe move the `if (!id)` guard to `getActivities` to prevent calls to `this.checkApiCallValidity`?

src/api/FileActivities.js
```diff
@@ -0,0 +1,106 @@
...
+import type { ElementsXhrError } from '../common/types/api';
+import type { FileActivity, FileActivityTypes } from '../common/types/feed';
+
+// We only show the latest reply in the UI
+const REPLY_LIMIT = 1;
+
+const getFileActivityQueryParams = (
+    fileID: string,
+    activityTypes?: FileActivityTypes[] = [],
+    enableReplies?: boolean = true,
```
tjuanitas, line 19: is the `enableReplies` parameter needed? all usages in this file are `true` and the function is not exported

src/api/FileActivities.js
```diff
@@ -0,0 +1,106 @@
...
+};
+
+class FileActivities extends Base {
+    /**
+     * API URL for file activities
+     *
+     * @param {string} [id] - a box file id
+     * @return {string} base url for files
+     */
+    getUrl(id?: string): string {
```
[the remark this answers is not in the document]
JChan106: Good point, thought we needed it because of the Base but I think we can remove this

src/constants.js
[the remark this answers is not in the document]
JChan106: Yeah looked at it again, and its pretty confusing. I removed the naming since it's only 'activity' for now. We can define a type in a different PR which uses it

src/api/Feed.js
```diff
@@ -95,6 +102,96 @@ const getItemWithPendingReply = <T: { replies?: Array<Comment> }>(item: T, reply
     return { replies: [...replies, reply], ...rest };
 };
 
+export const getParsedFileActivitiesResponse = (response?: { entries: FileActivity[] }) => {
+    if (!response || !response.entries || !response.entries.length) {
+        return { entries: [] };
+    }
+
+    const data = response.entries;
+    const parseReplies = (replies: Comment[]): Comment[] => {
```
jstoffan, line 111: Minor, but this helper can be moved outside of its parent function.

src/api/Feed.js
```diff
@@ -95,6 +102,96 @@ const getItemWithPendingReply = <T: { replies?: Array<Comment> }>(item: T, reply
...
+                    if (taskItem.assigned_to?.entries) {
+                        const assignedToEntries = taskItem.assigned_to.entries.map(entry => {
+                            const assignedToEntry = { ...entry };
+
+                            assignedToEntry.role = entry.role.toUpperCase();
+                            assignedToEntry.status = entry.status.toUpperCase();
+
+                            return assignedToEntry;
+                        });
+                        // $FlowFixMe Converting to uppercase makes role and status a string, which is incompatible with string literal
```
jstoffan, line 140: Not quite following this comment. Upper-casing shouldn't have an effect on the type.

src/api/FileActivities.js
```diff
@@ -0,0 +1,88 @@
...
+ */
+
+import Base from './Base';
+import { PERMISSION_CAN_COMMENT, PERMISSION_CAN_VIEW_ANNOTATIONS, ERROR_CODE_FETCH_ACTIVITY } from '../constants';
+import type { BoxItemPermission } from '../common/types/core';
+import type { ElementsXhrError } from '../common/types/api';
+import type { FileActivity, FileActivityTypes } from '../common/types/feed';
+
+// We only show the latest reply in the UI
+const REPLY_LIMIT = 1;
```
jstoffan, line 14: Minor, but this should be configurable via a parameter at some point in the future.

jstoffan (approved):

src/api/Feed.js
JChan106: Worded poorly on my part, but Flow assumes that `role` and `status` is a string because it's using the `toUpperCase` method. `role` and `status` are supposed to be string literals based on the typing used, so Flow gives an error

src/api/FileActivities.js
JChan106: I think we can configure this in the future should the value need to be dynamic. Currently we don't use any limit other than 1 so I think it's manageable for now.

jstoffan (approved):

JChan106:
Updated screenshots:

image

tjuanitas (approved):

PR #3319, opened 2023-05-02
fix(content-explorer): select all checkbox state matches selectedItems

pusch-e:


src/features/content-explorer/content-explorer/__tests__/ContentExplorer.test.js
```diff
@@ -453,6 +453,75 @@ describe('features/content-explorer/content-explorer/ContentExplorer', () => {
...
+        test('should set isSelectAllChecked to false if an item is unchecked while isSelectAllChecked is true', () => {
+            const clickedItemIndex = 0;
+            const clickedItemIndex2 = 1;
+            const items = [
+                { id: '1', name: 'item1' },
+                { id: '2', name: 'item2' },
+            ];
+            const selectedItems = { '1': items[clickedItemIndex], '2': items[clickedItemIndex2] };
+            const isSelectAllAllowed = true;
+            const isSelectAllChecked = true;
```
tjuanitas, line 505: nit thing: this might be more preference but I think single-use variables intended for objects don't provide additional value over using the literal value directly in the object.

something like
```
const nextValue = 123;
[...]
const object = { nextValue };
```

can sometimes be worse than
```
const object = { nextValue: 123 };
```

because now the developer needs to find where the variable was set

---

so when I was reading
```
wrapper.setState({ selectedItems, isSelectAllChecked });
```
my first impression was `isSelectAllChecked` was dynamic and I had to scroll up to see that the value was  fixed to `true`

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -181,17 +181,39 @@ class ContentExplorer extends Component {
...
         document.removeEventListener('click', this.handleDocumentClick, true);
     }
 
+    areAllCurrentItemsSelected = () => {
+        const { items } = this.props;
+        const { selectedItems } = this.state;
+
+        const someItemIsNotSelected = items.some(item => {
+            return !selectedItems[item.id];
+        });
```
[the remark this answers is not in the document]
pusch-e: Ah yes using .every is a better implementation. I originally used .every then changed to .some because I mistakenly thought that their time complexity would be different. I will change to use .every

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -181,17 +181,39 @@ class ContentExplorer extends Component {
...
+                this.setState({ isSelectAllChecked: areAllCurrentItemsSelected });
+            }
+        }
     }
 
     componentWillUnmount() {
         document.removeEventListener('click', this.handleDocumentClick, true);
     }
 
+    areAllCurrentItemsSelected = () => {
```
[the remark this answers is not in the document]
pusch-e: Yes this is true. I thought adding "current" in the name would make it extra clear, but I do not think it is necessary after reading your comment.

src/features/content-explorer/content-explorer/__tests__/ContentExplorer.test.js
pusch-e: That is a fair point. I will change to using the literal value in the object.

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -181,17 +181,39 @@ class ContentExplorer extends Component {
...
 
     componentDidUpdate({ initialFoldersPath: prevInitialFoldersPath }) {
-        const { initialFoldersPath } = this.props;
+        const { initialFoldersPath, isSelectAllAllowed } = this.props;
+        const { isSelectAllChecked } = this.state;
 
         if (prevInitialFoldersPath !== initialFoldersPath) {
             this.handleFoldersPathUpdated(initialFoldersPath);
         }
+        if (!this.isLoadingItems() && isSelectAllAllowed) {
```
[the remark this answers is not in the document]
pusch-e: I have tested this locally, and once the items do actually load, the checkbox will update again. So, the checkbox and the items remain in sync during this phase. I added this if statement because it is a better user experience for the checkbox to not update until the new items load. Without the isLoading check, the checkbox unchecks itself whenever items are loading and then updates once they are loaded in. So navigating between folders with everything selected would lead to a checked box going from checked to unchecked then checked again, which I thought looked worse.

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -181,17 +181,36 @@ class ContentExplorer extends Component {
...
+    areAllItemsSelected = () => {
+        const { items } = this.props;
+        const { selectedItems } = this.state;
+
+        return items.every(item => selectedItems[item.id]);
+    };
+
+    isLoadingItems = () => {
+        const { items } = this.props;
+        return items && items[0] && items[0].isLoading;
```
menaczar, line 211: items?.[0]?.isLoading

menaczar: isLoadingItems = () => this.props?.items?.[0]?.isLoading;

one liner

menaczar (approved):

tjuanitas (approved):

PR #3320, opened 2023-05-04
fix(pill-selector-dropdown): remove default value for isPositionDynamic

tjuanitas:
There's an issue in Admin Console where uses of `PillSelectorDropdown` are misaligned if they're used within small container. Setting `isPositionDynamic` to `false` by default so that components opt-in to using the Popper component.

**Before**
Screen Shot 2023-05-03 at 6 32 26 PM

**After**
Screen Shot 2023-05-03 at 6 32 34 PM

### DIFF
--- a/src/components/pill-selector-dropdown/PillSelectorDropdown.js
+++ b/src/components/pill-selector-dropdown/PillSelectorDropdown.js
@@ -38,6 +38,8 @@ type Props = {
     getPillImageUrl?: (data: { id: string, [key: string]: any }) => string | Promise<?string>,
     /** Passed in by `SelectorDropdown` for accessibility */
     inputProps: Object,
+    /** Option to enable dynamic positioning with popper */
+    isPositionDynamic?: boolean,
     /** Input label */
     label: React.Node,
     /** Called when pill selector input is blurred */
@@ -94,6 +96,7 @@ class PillSelectorDropdown extends React.Component<Props, State> {
         disabled: false,
         error: '',
         inputProps: {},
+        isPositionDynamic: false,
         label: '',
         onBlur: noop,
         onPillCreate: noop,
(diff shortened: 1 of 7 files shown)


src/components/pill-selector-dropdown/PillSelectorDropdown.js
```diff
@@ -38,6 +38,8 @@ type Props = {
     getPillImageUrl?: (data: { id: string, [key: string]: any }) => string | Promise<?string>,
     /** Passed in by `SelectorDropdown` for accessibility */
     inputProps: Object,
+    /** Option to enable dynamic positioning with popper */
+    isPositionDynamic?: boolean,
```
benjamin-shen, line 42: Should this be a required prop (like some of the other boolean props)? If not, should the default value be `undefined` instead?

tjuanitas: this came up from this PR: https://github.com/box/box-ui-elements/pull/3262 where `SelectorDropdown` defaults to false

we don't want to make it required but what's the advantage of making it `undefined`?

benjamin-shen: Ah I was thinking so we don't have `isPositionDynamic={false}` added to the snapshots and the prop could be added to the component with just `isPositionDynamic` instead of `isPositionDynamic={true}`. It's not a super strong suggestion, more of a question. WDYT?

tjuanitas: I think leaving it undefined makes sense since it's just a passthrough prop to `SelectorDropdown`. and then `SelectorDropdown` can handle the default value. that way we don't need to update in two places in the future.

benjamin-shen: Makes sense

tjuanitas: HALF A BOOLEAN??

benjamin-shen (approved):

pusch-e (approved):

PR #3321, opened 2023-05-04
feat(threaded-replies): reply to annotations

jcamden:
This is very much a draft. Names are not at all final (like BaseCommentSneaky). This is just a proof of concept for organization. Here is what has been done in this PR:

- I more-or-less hacked createAnnotationReply so that the URL has the required file_id param. I’m not sure, but maybe the most ideal solution here would be to make this endpoint conform to the other related endpoints by getting id from the body instead of doing this. But this is lowest LOE, for sure.
- I separated the ui and business logic of BaseComment into BaseComment and BaseCommentWrapper  respectively. This way, it will be easier to re-use some of BaseComment if/when comments are reborn in DSYS. Importantly, the wrapper will handle the differences between, e.g. onCommentEdit and onAnnotationsEdit , construct the many things that are needed (handleMessageUpdate, handleStatusUpdate , etc), and then forward those as props to the dumb component BaseComment.
- I broke Replies out into its own file. I handled the circular dependency issue that had led us to combine it with BaseComment.js by exposing children on BaseCommentWrapper, passing in Replies as a child, and then passing BaseCommentWrapper as a prop to Replies. This requires an additional wrapper around BaseCommentWrapper which has the temporary name BaseCommentSneaky.

Here is this code in action on dev vm:
May-08-2023 12-32-29

Here are the stories (so far) for BaseComment:
Screenshot 2023-05-04 at 5 16 00 PM
Screenshot 2023-05-04 at 5 16 03 PM

### DIFF
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
@@ -13,7 +13,7 @@ import Comment from '../comment';
 import TaskNew from '../task-new';
 import Version, { CollapsedVersion } from '../version';
 import withErrorHandling from '../../withErrorHandling';
-import { BaseComment } from '../comment/BaseComment';
+import BaseCommentWrapper from '../comment/BaseCommentWrapper';
 import {
     FEED_ITEM_TYPE_ANNOTATION,
     FEED_ITEM_TYPE_APP_ACTIVITY,
@@ -35,6 +35,9 @@ import type { SelectorItems, User } from '../../../../common/types/core';
 import type { GetAvatarUrlCallback, GetProfileUrlCallback } from '../../../common/flowTypes';
 import type { Translations } from '../../flowTypes';
 
+import { type OnAnnotationEdit, type OnCommentEdit } from '../comment/types';
+import AnnotationActivityLinkProvider from './AnnotationActivityLinkProvider';
+
 type Props = {
     activeFeedItem: FeedItem,
     activeFeedItemRef: { current: null | HTMLElement },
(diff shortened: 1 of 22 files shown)


src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -296,42 +299,73 @@ const ActiveState = ({
...
-                                        isCurrentVersion={currentFileVersionId === itemFileVersionId}
-                                        item={item}
+                                        getUserProfileUrl={getUserProfileUrl}
                                         mentionSelectorContacts={mentionSelectorContacts}
-                                        onEdit={onAnnotationEdit}
                                         onDelete={onAnnotationDelete}
-                                        onSelect={onAnnotationSelect}
-                                        onStatusChange={onAnnotationStatusChange}
+                                        // onEdit={onAnnotationEdit}
+                                        onEdit={onCommentEdit}
```
jcamden, line 315: Here's a todo.

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -11,186 +10,91 @@ import ActivityTimestamp from '../common/activity-timestamp';
...
+    isResolved: boolean,
 };
 
 // TODO: Replace and rename to Comment component once threaded replies refactor is fully implemented
 const BaseComment = (props: BaseCommentProps) => {
     const {
-        created_by,
-        created_at,
-        permissions = {},
+        // annotationActivityLink,
```
jcamden, line 59: Here's another important todo.

src/elements/content-sidebar/activity-feed/comment/BaseCommentSneaky.js
```diff
@@ -0,0 +1,60 @@
...
+                    isRepliesLoading={isRepliesLoading}
+                    onHideReplies={onHideReplies}
+                    onReplyCreate={onReplyCreate}
+                    onReplySelect={onSelect}
+                    onShowReplies={onShowReplies}
+                    replies={replies}
+                    repliesTotalCount={repliesTotalCount}
+                />
+            ) : (
+                undefined
```
jcamden, line 54: Only did this (as apposed to `replies && Replies`) to satisfy Flow. Lmk if you all know the trick :)

tjuanitas: this ended up working for me: `{hasReplies && replies && <Replies [...] />}`

one of the other `$FlowFixMe` might've fixed the issue you were seeing 🤷

JChan106 (commented):
Theres a lot going on in this PR, is it possible to split this up into multiple PRs to make it easier for reviewers?

PR #3318, opened 2023-04-28
feat(sharing): Change message for shared links for Canvas files

pjzabinski:

PR #3322, opened 2023-05-04
fix(usm): make item prop optional

tjuanitas:
the `item` prop was added for a Canvas use case in this PR: https://github.com/box/box-ui-elements/pull/3318

it was assumed that the component files in this features folder were only used by USM. however, there is a usage in the main webapp where only the `SharedLinkAccessLabel` component is imported and used in the sidebar when a folder is opened in the All Files page. When trying to bump BUIE in EUA, it is causing a type error because `item` is currently required.

This is setting `item` to optional instead of making a change in EUA to pass the `item` prop. The reasoning behind setting as optional instead of updating the component in EUA is so when Canvas is released, the changes in the above PR can be reverted without requiring another change in EUA.

### DIFF
--- a/src/features/unified-share-modal/SharedLinkAccessDescription.js
+++ b/src/features/unified-share-modal/SharedLinkAccessDescription.js
@@ -13,7 +13,7 @@ import messages from './messages';
 type Props = {
     accessLevel?: accessLevelType,
     enterpriseName?: string,
-    item: itemtype,
+    item?: itemtype,
     itemType: ItemType,
 };
 
@@ -23,7 +23,7 @@ const SharedLinkAccessDescription = ({ accessLevel, enterpriseName, item, itemTy
     switch (accessLevel) {
         case ANYONE_WITH_LINK:
             // TODO: temporary change to support Canvas not being truly public
-            if (isBoxCanvas(convertToBoxItem(item))) {
+            if (item && isBoxCanvas(convertToBoxItem(item))) {
                 description = messages.canvasPeopleWithLinkDescription;
             } else {
                 description = messages.peopleWithLinkDescription;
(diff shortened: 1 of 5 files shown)


QingyuChai (approved):
l g t m

src/features/unified-share-modal/SharedLinkAccessDescription.js
```diff
@@ -23,7 +23,7 @@ const SharedLinkAccessDescription = ({ accessLevel, enterpriseName, item, itemTy
     switch (accessLevel) {
         case ANYONE_WITH_LINK:
             // TODO: temporary change to support Canvas not being truly public
-            if (isBoxCanvas(convertToBoxItem(item))) {
+            if (item && isBoxCanvas(convertToBoxItem(item))) {
```
megansmith-box, line 26: Should we add a test case where item is undefined in src/features/unified-share-modal/__tests__/SharedLinkAccessDescription.test.js?

src/features/unified-share-modal/__tests__/SharedLinkAccessDescription.test.js
```diff
@@ -59,17 +59,16 @@ describe('features/unified-share-modal/SharedLinkAccessDescription', () => {
                 name: 'Box',
             },
         ].forEach(({ itemType, name }) => {
-            test('should render correct menu', () => {
-                const sharedLinkPermissionMenu = shallow(
```
tjuanitas, line 63: this must've been copied from another file but this is not a menu

src/features/unified-share-modal/SharedLinkAccessDescription.js
tjuanitas: I ended up reverting the tests from before to cover the `undefined` cases and then added a separate test to handle the new Canvas case

megansmith-box (approved):

PR #3324, opened 2023-05-09
fix(content-explorer): select all checkbox in empty folder

pusch-e:
Select All checkbox was automatically getting checked in an empty folder. This was because of logic with .every where an empty item list will return true still as there is no item unchecked. 
This new change makes it so that in an empty folder, the Select All checkbox is permanently unchecked, and it cannot be checked. This is because with no items, it is impossible to have any items checked.

Before:
Screen Shot 2023-05-09 at 10 48 10 AM

After:
Screen Shot 2023-05-09 at 10 45 44 AM


tjuanitas (approved):
LGTM

SkyisAakash (approved):

PR #3328, opened 2023-05-10
fix(content-explorer): increase grid button style specificity

tjuanitas:
If a developer is using the CDN links and multiple elements, then there can be conflicting styles based on the order of the stylesheets. e.g. order of `<link>` tags in https://codepen.io/box-platform/pen/rRZxYq

This is happening with the `ViewModeChangeButton` in ContentExplorer because it has the same specificity weight as the base BUIE button. The base button styles included in the other elements' stylesheets (like Sidebar, Preview, etc) are applied after the Explorer specific styles which means if they have the same specificity weight, then styles imported last take precedent.

**Before**
Screen Shot 2023-05-10 at 11 46 09 AM

**After**
Screen Shot 2023-05-10 at 11 46 33 AM

### DIFF
--- a/src/elements/common/sub-header/ViewModeChangeButton.scss
+++ b/src/elements/common/sub-header/ViewModeChangeButton.scss
@@ -1,4 +1,4 @@
-.bdl-ViewModeChangeButton {
+.btn.bdl-ViewModeChangeButton {
     padding: 5px 6px;
 
     svg {


src/elements/common/sub-header/ViewModeChangeButton.scss
```diff
@@ -1,4 +1,4 @@
-.bdl-ViewModeChangeButton {
+.btn.bdl-ViewModeChangeButton {
```
tjuanitas, line 1: same approach used for `AddButton` and `SortButton`: https://github.com/box/box-ui-elements/blob/master/src/elements/common/sub-header/AddButton.scss#L1

src/elements/common/sub-header/ViewModeChangeButton.scss
```diff
@@ -1,4 +1,4 @@
...
     padding: 5px 6px;
```
tjuanitas, line 2: conflicting with this: https://github.com/box/box-ui-elements/blob/e5aa7e5adc5e16fb91dad753f4c467dba75ec561/src/styles/common/_buttons.scss#L17

benjamin-shen (approved):

QingyuChai (approved):

bfoxx1906 (approved):
looks straight forward

PR #3329, opened 2023-05-10
fix(pill-selector-dropdown): pass showAvatars prop to RoundPill

tjuanitas:
The `showAvatars` prop from `PillSelector` was not being passed to the `RoundPill` component so the avatars were always being shown. The prop was added in this PR: https://github.com/box/box-ui-elements/pull/2171 and the description says the avatars can be hidden.

**Intended Implementation**
https://user-images.githubusercontent.com/1657938/83594159-5d823f00-a513-11ea-9b78-e80fee8bd0b4.png

**Before**
Screen Shot 2023-05-10 at 12 47 20 PM

**After**
Screen Shot 2023-05-10 at 12 46 48 PM

### DIFF
--- a/src/components/pill-selector-dropdown/PillSelector.js
+++ b/src/components/pill-selector-dropdown/PillSelector.js
@@ -242,7 +242,7 @@ class PillSelectorBase extends React.Component<Props, State> {
                                       onRemove={onRemove.bind(this, option, index)}
                                       // $FlowFixMe option.text is for backwards compatibility
                                       text={option.displayText || option.text}
-                                      showAvatar
+                                      showAvatar={showAvatars}
                                       id={option.id}
                                       hasWarning={option.hasWarning}
                                       isExternal={option.isExternalUser}
@@ -278,9 +278,7 @@ class PillSelectorBase extends React.Component<Props, State> {
                         {...rest}
                         {...inputProps}
                         autoComplete="off"
-                        className={classNames('bdl-PillSelector-input', 'pill-selector-input', className, {
-                            'bdl-PillSelector-input--showAvatars': showAvatars,
-                        })}
+                        className={classNames('bdl-PillSelector-input', 'pill-selector-input', className)}
                         disabled={disabled}
                         onInput={onInput}
                         placeholder={this.getNumSelected() === 0 ? placeholder : ''}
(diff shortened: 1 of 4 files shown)


src/components/pill-selector-dropdown/PillSelector.js
```diff
@@ -278,9 +278,7 @@ class PillSelectorBase extends React.Component<Props, State> {
                         {...rest}
                         {...inputProps}
                         autoComplete="off"
-                        className={classNames('bdl-PillSelector-input', 'pill-selector-input', className, {
-                            'bdl-PillSelector-input--showAvatars': showAvatars,
```
tjuanitas, line 282: not sure what this class was intended for but there's not reference of it in BUIE or GHE

src/features/unified-share-modal/ContactsField.js
```diff
@@ -209,6 +209,7 @@ class ContactsField extends React.Component<Props, State> {
                 placeholder={intl.formatMessage(commonMessages.pillSelectorPlaceholder)}
                 ref={fieldRef}
                 selectedOptions={selectedContacts}
+                showAvatars
```
tjuanitas, line 212: added so current behavior does not change since `showAvatar` was previously always set for `RoundPill` inside of `PillSelector`

benjamin-shen: Should we default `showAvatars` to true?

tjuanitas: I thought about that so it doesn't break current behavior but because the usages are so low, I was going to manually fix them in a bump.

one reason I was against it was because `RoundPill` (the component that does the avatar stuff) defaults to `false`. so having the a parent component default to `true` seemed odd: https://github.com/box/box-ui-elements/blob/master/src/components/pill-selector-dropdown/RoundPill.js#L41

the other reason:
```
// if we default to true, then components will need to do this
<PillSelectorDropdown
  showAvatars={false}
  showRoundPills
/>

----------------

// defaulting to undefined seems ideal because we can do:

// won't show avatars
<PillSelectorDropdown
  showRoundPills
/>

// will show avatars
<PillSelectorDropdown
  showAvatars
  showRoundPills
/>

// showAvatars and showRoundPills have the same usage which is good because they're used together related
```

benjamin-shen: All good points. Found this related post about boolean variable conventions: https://www.serendipidata.com/posts/naming-guidelines-for-boolean-variables

Maybe if we wanted to default to not showing avatars, we should have a prop `hideAvatars` instead that defaults to false. But since we already have `showAvatar` I'm good with sticking to `showAvatars` and defaulting to false.

benjamin-shen (approved):

TaoJiang98 (approved):

PR #3331, opened 2023-05-16
feat(threaded-replies): reply to annotations

jcamden:
- updated `createAnnotationReply` with necessary URL param for id
- modified `BaseComment` to represent annotations in addition to comments
- created `AnnotationActivityLinkProvider` for creating `AnnotationActivityLink` (to be passed to `BaseComment`)
- modified `onEdit` functions for both annotations and comments to take a single argument (an object) instead of several ordered arguments. (This makes it easier work with the differences between `onCommentEdit` and `onAnnotationEdit`.)
- created some basic stories for `BaseComment`

Note that this PR does NOT bring our activity feed ui up to spec (neither in terms of organization nor in terms of styles). This will be handled in one or several subsequent PRs.

Also note that may want to reduce the complexity of BaseComment since the CodeMetrics plugin currently rates it at 77, not including the 27 complexity Replies component that currently resides in the same file. A proposal for doing so can be found [here](https://cloud.box.com/s/ndcei1y0h409uqa87v7zv0yezq9qi143).
Screenshot 2023-05-16 at 1 30 41 PM

From Storybook:
Screenshot 2023-05-16 at 1 33 28 PM
Screenshot 2023-05-16 at 1 33 31 PM
(Please not that the `isDisabled` prop seems to have no effect.)
Screenshot 2023-05-16 at 1 33 39 PM
Screenshot 2023-05-16 at 1 33 46 PM
Screenshot 2023-05-16 at 1 33 50 PM
Screenshot 2023-05-16 at 1 33 43 PM
Screenshot 2023-05-16 at 1 33 53 PM

From dev vm:
Screenshot 2023-05-11 at 2 02 00 PM


src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -77,37 +74,37 @@ type BaseCommentProps = {
...
-        onSelect,
-        onReplyCreate,
-        onShowReplies,
-        onHideReplies,
-        replies = [],
-        repliesTotalCount = 0,
-        status,
-    } = props;
-
+export const BaseComment = ({
```
jcamden, line 77: I alphabetized the props. The real diff is only:
~~onEdit~~
onAnnotationEdit,
annotationActivityLink,
onCommentEdit,

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -157,23 +154,30 @@ const BaseComment = (props: BaseCommentProps) => {
...
+        } else if (onCommentEdit) {
+            onCommentEdit({ id, status: selectedStatus, hasMention: false, permissions });
+        }
     };
 
+    const onEdit = onCommentEdit ?? onAnnotationEdit;
+
     const canDelete = permissions.can_delete;
     const canEdit = onEdit !== noop && permissions.can_edit;
     const canResolve = onEdit !== noop && permissions.can_resolve;
```
patlm, line 183: I see this was already here but wondering if the noop check will cover the case of onCommentEdit and onAnnotationEdit both being null/undefined?

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -355,7 +370,7 @@ type RepliesProps = {
     translations?: Translations,
 };
 
-const Replies = ({
+export const Replies = ({
```
patlm, line 373: Is Replies used anywhere outside of BaseComment? If it isn't, I would advocate for not exporting it (if that wouldn't break anything).

I believe part of the argument for keeping it in this file aside from avoiding a major refractor is that a reply is sort of part of a comment component even from the way it arrives from the back-end. If they are going to be somewhat coupled for right now, I think it would make sense to only allow it to be accessed through the BaseComment.

If we need to export it only for testing, we could add a comment noting that.

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
patlm: Just got to where `onEdit={noop}` was deleted in Replies component and now neither of the edit props are passed in. I'm not familiar with how noop works though so maybe it will still work (because in your sceenshots/demo it seems to work). Just wanted to add this specific example to confirm.

JChan106 (commented):
Thanks for the quick changes @jcamden! There's a lot of changes in this PR, do you think it's possible to break this down into some smaller PRs to make it a bit easier to read for reviewers?

src/api/Annotations.js
```diff
@@ -309,7 +309,7 @@ export default class Annotations extends MarkerBasedApi {
             data: { data: { message } },
             errorCallback,
             successCallback,
-            url: this.getUrlWithRepliesForId(annotationId),
+            url: `${this.getUrlWithRepliesForId(annotationId)}?file_id=${fileId}`,
```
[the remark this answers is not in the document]
jcamden: Nah, same error:
```
"Query parameter 'file_id' is required"
```
Tried it with both
```
data: { data: { message }, file_id: fileId }
```
and
```
 data: { data: { message, file_id: fileId } }
```
just in case. We could look up where that endpoint is defined, but of course lowest LOE here is just to give it the query param that it's asking for.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -173,28 +190,15 @@ const ActiveState = ({
                                 ref={refValue}
                             >
                                 {hasNewThreadedReplies ? (
+                                    // TODO: legitimate, pre-existing typing issue that was previously undetected
+                                    // Conflict between BoxCommentPermissions and BoxTaskPermissions
+                                    // $FlowFixMe
```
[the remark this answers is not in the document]
jcamden: I think we'd probably need to differentiate which type of activity item we're dealing with using type guards. It'd be a good tech debt ticket, but probably not something to address as part of this task.

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
jcamden: They were both originally exported. I just moved the export inline so that you know its being exported as soon as you start reading the definition. You are right that Replies is being exported for testing.

About Replies living here, I definitely do not subscribe to the idea that the shape of data from the backend should dictate the composition of ui components. Typically, you'll want to build out leaves in their own files, and then gradually bring them together in increasingly abstract and trunkwise components. It's an awful idea to have needlessly massive components just for the purpose of trying to reflect data structures.

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
jcamden: Ah, I see what you're seeing here now, functionality wise. Yes, editing replies does not work. I think we would need to pass `onComment` edit through to `Replies` even for annotations. (Maybe another good reason to pass replies in as a child instead).

jcamden: Eventually, I found that noop is a default coming all the way down from ActivitySidebar, so I reverted this use noop. Tragically, passing onCommentEdit to BaseComment in Replies does not enable modifying replies. But on the bright side, I learned that modifying replies is out of scope here. Thanks for bringing this up.

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
jcamden: Just want to make sure I understand: is the idea that we want to add a comment to warn people off from using `Replies` in any component other than `BaseComment`? If so, it seems a little unnecessary to me. But I also don't mind adding it if you see value in that. Jlmk :)

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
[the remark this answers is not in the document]
jcamden: I created a ticket for this :)

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
bfoxx1906: Having the export won't hurt anything. I think it'll be fine to leave.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -296,42 +300,66 @@ const ActiveState = ({
...
-                                    onReplyDelete={onReplyDeleteHandler(item.id)}
-                                    onReplyEdit={onReplyUpdateHandler(item.id)}
-                                    onReplySelect={onCommentSelectHandler(item.id)}
-                                    onShowReplies={onShowRepliesHandler(item.id, item.type)}
-                                    repliesTotalCount={item.total_reply_count}
-                                    replies={item.replies}
-                                    translations={translations}
-                                >
-                                    <AnnotationActivity
+                                {hasNewThreadedReplies && onAnnotationSelect ? (
```
bfoxx1906, line 303: IF onAnnotationsSelect is null then we will default to the old behavior. Do we need to perform this null check here?

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -296,42 +300,66 @@ const ActiveState = ({
...
+                                    <BaseComment
+                                        {...commentAndAnnotationCommonProps}
+                                        annotationActivityLink={
+                                            <AnnotationActivityLinkProvider
+                                                item={item}
+                                                onSelect={onAnnotationSelect}
+                                                isCurrentVersion={currentFileVersionId === itemFileVersionId}
+                                            />
+                                        }
+                                        onAnnotationEdit={onAnnotationEdit}
```
bfoxx1906, line 316: Do we need to specify an onAnnotationEdit and an onCommentEdit?  Can we just have one onEdit function and assign the proper handler to it?

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -157,23 +154,32 @@ const BaseComment = (props: BaseCommentProps) => {
...
-    }: {
-        hasMention: boolean,
-        id: string,
-        text: string,
-    }): void => {
-        onEdit(messageID, text, undefined, hasMention, permissions);
+    const handleMessageUpdate = ({ id: messageID, text, hasMention }) => {
+        // Since we have to pass onCommentEdit through annotations (to Replies), onAnnotationEdit essentially overrides onCommentEdit
+        if (onAnnotationEdit) {
+            onAnnotationEdit({ id: messageID, text, permissions });
```
bfoxx1906, line 160: Annotations can also have mentions in them. Why does onCommentEdit have that flag passed but onAnnotationEdit doesn't. Might be a good reason but I just want to make sure we are aware that annotations have mentions too.

bfoxx1906 (commented):
left comments

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
jcamden: This check is here because `onSelect` is a required prop in `AnnotationActivityLinkProvider`.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
jcamden: Well, `onAnnotationEdit` and `onCommentEdit` edit take different args, so I'm resolving that within BaseComment.js. We could potentially do it in ActiveState (or in my previous PR I handled it in `BaseCommentWrapper`). Good eye though. The reason that both props are passed here was because I was trying out passing `onCommentEdit` down to `Replies` (within `BaseComment`) to see if that would enable modifying replies. It didn't. I thought I cleaned everything up, but `onCommentEdit` doesn't need to be defined here.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -296,42 +300,66 @@ const ActiveState = ({
...
+                                        isPending={item.isPending}
+                                        isRepliesLoading={item.isRepliesLoading}
                                         mentionSelectorContacts={mentionSelectorContacts}
-                                        onEdit={onAnnotationEdit}
-                                        onDelete={onAnnotationDelete}
-                                        onSelect={onAnnotationSelect}
-                                        onStatusChange={onAnnotationStatusChange}
-                                    />
-                                </ActivityThread>
+                                        onHideReplies={onHideRepliesHandler(item.id)}
```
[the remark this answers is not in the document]
jcamden: The Github diff is just confused because the `ActivityThread` props are similar to the `AnnotationActivity` props.

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
jcamden: I don't know. Aside from converting the multiple params into a single param with props, I didn't change the `onAnnotationEdit` API. The props correspond to the pre-existing params. I do think we should look into if/how annotations previously handled mentions, probably as part of the tech spike.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
jcamden: I removed `onCommentEdit` here.

src/elements/content-sidebar/activity-feed/activity-feed/ActivityThreadReplies.js
```diff
@@ -61,7 +61,7 @@ const ActivityThreadReplies = ({
                     getUserProfileUrl={getUserProfileUrl}
                     mentionSelectorContacts={mentionSelectorContacts}
                     onDelete={onDelete}
-                    onEdit={onEdit}
+                    onCommentEdit={onEdit}
```
JChan106, line 64: Why is this change needed?

src/elements/content-sidebar/activity-feed/activity-feed/AnnotationActivityLinkProvider.js
```diff
@@ -0,0 +1,43 @@
...
+type AnnotationActivityLinkProviderProps = {
+    isCurrentVersion: boolean,
+    item: Annotation,
+    onSelect: (annotation: Annotation) => void,
+};
+
+const AnnotationActivityLinkProvider = ({ isCurrentVersion, item, onSelect }: AnnotationActivityLinkProviderProps) => {
+    const { file_version, id, target } = item;
+
+    const isFileVersionUnavailable = file_version === null;
```
JChan106, line 18: How about if file_version is undefined or 0? Should we just use a negation instead?

src/elements/content-sidebar/activity-feed/activity-feed/AnnotationActivityLinkProvider.js
```diff
@@ -0,0 +1,43 @@
...
+
+    const linkMessage = isCurrentVersion
+        ? annotationsMessages.annotationActivityPageItem
+        : annotationsMessages.annotationActivityVersionLink;
+    const linkValue = isCurrentVersion ? target?.location.value : file_version?.version_number;
```
JChan106, line 23: does location need an optional chain?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityThreadReplies.js
jcamden: Because `BaseComment` no longer has an onEdit prop. It takes `onCommentEdit` and `onAnnotationEdit` and resolves the differences for the handlers, etc. which make use of `onEdit`.

src/elements/content-sidebar/activity-feed/activity-feed/AnnotationActivityLinkProvider.js
jcamden: Here (and below) I just copied the logic that was already being used for creating AnnotationActivityLink. That said, it could be worth reviewing. I'll take a look at this :)

jcamden: It's because the type for `file_version` on the `Annotation` type is `file_version: BoxItemVersionMini | null`.

jcamden: (It can't be undefined or 0.)

src/elements/content-sidebar/activity-feed/activity-feed/AnnotationActivityLinkProvider.js
jcamden: No, it doesn't. Target is a union of type:
```
export type Target = TargetRegion | TargetPoint;
```
Both of those types require location:
```
export type TargetRegion = {
    location: Page,
    shape?: Rect,
    type: 'region',
};

export type TargetPoint = {
    location: Page,
    type: 'point',
    x: number,
    y: number,
};
```

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -346,6 +365,7 @@ type RepliesProps = {
     isParentPending?: boolean,
     isRepliesLoading?: boolean,
     mentionSelectorContacts?: SelectorItems<>,
+    onEdit: Function,
```
JChan106, line 368: Can we type this in arrow notation to be more explicit with parameters and return type?

src/elements/content-sidebar/activity-feed/comment/types.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+import type { AnnotationPermission, BoxCommentPermission, FeedItemStatus } from '../../../../common/types/feed';
+
+export type OnAnnotationEdit = (args: { id: string, permissions: AnnotationPermission, text?: string }) => void;
+
+export type OnCommentEdit = (args: {
+    hasMention: boolean,
+    id: string,
+    onError?: Function,
```
JChan106, line 9: Same here

src/elements/content-sidebar/activity-feed/comment/types.js
```diff
@@ -0,0 +1,14 @@
...
+    onSuccess?: Function,
```
JChan106, line 10: Same here

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
jcamden: Yes, good call. This should only ever be `OnCommentEdit`.

src/elements/content-sidebar/activity-feed/comment/types.js
jcamden: That'd be better. Let me see if I can track these down.

src/api/Annotations.js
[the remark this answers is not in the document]
tjuanitas: since you received the error `"Query parameter 'file_id' is required"`, does that means this POST previously wasn't working?

src/elements/content-sidebar/activity-feed/comment/types.js
jcamden: Yeah, I think I looked at this initially, and unforch, it's pretty hard to nail down what was intended. `onCommentUpdate` is passed to `onCommentEdit` in `ActivityFeed`.
```
onCommentUpdate?: (
        id: string,
        text?: string,
        status?: FeedItemStatus,
        hasMention: boolean,
        permissions: BoxCommentPermission,
        onSuccess: ?Function,
        onError: ?Function,
    ) => void,
```
`ActivitySidebar's` `this.updateComment` is passed to `onCommentUpdate`, and the type there is:
```
updateComment = (
        id: string,
        text?: string,
        status?: FeedItemStatus,
        hasMention: boolean,
        permissions: BoxCommentPermission,
        onSuccess: ?Function,
        onError: ?Function,
    ): void => {
 ```
I haven't been able to find the onError and onSuccess params getting used, so I'm just really not sure if there were some more specific intentions there.

File: src/elements/content-sidebar/activity-feed/comment/types.js (hunk 15 above)
Comment (jcamden) [reply src/elements/content-sidebar/activity-feed/comment/types.js:10]: See above.

File: src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js [hunk 16]
```diff
@@ -296,42 +300,65 @@ const ActiveState = ({
...
-                                    onReplyEdit={onReplyUpdateHandler(item.id)}
-                                    onReplySelect={onCommentSelectHandler(item.id)}
-                                    onShowReplies={onShowRepliesHandler(item.id, item.type)}
-                                    repliesTotalCount={item.total_reply_count}
-                                    replies={item.replies}
-                                    translations={translations}
-                                >
-                                    <AnnotationActivity
+                                {hasNewThreadedReplies && onAnnotationSelect ? (
+                                    // TODO: legitimate, pre-existing typing issue that was previously undetected
```
bfoxx1906, line 304: So remember hasNewThreadedReplies will go away as a check as will ActivityThread. In this case are we saying that we don't even want to render BaseComment if onAnnotationsSelect is not passed in b/c right now it's an optional in this class. Should we default onSelect to a noop in the AnnotationActivityLinkProvider if nothing is passed in for that prop? If we don't do this then we probably need to make onAnnotationsSelect required otherwise annotations could potentially never be rendered. Lemme know your thoughts.

jcamden: Yeah, that's a good point. This is part of a larger problem with `ActiveState` (the same problem that is responsible for the `$FlowFixMe`s I had to add to this file): we don't know which type of activity item is being passed in. I think what we need to do is create a union for `ActivityFeedProps` (currently, just called `Props`, fyi). Maybe not exactly like this, but here's the idea:
```
type ActivityFeedAnnotationProps = ActivityFeedCommonProps & {
    items: FeedItemsWithAnnotations,
    onAnnotationEdit: OnAnnotationEdit,
    onAnnotationSelect: OnAnnotationSelect,
    onAnnotationStatusChang: OnAnnotationsSelect,
};

type ActivityFeedWithoutAnnotationProps = ActivityFeedCommonProps & {
    items: FeedItemsWithoutAnnotations,
};

type ActivityFeedProps = ActivityFeedAnnotationProps | ActivityFeedWithoutAnnotationProps;
```
This way, if it's possible for `item.type === FEED_ITEM_TYPE_ANNOTATION`, `onAnnotationSelect` (et al) also have to be defined, so we don't have to perform this check (and the types would prevent someone from passing in an annotation item without the annotation functions).

I'm not 100% sure how this would work with Flow. It might require some typeguards (more than the switch) to disambiguate which of the activity feed item types we have. Anyway, we could try it as part of this PR or as a follow up. Jlmk!

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -200,7 +206,17 @@ const BaseComment = (props: BaseCommentProps) => {
...
-                    <Avatar getAvatarUrl={getAvatarUrl} user={createdByUser} />
+                    <Avatar
+                        getAvatarUrl={getAvatarUrl}
+                        user={createdByUser}
+                        badgeIcon={
+                            annotationActivityLink ? (
+                                <IconAnnotation title={<FormattedMessage {...messages.annotationBadge} />} />
+                            ) : (
+                                undefined
+                            )
```
tjuanitas, line 217: does `&&` work?

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -284,6 +300,7 @@ const BaseComment = (props: BaseCommentProps) => {
                     </div>
                     <div className="bcs-Comment-timestamp">
                         <ActivityTimestamp date={createdAtTimestamp} />
+                        {annotationActivityLink && annotationActivityLink}
```
tjuanitas, line 303: what's this doing?

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -325,6 +342,8 @@ const BaseComment = (props: BaseCommentProps) => {
                     {...commentProps}
                     isParentPending={isPending}
                     isRepliesLoading={isRepliesLoading}
+                    // note that passing down onCommentEdit does not allow editing replies
+                    onEdit={noop}
```
tjuanitas, line 346: sorry unfamiliar here: are replies not editable?

why do we pass a prop to `Replies` if it's just a `noop` instead of passing the `noop` to `BaseComment` directly in the `Replies` implementation?

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -157,23 +154,32 @@ const BaseComment = (props: BaseCommentProps) => {
...
+        }
         commentFormSubmitHandler();
     };
 
     const handleStatusUpdate = (selectedStatus: FeedItemStatus): void => {
-        onEdit(id, undefined, selectedStatus, false, permissions);
+        if (onAnnotationEdit) {
+            onAnnotationEdit({ id, permissions });
+        } else if (onCommentEdit) {
+            onCommentEdit({ id, status: selectedStatus, hasMention: false, permissions });
```
bfoxx1906, line 176: We are passing an object here instead of parameters. This will cause the onCommentEdit method not to work properly. This ultimately calls this https://github.com/box/box-ui-elements/blob/master/src/api/Feed.js#L1907

bfoxx1906 (changes requested):

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
bfoxx1906: Also for the status update for annotations we will need to pass the actual status too right? Let's go back and verify that all the pieces are here for this to work.

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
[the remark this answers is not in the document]
jcamden: I personally loathe default exports, but I do when required. In this case, the named export pre-existed, so I'm just forwarding what was already there.

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
jcamden: Oops, yep, I think you're right about the issue with onCommentEdit there. If we don't want to change it to an object all the way up the chain (although I'm tempted), I need to spread the args in ActivityFeed, like this:
```
onCommentEdit={hasCommentPermission ? props => onCommentUpdate(...props) : noop}
```

jcamden: About annotation status, `onAnnotationEdit` doesn't take `status`. We'd need to look into how annotations handle `status`---but I wasn't sure that was within the scope of this PR.

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
jcamden: Yes, it does, actually. Thanks :)

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
jcamden: Yyyep, that's a good point.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
jcamden: We decided to break this out into a separate PR, since it's probably going to be somewhat nontrivial.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -296,42 +300,65 @@ const ActiveState = ({
                                 isFocused={isFocused}
```
JChan106, line 300: Adding 

`isHoverable
  hasNewThreadedReplies={hasNewThreadedReplies}
`
Should fix the spacing issues you've been seeing. Should also put the sidebar in a good state for other devs to look at.

jcamden: Yep, totally works. Thanks for that. This will definitely set a better stage for the upcoming UI tasks :)

JChan106 (approved):
LGTM

JChan106:
Another thing to make sure is that everything looks good with newThreadedReplies prop disabled.

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
```diff
@@ -397,7 +399,21 @@ class ActivityFeed extends React.Component<Props, State> {
                             onAnnotationStatusChange={onAnnotationStatusChange}
                             onAppActivityDelete={onAppActivityDelete}
                             onCommentDelete={hasCommentPermission ? onCommentDelete : noop}
-                            onCommentEdit={hasCommentPermission ? onCommentUpdate : noop}
+                            onCommentEdit={
+                                hasCommentPermission && onCommentUpdate
```
bfoxx1906, line 403: This is doing a lot. Can we treat this like the other props e.g. onCommentDelete, onAnnotationEdit and just create a new object argument type for onCommentEdit to ensure the right params are passed in.

Perhaps something like this added to types.js:
```

export type DiscussionItemType = {
    hasMention?: boolean,
    id: string,
    onError?: Function,
    onSuccess?: Function,
    permissions: BoxCommentPermission | AnnotationPermission,
    status?: FeedItemStatus,
    text?: string,
};
export type OnAnnotationEdit = DiscussionItemType => void;

export type OnCommentEdit = DiscussionItemType => void;

```

bfoxx1906 (commented):
Left a comment

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
jcamden: Hmm, yeah, I don't like this part of the PR either, but I felt like replacing the args with a single arg all the way up was too much of a change, especially because it would result in a mismatch between the comment-related function and other functions at the respective levels... if that makes sense. That said, I'm happy to do it. But I want to make sure we're on the same page: 

I have already created an `OnCommentEdit` type[ in types.js](https://github.com/box/box-ui-elements/pull/3331/files#diff-8006f38d16ee725c1654fe66120517d595055479ef90050057a83fb337c1dafa), although it is important to note that it _cannot_ be the same type as `OnAnnotationEdit`. So, in the area that we are dealing with, `onCommentEdit` does take a single arg with props. But I had to pick a place to transform it back to args (which is what I'm doing here) unless we wanted to change the comment-related function to take an arg with props everywhere (and then it would be kind of out of place alongside the other functions in those files that still take args). 

I'm super tired. Kid kept me up. So I hope that makes sense. Just lmk which way you'd like me to go :)

jcamden:
> Another thing to make sure is that everything looks good with newThreadedReplies prop disabled.

@JChan106 , I'm not sure I understand. Do you mean with the `enterprise preview threaded replies q1fy24` FF disabled?

JChan106:
> > Another thing to make sure is that everything looks good with newThreadedReplies prop disabled.
> 
> @JChan106 , I'm not sure I understand. Do you mean with the `enterprise preview threaded replies q1fy24` FF disabled?

Yeah, just checking that everything works as before when the FF is off.

bfoxx1906 (approved):

PR #3334, opened 2023-05-24
fix(usm): fix message for upgrade notice in share modal

tjuanitas:
Adding a space before the "Upgrade now" button

**Before**
Screen Shot 2023-05-24 at 4 17 04 PM

**After**
Screen Shot 2023-05-24 at 4 16 26 PM


TaoJiang98 (approved):
LGTM!

PR #3335, opened 2023-05-25
fix: align viewer permissions for shared links

tjuanitas:
There is misalignment in behavior between the main webapp and UI Elements for creating shared links as a Viewer role. In the main webapp, a Viewer can create a shared link on an item but cannot edit it or remove it. In Content Explorer and Content Picker, a Viewer cannot generate a shared link.
expected permissions: https://support.box.com/hc/en-us/articles/360044196413-Understanding-Collaborator-Permission-Levels

This is because there are instances in Content Explorer and Content Picker where both `can_share` and `can_set_share_access` item permissions are checked for fetching and creating shared links. But this should not be done as only `can_share` is needed to create shared links while `can_set_share_access` is used for setting the access level of the shared link, (access levels like "open to all visitors" or "only those in the enterprise", etc.)

If the `access` property is not set while fetching the shared link, the backend defaults the access level to the enterprise default setting. This allows the Viewer role to create a shared link and so we do not need to check for `can_set_share_access`. If `access` is something other than `undefined` (and I think technically `null`), then that's when we need to check for the `can_set_share_access` permission.


src/api/Item.js
```diff
@@ -286,29 +286,35 @@ class Item extends Base {
...
      *
      * @param {string|void} itemID - ID of item to share
      * @param {BoxItemPermission|void} itemPermissions - Permissions for item
-     * @param {boolean|void} canSkipSetShareAccess - skip the check for can_set_share_access when creating a new shared link
+     * @param {boolean|void} canSkipSetShareAccessPermission - skip `can_set_share_access` permission check
      * @throws {Error}
      * @return {void}
      */
-    validateRequest(itemID: ?string, itemPermissions: ?BoxItemPermission, canSkipSetShareAccess: boolean = false) {
+    validateSharedLinkRequest(
```
tjuanitas, line 297: this function is only intended to be used for sharing requests (`share()` and `updateSharedLink()`) not any type of request

src/api/Item.js
```diff
@@ -329,14 +335,14 @@ class Item extends Base {
...
+            const { id, permissions, shared_link: sharedLink }: BoxItem = item;
             this.id = id;
             this.successCallback = successCallback;
             this.errorCallback = errorCallback;
 
-            // if we use the default access level, we don't need permission to set the access level
-            const canSkipSetShareAccess = access === undefined;
-            this.validateRequest(id, permissions, canSkipSetShareAccess);
+            // skip permission check when creating links with default access level
+            const canSkipSetShareAccessPermission = !sharedLink && access === undefined;
```
tjuanitas, line 344: we only want to skip the permission check if the shared link does not already exist. i.e. the scenario in which a "viewer" role is trying to create a shared link

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -1275,33 +1275,23 @@ class ContentExplorer extends Component<Props, State> {
...
-        if (!newItem[FIELD_SHARED_LINK]) {
-            const { canSetShareAccess }: Props = this.props;
-            if (!newItem || !canSetShareAccess) {
-                return;
-            }
-
-            const { permissions, type } = newItem;
-            if (!permissions || !type) {
-                return;
-            }
```
tjuanitas, line 1290: couple things here:

1. we don't want to `return` here because the collection should still be updated
2. the `if (!newItem)`check is unnecessary since we already check `!newItem[FIELD_SHARED_LINK]` in the outer block
3. the `canSetShareAccess` prop should not affect whether a user can generate a share link, only whether the user can update the access level
    -  generating a share link is checked with the `canShare` prop which is handled in `shareCallback()`

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -1275,33 +1275,23 @@ class ContentExplorer extends Component<Props, State> {
...
             });
         } else {
             // update collection with existing shared link
-            this.updateCollection(currentCollection, newItem, () => this.setState({ isShareModalOpen: true }));
+            this.updateCollection(currentCollection, item, () => this.setState({ isShareModalOpen: true }));
         }
     };
 
     /**
-     * Chages the sort by and sort direction
```
tjuanitas, line 1304: intended for different method: `sort()`

src/utils/fields.js
```diff
@@ -174,7 +174,7 @@ const FILE_VERSIONS_FIELDS_TO_FETCH = [
     FIELD_VERSION_NUMBER,
 ];
 
-// Fields needed for Content Picker to show shared link permissions
```
tjuanitas, line 177: used for both Explorer and Picker

src/elements/content-picker/ContentPicker.js
```diff
@@ -940,7 +938,7 @@ class ContentPicker extends Component<Props, State> {
         // if no shared link currently exists, create a shared link with enterprise default
         if (!item[FIELD_SHARED_LINK]) {
             // $FlowFixMe
-            this.changeShareAccess(null, item);
+            this.changeShareAccess(undefined, item);
```
tjuanitas, line 941: using `undefined` over `null` because of this comment: https://github.com/box/box-ui-elements/blob/5909f1abe2ec3701531fcbcb780c8a3cce9c7545/src/api/Item.js#L328

HOWEVER, I think both actually work so not sure if I should change everything to `null` so it's more intentional than `undefined`

src/elements/content-picker/ContentPicker.js
```diff
@@ -974,7 +972,7 @@ class ContentPicker extends Component<Props, State> {
         }
 
         const { can_set_share_access }: BoxItemPermission = permissions;
-        if (!can_set_share_access) {
+        if (access !== undefined && !can_set_share_access) {
```
tjuanitas, line 975: the `can_set_share_access` permission is only checked if there is a specified access different than the enterprise default

src/elements/content-explorer/ContentExplorer.js
benjamin-shen: For 3 I think the naming can be more precise, something like `canGenerateSharedLink` rather than `canShare`, and `canUpdateSharedLinkAccess` or even `canUpdateShareAccess` rather than `canSetShareAccess`. And also `shareCallback` probably has a better name too

src/elements/content-picker/ContentPicker.js
benjamin-shen: I think undefined actually makes sense here since we're making it fall back to the default. I can see `access = null` being ambiguous

benjamin-shen: Should `access` in Item.js be of type `Access` as defined [here](https://github.com/box/box-ui-elements/blob/5909f1abe2ec3701531fcbcb780c8a3cce9c7545/src/common/types/core.js#L78)?

src/elements/content-explorer/ContentExplorer.js
tjuanitas: depending on the names you're referring to, it's probably not possible to change them:

- `canShare` is a prop that controls rendering the "Share" button
- `canSetShareAccess` is a prop that controls rendering a `select` element that allows the user to select an access level: [reference](https://user-images.githubusercontent.com/1075325/27887156-0940ee3e-6194-11e7-8e22-961139e82dfe.png)
- `can_share` (as a variable: `canShare`) and `can_set_share_access` (as a variable: `canSetShareAccess`) refer to item permissions from the backend

for the props, we can't change these names without a breaking change since developers pass these into the elements. and it's not worth changing the backend permissions

the `shareCallback` follows the pattern of the other callbacks: `renameCallback`, `deleteCallback`, `previewCallback`. not great names but consistent 🤷 

we can change the other stand alone variables though (like `canSkipSetShareAccessPermission`) but were there any instances in particular you were thinking about?

patlm (approved):
From Preview perspective, I didn't see any concerns!

benjamin-shen (approved):

PR #3340, opened 2023-05-26
fix(share-form): map called on undefined object

jpan-box:
BEFORE
A rejected promise occurs when `options.map` is called and `options` is undefined. The error does NOT result in a jest failure in node 14, but DOES result in a failure in node 18
```bash
nvm use 18.15
BABEL_ENV=test NODE_ENV=test yarn jest -c scripts/jest/jest.config.js src/features/unified-share-modal/__tests__/UnifiedShareForm.test.js -t "justification is allowed"

...

 RUNS  src/features/unified-share-modal/__tests__/UnifiedShareForm.test.js
/.../box-ui-elements/src/features/unified-share-modal/UnifiedShareForm.js:119
          justificationReasons: options.map(function (_ref2) {
                                        ^

TypeError: Cannot read properties of undefined (reading 'map')
    at map (/.../box-ui-elements/src/features/unified-share-modal/UnifiedShareForm.js:120:51)

Node.js v18.15.0
error Command failed with exit code 1.

```
before2
before1

CHANGE
Added a condition to default to empty array if options is undefined
Added a test to ensure `justificationReasons` is an empty array

AFTER
after1

### DIFF
--- a/src/features/unified-share-modal/UnifiedShareForm.js
+++ b/src/features/unified-share-modal/UnifiedShareForm.js
@@ -114,7 +114,7 @@ class UnifiedShareForm extends React.Component<USFProps, State> {
         this.setState({ isFetchingJustificationReasons: true });
 
         return getJustificationReasons(item.typedID, checkpoint)
-            .then(({ classificationLabelId, options }: GetJustificationReasonsResponse) => {
+            .then(({ classificationLabelId, options = [] }: GetJustificationReasonsResponse) => {
                 this.setState({
                     classificationLabelId,
                     justificationReasons: options.map(({ id, title }) => ({
--- a/src/features/unified-share-modal/flowTypes.js
+++ b/src/features/unified-share-modal/flowTypes.js
@@ -208,7 +208,7 @@ export type justificationReasonType = {
 };
 export type getJustificationReasonsResponseType = {
     classificationLabelId: string,
-    options: Array<justificationReasonType>,
+    options?: Array<justificationReasonType>,
 };
 
 // Prop types used in the invite section of the Unified Share Form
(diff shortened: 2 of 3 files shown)


src/features/unified-share-modal/UnifiedShareForm.js
```diff
@@ -117,10 +117,12 @@ class UnifiedShareForm extends React.Component<USFProps, State> {
             .then(({ classificationLabelId, options }: GetJustificationReasonsResponse) => {
```
greg-in-a-box, line 117: I think we should update `GetJustificationReasonsResponse` that `options` is nullable

greg-in-a-box (changes requested):

src/features/unified-share-modal/UnifiedShareForm.js
jstoffan: Could probably also do `options = {}` to simplify things a bit.

jpfranco, line 117: nit: Can be a bit less verbose if we set a default value like:
```
.then(({ classificationLabelId, options = [] }:
```

jpan-box: done both, but went with @jpfranco 's suggestion `options = []`. Using `{}` resulted in `options.map is not a function` error

src/features/unified-share-modal/flowTypes.js
```diff
@@ -208,7 +208,7 @@ export type justificationReasonType = {
 };
 export type getJustificationReasonsResponseType = {
     classificationLabelId: string,
-    options: Array<justificationReasonType>,
+    options: Array<justificationReasonType> | null,
```
megansmith-box, line 211: I think this is equivalent
```suggestion
    options: ?Array<justificationReasonType>,
```

greg-in-a-box, line 211:
```suggestion
    options?: Array<justificationReasonType>,
```

sorry i used the wrong word, meant to say "optional", since you were getting an undefined not null.

src/features/unified-share-modal/UnifiedShareForm.js
jstoffan: Whoops, my bad. I think my brain automatically parsed "options" as an object. Sorry about that.

greg-in-a-box (approved):

jpfranco (approved):

PR #3345, opened 2023-05-31
chore(deps): Bump react-beautiful-dnd to version 11.0.5

apieceofbart:
I want to bump version of react-beautiful-dnd to 11.0.5 in order to unify usage of react-redux which is a dependency of react-beautiful-dnd. 
This is a breaking change in a sense that because react-beautiful-dnd is a peer dependency any upstream repos (like admin-console or EndUserApp) would need to update their version of react-beautiful-dnd. I did test the change in admin-console and it looks fine.


src/components/draggable-list/DraggableList.tsx
```diff
@@ -22,6 +22,7 @@ const DraggableList = ({ children, className, listId, onDragEnd }: DraggableList
                 {droppableProvided => (
                     <div ref={droppableProvided.innerRef} className={className}>
                         {children}
+                        {droppableProvided.placeholder}
```
apieceofbart, line 25: This is to avoid warning in the console, since version 10 (I believe), Droppable need to use placeholder.

src/features/virtualized-table-renderers/__tests__/draggableRowRenderer.test.js
```diff
@@ -31,7 +31,7 @@ describe('features/virtualized-table-renderers/draggableRowRenderer', () => {
         };
         draggableProvided = { draggableProps: {}, dragHandleProps: {} };
         draggableSnapshot = { isDragging: false };
-        wrapper = getWrapper(rowRendererParams);
+        wrapper = getWrapper(rowRendererParams).dive();
```
apieceofbart, line 34: DragDropContext is now rendering ErrorBoundary as a top most component so I'm using dive to skip one level. This way I don't need to modify the rest of the code.

tjuanitas: would it work to add `.dive()` to the `shallow` in the implementation of `getWrapper` so that it doesn't need to be added to the individual instances?

tjuanitas (commented):
thanks for handling this!

it might better to change `feat` -> `chore`

since this is updating a peer dependency, could you include `BREAKING CHANGE` in the PR title? (we add to title because of how the squash merging works in this repo)

apieceofbart:
> thanks for handling this!
> 
> it might better to change `feat` -> `chore`
> 
> since this is updating a peer dependency, could you include `BREAKING CHANGE` in the PR title? (we add to title because of how the squash merging works in this repo)

I didn't go for the chore initially since according to [docs](https://github.com/box/box-ui-elements/blob/master/CONTRIBUTING.md) chores are not included in the release notes (unless I'm reading that wrong).
I have changed that now.

apieceofbart:
> thanks for handling this!
> 
> it might better to change `feat` -> `chore`
> 
> since this is updating a peer dependency, could you include `BREAKING CHANGE` in the PR title? (we add to title because of how the squash merging works in this repo)

I didn't go for the chore initially since according to [docs](https://github.com/box/box-ui-elements/blob/master/CONTRIBUTING.md) chores are not included in the release notes (unless I'm reading that wrong).
I have changed that now.

greg-in-a-box (approved):

tjuanitas (approved):

tjuanitas:
> > thanks for handling this!
> > it might better to change `feat` -> `chore`
> > since this is updating a peer dependency, could you include `BREAKING CHANGE` in the PR title? (we add to title because of how the squash merging works in this repo)
> 
> I didn't go for the chore initially since according to [docs](https://github.com/box/box-ui-elements/blob/master/CONTRIBUTING.md) chores are not included in the release notes (unless I'm reading that wrong). I have changed that now.

good point! I'm not actually sure what happens with `chore` + `BREAKING CHANGE` combined. feel free to change back if you want

PR #3347, opened 2023-05-31
chore(node): Upgrade to node 18

greg-in-a-box:
BREAKING CHANGE: Upgrade to node 18

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,8 @@
-
 aliases:
+  - &curl-install
+    name: Install curl
+    command: apt-get update && apt-get install curl -y
+
   - &yarn
     name: Installing dependencies
     command: yarn install --non-interactive --frozen-lockfile --cache-folder ~/.cache/yarn
@@ -14,21 +17,22 @@ aliases:
 
   - &restore-yarn-cache
     keys:
-      - yarn-{{ checksum "yarn.lock" }}-{{ checksum ".circleci/config.yml" }}
+      - yarn-{{ arch }}-{{ checksum "yarn.lock" }}-{{ checksum ".circleci/config.yml" }}
 
   - &save-yarn-cache
     paths:
       - node_modules
       - ~/.npm
       - ~/.cache
-    key: yarn-{{ checksum "yarn.lock" }}-{{ checksum ".circleci/config.yml" }}
+    key: yarn-{{ arch }}-{{ checksum "yarn.lock" }}-{{ checksum ".circleci/config.yml" }}
 
 defaults: &defaults
   working_directory: ~/buie
   docker:
-    - image: cypress/base:14.19.0
+    - image: cypress/included:12.13.0
 
 version: 2
+
 jobs:
   lint:
     <<: *defaults
@@ -39,6 +43,7 @@ jobs:
       - save-cache: *save-yarn-cache
       - run: *clean
       - run: *i18n
+      - run: *curl-install
       - run:
           name: Commit lint
           command: ./scripts/commitlint.sh
@@ -82,25 +87,6 @@ jobs:
       - run:
           name: E2E tests
           command: yarn test:e2e
-
-  visual-tests:
-    <<: *defaults
-    steps:
-      - checkout
-      - restore-cache: *restore-yarn-cache
-      - run: *yarn
-      - save-cache: *save-yarn-cache
-      - run:
-          name: Visual tests
-          command: yarn test:visuals
-      - run:
-          name: Copy diff artifacts
-          command: |
-            mkdir -p /tmp/visual-diff-artifacts
-            find src/. -name '*-diff.png' -exec cp {} /tmp/visual-diff-artifacts/ \;
-          when: on_fail
-      - store_artifacts:
-          path: /tmp/visual-diff-artifacts
 
 workflows:
   version: 2
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,1 +1,2 @@
 registry "https://registry.yarnpkg.com"
+--install.ignore-optional true
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -4,7 +4,7 @@ Our development setup assumes a LINUX/BSD environemnt.
 
 ## Project Setup
 
-1. Install Node v14.
+1. Install Node v18.
 2. Install Yarn package manager `https://yarnpkg.com/en/docs/install` v1.10 or higher.
 3. Fork the upstream repo `https://github.com/box/box-ui-elements` via github.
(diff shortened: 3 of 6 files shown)


.yarnrc
```diff
@@ -1 +1,2 @@
 registry "https://registry.yarnpkg.com"
+--install.ignore-optional true
```
greg-in-a-box, line 2: avoid installing fsevents

QingyuChai (approved):
looks straightforward enough for me - no breaking changes in the code between node 14 and 18?

tjuanitas (commented):
Development guide should be updated for node v18: https://github.com/box/box-ui-elements/blob/master/DEVELOPING.md

probably amend and force-push to keep a single commit, otherwise I don't think the BREAKING CHANGE will keep

QingyuChai:
lint and e2e failed, so gonna rerun

package.json
```diff
@@ -119,7 +119,7 @@
         }
     },
     "engines": {
-        "node": ">=14.0.0 <15.0.0",
+        "node": ">=18.16.0",
```
tjuanitas, line 122: I think `18.x` or `18.0.0` is safest for customers. those already on v18 won't need to upgrade again and is compatible with frontend-mono if teams plan to use a version between `18.14.2` and `18.16.0`

.circleci/config.yml
```diff
@@ -26,17 +26,18 @@ aliases:
 defaults: &defaults
   working_directory: ~/buie
   docker:
-    - image: cypress/base:14.19.0
+    - image: cimg/node:18.16.0
```
greg-in-a-box, line 29: updated the general image for over tasks, this image does not include the env or dependencies for cypress but is immediately available on circleci without downloading it.

https://circleci.com/developer/images/image/cimg/node

.circleci/config.yml
```diff
@@ -74,36 +75,19 @@ jobs:
...
-      - restore-cache: *restore-yarn-cache
+      - restore_cache: *restore-yarn-cache
       - run: *yarn
-      - save-cache: *save-yarn-cache
+      - save_cache: *save-yarn-cache
       - run:
           name: E2E tests
           command: yarn test:e2e
 
-  visual-tests:
```
greg-in-a-box, line 86: removed because it will get replaced later and its currently not being used

package.json
```diff
@@ -373,5 +373,8 @@
             "@hapi/address": "Version 2.1.2+ requires a polyfill for TextEncoder. Pinning to version 2.1.1 prevents IE11 from breaking.",
             "react-tether": "Version 2.x has too many breaking changes and requires forwardRef on all components"
         }
+    },
+    "resolutions": {
```
greg-in-a-box, line 377: There is a bug at 18.0.0 that requires this resolution otherwise storybook breaks

package.json
```diff
@@ -373,5 +373,8 @@
...
+        "ip": "1.1.8"
```
greg-in-a-box, line 378: There is a bug at 18.0.0 that requires this resolution otherwise storybook breaks

DEVELOPING.md
```diff
@@ -4,7 +4,7 @@ Our development setup assumes a LINUX/BSD environemnt.
 
 ## Project Setup
 
-1. Install Node v14.
+1. Install Node v18.
 2. Install Yarn package manager `https://yarnpkg.com/en/docs/install` v1.10 or higher.
```
QingyuChai, line 8: any chance we will require a higher yarn version? is 1.10 compatible with node 14?

.circleci/config.yml
```diff
@@ -74,36 +75,19 @@ jobs:
...
     steps:
       - checkout
-      - restore-cache: *restore-yarn-cache
+      - restore_cache: *restore-yarn-cache
       - run: *yarn
-      - save-cache: *save-yarn-cache
+      - save_cache: *save-yarn-cache
       - run:
           name: E2E tests
           command: yarn test:e2e
```
tjuanitas, line 88: the smallest of small things: there's an extra line break between these sections

DEVELOPING.md
greg-in-a-box: it didnt seem necessary to change the yarn version, I would assume its compatible with node 14 since thats what we have been using.  unless you meant 18, I would assume it wouldnt since the min require eng is 14

tjuanitas (approved):
looks good to me!

jstoffan (approved):

DEVELOPING.md
QingyuChai: yeah, I meant whether yarn 1.10 is compatible with node 18, mb!

QingyuChai (approved):

.yarnrc
jaknas: just curious why this is needed. fsevents is used in chokidar for native macOS syscalls. this should help with perf when watching files

tjuanitas: I think the version of fsevents that's being installed is requiring python2

jpan-box (dismissed):

TaoJiang98 (approved):

.circleci/config.yml
```diff
@@ -26,17 +30,19 @@ aliases:
...
+version: 2.1
 
-version: 2
 jobs:
   lint:
     <<: *defaults
     steps:
       - checkout
-      - restore-cache: *restore-yarn-cache
+#      - run: *chmod-working-dir
```
greg-in-a-box, line 42: commenting out caching related config for now, to separate node 18 bump, so that the issue for caching can be resolved after this PR goes through

.circleci/config.yml
```diff
@@ -1,5 +1,8 @@
-
 aliases:
+  - &curl-install
```
greg-in-a-box, line 2: image doesnt have curl installed by default

.circleci/config.yml
```diff
@@ -14,21 +17,22 @@ aliases:
...
       - ~/.cache
-    key: yarn-{{ checksum "yarn.lock" }}-{{ checksum ".circleci/config.yml" }}
+    key: yarn-{{ arch }}-{{ checksum "yarn.lock" }}-{{ checksum ".circleci/config.yml" }}
 
 defaults: &defaults
-  working_directory: ~/buie
+  working-directory: ~/buie
   docker:
-    - image: cypress/base:14.19.0
+    - image: cypress/included:12.13.0
```
greg-in-a-box, line 32: using only one image for all since the keys are the same for all jobs, otherwise throws permission error on subsequent runs

tjuanitas (dismissed):

tjuanitas (approved):

QingyuChai (approved):

PR #3353, opened 2023-06-09
feat(security-controls): Remove watermark learn more link

jpfranco:
Removing the "Learn more" link for Watermark restrictions on the Security Controls. This is being done for accessibility reasons since the controls will now be displayed within a hovercard for which no interactive elements are allowed.

Screenshot 2023-06-09 at 8 37 49 AM


src/features/classification/security-controls/utils.js
```diff
@@ -99,20 +96,9 @@ const getSharedLinkMessages = (controls: Controls): Array<MessageItem> => {
 const getWatermarkingMessages = (controls: Controls): Array<MessageItem> => {
     const items = [];
     const isWatermarkEnabled = getProp(controls, `${WATERMARK}.enabled`, false);
+
     if (isWatermarkEnabled) {
```
bxie-box, line 100: qq: (completely okay if we want to do this in another PR) but since we're here, could we clean up the watermarking check since it's fully GA? thank you!

jpfranco: I'll file a ticket for this since it's already a bit outside of the scope of what I'm currently working on

bxie-box (approved):

PR #3357, opened 2023-06-09
chore(ci): change regex for registry check in release script

greg-in-a-box:

### DIFF
--- a/scripts/cut_release.sh
+++ b/scripts/cut_release.sh
@@ -10,8 +10,8 @@ blue=$"\n\e[1;34m(ℹ) "
 end=$"\e[0m\n"
 
 # While running yarn, the registry changes to registry.yarnpkg.com which is a mirror to the public NPM registry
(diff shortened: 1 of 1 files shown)


scripts/cut_release.sh
```diff
@@ -10,8 +10,8 @@ blue=$"\n\e[1;34m(ℹ) "
 end=$"\e[0m\n"
 
 # While running yarn, the registry changes to registry.yarnpkg.com which is a mirror to the public NPM registry
-YARN_PUBLIC_REGISTRY_REGEX="^https://registry\.yarnpkg\.com/$"
-NPM_PUBLIC_REGISTRY_REGEX="^https://registry\.npmjs\.org/$"
+YARN_PUBLIC_REGISTRY_REGEX="^https://registry\.yarnpkg\.com\/?$"
+NPM_PUBLIC_REGISTRY_REGEX="^https://registry\.npmjs\.org\/?$"
```
tjuanitas, line 14: why not remove the trailing slash?

or if you escape the trailing slash, you could escape the ones after the protocol for consistency

PR #3360, opened 2023-06-09
chore(ci): unset OPENSSL_FORCE_FIPS_MODE

greg-in-a-box:

### DIFF
--- a/scripts/cut_release.sh
+++ b/scripts/cut_release.sh
@@ -111,6 +111,7 @@ lint_and_test() {
 
     # Tests
     printf "${blue}Running tests...${end}"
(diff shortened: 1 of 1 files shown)


scripts/cut_release.sh
```diff
@@ -111,6 +111,7 @@ lint_and_test() {
 
     # Tests
     printf "${blue}Running tests...${end}"
+    unset OPENSSL_FORCE_FIPS_MODE
```
tjuanitas, line 114: if this is used, do we still need the `--openssl-legacy-provider` flag on `yarn test`? or does this only work when used together?

PR #3361, opened 2023-06-09
feat(cloud-game): Adjust cloud game visual

Machoper:
## Summary of changes

- Make cloud objects bigger to accommodate small screens
- Add a max height of 40px to the message header and show tooltip on hover
- Prevent auto-scroll into view on load

## Demo
adjust_cloud_game_visual

### DIFF
--- a/src/features/security-cloud-game/SecurityCloudGame.js
+++ b/src/features/security-cloud-game/SecurityCloudGame.js
@@ -3,6 +3,7 @@ import React, { useEffect, useLayoutEffect, useRef, useState } from 'react';
 import { FormattedMessage, injectIntl } from 'react-intl';
 
 import FocusTrap from '../../components/focus-trap';
+import Tooltip from '../../components/tooltip';
 
 import DragCloud from './DragCloud';
 import DropCloud from './DropCloud';
@@ -12,8 +13,8 @@ import { checkOverlap, getGridPosition, getRandomCloudPosition } from './utils';
 import './SecurityCloudGame.scss';
 
 // pick these numbers to balance accessibility and game complexity
-const CLOUD_SIZE_RATIO = 5;
-const GRID_TRACK_SIZE_RATIO = 20;
+const CLOUD_SIZE_RATIO = 4;
+const GRID_TRACK_SIZE_RATIO = 16;
 
 const SecurityCloudGame = ({ height, intl: { formatMessage }, onValidDrop, width }) => {
     const [dropCloudPosition, setDropCloudPosition] = useState(null);
@@ -43,7 +44,7 @@ const SecurityCloudGame = ({ height, intl: { formatMessage }, onValidDrop, width
             gridTrackSize: minGameBoardLength / GRID_TRACK_SIZE_RATIO,
         });
 
-        messageElement.focus();
+        messageElement.focus({ preventScroll: true });
     }, [height, width]);
 
     useEffect(() => {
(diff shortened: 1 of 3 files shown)


src/features/security-cloud-game/SecurityCloudGame.js
```diff
@@ -197,6 +198,11 @@ const SecurityCloudGame = ({ height, intl: { formatMessage }, onValidDrop, width
         return <FormattedMessage {...messages.instructions} />;
     };
 
+    const hideMessageTooltip = () => {
```
tjuanitas, line 201: since this function is returning a boolean, could it be something like `isTooltipShown`?

from the current name I thought the method might be updating a state

src/features/security-cloud-game/SecurityCloudGame.js
```diff
@@ -12,8 +13,8 @@ import { checkOverlap, getGridPosition, getRandomCloudPosition } from './utils';
 import './SecurityCloudGame.scss';
 
 // pick these numbers to balance accessibility and game complexity
-const CLOUD_SIZE_RATIO = 5;
-const GRID_TRACK_SIZE_RATIO = 20;
+const CLOUD_SIZE_RATIO = 4;
+const GRID_TRACK_SIZE_RATIO = 16;
```
tjuanitas, line 17: just curious, how were these values determined? are there any min / max size considerations?

src/features/security-cloud-game/SecurityCloudGame.js
```diff
@@ -43,7 +44,7 @@ const SecurityCloudGame = ({ height, intl: { formatMessage }, onValidDrop, width
             gridTrackSize: minGameBoardLength / GRID_TRACK_SIZE_RATIO,
         });
 
-        messageElement.focus();
+        messageElement.focus({ preventScroll: true });
```
tjuanitas, line 47: are there any accessibility concerns if it becomes focused but is off-screen without auto-scroll?

src/features/security-cloud-game/SecurityCloudGame.scss
```diff
@@ -49,6 +53,10 @@
     left: -9999px;
 }
 
+.bdl-SecurityCloudGame-tooltip.bdl-Tooltip {
```
tjuanitas, line 56: is the `.bdl-Tooltip` part needed? or could it be removed to lower specificity?

src/features/security-cloud-game/SecurityCloudGame.scss
```diff
@@ -30,10 +30,14 @@
 
     .bdl-SecurityCloudGame-message {
         flex: none;
+        max-height: 40px;
```
tjuanitas, line 33: does the height change now? since the content doesn't wrap anymore would this be a fixed `height` instead of `max-height`?

tjuanitas (commented):
left some clarifying questions since I forgot how the game works but looks good overall!

src/features/security-cloud-game/SecurityCloudGame.js
Machoper: I removed focus-related code as it's not the best UX when an embed widget/cloud game is put at the bottom of a scrollable page or when there are multiple widgets/games on the page. So this function is no longer needed as it was only there to prevent the tooltip from showing up when the header is focused on load.

src/features/security-cloud-game/SecurityCloudGame.js
Machoper: I queried our embed widget analytics data for the height and width being used and determined that these values work better across all sizes (330 x 150 - 2560 x 1600 +).

src/features/security-cloud-game/SecurityCloudGame.js
Machoper: I removed this line due to the UX issue mentioned above, so it should not be a concern anymore.

src/features/security-cloud-game/SecurityCloudGame.scss
Machoper: Yes, it's needed. I tried removing `.bdl-Tooltip` but it didn't work.

tjuanitas (approved):
lgtm!!

greg-in-a-box (approved):

PR #3365, opened 2023-06-21
fix: update modal buttons alignments in element dialogs

tjuanitas:
This is updating the designs of the modals used in Content Explorer and Content Picker. There are three modal components that this is specifically targeting: `CreateFolderDialog.js`, `DeleteConfirmationDialog.js`, and `RenameDialog.js`

I decided to add `width: 100%;` to all text fields in the base elements stylesheet since `100%` seems to be the most ideal value for `width` in majority of cases for input.

At the moment, components are inheriting `width: 262px` from the [base components stylesheet](https://github.com/box/box-ui-elements/blob/master/src/styles/common/_forms.scss#L9-L22) but this seems to be an arbitrary value and components are overriding it. e.g.

- Task modal in Sidebar: https://github.com/box/box-ui-elements/blob/b9d196fcf330d38de87890ca8ccf4a5c8433d313/src/elements/content-sidebar/activity-feed/task-form/TaskForm.scss#L14
- Description in File Properties section in Sidebar: https://github.com/box/box-ui-elements/blob/b9d196fcf330d38de87890ca8ccf4a5c8433d313/src/features/item-details/ItemProperties.scss#L16
- Content Sharing / USM: https://github.com/box/box-ui-elements/blob/b9d196fcf330d38de87890ca8ccf4a5c8433d313/src/features/unified-share-modal/UnifiedShareModal.scss#L181

**Before**
Screenshot 2023-07-05 at 11 19 50 AM

**After**
Screenshot 2023-07-05 at 11 19 35 AM

### DIFF
--- a/src/elements/common/header/Header.scss
+++ b/src/elements/common/header/Header.scss
@@ -1,6 +1,6 @@
 @import '../variables';
 
-.be .be-header {
+.be-header {
     display: flex;
     flex: 0 0 70px;
     align-items: center;
@@ -15,11 +15,5 @@
     .be-search {
         flex: 1;
         padding-left: 20px;
-    }
-
-    input[type='search'] {
-        width: 100%;
-        font: inherit;
-        -webkit-appearance: none;
     }
 }
--- a/scripts/styleguide.styles.scss
+++ b/scripts/styleguide.styles.scss
@@ -2,6 +2,10 @@
     &.bce,
     &.bcp {
         height: 600px;
+
+        .bcu {
+            height: 100%;
+        }
     }
 
     &.bcpr {
--- a/src/components/i18n/FormattedCompMessage.js
+++ b/src/components/i18n/FormattedCompMessage.js
@@ -184,7 +184,7 @@ class FormattedCompMessage extends React.Component<Props, State> {
         // the reviewer finds and submit translation tickets to Jira and/or fixed
         // translations directly back to Mojito.
         // 3. It can be used by the planned "text experiment framework" to identify
-        // whole strings in the UI that can be A/B tested in various langauges without
+        // whole strings in the UI that can be A/B tested in various languages without
         // publishing new versions of the code.
         return React.createElement(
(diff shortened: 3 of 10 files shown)


tjuanitas:
note to self: the textarea in USM seems to render immediately while the rest of the UI is loading

src/elements/common/header/Header.scss
```diff
@@ -1,6 +1,6 @@
 @import '../variables';
 
-.be .be-header {
```
tjuanitas, line 3: the `.be` on this selector breaks the `.be-is-small &` selector below

src/elements/common/header/Header.scss
```diff
@@ -16,10 +16,4 @@
         flex: 1;
         padding-left: 20px;
     }
-
-    input[type='search'] {
-        width: 100%;
-        font: inherit;
-        -webkit-appearance: none;
-    }
```
tjuanitas, line 24: I'm not too sure what these lines were intended for. my assumption is this was related to a Safari issue but we don't target `input[type='search']` for any other components so I don't know why this was needed.

I think it's okay to remove the block now:

- `width: 100%;` is being added here: https://github.com/box/box-ui-elements/blob/8c0e2989690b882b1ce299566ac0105d576ad1ff/src/elements/common/base.scss#L25
- `font: inherit;` is applied here: https://github.com/box/box-ui-elements/blob/8c0e2989690b882b1ce299566ac0105d576ad1ff/src/elements/common/base.scss#L62
- `-webkit-appearance: none;` from my testing this wasn't having an impact in most recent version of browsers. I also don't think we want this based on this: https://github.com/box/box-ui-elements/blob/b9d196fcf330d38de87890ca8ccf4a5c8433d313/src/styles/common/_forms.scss#L106

tjuanitas:
I'll continue more testing of the elements to verify `width: 100%` has no unintended side effects but PR is ready for review now.

PR #3369, opened 2023-06-28
feat(uaa-integration): Filter activity types for UAA

JChan106:
This filters the call to file_activities depending on the parameters passed into the feed. Specifically for annotations, we previously threw an error if the specific file doesn't have the ability to add annotations. No call to file_activities would occur because of this. So now we just check if there's the proper permissions, if not, we filter annotations out.

image

### DIFF
--- a/src/api/Feed.js
+++ b/src/api/Feed.js
@@ -35,6 +35,7 @@ import {
     FILE_ACTIVITY_TYPE_TASK,
     HTTP_STATUS_CODE_CONFLICT,
     IS_ERROR_DISPLAYED,
+    PERMISSION_CAN_VIEW_ANNOTATIONS,
     TASK_NEW_APPROVED,
     TASK_NEW_COMPLETED,
     TASK_NEW_REJECTED,
@@ -518,13 +519,17 @@ class Feed extends Base {
 
         const versionsPromise = shouldShowVersions ? this.fetchVersions() : Promise.resolve();
         const currentVersionPromise = shouldShowVersions ? this.fetchCurrentVersion() : Promise.resolve();
+
+        const filteredActivityTypes = [];
+        if (shouldShowAnnotations && permissions[PERMISSION_CAN_VIEW_ANNOTATIONS]) {
+            filteredActivityTypes.push(FILE_ACTIVITY_TYPE_ANNOTATION);
+        }
+        if (shouldShowAppActivity) filteredActivityTypes.push(FILE_ACTIVITY_TYPE_APP_ACTIVITY);
+        filteredActivityTypes.push(FILE_ACTIVITY_TYPE_COMMENT);
+        if (shouldShowTasks) filteredActivityTypes.push(FILE_ACTIVITY_TYPE_TASK);
+
         const fileActivitiesPromise = shouldUseUAA
-            ? this.fetchFileActivities(permissions, [
-                  FILE_ACTIVITY_TYPE_ANNOTATION,
-                  FILE_ACTIVITY_TYPE_APP_ACTIVITY,
-                  FILE_ACTIVITY_TYPE_COMMENT,
-                  FILE_ACTIVITY_TYPE_TASK,
-              ])
+            ? this.fetchFileActivities(permissions, filteredActivityTypes, shouldShowReplies)
             : Promise.resolve();
 
         const handleFeedItems = (feedItems: FeedItems) => {
@@ -677,7 +682,11 @@ class Feed extends Base {
      * @param {FileActivityTypes[]} activityTypes - the activity types to filter by
      * @return {Promise} - the file comments
      */
-    fetchFileActivities(permissions: BoxItemPermission, activityTypes: FileActivityTypes[]): Promise<Object> {
+    fetchFileActivities(
+        permissions: BoxItemPermission,
+        activityTypes: FileActivityTypes[],
+        shouldShowReplies,
+    ): Promise<Object> {
         this.fileActivitiesAPI = new FileActivitiesAPI(this.options);
         return new Promise(resolve => {
             this.fileActivitiesAPI.getActivities({
@@ -686,6 +695,7 @@ class Feed extends Base {
                 permissions,
                 successCallback: resolve,
                 activityTypes,
+                shouldShowReplies,
             });
         });
     }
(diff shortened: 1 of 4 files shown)


src/api/FileActivities.js
```diff
@@ -5,18 +5,23 @@
...
+const getFileActivityQueryParams = (
+    fileID: string,
+    activityTypes?: FileActivityTypes[] = [],
+    shouldShowReplies?: boolean = false,
+) => {
     const baseEndpoint = `/file_activities?file_id=${fileID}`;
     const hasActivityTypes = !!activityTypes && !!activityTypes.length;
-    const enabledRepliesQueryParam = `&enable_replies=true&reply_limit=${REPLY_LIMIT}`;
+    const enableReplies = shouldShowReplies ? 'true' : 'false';
+    const enabledRepliesQueryParam = `&enable_replies=${enableReplies}&reply_limit=${REPLY_LIMIT}`;
```
patlm, line 24: minor - you should be able to embed the boolean shouldShowReplies directly into the string and it will automatically convert to the string version of true or false (that way enableReplies const can be removed). Like:
```
`&enable_replies=${shouldShowReplies}&reply_limit=${REPLY_LIMIT}`
```

src/api/FileActivities.js
```diff
@@ -67,7 +74,6 @@ class FileActivities extends Base {
             }
 
             this.checkApiCallValidity(PERMISSION_CAN_COMMENT, permissions, fileID);
-            this.checkApiCallValidity(PERMISSION_CAN_VIEW_ANNOTATIONS, permissions, fileID);
```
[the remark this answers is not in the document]
JChan106: Interesting point, I think this could add this in just so we don't have any permission concerns. However it is a drawback since if the annotations permission fails, the whole aggregated UAA call is not made.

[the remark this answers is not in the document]
patlm: That is a good point. Since it is already pre-checking the permission in the if block before calling this function, it seems like this now might be a question of what we want the experience to be like if from somewhere else this function is directly called in the future.

Would it be better for the annotation request to fail silently in that all of the activities are fetched and returned like expected (but there is an error in the "errors" list because of annotations in the API response).

Or for the request not to go through - and the developer has to recognize and catch that there isn't any activity being returned so they need to verify when annotations should be part of the activityTypes list.

src/api/FileActivities.js
JChan106: linter has problems with not making it a string, kept as is

src/api/Feed.js
```diff
@@ -518,13 +519,17 @@ class Feed extends Base {
...
         const versionsPromise = shouldShowVersions ? this.fetchVersions() : Promise.resolve();
         const currentVersionPromise = shouldShowVersions ? this.fetchCurrentVersion() : Promise.resolve();
+
+        const filteredActivityTypes = [];
+        if (shouldShowAnnotations && permissions[PERMISSION_CAN_VIEW_ANNOTATIONS]) {
+            filteredActivityTypes.push(FILE_ACTIVITY_TYPE_ANNOTATION);
+        }
+        if (shouldShowAppActivity) filteredActivityTypes.push(FILE_ACTIVITY_TYPE_APP_ACTIVITY);
+        filteredActivityTypes.push(FILE_ACTIVITY_TYPE_COMMENT);
+        if (shouldShowTasks) filteredActivityTypes.push(FILE_ACTIVITY_TYPE_TASK);
```
tjuanitas, line 529: not sure if this is more readable but:
```
const filteredActivityTypes = [
    FILE_ACTIVITY_TYPE_COMMENT,
    ...(shouldShowAnnotations && permissions[PERMISSION_CAN_VIEW_ANNOTATIONS] ? [FILE_ACTIVITY_TYPE_ANNOTATION] : []),
    ...(shouldShowAppActivity ? [FILE_ACTIVITY_TYPE_APP_ACTIVITY] : []),
    ...(shouldShowTasks ? [FILE_ACTIVITY_TYPE_TASK] : []),
];
```
and then it doesn't mutate the array

tjuanitas (dismissed):
makes sense to me

src/api/Feed.js
```diff
@@ -518,13 +519,17 @@ class Feed extends Base {
 
         const versionsPromise = shouldShowVersions ? this.fetchVersions() : Promise.resolve();
         const currentVersionPromise = shouldShowVersions ? this.fetchCurrentVersion() : Promise.resolve();
+
+        const filteredActivityTypes = [];
+        if (shouldShowAnnotations && permissions[PERMISSION_CAN_VIEW_ANNOTATIONS]) {
+            filteredActivityTypes.push(FILE_ACTIVITY_TYPE_ANNOTATION);
+        }
+        if (shouldShowAppActivity) filteredActivityTypes.push(FILE_ACTIVITY_TYPE_APP_ACTIVITY);
+        filteredActivityTypes.push(FILE_ACTIVITY_TYPE_COMMENT);
```
tjuanitas, line 528: would `PERMISSION_CAN_COMMENT` be checked for `FILE_ACTIVITY_TYPE_COMMENT`, similar to annotations?

src/api/Feed.js
JChan106: Good point about the mutations, I think I can store these in variables to make this a bit easier to read.

src/api/Feed.js
JChan106: Good point, will do to keep consistent

src/api/Feed.js
```diff
@@ -520,13 +520,18 @@ class Feed extends Base {
...
+        const annotationActivityType =
+            shouldShowAnnotations && permissions[PERMISSION_CAN_VIEW_ANNOTATIONS]
+                ? [FILE_ACTIVITY_TYPE_ANNOTATION]
+                : [];
+        const appActivityActivityType = shouldShowAppActivity ? [FILE_ACTIVITY_TYPE_APP_ACTIVITY] : [];
+        const taskActivityType = shouldShowTasks ? [FILE_ACTIVITY_TYPE_TASK] : [];
+        const filteredActivityTypes = [
+            ...annotationActivityType,
+            ...appActivityActivityType,
+            FILE_ACTIVITY_TYPE_COMMENT,
```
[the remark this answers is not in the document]
JChan106: I don't think so, the annotations check was added for certain file types that don't have annotations. Comments are always apart of the feed so it's not necessary.

tjuanitas (approved):

src/api/Feed.js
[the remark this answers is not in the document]
tjuanitas: gotcha. in that case, the `activityTypes.includes(FILE_ACTIVITY_TYPE_COMMENT` shouldn't be necessary since it'll always be included

patlm (approved):

PR #3371, opened 2023-06-30
fix(content-picker): check share permission for default shared links

tjuanitas:
Check if the item has the `can_share` permission before trying to create a shared link with default access level. There's a bug for roles that do not have permission to generate shared links such as Previewer.


src/api/WebLink.js
```diff
@@ -37,14 +36,14 @@ class WebLink extends Item {
      * Gets a Box weblink
      *
      * @param {string} id - Weblink id
-     * @param {(newItem: BoxItem) => void} successCallback - Function to call with results
```
tjuanitas, line 40: updated the type since there were issues with flow. this aligns with how it's written for `getFile` and `getFolder` methods

greg-in-a-box (approved):

joshmarnold (approved):

PR #3376, opened 2023-07-25
fix(metadata): Disable updates to metadata

joshmarnold:
### Summary

There is an option in the Admin Console that allows an admin to restrict which users have the ability to create cascading metadata. At some point in the not-so-distant past (sometime in the last 6-12 months as of July 2023) the feature was broken. Users who previously did not have the ability to create cascading metadata could now do so. 

### Background/History

When folder level metadata was introduced, it seems like it was intended to be synonymous with cascading metadata. That is not the case today. It’s true that most folder level metadata is cascaded, but it’s not the only use case. 

It also appears that with the way the feature was built, it not only restricted cascading metadata, but restricted ALL metadata adds/edits to folders. I believe this was a side effect of the previously mentioned belief that folder = cascade.

An important note to keep in mind with all of this is that the feature was built at the UI layer. Meaning, any user with edit capabilities, irrespective of the Admin Console option, could add/edit/delete folder level metadata or apply/remove cascading metadata via the API. In simpler terms, we put a roadblock in place, but this was NOT a security control.

The intent behind this feature is to allow web users to be editors of folders but not allow them to add cascading metadata to those folders. 

### Solution Requirements
There are 4 options an admin can select for this feature. Those are: “Enable for all managed users”, “Disable for all managed users”, “Enable for select users”, and “Enable for everyone except select users”. To make things simple, users in the following requirements will be denoted as “allowed users” or “disallowed users” regardless of how the status was obtained. We’ll also assume that all users have Editor permissions to the folder in question.</span></p><p style="font-size: 12pt"></p>

Scenario | Can View Metadata on a folder | Can apply a Metadata template to a folder* | Can edit/delete Metadata on a folder* | Can enable/disable Cascade Metadata | Can see if folder* Metadata is being cascaded
-- | -- | -- | -- | -- | --
Allowed user - No Metadata on folder | YES | YES | YES | n/a | n/a
Allowed user - Metadata on folder (no cascade) | YES | YES | YES | YES | YES
Allowed user - Metadata on folder with cascade | YES | YES | YES | YES | YES
Disallowed user - No Metadata on folder | YES | YES | YES | n/a | n/a
Disallowed user - Metadata on folder (no cascade) | YES | YES | YES | NO | YES
Disallowed user - Metadata on folder with cascade | YES | YES | NO | NO | YES


src/features/metadata-instance-editor/Instance.js
```diff
@@ -569,6 +571,7 @@ class Instance extends React.PureComponent<Props, State> {
                         data-resin-target="metadata-instanceedit"
                         onClick={this.toggleIsEditing}
                         type="button"
+                        data-testid="metadata-instance-edit-button"
```
JChan106, line 574: RTL doesn't recommend using extra test attributes unless necessary, do we need to add a testid to this button? Are we able to find the IconEdit and click that instead?

joshmarnold: Now using the aria-label

src/features/metadata-instance-editor/Instance.js
```diff
@@ -579,7 +581,13 @@ class Instance extends React.PureComponent<Props, State> {
     };
 
     render() {
-        const { cascadePolicy = {}, isDirty, isCascadingPolicyApplicable, isOpen, template }: Props = this.props;
+        const {
+            cascadePolicy = {},
+            isDirty,
+            isCascadingPolicyApplicable = false, // does user have cascading policy permissions
```
tjuanitas, line 587: can we move this to `static defaultProps` at the top?

src/features/metadata-instance-editor/Instance.js
```diff
@@ -209,6 +209,8 @@ class Instance extends React.PureComponent<Props, State> {
             onSave,
         }: Props = this.props;
         const { data: currentData, errors, isCascadingEnabled, isCascadingOverwritten }: State = this.state;
+        const hasCascadePolicy = !!cascadePolicy?.id;
+        const hasCascadePermission = !hasCascadePolicy || (hasCascadePolicy && isCascadingPolicyApplicable);
```
tjuanitas, line 213: I think this boolean logic is the same as:

```
const hasCascadePermission = !cascadePolicy?.id || isCascadingPolicyApplicable;
```

src/features/metadata-instance-editor/Instance.js
```diff
@@ -592,6 +600,9 @@ class Instance extends React.PureComponent<Props, State> {
         }: State = this.state;
         const isProperties = template.templateKey === TEMPLATE_CUSTOM_PROPERTIES;
         const isEditing = this.isEditing();
+        const hasCascadePolicy = !!cascadePolicy?.id;
+        const hasCascadePermission = !hasCascadePolicy || (hasCascadePolicy && isCascadingPolicyApplicable);
```
tjuanitas, line 604: same comment, isn't this:
```
const hasCascadePermission = !cascadePolicy?.id || isCascadingPolicyApplicable;
```

src/features/metadata-instance-editor/Instance.js
```diff
@@ -592,6 +600,9 @@ class Instance extends React.PureComponent<Props, State> {
...
+        const canEditInstance = !!(isEditing && hasCascadePermission);
```
tjuanitas, line 605: nit: `isEditing` and `hasCascadePermission` are booleans so you don't need the double bang
```suggestion
        const canEditInstance = isEditing && hasCascadePermission;
```

src/features/metadata-instance-editor/Instance.js
```diff
@@ -549,10 +551,10 @@ class Instance extends React.PureComponent<Props, State> {
         return this.canEdit() && isEditing;
     }
 
-    renderEditButton = () => {
+    renderEditButton = (canEditInstance: boolean) => {
         const { intl, isDirty }: Props = this.props;
         const { isBusy }: State = this.state;
-        const canEdit = this.canEdit();
+        const canEdit = this.canEdit() && canEditInstance;
```
tjuanitas, line 557: based on the method descriptions of `canEdit` and `isEditing`, the button only renders now when it's an editing mode, is that correct?
(since `canEditInstance = isEditing && hasCascadePermission`)

how does the component get into editing mode now?

src/features/metadata-instance-editor/Instance.js
```diff
@@ -549,10 +551,10 @@ class Instance extends React.PureComponent<Props, State> {
         return this.canEdit() && isEditing;
     }
 
-    renderEditButton = () => {
+    renderEditButton = (canEditInstance: boolean) => {
```
tjuanitas, line 554: should the boolean checks in `canEditInstance` be moved to the methods `canEdit`?

src/features/metadata-instance-editor/Instance.js
joshmarnold: isEditing can be undefined which is why I did this. Really, I should ensure that it's a boolean above

tjuanitas: `isEditing` being `undefined` sounds like a bug. do you know how that's happening? the methods should return booleans: https://github.com/box/box-ui-elements/blob/63d33f7e73f3cd53ebeaa7ae9804d02c5b9fb8f1/src/features/metadata-instance-editor/Instance.js#L534-L552

I was thinking with the `canEdit` prop but that should be required

src/features/metadata-instance-editor/Instance.js
joshmarnold: This logic has changed

src/features/metadata-instance-editor/Instance.js
```diff
@@ -207,8 +208,11 @@ class Instance extends React.PureComponent<Props, State> {
             isDirty,
             isCascadingPolicyApplicable,
             onSave,
+            template,
         }: Props = this.props;
         const { data: currentData, errors, isCascadingEnabled, isCascadingOverwritten }: State = this.state;
+        const isCustomInstance = template.templateKey === TEMPLATE_CUSTOM_PROPERTIES;
+        const hasCascadePermission = isCustomInstance ? true : isCascadingPolicyApplicable;
```
tjuanitas, line 215: is this saying all custom metadata is cascaded?

is the cascading policy only for the metadata templates created in admin console?

src/features/metadata-instance-editor/Instance.js
```diff
@@ -626,27 +633,26 @@ class Instance extends React.PureComponent<Props, State> {
                         <LoadingIndicatorWrapper isLoading={isBusy}>
                             <Form onValidSubmit={isDirty ? this.onSave : noop}>
                                 <div className="metadata-instance-editor-instance">
-                                    {isCascadingPolicyApplicable && (
-                                        <CascadePolicy
```
tjuanitas, line 630: this always shows the cascading policy, is that intentional? I assume the cascading policy doesn't apply when we add metadata to files

src/features/metadata-instance-editor/Instance.js
```diff
@@ -102,6 +102,7 @@ class Instance extends React.PureComponent<Props, State> {
     static defaultProps = {
         data: {},
         isDirty: false,
+        isCascadingPolicyApplicable: false,
```
tjuanitas, line 105: sorry do we even need to set this explicitly? shouldn't all usages be inherited from the default set in `Instances`?
https://github.com/box/box-ui-elements/blob/9ab6d78b09214d5928add084e5cd72ab8d8631b4/src/features/metadata-instance-editor/Instances.js#L22-L23

src/features/metadata-instance-editor/Instance.js
```diff
@@ -530,12 +534,16 @@ class Instance extends React.PureComponent<Props, State> {
      * @return {boolean} true if editable
      */
     canEdit(): boolean {
-        const { canEdit, onModification, onRemove, onSave }: Props = this.props;
-        return (
+        const { canEdit, onModification, onRemove, onSave, template, isCascadingPolicyApplicable }: Props = this.props;
+        const isCustom = template.templateKey === TEMPLATE_CUSTOM_PROPERTIES;
+        const hasCascadePermission = isCustom ? true : isCascadingPolicyApplicable;
```
tjuanitas, line 539: wouldn't this prevent editing metadata on files?

tjuanitas (commented):
have changes been tested with ContentSidebar?
https://github.com/box/box-ui-elements/blob/master/src/elements/content-sidebar/MetadataSidebar.js

src/features/metadata-instance-editor/Instance.js
joshmarnold: _is this saying all custom metadata has permission to cascade? (is the same true for metadata on files as well?)_
Cascading metadata policies cannot be applied to custom metadata, so we do not utilize cascade permissions when determining a user's ability to modify and save a custom template. Admittedly, the variable name here lacks clarity.

joshmarnold:
> have changes been tested with ContentSidebar? https://github.com/box/box-ui-elements/blob/master/src/elements/content-sidebar/MetadataSidebar.js


@tjuanitas As I was not previously aware of `MetadataSidebar`, I haven't had the opportunity to look into it or event test it manually. I also don't see it's use in EUA to try to test it out. As of now, I can only report that the tests are passing, which is positive news. Upon closer examination, I noticed that it's similar to `MetadataInstanceEditor`, where the `Instances` component is rendered. However, there is a big difference between these two components. It appears that the `MetadataSidebar` has is not utilizing the `isCascadingPolicyApplicable` parameter, which would have rendered it incompatible with cascading policies. Due to my lack of knowledge about the intended purpose of the `MetadataSidebar`, I cannot definitively state what the expected behavior should be or whether any updates are necessary.

src/features/metadata-instance-editor/Instance.js
joshmarnold: You're right but I suppose that it doesn't hurt to set it here as well?

src/features/metadata-instance-editor/Instance.js
joshmarnold: Yeah, it's fine that we're now always rendering the CascadePolicy. Internally, it checks if the user has permissions and shows messages depending on the permission.

File2023-07-28 at 09 14 33

File2023-07-28 at 09 14 52

src/features/metadata-instance-editor/Instance.js
joshmarnold: You're right that it does. This should not be the case. So, need to figure out how to get whether the template is on a file or folder.

Template has a property called “typeKey,” which is always an empty string when inspecting metadata on a folder and is set to “metadata_folder” when looking at metadata on a file. Although I’ve come across some instances of this being used in the monolith codebase, it’s not very clear on the specific conditions in which this is set. Therefore, I’m not 100% confident that this alone will effectively differentiate between folders and files when rendering the metadata.

Alternatively, I could pass in "who called" as 'isFromFolder', for example. I would rather not do this though

src/features/metadata-instance-editor/Instance.js
```diff
@@ -524,18 +530,41 @@ class Instance extends React.PureComponent<Props, State> {
...
+        const isCustom = template.templateKey === TEMPLATE_CUSTOM_PROPERTIES;
+        const hasCascadePermission = isCustom || !isFolderInstance ? true : isCascadingPolicyApplicable;
+
+        return !!(
             canEdit &&
             typeof onRemove === 'function' &&
             typeof onSave === 'function' &&
-            typeof onModification === 'function'
+            typeof onModification === 'function' &&
+            hasCascadePermission
```
tjuanitas, line 567: the changes here looks like it could utilize the existing `canEdit` prop. e.g. in the `FolderMetadataModal` component, there is a map function to construct the `editors` prop that's passed to `MetadataInstanceEditor`

in this map function seems like a good opportunity to set the `canEdit` property based on the `template` and `isCascadingPolicyApplicable` values

otherwise the behavior could be confusing that setting a prop `canEdit = true` would still return false for this method because it's looking at several additional factors

tjuanitas (commented):
I think you might be able to achieve the desired behavior in EUA using the `editors` prop that's being passed into `MetadataInstanceEditor`

for example, if the instance in the map has `template.templateKey === TEMPLATE_CUSTOM_PROPERTIES`, then you would omit the `cascadePolicy` property since that's not applied for custom instances. similarly, you could set the `canEdit` property based on `isFolderMetadataCascadeEnabled` / the split treatment.

this way when the values get to this `Instance.js` component, the props more accurately represent the desired behavior. e.g. the `canEdit` prop and the method `this.canEdit` should for the most part return the same value

src/features/metadata-instance-editor/Instance.js
```diff
@@ -524,18 +530,41 @@ class Instance extends React.PureComponent<Props, State> {
         return ops;
     }
 
+    /**
+     * Utility function to determine if instance should consider cascading policy
+     *
+     * @return {boolean} true if cascading policy is applicable
+     */
+    isCascadingPolicyApplicable = () => {
```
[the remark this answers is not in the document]
joshmarnold: I agree that it's confusing. This new method is appropriately named, as its purpose is to determine whether cascading applies to the metadata instance. The "prop" was initially given a misleading name, as it was used to enable or disable the cascade option, rather than determining its applicability to the instance.

Maybe I should rename the prop to something like `isCascadingPolicyEnabled`

src/features/metadata-instance-editor/Instance.js
joshmarnold: I went with your suggestion here and ended up utilizing the instance property for more than just canEdit. I'm also adding `isCascadingPolicyApplicable` and `isFolderInstance` here as well

src/features/metadata-instance-editor/Instance.js
```diff
@@ -52,8 +52,9 @@ type Props = {
     hasError: boolean,
     id: string,
     intl: Object,
-    isCascadingPolicyApplicable?: boolean,
+    isCascadingPolicyApplicable: boolean,
```
TylerGauntlett, line 55: I this intentional? `metadata.js` is still using `isCascadingPolicyApplicable?: boolean,`

src/features/metadata-instance-editor/Instance.js
```diff
@@ -590,8 +581,7 @@ class Instance extends React.PureComponent<Props, State> {
             shouldShowCascadeOptions,
             isCascadingOverwritten,
         }: State = this.state;
-        const isProperties = template.templateKey === TEMPLATE_CUSTOM_PROPERTIES;
-        const isEditing = this.isEditing();
```
TylerGauntlett, line 594: I would imagine the performance impact is minor but consider that you are now calling `isEditing()` n times instead of storing the result

src/features/metadata-instance-editor/__tests__/Instance.test.js
```diff
@@ -147,82 +148,149 @@ const userDefinedTemplateField = [
     },
 ];
 
-const intl = { formatMessage: () => {} };
+const intl = { formatMessage: jest.fn().mockReturnValue('Edit Metadata') };
```
tjuanitas, line 151: does this work: `const intl = { formatMessage: message => message.defaultMessage };`?

it would apply to additional message cases

src/features/metadata-instance-editor/__tests__/Instance.test.js
```diff
@@ -147,82 +148,149 @@ const userDefinedTemplateField = [
...
+                        fields,
+                    }}
+                />,
+            );
+
+            const editButton = screen.queryByLabelText('Edit Metadata');
+            if (editButton) {
+                fireEvent.click(editButton);
+            }
+            expect(container).toMatchSnapshot();
```
tjuanitas, line 227: could we switch to schecking if the component is in `TemplateInstance` with `toBeInTheDocument`?

I don't think the snapshots are helping in this case since the file is so large it's hard to verify

src/features/metadata-instance-editor/Instance.js
```diff
@@ -531,7 +521,8 @@ class Instance extends React.PureComponent<Props, State> {
      */
     canEdit(): boolean {
         const { canEdit, onModification, onRemove, onSave }: Props = this.props;
-        return (
+
+        return !!(
```
tjuanitas, line 525: I don't think we need this since each of the conditions between the && will return booleans

src/features/metadata-instance-editor/Instance.js
```diff
@@ -579,7 +570,7 @@ class Instance extends React.PureComponent<Props, State> {
     };
 
     render() {
-        const { cascadePolicy = {}, isDirty, isCascadingPolicyApplicable, isOpen, template }: Props = this.props;
+        const { isDirty, isOpen, template, isFolderInstance }: Props = this.props;
```
tjuanitas, line 573: 🔤 🙏

src/features/metadata-instance-editor/Instance.js
tjuanitas: agree, I think the original is cleaner too

src/features/metadata-instance-editor/Instance.js
```diff
@@ -219,19 +215,16 @@ class Instance extends React.PureComponent<Props, State> {
...
+
+        const cascadePolicyData = {
+            ...cascadePolicy,
+            isEnabled: isFolderInstance ? isCascadingEnabled : false,
+            overwrite: isCascadingOverwritten,
+        };
+
+        const clonedCurrentData = cloneDeep(currentData);
+        const JSONPatch = this.createJSONPatch(clonedCurrentData, originalData);
+        onSave(id, JSONPatch, cascadePolicyData, clonedCurrentData);
```
tjuanitas, line 227: little confused on why this change is needed. in the original implementation, it was only passing `cascadePolicyData` when the policy was applicable. but now it's passing `cascadePolicy` in all cases?

src/features/metadata-instance-editor/Instance.js
```diff
@@ -351,10 +344,10 @@ class Instance extends React.PureComponent<Props, State> {
      * @return {Object} - react title element
      */
     getTitle(): React.Node {
-        const { cascadePolicy = {}, hasError, isCascadingPolicyApplicable, template }: Props = this.props;
+        const { cascadePolicy = {}, hasError, template }: Props = this.props;
         const isProperties = template.templateKey === TEMPLATE_CUSTOM_PROPERTIES;
 
-        const type = isCascadingPolicyApplicable && cascadePolicy.id ? 'cascade' : 'default';
```
tjuanitas, line 357: just to confirm, we're removing `isCascadingPolicyApplicable` because if `cascadePolicy` exists then the former must be true, correct? and thus the former check is redundant?

src/features/metadata-instance-editor/__tests__/Instance.test.js
```diff
@@ -147,82 +148,149 @@ const userDefinedTemplateField = [
...
         const wrapper = shallow(
             <Instance
                 canEdit
                 data={data}
                 dataValue="value"
                 intl={intl}
+                cascadePolicy={{
+                    id: 'hello',
+                }}
+                isCascadingPolicyApplicable
```
tjuanitas, line 293: i see `isFolderInstance` is not added here, does the current implementation allow for the `isCascadingPolicyApplicable` and `isFolderInstance` props to work independently of each other? is that okay?

tjuanitas (commented):
changes were confirmed to work with `File` metadata in the sidebar right?

overall implementation makes sense but left minor comments

tjuanitas:
I also want to make sure the cases here are considered in the testing: https://opensource.box.com/box-ui-elements/#/Features?id=metadatainstanceeditor

i.e. showing the correct error states or messages

src/features/metadata-instance-editor/__tests__/Instance.test.js
joshmarnold: Changed to your suggestion.

src/features/metadata-instance-editor/Instance.js
joshmarnold: Removed the !!

src/features/metadata-instance-editor/__tests__/Instance.test.js
joshmarnold: Updated this to check when the instance is a template and is editing, that the correct toggle is showing

joshmarnold: We check that an instance is template or custom in the test above this one

src/features/metadata-instance-editor/Instance.js
joshmarnold: Relevant EUW service: https://git.dev.box.net/AppServices/end-user-web/blob/25e47e8a54c6a19f666d15e3ddf3ac09c158807a/src/server/controllers/metadata/saveInstances.js

please reference this as I think it will help understand what's going on here

src/features/metadata-instance-editor/__tests__/Instance.test.js
joshmarnold: They're testing that the footer renders properly. Whether or not the footer renders doesn't depend on these variables.

PR #3377, opened 2023-07-27
fix(button): add aria labels for accessibility

dsporysz-box:
Add an optional aria-label property to CloseButton and Button. Note that the "aria-label" property in CloseButton has a default value "close".

### DIFF
--- a/src/components/button/Button.tsx
+++ b/src/components/button/Button.tsx
@@ -11,8 +11,6 @@ export enum ButtonType {
     SUBMIT = 'submit',
 }
 export interface ButtonProps {
-    /** ariaLabel defines a string value that labels the current element */
-    ariaLabel?: string;
     /** Child components for the button, generally localized text */
     children?: React.ReactNode;
     /** Custom class for the button */
@@ -61,7 +59,6 @@ class Button extends React.Component<ButtonProps> {
 
     render() {
         const {
-            ariaLabel,
             children,
             className,
             icon,
@@ -95,7 +92,6 @@ class Button extends React.Component<ButtonProps> {
         let button = (
             // eslint-disable-next-line react/button-has-type
             <button
-                aria-label={ariaLabel}
                 ref={element => {
                     this.btnElement = element;
                     if (setRef) {
--- a/src/components/button/Button.tsx
+++ b/src/components/button/Button.tsx
@@ -11,6 +11,8 @@ export enum ButtonType {
     SUBMIT = 'submit',
 }
 export interface ButtonProps {
+    /** ariaLabel defines a string value that labels the current element */
+    ariaLabel?: string;
     /** Child components for the button, generally localized text */
     children?: React.ReactNode;
     /** Custom class for the button */
@@ -59,6 +61,7 @@ class Button extends React.Component<ButtonProps> {
 
     render() {
         const {
+            ariaLabel,
             children,
             className,
             icon,
@@ -92,6 +95,7 @@ class Button extends React.Component<ButtonProps> {
         let button = (
             // eslint-disable-next-line react/button-has-type
             <button
+                aria-label={ariaLabel}
                 ref={element => {
                     this.btnElement = element;
                     if (setRef) {
--- a/src/components/close-button/CloseButton.js.flow
+++ b/src/components/close-button/CloseButton.js.flow
@@ -9,15 +9,17 @@ import { bdlGray65 } from '../../styles/variables';
 import './CloseButton.scss';
 
 type Props = {
+    /** ariaLabel defines a string value that labels the current element */
+    ariaLabel?: string;
     /** Custom class for the close button */
     className?: string,
     /** onClick handler for the close button */
     onClick?: Function,
 }
 
-const CloseButton = ({ className, onClick }: Props) => {
+const CloseButton = ({ className, onClick, ariaLabel = 'close' }: Props) => {
     return (
-        <Button className={classNames('bdl-CloseButton', className)} onClick={onClick} type="button">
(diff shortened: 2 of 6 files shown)


src/components/button/Button.tsx
```diff
@@ -92,6 +95,7 @@ class Button extends React.Component<ButtonProps> {
         let button = (
             // eslint-disable-next-line react/button-has-type
             <button
+                aria-label={ariaLabel}
```
tjuanitas, line 98: the Button component uses the `rest` functionality to allow additional HTML attributes. this is how other components have been adding aria labels:
https://github.com/box/box-ui-elements/blob/d249a01de01dee59b8f53e64e1b569ddbe4f7c4f/src/components/button/Button.tsx#L75
https://github.com/box/box-ui-elements/blob/d249a01de01dee59b8f53e64e1b569ddbe4f7c4f/src/elements/common/sub-header/SortButton.js#L28
https://github.com/box/box-ui-elements/blob/d249a01de01dee59b8f53e64e1b569ddbe4f7c4f/src/elements/content-sidebar/SidebarNavSignButton.tsx#L59

are we able to continue using this pattern? any specific reason for the explicit prop now?

src/components/close-button/CloseButton.js.flow
```diff
@@ -9,15 +9,17 @@ import { bdlGray65 } from '../../styles/variables';
...
+    /** ariaLabel defines a string value that labels the current element */
+    ariaLabel?: string;
     /** Custom class for the close button */
     className?: string,
     /** onClick handler for the close button */
     onClick?: Function,
 }
 
-const CloseButton = ({ className, onClick }: Props) => {
+const CloseButton = ({ className, onClick, ariaLabel = 'close' }: Props) => {
```
tjuanitas, line 20: the `Close` message should be translated since this is read by screen readers

tjuanitas, line 20: nit: these props are sorted alphabetically

src/components/button/Button.tsx
dsporysz-box: Thank you for showing me examples. I made changes to use the `rest` functionality.

src/components/close-button/CloseButton.js.flow
dsporysz-box: I've removed the hardcoded value and I used the `IntlShape` for close message translation. Thanks a lot!

dsporysz-box: It's been removed. Thanks!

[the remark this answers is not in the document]
dsporysz-box: I've found a proper close message for `CloseButton` in `common/messages.js`. Now the label is properly translated and I've removed `aria-label` from props.

src/components/close-button/CloseButton.tsx
```diff
@@ -1,22 +1,29 @@
...
 import IconClose from '../../icons/general/IconClose';
 import { bdlGray65 } from '../../styles/variables';
 
+// @ts-ignore flow import
+import messages from '../../common/messages';
+
 import './CloseButton.scss';
 
 export interface CloseButtonProps {
+    intl: IntlShape;
```
tjuanitas, line 16: nit: this list is sorted alphabetically

src/components/close-button/CloseButton.tsx
```diff
@@ -1,22 +1,29 @@
...
     /** Custom class for the close button */
     className?: string;
     /** onClick handler for the close button */
     onClick?: Function;
 }
 
-const CloseButton = ({ className, onClick }: CloseButtonProps) => {
+const CloseButton = ({ intl, className, onClick }: CloseButtonProps) => {
```
tjuanitas, line 23: nit: can we sort these?

src/components/close-button/__tests__/CloseButton.test.tsx
```diff
@@ -37,4 +37,12 @@ describe('components/close-button/CloseButton', () => {
             expect(handleClick).toHaveBeenCalledTimes(1);
         });
     });
+
+    describe('accesability properties', () => {
+        test('should have an accesiblity label', () => {
```
tjuanitas, line 42: could we move this into the `render()` describe above? this test probably doesn't need to be in its own `describe`

tjuanitas, line 42:
```suggestion
        test('should have an accessibility label', () => {
```

src/components/close-button/CloseButton.tsx
```diff
@@ -27,4 +34,5 @@ const CloseButton = ({ className, onClick }: CloseButtonProps) => {
     );
 };
 
-export default CloseButton;
+export { CloseButton as CloseButtonBase };
```
tjuanitas, line 37: usually we do this export so that we can test the component without the `injectIntl` HOC. do we need to update the import in the test?

dsporysz-box: I removed unused export, leaving just the one with injectIntl.

src/components/close-button/__tests__/CloseButton.test.tsx
dsporysz-box: I moved it.

src/components/close-button/CloseButton.tsx
dsporysz-box: Sorted.

src/components/close-button/CloseButton.tsx
dsporysz-box: Sorted.

tjuanitas (approved):
lgtm thank you for the changes!

benjamin-shen (approved):

PR #3350, opened 2023-06-05
fix: update dist path and react version for elements test pages

tjuanitas:
changes include but not limited to:

- add `dist` to source path
- update react versions to `17.0.1`
- fix page titles
- remove values in `input` elements

PR #3378, opened 2023-08-04
fix: update source paths for test pages

tjuanitas:
I made a mistake by adding `dist` to the relative paths for the test pages in this PR: https://github.com/box/box-ui-elements/pull/3350

These files are intended to be tested with the `yarn start:dev` command and adding `dist` to the paths will cause issues because the sources will not be found.

<sup>i.e. everything breaks :(</sup>

### DIFF
--- a/src/api/Metadata.js
+++ b/src/api/Metadata.js
@@ -128,7 +128,7 @@ class Metadata extends File {
     /**
      * Returns the custom properties template
      *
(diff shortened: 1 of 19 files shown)


greg-in-a-box (approved):

jpan-box (approved):

bfoxx1906 (approved):
lgtm

PR #3379, opened 2023-08-04
fix(accessibility): skip uploadsManager in voiceover if not visible

bkepka-box:
Without this change voiceover is detecting uploadManager even if it is not visible on the screen.

### DIFF
--- a/src/elements/content-uploader/UploadsManager.scss
+++ b/src/elements/content-uploader/UploadsManager.scss
@@ -6,6 +6,7 @@
     width: 100%;
     max-height: 0;
     transition: max-height .5s;
+    display: none;
 
     .bcu-progress-container {
         margin-right: 0;
@@ -23,6 +24,7 @@
 
     &.bcu-is-visible {
         max-height: 60px;
+        display: initial;
     }
 
     .bcu-uploads-manager-item-list {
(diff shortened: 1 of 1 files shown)


joshmarnold (approved):

src/elements/content-uploader/UploadsManager.scss
```diff
@@ -3,6 +3,7 @@
 .be.bcu-uploads-manager-container {
     position: fixed;
     bottom: 0;
+    display: none;
     width: 100%;
     max-height: 0;
     transition: max-height .5s;
```
tjuanitas, line 9: there's a `transition` on this class, does it still work? I think there are conflicts when trying to animate and update the `display` property at the same time (I think some issues with `visibility` too)

bkepka-box: @tjuanitas It still works for me on Chrome and Safari

https://github.com/box/box-ui-elements/assets/141409011/bd384bcc-ed16-4b11-9008-4ced47d52f3d

tjuanitas: that transition occurs when toggling `bcu-is-expanded` which happens after `display: initial` is already applied. you'll want to test toggling `bcu-is-visible` which handles the transition between `display: none` -> `display: initial`. this will happen on drag-or-drop or when a file is uploaded

https://github.com/box/box-ui-elements/assets/7311041/c3e53f70-e1f8-4afc-8b53-b9677e0798de

PR #3380, opened 2023-08-04
fix(modal-dialog): add aria-modal to ModalDialog

dsporysz-box:
It's possible to escape a modal with Accessibility Tools and a screen reader by using Control + Option + Arrow Keys.

This PR adds `aria-modal="true"` prop to ModalDialog div.


tjuanitas (commented):
makes sense!

could you add line to this test? https://github.com/box/box-ui-elements/blob/baba4337f71a09b226cda52792115e9e387896c4/src/components/modal/__tests__/ModalDialog.test.js#L39-L40

dsporysz-box:
I've updated the aria props test.

arturfrombox (approved):

tjuanitas (approved):

PR #3383, opened 2023-08-11
fix: hide uploads manager progress bar for screen-readers if not visible

loonskai:
This PR fixes the issue of `OverallUploadsProgressBar` being accessible for screen readers when it's visually hidden

### DIFF
--- a/src/elements/content-uploader/OverallUploadsProgressBar.js
+++ b/src/elements/content-uploader/OverallUploadsProgressBar.js
@@ -88,13 +88,13 @@ const OverallUploadsProgressBar = ({
 
     return (
         <div
+            aria-hidden={!isVisible}
             className="bcu-overall-progress-bar"
             data-resin-target={isExpanded ? 'uploadcollapse' : 'uploadexpand'}
             onClick={onClick}
             onKeyDown={onKeyDown}
             role="button"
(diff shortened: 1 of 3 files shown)


src/elements/content-uploader/OverallUploadsProgressBar.js
```diff
@@ -94,6 +94,7 @@ const OverallUploadsProgressBar = ({
             onKeyDown={onKeyDown}
             role="button"
             tabIndex={isVisible ? '0' : '-1'}
+            aria-hidden={!isVisible}
```
dsporysz-box, line 97: nit: props should be sorted alphabetically

loonskai: I'm wondering if we should enable `react/jsx-sort-props` rule in this case? This is easy to miss if checked manually

dsporysz-box: I like the idea but I don't what effect it's going to have on BUIE users. I think it's a topic for another issue/PR. 😄

dsporysz-box (dismissed):
LGTM

dsporysz-box (approved):

JChan106 (approved):

PR #3384, opened 2023-08-11
chore: enable props alphabetic order

loonskai:
This PR suggests enabling `react/jsx-sort-props` ESLint rule in the project. Based on the observation from the previous PRs and common practice to define props in alphabetical order for better readability, enabling this rule should take off an additional friction with respect to code style


loonskai:
Closed as this rule is incompatible with TS file extensions.

PR #3385, opened 2023-08-11
feat(uaa-integration): Parse Versions Response from UAA

JChan106:
The UAA response gives a Versions activity event that is different than the v2 Versions API that we are currently using. We must parse the response and add the needed fields that our UI components need to render Version events. 

One big difference with this change is that the latest version events do not have `promotion` context, unlike the v2 API. So if the user has promoted a version and that version is the most recent version, they can only see that version was promoted by clicking the info button and looking at the version history. The team has had discussions around this and is okay with this tradeoff, we will look into adding the promotion context into the UAA response in the future.

UAA:
Multiple version events
image
Multiple upload collaborators
image
Version events with comments between
image
new file upload, no events
image

V2 Versions:
Multiple version events
image
Multiple upload collaborators
image
Version events with comments between
image
new file upload, no events
image

### DIFF
--- a/src/api/Feed.js
+++ b/src/api/Feed.js
@@ -190,7 +190,6 @@ export const getParsedFileActivitiesResponse = (response?: { entries: FileActivi
                 }
 
                 case FILE_ACTIVITY_TYPE_VERSION: {
-                    // $FlowFixMe
                     const versionsItem = { ...source[FILE_ACTIVITY_TYPE_VERSION] };
 
                     versionsItem.type = 'file_version';
@@ -224,30 +223,24 @@ export const getParsedFileActivitiesResponse = (response?: { entries: FileActivi
                             versionsItem.start?.created_at &&
                             versionsItem.start?.created_by
                         ) {
-                            const modifiedAt = versionsItem.start.created_at;
-                            const modifiedBy = { ...versionsItem.start.created_by };
-                            versionsItem.modified_at = modifiedAt;
-                            versionsItem.modified_by = modifiedBy;
+                            versionsItem.modified_at = versionsItem.start.created_at;
+                            versionsItem.modified_by = { ...versionsItem.start.created_by };
                         }
                         if (
                             versionsItem.action_type === 'trashed' &&
                             versionsItem.start?.trashed_at &&
                             versionsItem.start?.trashed_by
                         ) {
-                            const trashedAt = versionsItem.start.trashed_at;
-                            const trashedBy = { ...versionsItem.start.trashed_by };
-                            versionsItem.trashed_at = trashedAt;
-                            versionsItem.trashed_by = trashedBy;
+                            versionsItem.trashed_at = versionsItem.start.trashed_at;
+                            versionsItem.trashed_by = { ...versionsItem.start.trashed_by };
                         }
                         if (
                             versionsItem.action_type === 'restored' &&
                             versionsItem.start?.restored_at &&
                             versionsItem.start?.restored_by
                         ) {
-                            const restoredAt = versionsItem.start.restored_at;
-                            const restoredBy = { ...versionsItem.start.restored_by };
-                            versionsItem.restored_at = restoredAt;
-                            versionsItem.restored_by = restoredBy;
+                            versionsItem.restored_at = versionsItem.start.restored_at;
+                            versionsItem.restored_by = { ...versionsItem.start.restored_by };
                         }
                     }
 
--- a/src/api/Feed.js
+++ b/src/api/Feed.js
@@ -118,7 +118,6 @@ export const getParsedFileActivitiesResponse = (response?: { entries: FileActivi
     }
 
     const data = response.entries;
-    const versions = [];
 
     const parsedData: Array<Object> = data
         .map(item => {
@@ -196,6 +195,10 @@ export const getParsedFileActivitiesResponse = (response?: { entries: FileActivi
                     if (versionsItem.action_by) {
                         const collaborators = {};
 
+                        if (versionsItem.action_by.length === 1) {
+                            versionsItem.uploader_display_name = versionsItem.action_by[0].name;
+                        }
+
                         versionsItem.action_by.map(collaborator => {
                             collaborators[collaborator.id] = { ...collaborator };
                             return collaborator;
@@ -211,8 +214,42 @@ export const getParsedFileActivitiesResponse = (response?: { entries: FileActivi
                         versionsItem.version_start = versionsItem.start.number;
                     }
 
-                    versions.push(versionsItem);
-                    return null;
+                    if (versionsItem.version_start === versionsItem.version_end) {
+                        versionsItem.version_number = versionsItem.version_start;
+
+                        if (
+                            versionsItem.action_type === 'created' &&
+                            versionsItem.start?.created_at &&
+                            versionsItem.start?.created_by
+                        ) {
+                            const modifiedAt = versionsItem.start.created_at;
+                            const modifiedBy = { ...versionsItem.start.created_by };
+                            versionsItem.modified_at = modifiedAt;
+                            versionsItem.modified_by = modifiedBy;
+                        }
+                        if (
+                            versionsItem.action_type === 'trashed' &&
+                            versionsItem.start?.trashed_at &&
+                            versionsItem.start?.trashed_by
+                        ) {
+                            const trashedAt = versionsItem.start.trashed_at;
+                            const trashedBy = { ...versionsItem.start.trashed_by };
+                            versionsItem.trashed_at = trashedAt;
+                            versionsItem.trashed_by = trashedBy;
+                        }
+                        if (
+                            versionsItem.action_type === 'restored' &&
+                            versionsItem.start?.restored_at &&
+                            versionsItem.start?.restored_by
+                        ) {
+                            const restoredAt = versionsItem.start.restored_at;
+                            const restoredBy = { ...versionsItem.start.restored_by };
+                            versionsItem.restored_at = restoredAt;
+                            versionsItem.restored_by = restoredBy;
+                        }
+                    }
+
+                    return versionsItem;
                 }
 
                 default: {
@@ -220,9 +257,10 @@ export const getParsedFileActivitiesResponse = (response?: { entries: FileActivi
                 }
             }
         })
-        .filter(item => !!item);
+        .filter(item => !!item)
+        .reverse();
 
-    return [...versions, ...parsedData];
+    return parsedData;
 };
 
 class Feed extends Base {
--- a/src/api/Feed.js
+++ b/src/api/Feed.js
@@ -114,7 +114,7 @@ const parseReplies = (replies: Comment[]): Comment[] => {
 
 export const getParsedFileActivitiesResponse = (response?: { entries: FileActivity[] }) => {
     if (!response || !response.entries || !response.entries.length) {
-        return { entries: [] };
+        return [];
     }
 
     const data = response.entries;
--- a/src/api/Feed.js
+++ b/src/api/Feed.js
@@ -190,8 +190,10 @@ export const getParsedFileActivitiesResponse = (response?: { entries: FileActivi
                 }
 
                 case FILE_ACTIVITY_TYPE_VERSION: {
+                    // $FlowFixMe
                     const versionsItem = { ...source[FILE_ACTIVITY_TYPE_VERSION] };
 
+                    versionsItem.type = 'file_version';
                     if (versionsItem.action_by) {
                         const collaborators = {};
 
--- a/src/api/Feed.js
+++ b/src/api/Feed.js
@@ -33,6 +33,7 @@ import {
     FILE_ACTIVITY_TYPE_APP_ACTIVITY,
     FILE_ACTIVITY_TYPE_COMMENT,
     FILE_ACTIVITY_TYPE_TASK,
+    FILE_ACTIVITY_TYPE_VERSION,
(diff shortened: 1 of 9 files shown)


src/elements/content-sidebar/activity-feed/version/CollapsedVersion.js
```diff
@@ -83,9 +84,7 @@ const CollapsedVersion = (props: Props): React.Node => {
                         aria-label={intl.formatMessage(messages.getVersionInfo)}
                         className="bcs-Version-info"
                         data-resin-target={ACTIVITY_TARGETS.VERSION_CARD}
-                        onClick={() => {
-                            onInfo({ versions });
```
JChan106, line 87: What's passed into this cb is only used when `hasVersions` prop is `false` and `onVersionHistoryClick` prop is passed into BUIE. There doesn't seem to be a standardized format, since the CollapsedVersion and Version components pass back different data. Because UAA doesn't have the `versions` array like the v2 api, I decided to use the end version of the event and match what was passed in the `Version.js` component (`id` and `version_number`). This cb is not used in EUA.

src/api/Feed.js
```diff
@@ -184,14 +185,80 @@ export const getParsedFileActivitiesResponse = (response?: { entries: FileActivi
...
+                            versionsItem.start?.restored_by
+                        ) {
+                            const restoredAt = versionsItem.start.restored_at;
+                            const restoredBy = { ...versionsItem.start.restored_by };
+                            versionsItem.restored_at = restoredAt;
+                            versionsItem.restored_by = restoredBy;
+                        }
+                    }
+
+                    return versionsItem;
```
patlm, line 250: I was wondering if there would be a way to extract and organize some of this into separate functions to help reduce the length of the `getParsedFileActivitiesResponse` function? I'm also fine with leaving as-is if you don't think that would make sense here.

JChan106: Yeah, thought about it, there isn't really benefits to splitting though since the logic isn't reusable, and the switch statements make it more readable as well.

src/elements/content-sidebar/activity-feed/activity-feed/__tests__/ActivityFeed.test.js
```diff
@@ -148,6 +148,15 @@ describe('elements/content-sidebar/ActivityFeed/activity-feed/ActivityFeed', ()
         expect(wrapper.find('EmptyState').exists()).toBe(true);
     });
 
+    test('should not render empty state when UAA is enabled and there is 1 version (current version from file) and the start and end version are the same', () => {
```
patlm, line 151: Was this meant to be "should render empty.." instead of "should not render"?

JChan106: Good catch

patlm (dismissed):
Looks good to me!

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
```diff
@@ -275,12 +277,12 @@ const ActiveState = ({
                     case FEED_ITEM_TYPE_VERSION:
                         return (
                             <ActivityItem key={item.type + item.id} className="bcs-version-item" data-testid="version">
-                                {item.versions ? (
+                                {hasMultipleVersions(item) ? (
                                     // $FlowFixMe
-                                    <CollapsedVersion {...item} onInfo={onVersionInfo} />
+                                    <CollapsedVersion {...item} onInfo={onVersionInfo} shouldUseUAA={shouldUseUAA} />
```
mickr, line 282: Do we need the `shouldUseUAA` prop here? Could this also be handled in the `onVersionInfo` function that is passed to `onInfo`?

src/api/Feed.js
```diff
@@ -184,14 +185,80 @@ export const getParsedFileActivitiesResponse = (response?: { entries: FileActivi
...
+
+                    if (versionsItem.version_start === versionsItem.version_end) {
+                        versionsItem.version_number = versionsItem.version_start;
+
+                        if (
+                            versionsItem.action_type === 'created' &&
+                            versionsItem.start?.created_at &&
+                            versionsItem.start?.created_by
+                        ) {
+                            const modifiedAt = versionsItem.start.created_at;
```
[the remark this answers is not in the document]
JChan106: Not really, can change these to inline to shorten the logic

src/api/Feed.js
```diff
@@ -184,14 +185,80 @@ export const getParsedFileActivitiesResponse = (response?: { entries: FileActivi
                     return { ...source[FILE_ACTIVITY_TYPE_APP_ACTIVITY] };
                 }
 
+                case FILE_ACTIVITY_TYPE_VERSION: {
+                    // $FlowFixMe
```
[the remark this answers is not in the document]
JChan106: It was creating a lot of errors, because its not expecting a versions item with these fields from UAA. I was able to add some of the fields from UAA into the version type, so I'll remove this now.

src/elements/content-sidebar/activity-feed/activity-feed/ActiveState.js
JChan106: Could be, but this will be removed once the FF is gone, so the parameters passed in can stay in the CollapsedVersion component.

mickr (dismissed):

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
```diff
@@ -168,11 +168,20 @@ class ActivityFeed extends React.Component<Props, State> {
...
-        return feedItems.length === 0 || (feedItems.length === 1 && feedItems[0].type === ItemTypes.fileVersion);
+
+        return (
+            feedItems.length === 0 ||
+            (!shouldUseUAA && feedItems.length === 1 && feedItems[0].type === ItemTypes.fileVersion) ||
+            (!!shouldUseUAA &&
+                feedItems.length === 1 &&
+                feedItems[0].type === ItemTypes.fileVersion &&
+                feedItems[0].version_start === 1 &&
+                feedItems[0].version_end === 1)
```
tjuanitas, line 183: why are these needed but the previous implementation didn't check `feedItems[0].versions`?

src/api/__tests__/Feed.test.js
```diff
@@ -15,6 +15,7 @@ import {
     FILE_ACTIVITY_TYPE_TASK,
     IS_ERROR_DISPLAYED,
     TASK_MAX_GROUP_ASSIGNEES,
+    FILE_ACTIVITY_TYPE_VERSION,
```
tjuanitas, line 18: 🔤

src/api/Feed.js
```diff
@@ -184,14 +185,73 @@ export const getParsedFileActivitiesResponse = (response?: { entries: FileActivi
...
                     return null;
                 }
             }
         })
-        .filter(item => !!item);
+        .filter(item => !!item)
+        .reverse();
 
-    return { entries: parsedData };
+    return parsedData;
```
tjuanitas, line 254: you could move the `return` to the `map` directly instead of needing the variable

src/api/Feed.js
```diff
@@ -184,14 +185,73 @@ export const getParsedFileActivitiesResponse = (response?: { entries: FileActivi
                     return { ...source[FILE_ACTIVITY_TYPE_APP_ACTIVITY] };
                 }
 
+                case FILE_ACTIVITY_TYPE_VERSION: {
+                    const versionsItem = { ...source[FILE_ACTIVITY_TYPE_VERSION] };
+
+                    versionsItem.type = 'file_version';
```
tjuanitas, line 191: you can move this to inside the declaration of the object. unless it's cleaner to keep it outside so it remains consistent with the other cases

src/api/Feed.js
```diff
@@ -184,14 +185,73 @@ export const getParsedFileActivitiesResponse = (response?: { entries: FileActivi
...
+                    if (versionsItem.action_by) {
+                        const collaborators = {};
+
+                        if (versionsItem.action_by.length === 1) {
+                            versionsItem.uploader_display_name = versionsItem.action_by[0].name;
+                        }
+
+                        versionsItem.action_by.map(collaborator => {
```
tjuanitas, line 199: it's a little odd to use `map` without intending to use the returned array. I think this might be a good use case for `reduce`:

```
versionsItem.collaborators = versionsItem.action_by.reduce( ... , {});
```

src/api/Feed.js
```diff
@@ -184,14 +185,73 @@ export const getParsedFileActivitiesResponse = (response?: { entries: FileActivi
...
+                            collaborators[collaborator.id] = { ...collaborator };
+                            return collaborator;
+                        });
+
+                        versionsItem.collaborators = collaborators;
+                    }
+                    if (versionsItem.end?.number !== undefined) {
```
tjuanitas, line 206: is this okay?
```suggestion
                    if (versionsItem.end?.number) {
```
same for `versionsItem.start`

src/api/Feed.js
```diff
@@ -184,14 +185,73 @@ export const getParsedFileActivitiesResponse = (response?: { entries: FileActivi
...
+                        versionsItem.version_end = versionsItem.end.number;
+                        versionsItem.id = versionsItem.end.id;
+                    }
+                    if (versionsItem.start?.number !== undefined) {
+                        versionsItem.version_start = versionsItem.start.number;
+                    }
+
+                    if (versionsItem.version_start === versionsItem.version_end) {
+                        versionsItem.version_number = versionsItem.version_start;
```
tjuanitas, line 215: what's `version_number` used for and is it set when `versionsItem.version_start !== versionsItem.version_end`?

src/api/Feed.js
tjuanitas, line 191: constants for this and `action_type`s below?

src/elements/content-sidebar/activity-feed/activity-feed/__tests__/ActivityFeed.test.js
```diff
@@ -148,6 +148,24 @@ describe('elements/content-sidebar/ActivityFeed/activity-feed/ActivityFeed', ()
...
+    test('should render empty state when UAA is enabled and there is 1 version (current version from file) and the start and end version are the same', () => {
+        const wrapper = getWrapper({
+            currentUser,
+            feedItems: [{ ...first_version, version_start: 1, version_end: 1 }],
+            shouldUseUAA: true,
+        });
+        expect(wrapper.find('EmptyState').exists()).toBe(true);
+    });
+
+    test('should not render empty state when UAA is enabled and there is 1 version (current version from file) and the start and end version are different', () => {
```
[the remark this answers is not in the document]
JChan106: Ah meant 1 version event, will update

src/api/Feed.js
JChan106: version_number is used when there is an event with a single version (`Version.js`). It is not set when start !== end, if start === end, we would use `CollapsedVersion.js`. Which is a range of versions

src/api/Feed.js
JChan106: Yeah it should be okay, version number shouldn't be 0

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
JChan106: It only had to check the length === 1 because multiple versions would be multiple events. With UAA, you can have multiple versions on a single event

src/api/Feed.js
JChan106: I prefer having the variable for the typing and debugging tradeoffs

src/elements/content-sidebar/activity-feed/version/CollapsedVersion.js
JChan106: yup

src/api/Feed.js
JChan106: Leaning for keeping it for now since we've used `map` in a few other place. Going through another round of manual testing for the logic change also is a big time sink.

src/api/Feed.js
JChan106: I will probably keep it consistent

src/api/Feed.js
tjuanitas: semantically it feels a little wrong to use a `map` in this case over something like a `forEach` since the goal is to populate a separate object instead of using the returned array

with the reduce it could be:
```
versionsItem.collaborators = versionsItem.action_by.reduce((collaborators, collaborator) =>
    ({ ...collaborators, [collaborator.id]: { ...collaborator } }), {});
```

although you could leave it as-is if the current implementation is easier to read

src/constants.js
```diff
@@ -471,6 +471,11 @@ export const ACTIVITY_FILTER_OPTION_RESOLVED: 'resolved' = 'resolved';
 export const ACTIVITY_FILTER_OPTION_TASKS: 'tasks' = 'tasks';
 export const ACTIVITY_FILTER_OPTION_UNRESOLVED: 'open' = 'open';
 
+/* ------------------ File Activity Action Types ----------- */
+export const ACTION_TYPE_CREATED: 'created' = 'created';
```
tjuanitas, line 475: I'm not sure if there's a concept of "action type" for other features but if there was we could change to `FILE_ACTIVITY_ACTION_TYPE_[...]` to be more specific

tjuanitas (approved):
lgtm!

mickr (approved):

patlm (approved):

PR #3386, opened 2023-08-16
feat(uaa-integration): Parse App Activity item

JChan106:
The UAA response gives a app activity event that is different than the v2 app activity API that we are currently using. We must parse the response and add the needed fields that our UI components need to render app activity events properly.

### DIFF
--- a/src/api/Feed.js
+++ b/src/api/Feed.js
@@ -181,7 +181,11 @@ export const getParsedFileActivitiesResponse = (response?: { entries: FileActivi
                     return annotationItem;
                 }
                 case FILE_ACTIVITY_TYPE_APP_ACTIVITY: {
-                    return { ...source[FILE_ACTIVITY_TYPE_APP_ACTIVITY] };
+                    const appActivityItem = { ...source[FILE_ACTIVITY_TYPE_APP_ACTIVITY] };
+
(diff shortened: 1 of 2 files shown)


src/api/Feed.js
```diff
@@ -181,7 +181,11 @@ export const getParsedFileActivitiesResponse = (response?: { entries: FileActivi
                     return annotationItem;
                 }
                 case FILE_ACTIVITY_TYPE_APP_ACTIVITY: {
-                    return { ...source[FILE_ACTIVITY_TYPE_APP_ACTIVITY] };
+                    const appActivityItem = { ...source[FILE_ACTIVITY_TYPE_APP_ACTIVITY] };
+
+                    appActivityItem.created_at = appActivityItem.occurred_at;
```
patlm, line 186: In the API schema, it looks like the schema shows a created_at and an occurred_at property. What is the reasoning behind setting created_at and occurred_at to the same value? Does one of them not have a value?

patlm (approved):

src/api/Feed.js
JChan106: created_at does not have a value in file_activities

karelee7 (approved):

PR #3387, opened 2023-08-16
fix(modify-reply): Map object to params

patlm:
* Fix bug with modifying a reply to a comment by mapping the single param to the multiple params used in props above

Notes
* See below for help tracing props
* `Comment.js and BaseComment.js` <- `ActivityThreadReplies.js` <- `ActivityThread.js`
* The props is continued to passed but ActivityThread shows the definition for the onEdit function params expected in components above it.

Issue
* The issue was introduced in this commit: https://github.com/box/box-ui-elements/commit/4f65525e0b502b809332717e537ca075f7a40321
* Before: https://cloud.box.com/s/pmgbtcg04fqjnvf38yok5dmk3wp3z6b2
* After: https://cloud.box.com/s/bwakotinhvlveplrosnpztpjzdz74ju1

### DIFF
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActivityThreadReplies.js
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActivityThreadReplies.js
@@ -1,5 +1,6 @@
 // @flow
 import React from 'react';
+import noop from 'lodash/noop';
 import Comment from '../comment';
 import LoadingIndicator from '../../../../components/loading-indicator';
 import { BaseComment } from '../comment/BaseComment';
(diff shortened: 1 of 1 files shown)


JChan106 (commented):
Could you link what commit this bug came from, and can you also post examples of the before and after of this change?

JChan106:
Looks like theres an issue with the double active states?

patlm:
I would guess the double active state is being caused by trying to use the Canvas TR experience inside the Preview sidebar.

However, I did switch to master branch to check if the double active state was already existing. I found that it is already occurring on the main branch as shown in [this video](https://cloud.box.com/s/2jd782zt8hfskn2kt922fe3mmoloanzw). The issue wasn't showing up in Canvas on staging so seems like just issue with it being put in Preview sidebar.

patlm:
I think there is some confusion. We do have threaded replies in the preview sidebar (that is what is shown in the videos). There are just styling differences and so the styling in the video doesn't accurately reflect Canvas. Because the issue isn't related to the styling, the tests are still valid. If you compare the videos below, you can see before/after the fix and see there is not any regression in styling and that the issue is resolved.
* Before: https://cloud.box.com/s/2jd782zt8hfskn2kt922fe3mmoloanzw
* After: https://cloud.box.com/s/bwakotinhvlveplrosnpztpjzdz74ju1

I will connect with you on slack about Canvas setup.

JChan106 (approved):

karelee7 (approved):

PR #3388, opened 2023-08-17
fix(flyout): set default role for Flyout to 'dialog'

loonskai:
~This PR extends `Flyout` component with `overlayRole` prop in order to pass a custom role for popup content. It includes updates for both overlay `role` attribute as well as `aria-haspopup` attribute for button container.~
This PR fixes `Flyout` container component attribute `aria-haspopup` to `dialog` to match the role of the overlay component.

### DIFF
--- a/src/components/flyout/Flyout.js
+++ b/src/components/flyout/Flyout.js
@@ -54,6 +54,13 @@ const positions = {
 };
 
 /**
+ * The set of container's aria-haspopup attribute values which are compatible with popup's role attribute according to MDN: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-haspopup
+ */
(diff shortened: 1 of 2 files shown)


tjuanitas (commented):
are there use cases for the `Flyout` component with a different role than `dialog`? i want to make sure developers don't misuse the component

loonskai:
> are there use cases for the `Flyout` component with a role different than `dialog`? i want to confirm developers don't misuse the component by adding non-applicable roles instead of using a more specialized component

You're right, overlay components already assume the role to be `dialog` only so my last changes are redundant a bit, applied the fix

tjuanitas (approved):
LGTM, thank you Siarhei!

dsporysz-box (approved):

PR #3394, opened 2023-09-01
feat(uaa-integration): Handle no comment permission

JChan106:
When comment permission is false, we don't want to block the file_activities call, so we exclude `comments` from the parameters.

### DIFF
--- a/src/api/Feed.js
+++ b/src/api/Feed.js
@@ -41,6 +41,7 @@ import {
     HTTP_STATUS_CODE_CONFLICT,
     IS_ERROR_DISPLAYED,
     PERMISSION_CAN_VIEW_ANNOTATIONS,
+    PERMISSION_CAN_COMMENT,
     TASK_NEW_APPROVED,
     TASK_NEW_COMPLETED,
     TASK_NEW_REJECTED,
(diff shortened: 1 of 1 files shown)


src/api/Feed.js
```diff
@@ -595,17 +596,20 @@ class Feed extends Base {
...
             ...taskActivityType,
             ...versionsActivityType,
         ];
 
-        const fileActivitiesPromise = shouldUseUAA
-            ? this.fetchFileActivities(permissions, filteredActivityTypes, shouldShowReplies)
-            : Promise.resolve();
+        const fileActivitiesPromise =
+            // Only fetch when activity types are explicitly stated
+            shouldUseUAA && filteredActivityTypes.length
```
bfoxx1906, line 610: Are we checking the length because it's actually possible to get to this part of the code and the filteredActivityTypes array is empty?

JChan106: Yes, if all ff are off and the comments are disabled, it's possible to have an empty array. The api handles that as fetching all types, which is what we don't want

patlm (approved):

bfoxx1906 (approved):

PR #3397, opened 2023-09-08
fix(preview): Change focus indicator for actions in version panel

bkepka-box:
Change focus indicator of actions button in version panel because the original one is difficult to be seen
Before:
image
After:
image

### DIFF
--- a/src/elements/content-sidebar/versions/VersionsItem.scss
+++ b/src/elements/content-sidebar/versions/VersionsItem.scss
@@ -1,4 +1,5 @@
 @import '../../common/variables';
+@import '../../../styles/constants/buttons';
 
 .bcs-VersionsItem {
     $bcs-VersionsItemButton-horizontal: 10px;
@@ -24,7 +25,7 @@
 
         &:focus {
             border: $bdl-btn-border-width solid $primary-color;
-            box-shadow: inset 0 0 0 1px fade-out($white, .2), 0 1px 2px fade-out($black, .9);
+            box-shadow: $bdl-btn-primary-box-shadow;
         }
     }
 }
--- a/src/elements/content-sidebar/versions/VersionsItem.scss
+++ b/src/elements/content-sidebar/versions/VersionsItem.scss
@@ -22,11 +22,9 @@
         top: $bcs-VersionsItemButton-vertical;
         right: $bcs-VersionsItemButton-horizontal;
 
-        &:not(.is-disabled) {
-            &:focus {
-                border: $bdl-btn-border-width solid $primary-color;
-                box-shadow: inset 0 0 0 1px fade-out($white, .2), 0 1px 2px fade-out($black, .9);
-            }
+        &:focus {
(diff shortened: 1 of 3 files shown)


jcamden (approved):
Looks good.

src/elements/content-sidebar/versions/VersionsItem.scss
```diff
@@ -21,6 +21,13 @@
         position: absolute;
         top: $bcs-VersionsItemButton-vertical;
         right: $bcs-VersionsItemButton-horizontal;
+
+        &:not(.is-disabled) {
```
tjuanitas, line 25: small nit but neither `bcs-VersionsItemActions-toggle` or the `PlainButton` uses the `is-disabled` class. that's set for the regular Button component

karelee7 (approved):
Looks good! Minor: May also be a good idea to include a short description of the change in the PR description with the image, and also move the "Before" picture to be in the description.

src/elements/content-sidebar/versions/VersionsItem.scss
```diff
@@ -21,6 +21,11 @@
         position: absolute;
         top: $bcs-VersionsItemButton-vertical;
         right: $bcs-VersionsItemButton-horizontal;
+
+        &:focus {
+            border: $bdl-btn-border-width solid $primary-color;
+            box-shadow: inset 0 0 0 1px fade-out($white, .2), 0 1px 2px fade-out($black, .9);
```
bfoxx1906, line 27: Is this where you copied this from https://github.com/box/box-ui-elements/blob/4d71c6279313b24e9529c9f8b769348453a57226/src/styles/common/_buttons.scss#L179

mickr (dismissed):

src/elements/content-sidebar/versions/VersionsItem.scss
bfoxx1906: So I'm guessing that those values are supposed to be consistent i.e. if one changes the other should change as well? If that's the case I'd recommend creating a mixin for this and use @include in both places. If they don't have to be the same based on our overall design scheme then we can just leave it alone. Thoughts?

bkepka-box: The main problem was that this button has "a low visibility focus indicator." I think it doesn't need to be exactly the same as primary button one but I can move it to the mixin to keep consistency.

tjuanitas (approved):
lgtm

src/elements/content-sidebar/versions/VersionsItem.scss
```diff
@@ -1,4 +1,5 @@
 @import '../../common/variables';
+@import '../../../styles/constants/buttons';
```
tjuanitas, line 2: nit:
an explicit import isn't needed since the main `variables` file will import this:
https://github.com/box/box-ui-elements/blob/master/src/styles/_variables.scss#L7

PR #3402, opened 2023-09-19
feat(threaded-replies): delete threaded reply comments

TylerGauntlett:
### What This Does

The option for deleting a thread reply is currently selectable from the dropdown, but this action does not do anything. The goal of this PR is to connect this dropdown option with the existing delete code

### High Level

`onDelete` within `BaseComment` is called to trigger the confirmation modal and ultimately delete a comment. In order to recycle this logic, I added a new `onReplyDelete` prop to the `BaseComment` class. This is then passed to the `Replies` component, where it can be passed back into `BaseComment` as `onDelete`. This pattern follows other reply actions (see `onReplyCreate`).

### Demo

https://github.com/box/box-ui-elements/assets/13861069/364035b9-a6b8-4c4d-a879-1e2e4de95035


src/elements/content-sidebar/activity-feed/comment/__tests__/Replies.test.js
```diff
@@ -170,6 +172,22 @@ describe('elements/content-sidebar/ActivityFeed/comment/Replies', () => {
         expect(replyCreate).toBeCalledWith('Batman');
     });
 
+    test('should call onReplyDelete when reply is deleted', () => {
+        getWrapper({ replies: [comment] });
```
JChan106, line 176: Could we add another reply and make sure that the chosen reply is the correct reply that is deleted?

JChan106 (commented):
Have we checked that other users or even the owner of the file, cannot delete replies that aren't theirs?

Also do we know if this handles annotation replies? Or is it only comment replies?

src/elements/content-sidebar/activity-feed/comment/__tests__/Replies.test.js
bfoxx1906: I think line 188 should account for that. But it would be nice to add a few comments with different ids and just select a comment from the list in the replies component, and verify that reply delete was called with the right id..

bfoxx1906 (commented):
Should we also add a test to make sure that th

src/elements/content-sidebar/activity-feed/comment/__tests__/Replies.test.js
TylerGauntlett: Yea I can try that. I was running into issues with the test ids being duplicates whenever multiple comments were added but can find a way to work around that.

TylerGauntlett: @JChan106 does the current dropdown logic account for proper users? This change only makes the dropdown button work. If not, I'll update the dropdown logic to include.

TylerGauntlett: I updated my test to include multiple comments and verify that the expected comment is clicked. 

I began writing a test to ensure the delete option shows/hides based on the user (ie permissions) but found a much more comprehensive test for this already exists https://github.com/box/box-ui-elements/blob/master/src/elements/content-sidebar/activity-feed/comment/__tests__/BaseComment.test.js#L116

TylerGauntlett:
Rebased on master

bfoxx1906 (dismissed):

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -446,7 +451,7 @@ export const Replies = ({
                                     mentionSelectorContacts={mentionSelectorContacts}
                                     onCommentEdit={onCommentEdit}
                                     onSelect={onReplySelect}
-                                    onDelete={noop}
```
jcamden, line 449: I don't see any codepaths that were particular to this being `noop`---which surprising since `onEdit` has some `noop`-specific paths, like  `const canEdit = onEdit !== noop && (permissions.can_edit ?? false);`. Mentioning it juuust in case.

src/elements/content-sidebar/activity-feed/comment/__tests__/Replies.test.js
```diff
@@ -170,6 +172,27 @@ describe('elements/content-sidebar/ActivityFeed/comment/Replies', () => {
         expect(replyCreate).toBeCalledWith('Batman');
     });
 
+    test('should call onReplyDelete when reply is deleted', () => {
+        getWrapper({
+            replies: [comment, comment2],
+        });
+
+        const [menu] = screen.getAllByTestId('comment-actions-menu');
```
jcamden, line 180: Can you help me understand why `getAll-` here instead of just `get-`?

jcamden: Also, lets `getByLabel("Options")` vis-a-vis https://testing-library.com/docs/queries/about/#priority We get some free a11y verification that way :)

src/elements/content-sidebar/activity-feed/comment/__tests__/Replies.test.js
```diff
@@ -170,6 +172,27 @@ describe('elements/content-sidebar/ActivityFeed/comment/Replies', () => {
...
+
+        fireEvent.click(menu);
+
+        expect(screen.getByTestId('delete-comment')).toBeInTheDocument();
```
jcamden, line 184: I wonder if we should add and select by a label here. I think we want/need that for a11y.

If not, should we use `getByText`? If we're worried about "Delete" showing up in the comments themselves, we could use `within`.

src/elements/content-sidebar/activity-feed/comment/__tests__/Replies.test.js
```diff
@@ -170,6 +172,27 @@ describe('elements/content-sidebar/ActivityFeed/comment/Replies', () => {
...
+        fireEvent.click(screen.getByTestId('delete-comment'));
+
+        expect(screen.getByTestId('bcs-delete-confirmation-delete')).toBeInTheDocument();
```
jcamden, line 187: Since this is a button, it should probably be `getByLabelText` as well.

src/elements/content-sidebar/activity-feed/comment/__tests__/Replies.test.js
```diff
@@ -170,6 +172,27 @@ describe('elements/content-sidebar/ActivityFeed/comment/Replies', () => {
...
+        fireEvent.click(screen.getByTestId('bcs-delete-confirmation-delete'));
+
+        expect(replyDelete).toBeCalledTimes(1);
```
jcamden, line 190: Implementation detail; but I get that, at this level, it doesn't make sense to test `replyDelete` itself 👍

src/elements/content-sidebar/activity-feed/common/delete-confirmation/DeleteConfirmation.js
```diff
@@ -60,6 +60,7 @@ class DeleteConfirmation extends React.Component<Props> {
                     <Button
                         className="bcs-DeleteConfirmation-cancel"
                         onClick={onDeleteCancel}
+                        data-testid="bcs-delete-confirmation-cancel"
```
jcamden, line 63: Ah yes, here they are. Can we add labels instead and make Rob Smith happy?

src/elements/content-sidebar/activity-feed/common/delete-confirmation/DeleteConfirmation.js
```diff
@@ -58,16 +58,20 @@ class DeleteConfirmation extends React.Component<Props> {
                 </div>
                 <div>
                     <Button
+                        aria-label={commonMessages.cancel.defaultMessage}
                         className="bcs-DeleteConfirmation-cancel"
                         onClick={onDeleteCancel}
+                        data-testid="bcs-delete-confirmation-cancel"
```
jcamden, line 64: I think this is not needed in light of the label. Removing it will help push people towards selecting by label :)

src/elements/content-sidebar/activity-feed/common/delete-confirmation/DeleteConfirmation.js
```diff
@@ -58,16 +58,20 @@ class DeleteConfirmation extends React.Component<Props> {
...
                         type="button"
                         data-resin-target={ACTIVITY_TARGETS.INLINE_DELETE_CANCEL}
                     >
                         <FormattedMessage {...commonMessages.cancel} />
                     </Button>
                     <PrimaryButton
+                        aria-label={commonMessages.delete.defaultMessage}
                         className="bcs-DeleteConfirmation-delete"
                         onClick={onDeleteConfirm}
+                        data-testid="bcs-delete-confirmation-delete"
```
jcamden, line 74: Same thing here: label instead of testid 👍

src/elements/content-sidebar/activity-feed/comment/__tests__/Replies.test.js
jcamden: Oh yeah, I just meant check anything to do with `replyDelete` being called. Ideally, we would assert what the effects of `replyDelete` being called are for the user. But it's super practical to do that here.

jcamden (approved):
Looks good!

bfoxx1906 (approved):

PR #3404, opened 2023-09-21
fix(preview): Add accessibility to annotation badge type icon

aostaszewskiBox:


src/icons/two-toned/IconAnnotation.tsx
```diff
@@ -13,7 +13,15 @@ const IconAnnotation = ({ className = '', height = 24, title, width = 24 }: TwoT
...
     >
         <g fillRule="nonzero" fill="none">
-            <circle className="background-color" fill={bdlYellorange} cx="16.162" cy="16.162" r="16" />
+            <circle
+                className="background-color"
+                fill={bdlYellorange}
+                cx="16.162"
+                cy="16.162"
+                r="16"
+                aria-label="Inline comment"
```
arturfrombox, line 22: can we sort props in alphabetical order ?

src/icons/two-toned/IconAnnotation.tsx
```diff
@@ -13,7 +13,15 @@ const IconAnnotation = ({ className = '', height = 24, title, width = 24 }: TwoT
         width={width}
     >
         <g fillRule="nonzero" fill="none">
-            <circle className="background-color" fill={bdlYellorange} cx="16.162" cy="16.162" r="16" />
+            <circle
+                aria-label="Inline comment"
```
arturfrombox, line 17: do we need translation here ?

tjuanitas, line 17: this seems like specific text for an icon that could have multiple use cases. for example, Notes or Canvas could be using this icon in a way that's not related to inline comments

could this be handled in a more general way?

src/icons/two-toned/IconAnnotation.tsx
```diff
@@ -13,7 +13,15 @@ const IconAnnotation = ({ className = '', height = 24, title, width = 24 }: TwoT
...
+                className="background-color"
+                cx="16.162"
+                cy="16.162"
+                fill={bdlYellorange}
+                r="16"
+                tabIndex={0}
```
tjuanitas, line 23: why do we add these changes to the background part of the icon rather than one level higher to include all parts of the icon?

tjuanitas (commented):
I'm wondering if this is the best way to add accessibility for this icon. the `IconAnnotation` component takes a `title` prop which is passed to `AccessibleSVG`, would that resolve the issue your seeing?
https://github.com/box/box-ui-elements/blob/09110051c86f29a358c99c44ecba330fffb050dc/src/components/accessible-svg/AccessibleSVG.tsx#L40-L46

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -210,7 +210,9 @@ export const BaseComment = ({
                         user={createdByUser}
                         badgeIcon={
                             annotationActivityLink && (
-                                <IconAnnotation title={<FormattedMessage {...messages.annotationBadge} />} />
```
loonskai, line 213: Maybe it's worth checking with products if we want to update this title? They probably didn't know it existed as the icon wasn't focusable. If we need to update then `annotationBadge` should also be removed from messages if it's not used anywhere else

src/elements/content-sidebar/activity-feed/task-new/Task.js
```diff
@@ -279,7 +279,15 @@ class Task extends React.Component<Props, State> {
                     data-testid="task-card"
                 >
                     <Media.Figure className="bcs-Task-avatar">
-                        <Avatar getAvatarUrl={getAvatarUrl} user={createdByUser} badgeIcon={<TaskTypeIcon />} />
+                        <Avatar
+                            getAvatarUrl={getAvatarUrl}
+                            user={createdByUser}
+                            badgeIcon={
```
loonskai, line 285: nitpick: props alphabetical order

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
aostaszewskiBox: this is unused part of code since this icon was not focusable and now we use another one

src/icons/two-toned/IconTaskApproval.tsx
```diff
@@ -11,6 +11,7 @@ const IconTaskApproval = ({ className = '', height = 24, title, width = 24 }: Tw
         title={title}
         viewBox="0 0 32 32"
         width={width}
+        tabIndex={0}
```
tjuanitas, line 14: this makes all usages of this icon tabbable, is that what we want? for example, this is used in task service client and EUA

tjuanitas (commented):
hey sorry I'm trying to better understand what we're trying to accomplish with this change. why do we need to make the icon focusable? is the the user able to interact with this?

src/icons/two-toned/IconTaskApproval.tsx
aostaszewskiBox: It is intended to be tabbable since if it is not, it won't be visible for screen reader. Those Icons are also used in content-sidebar in comment section which needs to be accessible.

tjuanitas: the screen reader can still read aria-labels even if the icon itself is not explicitly tabbable. I don't think we need to make the icon focusable otherwise we'd have accessibility issues with every icon used in BUIE since none of them currently have `tabIndex`, don't you think?

for example, we used the `title` prop to resolve accessibility issues for the "Global" icon used on Avatars:
https://github.com/box/box-ui-elements/blob/69293d8903f17d136aa567fd674e2fe263873d86/src/components/avatar/Avatar.tsx#L99

and this is read by screen readers without using tabIndex:
Screenshot 2023-09-25 at 2 06 39 PM

tjuanitas: this change would make all icons, regardless of it's usage, tabbable

e.g. if this icon is used inside of a button, such as for an IconButton, the component would now focus twice: once for the button focus and once for the icon focus. that doesn't seem right from a user experience perspective

aostaszewskiBox: You are right, accessibility is working also when only title is passed. I did test it wrong previously so tabIndex is deleted

arturfrombox (approved):

loonskai (approved):

src/elements/content-sidebar/activity-feed/comment/messages.js
```diff
@@ -62,6 +57,11 @@ const messages = defineMessages({
         defaultMessage: 'See {repliesToLoadCount, plural, one {# reply} other {# replies}}',
         description: 'Text to show to get more replies of comment or annotation',
     },
+    inlineCommentAnnotationAccessibleSVGTitle: {
+        id: 'be.contentSidebar.activityFeed.comment.IconAnnotation',
```
tjuanitas, line 61: minor - the ID was changed:
```suggestion
    inlineCommentAnnotationAccessibleSVGTitle: {
        id: 'be.contentSidebar.activityFeed.comment.inlineCommentAnnotationAccessibleSVGTitle',
```

src/elements/content-sidebar/activity-feed/task-new/messages.js
```diff
@@ -175,6 +175,16 @@ const messages = defineMessages({
...
     },
+    generalTaskAnnotationAccessibleSVGTitle: {
+        id: 'be.contentSidebar.activityFeed.task.generalTaskAnnotationAccessibleSVGTitle',
+        defaultMessage: 'General Task',
+        description: 'Text for aria label of general task circle type icon',
+    },
+    approvalTaskAnnotationAccessibleSVGTitle: {
+        id: 'be.contentSidebar.activityFeed.task.approvalTaskAnnotationAccessibleSVGTitle',
+        defaultMessage: 'Approval Task',
+        description: 'Text for aria label of approval taskcircle type icon',
```
tjuanitas, line 186: super nit:
```suggestion
        description: 'Text for aria label of approval task circle type icon',
```

src/elements/content-sidebar/activity-feed/comment/messages.js
```diff
@@ -62,6 +57,11 @@ const messages = defineMessages({
         defaultMessage: 'See {repliesToLoadCount, plural, one {# reply} other {# replies}}',
         description: 'Text to show to get more replies of comment or annotation',
     },
+    inlineCommentAnnotationAccessibleSVGTitle: {
```
tjuanitas, line 60: another really minor thing:
```suggestion
    inlineCommentAnnotationIconTitle: {
```
`AccessibleSVG` refers to an internal component name, I think it's okay to go a little more general and call this `Icon` so that the message isn't coupled to a specific component - i.e. allows for reusability and implementation changes

(same comment for the other messages)

tjuanitas (commented):
some minor comments but logic makes sense to me now 👍 

thanks!

src/elements/content-sidebar/activity-feed/comment/messages.js
```diff
@@ -62,6 +57,11 @@ const messages = defineMessages({
         defaultMessage: 'See {repliesToLoadCount, plural, one {# reply} other {# replies}}',
         description: 'Text to show to get more replies of comment or annotation',
     },
+    inlineCommentAnnotationIconTitle: {
+        id: 'be.contentSidebar.activityFeed.comment.inlineCommentAnnotationIconTitle',
+        defaultMessage: 'Inline Comment',
+        description: 'Text for aria label of circle type icon',
```
karelee7, line 63: nit: is there a more clear/descriptive name for what this circle type icon is?

bkepka-box (approved):

arturfrombox (approved):

karelee7 (approved):
Looks good! If you could update the. PR description with a short summary and any relevant screenshots/images that would be great!

aostaszewskiBox:
Added accessibility to svg icons in content-sidebar

PR #3405, opened 2023-09-21
feat(activityitems): new layout

jcamden:
The purpose of this task is to bring `BaseComment` into alignment with the layout for TR comments in [designs](https://www.figma.com/file/EyW24R6Bg9IkiQ33urle5l/Preview-Threaded-Replies-and-Annotations?node-id=1169%3A189350&mode=dev). The nesting of wrappers to apply the new layout prompted a request to break `BaseComment` down into its constituent components. That is the content of this PR.

Please note that bringing activity item styles into full alignment with the designs is beyond the scope of this PR. I did only what I subjectively considered to be the minimum to accommodate the new layout. (See [this document](https://cloud.box.com/s/04xkqx19dfkj5v4v069waqwh7locfzt3) for an attempt at cataloguing all incoming UI diffs).

legacy (on dev pod)
Screenshot 2023-09-27 at 12 45 55 PM

new layout (on dev pod)
TR-new-layout

If helpful, some screenshots of stories:
Screenshot 2023-09-26 at 3 46 53 PM
Screenshot 2023-09-26 at 3 46 58 PM
Screenshot 2023-09-26 at 3 47 32 PM
Screenshot 2023-09-26 at 4 05 03 PM
Screenshot 2023-09-26 at 3 47 42 PM
Screenshot 2023-09-26 at 3 47 45 PM

### DIFF
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
@@ -41,7 +41,7 @@ import { type OnAnnotationEdit } from '../comment/types';
 
 import './ActivityFeed.scss';
 
-type Props = {
+export type ActivityFeedProps = {
     activeFeedEntryId?: string,
     activeFeedEntryType?: FocusableFeedItemType,
     activityFeedError: ?Errors,
@@ -49,6 +49,7 @@ type Props = {
     contactsLoaded?: boolean,
     currentUser?: User,
     feedItems?: FeedItems,
+    // $FlowFixMe file.file_version (BoxItem[BoxItemVersion]) has many apparently long-standing type errors
     file: BoxItem,
     getApproverWithQuery?: Function,
     getAvatarUrl: GetAvatarUrlCallback,
@@ -104,7 +105,7 @@ type State = {
     selectedItemId: string | null,
 };
 
-class ActivityFeed extends React.Component<Props, State> {
+class ActivityFeed extends React.Component<ActivityFeedProps, State> {
     state = {
         isScrolled: false,
         isInputOpen: false,
@@ -119,7 +120,7 @@ class ActivityFeed extends React.Component<Props, State> {
         this.resetFeedScroll();
     }
 
-    componentDidUpdate(prevProps: Props, prevState: State) {
+    componentDidUpdate(prevProps: ActivityFeedProps, prevState: State) {
         const {
             activeFeedEntryId: prevActiveFeedEntryId,
             currentUser: prevCurrentUser,
@@ -168,7 +169,7 @@ class ActivityFeed extends React.Component<Props, State> {
      * @param {object} currentUser - The user that is logged into the account
      * @param {object} feedItems - Items in the activity feed
      */
-    isEmpty = ({ feedItems, shouldUseUAA }: Props = this.props): boolean => {
+    isEmpty = ({ feedItems, shouldUseUAA }: ActivityFeedProps = this.props): boolean => {
         if (feedItems === undefined) {
             return false;
         }
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActivityItem.scss
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActivityItem.scss
@@ -6,7 +6,7 @@
 
         // This brittle selector is here in order to only select the parent BaseComment and not the BaseComments which constitute its replies.
         & > div {
-            padding: 6px 12px $bdl-grid-unit * 5;
+            padding: 6px $bdl-grid-unit * 3 $bdl-grid-unit * 5;
         }
 
         &.bcs-is-hoverable:hover {
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActivityItem.scss
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActivityItem.scss
@@ -4,9 +4,9 @@
     &.hasNewThreadedReplies {
         padding: 2px $bdl-grid-unit * 2;
 
+        // This brittle selector is here in order to only select the parent BaseComment and not the BaseComments which constitute its replies.
         & > div {
-            padding: $bdl-grid-unit * 2;
-            padding-right: $bdl-grid-unit * 4;
+            padding: 6px 12px $bdl-grid-unit * 5;
         }
 
         &.bcs-is-hoverable:hover {
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActivityItem.scss
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActivityItem.scss
@@ -6,7 +6,7 @@
 
         // This brittle selector is here in order to only select the parent BaseComment and not the BaseComments which constitute its replies.
         & > div {
-            padding: 6px $bdl-grid-unit * 3 $bdl-grid-unit * 5;
+            padding: $bdl-grid-unit * 3 $bdl-grid-unit * 3 $bdl-grid-unit * 5;
         }
 
         &.bcs-is-hoverable:hover {
--- a/src/elements/content-sidebar/activity-feed/activity-feed/stories/ActivityFeed.stories.js
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/stories/ActivityFeed.stories.js
@@ -1,0 +1,70 @@
+// @flow
+import * as React from 'react';
+
+import { IntlProvider } from 'react-intl';
+
+import ActivityFeed, { type ActivityFeedProps } from '../ActivityFeed';
+import {
+    TIME_STRING_SEPT_27_2017,
+    TIME_STRING_SEPT_28_2017,
+    currentUser,
+    comment,
+    replies,
+    repliesProps,
+} from '../../comment/stories/common';
+import { FEED_ITEM_TYPE_VERSION } from '../../../../../constants';
+
+import type { FeedItems, Comment } from '../../../../../common/types/feed';
+
+const file = {
+    id: '12345',
+    permissions: {
+        can_comment: true,
+    },
+    created_at: TIME_STRING_SEPT_27_2017,
+    modified_at: TIME_STRING_SEPT_28_2017,
+    file_version: {
+        id: 987,
+        type: FEED_ITEM_TYPE_VERSION,
+        created_at: TIME_STRING_SEPT_28_2017,
+        modified_at: null,
+        modified_by: null,
+        trashed_at: null,
+        version_number: 1,
+    },
+    version_number: '3',
+};
+
+const activityFeedComment: Comment = {
+    ...comment,
+    replies,
+};
+
+const comments = {
+    total_count: 1,
+    entries: [activityFeedComment],
+};
+
+const feedItems: FeedItems = [...comments.entries];
+
+const getTemplate = customProps => (props: ActivityFeedProps) => (
+    <IntlProvider locale="en">
+        <ActivityFeed
+            currentUser={currentUser}
+            feedItems={feedItems}
+            file={file}
+            {...repliesProps}
+            {...props}
+            {...customProps}
+        />
+    </IntlProvider>
+);
+
+export const LegacyComment = getTemplate({ hasNewThreadedReplies: false });
+
+export const ThreadedRepliesComment = getTemplate({ hasNewThreadedReplies: true });
+
+export default {
+    title: 'Components|ActivityFeed',
+    component: ActivityFeed,
+};
--- a/src/elements/content-sidebar/activity-feed/comment/BaseComment.js
+++ b/src/elements/content-sidebar/activity-feed/comment/BaseComment.js
@@ -1,45 +1,35 @@
 // @flow
 import * as React from 'react';
+
 import classNames from 'classnames';
 import noop from 'lodash/noop';
-import TetherComponent from 'react-tether';
-import { FormattedMessage } from 'react-intl';
+
+import { BaseCommentMenuWrapper } from './components/BaseCommentMenuWrapper';
+import { BaseCommentInfo } from './components/BaseCommentInfo';
+import { COMMENT_STATUS_RESOLVED } from '../../../../constants';
+
 import ActivityError from '../common/activity-error';
 import ActivityMessage from '../common/activity-message';
-import ActivityStatus from '../common/activity-status';
-import ActivityTimestamp from '../common/activity-timestamp';
-import Avatar from '../Avatar';
-import Checkmark16 from '../../../../icon/line/Checkmark16';
 import CommentForm from '../comment-form';
 import CreateReply from './CreateReply';
-import DeleteConfirmation from '../common/delete-confirmation';
-import IconAnnotation from '../../../../icons/two-toned/IconAnnotation';
 import LoadingIndicator from '../../../../components/loading-indicator';
-import Media from '../../../../components/media';
-import messages from './messages';
-import Pencil16 from '../../../../icon/line/Pencil16';
 import RepliesToggle from './RepliesToggle';
-import Trash16 from '../../../../icon/line/Trash16';
-import UserLink from '../common/user-link';
-import X16 from '../../../../icon/fill/X16';
-import { ACTIVITY_TARGETS } from '../../../common/interactionTargets';
-import { COMMENT_STATUS_OPEN, COMMENT_STATUS_RESOLVED, PLACEHOLDER_USER } from '../../../../constants';
-import { MenuItem } from '../../../../components/menu';
-import { type OnAnnotationEdit, type OnCommentEdit } from './types';
+
 import type {
     ActionItemError,
     BoxCommentPermission,
     FeedItemStatus,
     Comment as CommentType,
 } from '../../../../common/types/feed';
-
 import type { GetAvatarUrlCallback, GetProfileUrlCallback } from '../../../common/flowTypes';
+import type { OnAnnotationEdit, OnCommentEdit } from './types';
 import type { SelectorItems, User } from '../../../../common/types/core';
 import type { Translations } from '../../flowTypes';
 
 import './BaseComment.scss';
 import './Replies.scss';
 import './Comment.scss';
+import ActivityStatus from '../common/activity-status';
 
 export type BaseCommentProps = {
     annotationActivityLink?: React.Element<any>,
@@ -106,37 +96,8 @@ export const BaseComment = ({
     translatedTaggedMessage,
     translations,
 }: BaseCommentProps) => {
-    const [isConfirmingDelete, setIsConfirmingDelete] = React.useState<boolean>(false);
     const [isEditing, setIsEditing] = React.useState<boolean>(false);
     const [isInputOpen, setIsInputOpen] = React.useState<boolean>(false);
-
-    const handleDeleteConfirm = (): void => {
-        onDelete({ id, permissions });
-        onSelect(false);
-    };
-
-    const handleDeleteCancel = (): void => {
-        setIsConfirmingDelete(false);
-        onSelect(false);
-    };
-
-    const handleDeleteClick = (): void => {
-        setIsConfirmingDelete(true);
-        onSelect(true);
-    };
-
-    const handleEditClick = (): void => {
-        setIsEditing(true);
-        setIsInputOpen(true);
-        onSelect(true);
-    };
-
-    const handleMenuClose = (): void => {
-        if (isConfirmingDelete || isEditing || isInputOpen) {
-            return;
-        }
-        onSelect(false);
-    };
 
     const commentFormFocusHandler = (): void => {
         setIsInputOpen(true);
@@ -170,22 +131,13 @@ export const BaseComment = ({
         commentFormSubmitHandler();
     };
 
-    const handleStatusUpdate = (selectedStatus: FeedItemStatus): void => {
-        if (onAnnotationEdit) {
-            onAnnotationEdit({ id, permissions });
-        } else if (onCommentEdit) {
-            onCommentEdit({ id, status: selectedStatus, hasMention: false, permissions });
-        }
-    };
-
     // Since we have to pass onCommentEdit through annotations (to Replies), onAnnotationEdit essentially overrides onCommentEdit
     const onEdit = onAnnotationEdit ?? onCommentEdit;
 
-    const canDelete = permissions.can_delete;
-    const canEdit = onEdit !== noop && permissions.can_edit;
-    const canResolve = onEdit !== noop && permissions.can_resolve;
-    const createdAtTimestamp = new Date(created_at).getTime();
-    const createdByUser = created_by || PLACEHOLDER_USER;
+    const canDelete = !!permissions.can_delete;
+    const canEdit = onEdit !== noop && !!permissions.can_edit;
+    const canResolve = onEdit !== noop && !!permissions.can_resolve;
+
     const isEdited = modified_at !== undefined && modified_at !== created_at;
     const isMenuVisible = (canDelete || canEdit || canResolve) && !isPending;
     const isResolved = status === COMMENT_STATUS_RESOLVED;
@@ -201,110 +153,41 @@ export const BaseComment = ({
     return (
         // TODO: Change className to bcs-Comment once FF is removed
         <div className="bcs-BaseComment">
-            <Media
+            <div
                 className={classNames('bcs-Comment-media', {
                     'bcs-is-pending': isPending || error,
                 })}
             >
-                <Media.Figure>
-                    <Avatar
-                        badgeIcon={
-                            annotationActivityLink && (
-                                <IconAnnotation
-                                    title={<FormattedMessage {...messages.inlineCommentAnnotationIconTitle} />}
-                                />
-                            )
-                        }
+                <div className="bcs-BaseComment-header">
+                    <BaseCommentInfo
+                        annotationActivityLink={annotationActivityLink}
+                        created_at={created_at}
+                        created_by={created_by}
                         getAvatarUrl={getAvatarUrl}
-                        user={createdByUser}
+                        getUserProfileUrl={getUserProfileUrl}
+                        status={status}
                     />
-                </Media.Figure>
-                <Media.Body>
                     {isMenuVisible && (
-                        <TetherComponent
-                            attachment="top right"
-                            className="bcs-Comment-deleteConfirmationModal"
-                            constraints={[{ to: 'scrollParent', attachment: 'together' }]}
-                            targetAttachment="bottom right"
-                        >
-                            <Media.Menu
-                                isDisabled={isConfirmingDelete}
-                                data-testid="comment-actions-menu"
-                                dropdownProps={{
-                                    onMenuOpen: () => onSelect(true),
-                                    onMenuClose: handleMenuClose,
-                                }}
-                                menuProps={{
-                                    'data-resin-component': ACTIVITY_TARGETS.COMMENT_OPTIONS,
-                                }}
-                            >
-                                {canResolve && isResolved && (
-                                    <MenuItem
-                                        className="bcs-Comment-unresolveComment"
-                                        data-resin-target={ACTIVITY_TARGETS.COMMENT_OPTIONS_EDIT}
-                                        data-testid="unresolve-comment"
-                                        onClick={() => handleStatusUpdate(COMMENT_STATUS_OPEN)}
-                                    >
-                                        <X16 />
-                                        <FormattedMessage {...messages.commentUnresolveMenuItem} />
-                                    </MenuItem>
-                                )}
-                                {canResolve && !isResolved && (
-                                    <MenuItem
-                                        data-resin-target={ACTIVITY_TARGETS.COMMENT_OPTIONS_EDIT}
-                                        data-testid="resolve-comment"
-                                        onClick={() => handleStatusUpdate(COMMENT_STATUS_RESOLVED)}
-                                    >
-                                        <Checkmark16 />
-                                        <FormattedMessage {...messages.commentResolveMenuItem} />
-                                    </MenuItem>
-                                )}
-                                {canEdit && (
-                                    <MenuItem
-                                        data-resin-target={ACTIVITY_TARGETS.COMMENT_OPTIONS_EDIT}
-                                        data-testid="edit-comment"
-                                        onClick={handleEditClick}
-                                    >
-                                        <Pencil16 />
-                                        <FormattedMessage {...messages.commentEditMenuItem} />
-                                    </MenuItem>
-                                )}
-                                {canDelete && (
-                                    <MenuItem
-                                        aria-label={messages.commentDeleteMenuItem.defaultMessage}
-                                        data-resin-target={ACTIVITY_TARGETS.COMMENT_OPTIONS_DELETE}
-                                        data-testid="delete-comment"
-                                        onClick={handleDeleteClick}
-                                    >
-                                        <Trash16 />
-                                        <FormattedMessage {...messages.commentDeleteMenuItem} />
-                                    </MenuItem>
-                                )}
-                            </Media.Menu>
-                            {isConfirmingDelete && (
-                                <DeleteConfirmation
-                                    data-resin-component={ACTIVITY_TARGETS.COMMENT_OPTIONS}
-                                    isOpen={isConfirmingDelete}
-                                    message={messages.commentDeletePrompt}
-                                    onDeleteCancel={handleDeleteCancel}
-                                    onDeleteConfirm={handleDeleteConfirm}
-                                />
-                            )}
-                        </TetherComponent>
-                    )}
-                    <div className="bcs-Comment-headline">
-                        <UserLink
-                            data-resin-target={ACTIVITY_TARGETS.PROFILE}
-                            id={createdByUser.id}
-                            name={createdByUser.name}
-                            getUserProfileUrl={getUserProfileUrl}
+                        <BaseCommentMenuWrapper
+                            canDelete={canDelete}
+                            canEdit={canEdit}
+                            canResolve={canResolve}
+                            id={id}
+                            isEditing={isEditing}
+                            isInputOpen={isInputOpen}
+                            isResolved={isResolved}
+                            onAnnotationEdit={onAnnotationEdit}
+                            onCommentEdit={onCommentEdit}
+                            onDelete={onDelete}
+                            onSelect={onSelect}
+                            permissions={permissions}
+                            setIsEditing={setIsEditing}
+                            setIsInputOpen={setIsInputOpen}
                         />
-                    </div>
-                    <div className="bcs-Comment-timestamp">
-                        <ActivityTimestamp date={createdAtTimestamp} />
-                        {annotationActivityLink}
-                    </div>
-                    <ActivityStatus status={status} />
+                    )}
+                </div>
+                <ActivityStatus status={status} />
+                <div className="bcs-BaseComment-content">
                     {isEditing ? (
                         <CommentForm
                             className={classNames('bcs-Comment-editor', {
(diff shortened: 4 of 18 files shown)


src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
```diff
@@ -41,7 +41,7 @@ import { type OnAnnotationEdit } from '../comment/types';
 
 import './ActivityFeed.scss';
 
-type Props = {
+export type ActivityFeedProps = {
```
ahorowitz123, line 44: nit: if the reason for changing the name here is because its being exported I would probably change the name at the point of export in the index folder instead

ahorowitz123:
@jcamden can you add before and after gifs and a link to the designs in the description?

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.js
jcamden: I don't quite see the advantage, and I think there is a disadvantage to renaming exports in terms of readability. I have a strong preference for providing a good name where I variable is defined.

jcamden:
> @jcamden can you add before and after gifs, a link to the designs, and a list of the changes you made in the description?

Whoops! I completely lost my PR description. (I copied it out when I needed to---oh, blablabla.) I'll add that asap.

mickr:
@jcamden can you update the description of the intent of this PR?

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -199,139 +199,154 @@ export const BaseComment = ({
...
+                                    badgeIcon={
+                                        annotationActivityLink && (
+                                            <IconAnnotation
+                                                title={<FormattedMessage {...messages.annotationBadge} />}
+                                            />
+                                        )
+                                    }
+                                />
+                            </div>
+                            <div className="bcs-BaseComment-MetaRow-Metadata-Data">
```
jstoffan, line 223: We seem to have delved too greedily and too deep with the BaseComment component. The nesting is getting a bit intense and the naming is losing its meaning. It looks like we may need to refactor into a few smaller components at this point.

jcamden: I 100% agree. Sorry I haven't added a PR description yet. I copied it out while I was making some changes, and forgot to copy it back in, but in the original description I said this:

> Although a re-architecture of BaseComment may be in order, for the purposes of this task, I tried to constrain myself to doing the minimum required to implement the [new layout](https://www.figma.com/file/EyW24R6Bg9IkiQ33urle5l/Preview-Threaded-Replies-and-Annotations?node-id=1169%3A189350&mode=dev) of comments and annotations.

The reason I say this is because I have already proposed re-architecturing of BaseComment long ago but it was not perceived to be worth it at the time. I mentioned this to Adam in a DM on Monday to get his advice, he advised to proceed as is, and try proposing re-architecturing again in-between tasks. That sounded wise to me. What d'you think?

jstoffan: Re-architecture and refactoring are sometimes different. I haven't seen the prior proposal, so I can't say. In broad terms, we should take the time to build things the best way we can within reason. I would imagine extracting the new layout to its own component(s), along with maybe the menu, would help a lot at relatively low cost.

jcamden: That sounds pretty good. We've discussed just that refactor. I think that's a good place to start, and I'd be happy to do it in this PR. 

I also recommended [this re-architecture](https://github.com/box/box-ui-elements/pull/3321) back in the day. Probably the key point there is solving the circular dependency on Replies by passing it in as a child. That's a more major overhaul. Do you think it's worth brining that up again?

bfoxx1906: The overall architecture as a conept is ok but we can break BaseComment down into constituent components. Right now we just want to make sure that everything is ready for the Boxer release because I would be shocked if we don't have more changes to make after people start using it.

jcamden: Maybe the overall architecture is okay, but certainly not great. It would certainly be better to define the Replies component in its own file rather than defining it in BaseComment.js. But I also see that it might not be worth it. Super happy about the green light for re-organizing into constituent components though, and I will get on top of that right away :)

src/elements/content-sidebar/activity-feed/activity-feed/ActivityItem.scss
```diff
@@ -4,9 +4,9 @@
     &.hasNewThreadedReplies {
         padding: 2px $bdl-grid-unit * 2;
 
+        // This brittle selector is here in order to only select the parent BaseComment and not the BaseComments which constitute its replies.
         & > div {
```
shahzadaziz, line 8: Any way to make it less brittle ?

src/elements/content-sidebar/activity-feed/activity-feed/stories/ActivityFeed.stories.js
```diff
@@ -0,0 +1,70 @@
...
+            {...customProps}
+        />
+    </IntlProvider>
+);
+
+export const LegacyComment = getTemplate({ hasNewThreadedReplies: false });
+
+export const ThreadedRepliesComment = getTemplate({ hasNewThreadedReplies: true });
+
+export default {
```
shahzadaziz, line 67: nit: any reason why we doing an object as default export ? Is that the recommendation ?

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -199,139 +199,154 @@ export const BaseComment = ({
...
-                                data-testid="comment-actions-menu"
-                                dropdownProps={{
-                                    onMenuOpen: () => onSelect(true),
-                                    onMenuClose: handleMenuClose,
-                                }}
-                                menuProps={{
-                                    'data-resin-component': ACTIVITY_TARGETS.COMMENT_OPTIONS,
-                                }}
+                <div className="bcs-BaseComment-MainComment">
+                    <div className="bcs-BaseComment-MetaRow">
```
bfoxx1906, line 208: This would be a good candidate for extraction. As long as we can make sure we stay within our timeframes I'm good with it.

src/elements/content-sidebar/activity-feed/activity-feed/ActivityItem.scss
jcamden: Yes, actually. And I would love to do so. I think we need to restructure a bit so that the focus-implying border is on a wrapper for the BaseComment rather than on the BaseComment itself. Then we don't need to distinguish the first BaseComment from its replies.

src/elements/content-sidebar/activity-feed/activity-feed/stories/ActivityFeed.stories.js
jcamden: Yeah, I'm really not sure. I just copied this from other Storybook files in BUIE. Storybook is so far behind BUIE that I am not even really clear on what the story format was back then. It would be great to upgrade from 5 => 7. I actually did that as part of my Node 18 POC (even though it turned out to be unnecessary). There was a bit to do, overhauling configs in the .storybook directory, but not an overwhelming amount. If we get on latest, we can adopt latest story format standards (and it's still backwards compatible with old story formats IIRC).

src/elements/content-sidebar/activity-feed/comment/components/BaseCommentMetadata.js
```diff
@@ -0,0 +1,73 @@
...
+export interface BaseCommentMetadataProps {
+    annotationActivityLink?: React.Element<any> | typeof undefined;
+    created_at: string | number;
+    created_by: User;
+    getAvatarUrl: GetAvatarUrlCallback;
+    getUserProfileUrl?: GetProfileUrlCallback | typeof undefined;
+    status?: FeedItemStatus | typeof undefined;
+}
+
+export const BaseCommentMetadata = ({
```
jstoffan, line 30: We may want to reconsider the use of "metadata" and "data" in these components, since they're both generic and super overloaded. Maybe something like `BaseCommentCitation` or `BaseCommentByline`.

src/elements/content-sidebar/activity-feed/comment/components/BaseCommentMetadata.js
```diff
@@ -0,0 +1,73 @@
...
+                    user={createdByUser}
+                    badgeIcon={
+                        annotationActivityLink && (
+                            <IconAnnotation title={<FormattedMessage {...messages.annotationBadge} />} />
+                        )
+                    }
+                />
+            </div>
+            <div className="bcs-BaseComment-Metadata-Data">
+                <div className="bcs-Comment-headline">
```
jstoffan, line 55: SUIT conventions call for every class name to be scoped to the root name of the component it's in. In this case, it would be `bcs-BaseCommentMetadata-headline`.

src/elements/content-sidebar/activity-feed/comment/components/BaseCommentMetadata.js
```diff
@@ -0,0 +1,73 @@
...
+    getAvatarUrl,
+    getUserProfileUrl,
+    status,
+}: BaseCommentMetadataProps) => {
+    const createdByUser = created_by || PLACEHOLDER_USER;
+    const createdAtTimestamp = new Date(created_at).getTime();
+
+    return (
+        <div className="bcs-BaseComment-Metadata">
+            <div className="bcs-BaseComment-Metadata-Avatar">
```
jstoffan, line 43: SUIT calls for non-component child classes to be lower-cased: `bcs-BaseCommentMetadata-avatar`.

src/elements/content-sidebar/activity-feed/comment/components/BaseCommentMetadata.js
```diff
@@ -0,0 +1,73 @@
...
+                    getAvatarUrl={getAvatarUrl}
+                    user={createdByUser}
+                    badgeIcon={
+                        annotationActivityLink && (
+                            <IconAnnotation title={<FormattedMessage {...messages.annotationBadge} />} />
+                        )
+                    }
+                />
+            </div>
+            <div className="bcs-BaseComment-Metadata-Data">
```
jstoffan, line 54: Can we brainstorm something other than "metadata data" for this information?

src/elements/content-sidebar/activity-feed/comment/components/BaseCommentMenuWrapper.js
```diff
@@ -0,0 +1,95 @@
...
+export const BaseCommentMenuWrapper = ({
+    canDelete,
+    canEdit,
+    canResolve,
+    id,
+    isEditing,
+    isInputOpen,
+    isResolved,
+    onAnnotationEdit,
+    onCommentEdit,
```
jstoffan, line 34: Should these be combined into `onEdit`? We can allow the parent to figure out comment versus annotation.

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -168,22 +128,13 @@ export const BaseComment = ({
...
-
     // Since we have to pass onCommentEdit through annotations (to Replies), onAnnotationEdit essentially overrides onCommentEdit
     const onEdit = onAnnotationEdit ?? onCommentEdit;
 
-    const canDelete = permissions.can_delete;
-    const canEdit = onEdit !== noop && permissions.can_edit;
-    const canResolve = onEdit !== noop && permissions.can_resolve;
-    const createdAtTimestamp = new Date(created_at).getTime();
-    const createdByUser = created_by || PLACEHOLDER_USER;
+    const canDelete = permissions.can_delete ?? false;
```
jstoffan, line 134: We should be cautious about changing business logic in a refactor. Are these values nullable now? If not, is the `?? false` needed?

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -204,147 +155,66 @@ export const BaseComment = ({
...
-                                        >
-                                            <Trash16 />
-                                            <FormattedMessage {...messages.commentDeleteMenuItem} />
-                                        </MenuItem>
-                                    )}
-                                </Media.Menu>
+                <div className="bcs-BaseComment-Header">
+                    <BaseCommentMetadata
+                        annotationActivityLink={annotationActivityLink}
+                        created_at={created_at}
```
jstoffan, line 161: We should prefer camelCase for prop names.

src/elements/content-sidebar/activity-feed/comment/components/BaseCommentMenu.js
```diff
@@ -0,0 +1,125 @@
...
+    handleDeleteConfirm,
+    handleEditClick,
+    handleMenuClose,
+    handleStatusUpdate,
+    isConfirmingDelete,
+    isResolved,
+    onSelect,
+}: BaseCommentMenuProps) => {
+    return (
+        <div className="BaseCommentMenu">
```
jstoffan, line 52: Is this wrapping div needed or can we add a class to the `Media.Menu` component?

tsconfig.json
```diff
@@ -11,5 +11,11 @@
         }
     },
     "files": [".storybook/typings.d.ts"],
-    "include": ["src/**/*.ts", "src/**/*.tsx", "examples/src/**.tsx", "types.ts"]
+    "include": [
+        "src/**/*.ts",
+        "src/**/*.tsx",
+        "examples/src/**.tsx",
+        "types.ts",
+        "src/elements/content-sidebar/activity-feed/comment/components/BaseCommentMenu.js"
```
jstoffan, line 19: What's the reason for including this file here?

src/elements/content-sidebar/activity-feed/comment/components/BaseCommentMenuWrapper.js
```diff
@@ -0,0 +1,95 @@
...
+    onAnnotationEdit?: OnAnnotationEdit | typeof undefined;
+    onCommentEdit: OnCommentEdit;
+    onDelete: ({ id: string, permissions?: BoxCommentPermission }) => any;
+    onSelect: (isSelected: boolean) => void;
+    permissions: BoxCommentPermission;
+    setIsEditing: ((boolean => boolean) | boolean) => void;
+    setIsInputOpen: ((boolean => boolean) | boolean) => void;
+}
+
+export const BaseCommentMenuWrapper = ({
```
jstoffan, line 25: In this case, this component is acting as a state container, so "Container" seems preferable.

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -1,45 +1,34 @@
...
-
 import type { GetAvatarUrlCallback, GetProfileUrlCallback } from '../../../common/flowTypes';
+import type { OnAnnotationEdit, OnCommentEdit } from './types';
 import type { SelectorItems, User } from '../../../../common/types/core';
 import type { Translations } from '../../flowTypes';
 
 import './BaseComment.scss';
-import './Replies.scss';
 import './Comment.scss';
+import './Replies.scss';
```
jstoffan, line 31: Import order for CSS can have side effects. Is this needed?

src/elements/content-sidebar/activity-feed/comment/components/stories/BaseCommentMenu.stories.js
```diff
@@ -0,0 +1,42 @@
+// @flow
+import * as React from 'react';
```
jstoffan, line 2: We may want to wait to start down this path, especially since this change is going to be quite large due the refactor. Do we need stories for low-level components that are scoped to a single "experience"?

src/elements/content-sidebar/activity-feed/comment/components/__tests__/BaseCommentMenu.test.js
```diff
@@ -0,0 +1,100 @@
...
+    isResolved: false,
+    onSelect: jest.fn(),
+};
+
+const openMenu = (wrapper: RenderResult) => {
+    fireEvent.click(wrapper.queryByTestId('comment-actions-menu'));
+};
+
+// eslint-disable-next-line import/prefer-default-export
+export const getWrapper = props =>
```
jstoffan, line 37: Why are we exporting from a unit test?

src/elements/content-sidebar/activity-feed/comment/components/__tests__/BaseCommentMenu.test.js
```diff
@@ -0,0 +1,100 @@
...
+        ${false}   | ${true}    | ${'should NOT'}
+        ${true}    | ${false}   | ${'should NOT'}
+        ${false}   | ${false}   | ${'should NOT'}
+    `(
+        `$should render unresolve menu item when canResolve us $canResolve and isResolved is $isResolved`,
+        ({ canResolve, isResolved, should }) => {
+            const wrapper = getWrapper({ canResolve, isResolved });
+            const message = localize(messages.commentUnresolveMenuItem.id);
+            openMenu(wrapper);
+            if (should === 'should') {
```
jstoffan, line 75: We typically want to avoid introducing logic in our tests. Best to make these two sets of independent tests.

jstoffan (commented):
May be worth moving the refactor to its own PR.

src/elements/content-sidebar/activity-feed/activity-feed/ActivityItem.scss
ahorowitz123: This focus border should only be on the outer BaseComment? In that case I agree, it shouldn't be part of BaseComment and should be factored out to a wrapper

src/elements/content-sidebar/activity-feed/comment/components/BaseCommentMetadata.js
jcamden: Hey, I see you're looking at this. FYI, still a work-in-progess, but I do appreciate it. Re: this class, it's not a new class that I'm creating. It's a holdover from Comment.

src/elements/content-sidebar/activity-feed/comment/components/BaseCommentMenuWrapper.js
jcamden: That's part of why this wrapper exists. This way, the dumb component `BaseCommentMenu` doesn't need to know about `onCommentEdit` and `onAnnotationEdit`. It just receives `handleStatusUpdate`.

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
jcamden: Oh, that's a good point. You're right that I was just naively alphabetizing. There was no effect, but not needed. I'll revert this.

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
jcamden: In this case, we really just need a boolean. It is not necessary to continue to allow it to be nullable after this point.

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
jcamden: Indeed. I'm not 100% sure where this originates, but it's pretty entrenched. Search for `created_at` yielded 243 results. Where do you recommend that I transform this? Just on the new components? Or at BaseComment? Or some parent of BaseComment. Jlmk :)

src/elements/content-sidebar/activity-feed/comment/components/BaseCommentMenuWrapper.js
jcamden: For definitions of wrappers versus containers, I've been going by the highest voted answer [here](https://stackoverflow.com/questions/4059163/css-language-speak-container-vs-wrapper) for a while. I've been thinking of Redux containers as a special convention outside of these definitions. What do you think about these definitions?

src/elements/content-sidebar/activity-feed/comment/components/__tests__/BaseCommentMenu.test.js
jcamden: Ah, it was just for dev. I was recycling these temporarily. The exports and unused defaultProps have since been removed.

src/elements/content-sidebar/activity-feed/comment/components/__tests__/BaseCommentMenu.test.js
jcamden: I think I disagree. As I wrote this today, I recalled our conversation about this a couple years back. I agreed at the time that it was better to separate them; but when I thought of this `should` param this morning, I thought it was a pretty nice solution. It's not exactly fearsome logic :)

src/elements/content-sidebar/activity-feed/comment/components/stories/BaseCommentMenu.stories.js
jcamden: The main reason I created these stories---aside from just liking Storybook development---is because Preview kinks (like monoproxy) had not been worked out for dev pods. During planning, we were going to suspend working on these tickets, but I suggested we could go ahead using Storybook. I think a trail has since been blazed to use dev pods for Preview dev, and I plan to pick up that trail to better confirm these changes before I move this PR out of draft. But, in sum, the stories were the means by which I developed these components.

tsconfig.json
jcamden: Ah, that's just IDE shenanigans. Thanks for catching that.

src/elements/content-sidebar/activity-feed/comment/components/BaseCommentMenu.js
jcamden: Ah, you're right. The lone override in this class doesn't need to apply to the `TetherComponent` here, nor to the `DropdownMenu` in `MediaMenu`, so it's okay to just let `MediaMenu` forward those classes to the `PlanButton` (as it does).

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
jcamden: Maybe actually the better syntax would be `!!can_delete`, come to think of it.

src/elements/content-sidebar/activity-feed/comment/components/BaseCommentMenuWrapper.js
jstoffan: Yeah, naming is always squishy. SUIT throws a wrench in here, as well, since `-` is use to indicate that something is a descendant. Ideally, components would start with a base name, like `Comment`, which would be the "wrapper" and everything inside it (e.g. `Comment-text`) would be descendants.

It's not a major issue either way, so long as the conventions are followed otherwise.

src/elements/content-sidebar/activity-feed/comment/components/__tests__/BaseCommentMenu.test.js
jcamden: I was thinking about this. I definitely see the point of excluding, as much as possible, logic from tests since that new logic is then untested. Something as simple as this though, I'm on the fence about. I see the value in not wanting to set a precedent for untested logic in unit tests. But on the other hand, I like that these test are slightly easier to read and maintain since they aren't duplicated-ish. Both tradeoffs seem extremely minor, so kind of hard to decide. Could happily go your way :)

src/elements/content-sidebar/activity-feed/activity-feed/ActivityItem.scss
jcamden: Heads up, I have not handled this in this PR. Would it be alright to break this out into a separate task?

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -201,108 +153,41 @@ export const BaseComment = ({
...
         <div className="bcs-BaseComment">
-            <Media
+            <div
                 className={classNames('bcs-Comment-media', {
                     'bcs-is-pending': isPending || error,
                 })}
             >
-                <Media.Figure>
-                    <Avatar
+                <div className="bcs-BaseComment-header">
```
jcamden, line 161: Ultimately, this section came to have three top-level components ("metadata", menu, and status). As such, I think it'd probably be most ideal to break it out into a separate component. That said, lots of threaded replies work is lined up for the new sprint which starts today. In order to avoid those tasks from having to fork my fork, we're hoping to limit the scope for now.

src/elements/content-sidebar/activity-feed/comment/components/BaseCommentMetadata.js
jcamden: Surprisingly tough to think of a good name here. This is less than a citation but more than a byline. Eventually, I asked ChatGPT, but tragically, I think its best suggestion was metadata. Another decent candidate, but still not great, is info... It's generic enough to capture author, timestamp (and sometimes activity link); but it's also kind of too generic.

src/elements/content-sidebar/activity-feed/activity-feed/ActivityItem.scss
bfoxx1906: Yea I think that will work.

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
bfoxx1906: I agree. Let's limit the scope to what you have now. The biggest priority for us relative to the Boxer release are the UI updates.

src/elements/content-sidebar/activity-feed/activity-feed/stories/ActivityFeed.stories.js
```diff
@@ -0,0 +1,70 @@
...
+    TIME_STRING_SEPT_27_2017,
+    TIME_STRING_SEPT_28_2017,
+    currentUser,
+    comment,
+    replies,
+    repliesProps,
+} from '../../comment/stories/common';
+import { FEED_ITEM_TYPE_VERSION } from '../../../../../constants';
+
+import { type FeedItems, type Comment } from '../../../../../common/types/feed';
```
jcamden, line 17: nit: could combine to `import type {foo, bar}`; I'll implement this if there are other changes needed.

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -170,22 +131,13 @@ export const BaseComment = ({
...
     const onEdit = onAnnotationEdit ?? onCommentEdit;
 
-    const canDelete = permissions.can_delete;
-    const canEdit = onEdit !== noop && permissions.can_edit;
-    const canResolve = onEdit !== noop && permissions.can_resolve;
-    const createdAtTimestamp = new Date(created_at).getTime();
-    const createdByUser = created_by || PLACEHOLDER_USER;
+    const canDelete = !!permissions.can_delete;
+    const canEdit = onEdit !== noop && !!permissions.can_edit;
+    const canResolve = onEdit !== noop && !!permissions.can_resolve;
```
jcamden, line 139: Oops, that's a mistake. Only `canDelete` needed `!!`.

src/elements/content-sidebar/activity-feed/comment/components/BaseCommentMetadata.js
TylerGauntlett: Would `BaseCommentSupplementalData` be any better?

src/elements/content-sidebar/activity-feed/comment/components/BaseCommentMetadata.js
TylerGauntlett: Potentially "supplemental data" here too

TylerGauntlett: Potentially "supplemental data" here too

src/elements/content-sidebar/activity-feed/activity-feed/ActivityItem.scss
```diff
@@ -4,9 +4,9 @@
...
-            padding: $bdl-grid-unit * 2;
-            padding-right: $bdl-grid-unit * 4;
+            padding: 6px 12px $bdl-grid-unit * 5;
```
TylerGauntlett, line 9: Are there DSYS tokens we could use in place of the px value?

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -170,22 +131,13 @@ export const BaseComment = ({
...
-        } else if (onCommentEdit) {
-            onCommentEdit({ id, status: selectedStatus, hasMention: false, permissions });
-        }
-    };
-
     // Since we have to pass onCommentEdit through annotations (to Replies), onAnnotationEdit essentially overrides onCommentEdit
     const onEdit = onAnnotationEdit ?? onCommentEdit;
 
-    const canDelete = permissions.can_delete;
+    const canDelete = !!permissions.can_delete;
```
TylerGauntlett, line 137: Does the API return non-bool values? I've only seen bool permissions

src/elements/content-sidebar/activity-feed/comment/components/BaseCommentMetadata.scss
```diff
@@ -0,0 +1,38 @@
...
+.bcs-BaseCommentMetadata-avatar {
+    display: flex;
+    flex-direction: column;
+    height: 100%;
+}
+
+.bcs-BaseCommentMetadata-data {
+    display: flex;
+    flex-direction: column;
+    margin-left: 10px;
```
TylerGauntlett, line 24: Consider DSYS token here too (if available)

src/elements/content-sidebar/activity-feed/comment/components/__tests__/BaseCommentMenu.test.js
bfoxx1906: That's not a bad idea @patlm

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
jcamden: The type on the permissions object is `boolean | undefined`.
```
type BoxCommentPermission = {
    can_delete?: boolean,
    can_edit?: boolean,
    can_reply?: boolean,
    can_resolve?: boolean,
};
```
Turns out that's true for the other permissions too, so I've done that as well (below). I'm assuming it can be undefined because the object can get instantiated without or before a 200.

src/elements/content-sidebar/activity-feed/comment/components/BaseCommentMetadata.js
jcamden: That's not a bad idea. I'll try that refactor right now, and see if I can keep everything working :)

src/elements/content-sidebar/activity-feed/comment/components/BaseCommentMetadata.scss
jcamden: See my comment above :)

src/elements/content-sidebar/activity-feed/comment/components/__tests__/BaseCommentMenu.test.js
jcamden: @patim, that works well for cases where we're making assertions about the state of something that is rendered but not so well when we're asserting whether or not something is rendered at all, I think.

src/elements/content-sidebar/activity-feed/comment/components/__tests__/BaseCommentMenu.test.js
```diff
@@ -0,0 +1,85 @@
...
+                expect(wrapper.getByText(message)).toBeInTheDocument();
+            } else {
+                expect(wrapper.queryByText(message)).not.toBeInTheDocument();
```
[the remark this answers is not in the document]
jcamden: For exactly that reason, haha! If we use `getByText` in the second assertion, the test would fail.

src/elements/content-sidebar/activity-feed/comment/components/BaseCommentMetadata.js
jcamden: Actually, I think if we're going that generic, then `Info` is the same but more concise, and has the advantage of avoiding, `SupplementalData-data`. All pretty so-so names, but sometimes it's just hard to think of good ones.

bfoxx1906: What about just calling this BaseCommentHeader instead of BaseCommentMetadata. This component is a UI construct that's more just a collection of ui elements(not necessarily comment data) that goes at the top of a comment.

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
bfoxx1906: @jcamden the prop itself could be createdAt. The value doesn't though. We've used that other places in the code.

src/elements/content-sidebar/activity-feed/comment/BaseComment.scss
```diff
@@ -3,28 +3,51 @@
...
+        width: 100%;
+        min-height: $bdl-grid-unit * 12;
     }
 }
+
+.bcs-BaseComment-header {
+    display: flex;
+    justify-content: space-between;
+    width: 100%;
+    padding: 2;
```
mickr, line 47: ❓ Do we need a unit here?

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
jcamden: Right, just wondering, if we do refactor the prop name, where should we do the transform? Should we refactor the pre-existing prop in BaseComment? Should I only transform it in the new component that I created? Or should I transform at the parent of BaseComment or some other ancestor?

jcamden: My many questions about this probably reflect the fact that I lean towards keep the prop name consistent as it's being drilled. Imo, it's better to change it everywhere than the have suddenly transform at some arbitrary point.

src/elements/content-sidebar/activity-feed/comment/BaseComment.scss
jcamden: Apparently not, haha! Lemme look into this. Good catch.

jcamden: Comparing to designs, as far as I can tell, that shouldn't be there. Doesn't seem to be any padding that's unique to the header (although it's kinda hard to tell for sure...). I'll remove that 👍

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
bfoxx1906: I see your point. This is based on a type coming off of BaseFeedItem ready that was already there. One difference is that I don't think it's ever explicitly passed to a component. It's just  extracted via passed down props for various operations. I think in this case it makes sense to stay consistent with our conventions and just use createdAt={created_at}

TylerGauntlett (approved):

ahorowitz123 (approved):

src/elements/content-sidebar/activity-feed/comment/components/__tests__/BaseCommentMenu.test.js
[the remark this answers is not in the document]
tjuanitas: @jcamden you could do something like this:
```
test.each`
        canResolve | isResolved | exists
        ${true}    | ${true}    | ${true}
        ...
        ...
        expect(!!wrapper.queryByText(message)).toBe(exists);
```

not sure if it's frowned upon to use a `!!` in an assertion like that but would resolve this thread and thread above

PR #3407, opened 2023-09-21
feat(preview): Focus on comment form for replies and modifying

ahorowitz123:
- Used draft.js functionality in `EditorState` to force focus on the comment form when it is open from either initiating a new reply or modifying and existing comment/reply
- Added `shouldFocusOnOpen` prop to `CommentForm` which allows callers to opt in to this behavior

*Note regarding tests* -- While it would be better to simulate click events and check browser focus of the textbox element, the `EditorState.moveFocusToEnd` uses a lot of window functions which don't exist in our current version of jsdom. I tried upgrading jsdom which solved the issue but focusing on the text box still wasn't working correctly. Given that this function is part of draft.js, I opted to mock it and check that was called.

focus on comment form

### DIFF
--- a/src/elements/content-sidebar/activity-feed/comment-form/__tests__/CommentForm.test.js
+++ b/src/elements/content-sidebar/activity-feed/comment-form/__tests__/CommentForm.test.js
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { mount } from 'enzyme';
-import { render as renderRTLFunc } from '@testing-library/react';
+import { render } from '@testing-library/react';
 import { EditorState } from 'draft-js';
 
 import Button from '../../../../../components/button/Button';
@@ -18,12 +18,12 @@ describe('elements/content-sidebar/ActivityFeed/comment-form/CommentForm', () =>
         <CommentForm getMentionWithQuery={() => {}} intl={intlFake} user={{ id: 123, name: 'foo bar' }} {...props} />
     );
 
-    const render = props => mount(getComponent(props));
+    const getWrapper = props => mount(getComponent(props));
 
-    const renderRTL = props => renderRTLFunc(getComponent(props));
+    const getWrapperRTL = props => render(getComponent(props));
 
     test('should correctly render initial state', () => {
-        const wrapper = render();
+        const wrapper = getWrapper();
 
         expect(wrapper.find('[contentEditable]').length).toEqual(1);
         expect(wrapper.find('.bcs-CommentFormControls').length).toEqual(0);
@@ -33,7 +33,7 @@ describe('elements/content-sidebar/ActivityFeed/comment-form/CommentForm', () =>
     test('should call onFocus handler when input is focused', () => {
         const onFocusSpy = jest.fn();
 
-        const wrapper = render({ onFocus: onFocusSpy });
+        const wrapper = getWrapper({ onFocus: onFocusSpy });
 
         const mentionSelector = wrapper.find('[contentEditable]');
         mentionSelector.simulate('focus');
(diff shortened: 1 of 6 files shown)


JChan106 (commented):
Have we checked if this works with multiple comments and opening the reply form between them to make sure the focus switches properly?

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -302,22 +302,23 @@ export const BaseComment = ({
...
-                            isEditing={isEditing}
-                            entityId={id}
+                            shouldFocusOnOpen
                             tagged_message={tagged_message}
-                            getAvatarUrl={getAvatarUrl}
-                            mentionSelectorContacts={mentionSelectorContacts}
-                            getMentionWithQuery={getMentionWithQuery}
+                            updateComment={handleMessageUpdate}
+                            // $FlowFixMe
+                            user={currentUser}
```
jcamden, line 321: +1 for alphabetized props. Support it, but it is admittedly a little hard to read the diff. I can't see any changes, but I just wanted to confirm that.

src/elements/content-sidebar/activity-feed/comment/CreateReply.js
```diff
@@ -49,15 +49,16 @@ const CreateReply = ({
...
-                    showTip={false}
+                    mentionSelectorContacts={mentionSelectorContacts}
                     onCancel={onCancel}
                     onFocus={onFocus}
-                    createComment={handleSubmit}
-                    mentionSelectorContacts={mentionSelectorContacts}
-                    getMentionWithQuery={getMentionWithQuery}
                     placeholder={placeholder}
+                    shouldFocusOnOpen
+                    showTip={false}
```
jcamden, line 61: Same here: just alphabetization?

src/elements/content-sidebar/activity-feed/comment/__tests__/BaseComment.test.js
```diff
@@ -55,6 +55,10 @@ export const getWrapper = props =>
     );
 
 describe('elements/content-sidebar/ActivityFeed/comment/BaseComment', () => {
+    beforeEach(() => {
+        EditorState.moveFocusToEnd = editor => editor;
```
jcamden, line 59: Wh... wha? Just a function that returns its arg? Is this how you are clearing mocks in between tests?

jcamden (approved):
Looks good! Left some inconsequential comments.

src/elements/content-sidebar/activity-feed/comment-form/__tests__/CommentForm.test.js
```diff
@@ -12,15 +14,13 @@ const intlFake = {
...
-                {...props}
-            />,
-        );
+    const getComponent = (props = {}) => (
+        <CommentForm getMentionWithQuery={() => {}} intl={intlFake} user={{ id: 123, name: 'foo bar' }} {...props} />
+    );
+
+    const render = props => mount(getComponent(props));
+
+    const renderRTL = props => renderRTLFunc(getComponent(props));
```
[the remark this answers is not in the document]
ahorowitz123: Good call. I just copied what was already there but I'll update this

src/elements/content-sidebar/activity-feed/comment/CreateReply.js
ahorowitz123: Same as comment above

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
ahorowitz123: I only added `shouldFocusOnOpen`. The rest is just alphabetizing.

src/elements/content-sidebar/activity-feed/comment/__tests__/BaseComment.test.js
ahorowitz123: I guess I don't need to set this before every test and can just do it once. Good call.

But yeah this is how I'm mocking it. It needs to return an editor to get other tests to work. But also breaks in jsdom if we try to actually execute the real function

ahorowitz123:
> Have we checked if this works with multiple comments and opening the reply form between them to make sure the focus switches properly?

Yes since the function is called on component mount, every time a new component is mounted it gets focus

jcamden (approved):

joshmarnold (approved):

PR #3411, opened 2023-09-25
fix(activity-feed-zoom): add a min-height prop to the items container

dsporysz-box:
The activity feed with comments list doesn't scale with zoom correctly. At 400% zoom, the top header and the comment form cover the whole screen making the comment history not visible and it's not possible to scroll it into a view.

This PR adds a min-height property to the comments container so it does not get squished to zero height. The min-height is the same as the comment form.

Before:
zoom400before
After:
zoom400after

### DIFF
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
@@ -49,7 +49,7 @@
         flex: 1 1 auto;
         flex-direction: column;
         width: 100%; // Explicit width required since flex-direction is column
-        min-height: 93px;
+        min-height: $sidebarContentMinHeight;
         overflow-x: hidden;
         overflow-y: auto;
 
@@ -134,7 +134,7 @@
 
     .bcs-activity-feed-comment-input {
         flex: 0 0 auto;
-        min-height: 93px;
+        min-height: $sidebarContentMinHeight;
         max-height: 500px;
         overflow: hidden;
         background-color: $white;
--- a/src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
+++ b/src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
@@ -49,6 +49,7 @@
         flex: 1 1 auto;
         flex-direction: column;
         width: 100%; // Explicit width required since flex-direction is column
+        min-height: 93px;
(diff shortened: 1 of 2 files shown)


bkepka-box (dismissed):

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
```diff
@@ -49,6 +49,7 @@
         flex: 1 1 auto;
         flex-direction: column;
         width: 100%; // Explicit width required since flex-direction is column
+        min-height: 93px;
```
loonskai, line 52: May be a good candidate for a CSS variable

loonskai (dismissed):

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
dsporysz-box: Thanks, I added a var

loonskai (approved):

bkepka-box (approved):

tjuanitas (approved):

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
```diff
@@ -49,6 +49,7 @@
         flex: 1 1 auto;
         flex-direction: column;
         width: 100%; // Explicit width required since flex-direction is column
+        min-height: $sidebarContentMinHeight;
```
tjuanitas, line 52: I didn't notice this before but I see now there's two vertical scroll bars that would appear at the same time

multiple scroll bars in the same axis can have some unintuitive behavior to users, is that okay? is that an accessibility concern?

tjuanitas (commented):
sorry I had a thought come to mind after I approved

src/elements/content-sidebar/activity-feed/activity-feed/ActivityFeed.scss
dsporysz-box: @tjuanitas I've reached out to the Product Team and this solution has been accepted. Two scroll bars appear only at higher zoom and this solution lets us avoid impacting the broader user experience. Great attention to the details! Thanks!

tjuanitas: great, sounds good to me!

PR #3414, opened 2023-09-27
fix(preview): Ensure correct promoted by field is shown

joshmarnold:
"<Owner> promoted .." was getting displayed instead of "<Promoter> promoted ..". We want to ensure whoever promoted the current version is shown, not the owner.

File2023-09-27 at 14 30 06


src/elements/common/selectors/version.js
```diff
@@ -34,11 +34,12 @@ const getVersionAction = ({ restored_at, trashed_at, version_promoted }: $Shape<
 
 const getVersionUser = ({
     modified_by,
+    promoted_by,
     restored_by,
     trashed_by,
     uploader_display_name,
 }: $Shape<BoxItemVersion>): User => {
-    const { name, id, ...rest } = restored_by || trashed_by || modified_by || PLACEHOLDER_USER;
+    const { name, id, ...rest } = restored_by || trashed_by || promoted_by || modified_by || PLACEHOLDER_USER;
```
greg-in-a-box, line 42: I think we should update the test for this selector to include `promoted_by`.  Do we need any other test to ensure that `promoted_by` is being used on the sidebar?

greg-in-a-box (changes requested):
I think we need some tests added since this is a bug we are trying to fix and ensure it always correct.

src/utils/fields.js
```diff
@@ -150,6 +151,7 @@ const FILE_VERSION_FIELDS_TO_FETCH = [
     FIELD_MODIFIED_BY,
     FIELD_RESTORED_FROM,
     FIELD_SIZE,
+    FIELD_PROMOTED_BY,
```
tjuanitas, line 154: 🔤

src/utils/fields.js
```diff
@@ -11,6 +11,7 @@ import {
     FIELD_NAME,
     FIELD_TYPE,
     FIELD_SIZE,
+    FIELD_PROMOTED_BY,
```
tjuanitas, line 14: could you move to this the `[...]_BY` section below to keep the user fields together

```
FIELD_CREATED_BY,
FIELD_MODIFIED_BY,
FIELD_OWNED_BY,
FIELD_RESTORED_BY,
FIELD_TRASHED_BY,
```

src/utils/__tests__/fields.test.js
```diff
@@ -19,6 +19,7 @@ import {
     FIELD_PARENT,
     FIELD_EXTENSION,
     FIELD_PERMISSIONS,
+    FIELD_PROMOTED_BY,
```
tjuanitas, line 22: same comment about grouping with the other `[...]_BY` fields

src/constants.js
```diff
@@ -90,6 +90,7 @@ export const FIELD_DATE: 'date' = 'date';
 export const FIELD_NAME: 'name' = 'name';
 export const FIELD_TYPE = 'type';
 export const FIELD_SIZE: 'size' = 'size';
+export const FIELD_PROMOTED_BY = 'promoted_by';
```
tjuanitas, line 93: also near the other `[...]_BY`

joshmarnold: Good catch, didn't see this sorting. Thought it was random. Kinda weird way to order these. Would prefer alphabetically.

src/utils/__tests__/fields.test.js
joshmarnold: This list is sorted alphabetically

src/utils/fields.js
joshmarnold: Yeah, np

src/utils/fields.js
joshmarnold: To be fair, it's not truly alphabetical lol

src/elements/common/selectors/version.js
joshmarnold: Update the test for this code to include the promoted_by. I don't see anywhere else that needs updated

greg-in-a-box: I was thinking having a test at this level too https://github.com/box/box-ui-elements/blob/f950772922fd75089360de76b3ed39ecb035f678/src/elements/content-sidebar/versions/VersionsItem.js#L141-L146 because thats where its changing the sentence that will display the name you want to present

greg-in-a-box: assuming this is the component that is shown in your image in your description

joshmarnold: Included this file as well

greg-in-a-box (approved):

karelee7 (approved):

PR #3419, opened 2023-09-29
fix: disable copy in textInputWithCopyButton when button is disabled

kdesai-box:
Currently textInputWithCopyButton allows the user to use ctrl c to copy the text input even when button is disabled. This change adds a disableCopy prop that will completely disable the ability to copy when set to true.

### DIFF
--- a/src/components/text-input-with-copy-button/TextInputWithCopyButton.js
+++ b/src/components/text-input-with-copy-button/TextInputWithCopyButton.js
@@ -26,6 +26,8 @@ type Props = {
     buttonProps?: Object,
     buttonSuccessText?: string | React.Node,
     className: string,
+    /** Disables copy function when copyDisabled is true */
+    copyDisabled?: boolean,
     disabled?: boolean,
     /** Label displayed for the text input */
     // TODO: Make label required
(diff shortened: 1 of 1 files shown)


tjuanitas:
hi @kdesai-box!

does the `disabled` prop not work for this use case? https://github.com/box/box-ui-elements/blob/master/src/components/text-input-with-copy-button/TextInputWithCopyButton.js#L29

Screenshot 2023-09-29 at 4 49 17 PM

btw the blueprint team is happy to work with you if you're interested in a blueprint equivalent

kdesai-box:
hi @tjuanitas 
I was advised not to use the disabled prop to fix this ctrl c bug, however it seems to work fine and pass all tests!

tjuanitas (commented):
interesting bug! I didn't realize that a `div` could have an `onCopy` event. one minor comment but cool find 👍 

> I was advised not to use the disabled prop to fix this ctrl c bug

any particular reason not to? maybe it's still a valid concern

kdesai-box:
> interesting bug! I didn't realize that a `div` could have an `onCopy` event. one minor comment but cool find 👍
> 
> > I was advised not to use the disabled prop to fix this ctrl c bug
> 
> any particular reason not to? maybe it's still a valid concern

We weren't sure how this would effect other projects using this particular button and thought it best to just create a new prop for our specific use. However, now that I look at it again if the intent is that the button should be completely disabled its probably best to use the original disable prop.

tjuanitas (approved):
lgtm! I would update the PR title and use `fix` instead of `feat` now

also the repo requires two approvals, if you have a teammate already part of the public Box org they can review

jsenar (approved):
lgtm

PR #3420, opened 2023-09-29
feat(content-answers): intelligence api

greg-in-a-box:


src/api/Intelligence.js
```diff
@@ -0,0 +1,69 @@
...
+ * @file Helper for the Box AI API endpoint
+ * @author Box
+ */
+
+import Base from './Base';
+import { ERROR_CODE_INTELLIGENCE } from '../constants';
+import type { BoxItem } from '../common/types/core';
+import type { ElementsErrorCallback } from '../common/types/api';
+
+class Intelligence extends Base {
```
megansmith-box, line 12: I see other recent Apis extend `MarkerBasedApi` which extends `Base` , do you think we'd ever need that functionality for Intelligence API?

src/api/Intelligence.js
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * @property {Function}
+     */
+    errorCallback: ElementsErrorCallback;
+
+    /**
+     * API endpoint to ask ai a question
+     *
+     * @param {string} prompt - Question
+     * @param {Array<object>} items - Array of items to ask about
```
megansmith-box, line 27: Are we missing the last 2 params here?

src/api/Intelligence.js
```diff
@@ -0,0 +1,69 @@
...
+        const body = {
+            data: {
+                mode: 'single_item_qa',
+                prompt,
+                items,
+            },
+        };
+
+        this.successCallback = successCallback;
+        this.errorCallback = errorCallback;
```
megansmith-box, line 56: Why do these need to be set as variables of the class?

megansmith-box: Should we also validate that successCallback and errorCallback are valid or have a default value?

src/api/Intelligence.js
```diff
@@ -0,0 +1,69 @@
...
+
+        await this.xhr
+            .post({
+                url,
+                id: items[0].id,
+                data: body,
+            })
+            .then(this.successCallback)
+            .catch(this.errorHandler);
```
megansmith-box, line 65: should this be errorCallback?

src/api/Intelligence.js
```diff
@@ -0,0 +1,69 @@
...
+            },
+        };
+
+        this.successCallback = successCallback;
+        this.errorCallback = errorCallback;
+
+        await this.xhr
+            .post({
+                url,
+                id: items[0].id,
```
megansmith-box, line 61: Why is id = to the first items id?

src/api/Intelligence.js
```diff
@@ -0,0 +1,69 @@
...
+                data: body,
```
megansmith-box, line 62: This seems kind of strange, should it be `data.data`?

src/api/Intelligence.js
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * API endpoint to ask ai a question
+     *
+     * @param {string} prompt - Question
+     * @param {Array<object>} items - Array of items to ask about
+     * @return {Promise}
+     */
+    async ask(prompt: string, items: Array<BoxItem>, successCallback: Function, errorCallback: ElementsErrorCallback) {
+        this.errorCode = ERROR_CODE_INTELLIGENCE;
```
megansmith-box, line 31: Why is this error code set here like this?

src/api/Intelligence.js
greg-in-a-box: the errorHandler calls errorCallback

src/api/Intelligence.js
greg-in-a-box: pattern is used in other clients, its been set for the base class to use

src/api/Intelligence.js
greg-in-a-box: I honestly dont know why it even needs the id, its a requirement of the the xhr.post, but the items array will always have one item for now till this client is used for multi. items is required by the intelligence request it handles multi only

src/api/Intelligence.js
greg-in-a-box: this.errorCode is used by the base class to be applied when the errorHandler is being used

src/api/Intelligence.js
greg-in-a-box: I dont think so

src/api/Intelligence.js
greg-in-a-box: if anything I think I should be using `successHandler`

src/api/Intelligence.js
greg-in-a-box: https://github.com/box/box-ui-elements/blob/master/src/api/Base.js#L182
https://github.com/box/box-ui-elements/blob/master/src/api/Base.js#L194
Base class is handling the validation of those callbacks

src/api/Intelligence.js
greg-in-a-box: looked it up, it seems to be using the id to get a writetoken to do whatever it needs to do

TaoJiang98: yeah since the current mode we are using is 'single_item_qa', the array item will always has one item only.

TaoJiang98 (approved):
lgtm!

src/api/Intelligence.js
megansmith-box: [nit] should we validate that items length is always 1?

megansmith-box (approved):

bfoxx1906 (approved):
previous reviews make sense

PR #3423, opened 2023-10-02
feat(threaded-replies): added resolve/unresolve for annotations

TylerGauntlett:
### Overview
- Allows users to resolve/unresolve annotations when they have correct permissions
- Adds two new types for `onStatusChange` events (`OnCommentStatusChange` and `OnCommentStatusChange`)
- Refactored default properties in `OnCommentStatusChange` to be automatically appended in from the parent component instead of conditionally added in the child
- Updating status of comments was refactored but functionally unchanged

### Demo
Screen Recording 2023-10-03 at 10 36 52 am

### Out of Scope
- Annotations are currently returning `can_resolve = false` regardless of comment owner on resolution. A separate ticket is tracking this bug

### DIFF
--- a/src/elements/content-sidebar/ActivitySidebar.js
+++ b/src/elements/content-sidebar/ActivitySidebar.js
@@ -73,7 +73,7 @@ import type { Translations } from './flowTypes';
 import type { FeatureConfig } from '../common/feature-checking';
 import './ActivitySidebar.scss';
 
-import { type OnAnnotationEdit } from './activity-feed/comment/types';
+import type { OnAnnotationEdit, OnAnnotationStatusChange } from './activity-feed/comment/types';
 
 type ExternalProps = {
     activeFeedEntryId?: string,
@@ -224,7 +224,7 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         this.fetchFeedItems();
     };
 
-    handleAnnotationStatusChange = (id: string, status: FeedItemStatus, permissions: AnnotationPermission) => {
+    handleAnnotationStatusChange: OnAnnotationStatusChange = ({ id, permissions, status }) => {
         const { api, emitAnnotationUpdateEvent, file } = this.props;
 
         emitAnnotationUpdateEvent({ id, status }, true);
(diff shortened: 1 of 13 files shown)


src/elements/content-sidebar/activity-feed/comment/types.js
```diff
@@ -12,3 +18,10 @@ export type OnCommentEdit = (args: {
     status?: FeedItemStatus,
     text?: string,
 }) => void;
+
+export type OnCommentStatusChange = (args: {
+    hasMention?: boolean,
```
TylerGauntlett, line 23: This was done to make the types work correctly. Refactor so types work more intentionally, potentially deduplicating `OnAnnotationStatusChange` and `OnCommentStatusChange`

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -224,7 +224,7 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         this.fetchFeedItems();
     };
 
-    handleAnnotationStatusChange = (id: string, status: FeedItemStatus, permissions: AnnotationPermission) => {
+    handleAnnotationStatusChange: OnAnnotationStatusChange = ({ id, permissions, status }) => {
         const { api, emitAnnotationUpdateEvent, file } = this.props;
```
bfoxx1906, line 228: I'd recommend not using this and doing the same way you did for onCommentStatusChangeHandler in ActiveState. The folks from Poland put this here before for their ActivityThread implementation but that won't be used anymore once we push out our code. It's best stay as consistent as possible between annotations and comments functionality. What are your thoughts on this?

TylerGauntlett: Im open to going that route. If I do that, I think the `handleAnnotationStatusChange` and `handleAnnotationEdit` should be collapsed into one function. I can conditionally build that `description` object when `text` is passed in. If they are not combined back together, I think it makes more sense to use the function that specifically does this.

TylerGauntlett: After discussion, we will proceed with the current code structure in the scope of this PR. A new ticket will be opened to handle standardising comments and annotations across all component.

bfoxx1906 (approved):
@TylerGauntlett will create a new ticket to address some of the architecture issues we discussed.

shahzadaziz (approved):

PR #3424, opened 2023-10-02
chore(mergify): auto moji

greg-in-a-box:
I put it through the mergify config editor and it reformatted the file


tjuanitas (dismissed):
lgtm

.mergify.yml
```diff
@@ -1,28 +1,41 @@
...
-        name: default
-  - name: Automatic head deletion
-    conditions: []
-    actions:
-      delete_head_branch: {}
+    - name: Automatic boxmoji merge
+      conditions:
+          - author=boxmoji
+          - base=master
+          - files~=^[a-z]{2}-[A-Z]{2}\.properties$
```
benjamin-shen, line 13: Is the matcher supposed to match the full file path, including directory names (i.e. files should start with `i18n/`)? or is it just the filename?

.mergify.yml
```diff
@@ -1,28 +1,41 @@
...
+          - status-success="Validate"
+          - status-success=license/cla
+          - status-success=lint_test_build
+          - title~=^(fix)\(118n\)?:\supdate translations$
```
benjamin-shen, line 17:
```suggestion
          - title~=^(fix)\(i18n\)?:\supdate translations$
```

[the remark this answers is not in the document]
tjuanitas: hmm, on second thought. maybe the current space is correct because we don't want to allow tabs or something in that string?

.mergify.yml
greg-in-a-box: no examples given 
image

.mergify.yml
greg-in-a-box: fixed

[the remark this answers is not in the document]
greg-in-a-box: undone, mergify seems to be real fickle about the formatting for regex in that title slot...i even tried removing the other `\s` but it didnt like that

PR #3426, opened 2023-10-03
feat(threaded-replies): reload sidebar for filters

karelee7:
### Overview
* Activity items in the preview sidebar are now being fetched via UAA when using the filter
* Update existing unit tests and added new unit tests

### Demo
**Example flow of using the "Unresolved" filter:**
Screenshot 2023-10-03 at 3 28 14 PM
Screenshot 2023-10-03 at 3 28 46 PM
Screenshot 2023-10-03 at 3 28 58 PM

**Video demo of all filters:** 
https://github.com/box/box-ui-elements/assets/95440409/c934d82b-208e-492d-a591-94f9f6d8df34

**Example of concurrent comment adding on the same file:**
Logged in as User 1
Screenshot 2023-10-04 at 12 54 54 PM

Logged in and added comment as User 2
Screenshot 2023-10-04 at 12 55 12 PM

Apply "Unresolved" filter as User 1
Screenshot 2023-10-04 at 1 02 16 PM


src/elements/content-sidebar/__tests__/ActivitySidebar.test.js
```diff
@@ -1496,35 +1501,220 @@ describe('elements/content-sidebar/ActivitySidebar', () => {
...
+            restored_at: null,
+            restored_by: null,
+            retention: null,
+            size: 4159,
+            trashed_at: null,
+            trashed_by: null,
+            type: 'file_version',
+            uploader_display_name: 'John Doe',
+            version_number: undefined,
+        };
```
patlm, line 1700: These expected values look like they may be copied from the fixture. I wonder if instead of copying, if the expected variables here could be set as deep clones from the fixture? I think this would help with duplication while avoiding the concerns with potential side effects and mutability of the objects.

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -1108,26 +1114,61 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+            }
+
+            // Filter is completed on two properties (status and type) because filtering on comments (resolved vs. unresolved)
+            // requires looking at item status to see if it is open or resolved. Item type is also used to keep versions in the feed.
+            // Task also has a status but it's status will be "NOT_STARTED" or "COMPLETED" so it will not conflict with comment's status.
+            const filteredActivity = parsedFeedItems.filter(
+                item =>
+                    item.status === feedItemsStatusFilter ||
+                    item.type === FEED_ITEM_TYPE_VERSION ||
+                    item.type === feedItemsStatusFilter,
```
patlm, line 1168: I think it makes sense why the filtering here is still included in this PR. However, I think it would be helpful to evaluate what changes would need to be made to the API or code here to be able to pass the feed items directly from the API.

karelee7: Yeah, I believe that the api currently only takes the above params; I think the status is a param in the endpoint at the moment (see [here](https://git.dev.box.net/AppServices/end-user-web/blob/93bf1d03c92b2c64596787d6cc07114ceb12e395/schemas/v2/m0/endpoints/uaa/getActivityForFileId.json#L88)), but it's not attached to anything so at the moment, it doesn't do anything. @bfoxx1906 Do you know more about this endpoint? I think we can change the endpoint if needed, but it's probably out of scope for this PR

patlm: Yeah, I agree that it is out of scope. I looked into the status param on Friday and saw it wasn't attached.

karelee7: Or @Priyanka-Mudiganti do you know what else we would have to change on the endpoint to get what we need directly from the api?

src/elements/content-sidebar/__tests__/ActivitySidebar.test.js
karelee7: Hm yeah, let me see if I can deep clone these specific fields from the fixture; if not, I can try copying it over from the fixture to help with duplication!

karelee7: Hmm, have yet to find a way too deep clone specific properties from an object; unfortunately some of the items aren't exactly the same as the ones in the fixture - if mutability is a concern, do you think something like this would suffice?
```
const expectedAnnotationOpen = {
   created_at: annotationOpen.source.annotation.created_at,
   created_by: annotationOpen.source.annotation.created_by,
   id: annotationOpen.source.annotation.id,
   modified_at: annotationOpen.source.annotation.modified_at,
   permissions: annotationOpen.source.annotation.permissions,
   status: annotationOpen.source.annotation.status,
 };

karelee7: As discussed, using something like `expectedVersionItem = cloneDeep(filterableActivityFeedItems.versionItem.source)` and cloning specific properties worked for the tests!

patlm (approved):

PR #3436, opened 2023-10-09
fix(presence): UI not showing relative time correctly

dadu109:


tjuanitas (commented):
couple questions:

- the issue in Notes only recently started happening but this code hasn't been changed in years, why are we seeing this now?

- the Presence feature is also used in the main webapp, is the issue reproducible there? how will these changes affect that experience?

im wondering if this is an issue with Notes rather than an issue with the components

src/features/presence/utils/__tests__/presenceUtils.test.js
```diff
@@ -157,4 +157,57 @@ describe('features/presence/utils/sortByActivity', () => {
...
+            },
+            {
+                ms: 1000 * 60 * 60 * 24 * 30,
+                expected: { unit: 'month', value: 1 },
+            },
+            {
+                ms: 1000 * 60 * 60 * 24 * 30 * 10,
+                expected: { unit: 'month', value: 10 },
+            },
+        ].forEach(({ ms, expected }) => {
```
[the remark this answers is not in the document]
rlyonbox: @tjuanitas the issue started for Notes when we upgraded react-intl from 2.x to 6.x, which [included a breaking change to the formatRelative/formatRelativeTime APIs](https://github.com/formatjs/formatjs/blob/main/website/docs/react-intl/upgrade-guide-3.x.md).

EUA still uses 2.9.0 so this issue will occur there when they upgrade as well.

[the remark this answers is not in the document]
tjuanitas: okay that makes sense. I see the issue happening here now: https://opensource.box.com/box-ui-elements/#/Features?id=presence

src/features/presence/PresenceAvatarTooltipContent.tsx
```diff
@@ -14,11 +14,12 @@ export type Props = {
 } & WrappedComponentProps;
 
 function PresenceAvatarTooltipContent({ name, interactedAt, interactionType, intl, isActive }: Props): JSX.Element {
-    const lastActionMessage = determineInteractionMessage(interactionType);
+    const lastActionMessage = determineInteractionMessage(interactionType, interactedAt);
+    const { unit, value } = convertMillisecondsToUnitAndValue(Math.abs(interactedAt - Date.now()));
```
tjuanitas, line 18: maybe swap the order?
```suggestion
    const { unit, value } = convertMillisecondsToUnitAndValue(Date.now() - interactedAt);
```

then you don't need the Math.abs since Date.now() should always be >= than interactedAt

src/features/presence/utils/presenceUtils.js
```diff
@@ -60,4 +64,24 @@ function determineInteractionMessage(interactionType, interactedAt) {
     return message;
 }
 
-export { sortByActivity, getLastActionTimeMS, determineInteractionMessage };
+function convertMillisecondsToUnitAndValue(ms) {
+    const seconds = ms / 1000;
```
tjuanitas, line 68: could we keep it in milliseconds instead of converting to seconds? I think for consistency since the other functions use milliseconds

as well as the constants can stay in the same unit:

```
const MS_IN_A_SECOND = 1000;
const MS_IN_A_MINUTE = MS_IN_A_SECOND * 60;
const MS_IN_A_HOUR = MS_IN_A_MINUTE * 60;
const MS_IN_A_DAY = MS_IN_A_HOUR * 24;
const MS_IN_A_MONTH = MS_IN_A_DAY * 30;
```

src/features/presence/utils/presenceUtils.js
```diff
@@ -60,4 +64,24 @@ function determineInteractionMessage(interactionType, interactedAt) {
...
+    const abs = Math.abs(seconds);
```
tjuanitas, line 69: might be redundant to do an absolute check here since there's one when the function is called

src/features/presence/utils/presenceUtils.js
```diff
@@ -60,4 +64,24 @@ function determineInteractionMessage(interactionType, interactedAt) {
...
+        return { unit: 'second', value: seconds };
+    }
+    if (abs < SEC_IN_HOUR) {
+        return { unit: 'minute', value: Math.round(seconds / SEC_IN_MINUTE) };
+    }
+    if (abs < SEC_IN_DAY) {
+        return { unit: 'hour', value: Math.round(seconds / SEC_IN_HOUR) };
+    }
+
+    if (abs < SEC_IN_MONTH) {
```
tjuanitas, line 80: months don't share the same amount of days so this calculation wouldn't be technically accurate. is that okay? is this behavior different than the react-intl 2.x behavior?

src/features/presence/utils/presenceUtils.js
```diff
@@ -60,4 +64,24 @@ function determineInteractionMessage(interactionType, interactedAt) {
...
+        return { unit: 'day', value: Math.round(seconds / SEC_IN_DAY) };
+    }
+
+    return { unit: 'month', value: Math.round(seconds / SEC_IN_MONTH) };
```
tjuanitas, line 84: do we need to handle years?

src/features/presence/utils/presenceUtils.js
```diff
@@ -60,4 +64,24 @@ function determineInteractionMessage(interactionType, interactedAt) {
...
+        return { unit: 'minute', value: Math.round(seconds / SEC_IN_MINUTE) };
+    }
+    if (abs < SEC_IN_DAY) {
+        return { unit: 'hour', value: Math.round(seconds / SEC_IN_HOUR) };
+    }
+
+    if (abs < SEC_IN_MONTH) {
+        return { unit: 'day', value: Math.round(seconds / SEC_IN_DAY) };
+    }
+
```
tjuanitas, line 83: super small nit:
could we make the white space around this `if` block be consistent the other `if` blocks in this function?

soygitana:
Hi all, 
Thanks for catching and addressing this issue. I was aware of the change in the way relative time is handled (see [documentation](https://formatjs.io/docs/react-intl/upgrade-guide-3x#formattedrelativetime)). I have made the necessary changes in the upgraded repositories that use relative time, for example, in the [Canvas repository](https://git.dev.box.net/Canvas/canvas-frontend/pull/1557/files#diff-2380ecc7f1bd51aff663714fa4bb044d3097baba229d1fa592246bc63bfee5b2). However, I didn't realize that the UI element used in the upgraded repository would also require an upgraded version of react-intl. 
Not all repositories have been upgraded yet, so we will need to provide support for both versions below 3 and versions 3 and above. Once all repos are upgraded, I will upgrade BUIE and support for version below 3 will be removed.

soygitana:
@dadu109 This PR can be closed in favor of #3466

soygitana:
Closed in favor of https://github.com/box/box-ui-elements/pull/3466

PR #3439, opened 2023-10-11
fix(activity-thread): Hide Modify option on comment if resolved

shahzadaziz:

### DIFF
--- a/src/elements/content-sidebar/activity-feed/comment/components/BaseCommentMenu.js
+++ b/src/elements/content-sidebar/activity-feed/comment/components/BaseCommentMenu.js
@@ -88,7 +88,7 @@ export const BaseCommentMenu = ({
                         <FormattedMessage {...messages.commentResolveMenuItem} />
                     </MenuItem>
                 )}
-                {canEdit && (
+                {canEdit && !isResolved && (
                     <MenuItem
                         data-resin-target={ACTIVITY_TARGETS.COMMENT_OPTIONS_EDIT}
                         data-testid="edit-comment"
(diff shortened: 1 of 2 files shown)


src/elements/content-sidebar/activity-feed/comment/components/BaseCommentMenu.js
```diff
@@ -88,7 +88,7 @@ export const BaseCommentMenu = ({
                         <FormattedMessage {...messages.commentResolveMenuItem} />
                     </MenuItem>
                 )}
-                {canEdit && (
+                {canEdit && !isResolved && (
```
JChan106, line 91: Is it not possible to have a backend change for this? If canEdit is false if the status is resolved on the backend, then we wouldn't need this extra flag on the FE.

shahzadaziz: Ya I have been looking into EUW endpoints and extending permissions to add `isResolved` to both top level comment and replies. That seems like a better place put these checks. 

That work is currently blocked since EUW local dev setup is broken but we can disregard this PR if that is the direction we agree on.

JChan106: Yeah I agree with the direction of extending permissions. This will also prevent users from bypassing the ui and hitting the endpoint directly.

patlm: On the back-end canEdit should already be checking resolution status for top level and replies. I would be curious why it isn't coming through.

JChan106: @shahzadaziz Before your changes, did the menu item to modify comment show up when the comment is resolved?

PR #3440, opened 2023-10-12
feat(pill-selector): added more options for text input

AsemK:
This PR adds two new props to the `PillSelector` component. The two props are needed for the `PillSelectorDropdown` component. For now, they can be supplied to `PillSelector` through the `inputProps` prop.
## `isInputDisabled`
Allows disabling the `textarea` element without disabling the whole `PillSelector`. Usecase: our application needs to be able to set a maximum number of pills so that the user cannot add more than this number but can add less. The plan is to use this prop to disable the input once the maximum number of pills is reached. The user cannot then add new pills, but can still remove existing pills. Once the number of pills goes below the maximum, we enable the input again.

## `isInputFocusedNextLine`
Gives the `textarea` its own separate line when the user focuses the component to input text. The new line disappears when you focus away. This is very useful when the component is used in a limited width (as the case in our application). See the GIFs below:

| Before | After |
| ------ | ----- |
| before | after |

### DIFF
--- a/src/components/pill-selector-dropdown/PillSelector.js
+++ b/src/components/pill-selector-dropdown/PillSelector.js
@@ -31,6 +31,10 @@ type Props = {
     getPillImageUrl?: (data: { id: string | number, [key: string]: any }) => string,
     innerRef?: React.Ref<any>,
     inputProps: Object,
+    /** Allows disabling the textarea element without disabling the whole PillSelector */
+    isInputDisabled?: boolean,
+    /** Whether to show textarea in next line when focused */
+    isInputFocusedNextLine?: boolean,
     onInput: Function,
     onRemove: Function,
     onSuggestedPillAdd?: Function,
@@ -194,6 +198,8 @@ class PillSelectorBase extends React.Component<Props, State> {
             getPillClassName,
             getPillImageUrl,
             inputProps,
+            isInputDisabled,
+            isInputFocusedNextLine,
             onInput,
             onRemove,
             onSuggestedPillAdd,
@@ -284,8 +290,10 @@ class PillSelectorBase extends React.Component<Props, State> {
                         {...rest}
                         {...inputProps}
                         autoComplete="off"
-                        className={classNames('bdl-PillSelector-input', 'pill-selector-input', className)}
-                        disabled={disabled}
+                        className={classNames('bdl-PillSelector-input', 'pill-selector-input', className, {
+                            'bdl-PillSelector-input--nextLine': isInputFocusedNextLine,
+                        })}
+                        disabled={disabled || isInputDisabled}
                         onInput={onInput}
                         placeholder={this.getNumSelected() === 0 ? placeholder : ''}
                         ref={input => {
--- a/src/components/pill-selector-dropdown/PillSelector.scss
+++ b/src/components/pill-selector-dropdown/PillSelector.scss
@@ -50,6 +50,10 @@
             border-color: $primary-color;
             outline: 0;
             box-shadow: none;
+
+            .bdl-PillSelector-input--nextLine {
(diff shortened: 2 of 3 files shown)


src/components/pill-selector-dropdown/PillSelector.js
```diff
@@ -30,6 +30,9 @@ type Props = {
     /** Function to retrieve the image URL associated with a pill */
     getPillImageUrl?: (data: { id: string | number, [key: string]: any }) => string,
     innerRef?: React.Ref<any>,
+    inputDisabled?: boolean,
```
Multimo, line 33:
```suggestion
    /** Allows disabling the textarea element without disabling the whole PillSelector */
    inputDisabled?: boolean,
```

Multimo (approved):
Added a small comment, might help someone looking at the code see the different between inputDisabled and disabled :)

src/components/pill-selector-dropdown/PillSelector.js
tjuanitas, line 33: we should use an is / can / has prefix for booleans
```suggestion
    isInputDisabled?: boolean,
```

src/components/pill-selector-dropdown/PillSelector.js
```diff
@@ -284,8 +289,10 @@ class PillSelectorBase extends React.Component<Props, State> {
                         {...rest}
                         {...inputProps}
                         autoComplete="off"
-                        className={classNames('bdl-PillSelector-input', 'pill-selector-input', className)}
-                        disabled={disabled}
+                        className={classNames('bdl-PillSelector-input', 'pill-selector-input', className, {
+                            'input-in-new-line': inputInNewLine,
```
tjuanitas, line 293: new classnames should follow SUIT naming:

https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md

this could be a modifier:
bdl-PillSelector-input--nextLine

src/components/pill-selector-dropdown/PillSelector.scss
```diff
@@ -50,6 +50,10 @@
             border-color: $primary-color;
             outline: 0;
             box-shadow: none;
+
+            .bdl-PillSelector-input.input-in-new-line {
```
tjuanitas, line 54: we should decrease the specificity if possible. otherwise it could be hard to override in the future

src/components/pill-selector-dropdown/__tests__/PillSelector.test.js
```diff
@@ -41,6 +41,13 @@ describe('components/pill-selector-dropdown/PillSelector', () => {
             expect(wrapper).toMatchSnapshot();
         });
 
+        test('should render with disabled input', () => {
+            const wrapper = shallow(<PillSelector inputDisabled onInput={() => {}} onRemove={() => {}} />);
```
tjuanitas, line 45: what about using the existing stubs for these callbacks?

src/components/pill-selector-dropdown/__tests__/PillSelector.test.js
```diff
@@ -41,6 +41,13 @@ describe('components/pill-selector-dropdown/PillSelector', () => {
...
+            const input = wrapper.find('textarea');
+
+            expect(input.prop('disabled')).toBeTruthy();
```
tjuanitas, line 48: we can be more explicit
```suggestion
            expect(input.prop('disabled')).toBe(true);
```

AsemK:
@Multimo @tjuanitas 
Thanks for the review! I made all the proposed changes. Please review again.

Multimo (approved):

greg-in-a-box (approved):

PR #3442, opened 2023-10-12
feat(content-answers): add scrolling to bottom

TaoJiang98:
The modal content will scroll all the way to the bottom when

- opens the modal
- submit a question and the answer is loading
- the answer finishes loading and generated completely


greg-in-a-box (changes requested):

src/features/content-answers/Answer.tsx
```diff
@@ -7,10 +7,17 @@ import './Answer.scss';
...
     isLoading: boolean;
 };
 
-const Answer = ({ answer, isLoading }: Props) => {
+const Answer = ({ answer, handleScrollToBottom, isLoading }: Props) => {
+    useEffect(() => {
+        if (answer) {
+            handleScrollToBottom('smooth');
+        }
+    }, [answer, handleScrollToBottom]);
```
megansmith-box, line 19: I feel like `handleScrollToBottom` should not be in the dependency array?

src/features/content-answers/ContentAnswersModalChat.tsx
```diff
@@ -0,0 +1,47 @@
...
+};
+
+const ContentAnswersModalChat = ({ currentUser, fileName, handleScrollToBottom, isLoading, questions }: Props) => {
+    const throttledScrollToBottom = throttle(handleScrollToBottom, 200);
+
+    return (
+        <div className="bdl-ContentAnswersModalChat">
+            <WelcomeMessage fileName={fileName} />
+            <ul>
+                {questions &&
```
megansmith-box, line 29: questions.length?

src/features/content-answers/ContentAnswersModalContent.tsx
```diff
@@ -1,38 +1,77 @@
...
+        if (contentRef.current) {
+            const { scrollTop, clientHeight, scrollHeight } = contentRef.current;
+            const threshold = scrollHeight - (scrollTop + clientHeight);
+            userScrolled.current = threshold > 5;
+        }
+    };
+
+    // Scroll to the bottom on chat load
+    React.useEffect(() => {
+        setTimeout(() => {
```
megansmith-box, line 46: Why setTimeout?

src/features/content-answers/ContentAnswersModalContent.tsx
```diff
@@ -1,38 +1,77 @@
...
+        }, 0);
+    }, [handleScrollToBottom]);
+
+    // Scroll to the bottom if a new answer is loading
+    React.useEffect(() => {
+        if (isLoading) {
+            userScrolled.current = false;
+            handleScrollToBottom('smooth');
+        }
+    }, [handleScrollToBottom, isLoading]);
```
megansmith-box, line 57: I don't believe we should pass handleScrollToBottom to the dependency array?

src/features/content-answers/ContentAnswersModalFooter.tsx
```diff
@@ -46,6 +46,8 @@ const ContentAnswersModalFooter = ({ currentUser, intl, isLoading, onAsk }: Prop
     const handleOnAsk = useCallback(() => {
         if (!isSubmitDisabled && prompt) {
             onAsk(prompt);
+            setPrompt('');
```
megansmith-box, line 49: Can we remove these two lines from `handleKeyDown `?

src/features/content-answers/ContentAnswersModalContent.tsx
```diff
@@ -1,38 +1,77 @@
...
+    currentUser?: User;
     fileName: string;
     isLoading: boolean;
     questions: QuestionType[];
 };
 
 const ContentAnswersModalContent = ({ currentUser, fileName, isLoading, questions }: Props) => {
+    const contentRef = React.useRef<HTMLDivElement | null>(null);
+    const messagesEndRef = React.useRef<HTMLDivElement | null>(null);
+    const userScrolled = React.useRef(false);
```
megansmith-box, line 24: Where is userScrolled added as a ref to an element?

src/features/content-answers/ContentAnswersModalContent.tsx
```diff
@@ -1,38 +1,77 @@
...
-import './ContentAnswersModalContent.scss';
+export type QuestionType = {
+    prompt: string;
+    answer?: string;
+    createdAt?: string;
+};
 
 type Props = {
-    currentUser: User;
+    currentUser?: User;
```
megansmith-box, line 15: I believe currentUser will always be there, it's not optional?

src/features/content-answers/Answer.tsx
TaoJiang98: There is a rule in react that we need to include every variables and functions used inside of useEffect to be included inside of the dependency array. And here to prevent infinitely re-rendering of calling handleScrollToBottom function, we used useCallback to keep the identity same in each render

src/features/content-answers/ContentAnswersModalChat.tsx
TaoJiang98: I simplified the condition check to `questions?.map`. But I think we need to leave the <ul> tag outside since we don't want to map on that and instead we want to map <li> tag for each pair of question and answer inside of the <ul> tag.

src/features/content-answers/ContentAnswersModalContent.tsx
TaoJiang98: There is [a known issue](https://stackoverflow.com/questions/71181018/scrollintoview-doesnt-scroll-anywhere) for scrollingIntoView function that will not scroll to anywhere when it is invoked before the changed state is rendered. And by using setTimeout here, it would help to put handleScrollToBottom function at the end of the event loop which prevents this above concern.

src/features/content-answers/ContentAnswersModalContent.tsx
TaoJiang98: I don't think that we need userScrolled here so I removed it. And I will push a new commit without it.

src/features/content-answers/ContentAnswersModalContent.tsx
TaoJiang98: I think that could be a chance that the fetchCurrentUser function failed and maybe that is reason why the currentUser is set to optional. And I checked another usage in buie which also sets to optional.
https://github.com/box/box-ui-elements/blob/22bb03133fd84b13400757f78e00536a5461c648/src/elements/content-sidebar/Sidebar.js#L36

greg-in-a-box: when you initially fetch for the current user while youre loading the current user its not there yet thats why it has to be optional otherwise exceptions will throw because its trying to use it when still not available because the request hasnt returned yet

src/features/content-answers/ContentAnswersModalContent.tsx
greg-in-a-box: when you use useRef, it doesnt need to ref an element, just a value that doesnt require rendering, so in this case its keeping track of the users position of the scrollbar so it can check if youre at the bottom or not


reference: https://react.dev/reference/react/useRef

src/features/content-answers/ContentAnswersModalContent.tsx
```diff
@@ -13,28 +13,60 @@ import { User } from '../../../common/types/core';
...
-                    questions.map(({ prompt, error, answer = '' }, index) => {
-                        const hasError = !!error;
-                        return (
-                            <li key={index}>
-                                <Question currentUser={currentUser} prompt={prompt} />
-                                {hasError ? <InlineError /> : <Answer answer={answer} isLoading={isLoading} />}
-                            </li>
-                        );
-                    })}
+                {questions?.map(({ prompt, error, answer = '' }, index) => {
```
tjuanitas, line 51: nit: sorry didn't see this before but `questions` is a required prop and it's defaulted to an empty array so the `?.` check isn't necessary

src/features/content-answers/ContentAnswersModalContent.tsx
```diff
@@ -13,28 +13,60 @@ import { User } from '../../../common/types/core';
...
 const ContentAnswersModalContent = ({ currentUser, fileName, isLoading, questions }: Props) => {
+    const messagesEndRef = useRef<HTMLDivElement | null>(null);
+
+    const handleScrollToBottom = useCallback(behavior => {
+        if (messagesEndRef.current && messagesEndRef.current.scrollIntoView) {
+            messagesEndRef.current.scrollIntoView({ behavior });
+        }
+    }, []);
+
+    const throttledScrollToBottom = throttle(handleScrollToBottom, 1000);
```
tjuanitas, line 31: is this 1 second? sounds sort of long, is that to fix the scrolling issue?

tjuanitas (dismissed):
lgtm

src/features/content-answers/ContentAnswersModalContent.tsx
TaoJiang98: actually just got an update that this throttling is for streaming mode, so we could remove this logic

tjuanitas (approved):
lgtm but would confirm that @megansmith-box concerns are addressed

src/features/content-answers/__tests__/ContentAnswersModalContent.test.tsx
```diff
@@ -61,6 +66,35 @@ describe('features/content-answers/ContentAnswersModalContent', () => {
...
+    test('scrollIntoView function is called with smooth behavior when the answer is updated', () => {
+        const { rerender } = renderComponent({ isLoading: true });
+
+        rerender(
+            <ContentAnswersModalContent
+                currentUser={mockCurrentUser}
+                fileName={mockFile.name}
+                isLoading={false}
+                questions={mockQuestionsWithAnswer}
+            />,
```
greg-in-a-box, line 91: do we need this `,` here?

ghost (approved):

greg-in-a-box (approved):

PR #3444, opened 2023-10-13
feat(content-answers): add inline error

TaoJiang98:
Screenshot 2023-10-13 at 2 33 59 PM

### DIFF
--- a/src/features/content-answers/Answer.tsx
+++ b/src/features/content-answers/Answer.tsx
@@ -1,20 +1,26 @@
 import * as React from 'react';
 
 import ContentAnswersGridCard from './ContentAnswersGridCard';
+import InlineError from './InlineError';
 import LoadingElement from './LoadingElement';
+// @ts-ignore: no ts definition
+// eslint-disable-next-line import/named
+import { ElementsXhrError } from '../../../common/types/api';
 
 import './Answer.scss';
 
 type Props = {
     answer?: string;
+    error: ElementsXhrError;
     isLoading: boolean;
 };
 
-const Answer = ({ answer, isLoading }: Props) => {
+const Answer = ({ answer, error, isLoading }: Props) => {
     return (
         <div className="bdl-Answer">
             {answer && <ContentAnswersGridCard>{answer}</ContentAnswersGridCard>}
             {!answer && isLoading && <LoadingElement />}
+            {error && <InlineError error={error} />}
         </div>
     );
 };
--- a/src/features/content-answers/Answer.tsx
+++ b/src/features/content-answers/Answer.tsx
@@ -19,8 +19,8 @@ const Answer = ({ answer, error, isLoading }: Props) => {
     return (
         <div className="bdl-Answer">
             {answer && <ContentAnswersGridCard>{answer}</ContentAnswersGridCard>}
-            {!answer && isLoading && <LoadingElement />}
-            {error && <InlineError error={error} />}
+            {!answer && !error && isLoading && <LoadingElement />}
+            {error && <InlineError />}
         </div>
     );
 };
--- a/src/features/content-answers/Answer.tsx
+++ b/src/features/content-answers/Answer.tsx
@@ -1,26 +1,20 @@
 import * as React from 'react';
 
 import ContentAnswersGridCard from './ContentAnswersGridCard';
-import InlineError from './InlineError';
 import LoadingElement from './LoadingElement';
-// @ts-ignore: no ts definition
-// eslint-disable-next-line import/named
-import { ElementsXhrError } from '../../../common/types/api';
 
 import './Answer.scss';
 
 type Props = {
     answer?: string;
-    error?: ElementsXhrError;
     isLoading: boolean;
 };
 
-const Answer = ({ answer, error, isLoading }: Props) => {
+const Answer = ({ answer, isLoading }: Props) => {
     return (
-        <div className="bdl-Answer">
+        <div className="bdl-Answer" data-testid="Answer">
             {answer && <ContentAnswersGridCard>{answer}</ContentAnswersGridCard>}
-            {!answer && !error && isLoading && <LoadingElement />}
-            {error && <InlineError />}
+            {!answer && isLoading && <LoadingElement />}
         </div>
     );
 };
--- a/src/features/content-answers/Answer.tsx
+++ b/src/features/content-answers/Answer.tsx
@@ -11,7 +11,7 @@ import './Answer.scss';
 
 type Props = {
     answer?: string;
-    error: ElementsXhrError;
+    error?: ElementsXhrError;
     isLoading: boolean;
 };
 
--- a/src/features/content-answers/ContentAnswersModal.tsx
+++ b/src/features/content-answers/ContentAnswersModal.tsx
@@ -42,27 +42,24 @@ const ContentAnswersModal = ({ api, currentUser, file, isOpen, onRequestClose }:
     const [isLoading, setIsLoading] = useState<boolean>(false);
     const [questions, setQuestions] = useState<QuestionType[]>([]);
 
-    const handleSuccessCallback = useCallback((prevQuestions, prompt, response): void => {
-        const updatedQuestion = {
+    const handleSuccessCallback = useCallback((response): void => {
+        const question = {
             answer: response.data.answer,
             createdAt: response.data.created_at,
-            prompt,
         };
 
-        setQuestions([...prevQuestions, updatedQuestion]);
+        setQuestions(prevState => {
+            const lastQuestion = prevState[prevState.length - 1];
+            return [...prevState.slice(0, -1), { ...lastQuestion, ...question }];
+        });
     }, []);
 
-    const handleErrorCallback = useCallback(
-        (error: ElementsXhrError, prevQuestions: QuestionType[], prompt: string): void => {
-            const updatedQuestion = {
-                error,
-                prompt,
-            };
-
-            setQuestions([...prevQuestions, updatedQuestion]);
-        },
-        [],
-    );
+    const handleErrorCallback = useCallback((error: ElementsXhrError): void => {
+        setQuestions(prevState => {
+            const lastQuestion = prevState[prevState.length - 1];
+            return [...prevState.slice(0, -1), { ...lastQuestion, error }];
+        });
+    }, []);
 
     const handleOnAsk = useCallback(
         async (prompt: string) => {
@@ -73,14 +70,13 @@ const ContentAnswersModal = ({ api, currentUser, file, isOpen, onRequestClose }:
                     type: 'file',
                 },
             ];
-            const prevQuestions = [...questions];
-            setQuestions([...prevQuestions, { prompt }]);
+            setQuestions([...questions, { prompt }]);
             setIsLoading(true);
             try {
                 const response = await api.getIntelligenceAPI(true).ask(prompt, items);
-                handleSuccessCallback(prevQuestions, prompt, response);
+                handleSuccessCallback(response);
             } catch (e) {
-                handleErrorCallback(e, prevQuestions, prompt);
+                handleErrorCallback(e);
             }
             setIsLoading(false);
         },
--- a/src/features/content-answers/ContentAnswersModal.tsx
+++ b/src/features/content-answers/ContentAnswersModal.tsx
@@ -42,22 +42,27 @@ const ContentAnswersModal = ({ api, currentUser, file, isOpen, onRequestClose }:
     const [isLoading, setIsLoading] = useState<boolean>(false);
     const [questions, setQuestions] = useState<QuestionType[]>([]);
 
-    const handleSuccessCallback = useCallback((response, updatedQuestions): void => {
-        const lastQuestion = updatedQuestions[updatedQuestions.length - 1];
+    const handleSuccessCallback = useCallback((prevQuestions, prompt, response): void => {
+        const updatedQuestion = {
+            answer: response.data.answer,
+            createdAt: response.data.created_at,
+            prompt,
+        };
 
-        if (lastQuestion && lastQuestion.prompt) {
-            lastQuestion.answer = response.data.answer;
-            lastQuestion.createdAt = response.data.created_at;
-        }
+        setQuestions([...prevQuestions, updatedQuestion]);
     }, []);
 
-    const handleErrorCallback = useCallback((error: ElementsXhrError, updatedQuestions: QuestionType[]): void => {
-        const lastQuestion = updatedQuestions[updatedQuestions.length - 1];
+    const handleErrorCallback = useCallback(
+        (error: ElementsXhrError, prevQuestions: QuestionType[], prompt: string): void => {
+            const updatedQuestion = {
+                error,
+                prompt,
+            };
 
-        if (lastQuestion && lastQuestion.prompt) {
-            lastQuestion.error = error;
-        }
-    }, []);
+            setQuestions([...prevQuestions, updatedQuestion]);
+        },
+        [],
+    );
 
     const handleOnAsk = useCallback(
         async (prompt: string) => {
@@ -68,14 +73,14 @@ const ContentAnswersModal = ({ api, currentUser, file, isOpen, onRequestClose }:
                     type: 'file',
                 },
             ];
-            const updatedQuestions = [...questions, { prompt }];
-            setQuestions(updatedQuestions);
+            const prevQuestions = [...questions];
+            setQuestions([...prevQuestions, { prompt }]);
             setIsLoading(true);
             try {
                 const response = await api.getIntelligenceAPI(true).ask(prompt, items);
-                handleSuccessCallback(response, updatedQuestions);
+                handleSuccessCallback(prevQuestions, prompt, response);
             } catch (e) {
-                handleErrorCallback(e, updatedQuestions);
+                handleErrorCallback(e, prevQuestions, prompt);
             }
             setIsLoading(false);
         },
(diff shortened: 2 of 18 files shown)


src/features/content-answers/Answer.tsx
```diff
@@ -1,20 +1,26 @@
...
+    error?: ElementsXhrError;
     isLoading: boolean;
 };
 
-const Answer = ({ answer, isLoading }: Props) => {
+const Answer = ({ answer, error, isLoading }: Props) => {
     return (
         <div className="bdl-Answer">
             {answer && <ContentAnswersGridCard>{answer}</ContentAnswersGridCard>}
             {!answer && isLoading && <LoadingElement />}
```
greg-in-a-box, line 22: We need to prevent these were rendering if there is an error, definitely the first line but the second line is debatable depending on how isLoading is handle but I think its safe to say both should not render if there is an error

src/features/content-answers/ContentAnswersModal.tsx
```diff
@@ -26,6 +26,7 @@ export type QuestionType = {
     prompt: string;
     answer?: string;
     createdAt?: string;
+    error?: string;
```
greg-in-a-box, line 29: I didnt notice this before but abc

src/features/content-answers/ContentAnswersModal.tsx
```diff
@@ -53,11 +54,15 @@ const ContentAnswersModal = ({ api, currentUser, file, isOpen, onRequestClose }:
...
         setIsLoading(false);
-        if (error) {
-            // TODO: render error component
+
+        const lastQuestion = updatedQuestions.pop();
+        if (lastQuestion && lastQuestion.prompt) {
+            lastQuestion.error = error;
         }
+
+        setQuestions([...updatedQuestions, lastQuestion]);
```
greg-in-a-box, line 65: should we still update state if there isnt a `lastQuestion`?

src/features/content-answers/ContentAnswersModal.tsx
```diff
@@ -53,11 +54,15 @@ const ContentAnswersModal = ({ api, currentUser, file, isOpen, onRequestClose }:
         setQuestions([...updatedQuestions, lastQuestion]);
     }, []);
 
-    const handleErrorCallback = useCallback((error: ElementsXhrError): void => {
+    const handleErrorCallback = useCallback((error: ElementsXhrError, updatedQuestions): void => {
```
greg-in-a-box, line 57: since we're already adding the error type, can we add the updatedQuestions type?

src/features/content-answers/__tests__/InlineError.test.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { ElementsXhrError } from '../../../common/types/api';
+
+import messages from '../messages';
+
+jest.mock('react-intl', () => jest.requireActual('react-intl'));
+
+describe('features/content-answers/InlineError', () => {
+    const IntlWrapper = ({ children }: { children?: React.ReactNode }) => {
+        return <IntlProvider locale="en">{children}</IntlProvider>;
+    };
```
greg-in-a-box, line 17: we dont need the intl wrapper here since youre not apply it directly to this component 
```suggestion

```

src/features/content-answers/InlineError.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+import messages from './messages';
+
+import './InlineError.scss';
+
+type Props = {
+    error: ElementsXhrError;
+};
+
+const InlineError = ({ error }: Props) => {
```
greg-in-a-box, line 19: do we have to pass in the `error`? you're already controlling when to render component in the parent component

src/features/content-answers/__tests__/Answer.test.tsx
```diff
@@ -2,21 +2,32 @@ import React from 'react';
...
 
     test('should render the loading element', () => {
-        renderComponent({ answer: '', isLoading: true });
+        renderComponent({ answer: '', error: undefined, isLoading: true });
 
         const loadingElement = screen.getByTestId('LoadingElement');
         expect(loadingElement).toBeInTheDocument();
     });
+
+    test('should render inline error', () => {
```
greg-in-a-box, line 27: I think we should check if the other two have rendered or not

src/features/content-answers/__tests__/Answer.test.tsx
```diff
@@ -2,21 +2,32 @@ import React from 'react';
...
+        renderComponent({ answer: '', error: new Error('error'), isLoading: true });
```
greg-in-a-box, line 28: I think you need to do something like a test.each to check all the possibilities and then verify the expected behavior in each of them

greg-in-a-box (commented):
left some comments

src/features/content-answers/ContentAnswersModal.tsx
TaoJiang98: It should be guaranteed that there is always a lastQuestion in the question array. The reason is that whenever we submit a question, the new question object with the prompt only will be appended into the question array.

src/features/content-answers/InlineError.tsx
TaoJiang98: removed error prop, but in the future we could possibly pass error and make the inlineError component show different error messages based on the error prop

src/features/content-answers/__tests__/InlineError.test.tsx
TaoJiang98: this is to handle the console.err `[React Intl] Could not find required `intl` object. <IntlProvider> needs to exist in the component ancestry. Using default message as fallback.`

TaoJiang98: there is a global mock for FormattedMessage shown below https://github.com/box/box-ui-elements/blob/4f6c21941c2f0d05a830ffd2c3dfdca4b946895c/src/__mocks__/react-intl.js#L16-L17

In order to test the text message, we need to mock it again and use the actual react-intl which requires IntlWrapper.

src/features/content-answers/Answer.tsx
```diff
@@ -1,20 +1,26 @@
...
     answer?: string;
+    error?: ElementsXhrError;
     isLoading: boolean;
 };
 
-const Answer = ({ answer, isLoading }: Props) => {
+const Answer = ({ answer, error, isLoading }: Props) => {
     return (
         <div className="bdl-Answer">
             {answer && <ContentAnswersGridCard>{answer}</ContentAnswersGridCard>}
```
greg-in-a-box, line 21: I dont think we should give the opportunity to provide an answer and an error at the same time where it renders both

greg-in-a-box (changes requested):

src/features/content-answers/Answer.tsx
TaoJiang98: do you suggest that only pass answer prop here and if the error is not undefined, we display InlineError directly in `ContentAnswersModalContent` component?

src/features/content-answers/__mocks__/mocks.ts
```diff
@@ -26,6 +26,14 @@ export const mockQuestionsWithAnswer: QuestionType[] = [
     },
 ];
 
+export const mockQuestionsWithError: QuestionType[] = [
+    {
+        prompt: 'summarize',
+        answer: '',
+        error: new Error('error'),
```
greg-in-a-box, line 33: Is this what the actual return is for a xhr error ?

src/features/content-answers/__tests__/Answer.test.tsx
```diff
@@ -6,17 +6,19 @@ import Answer from '../Answer';
...
+        expect(screen.queryByTestId('LoadingElement')).not.toBeInTheDocument();
     });
 
-    test('should render the loading element', () => {
+    test('should only render the Loading Element', () => {
         renderComponent({ answer: '', isLoading: true });
 
         const loadingElement = screen.getByTestId('LoadingElement');
         expect(loadingElement).toBeInTheDocument();
+        expect(screen.queryAllByTestId('content-answers-grid-card')).toHaveLength(1);
```
greg-in-a-box, line 22: I think this should assert if the answer is there not the card, its easier to understand in terms of readability. noone is going to know that the card is actually under the LoadingElement too unless they check in that component too

src/features/content-answers/ContentAnswersModal.tsx
```diff
@@ -48,15 +49,17 @@ const ContentAnswersModal = ({ api, currentUser, file, isOpen, onRequestClose }:
...
-    const handleErrorCallback = useCallback((error: ElementsXhrError): void => {
+    const handleErrorCallback = useCallback((error: ElementsXhrError, updatedQuestions: QuestionType[]): void => {
         setIsLoading(false);
-        if (error) {
-            // TODO: render error component
+
+        const lastQuestion = updatedQuestions.pop();
+        if (lastQuestion && lastQuestion.prompt) {
+            lastQuestion.error = error;
+            setQuestions([...updatedQuestions, lastQuestion]);
```
greg-in-a-box, line 62: I think you need a test to handle this logic

greg-in-a-box (changes requested):

src/features/content-answers/__tests__/Answer.test.tsx
TaoJiang98: since the answer prop here is given as ''. I think it does not make sense to check if text '' not in the document. Instead, I checked length of children list under Answer component is 1 which should be loading element.

src/features/content-answers/messages.ts
```diff
@@ -32,6 +32,11 @@ const messages = defineMessages({
             'Disabled tooltip message for Content Answers entry point button when the file type is not supported',
         id: 'boxui.contentAnswers.disabledTooltipFileNotCompatible',
     },
+    inlineErrorText: {
+        defaultMessage: 'The Box AI service was unavailable.',
```
megansmith-box, line 36: Should it say, `is` instead of `was`?

TaoJiang98: When the error comes out, there would be a retry button for user to click. And I think it would make sense saying that Box AI service was unavailable and it is worth to retry.

src/features/content-answers/__tests__/ContentAnswersModalContent.test.tsx
```diff
@@ -59,4 +60,11 @@ describe('features/content-answers/ContentAnswersModalContent', () => {
         const loadingElement = screen.queryAllByTestId('LoadingElement');
         expect(loadingElement.length).toEqual(1);
     });
+
+    test('should render only the inline error which not loading and there is an error', () => {
```
tjuanitas, line 64: `when` instead of `which`?

src/features/content-answers/__tests__/InlineError.test.tsx
```diff
@@ -0,0 +1,22 @@
...
+import InlineError from '../InlineError';
+
+import messages from '../messages';
+
+jest.mock('react-intl', () => jest.requireActual('react-intl'));
+
+describe('features/content-answers/InlineError', () => {
+    const IntlWrapper = ({ children }: { children?: React.ReactNode }) => {
+        return <IntlProvider locale="en">{children}</IntlProvider>;
+    };
```
tjuanitas, line 14: does this work:
```
jest.mock('react-intl', () => ({
    ...jest.requireActual('react-intl'),
    FormattedMessage: ({ defaultMessage }: { defaultMessage: string }) => <span>{defaultMessage}</span>,
}));
```

or does it still need the whole wrapper? when does the console error occur? why does a test like `ContentAnalyticsErrorState.test.tsx` not need it?

src/features/content-answers/ContentAnswersModal.tsx
```diff
@@ -48,15 +49,17 @@ const ContentAnswersModal = ({ api, currentUser, file, isOpen, onRequestClose }:
...
-        setQuestions([...updatedQuestions, lastQuestion]);
     }, []);
 
-    const handleErrorCallback = useCallback((error: ElementsXhrError): void => {
+    const handleErrorCallback = useCallback((error: ElementsXhrError, updatedQuestions: QuestionType[]): void => {
         setIsLoading(false);
-        if (error) {
-            // TODO: render error component
+
+        const lastQuestion = updatedQuestions.pop();
```
tjuanitas, line 59: we should probably avoid object mutation here since the reference to `updatedQuestions` is used in multiple places which could cause unexpected issues

maybe a `slice()` or length solution:
```
const [lastQuestion] = updatedQuestions.slice(-1);

// or

const lastQuestion = updatedQuestions[updatedQuestions.length - 1];
```

src/features/content-answers/ContentAnswersModal.tsx
```diff
@@ -48,15 +49,17 @@ const ContentAnswersModal = ({ api, currentUser, file, isOpen, onRequestClose }:
...
             lastQuestion.createdAt = response.data.created_at;
+            setQuestions([...updatedQuestions, lastQuestion]);
         }
-
-        setQuestions([...updatedQuestions, lastQuestion]);
     }, []);
 
-    const handleErrorCallback = useCallback((error: ElementsXhrError): void => {
+    const handleErrorCallback = useCallback((error: ElementsXhrError, updatedQuestions: QuestionType[]): void => {
         setIsLoading(false);
```
tjuanitas, line 57: why is `setIsLoading` called at the beginning of the function instead of the end of the logic or at the end of `handleOnAsk`?

src/features/content-answers/ContentAnswersModalContent.tsx
```diff
@@ -24,11 +25,13 @@ const ContentAnswersModalContent = ({ currentUser, fileName, isLoading, question
...
                 {questions &&
-                    questions.map(({ prompt, answer = '' }, index) => {
+                    questions.map(({ prompt, error, answer = '' }, index) => {
+                        const hasError = !!error;
                         return (
                             <li key={index}>
                                 <Question currentUser={currentUser} prompt={prompt} />
-                                <Answer answer={answer} isLoading={isLoading} />
+                                {!hasError && <Answer answer={answer} isLoading={isLoading} />}
+                                {hasError && !isLoading && <InlineError />}
```
tjuanitas, line 34: do we want to remove the `isLoading` check for `InlineError`? wouldn't this affect all "error answers"?

**Stage 1**
> Q: summarize this document

**Stage 2**
> Q: summarize this document
A: ERROR

**Stage 3**
> Q: summarize this document
A: ERROR
Q: summarize this document in 200 words

**Stage 4**
> Q: summarize this document
A: LOADING
Q: summarize this document in 200 words
A: LOADING

**Stage 5**
> Q: summarize this document
A: ERROR
Q: summarize this document in 200 words
A: blah blah blah

tjuanitas:
```
{hasError ? <InlineError /> : <Answer answer={answer} isLoading={isLoading} />}
```

src/features/content-answers/__tests__/InlineError.test.tsx
TaoJiang98: yeah I think this works. The console error occurs when I only mocks 'react-intl'. The 
test ContentAnalyticsErrorState.test.tsx did not need it because it was testing the element instead of the text.

src/features/content-answers/ContentAnswersModal.tsx
```diff
@@ -76,8 +75,9 @@ const ContentAnswersModal = ({ api, currentUser, file, isOpen, onRequestClose }:
                 const response = await api.getIntelligenceAPI(true).ask(prompt, items);
                 handleSuccessCallback(response, updatedQuestions);
             } catch (e) {
-                handleErrorCallback(e);
+                handleErrorCallback(e, updatedQuestions);
```
tjuanitas, line 78: sorry it might be out of scope but I didn't realize this before: are these "callbacks" necessary?

why not something like:
```
const handleOnAsk = useCallback(
    ...
    let nextQuestion = { prompt };
    setQuestions([...questions, nextQuestion]);
    setIsLoading(true);
    try {
        const response = await api.getIntelligenceAPI(true).ask(prompt, items);
        nextQuestion = {
            answer: response.data.answer,
            createdAt: response.data.created_at,
            prompt,
        };
    } catch (error) {
        nextQuestion = {
            error,
            prompt,
        };
    }
    setQuestions([...questions, nextQuestion]);
    setIsLoading(false);
```
that way you don't need `handleSuccessCallback` and `handleErrorCallback` and avoids the object mutation

TaoJiang98: This is a very great point! I think previously it is written as `handleSuccessCallback` and `handleErrorCallback` to better address which part need to work on for following small tasks. 

When bring everything back to try catch, it still needs to update the last question either in place or pop up and then append it back since we need to append the question with prompt into the array at the very beginning.

TaoJiang98: removed the object mutation logic

src/features/content-answers/__tests__/ContentAnswersModal.test.tsx
```diff
@@ -19,17 +26,18 @@ describe('features/content-answers/ContentAnswersModal', () => {
                 />
             </APIContext.Provider>,
         );
+    };
 
     test('should render the header icon', () => {
-        renderComponent();
+        renderComponent(mockApi);
```
greg-in-a-box, line 32: cant mockApi be defaulted since you have to use it more than once?

greg-in-a-box (changes requested):

tjuanitas (approved):

greg-in-a-box (approved):

ghost (approved):

PR #3445, opened 2023-10-17
feat(activitysidebar): collapse replies on resolve

ahorowitz123:
- Collapse replies on comment/annotation resolve to show only one comment

comment resolve collapse

annotation resolve collapse

### DIFF
--- a/src/elements/content-sidebar/ActivitySidebar.js
+++ b/src/elements/content-sidebar/ActivitySidebar.js
@@ -1143,8 +1143,9 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
     hideRepliesForFeedItem = (feedItemId: string) => {
         const { feedItems } = this.state;
         const feedItem = feedItems.find((item: FeedItem) => item.id === feedItemId);
-        const lastReply = feedItem.replies.slice(-1);
-        if (lastReply.length > 1) {
+        const { replies } = feedItem;
+        if (replies.length > 1) {
+            const lastReply = feedItem.replies.slice(-1);
             this.updateReplies(feedItemId, lastReply);
         }
     };
--- a/src/elements/content-sidebar/ActivitySidebar.js
+++ b/src/elements/content-sidebar/ActivitySidebar.js
@@ -1144,7 +1144,7 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         const { feedItems } = this.state;
         const feedItem = feedItems.find((item: FeedItem) => item.id === feedItemId);
         const { replies } = feedItem;
-        if (replies.length > 1) {
+        if (replies && replies.length > 1) {
             const lastReply = feedItem.replies.slice(-1);
             this.updateReplies(feedItemId, lastReply);
         }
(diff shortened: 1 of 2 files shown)


src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -1131,6 +1140,16 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         });
     };
 
+    hideRepliesForFeedItem = (feedItemId: string) => {
+        const { feedItems } = this.state;
+        const feedItem = feedItems.find((item: FeedItem) => item.id === feedItemId);
```
TylerGauntlett, line 1145: Any risk that `feedItemId` could not be contained in `feedItems`? If so, that'd result in `feedItem` would result in `undefined`. Destructoring `replies` would throw an error in this scenario

Consider the following:

```js
const { replies } = feedItems || {}
```

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -1131,6 +1140,16 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
...
+        const { replies } = feedItem;
+        if (replies.length > 1) {
+            const lastReply = feedItem.replies.slice(-1);
```
TylerGauntlett, line 1148: `replies` has already been destructured. Consider using it directly instead of accessing via sub property

`const lastReply = replies.slice(-1);`

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -237,6 +238,10 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
             (annotation: Annotation) => {
                 emitAnnotationUpdateEvent(annotation);
                 this.feedSuccessCallback();
+
```
bfoxx1906, line 241: Instead of putting this logic in ActivitySidebar can we not toggle the replies in BaseComment in accordance with the resolution status of that Comment?

src/elements/content-sidebar/ActivitySidebar.js
```diff
@@ -1131,6 +1140,18 @@ class ActivitySidebar extends React.PureComponent<Props, State> {
         });
     };
 
+    hideRepliesForFeedItem = (feedItemId: string) => {
+        const { feedItems = [] } = this.state;
+        // unfortunately, casting to any is the only way to avoid flow issues resulting
+        // from replies not existing on certain FeedItem types
+        const feedItem: any = feedItems.find(item => item.id === feedItemId);
```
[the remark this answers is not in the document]
ahorowitz123: I went down that same rabbit hole...

src/elements/content-sidebar/ActivitySidebar.js
ahorowitz123: Yeah we definitely could add a `useEffect` to BaseComment that would take care of this. Should I do that instead?

bfoxx1906: That would be ideal if possible. We're doing a lot in ActivitySidebar already.

ahorowitz123: I like this solution. I'm going to code this up in a separate branch

ahorowitz123: Going to do this in a new branch since its a completely different solution

ahorowitz123:
Creating a new branch/PR for the BaseComment approach. Leaving this PR and branch around until the other one is approved an merged

PR #3447, opened 2023-10-18
Hide reply

TylerGauntlett:
### What this does
* Hides the "Reply" button when the comment is in a resolved state
* Visually maintains margins/padding when "Reply" button is displayed

### Callouts
* Margins had to be changed around a bit to properly apply bottom margin when the "Reply" button is shown vs hidden


mickr (approved):
lgtm

PR #3448, opened 2023-10-18
feat(content-answers): button focusing

greg-in-a-box:
focus button when qa modal closes after a question has been submitted

### DIFF
--- a/src/features/content-answers/ContentAnswers.tsx
+++ b/src/features/content-answers/ContentAnswers.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useCallback, useState } from 'react';
 
 import getProp from 'lodash/get';
 import ContentAnswersModal from './ContentAnswersModal';
@@ -17,14 +17,19 @@ type Props = {
 
 const ContentAnswers = ({ file }: Props) => {
     const [isModalOpen, setIsModalOpen] = useState(false);
+    const [isHighlighted, setIsHighlighted] = useState(false);
 
-    const handleClick = () => {
+    const handleClick = useCallback(() => {
         setIsModalOpen(true);
-    };
+    }, [setIsModalOpen]);
 
-    const handleClose = () => {
+    const handleClose = useCallback(() => {
         setIsModalOpen(false);
-    };
+    }, [setIsModalOpen]);
+
+    const handleAsk = useCallback(() => {
+        setIsHighlighted(true);
+    }, [setIsHighlighted]);
 
     const currentExtension = getProp(file, 'extension');
     return (
@@ -32,12 +37,15 @@ const ContentAnswers = ({ file }: Props) => {
             <ContentAnswersOpenButton
                 data-testid="content-answers-open-button"
                 fileExtension={currentExtension}
+                isHighlighted={isHighlighted}
+                isModalOpen={isModalOpen}
                 onClick={handleClick}
             />
             <ContentAnswersModal
                 data-testid="content-answers-modal"
                 file={file}
                 isOpen={isModalOpen}
+                onAsk={handleAsk}
                 onRequestClose={handleClose}
             />
         </div>
--- a/src/features/content-answers/ContentAnswers.tsx
+++ b/src/features/content-answers/ContentAnswers.tsx
@@ -17,6 +17,7 @@ type Props = {
 
 const ContentAnswers = ({ file }: Props) => {
     const [isModalOpen, setIsModalOpen] = useState(false);
+    const [hasQuestions, setHasQuestions] = useState(false);
     const [isHighlighted, setIsHighlighted] = useState(false);
 
     const handleClick = useCallback(() => {
@@ -25,11 +26,14 @@ const ContentAnswers = ({ file }: Props) => {
 
     const handleClose = useCallback(() => {
         setIsModalOpen(false);
-    }, [setIsModalOpen]);
+        if (hasQuestions) {
+            setIsHighlighted(true);
+        }
+    }, [hasQuestions]);
 
     const handleAsk = useCallback(() => {
-        setIsHighlighted(true);
-    }, [setIsHighlighted]);
+        setHasQuestions(true);
+    }, []);
 
     const currentExtension = getProp(file, 'extension');
     return (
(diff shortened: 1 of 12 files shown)


src/features/content-answers/ContentAnswersOpenButton.tsx
```diff
@@ -13,35 +14,58 @@ import './ContentAnswersOpenButton.scss';
...
+                className={openButtonClassNames}
                 data-testid="content-answers-open-button"
                 isDisabled={!isAllowedFileType(fileExtension)}
                 onClick={onClick}
+                setRef={(ref: HTMLButtonElement) => {
+                    buttonRef.current = ref;
+                }}
             >
-                <BoxAiLogo width={20} height={20} />
+                <BoxAiLogo className="bdl-BoxAIIcon" width={20} height={20} />
```
tjuanitas, line 68: the icon is a descendant of the root component `ContentAnswersOpenButton` so it would be something like: `bdl-ContentAnswersOpenButton-icon`

src/features/content-answers/ContentAnswers.tsx
```diff
@@ -17,27 +17,35 @@ type Props = {
...
-    };
+    }, [setIsModalOpen]);
 
-    const handleClose = () => {
+    const handleClose = useCallback(() => {
         setIsModalOpen(false);
-    };
+    }, [setIsModalOpen]);
+
+    const handleAsk = useCallback(() => {
```
tjuanitas, line 30: do these need the `useCallback`? is there a benefit to them?

src/features/content-answers/ContentAnswersModal.tsx
```diff
@@ -62,6 +63,7 @@ const ContentAnswersModal = ({ api, currentUser, file, isOpen, onRequestClose }:
 
     const handleOnAsk = useCallback(
         async (prompt: string) => {
+            onAsk();
```
tjuanitas, line 66: this also highlights if the first question returns an error, is that okay? or do we want to move until after a successful response?

src/features/content-answers/ContentAnswersOpenButton.scss
```diff
@@ -17,4 +31,28 @@
...
+
+    &.bdl-ContentAnswersOpenButton--hasQuestions:not(.bdl-is-disabled):focus,
+    &.bdl-ContentAnswersOpenButton--hasQuestions:not(.is-disabled):focus {
+        @include BackgroundGradientHover;
+        @include ButtonOutline;
+    }
+
+    &-tooltip {
+        text-align: center;
+    }
```
tjuanitas, line 57: this wouldn't work because `bdl-ContentAnswersOpenButton-tooltip` is currently a wrapper around the button class `bdl-ContentAnswersOpenButton` so it's not a descendent element. technically `bdl-ContentAnswersOpenButton` should be moved to the root element in SUIT but that probably introduces more headache then it resolves

but two thoughts:
1. do we actually need `text-align: center`? I know it's used in the EUA version but it doesn't follow the pattern we use for the rest of BUIE / Preview to center text in a tooltip
2. `&-tooltip` hurts searchability when searching for `bdl-ContentAnswersOpenButton-tooltip` in GitHub or IDE

src/features/content-answers/ContentAnswersOpenButton.scss
```diff
@@ -17,4 +31,28 @@
...
+            }
+        }
+    }
+
+    &.bdl-ContentAnswersOpenButton--hasQuestions:hover {
+        @include BackgroundGradientHover;
+    }
+
+    &.bdl-ContentAnswersOpenButton--hasQuestions:not(.bdl-is-disabled):focus,
+    &.bdl-ContentAnswersOpenButton--hasQuestions:not(.is-disabled):focus {
```
tjuanitas, line 50: the eventual goal is to replace `is-disabled` with `bdl-is-disabled` so it shouldn't need both. I think you can remove the second line

src/features/content-answers/ContentAnswersOpenButton.scss
```diff
@@ -1,5 +1,19 @@
...
+}
+
+@mixin BackgroundGradientHover {
+    background: linear-gradient(0deg, #2223, #2223), linear-gradient(135deg, #fe01da 0%, $bdl-light-blue 100%);
+    border: 1px solid #2222224d;
+}
+
+@mixin ButtonOutline {
+    box-shadow: 0 0 0 1px $white, 0 0 0 3px $bdl-light-blue;
+}
```
tjuanitas, line 15: nit: mixins can also follow SUIT naming:
`bdl-ContentAnswersOpenButton-backgroundGradient`
`bdl-ContentAnswersOpenButton-onHover-backgroundGradient`

tjuanitas, line 15: is this needed as a mixin or just apply this one directly?

src/features/content-answers/ContentAnswersModal.tsx
greg-in-a-box: the other one does that also, so it meets parity

src/features/content-answers/ContentAnswers.tsx
greg-in-a-box: the performance difference is probably incredible low from our point of view but I dont see it hurting us in the short term.

tjuanitas (dismissed):
some nits but good to me

src/features/content-answers/ContentAnswersModal.tsx
```diff
@@ -34,10 +34,11 @@ type Props = {
     currentUser?: User;
     file: BoxItem;
     isOpen: boolean;
+    onAsk: Function;
     onRequestClose: () => void;
```
tjuanitas, line 38: nit: do we want to be more specific than `Function` similar to `onRequestClose`?

tjuanitas (dismissed):
looks great!

ghost (dismissed):

TaoJiang98 (dismissed):

TaoJiang98 (dismissed):

src/features/content-answers/ContentAnswersModal.tsx
greg-in-a-box: updated to change all of the callbacks in here

tjuanitas (approved):

TaoJiang98 (approved):

ghost (approved):

PR #3449, opened 2023-10-19
feat(content-explorer-modal): allow rendering within parent

dmazaleuskaya:
Extended `ContentExplorerModalContainer` and `ContentExplorerModal` components to accept optional `shouldNotUsePortal` to that it's possible to render the modal within the parent, not only in the Portal.


tnastula (approved):

greg-in-a-box (changes requested):
In `ContentExplorerModalContainer`, there are two modals, but the prop you added is applying only to the first one, does the second one also need to do the same thing? 

There should also be a test from this level to show that the `ContentExplorerModal` is actually rendering within the parent and not from a portal, since its a feature that is being added.

dmazaleuskaya:
Close in favour of https://github.com/box/box-ui-elements/pull/3501

PR #3450, opened 2023-10-19
feat(threaded-replies): Collapse replies on resolve v2

ahorowitz123:
- Collapse replies on comment/annotation resolve to show only one comment

comment resolve collapse

annotation resolve collapse

### DIFF
--- a/src/elements/content-sidebar/activity-feed/comment/BaseComment.js
+++ b/src/elements/content-sidebar/activity-feed/comment/BaseComment.js
@@ -101,6 +101,21 @@ export const BaseComment = ({
     const [isEditing, setIsEditing] = React.useState<boolean>(false);
     const [isInputOpen, setIsInputOpen] = React.useState<boolean>(false);
 
+    const prevStatus = React.useRef<FeedItemStatus | undefined>(status);
+
+    React.useEffect(() => {
+        if (
+            onHideReplies &&
+            status === COMMENT_STATUS_RESOLVED &&
+            status !== prevStatus.current &&
+            repliesTotalCount === replies.length
+        ) {
+            onHideReplies(replies.splice(-1));
+        }
+
+        prevStatus.current = status;
+    }, [onHideReplies, replies, repliesTotalCount, status]);
+
     const commentFormFocusHandler = (): void => {
         setIsInputOpen(true);
         onSelect(true);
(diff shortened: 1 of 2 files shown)


src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -101,6 +101,22 @@ export const BaseComment = ({
     const [isEditing, setIsEditing] = React.useState<boolean>(false);
     const [isInputOpen, setIsInputOpen] = React.useState<boolean>(false);
 
+    const prevStatus = React.useRef<?FeedItemStatus>(status);
```
bfoxx1906, line 104: Do we need the ref here? Not sure if the previous status matters or not, we just need to make sure that the comment is in the resolved state if we want to call onHideReplies and replies and repliesTotalCount shouldn't change if the Comment is resolved.

ahorowitz123: The issue is we still want to allow the user to expand or hide the replies manually even if its in the resolved state. If we remove the ref then the that functionality won't work anymore since the comments will always be forced to hide when the comment is resolved

bfoxx1906: Ahh ok that's right. I re-read the AC. Looks good

bfoxx1906 (dismissed):

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
ahorowitz123: I was forced to add those other dependencies to `useEffect` by eslint. Initially I just wanted `status` as the dependency.

src/elements/content-sidebar/activity-feed/comment/BaseComment.js
```diff
@@ -101,6 +101,22 @@ export const BaseComment = ({
...
+
+    React.useEffect(() => {
+        if (
```
JChan106, line 107: If statement is a little long, maybe we can store the statements in 2 boolean variables like `newStatusIsResolved` and `repliesIsExpanded` to improve readability?

JChan106 (approved):
lgtm!

bfoxx1906 (approved):
lgtm

PR #3461, opened 2023-11-08
chore(storybook): Bump storybook to 7

greg-in-a-box:
For the most part, storybook has automigration scripts that run when you run the primary migration command.  
Such as the following: 
1. installing necessary packages 
2. changes some commands 
3. there was more but i already deleted the migration log

Things I had to do manually:
1. bump storybook to latest
4. bump eslint to 7.X (8 required additional effort to get to work that seemed uncessary)
5. bump typescript to latest because of surfaced errors from bumping storybook
6. moved old storybook files to storybook-old folder 
7. removed locale related configuration since it wasnt working before and it wont work until react-intl is bumped
8. moved globals to preview.ts
9. Updated ContentExplorer story

PR #3466, opened 2023-11-28
feat(react-intl): Bump react-intl package [BREAKING CHANGE]

soygitana:
BREAKING CHANGE
IMPORTANT:

- Minimal set of changes to solve the problem. 

Changes:

1. `react-intl` lib has been upgrade to v6.4.2

2. I have to change syntax of some messages and update related snapshots. Single quote which were use to wrap {placeholder} have been replaced by double quote. This is related to Message Format syntax changes in later versions of react-intl. Escape character has been changed to apostrophe ('). ([a9984c3](https://github.com/box/box-ui-elements/pull/3466/commits/a9984c362553c8c9942adbb534220edb46f39d37))

e.g.: now "'{foo}'" will output: "{foo}" in the formatted output instead of interpreting it as a foo argument.

❗ From now, please do not wrap {placeholder} in single quote. If you want to use quotes please use double quote (“{name}”). 

3. `formatRelativeTime` has been fixed (check [comment](https://github.com/box/box-ui-elements/pull/3466#issuecomment-1918775431) ⬇️  for details)

4. **Legacy support has been completely removed.**

 ---
Testing:

Make sure application and strings are not broken when language is changed. 

---
For more information, please check: 

- [official upgrade guide](https://formatjs.io/docs/react-intl/upgrade-guide-3x)
- upgrade process review guide in collaborative spreadsheet


soygitana:
[a12337a](https://github.com/box/box-ui-elements/pull/3466/commits/a12337a9ce973deec23ca1994fcc559a3655687b):

`react-intl` has been rewritten in TS. Now, `react-intl` exposes `IntlShape` as an interface. I removed the `InjectIntlProvidedProps`, which is not provided by the package. I used IntlShape with Flow.

soygitana:
@greg-in-a-box @jstoffan @box/webapp 

It turned out that the `TS` version in BUIE (`"^3.7.5"`) is not sufficient for `react-intl` version 6 (v6.4 requires TS v4, and v6.5 requires TS v5). As a result, there will be type incompatibility after upgrading react-intl (please refer to the lint output). Is there a plan to upgrade TS in BUIE? If not, what would be the best solution?

A. Creating a temporary workaround by generating `d.ts` for `react-intl`
B. Suspending the `react-intl` upgrade until `TS` is upgraded

greg-in-a-box:
> @greg-in-a-box @jstoffan @box/webapp
> 
> It turned out that the `TS` version in BUIE (`"^3.7.5"`) is not sufficient for `react-intl` version 6 (v6.4 requires TS v4, and v6.5 requires TS v5). As a result, there will be type incompatibility after upgrading react-intl (please refer to the lint output). Is there a plan to upgrade TS in BUIE? If not, what would be the best solution?
> 
> A. Creating a temporary workaround by generating `d.ts` for `react-intl` B. Suspending the `react-intl` upgrade until `TS` is upgraded

you can upgrade to `5.2.2` since this PR https://github.com/box/box-ui-elements/pull/3461, we are upgrading it because of storybook being upgrade

soygitana:
[960fc2d](https://github.com/box/box-ui-elements/pull/3466/commits/960fc2d90b7a16aa0c0c5a3f1d93d6d663f692c9):

I have to add resolution for `react-intl` to package.json file ( `"**/react-intl/**/@types/react": "^16.9.18"`). `react-intl` has set `@types/react` to version 16 || 17 || 18 (https://github.com/formatjs/formatjs/blob/main/package.json#L61) so yarn install the latest one, however `@types/react` 17 i 18 are incompatible. It caused a mismatch of types in some components.

soygitana:
> > @greg-in-a-box @jstoffan @box/webapp
> > It turned out that the `TS` version in BUIE (`"^3.7.5"`) is not sufficient for `react-intl` version 6 (v6.4 requires TS v4, and v6.5 requires TS v5). As a result, there will be type incompatibility after upgrading react-intl (please refer to the lint output). Is there a plan to upgrade TS in BUIE? If not, what would be the best solution?
> > A. Creating a temporary workaround by generating `d.ts` for `react-intl` B. Suspending the `react-intl` upgrade until `TS` is upgraded
> 
> you can upgrade to `5.2.2` since this PR #3461, we are upgrading it because of storybook being upgrade

@greg-in-a-box Great! I can see that `#3461` has been already merged so I merged master into this branch and updated `yarn.lock`. TS upgrade and adding resolution for react-intl([960fc2d](https://github.com/box/box-ui-elements/pull/3466/commits/960fc2d90b7a16aa0c0c5a3f1d93d6d663f692c9)) solved TS issues.

src/components/breadcrumb/Breadcrumb.js
```diff
@@ -1,6 +1,7 @@
 // @flow
 import * as React from 'react';
-import { injectIntl, type IntlShape } from 'react-intl';
```
jstoffan: Can we retain the unified import and use it in the other affected modules?

greg-in-a-box: Does flow treat those imports the same way TS does?

package.json
```diff
@@ -372,6 +372,7 @@
         }
     },
     "resolutions": {
-        "ip": "1.1.8"
+        "ip": "1.1.8",
+        "**/react-intl/**/@types/react": "^16.9.18"
```
jstoffan, line 376: Should this be some version of 17 to match our other repositories?

package.json
```diff
@@ -276,7 +276,7 @@
         "react-dom": "^17.0.1",
         "react-draggable": "^3.3.0",
         "react-immutable-proptypes": "^2.1.0",
-        "react-intl": "^3.11.0",
+        "react-intl": "^6.4.2",
```
jstoffan, line 279: Should this use an exact version to match what we have in our other repositories?

scripts/check_generated_files.sh
```diff
@@ -11,7 +11,7 @@ end=$"\e[0m\n"
 
 check_generated_files() {
     printf "${blue}--------------------------------------${end}"
-    printf "${blue}Checking react-intl v3 usage${end}"
+    printf "${blue}Checking react-intl >= 3 usage${end}"
```
jstoffan, line 14: Can we remove these scripts entirely? We no longer support legacy versions, correct?

src/components/i18n/FormattedCompMessage.js
```diff
@@ -1,4 +1,5 @@
 // @flow
+// @deprecated, use FormattedMessage from react-intl v6 instead.
```
jstoffan, line 2: Can we delete these components at this point since the upgrade itself is already a breaking change?

src/components/time/ReadableTime.js
```diff
@@ -79,7 +79,7 @@ const ReadableTime = ({
     const timeDiff = timestamp - Date.now();
     if (Math.abs(timeDiff) <= relativeThreshold) {
         if (intl.formatRelativeTime) {
-            // react-intl v3
+            // react-intl >= 3
             output = intl.formatRelativeTime(timeDiff);
         } else {
```
jstoffan, line 84: Can we remove all support for legacy versions?

jstoffan (commented):
@soygitana, the change itself looks good, but I'm wondering if we can drop all the cruft supporting legacy versions of react-intl. 

Please also update the PR title and commit messages to something other than `chore` and ensure it's marked as a `BREAKING CHANGE` per the [semantic conventions](https://www.conventionalcommits.org).

src/components/breadcrumb/Breadcrumb.js
soygitana: @greg-in-a-box I believe Flow doesn't directly support the 'import type' syntax. However, I continue to use it in the same way as before the upgrade, ('import type { InjectIntlProvidedProps } from 'react-intl'). No errors have been reported, due to settings in .vscode.

soygitana:
> @soygitana, the change itself looks good, but I'm wondering if we can drop all the cruft supporting legacy versions of react-intl.
> 
> Please also update the PR title and commit messages to something other than `chore` and ensure it's marked as a `BREAKING CHANGE` per the [semantic conventions](https://www.conventionalcommits.org).

Regarding supporting legacy versions of react-intl:

1. We cannot remove the FormattedCompMessage or Plural component yet. Even though it is marked as deprecated, it is still used across all Box React apps. We have a linter rule that catches this error and reports it to code owners, but I believe it will take some time to remove it across all repositories.

2. General support for legacy versions (conditional code, etc.): Perhaps it would be better to handle this in a separate PR? If not, I believe we will need to merge and release all React repositories before merging BUIE.

soygitana:
> > @soygitana, the change itself looks good, but I'm wondering if we can drop all the cruft supporting legacy versions of react-intl.
> > Please also update the PR title and commit messages to something other than `chore` and ensure it's marked as a `BREAKING CHANGE` per the [semantic conventions](https://www.conventionalcommits.org).
> 
> @jstoffan Regarding supporting legacy versions of react-intl:
> 
> 1. We cannot remove the FormattedCompMessage or Plural component yet. Even though it is marked as deprecated, it is still used across all Box React apps. We have a linter rule that catches this error and reports it to code owners, but I believe it will take some time to remove it across all repositories.
> 2. General support for legacy versions (conditional code, etc.): Perhaps it would be better to handle this in a separate pull request? If not, will we need to merge and release all React repositories before merging BUIE? or if BUIE is consumed as a package in React repositories, we can merge and release it independently? Subsequently, we can upgrade the BUIE package in all repositories once they are updated.
> 
> ❗ @greg-in-a-box @jstoffan Let's determine whether [this pull request](https://github.com/box/box-ui-elements/pull/3436) is necessary. Dawid has added support for react-intl v6 (to solve issue in Notes), which includes a breaking change related to formatRelative. If we remove support for react-intl v2 in this (#3466) pull request and ensure proper support for v6, I believe [this pull request](https://github.com/box/box-ui-elements/pull/3436) can be closed, correct?

I have agreed with @greg-in-a-box that it would be cleaner and more accurate for me to incorporate those changes from #3436 into this react-intl upgrade pull request. 
For context: The changes from #3436 must be included in this pull request, as they are related to breaking changes in the FormattedRelative API. Failure to do so will result in issues with some apps (e.g., Notes, EUA) caused by BUIE.

src/utils/relativeTime.ts
```diff
@@ -0,0 +1,30 @@
...
+    if (Math.abs(diff) > DAY_IN_MS) {
+        return { value: Math.trunc(diff / DAY_IN_MS), unit: 'day' };
+    }
+    if (Math.abs(diff) > HOUR_IN_MS) {
+        return { value: Math.trunc((diff % DAY_IN_MS) / HOUR_IN_MS), unit: 'hour' };
+    }
+    if (Math.abs(diff) > MIN_IN_MS) {
+        return { value: Math.trunc((diff % HOUR_IN_MS) / MIN_IN_MS), unit: 'minute' };
+    }
+    return { value: Math.trunc((diff % MIN_IN_MS) / SEC_IN_MS), unit: 'second' };
```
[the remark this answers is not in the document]
rlyonbox: What happens when diff is exactly one day? I think it's incorrect:
```javascript
timeFromNow(8.64e7) === { value: 0, unit: 'hour' }
```

soygitana:
[cb22936](https://github.com/box/box-ui-elements/pull/3466/commits/cb22936664c0aa07866e3a07c8aeb898552cc0ee)
[2305977](https://github.com/box/box-ui-elements/pull/3466/commits/2305977cd43437d6ba203ade2468b80125fd4df7)
[11c0afc](https://github.com/box/box-ui-elements/pull/3466/commits/11c0afc8511d5629b9be042484dcbb2a3030a8ab)

In order to address issues related to `formattedRelativeTime` (described [here](https://github.com/box/box-ui-elements/pull/3466#issuecomment-1918775431)), I have added a helper function that can be reused to correctly format relative time. Subsequently, I had to fix components which use `formattedRelativeTime` as it was implemented incorrectly.

Please refer to the screenshot for context:

**PresenceAvatarTooltipContent**

Currently on **MASTER** (assuming `react-intl v6` and `formattedRelativeTime` are used):

Screenshot 2024-01-31 at 15 29 23

After fix in this PR:

Screenshot 2024-01-31 at 15 19 56

jstoffan (dismissed):
LGTM. Once this merges, consumers will be required to upgrade to get new changes to BUIE. Are we ready to upgrade our other first-party repositories in short order?

soygitana:
@jstoffan https://github.com/box/box-ui-elements/pull/3466#discussion_r1474693079:

https://github.com/box/box-ui-elements/pull/3466#issuecomment-1838418400
BUIE uses exact version: "@types/react": "^16.9.18"

soygitana:
> LGTM. Once this merges, consumers will be required to upgrade to get new changes to BUIE. Are we ready to upgrade our other first-party repositories in short order?

@jstoffan Currently, only repositories with pending merges are EUA + Workflow (open pull requests); all other repositories have been merged and are functioning correctly in production.

I had a conversation with @greg-in-a-box, and he mentioned that a stable release of BUIE was just made last week. He also indicated that there won't be a stable release of BUIE anytime soon (please correct me if I'm wrong, @greg-in-a-box).

From our understanding, only annotations and preview require a stable version of BUIE (correct?). Can we delay updating those two packages, or is it necessary to do so before deploying our internal repositories?

jstoffan (approved):
LGTM. Please verify that any remaining internal teams who upgrade BUIE regularly are ready to upgrade react-intl as well.

greg-in-a-box (approved):

PR #3467, opened 2023-11-28
feat(content-sidebar): allow custom tooltip for the sidebar sign button

sstefancu:
- extend the Sign sidebar button to allow the use of an external custom ftux tooltip component in addition to the BUIE default one
- replace a word breaking hyphen character with its non-breaking variation in the tooltip message

Before:
Screenshot 2023-11-28 at 18 34 51

After (example): 
Screenshot 2023-11-28 at 18 42 33

### DIFF
--- a/src/elements/content-sidebar/SidebarNavSignButton.tsx
+++ b/src/elements/content-sidebar/SidebarNavSignButton.tsx
@@ -21,16 +21,23 @@ export type Props = PlainButtonProps & {
         onComplete: () => void;
         onShow: () => void;
     };
+    CustomFtuxTooltip?: React.ElementType;
+};
+
+type SignButtonProps = {
+    disableTooltip?: boolean;
 };
 
 export const PlaceholderTooltip = ({ children }: { children: React.ReactNode }) => children;
 
-export function SidebarNavSignButton({ blockedReason, intl, status, targetingApi, ...rest }: Props) {
+export function SidebarNavSignButton({ blockedReason, intl, status, targetingApi, CustomFtuxTooltip, ...rest }: Props) {
     const isSignDisabled = !!blockedReason;
     const isTargeted = targetingApi && targetingApi.canShow;
     const FtuxTooltip = !isSignDisabled && isTargeted ? TargetedClickThroughGuideTooltip : PlaceholderTooltip;
     const label = intl.formatMessage(status === 'active' ? messages.boxSignSignature : messages.boxSignRequest);
     const buttonClassName = classnames('bcs-SidebarNavSignButton', { 'bdl-is-disabled': isSignDisabled });
+    const ftuxTitle = intl.formatMessage(messages.boxSignFtuxTitle);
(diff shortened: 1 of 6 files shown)


src/elements/content-sidebar/messages.js
```diff
@@ -39,7 +39,7 @@ const messages = defineMessages({
     },
     boxSignFtuxTitle: {
         id: 'be.contentSidebar.boxSignFtuxTitle',
-        defaultMessage: 'Box Sign - Secure, seamless e-signatures in Box',
+        defaultMessage: 'Box Sign - Secure, seamless e‑signatures in Box',
```
tjuanitas, line 42: is this change intentional? while the dash characters appear the same, I don't see usages of this particular ASCII character in BUIE or some of the other frontend repos

src/elements/content-sidebar/SidebarNavSignButton.tsx
```diff
@@ -21,16 +21,23 @@ export type Props = PlainButtonProps & {
         onComplete: () => void;
         onShow: () => void;
     };
+    CustomFtuxTooltip?: React.ElementType;
```
tjuanitas, line 24: my understanding is that blueprint availability in BUIE is coming soon™ 🤞 

Is this change needed soon or can it wait until after a blueprint->BUIE solution has been implemented? Adding custom components through props to enable blueprint might set an undesirable precedent

src/elements/content-sidebar/messages.js
sstefancu: The change is intentional and replaces the standard breaking hyphen character with its non breaking variant (U+2011) since "e-signatures" it's a single word and should not be broken by browsers like in the image below:

Screenshot 2023-11-28 at 18 42 33

src/elements/content-sidebar/SidebarNavSignButton.tsx
sstefancu: this is was not designed as a "enable blueprint" change; it does not have any blueprint specifics and will allow any kind of custom component (with the required interface) to be passed and used;

this is a pretty standard & popular pattern in UI libraries that allow users to use their own custom components that may be visually different but also because they may encapsulate custom logic which does not belong inside the library 

this is my use case as well, we do intend to pass a custom component that happens to also use a blueprint component however the most important benefit is that it encapsulates specific logic

tjuanitas: I understand the pattern but I'm unsure if it's needed in this case. This component and the FTUX tooltip are only used in the All Files page. Rather than adding a custom component, does it make sense to update the regular FTUX tooltip directly in this repo for all use cases? Is there logic being passed that shouldn't exist here?

i.e. if the `CustomFtuxTooltip` change went live in EUA, there wouldn't be any uses of the regular `FtuxTooltip` component. But even if there were additional usages, why wouldn't we want them to be the same for all uses?

separate question: is the custom one using the same `targetingApi` that's currently used? the FTUX experiences are designed so that multiple experiences are not showing at the same time

PR #3468, opened 2023-11-28
fix(content-explorer): allow onDelete callback to be success on 404

greg-in-a-box:
When `ContentExplorer` is being used at the same time and one user deletes an item, it locked the other user that is attempting to delete the same item in a loading state permanently. 

By, checking for the status code of the request from the `DELETE` request is a `404`, we can continue with the successHandler to release the user from the loading state and perform the refresh as if they deleted it.


jpan-box:
Are users actually shown any sort of success banner during a delete? or does it just end up being a refresh?

I think it could confusing for audit purposes if one user _thinks_ they deleted the content due to the success response from BUIE, but in reality some other user deleted it prior to the first user. 

However, short of implementing custom error messaging I think this is a great compromise for now, and unblocks the users.

jpan-box (dismissed):

greg-in-a-box:
> Are users actually shown any sort of success banner during a delete? or does it just end up being a refresh?
> 
> I think it could confusing for audit purposes if one user _thinks_ they deleted the content due to the success response from BUIE, but in reality some other user deleted it prior to the first user.
> 
> However, short of implementing custom error messaging I think this is a great compromise for now, and unblocks the users.

I dont recall a prompt appearing after the delete action has concluded, from what I do recall, it closes the deleteConfirmationModal then refreshes the folder

src/api/Item.js
```diff
@@ -200,7 +201,12 @@ class Item extends Base {
             .delete({ url })
             .then(this.deleteSuccessHandler)
             .catch((e: $AxiosError<any>) => {
-                this.errorHandler(e);
+                const { response } = e;
+                if (response?.status === HTTP_STATUS_CODE_NOT_FOUND) {
+                    this.deleteSuccessHandler();
```
tjuanitas, line 206: could a 404 happen if the user loses access to the content instead of it being deleted? in that case, it wouldn't be a success right?

tjuanitas: for files that were deleted by other users, it seems more accurate to call `this.errorHandler(e);` since the deletion wasn't a result of this api call

tjuanitas, line 206: this seems odd to do semantically since it goes through the success flow on an error. what about passing in an `errorCallback` to `deleteItem` that refreshes the collection?
https://github.com/box/box-ui-elements/blob/6aef87b8ca7316e871abaa0584abf6e4c918e65c/src/elements/content-explorer/ContentExplorer.js#L1088-L1091

```
this.api.getAPI(type).deleteItem(
    selected,
    () => {
        onDelete(cloneDeep([selected]));
        this.refreshCollection();
    },
    () => {
        this.refreshCollection();
    },
);
```

greg-in-a-box: thats what I ended up doing after reading your comment

jpan-box (approved):

src/elements/content-explorer/ContentExplorer.js
```diff
@@ -1085,10 +1085,16 @@ class ContentExplorer extends Component<Props, State> {
...
-            this.refreshCollection();
-        });
+        this.api.getAPI(type).deleteItem(
+            selected,
+            () => {
+                onDelete(cloneDeep([selected]));
+                this.refreshCollection();
+            },
+            () => {
+                this.refreshCollection();
```
bfoxx1906, line 1095: what happens if it's not a 404?

bfoxx1906 (approved):

tjuanitas (approved):

PR #3473, opened 2023-12-07
feat(label-primitive): pass properties to label

hkisel-box:
This change is required to link Label with input/seelctor/date picker and have "name" under accessibility tab.

### DIFF
--- a/src/components/label/HiddenLabel.tsx
+++ b/src/components/label/HiddenLabel.tsx
@@ -11,8 +11,8 @@ export interface HiddenLabelProps {
     labelElProps?: React.ComponentPropsWithoutRef<'label'>;
 }
 
-const HiddenLabel = ({ children, labelContent, labelElProps }: HiddenLabelProps) => (
-    <LabelPrimitive className="accessibility-hidden" labelContent={labelContent} labelElProps={labelElProps}>
+const HiddenLabel = ({ children, ...rest }: HiddenLabelProps) => (
+    <LabelPrimitive className="accessibility-hidden" {...rest}>
         {children}
     </LabelPrimitive>
 );
--- a/src/components/label/HiddenLabel.tsx
+++ b/src/components/label/HiddenLabel.tsx
@@ -7,10 +7,12 @@ export interface HiddenLabelProps {
     children: React.ReactElement;
     /** Text content of the label */
     labelContent: React.ReactNode;
+    /** Optional props for the label element */
+    labelElProps?: React.ComponentPropsWithoutRef<'label'>;
 }
 
-const HiddenLabel = ({ children, labelContent }: HiddenLabelProps) => (
-    <LabelPrimitive className="accessibility-hidden" labelContent={labelContent}>
+const HiddenLabel = ({ children, labelContent, labelElProps }: HiddenLabelProps) => (
+    <LabelPrimitive className="accessibility-hidden" labelContent={labelContent} labelElProps={labelElProps}>
         {children}
     </LabelPrimitive>
 );
--- a/src/components/label/Label.tsx
+++ b/src/components/label/Label.tsx
@@ -42,7 +42,7 @@ const Label = ({
     showOptionalText,
     hideLabel,
     children,
-    labelElProps = {},
+    labelElProps,
 }: LabelProps) => {
     const labelContent = [
         <span key="labelText">{text}</span>,
--- a/src/components/label/Label.tsx
+++ b/src/components/label/Label.tsx
@@ -30,9 +30,20 @@ export interface LabelProps {
     text: React.ReactNode;
     /** Optional tooltip text for the label */
     tooltip?: React.ReactNode;
+    /** Optional peopierties for label element */
+    labelElProps?: Object;
 }
 
-const Label = ({ text, tooltip, infoTooltip, infoIconProps, showOptionalText, hideLabel, children }: LabelProps) => {
+const Label = ({
+    text,
+    tooltip,
+    infoTooltip,
+    infoIconProps,
+    showOptionalText,
+    hideLabel,
+    children,
+    labelElProps = {},
+}: LabelProps) => {
     const labelContent = [
         <span key="labelText">{text}</span>,
         showOptionalText ? <OptionalFormattedMessage key="optionalMessage" /> : null,
@@ -53,7 +64,7 @@ const Label = ({ text, tooltip, infoTooltip, infoIconProps, showOptionalText, hi
     }
 
     return (
-        <StandardLabel labelContent={labelContent} tooltip={tooltip}>
+        <StandardLabel labelContent={labelContent} tooltip={tooltip} labelElProps={labelElProps}>
             {children}
         </StandardLabel>
     );
--- a/src/components/label/Label.tsx
+++ b/src/components/label/Label.tsx
@@ -24,14 +24,14 @@ export interface LabelProps {
     infoIconProps?: Record<string, any>; // eslint-disable-line @typescript-eslint/no-explicit-any
     /** Tooltip text for the info icon */
     infoTooltip?: React.ReactNode;
+    /** Optional props for the label element */
+    labelElProps?: Object;
     /** Whether to show the `(Optional)` text next to the label for an optional field */
     showOptionalText?: boolean;
     /** The label text */
     text: React.ReactNode;
     /** Optional tooltip text for the label */
     tooltip?: React.ReactNode;
-    /** Optional peopierties for label element */
-    labelElProps?: Object;
 }
 
 const Label = ({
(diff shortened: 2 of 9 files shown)


src/components/label/Label.tsx
```diff
@@ -30,9 +30,20 @@ export interface LabelProps {
     text: React.ReactNode;
     /** Optional tooltip text for the label */
     tooltip?: React.ReactNode;
+    /** Optional peopierties for label element */
+    labelElProps?: Object;
```
tjuanitas, line 34:
```suggestion
    /** Optional props for the label element */
    labelElProps?: Object;
```

same for the other props list

src/components/label/Label.tsx
```diff
@@ -32,7 +34,16 @@ export interface LabelProps {
...
-const Label = ({ text, tooltip, infoTooltip, infoIconProps, showOptionalText, hideLabel, children }: LabelProps) => {
+const Label = ({
+    text,
+    tooltip,
+    infoTooltip,
+    infoIconProps,
+    showOptionalText,
+    hideLabel,
+    children,
+    labelElProps = {},
```
tjuanitas, line 45: nit: should be okay to omit the default value `{}` since the prop type in `StandardLabel` and `LabelPrimitive` is optional and the spread operator in `{...labelElProps}` will handle `undefined`

tjuanitas (dismissed):
lgtm, thanks Anna!

tjuanitas (dismissed):

stanholub (approved):
lgtm!

src/components/label/Label.tsx
```diff
@@ -24,6 +24,8 @@ export interface LabelProps {
     infoIconProps?: Record<string, any>; // eslint-disable-line @typescript-eslint/no-explicit-any
     /** Tooltip text for the info icon */
     infoTooltip?: React.ReactNode;
+    /** Optional props for the label element */
+    labelElProps?: Object;
```
gdangelo, line 28: We can use a more appropriate and specific type here instead of using a generic `Object`. Either `React.ComponentPropsWithoutRef<'label'>` or `React.LabelHTMLAttributes<HTMLLabelElement>`.

gdangelo: Another strategy would be to extend `React.ComponentPropsWithoutRef<'label'>` to allow `<Label>` to take all the normal props of `<label>` instead of having to pass them through an extra prop.

```tsx
export interface LabelProps extends React.ComponentPropsWithoutRef<'label'> {
    //...
}

const Label = ({
    text,
    tooltip,
    infoTooltip,
    infoIconProps,
    showOptionalText,
    hideLabel,
    children,
    ...labelProps
}: LabelProps) => {
    //...

    return (
        <StandardLabel labelContent={labelContent} tooltip={tooltip} {...labelProps}>
            {children}
        </StandardLabel>
    );
};
```

src/components/label/Label.tsx
```diff
@@ -53,7 +64,7 @@ const Label = ({ text, tooltip, infoTooltip, infoIconProps, showOptionalText, hi
     }
 
     return (
-        <StandardLabel labelContent={labelContent} tooltip={tooltip}>
+        <StandardLabel labelContent={labelContent} tooltip={tooltip} labelElProps={labelElProps}>
```
gdangelo, line 67: What about `<HiddenLabel>` also rendered by this component? Should we also pass it the html label attributes?

src/components/label/LabelPrimitive.tsx
```diff
@@ -8,10 +8,12 @@ export interface LabelPrimitiveProps {
     className?: string;
     /** Text content of the label */
     labelContent: React.ReactNode;
+    /** Optional props for the label element */
+    labelElProps?: Object;
```
gdangelo, line 12: Same here, we could simply extend `React.ComponentPropsWithoutRef<'label'>`. 

imo it's preferable to extend those and to add an explicit prop for the `span` element, instead of the other way around, since `label` is the outer/main element rendered by that component. In other words, from a consumer-perspective it is not clear, as of now, that everything else that you pass to `LabelPrimitive` component is forwarded to the inner `span` element.

src/components/label/StandardLabel.tsx
```diff
@@ -8,12 +8,18 @@ export interface StandardLabelProps {
     children: React.ReactElement;
     /** Text content of the label */
     labelContent: React.ReactNode;
+    /** Optional props for the label element */
+    labelElProps?: Object;
```
gdangelo, line 12: Same comment here about extending instead of adding a new prop.

gdangelo (commented):
@hkisel-box Thanks for adding that extra prop to the `Label` component. However, I've left some comments below about an alternative way of extending the component with the regular `label` html attributes, which imo is more appropriate. Let me know what you think?

Also, I see that we also have a `.flow` file for each component. Should we also update that one? Or is it autogenerated somehow?

src/components/label/LabelPrimitive.tsx
hkisel-box: and what is the process for breaking changes, except that it should be indicated with prefix in commits and PR title? don't see anything else in documentation.

tjuanitas: Greg makes a lot of great points but I think you're right that it's a little tricky because of the possible impact. at the moment, we're trying to limit breaking changes since teams are encouraged to migrate to blueprint and so we don't want to cause concern if teams need to still upgrade to BUIE.

I'm not sure why the `{...rest}` was added to the `span` initially (the original implementation is from box-react-ui in 2016). Based on the explicit usages in `StandardLabel` and `HiddenLabel`, it shouldn't be doing anything lol. But I do see tooltips is utilizing the `...rest` props by adding additional props (`componentProps`) to the child component: https://github.com/box/box-ui-elements/blob/13ed6dadab00ba48127c0e3f72feb6b104f81f08/src/components/tooltip/Tooltip.tsx#L385
(not sure if there is an accessibility concern to move the attributes from the `span` to the `label` or if we need to try to keep the tooltip specific ones on the `span`)

up to you if you'd like to keep as-is or update it, just be cautious around the Tooltip use case. the process for breaking changes is just adding the prefix in the commit and PR, no other action needed

hkisel-box: I would leave as-is, and added more precise typing for `labelElProps` props, because we are migrating to `Blueprint` already.
If we decide that it is something that should be refactored, I think that we can create another task for that, and we also need to work on fix for running forms-client linked to BUIE, to test these changes. SWE from BUIE helped me to run that with workaround to test current fix.

src/components/label/HiddenLabel.tsx
```diff
@@ -7,10 +7,12 @@ export interface HiddenLabelProps {
...
     /** Text content of the label */
     labelContent: React.ReactNode;
+    /** Optional props for the label element */
+    labelElProps?: React.ComponentPropsWithoutRef<'label'>;
 }
 
-const HiddenLabel = ({ children, labelContent }: HiddenLabelProps) => (
-    <LabelPrimitive className="accessibility-hidden" labelContent={labelContent}>
+const HiddenLabel = ({ children, labelContent, labelElProps }: HiddenLabelProps) => (
+    <LabelPrimitive className="accessibility-hidden" labelContent={labelContent} labelElProps={labelElProps}>
```
gdangelo, line 15: **nit**:

```suggestion
const HiddenLabel = ({ children, ...rest }: HiddenLabelProps) => (
    <LabelPrimitive className="accessibility-hidden" {...rest}>
```

src/components/label/StandardLabel.tsx
```diff
@@ -8,12 +8,18 @@ export interface StandardLabelProps {
...
+    labelElProps?: React.ComponentPropsWithoutRef<'label'>;
     /** Optional tooltip text for the label */
     tooltip?: React.ReactNode;
 }
 
-const StandardLabel = ({ children, labelContent, tooltip }: StandardLabelProps) => {
-    const label = <LabelPrimitive labelContent={labelContent}>{children}</LabelPrimitive>;
+const StandardLabel = ({ children, labelElProps, labelContent, tooltip }: StandardLabelProps) => {
+    const label = (
+        <LabelPrimitive labelElProps={labelElProps} labelContent={labelContent}>
```
gdangelo, line 19: **nit**:

```suggestion
const StandardLabel = ({ children, tooltip, ...rest }: StandardLabelProps) => {
    const label = (
        <LabelPrimitive {...rest}>
```

gdangelo (approved):
LGTM

tjuanitas (approved):

greg-in-a-box (approved):

mbas-box (approved):

PR #3474, opened 2023-12-07
feat(content-explorer): Add auto selected items to content explorer

SkyisAakash:

### DIFF
--- a/src/features/content-explorer/content-explorer-modal-container/ContentExplorerModalContainer.js
+++ b/src/features/content-explorer/content-explorer-modal-container/ContentExplorerModalContainer.js
@@ -70,6 +70,10 @@ class ContentExplorerModalContainer extends Component {
         initialFoldersPath: FoldersPathPropType.isRequired,
         /** Initial items that will show up as selected */
         initialSelectedItems: PropTypes.object,
+        /** Items that will show up as selected */
+        autoSelectedItems: PropTypes.object,
+        // Called when items are deselected
+        deselectAutoSelectedItem: PropTypes.func,
         /**
          * Called when the current folder changes
          *
--- a/src/features/content-explorer/content-explorer-modal-container/ContentExplorerModalContainer.js
+++ b/src/features/content-explorer/content-explorer-modal-container/ContentExplorerModalContainer.js
@@ -72,8 +72,6 @@ class ContentExplorerModalContainer extends Component {
         initialSelectedItems: PropTypes.object,
         /** Items that will show up as selected */
         autoSelectedItems: PropTypes.object,
-        /** Called when items are deselected */
-        deselectAutoSelectedItem: PropTypes.func,
         /**
          * Called when the current folder changes
          *
--- a/src/features/content-explorer/content-explorer-modal-container/ContentExplorerModalContainer.js
+++ b/src/features/content-explorer/content-explorer-modal-container/ContentExplorerModalContainer.js
@@ -71,7 +71,7 @@ class ContentExplorerModalContainer extends Component {
         /** Initial items that will show up as selected */
         initialSelectedItems: PropTypes.object,
         /** Items that will show up as selected */
-        autoSelectedItems: PropTypes.object,
+        controlledSelectedItems: PropTypes.object,
         /**
          * Called when the current folder changes
          *
--- a/src/features/content-explorer/content-explorer-modal-container/ContentExplorerModalContainer.js
+++ b/src/features/content-explorer/content-explorer-modal-container/ContentExplorerModalContainer.js
@@ -72,7 +72,7 @@ class ContentExplorerModalContainer extends Component {
         initialSelectedItems: PropTypes.object,
         /** Items that will show up as selected */
         autoSelectedItems: PropTypes.object,
-        // Called when items are deselected
+        /** Called when items are deselected */
         deselectAutoSelectedItem: PropTypes.func,
         /**
          * Called when the current folder changes
--- a/src/features/content-explorer/content-explorer-modal/ContentExplorerModal.js
+++ b/src/features/content-explorer/content-explorer-modal/ContentExplorerModal.js
@@ -17,7 +17,6 @@ type Props = {
     className?: string,
     customInput?: React.ComponentType<any>,
     description?: string,
-    deselectAutoSelectedItem?: Function,
     hasFolderTreeBreadcrumbs: boolean,
     includeSubfoldersProps?: Object,
     isNoSelectionAllowed?: boolean,
--- a/src/features/content-explorer/content-explorer-modal/ContentExplorerModal.js
+++ b/src/features/content-explorer/content-explorer-modal/ContentExplorerModal.js
@@ -13,13 +13,13 @@ import './ContentExplorerModal.scss';
 type Props = {
     additionalColumns?: Array<React.ComponentType<Column>>,
     /** Items that will show up as selected */
-    autoSelectedItems: PropTypes.object,
+    autoSelectedItems: Object,
     breadcrumbProps?: BreadcrumbProps,
     className?: string,
     customInput?: React.ComponentType<any>,
     description?: string,
     // Called when items are deselected
-    deselectAutoSelectedItem: PropTypes.func,
+    deselectAutoSelectedItem: Function,
     hasFolderTreeBreadcrumbs: boolean,
     includeSubfoldersProps?: Object,
     isNoSelectionAllowed?: boolean,
--- a/src/features/content-explorer/content-explorer-modal/ContentExplorerModal.js
+++ b/src/features/content-explorer/content-explorer-modal/ContentExplorerModal.js
@@ -12,9 +12,9 @@ import './ContentExplorerModal.scss';
 
 type Props = {
     additionalColumns?: Array<React.ComponentType<Column>>,
-    autoSelectedItems?: Object,
     breadcrumbProps?: BreadcrumbProps,
     className?: string,
+    controlledSelectedItems?: Object,
     customInput?: React.ComponentType<any>,
     description?: string,
     hasFolderTreeBreadcrumbs: boolean,
--- a/src/features/content-explorer/content-explorer-modal/ContentExplorerModal.js
+++ b/src/features/content-explorer/content-explorer-modal/ContentExplorerModal.js
@@ -12,14 +12,12 @@ import './ContentExplorerModal.scss';
 
 type Props = {
     additionalColumns?: Array<React.ComponentType<Column>>,
-    /** Items that will show up as selected */
-    autoSelectedItems: Object,
+    autoSelectedItems?: Object,
     breadcrumbProps?: BreadcrumbProps,
     className?: string,
     customInput?: React.ComponentType<any>,
     description?: string,
-    // Called when items are deselected
-    deselectAutoSelectedItem: Function,
+    deselectAutoSelectedItem?: Function,
     hasFolderTreeBreadcrumbs: boolean,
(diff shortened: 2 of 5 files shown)


src/features/content-explorer/content-explorer-modal-container/ContentExplorerModalContainer.js
```diff
@@ -70,6 +70,10 @@ class ContentExplorerModalContainer extends Component {
         initialFoldersPath: FoldersPathPropType.isRequired,
         /** Initial items that will show up as selected */
         initialSelectedItems: PropTypes.object,
+        /** Items that will show up as selected */
+        autoSelectedItems: PropTypes.object,
```
tjuanitas, line 74: for my understanding: does an item in `autoSelectedItems` only deselect when used with `deselectAutoSelectedItem`?

i.e. what happens if a value is passed for the `autoSelectedItems` prop but no callback is passed for the `deselectAutoSelectedItem` prop?

src/features/content-explorer/content-explorer-modal-container/ContentExplorerModalContainer.js
```diff
@@ -70,6 +70,10 @@ class ContentExplorerModalContainer extends Component {
...
+        // Called when items are deselected
```
tjuanitas, line 75: nit: the comment style for these props are `/** ... */`
```
/** Called when an item is deselected */
```

src/features/content-explorer/content-explorer-modal/ContentExplorerModal.js
```diff
@@ -12,10 +12,14 @@ import './ContentExplorerModal.scss';
...
 type Props = {
     additionalColumns?: Array<React.ComponentType<Column>>,
+    /** Items that will show up as selected */
+    autoSelectedItems: Object,
     breadcrumbProps?: BreadcrumbProps,
     className?: string,
     customInput?: React.ComponentType<any>,
     description?: string,
+    // Called when items are deselected
+    deselectAutoSelectedItem: Function,
```
tjuanitas, line 22: nit: the comments aren't included in the first. my guess because the container already has the descriptions for them

tjuanitas, line 22: are these optional props?
```
    autoSelectedItems?: Object,
    ...
    deselectAutoSelectedItem?: Function,
```

src/features/content-explorer/content-explorer-modal-container/ContentExplorerModalContainer.js
tjuanitas: same for the other files

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -492,9 +498,11 @@ class ContentExplorer extends Component {
             listWidth,
             listHeight,
             searchInputProps,
+            autoSelectedItems,
```
tjuanitas, line 501: nit: this list is almost sorted. could you move this to the top?

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -492,9 +498,11 @@ class ContentExplorer extends Component {
...
             ...rest
         } = this.props;
         const { isInSearchMode, foldersPath, selectedItems, isSelectAllChecked } = this.state;
+        const allSelectedItems = { ...selectedItems, ...autoSelectedItems };
```
tjuanitas, line 505: is the internal state for the initial render out of sync with the UI? is that okay? is this prop intended to be used in combination with `initialSelectedItems` for those cases?

https://github.com/box/box-ui-elements/blob/529f7c54def5dd885842ab7e33f2200b4f1e5fff/src/features/content-explorer/content-explorer/ContentExplorer.js#L176

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -492,9 +498,11 @@ class ContentExplorer extends Component {
...
         const isViewingSearchResults = isInSearchMode && foldersPath.length === 1;
         const currentFolder = this.getCurrentFolder();
         const contentExplorerProps = omit(rest, [
```
tjuanitas, line 508: this should include `deselectAutoSelectedItem`

tjuanitas (commented):
I think consider using `onSelectedItemsUpdate` instead of adding `deselectAutoSelectedItem` if possible. they look similar but I might not have all the context

src/features/content-explorer/content-explorer/ContentExplorer.js
tjuanitas: I feel like there are other cases where it's possible for the internal state to be out of sync with the UI if the user tries to deselect an item from `autoSelectedItems`

src/features/content-explorer/content-explorer-modal-container/ContentExplorerModalContainer.js
SkyisAakash: No, autoSelectedItems are not disabled. That means that user can deselect an autoSelectedItem. deselectAutoSelectedItem prop is used to clear autoSelectedItem in parent component when methods like `deselectAll` are called in contentExplorer

src/features/content-explorer/content-explorer-modal/ContentExplorerModal.js
SkyisAakash: Yes

src/features/content-explorer/content-explorer-modal-container/ContentExplorerModalContainer.js
```diff
@@ -70,6 +70,10 @@ class ContentExplorerModalContainer extends Component {
...
+        deselectAutoSelectedItem: PropTypes.func,
```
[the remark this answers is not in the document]
SkyisAakash: This is a good point, I think using `onSelectedItemsUpdate` should be sufficient. I have removed prop `deselectAutoSelectedItem` so now there is only one new prop introduced. :)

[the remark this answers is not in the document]
SkyisAakash: Done

src/features/content-explorer/content-explorer/ContentExplorer.js
SkyisAakash: The prop autoSelectedItems is meant for allowing component to have certain items selected without user having to click on those after initial render. So ideal usecase would be if we want to trigger selecting certain items as part of an action/event. That said, if autoSelectedItems has items along with autoSelectedItems then both of them would appear selected. Perhaps we want to make it clear in description that these two should be use exclusively? 
If user tries to deselect an item from autoSelectedItems then we should expect parent component to handle updating value of prop autoSelectedItems based on their requirement.

src/features/content-explorer/content-explorer/__tests__/ContentExplorer.test.js
```diff
@@ -739,6 +749,13 @@ describe('features/content-explorer/content-explorer/ContentExplorer', () => {
             expect(result).toStrictEqual({});
         });
 
+        test('should remove items from autoSelectedItems when unselectAll is called', () => {
+            const wrapper = renderComponent({ items });
+            wrapper.setState({ selectedItems });
+            const result = wrapper.instance().unselectAll();
+            expect(result).toStrictEqual({});
+        });
```
tjuanitas, line 757: are we okay to remove this test? it looks the same as the one above

src/features/content-explorer/content-explorer/__tests__/ContentExplorer.test.js
```diff
@@ -787,4 +804,23 @@ describe('features/content-explorer/content-explorer/ContentExplorer', () => {
...
+            const autoSelectedItems = { item1: { id: 'item1', name: 'name1' } };
+            const mockEvent = { stopPropagation: () => {} };
+
+            const wrapper = renderComponent({
+                items,
+                autoSelectedItems,
+                contentExplorerMode: ContentExplorerModes.MULTI_SELECT,
+            });
+            wrapper.instance().handleItemClick({ event: mockEvent, index: 1 });
+            expect(Object.keys(wrapper.state('selectedItems')).length).toEqual(2);
```
tjuanitas, line 823: tiny nit: we've been trying to use `toBe` for primitive values
```suggestion
            expect(Object.keys(wrapper.state('selectedItems')).length).toBe(2);
```

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -26,6 +26,8 @@ class ContentExplorer extends Component {
          * Each column has to be a Column element
          */
         additionalColumns: PropTypes.arrayOf(PropTypes.element),
+        /** Items that will show up as selected */
+        autoSelectedItems: PropTypes.object,
```
tjuanitas, line 30: what are your thoughts on something like `controlledSelectedItems`? to convey that the parent component can force a selection over the user

tjuanitas (commented):
I had some confusion earlier but it makes sense now 👍

src/features/content-explorer/content-explorer/__tests__/ContentExplorer.test.js
SkyisAakash: Ah yes, missed removing this one while removing the additional prop.

src/features/content-explorer/content-explorer/ContentExplorer.js
SkyisAakash: That is more descriptive yes.

SkyisAakash: Updated

tjuanitas (approved):

src/features/content-explorer/content-explorer/ContentExplorer.js
```diff
@@ -329,9 +331,10 @@ class ContentExplorer extends Component {
     };
 
     handleItemClick = ({ event, index }) => {
-        const { contentExplorerMode, items, onSelectItem, onSelectedItemsUpdate } = this.props;
+        const { contentExplorerMode, items, onSelectItem, onSelectedItemsUpdate, controlledSelectedItems } = this.props;
         const { selectedItems } = this.state;
         const item = items[index];
+        const allSelectedItems = { ...selectedItems, ...controlledSelectedItems };
```
apieceofbart, line 337: I was just wondering, since you're doing this in few place, maybe instead create a method:
```
    getAllSelectedItems = () => {
        return { ...this.state.selectedItems, ...this.props.controlledSelectedItems };
    };
 ```

File: src/features/content-explorer/content-explorer/ContentExplorer.js [hunk 12]
```diff
@@ -520,18 +524,18 @@ class ContentExplorer extends Component {
             // is selected and that item's isActionDisabled is false, we enable the action button
             areActionButtonsDisabled =
                 (selectedItemsIds.length === 0 && !isNoSelectionAllowed) ||
-                (selectedItemsIds.length === 1 && selectedItems[selectedItemsIds[0]].isActionDisabled);
+                (selectedItemsIds.length === 1 && allSelectedItems[selectedItemsIds[0]].isActionDisabled);
```
apieceofbart, line 527: I was wondering if we could extract `allSelectedItems[selectedItemsIds[0]].isActionDisabled` to a variable since it's used in 3 places below (I know, this was like that before):
```
   const isActionDisabled = allSelectedItems[selectedItemsIds[0]]?.isActionDisabled;
```

apieceofbart (approved):

PR #3475, opened 2023-12-14
fix(search-form): extract search actions component

tjuanitas:
This should address an issue in the features version of ContentExplorer where passing a `searchInputProps` prop prevented the search input to not submit. I honestly couldn't figure out the exact root cause but extracting the `makeLoadable` HOC out of the `render` method seemed to fix it.

### DIFF
--- a/src/components/search-form/SearchActions.tsx
+++ b/src/components/search-form/SearchActions.tsx
@@ -1,0 +1,50 @@
+import * as React from 'react';
+import { injectIntl, IntlShape } from 'react-intl';
+
+import ClearBadge16 from '../../icon/fill/ClearBadge16';
+import Search16 from '../../icon/fill/Search16';
+import makeLoadable from '../loading-indicator/makeLoadable';
+
+import messages from './messages';
+
+export interface SearchActionsProps {
+    /** Whether to render an interactive search button */
+    hasSubmitAction: boolean;
+    /** Intl object */
+    intl: IntlShape;
+    /** Called when clear button is clicked */
+    onClear: (event: React.SyntheticEvent<HTMLButtonElement>) => void;
+}
+
+const SearchActions = ({ hasSubmitAction, intl, onClear }: SearchActionsProps) => {
+    const { formatMessage } = intl;
+
+    return (
+        <div className="action-buttons">
+            {hasSubmitAction ? (
+                <button
+                    type="submit"
+                    className="action-button search-button"
+                    title={formatMessage(messages.searchButtonTitle)}
+                >
+                    <Search16 />
+                </button>
+            ) : (
+                <div className="action-button search-button">
+                    <Search16 />
+                </div>
+            )}
+            <button
+                className="action-button clear-button"
+                onClick={onClear}
+                title={formatMessage(messages.clearButtonTitle)}
+                type="button"
+            >
+                <ClearBadge16 />
+            </button>
+        </div>
+    );
+};
+
+export { SearchActions as SearchActionsBase };
+export default makeLoadable(injectIntl(SearchActions));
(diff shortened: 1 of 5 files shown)


src/components/search-form/SearchForm.js
```diff
@@ -177,35 +160,6 @@ class SearchFormBase extends React.Component<Props, State> {
...
-                    onClick={this.onClearHandler}
-                    title={formatMessage(messages.clearButtonTitle)}
-                    type="button"
-                >
-                    <ClearBadge16 />
-                </button>
-            </div>
-        );
-
-        const LoadableSearchActions = makeLoadable(SearchActions);
```
tjuanitas, line 207: legacy docs state to not use HOCs in the render method: https://legacy.reactjs.org/docs/higher-order-components.html#dont-use-hocs-inside-the-render-method

I don't know if this is the direct cause of the original issue but it ended up fixing it without changing the other logic

src/components/search-form/__tests__/SearchForm.test.js
```diff
@@ -124,8 +124,12 @@ describe('components/search-form/SearchForm', () => {
...
     test('should set the onClearHandler to the clear button onClick prop', () => {
         const wrapper = shallow(<SearchForm intl={intlShape} />).shallow();
-        const lodableComponent = wrapper.find('LoadableSearchActions').shallow();
-        const searchActions = lodableComponent.shallow();
+        // Sift through the nested HOCs to find the correct element
+        const searchActions = wrapper
+            .find('LoadableSearchActions')
+            .shallow()
+            .shallow()
+            .shallow();
```
tjuanitas, line 132: admittedly not my best work...I briefly looked into using `mount` and alternatives but I selfishly didn't want to spend too much time on this component and changes

greg-in-a-box: shouldn't we being use two `dive()`s here instead of two `shallow()`?

greg-in-a-box: Also should this test be here? shouldn't it be in a new test file for `SearchActions` this way you dont have to go through the HOC?

greg-in-a-box (commented):
left some comments

src/components/search-form/__tests__/SearchForm.test.js
tjuanitas: > shouldn't we being use two dive()s here instead of two shallow()?

i mainly copied from the existing test since I wasn't sure the exact differences. but it looks like `dive()` is preferred?https://github.com/enzymejs/enzyme/issues/1798

i'll update this to `dives()`

tjuanitas: > Also should this test be here? shouldn't it be in a new test file for SearchActions this way you dont have to go through the HOC?

yeah this was one of the alternatives I was looking into but then moving it to a new `SearchActions` file I think sort of defeats the intention of the original test since it's testing the `onClearHandler` in `SearchForm` (the parent component)

tjuanitas: I think the ideal way to test it would be to convert this to RTL I guess and test the button click functionality? but then I was like: ahhh we'll save that for another day...

src/components/search-form/SearchActions.tsx
```diff
@@ -0,0 +1,50 @@
...
+export interface SearchActionsProps {
+    /** Whether to render an interactive search button */
+    hasSubmitAction: boolean;
+    /** Intl object */
+    intl: IntlShape;
+    /** Called when clear button is clicked */
+    onClear: (event: React.SyntheticEvent<HTMLButtonElement>) => void;
+}
+
+const SearchActions = ({ hasSubmitAction, intl, onClear }: SearchActionsProps) => {
```
tjuanitas, line 19: fyi this component is just copied from `SearchForm`. leaving class names and whatnot as the same so it doesn't break existing functionality for apps

pusch-e (approved):

menaczar (approved):

PR #3478, opened 2023-12-20
fix(content-sidebar): increase z-index to stack above annotations

tjuanitas:
### Context
When the sidebar is opened on medium sized screens, annotations on the file will render above the sidebar.

Updating the `z-index` value of `.bcs.bcs-is-open` will layer the sidebar above the annotations. See [BaseAnnotator.scss](https://github.com/box/box-annotations/blob/master/src/common/BaseAnnotator.scss) for `z-index` values.

Screenshot 2023-12-20 at 10 30 54 AM

---

### Before

Closed sidebar

Screenshot 2023-12-20 at 9 58 43 AM

Opened sidebar

Screenshot 2023-12-20 at 9 58 29 AM

---

### After

Screenshot 2023-12-20 at 11 07 14 AM

### DIFF
--- a/src/elements/content-sidebar/ContentSidebar.scss
+++ b/src/elements/content-sidebar/ContentSidebar.scss
@@ -43,7 +43,7 @@ $sidebarWidth: $sidebarTabsWidth + $sidebarContentWidth;
             &.bcs-is-open {
                 position: absolute;
                 bottom: 0;
-                z-index: 2;
+                z-index: 5; // Stack sidebar above annotations. See z-index values https://github.com/box/box-annotations/blob/master/src/common/BaseAnnotator.scss
                 max-height: 100%;
             }
         }


src/elements/content-sidebar/ContentSidebar.scss
```diff
@@ -32,6 +32,7 @@ $sidebarWidth: $sidebarTabsWidth + $sidebarContentWidth;
 .is-responsive-web {
     @include breakpoint($medium-screen) {
         .be.bcs {
+            z-index: 5; // Stack Content Sidebar above Box Annotations. See z-index values https://github.com/box/box-annotations/blob/master/src/common/BaseAnnotator.scss
```
tjuanitas, line 35: verified no issues for modals, tooltips, and dropdowns

benjamin-shen (approved):
Do we expect annotations to have even more layers in the future? Should we proactively bump up this z-index further?

megansmith-box (approved):

tjuanitas:
> Do we expect annotations to have even more layers in the future? Should we proactively bump up this z-index further?

would have to ask Preview but I don't think there are intentions of enhancing annotations with more layers in the foreseeable future. the z-indexes were updated in this PR: https://github.com/box/box-annotations/pull/703 but it looks like to address a one-off issue

src/elements/content-sidebar/ContentSidebar.scss
```diff
@@ -43,7 +43,7 @@ $sidebarWidth: $sidebarTabsWidth + $sidebarContentWidth;
             &.bcs-is-open {
                 position: absolute;
                 bottom: 0;
-                z-index: 2;
+                z-index: 5; // Stack sidebar above annotations. See z-index values https://github.com/box/box-annotations/blob/master/src/common/BaseAnnotator.scss
```
mickr, line 46: Should we make it higher? I'm not sure we ever want something at a higher z-index

mickr (approved):

src/elements/content-sidebar/ContentSidebar.scss
tjuanitas: yeah that makes sense but also not sure what is a safe and intentional value to set it to. I went back n forth on some alternatives such as adding a `z-index` to the preview container so it creates a stacking context always contained under the sidebar

but ultimately I wasn't confident with how these changes might interact with floating components like tooltips, modals, menus, flyouts, etc. also as we move to blueprint components (coming soon™) how those might interact with the stacking contexts of preview and sidebar

at the moment this seems to be the safest thing and as we handle migrating UI elements to blueprint, we can revisit the staking contexts

PR #3484, opened 2024-01-09
feat(docgen): docgen sidebar base

rustam-e:
Added: a docgen sidebar
1. loads if the previewed file is a docgen template
2. renders list of tags
3. includes empty, loading and error states

Screenshot 2024-01-29 at 14 51 56


src/constants.js
```diff
@@ -396,6 +396,7 @@ export const SIDEBAR_VIEW_DETAILS: 'details' = 'details';
 export const SIDEBAR_VIEW_METADATA: 'metadata' = 'metadata';
 export const SIDEBAR_VIEW_ACTIVITY: 'activity' = 'activity';
 export const SIDEBAR_VIEW_VERSIONS: 'versions' = 'versions';
+export const SIDEBAR_DOCGEN: 'docgen' = 'docgen';
```
tjuanitas, line 399: nit:
```suggestion
export const SIDEBAR_VIEW_DOCGEN: 'docgen' = 'docgen';
```
aligns with the other views

src/elements/common/messages.js
```diff
@@ -417,6 +417,11 @@ const messages = defineMessages({
         description: 'Title for the preview activity feed.',
         defaultMessage: 'Activity',
     },
+    sidebarDocgenTitle: {
+        id: 'be.sidebarDocgenTitle',
+        description: 'Title for the DocGen sidebar tab.',
+        defaultMessage: 'Doc Gen Tags',
```
tjuanitas, line 423: interesting, is this copy in the design? would normal users know these abbreviations?

src/elements/content-sidebar/ContentSidebar.js
```diff
@@ -24,6 +24,7 @@ import { withFeatureProvider } from '../common/feature-checking';
 import { withLogger } from '../common/logger';
 
 import type { DetailsSidebarProps } from './DetailsSidebar';
+import type { DocgenSidebarProps } from './DocgenSidebar';
```
tjuanitas, line 27: would the casing be `DocGen` rather than treated as one word?

src/elements/content-sidebar/ContentSidebar.js
```diff
@@ -50,6 +51,7 @@ type Props = {
     currentUser?: User,
     defaultView: string,
     detailsSidebarProps: DetailsSidebarProps,
+    docgenPreviewSidebarProps: DocgenSidebarProps,
```
tjuanitas, line 54: should `Preview` be removed from this name or is this intentional? i.e. is this related to ContentPreview or some other use of the word?

src/elements/content-sidebar/DocgenSidebar.js
```diff
@@ -0,0 +1,86 @@
+// @ts-check
+/**
+ * @flow
+ * @file Sign sidebar component
```
tjuanitas, line 4: should it be?
```suggestion
 * @file Doc Gen sidebar component
```

we might even change to use the full name. I'm not familiar with what this is though 
 document generation / generator?

src/elements/content-sidebar/DocgenSidebar.js
```diff
@@ -0,0 +1,86 @@
...
+import { EVENT_JS_READY } from '../common/logger/constants';
+import SidebarContent from './SidebarContent';
+
+type EUAProps = {
+    enabled: boolean,
+    getDocgenTags: Function,
+};
+
+type ExternalProps = {
+    // isFeatureEnabled: boolean,
```
tjuanitas, line 26: do we need this commented code or can we remove?

src/elements/content-sidebar/DocgenSidebar.js
```diff
@@ -0,0 +1,86 @@
...
+    WithLoggerProps;
+
+type State = { tags: Array<DocgenTag> };
+
+const MARK_NAME_JS_READY = `${ORIGIN_METADATA_SIDEBAR}_${EVENT_JS_READY}`;
+
+mark(MARK_NAME_JS_READY);
+
+// TO DO: implement actual sidebar content in separate tickets
+class DocgenSidebar extends React.PureComponent<Props, State> {
```
tjuanitas, line 49: same comment about abbreviation if this should be `DocGenSidebar` and similar places

src/elements/content-sidebar/DocgenSidebar.js
```diff
@@ -0,0 +1,86 @@
...
+import type { WithLoggerProps } from '../../common/types/logging';
+import { ORIGIN_METADATA_SIDEBAR, SIDEBAR_VIEW_METADATA } from '../../constants';
+import { mark } from '../../utils/performance';
+import { withAPIContext } from '../common/api-context';
+import { withErrorBoundary } from '../common/error-boundary';
+import { withLogger } from '../common/logger';
+import { EVENT_JS_READY } from '../common/logger/constants';
+import SidebarContent from './SidebarContent';
+
+type EUAProps = {
```
tjuanitas, line 20: don't think it should have knowledge of the service name

src/elements/content-sidebar/DocgenSidebar.js
```diff
@@ -0,0 +1,86 @@
...
+    enabled: boolean,
+    getDocgenTags: Function,
```
tjuanitas, line 22: could these be moved to `ExternalProps` directly? is the nested `docgenPreviewSidebarProps` needed?

src/elements/content-sidebar/DocgenSidebar.js
```diff
@@ -0,0 +1,86 @@
...
+        });
+    }
+
+    render() {
+        const { isLoading } = this.props;
+        const { tags } = this.state;
+
+        return (
+            <SidebarContent sidebarView={SIDEBAR_VIEW_METADATA} title="Box Docgen">
+                <input placeholder="Search" />
```
tjuanitas, line 66: translations needed for `title` and `placeholder`?

src/elements/content-sidebar/DocgenSidebar.js
```diff
@@ -0,0 +1,86 @@
...
+                {isLoading && <div>Loading</div>}
+                {tags.length > 0 ? (
+                    <div>
+                        {tags.map(tag => (
+                            <p key={tag.json_paths[0]}>{tag.json_paths[0]}</p>
+                        ))}
+                    </div>
+                ) : (
+                    <div>No Tags found</div>
```
tjuanitas, line 75: translations needed here too? although maybe this isn't the final design?

src/elements/content-sidebar/SidebarNav.js
```diff
@@ -100,6 +104,15 @@ const SidebarNav = ({
                             <IconMetadataThick />
                         </SidebarNavButton>
                     )}
+                    {isDocgenTemplate && (
+                        <SidebarNavButton
+                            data-resin-target={SIDEBAR_NAV_TARGETS.DOCGEN}
+                            sidebarView={SIDEBAR_DOCGEN}
+                            tooltip="Box Doc Gen"
+                        >
+                            <Tag16 className="bcs-SidebarNavSignButton-icon" />
```
tjuanitas, line 113: the class name was intended for the Sign button and doesn't follow SUIT naming for this file. should there be a general class created?

src/elements/content-sidebar/SidebarNav.js
```diff
@@ -19,14 +19,16 @@ import SidebarToggle from './SidebarToggle';
...
+    SIDEBAR_DOCGEN,
     SIDEBAR_VIEW_ACTIVITY,
     SIDEBAR_VIEW_DETAILS,
     SIDEBAR_VIEW_METADATA,
+    SIDEBAR_VIEW_SKILLS,
 } from '../../constants';
 import { useFeatureConfig } from '../common/feature-checking';
 import type { NavigateOptions, AdditionalSidebarTab } from './flowTypes';
 import './SidebarNav.scss';
+import Tag16 from '../../icon/fill/Tag16';
```
tjuanitas, line 31: nit: this import could be moved up in this list next to the components

src/elements/content-sidebar/Sidebar.js
```diff
@@ -259,6 +319,7 @@ class Sidebar extends React.Component<Props, State> {
                                 hasMetadata={hasMetadata}
                                 hasSkills={hasSkills}
                                 isOpen={isOpen}
+                                isDocgenTemplate={this.state.isDocgenTemplate}
```
tjuanitas, line 322: nit: alpha sort this prop

src/elements/content-sidebar/SidebarNav.js
```diff
@@ -53,6 +56,7 @@ const SidebarNav = ({
     intl,
     isOpen,
     onNavigate,
+    isDocgenTemplate = false,
```
tjuanitas, line 59: nit: could alphasort here as well

src/elements/content-sidebar/__tests__/DocgenSidebar.test.js
```diff
@@ -0,0 +1,69 @@
...
+                ],
+            },
+        }),
+    ),
+};
+
+const defaultProps = {
+    docgenPreviewSidebarProps,
+};
+describe('elements/content-sidebar/Docgen/DocgenSidebar', () => {
```
tjuanitas, line 40:
```suggestion
describe('elements/content-sidebar/DocgenSidebar', () => {
```

src/elements/content-sidebar/SidebarPanels.js
```diff
@@ -69,6 +73,7 @@ const MARK_NAME_JS_LOADING_DETAILS = `${ORIGIN_DETAILS_SIDEBAR}${BASE_EVENT_NAME
 const MARK_NAME_JS_LOADING_ACTIVITY = `${ORIGIN_ACTIVITY_SIDEBAR}${BASE_EVENT_NAME}`;
 const MARK_NAME_JS_LOADING_SKILLS = `${ORIGIN_SKILLS_SIDEBAR}${BASE_EVENT_NAME}`;
 const MARK_NAME_JS_LOADING_METADATA = `${ORIGIN_METADATA_SIDEBAR}${BASE_EVENT_NAME}`;
+const MARK_NAME_JS_LOADING_DOCGEN = `docgen_sidebar${BASE_EVENT_NAME}`;
```
tjuanitas, line 76: could we match and use a constant here?

src/elements/content-sidebar/Sidebar.js
```diff
@@ -91,14 +96,63 @@ class Sidebar extends React.Component<Props, State> {
...
+            this.navigateToDefaultTab();
+        }
+    }
+
+    componentDidMount() {
+        // if docgen feature is enabled, load metadata to check whether file is a docgen template
+        if (this.props.features?.docgen?.enabled) {
+            this.fetchMetadata();
+        }
+    }
```
tjuanitas, line 151: i'm not familiar with how the Doc Gen feature but should this logic be moved to the DocGenSidebar component?

tjuanitas (commented):
took a look but I don't have a lot of context on this feature. do you have a teammate that's able to review as well? also really important to get Preview's input

src/elements/common/messages.js
rustam-e: yes this is the name of the product

src/elements/content-sidebar/DocgenSidebar.js
rustam-e: this is just placeholder content, the actual content will be implemented separately

src/elements/content-sidebar/DocgenSidebar.js
rustam-e: this is just placeholder content, the actual content will be implemented separately

src/elements/content-sidebar/Sidebar.js
rustam-e: no because this logic determines whether to enable the docgen sidebar in the first place, we have a separate ticket investigating whether this logic can be returned via monolith call that preview makes in order to avoid making extra metadata call - in that case most of this logic could be removed

src/elements/content-sidebar/DocgenSidebar.js
```diff
@@ -0,0 +1,83 @@
+// @ts-check
```
jackinf, line 1: is this VSCode specific? 
(yes: suggest removal)

src/elements/content-sidebar/__tests__/DocgenSidebar.test.js
```diff
@@ -0,0 +1,69 @@
...
+
+        await wrapper.update();
+
+        const tagList = wrapper.find('p');
+        expect(tagList).toHaveLength(4);
+
+        expect(wrapper).toMatchSnapshot();
+    });
+
+    // TO DO should render empty state
```
jackinf, line 64: do you need to resolve these todo-s?

src/elements/content-sidebar/DocgenSidebar.js
jackinf: if you want you could add a js comment with a jira ticket that's going to remove/replace this hardcoded text

src/elements/content-sidebar/DocgenSidebar.js
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    render() {
+        const { isLoading } = this.props;
+        const { tags } = this.state;
+
+        return (
+            <SidebarContent sidebarView={SIDEBAR_VIEW_METADATA} title="Box DocGen">
+                <input placeholder="Search" />
+                {isLoading && <div>Loading</div>}
```
jackinf, line 64: translation for this one as well?

src/elements/content-sidebar/__tests__/DocgenSidebar.test.js
rustam-e: will be handled later - I wanted to work iteratively instead of creating a single huge PR

src/elements/content-sidebar/DocgenSidebar.js
rustam-e: the purpose of this PR is to display sidebar tab by default and add a wrapper page for the Docgen Sidebar - all of the content inside DocgenSidebar component is just dummy content that will be reworked separately.

I'm trying to split the epic I'm working on into small units of work that are easy to review instead of having entire epic in 1 PR

jackinf (approved):

src/elements/content-sidebar/DocgenSidebar.js
jackinf: great!

src/elements/content-sidebar/DocGenSidebar/TagTree.tsx
```diff
@@ -0,0 +1,29 @@
...
+import './DocGenSidebar.scss';
+import { JsonData } from './types';
+
+interface TagTreeProps {
+    data: JsonData | string[];
+    level?: number;
+}
+
+const TagTree = ({ data, level = 0 }: TagTreeProps) => {
+    if (Array.isArray(data)) {
```
jackinf, line 11: in the interface, you say that it can also be JsonData. Should ```Array.isArray(data)``` return false if `data` is JsonData?

src/elements/content-sidebar/DocGenSidebar/NoTagsAvailable.tsx
```diff
@@ -0,0 +1,13 @@
+import * as React from 'react';
+import NoTagsIcon from './NoTagsIcon';
+
+const NoTagsAvailable = () => (
+    <div className="docgen-empty-state">
+        <NoTagsIcon className="docgen-empty-state--icon" />
+        <div>This document has no tags</div>
```
jackinf, line 7: Are translations required?

src/elements/content-sidebar/DocGenSidebar/TagTree.tsx
jackinf: discussed

jackinf (approved):

src/elements/content-sidebar/DocGenSidebar/NoTagsAvailable.tsx
rustam-e: good catch

src/elements/content-sidebar/Sidebar.js
```diff
@@ -91,14 +96,68 @@ class Sidebar extends React.Component<Props, State> {
...
+        if (location.pathname === `/${SIDEBAR_VIEW_DOCGEN}`) {
+            history.push('/');
+        }
+    }
+
+    fetchMetadataSuccessCallback = ({ editors }: { editors: Array<MetadataEditor> }) => {
+        const { history } = this.props;
+        if (editors && editors.length > 0) {
+            const template = editors[0]?.template;
+            const scopeIsGlobal = template.scope === 'global';
```
bfoxx1906, line 120: If editors[0] is null or undefined won't template by undefined? If that's the case then unless you do template?.scope below you'll get an error right?

src/elements/content-sidebar/Sidebar.js
```diff
@@ -91,14 +96,68 @@ class Sidebar extends React.Component<Props, State> {
...
+                this.setState({ ...this.state, isDocGenTemplate });
+                // navigate to docgen tab by default
+                history.push(`/${SIDEBAR_VIEW_DOCGEN}`);
+            }
+        } else {
+            this.disableDocgenTab();
+        }
+    };
+
+    fetchMetadata(): void {
```
bfoxx1906, line 134: Ok so my biggest concern here is the overall approach. We're treating the docgen side bar panel different from the rest of them e.g. acitivty, versions, metadata etc as those are controlled by props passed into the component and not by the component iteself making an API call to determine whether or not the docgen panel should be shown or not. Is there a rreason why we decided to go this route instead of having the consumer of the sidebar pass something along the lines of hasDocGenTemplate to the ContentSidebar that is set based on the API calls you've added here?

src/elements/content-sidebar/DocGenSidebar/DocGenSidebar.tsx
```diff
@@ -0,0 +1,162 @@
...
+        },
+        jsonPaths: {
+            textTree: {},
+            imageTree: {},
+        },
+    });
+
+    const tagsToJsonPaths = (tags: DocGenTag[]) => {
+        /* eslint-disable  @typescript-eslint/no-explicit-any */
+        const jsonPathsMap: any = {};
```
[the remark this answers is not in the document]
rustam-e: forgot to update, nice catch

src/elements/content-sidebar/Sidebar.js
rustam-e: indeed we have tickets in place for getting this information from preview calls in monolith instead of making a separate call to metadata from FE. I used this approach here in order to confirm the functionality working and unblock the e2e testing work. I can move the logic into preview page FE but eventually we'll need to remove it for performance reasons there as well to avoid unnecessary calls to metadata api from frontend if we can leverage some call preview team is making (that is if it is making a call to metadata api) on file view.

But you're right it'd be better to move the call logic into consumer instead, will do

rustam-e: I removed the metadata fetching logic out of BUIE - now it's done in the client. Later, we'll remove the call to 'checkDocGenTemplate' completely

src/elements/content-sidebar/Sidebar.js
rustam-e: removed from this PR

src/elements/content-sidebar/SidebarNav.js
```diff
@@ -101,6 +105,15 @@ const SidebarNav = ({
                             <IconMetadataThick />
                         </SidebarNavButton>
                     )}
+                    {isDocGenTemplate && (
```
bfoxx1906, line 108: Did we need to add tests for this in SidebarNav.test.js

src/elements/content-sidebar/SidebarUtils.js
```diff
@@ -151,6 +152,8 @@ class SidebarUtils {
                 return <FormattedMessage {...messages.sidebarMetadataTitle} />;
             case SIDEBAR_VIEW_ACTIVITY:
                 return <FormattedMessage {...messages.sidebarActivityTitle} />;
+            case SIDEBAR_VIEW_DOCGEN:
```
bfoxx1906, line 155: Do we have tests for this in the SidebarUtils test file? In general let's make sure that we have tests for all new functionality we've added.

src/elements/content-sidebar/Sidebar.js
```diff
@@ -111,6 +123,19 @@ class Sidebar extends React.Component<Props, State> {
             this.setForcedByLocation();
             this.setState({ isDirty: true });
         }
+        // need to re-check if file is a docgen-template on file change
+        if (file.id !== prevFile.id && features?.docgen?.enabled && features?.docgen?.checkDocGenTemplate) {
```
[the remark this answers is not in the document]
rustam-e: once we get the information on whether the file is a docgen template from BE, we'll just pass the isDocGenTemplate prop as enabled - there'll be no need for anything other than `enabled` and `getDocGenTags` props, i.e. the checkDocGenTemplate and isDocGenTemplate props will be removed before we enable the split feature

if that is ok to keep in features then I'd leave it there, if you think it makes sense to pass getDocGenTags callback separately then I'll do that

src/elements/content-sidebar/DocGenSidebar/DocGenSidebar.tsx
```diff
@@ -0,0 +1,162 @@
...
+    };
+
+    React.useEffect(() => {
+        loadTags();
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, []);
+
+    const { hasError, tags, loading } = sidebarState;
+
+    const hasNoTags = tags.image.length + tags.text.length === 0;
```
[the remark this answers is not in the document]
rustam-e: double checked - indeed it can be undefined

src/elements/content-sidebar/DocGenSidebar/DocGenSidebar.tsx
```diff
@@ -0,0 +1,162 @@
...
+                        setSidebarState({ ...sidebarState, loading: false });
+                    }
+                })
+                .catch(() => setSidebarState({ ...sidebarState, loading: false, hasError: true }));
+        }
+    };
+
+    React.useEffect(() => {
+        loadTags();
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
rustam-e: no because we still get an infinite loop since loadTags resets the sidebarState which it depends on, which causes re-instantiation of loadTags which causes it to be called again.

src/elements/content-sidebar/DocGenSidebar/TagTree.tsx
```diff
@@ -0,0 +1,29 @@
...
+        return null;
+    }
+
+    return (
```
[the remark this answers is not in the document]
rustam-e: if it's an object, it should iterate recursively through it. If it's null - ts should protect use from null values at top level.

src/elements/content-sidebar/__tests__/Sidebar.test.js
```diff
@@ -19,7 +19,14 @@ describe('elements/content-sidebar/Sidebar', () => {
...
     };
 
-    const getWrapper = props => shallow(<Sidebar file={file} location={{ pathname: '/' }} {...props} />);
+    const features = {
+        docgen: {
+            enabled: false,
+        },
+    };
+
+    const getWrapper = props =>
```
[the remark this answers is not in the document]
rustam-e: done

src/elements/content-sidebar/SidebarNav.js
rustam-e: done

src/elements/content-sidebar/Sidebar.js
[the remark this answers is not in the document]
bfoxx1906: I'd recommend passing the separately. That way we're not breaking the paradigm.

src/elements/content-sidebar/DocGenSidebar/TagTree.tsx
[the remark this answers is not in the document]
bfoxx1906: So TS wont' actually protect you at runtime. Once the code gets transpiled it's pure JS and it doesn't know anything about types. Whenever I see recursion I always wanna make sure that we don't have any issues with not breaking out but based on what I see I think we should be fine even if data is null. It looks like it will just  throw an exception on Object.keys which will cause it to break out of the recursion.

bfoxx1906 (commented):
Left one comment about using features. Other than that it should be good.

src/elements/content-sidebar/Sidebar.js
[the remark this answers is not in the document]
rustam-e: addressed - moved all logic relating to docgen sidebar into a docGenSidebarProps props

src/elements/content-sidebar/Sidebar.js
```diff
@@ -111,8 +125,31 @@ class Sidebar extends React.Component<Props, State> {
...
+
+        this.handleDocgenTemplateOnUpdate(prevProps);
     }
 
+    handleDocgenTemplateOnUpdate = (prevProps: Props) => {
+        const { history, location, file, api, metadataSidebarProps, docGenSidebarProps } = this.props;
+        const { file: prevFile, docGenSidebarProps: prevDocGenSidebarProps }: Props = prevProps;
+        // need to re-check if file is a docgen-template on file change
+        if (file.id !== prevFile.id && docGenSidebarProps?.enabled && docGenSidebarProps?.checkDocGenTemplate) {
+            docGenSidebarProps.checkDocGenTemplate(api, file, metadataSidebarProps.isFeatureEnabled);
```
bfoxx1906, line 137: Should we have an await here? Is checkDocGenTemplate async?

bfoxx1906 (commented):
Left a few more comments.

src/elements/content-sidebar/Sidebar.js
rustam-e: no that's not necessary, checkDocGenTemplate will reset the isDocgenTemplate prop which will be responsible for deciding whether to display or not display the sidebar and whether to navigate to the docgen tab

src/elements/content-sidebar/DocGenSidebar/DocGenSidebar.tsx
```diff
@@ -0,0 +1,161 @@
...
+    }, []);
+
+    const { hasError, tags, loading } = sidebarState;
+
+    const hasNoTags = tags.image.length + tags.text.length === 0;
+
+    return (
+        <SidebarContent sidebarView={SIDEBAR_VIEW_METADATA} title={intl.formatMessage(messages.docgenTags)}>
+            <div className={classNames('docgen-sidebar', { center: hasNoTags || hasError || loading })}>
```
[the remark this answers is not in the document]
rustam-e: no because that breaks the underline in the header. Also its child with class 'bcs-scroll-content-wrapper' has position absolute which breaks styling for the padding.

src/elements/content-sidebar/DocGenSidebar/NoTagsIcon.tsx
```diff
@@ -0,0 +1,99 @@
+import * as React from 'react';
+import AccessibleSVG, { SVGProps } from '../../../components/accessible-svg/AccessibleSVG';
+
+const NoTagsIcon = ({ className = '', height = 140, width = 141 }: SVGProps) => (
```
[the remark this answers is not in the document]
rustam-e: i exported it from figma designs - I'll confirm with him to see if it's a shared icon in box or a custom case

src/elements/content-sidebar/DocGenSidebar/DocGenSidebar.tsx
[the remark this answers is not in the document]
rustam-e: so you want to have the component file in the folder above and all of the components in it in a folder of the same name?

src/elements/content-sidebar/Sidebar.js
```diff
@@ -96,6 +102,14 @@ class Sidebar extends React.Component<Props, State> {
         this.setForcedByLocation();
     }
 
+    componentDidMount() {
+        const { file, api, metadataSidebarProps, docGenSidebarProps }: Props = this.props;
+        // if docgen feature is enabled, load metadata to check whether file is a docgen template
+        if (docGenSidebarProps?.enabled && docGenSidebarProps?.checkDocGenTemplate) {
+            docGenSidebarProps.checkDocGenTemplate(api, file, metadataSidebarProps.isFeatureEnabled);
+        }
```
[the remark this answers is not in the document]
rustam-e: it uses calls metadata endpoint from EUA and sets whether a file is docgenTemplate

src/elements/content-sidebar/Sidebar.js
```diff
@@ -111,8 +125,31 @@ class Sidebar extends React.Component<Props, State> {
             this.setForcedByLocation();
             this.setState({ isDirty: true });
         }
+
+        this.handleDocgenTemplateOnUpdate(prevProps);
     }
 
+    handleDocgenTemplateOnUpdate = (prevProps: Props) => {
```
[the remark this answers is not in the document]
rustam-e: as far as I can tell the sidebar is open on load and does not depend on url params. Could you please elaborate on some edge cases you're considering? how they would be impacted?

src/elements/content-sidebar/Sidebar.js
```diff
@@ -111,8 +125,31 @@ class Sidebar extends React.Component<Props, State> {
...
+            docGenSidebarProps?.enabled &&
+            prevDocGenSidebarProps?.isDocGenTemplate !== docGenSidebarProps?.isDocGenTemplate
+        ) {
+            if (docGenSidebarProps.isDocGenTemplate) {
+                // navigate to docgen tab
+                history.push(`/${SIDEBAR_VIEW_DOCGEN}`);
+            } else if (location.pathname === `/${SIDEBAR_VIEW_DOCGEN}`) {
+                history.push('/');
+            }
+        }
```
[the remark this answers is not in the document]
rustam-e: this behavior is agreed upon with product. The DocGen view is the default only for DocGen templates. All other files are not affected.

src/elements/content-sidebar/Sidebar.js
```diff
@@ -22,20 +22,25 @@ import { withFeatureConsumer } from '../common/feature-checking';
...
 import type { MetadataEditor } from '../../common/types/metadata';
 import type { BoxItem, User } from '../../common/types/core';
 import type { Errors } from '../common/flowTypes';
+import { SIDEBAR_VIEW_DOCGEN } from '../../constants';
+import API from '../../api';
 
 type Props = {
     activitySidebarProps: ActivitySidebarProps,
     additionalTabs?: Array<AdditionalSidebarTab>,
+    api: API,
```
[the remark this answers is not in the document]
rustam-e: ContentSidebar initializes the api and passes it via APIContext.Provider

src/elements/content-sidebar/DocGenSidebar/DocGenSidebar.tsx
```diff
@@ -0,0 +1,161 @@
...
+            });
+        });
+
+        return jsonPathsMap;
+    };
+
+    const loadTags = () => {
+        if (getDocGenTags) {
+            setSidebarState({ ...sidebarState, loading: true });
+            getDocGenTags()
```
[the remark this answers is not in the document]
rustam-e: now it calls a BFF endpoint

src/elements/content-sidebar/DocGenSidebar/Error.tsx
```diff
@@ -0,0 +1,27 @@
...
+import messages from './messages';
+import RefreshIcon from './RefreshIcon';
+
+type Props = {
+    onClick: () => void;
+};
+
+const Error = ({ onClick }: Props) => (
+    <div className="bcs-DocGen-error-state">
+        <RefreshIcon className="bcs-DocGen-error-state--icon" />
```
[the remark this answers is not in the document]
rustam-e: imo `bcs-TagsError` is more general than `bcs-DocGen-error-state`

src/elements/content-sidebar/DocGenSidebar/TagTree.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+interface TagTreeProps {
+    data: JsonPathsMap;
+    level?: number;
+}
+
+const TagTree = ({ data, level = 0 }: TagTreeProps) => {
+    if (Array.isArray(data) || !data) {
+        return null;
+    }
```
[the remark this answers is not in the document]
rustam-e: leaf nodes used to be arrays but they no longer are, fixed

src/elements/content-sidebar/Sidebar.js
```diff
@@ -96,6 +101,14 @@ class Sidebar extends React.Component<Props, State> {
         this.setForcedByLocation();
     }
 
+    componentDidMount() {
+        const { file, api, metadataSidebarProps, docGenSidebarProps }: Props = this.props;
+        // if docgen feature is enabled, load metadata to check whether file is a docgen template
+        if (docGenSidebarProps.enabled) {
+            docGenSidebarProps.checkDocGenTemplate(api, file, metadataSidebarProps.isFeatureEnabled);
+        }
```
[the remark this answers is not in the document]
rustam-e: I didn't use the logic in ContentSidebar because it only fetches metadata in following case:
`        const canHaveMetadataSidebar = !isFeatureEnabled && SidebarUtils.canHaveMetadataSidebar(this.props);
`
which is confusing to me since it seems to be the opposite of metadata being enabled. So I didn't want to change the existing logic that would affect metadata tab (since I don't understand it) and created my own.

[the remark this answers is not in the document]
rustam-e: but if it is indeed a mistake then I can try updating that indeed - I raised it with the metadata team

src/elements/content-sidebar/ContentSidebar.js
```diff
@@ -104,6 +106,7 @@ class ContentSidebar extends React.Component<Props, State> {
         clientName: CLIENT_NAME_CONTENT_SIDEBAR,
         defaultView: '',
         detailsSidebarProps: {},
+        docGenSidebarProps: { enabled: false },
```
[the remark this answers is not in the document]
rustam-e: https://git.dev.box.net/Box/EndUserApp/pull/11066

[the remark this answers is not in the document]
rustam-e: in other words it's not enough for a docgen feature to be enabled to dispaly the sidebar tab - the file also needs to be a template

src/elements/content-sidebar/Sidebar.js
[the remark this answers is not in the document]
rustam-e: I discussed it with @bfoxx1906  and decided t o leave it as it is - the code in the current implementation will be removed once we retrieve information of whether a file is a DocGen template from backend - the logic in ContentSidebar is quite old and affecting state of that component and Metadata sidebar so I'd like to avoid modifying it since I'll have to revert back once the logic is no longer necessary.

bfoxx1906 (approved):
My concerns have been addressed and we confirmed that a lot of the docgen logic will eventually be moved to the backed api. Make sure @tjuanitas 's comments are addressed and we should be good to go.

rustam-e:
Screenshot 2024-03-27 at 18 17 47

tjuanitas (approved):

